PR #341, opened 2023-04-05
OpenAPI passthrough token auth provider

gitri-ms:
### Motivation and Context

The OpenAPI skill accepts a callback function as a parameter, to be used for adding auth info to HTTP requests before sending. In the simplest scenario, the caller would have an access token to add to the requests. This change provides a class/function that callers can use to pass a pre-acquired token to the HTTP request's auth header, and also serves as a reference for future OpenAPI auth providers.

### Description

Implement the TokenAuthenticationProvider class. The constructor accepts an access token, which is added to the HTTP request message provided when calling AuthenticateRequestAsync. The function signature matches the signature of the AuthenticateRequestAsyncCallback delegate type.

### Contribution Checklist


dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Authentication/TokenAuthenticationProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Microsoft.SemanticKernel.Skills.OpenAPI.Authentication;
+
+/// <summary>
+/// Stores an access token and uses it to authenticate HTTP requests.
+/// This class implements the <see cref="AuthenticateRequestAsyncCallback"/> delegate.
+/// </summary>
+public class TokenAuthenticationProvider
+{
+    private readonly string _accessToken;
```
adrianwyatt, line 15: Let's chat about securely storing tokens in memory.

adrianwyatt (changes requested):

adrianwyatt (approved):

PR #354, opened 2023-04-06
Mitigate support for OpenAPI 3.1

SergeyMenshykh:
### Motivation and Context

This PR adds support for OpenAPI 3.1 spec to OpenAPI skills functionality that may be useful in scenarios when OpenAPI 3.1 skills are imported directly from OpenAPI documents or indirectly through ChatGPT plugins.

### Description

To support the OpenAPI 3.1 specification, the OpenApiDocumentParser class performs the following actions:
- Adapts/converts YAML content to JSON for further processing using System.Text.Json API.
- Tries to find the 'openapi' property.
- Downgrade the property value to latest supported version - 3.0.1 if the version in the document is higher.
- Passes the updated JSON OpenAPI document to Microsoft.OpenAPI.NET library for parsing.

Other small changes introduced in this PR:
- Cancellation toke is added to all ImportChatGptPluginSkill* and ImportOpenApiSkill* methods.
- All the ImportChatGptPluginSkill* and ImportOpenApiSkill* methods are made asynchronous.
- ReadAsync method of OpenApiReader class is used Read one.

### Contribution Checklist


dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
```diff
@@ -65,7 +72,7 @@ public static class KernelOpenApiExtensions
                 throw new MissingManifestResourceException($"Unable to load OpenApi skill from url '{url}'.");
             }
 
-            return kernel.RegisterOpenApiSkill(stream, skillName, authCallback);
+            return await kernel.RegisterOpenApiSkillAsync(stream, skillName, authCallback, cancellationToken);
```
lemillermicrosoft, line 75: Curious if the await here has an impact, is it not better to just return the task when possible?

lemillermicrosoft (dismissed):
Overall LGTM.

dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
SergeyMenshykh: On the one hand, there are inefficiencies introduced by compiler that generates a state-machine code for async/await methods. On the other hand, there are a few pitfalls you need to make sure not fall into when skipping async/await keywords. I’ll follow whichever adopted approach, but I would be more inclined to use async/await to save my dev time by not thinking/analyzing consequences of omitting them, up to the point when it becomes an issue, if at all, and only in this case will think about optimization.

lemillermicrosoft: @shawncal in case we have a team preference.

shawncal: Where there's only one await and it's on the final statement, I prefer to remove async and the await, just return the Task directly.

shawncal (commented):
Just 2 comments:
 - async/await thing mentioned below
 - where we are adding awaits, please add start adding .ConfigureAwait(false) as a general good practice (we'll need to make this change in the whole kernel project at some point)

dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
lemillermicrosoft: Sounds good, so if there is other await in the method already and we can't remove the `async` then it's fine to `return await` . Thanks!

SergeyMenshykh:
> Just 2 comments:
> 
> * async/await thing mentioned below
> * where we are adding awaits, please add start adding .ConfigureAwait(false) as a general good practice (we'll need to make this change in the whole kernel project at some point)

Removed the awaits where it was possible and added ConfigureAwait(false) to all awaits in OpenAPI sills functionality.

dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -21,21 +26,88 @@ namespace Microsoft.SemanticKernel.Skills.OpenAPI.OpenApi;
 internal class OpenApiDocumentParser : IOpenApiDocumentParser
 {
     /// <inheritdoc/>
-    public IList<RestApiOperation> Parse(Stream stream)
+    public async Task<IList<RestApiOperation>> ParseAsync(Stream stream, CancellationToken cancellationToken = default)
     {
-        var openApiDocument = new OpenApiStreamReader().Read(stream, out var diagnostic);
+        var jsonObject = await this.DowngradeDocumentVersionToSuportedOneAsync(stream, cancellationToken).ConfigureAwait(false);
```
adrianwyatt, line 31: Could you also update the title of the PR to replace "Support" with "Mitigate" since we don't *technically* support true OpenAPI 3.1. Of course, let me know if I'm misunderstanding.

adrianwyatt (approved):

PR #364, opened 2023-04-07
Improve function availability and planning skill template

lemillermicrosoft:
### Motivation and Context

Today, planner can create different plans for the same goal and function input. This is undesirable, as it'd be best to have consistent plan generation as temperature is 0. The reason for these difference, is because the use of ConcurrentDictionary would result in non-deterministic function manual ordering. This change adds sorting to local manual retrieval to ensure consistent plan generation. Additionally, issues have been raised showing DV3 being unable to use available functions without additional whitespace in the prompt, so adding that as well.

### Description

      - Order the available functions by skill name and function name for better readability
      - Add blank lines to the planning skill template for better formatting
      - Add a pragma directive to suppress a warning for the namespace of the extension methods
      -[Test] Convert the result of GetAvailableFunctionsAsync to a list to avoid multiple enumerations

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/Planning/SKContextPlanningExtensions.cs
+++ b/dotnet/src/SemanticKernel/Planning/SKContextPlanningExtensions.cs
@@ -9,8 +9,10 @@ using Microsoft.SemanticKernel.Memory;
 using Microsoft.SemanticKernel.SkillDefinition;
 using static Microsoft.SemanticKernel.CoreSkills.PlannerSkill;
 
+#pragma warning disable IDE0130
 // ReSharper disable once CheckNamespace // Extension methods
 namespace Microsoft.SemanticKernel.Orchestration;
+#pragma warning restore IDE0130
 
 internal static class SKContextPlanningExtensions
 {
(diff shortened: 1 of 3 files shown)


dotnet/src/SemanticKernel/Planning/SKContextPlanningExtensions.cs
```diff
@@ -9,8 +9,10 @@
 using Microsoft.SemanticKernel.SkillDefinition;
 using static Microsoft.SemanticKernel.CoreSkills.PlannerSkill;
 
+#pragma warning disable IDE0130
 // ReSharper disable once CheckNamespace // Extension methods
 namespace Microsoft.SemanticKernel.Orchestration;
+#pragma warning restore IDE0130
```
RogerBarreto, line 15: Can you change the namespace for this one to `SemanticKernel.UnitTests.Planning` and remove the pragmas and ReSharper disable comments

lemillermicrosoft: No, we explicitly want to keep this in the same orchestration namespace to reduce what devs are required to import.

lemillermicrosoft: ((This was requested of me in previous reviews, just now adding the IDE pragma so I don't have to keep fighting it))

shawncal (approved):

PR #366, opened 2023-04-07
Mark example links with ticks to avoid markdown link checking.

adrianwyatt:
### Motivation and Context
The markdown link checker is failing with some links in the CopilotChat READMEs.

### Description
Make sure all example links, such as contoso and env files", and annotated with ticks to avoid the link checker trying to validate them.


samples/apps/copilot-chat-app/WebApp/README.md
```diff
@@ -25,13 +25,14 @@ The Copilot Chat sameple showcases how to build an enriched intelligent app, wit
 ## Running the sample
```
lemillermicrosoft, line 25: Up above ^^
from
>    - Select Single-page application (SPA) as platform type, and the Redirect URI will be http://localhost:3000
to
>    - Select Single-page application (SPA) as platform type, and the Redirect URI will be `http://localhost:3000`

lemillermicrosoft (approved):
Thanks Adrian!

PR #371, opened 2023-04-07
Simplifying state + removing unused components

teresaqhoang:
### Motivation and Context

Removing unused parts of app, including chatSlice from redux and filter icon in ChatList

### Description

Consolidated ChatState into ConversationsState, will manage state from conversations array.
Commenting out Filter icon under "Conversations" list until we're ready to implement
Changed login error to be more helpful to error cases we've observed

### Contribution Checklist


samples/apps/copilot-chat-app/WebApp/src/components/Login.tsx
```diff
@@ -20,8 +20,12 @@ export const Login: React.FC = () => {
 
     const handleError = (error: any) => {
         console.error(error);
-        setErrorMessage((error as Error).message);
-    }
+        setErrorMessage(
+            `Login failed. Check that you have a valid REACT_APP_CHAT_CLIENT_ID set in your .env file. See ${
```
craigomatic, line 24: Would it make sense to align this ID more explicitly with AAD? Maybe REACT_APP_AAD_CLIENT_ID?

If we do that here, it would be good to align the prior auth sample (in another PR) with the new id. We used REACT_APP_GRAPH_CLIENT_ID there.

craigomatic (dismissed):
Looks good, I do think we should consider making the constant in the .env file more meaningful re AAD, but that should not block this PR.

samples/apps/copilot-chat-app/WebApp/src/components/Login.tsx
teresaqhoang: I updated it to REACT_APP_AAD_CLIENT_ID for copilot-chat, but left REACT_APP_GRAPH_CLIENT_ID in the auth app sample since it's referenced in the video walk through and want to keep it consistent.

adrianwyatt (approved):

PR #380, opened 2023-04-10
Rename SKWebApi to CopilotChatApi

adrianwyatt:
### Motivation and Context
"SKWebApi" is not descriptive enough for intention of the application.

### Description
Renamed SKWebApi to CopilotChatApi everywhere.


dotnet/SK-dotnet.sln
glahaye, line 1: The goal for the backend is to be SK as a service, not just the backend behind the chat app.

The rename here actually goes away from that and won't sit well when we have SKaaS deployed internally to support various apps, and even less when users will be able to deploy the backend to Azure or a container to support their apps!

samples/apps/copilot-chat-app/WebApp/yarn.lock
craigomatic, line 1: It's a little strange to see many changes in this file, given everything else in the repo is a rename maybe this change isn't needed?

adrianwyatt: Not sure what happened here - re-copied from 'main'.

dotnet/SK-dotnet.sln
adrianwyatt: Let's keep it in the CopilotChat naming paradigm while it is located in samples/apps/copilot-chat-app. When we're ready to consider the webapi as sufficiently generalized, let's discuss moving it out of the copilot chat sample directory and consider renaming again then.

Issue #159, opened 2023-03-26
Improve Planner prompt to support invalid ASK


gogodjzhu:
It appears that the Microsoft.SemanticKernel.CoreSkills.SemanticFunctionConstants.FunctionFlowFunctionDefinition prompt always attempts to generate a plan for a given goal, even if there are no appropriate skills available to fulfill it. This can result in some absurd plans being generated.

To address this, there are two potential options:

1. Return an error message when there are no suitable skills available to fulfill the goal. This would help prevent the generation of ridiculous plans and provide feedback to the user that the goal cannot be accomplished with the current set of skills. 
2. Leave the plan verification task to the user. 

Thx.

evchaki:
Closing with PR now in main

evchaki:
There is still additional work to be done.  Opening this back up

PR #406, opened 2023-04-11
Improve Planner to Support invalid ASK - Issue 159

RogerBarreto:
### Motivation and Context

Resolve Related Issue: [Improve Planner prompt to support invalid ASK](https://github.com/microsoft/semantic-kernel/issues/159)

This change mitigate the potential of generated plans with non existing Skills / Functions not provided.

If you ask for a Plan with insufficient Skills to fulfill, it will fail with a reason why. `context.ErrorDescription`

### Description

This change makes an additional query to the LLM providing only the available skills and the Goal without mixing it with the plan rules. 

This provides clarity when asking for plans that cannot be created and avoid the potential of LLM wrong assumptions, hallucinations using the few shots on the planning creation prompt.

This changes adds also a post verification when the plan is created to ensure the functions returned by the plan are indeed present in the SkillCollection.

### Contribution Checklist


dotnet/src/SemanticKernel/CoreSkills/PlannerSkill.cs
```diff
@@ -242,19 +260,96 @@ public async Task<SKContext> CreatePlanAsync(string goal, SKContext context)
         context.Variables.Set("available_functions", relevantFunctionsManual);
         // TODO - consider adding the relevancy score for functions added to manual
 
-        var plan = await this._functionFlowFunction.InvokeAsync(context);
+        if (await this.VerifyCanCreatePlanForGoalAsync(goal, context))
```
shawncal, line 263: I don't think we should hit the model before creating a plan, just to ask if it can. Extra time, expense. This should be merged into the Planner prompt itself: "Create a plan. If you can't generate a plan, respond with..."

lemillermicrosoft: I think there are some other ways we can do this, let's connect.

MovGP0:
See also #523 which wants to remove the planner altogether

lemillermicrosoft:
> See also #523 which wants to remove the planner altogether

PlannerSkill - not Planner though :) But yeah, we will still want to do *something* to address this issue. Did not do in that PR to reduce scope of changes.

shawncal:
@RogerBarreto I think we can close this draft. Can you confirm?

PR #409, opened 2023-04-11
.NET Standard 2.0 Support (SK main project)

shawncal:
### Motivation and Context
SK currently targets .netstandard2.1, which makes it incompatible with some .net implementations, including .net Framework. Some partners require this compatibility.

### Description
Changing the main SK project to target netstandard2.0 for greatest reach. Other projects to follow.

Some .net APIs are not available, and have been replaced with equivalent extension methods, or require additional nugets for compatability.


dluc (approved):
Tested on MacOS: built nugets locally and installed on a .NET 7 app, ran a prompt, looking good. Tested tokenizer: ok.

kheiman61:
Does this mean official support for use of SK in .NET Framework scenarios may be coming? I ran into this limitation recently when trying to use SK with a .NET Framework project on a platform that is pegged at .NET 4.6.2 currently.

PR #414, opened 2023-04-11
Ref app speech input

amsacha:
### Motivation and Context
Provide a reference implementation for speech to text input for communicating with the AI model.

### Description
Added UI to the frontend (visible as a microphone symbol) that you can click on to provide speech input.
Hooked up the backend to communicate with the speech SDK via the Microsoft.CognitiveServices.Speech package.
The backend loads the subscription key and region to generate a authorization token. This token is requested by the frontend whenever speech input is needed. The token allows the front end to generate a speech config and use the speech sdk to process the speech input to text. This text input is then passed along to the app as usual.

### Contribution Checklist


samples/apps/copilot-chat-app/webapi/Controllers/SpeechController.cs
```diff
@@ -0,0 +1,55 @@
+﻿
+// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Configuration;
+using SemanticKernel.Service.Config;
+
+namespace SemanticKernel.Service.Controllers;
+
+public class TokenResponse
```
glahaye, line 10: All public entities should have /// comments

samples/apps/copilot-chat-app/webapi/Controllers/SpeechController.cs
```diff
@@ -0,0 +1,55 @@
...
+        this._configuration = configuration;
+    }
+
+    [Route("speechToken")]
+    [HttpGet]
+    public ActionResult<TokenResponse> Get()
+    {
+        AzureSpeechServiceConfig azureSpeechConfig = this._configuration.GetSection("AzureSpeechConfig").Get<AzureSpeechServiceConfig>();
+
+        if (true)
```
glahaye, line 32: Why do you keep this if (true) here?

samples/apps/copilot-chat-app/webapi/Controllers/SpeechController.cs
```diff
@@ -0,0 +1,55 @@
+﻿
```
glahaye, line 1: I don't feel too hot about adding a controller to give out tokens in the SK service.

Talk to Teresa; she's working on something where she can request tokens from the frontend for various services and then pass them to the backend via a new header.

Also, more broadly, this feels like something that should simply be gated by AAD scopes.

I understand we're getting diverging requests and opinions about what the backend service and the frontend app are supposed to be and do.  It's becoming evident that we (the whole services and app crew) need to clarify this before going much further.

I'm afraid that by trying to "simplify" the backend service, we're actually making things more complicated!

samples/apps/copilot-chat-app/WebApp/src/components/chat/ChatInput.tsx
```diff
@@ -59,6 +61,30 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
...
     const [value, setValue] = React.useState('');
     const [previousValue, setPreviousValue] = React.useState('');
+    const [recognizer, setRecognizer] = React.useState<speechSdk.SpeechRecognizer>();
+    const [isListening, setIsListening] = React.useState(false);
+    const speechService = useSKSpeechService(process.env.REACT_APP_BACKEND_URI  as string);
+
+    React.useEffect(() => {
+        if (recognizer) return;
+        void (async () => {
+            const newRecognizer = await speechService.getSpeechRecognizerAsync();;
```
SergeyMenshykh, line 71: nit: double semicolon.

glahaye (changes requested):
The backend service should not need to know about the services invoked by the frontend. Just use an AAD token to the speech service and there's no need to Modify the ASP.Net service

samples/apps/copilot-chat-app/webapi/Controllers/SpeechController.cs
amsacha: Resolved Offline; We need special admin privileges to make this possible, which isnt great for a reference implementation.

samples/apps/copilot-chat-app/README.md
```diff
@@ -69,6 +69,13 @@ and these components are functional:
               option is appropriate for your instance.    
             * You will insert your Azure endpoint key during build of the backend
               API Server
+         
+         * If you are using speech to text as input option then under the `AzureSpeechConfig` block, make sure the following configuration
```
glahaye, line 73: If you are using speech-to-text as an input option, then under the `AzureSpeechConfig` block, make sure the following configuration matches your instance

samples/apps/copilot-chat-app/webapi/Controllers/SpeechController.cs
```diff
@@ -0,0 +1,58 @@
...
+/// Token Response is a simple wrapper around the token and region
+/// </summary>
+public class TokenResponse
+{
+    public string? token { get; set; }
+    public string? region { get; set; }
+}
+
+[ApiController]
+public class TokenController : ControllerBase
```
glahaye, line 18: Please make class name and file name match.  Maybe SpeechTokenController??

samples/apps/copilot-chat-app/webapi/Controllers/SpeechController.cs
```diff
@@ -0,0 +1,58 @@
...
+    private readonly ILogger<TokenController> _logger;
+
+    public TokenController(IConfiguration configuration, ILogger<TokenController> logger)
+    {
+        this._configuration = configuration;
+        this._logger = logger;
+    }
+
+    /// <summary>
+    /// Use the Azure Speech Config key to return a authorization token and region as a Token Response.
```
glahaye, line 30: Nit: a authorization -> an authorization

glahaye (approved):

adrianwyatt:
> The backend service should not need to know about the services invoked by the frontend. Just use an AAD token to the speech service and there's no need to Modify the ASP.Net service

We'll be using the pattern of using a subscription key to generate an access token as the first version of this feature. We don't want to expose the subscription key in the FE, so we'll be using the BE as a token manager. This lights up the feature and gives us some room to optimize our auth story later in this sprint.

samples/apps/copilot-chat-app/WebApp/package-lock.json
adrianwyatt, line 1: Is this file supposed to be here?

samples/apps/copilot-chat-app/WebApp/yarn.lock
adrianwyatt, line 1: Is including this file intentional?

samples/apps/copilot-chat-app/webapi/Controllers/SpeechTokenController.cs
```diff
@@ -0,0 +1,58 @@
...
+    public SpeechTokenController(IConfiguration configuration, ILogger<SpeechTokenController> logger)
+    {
+        this._configuration = configuration;
+        this._logger = logger;
+    }
+
+    /// <summary>
+    /// Use the Azure Speech Config key to return an authorization token and region as a Token Response.
+    /// </summary>
+    [Route("speechToken")]
```
adrianwyatt, line 32: What the intended route here? I believe this wil lend up as http://localhost:3000/SpeechToken/speechToken yeah? I recommend renaming the controller to "TokenController" and having this route be "speech", so that would give us `http://localhost:3000/token/speech`.

samples/apps/copilot-chat-app/webapi/Controllers/SpeechTokenController.cs
```diff
@@ -0,0 +1,58 @@
...
+        string subscriptionKey = azureSpeechConfig.Key;
+
+        var token = this.FetchTokenAsync(FetchTokenUri, subscriptionKey).Result;
+
+        return new SpeechTokenResponse { token = token, region = azureSpeechConfig.Region };
+    }
+
+    private async Task<string> FetchTokenAsync(string fetchUri, string subscriptionKey)
+    {
+        using (var client = new HttpClient())
```
adrianwyatt, line 48: Please add a "TODO" for getting this client from the DI: "// TODO: get the HttpClient from the DI container"

samples/apps/copilot-chat-app/webapi/Controllers/SpeechTokenController.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            client.DefaultRequestHeaders.Add("Ocp-Apim-Subscription-Key", subscriptionKey);
+            UriBuilder uriBuilder = new UriBuilder(fetchUri);
+
+            var result = await client.PostAsync(uriBuilder.Uri, null);
```
adrianwyatt, line 53: Please check for a successful result, so we'll throw an exception on failure.
`result.EnsureSuccessStatusCode()`.

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/appsettings.json
```diff
@@ -48,6 +48,16 @@
     }
   },
 
+  "AzureSpeechConfig": {
+    "Label": "AzureSpeech",
+    "Region": "westus2", // Your Speech SDK subscription's region(e.g. westus2)
```
adrianwyatt, line 53: Since the region will always be specific to the user, I recommend removing the default value here.

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/README.md
```diff
@@ -69,6 +69,12 @@ and these components are functional:
               option is appropriate for your instance.    
             * You will insert your Azure endpoint key during build of the backend
               API Server
+         
+         * If you are using speech to text as input option then under the `AzureSpeechConfig` block, make sure the following configuration matches your instance:
+            * `"Region": "westus2",` or whichever region is appropriate
+              for your speech sdk instance.
+            * You will insert your Azure speech key during build of the backend
```
adrianwyatt, line 76: Is this true? Don't we recommend setting it as a user-secrets?

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Controllers/SpeechTokenController.cs
amsacha: No, the eventual path will be http://localhost:40443/speechToken, regardless of what the class is actually called the commandURL will be the baseURL_forApp + route

samples/apps/copilot-chat-app/WebApp/yarn.lock
amsacha: will remove

samples/apps/copilot-chat-app/WebApp/package-lock.json
amsacha: will remove

samples/apps/copilot-chat-app/README.md
amsacha: This is just copying the recommendations listed for the Embedding and Completion blocks, will update them

samples/apps/copilot-chat-app/webapi/Controllers/SpeechTokenController.cs
amsacha: No, the URL ends up being http://localhost:40443/speechToken, the class name isnt amended to the path, just the route itself.

PR #425, opened 2023-04-12
Add customizable metadata field to MemoryRecord and memory retrieval with embeddings is optional

awharrison-28:
### Motivation and Context
This PR addresses 3 issues. 

1. MemoryRecord now has a ValueString field - this is to support customizable metadata from the user. It is up to the user to serialize and deserialize this information. 
2. Retrieving the embedding from an IMemoryStore is optional and defaults to false. In the vast majority of cases, it is unnecessary and wasteful to pass the vector around the application, especially over a network. It is primarily used for similarity comparison which performed server-side for vector optimized data stores (ex: Qdrant).
3. Refactored SqliteMemoryStore so that connections are disposed better. There were multiple cases of residual connections causing intermittent, unexpected unit test behavior. 

### Description
- `IMemoryStore` `Get` and `GetNearest` calls have a new boolean parameter called `withEmbedding`. This indicates whether or not the response should include the vector associated with the memory. For all existing memory connectors, this defaults to false. This boolean is propagated to `ISemanticTextMemory` `Get` and `Search` calls as well.
- `MemoryRecord` has an optional field called `valueString` to allow the user to supply a custom metadata string. This is field is added to `ISemanticTextMemory` record creation as well. 
- Connectors.Sqlite.Memory `Database` is no longer a static class. The connection is supplied by the `SqliteMemoryStore` which also ensures its proper disposal.
- Added a few missing XML docs where necessary. 
- Added tests for `valueString` and `withEmbeddings`.
### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.Memory.CosmosDB/CosmosMemoryStore.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.CosmosDB/CosmosMemoryStore.cs
@@ -119,7 +119,7 @@ public class CosmosMemoryStore : IMemoryStore
     }
 
     /// <inheritdoc />
-    public async Task<MemoryRecord?> GetAsync(string collectionName, string key, CancellationToken cancel = default)
+    public async Task<MemoryRecord?> GetAsync(string collectionName, string key, bool withEmbedding = false, CancellationToken cancel = default)
     {
         var id = this.ToCosmosFriendlyId(key);
         var partitionKey = PartitionKey.None;
@@ -141,7 +141,7 @@ public class CosmosMemoryStore : IMemoryStore
         {
             var result = response.Resource;
 
-            var vector = System.Text.Json.JsonSerializer.Deserialize<float[]>(result.EmbeddingString);
+            float[]? vector = withEmbedding ? System.Text.Json.JsonSerializer.Deserialize<float[]>(result.EmbeddingString) : System.Array.Empty<float>();
 
             if (vector != null)
             {
@@ -157,12 +157,12 @@ public class CosmosMemoryStore : IMemoryStore
     }
 
     /// <inheritdoc/>
-    public async IAsyncEnumerable<MemoryRecord> GetBatchAsync(string collectionName, IEnumerable<string> keys,
+    public async IAsyncEnumerable<MemoryRecord> GetBatchAsync(string collectionName, IEnumerable<string> keys, bool withEmbeddings = false,
         [EnumeratorCancellation] CancellationToken cancel = default)
     {
         foreach (var key in keys)
         {
-            var record = await this.GetAsync(collectionName, key, cancel);
+            var record = await this.GetAsync(collectionName, key, withEmbeddings, cancel);
 
             if (record != null)
             {
@@ -241,6 +241,7 @@ public class CosmosMemoryStore : IMemoryStore
         Embedding<float> embedding,
         int limit,
         double minRelevanceScore = 0,
+        bool withEmbeddings = false,
         [EnumeratorCancellation] CancellationToken cancel = default)
     {
         {
@@ -252,15 +253,16 @@ public class CosmosMemoryStore : IMemoryStore
             var collectionMemories = new List<MemoryRecord>();
             TopNCollection<MemoryRecord> embeddings = new(limit);
 
-            await foreach (var entry in this.GetAllAsync(collectionName, cancel))
+            await foreach (var record in this.GetAllAsync(collectionName, cancel))
             {
-                if (entry != null)
+                if (record != null)
                 {
                     double similarity = embedding
                         .AsReadOnlySpan()
-                        .CosineSimilarity(entry.Embedding.AsReadOnlySpan());
+                        .CosineSimilarity(record.Embedding.AsReadOnlySpan());
                     if (similarity >= minRelevanceScore)
                     {
+                        var entry = withEmbeddings ? record : MemoryRecord.FromMetadata(record.Metadata, Embedding<float>.Empty, record.Key, record.Timestamp);
                         embeddings.Add(new(entry, similarity));
                     }
                 }
@@ -276,7 +278,7 @@ public class CosmosMemoryStore : IMemoryStore
     }
 
     /// <inheritdoc/>
-    public async Task<(MemoryRecord, double)?> GetNearestMatchAsync(string collectionName, Embedding<float> embedding, double minRelevanceScore = 0,
+    public async Task<(MemoryRecord, double)?> GetNearestMatchAsync(string collectionName, Embedding<float> embedding, double minRelevanceScore = 0, bool withEmbedding = false,
         CancellationToken cancel = default)
     {
         return await this.GetNearestMatchesAsync(
@@ -284,6 +286,7 @@ public class CosmosMemoryStore : IMemoryStore
             embedding: embedding,
             limit: 1,
             minRelevanceScore: minRelevanceScore,
+            withEmbeddings: withEmbedding,
             cancel: cancel).FirstOrDefaultAsync(cancellationToken: cancel);
     }
 
--- a/dotnet/src/SemanticKernel/Memory/ISemanticTextMemory.cs
+++ b/dotnet/src/SemanticKernel/Memory/ISemanticTextMemory.cs
@@ -20,7 +20,8 @@ public interface ISemanticTextMemory
     /// <param name="description">Optional description.</param>
     /// <param name="valueString">Optional value string for saving custom metadata.</param>
     /// <param name="cancel">Cancellation token.</param>
-    public Task SaveInformationAsync(
+    /// <returns>Unique identifier of the saved memory record.</returns>
+    public Task<string> SaveInformationAsync(
         string collection,
         string text,
         string id,
@@ -38,7 +39,8 @@ public interface ISemanticTextMemory
     /// <param name="description">Optional description.</param>
     /// <param name="valueString">Optional value string for saving custom metadata.</param>
     /// <param name="cancel">Cancellation token</param>
-    public Task SaveReferenceAsync(
+    /// <returns>Unique identifier of the saved memory record.</returns>
+    public Task<string> SaveReferenceAsync(
         string collection,
         string text,
         string externalId,
--- a/dotnet/src/SemanticKernel/Memory/ISemanticTextMemory.cs
+++ b/dotnet/src/SemanticKernel/Memory/ISemanticTextMemory.cs
@@ -14,26 +14,29 @@ public interface ISemanticTextMemory
     /// <summary>
     /// Save some information into the semantic memory, keeping a copy of the source information.
     /// </summary>
-    /// <param name="collection">Collection where to save the information</param>
-    /// <param name="id">Unique identifier</param>
-    /// <param name="text">Information to save</param>
-    /// <param name="description">Optional description</param>
-    /// <param name="cancel">Cancellation token</param>
+    /// <param name="collection">Collection where to save the information.</param>
+    /// <param name="id">Unique identifier.</param>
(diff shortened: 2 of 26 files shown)


shawncal (approved):

dotnet/src/SemanticKernel/Diagnostics/NullableAttributes.cs
```diff
@@ -1,6 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
```
shawncal, line 1: This one needs to stay. We should see if we can stop the tools from flagging it.

dotnet/src/SemanticKernel/Memory/ISemanticTextMemory.cs
```diff
@@ -14,54 +14,61 @@ public interface ISemanticTextMemory
...
+    /// <param name="description">Optional description.</param>
+    /// <param name="valueString">Optional value string for saving custom metadata.</param>
+    /// <param name="cancel">Cancellation token.</param>
+    /// <returns>Unique identifier of the saved memory record.</returns>
+    public Task<string> SaveInformationAsync(
         string collection,
         string text,
         string id,
         string? description = null,
+        string? valueString = null,
```
shawncal, line 29: valueString <== sounds to me like the "actual value" of the thing being stored. Perhaps metadataString? Or additionalContext?

Dare I ask: should this be a string or a Dict<string, string>?

dotnet/src/SemanticKernel/Memory/MemoryRecord.cs
```diff
@@ -110,28 +116,45 @@ public class MemoryRecord : DataEntryBase
...
-    /// <param name="timestamp">optional timestamp</param>
+    /// <param name="embedding">Optional embedding associated with a memory record.</param>
+    /// <param name="key">Optional existing database key.</param>
+    /// <param name="timestamp">optional timestamp.</param>
     /// <returns></returns>
     /// <exception cref="MemoryException"></exception>
     public static MemoryRecord FromJson(
         string json,
-        Embedding<float> embedding,
+        Embedding<float>? embedding,
```
shawncal, line 126: Shouldn't this also be the json-encoded (string) embedding? Or are we keeping those separate?

dotnet/src/SemanticKernel/Memory/MemoryRecord.cs
```diff
@@ -110,28 +116,45 @@ public class MemoryRecord : DataEntryBase
...
+    /// <param name="key">Optional existing database key.</param>
+    /// <param name="timestamp">optional timestamp.</param>
+    /// <returns></returns>
+    public static MemoryRecord FromMetadata(
+        MemoryRecordMetadata metadata,
+        Embedding<float>? embedding,
+        string? key = null,
+        DateTimeOffset? timestamp = null)
+    {
+        return new MemoryRecord(metadata, embedding ?? Embedding<float>.Empty, key, timestamp);
```
shawncal, line 155: Do we want empty embeddings in MemoryRecord? Or should this property be nullable too?

dotnet/src/SemanticKernel/Memory/MemoryRecordMetadata.cs
```diff
@@ -47,6 +47,13 @@ public class MemoryRecordMetadata : ICloneable
     [JsonPropertyName("text")]
     public string Text { get; }
 
+    /// <summary>
+    /// Field for saving custom metadata with a memory.
+    /// </summary>
+    [JsonInclude]
+    [JsonPropertyName("value_string")]
+    public string ValueString { get; }
```
shawncal, line 55: Same question: is this meant to be "the value" or just nice things we may want to know about for later? Should it be a string or a property bag (dict)?

samples/dotnet/kernel-syntax-examples/Example23_ReadOnlyMemoryStore.cs
```diff
@@ -76,12 +76,12 @@ public Task<bool> DoesCollectionExistAsync(string collectionName, CancellationTo
             throw new System.NotImplementedException();
         }
 
-        public Task<MemoryRecord?> GetAsync(string collectionName, string key, CancellationToken cancel = default)
+        public Task<MemoryRecord?> GetAsync(string collectionName, string key, bool withEmbedding = true, CancellationToken cancel = default)
```
dluc, line 79: why is the bool true by default?

dotnet/src/SemanticKernel/Memory/ISemanticTextMemory.cs
awharrison-28: Happy to rename it to customString, or additionalMetadata.

I do not want to make this a dictionary. The next best option I think is to make a dictionary<string, object> as we shouldn't limit the schema of the metadata. It should be up to the user whether they want to serialize an object, or have a  simple string, or nothing at all for this field. However, when you start adding dictionaries, that's additional steps to add fields, serialize, and deserialize individual entries, and there are no guarantees that they can be properly serialized/deserialized at the storage level.

dotnet/src/SemanticKernel/Memory/MemoryRecord.cs
awharrison-28: The embedding should be separate from the metadata - perhaps I should rename this to FromJsonMetadata. Constructing a memory record from existing memories should be embedding optional.

dotnet/src/SemanticKernel/Memory/MemoryRecordMetadata.cs
awharrison-28: Commented above. It's meant to support any additional metadata that the user may want to associate with the memory. I am strongly in favor of a string - placing the responsibility of serialization/deserialization of the custom fields on the application.

samples/dotnet/kernel-syntax-examples/Example23_ReadOnlyMemoryStore.cs
awharrison-28: It doesn't really make sense for this simple example to ever not return the embedding, it's just too basic. I can call that out in a comment.

dotnet/src/SemanticKernel/Memory/MemoryRecord.cs
awharrison-28: I found making the property nullable to be very tedious in terms of null checks EVERYWHERE. An empty embedding serves no purpose and cannot be used for similarity comparisons, so functionally it ensures correct behavior. From my perspective, psychologically, it gives me a better sense that the MemoryRecord was constructed with purpose from something existing as opposed to an accidentally missed field.

dotnet/src/SemanticKernel/Memory/ISemanticTextMemory.cs
awharrison-28: renamed to additionalMetadata

dotnet/src/SemanticKernel/Memory/MemoryRecordMetadata.cs
awharrison-28: renamed to additionalMetadata

samples/dotnet/kernel-syntax-examples/Example23_ReadOnlyMemoryStore.cs
awharrison-28: added comment explaining that the value isn't used and set = default

dotnet/src/SemanticKernel/Memory/MemoryRecord.cs
awharrison-28: renamed to FromJsonMetadata to make more clear

dotnet/src/SemanticKernel/Diagnostics/NullableAttributes.cs
awharrison-28: Brought it back, but it's keeping me from passing dotnet-format

dotnet/src/SemanticKernel/Planning/SKContextPlanningExtensions.cs
```diff
@@ -79,7 +79,7 @@ internal static class SKContextPlanningExtensions
             await RememberFunctionsAsync(context, availableFunctions);
 
             // Search for functions that match the semantic query.
-            var memories = context.Memory.SearchAsync(PlannerMemoryCollectionName, semanticQuery!, config.MaxRelevantFunctions, config.RelevancyThreshold.Value,
+            var memories = context.Memory.SearchAsync(PlannerMemoryCollectionName, semanticQuery!, config.MaxRelevantFunctions, config.RelevancyThreshold.Value, false,
```
adrianwyatt, line 82: Could we add parameter labels to this call for clarity?

dotnet/src/SemanticKernel/Planning/SKContextPlanningExtensions.cs
```diff
@@ -135,13 +135,13 @@ internal static async Task RememberFunctionsAsync(SKContext context, List<Functi
             var key = string.IsNullOrEmpty(function.Description) ? functionName : function.Description;
 
             // It'd be nice if there were a saveIfNotExists method on the memory interface
-            var memoryEntry = await context.Memory.GetAsync(PlannerMemoryCollectionName, key, context.CancellationToken);
+            var memoryEntry = await context.Memory.GetAsync(PlannerMemoryCollectionName, key, false, context.CancellationToken);
```
adrianwyatt, line 138: Could we add parameter labels to this call for clarity?

samples/dotnet/kernel-syntax-examples/Example23_ReadOnlyMemoryStore.cs
```diff
@@ -76,13 +76,15 @@ public Task<bool> DoesCollectionExistAsync(string collectionName, CancellationTo
             throw new System.NotImplementedException();
         }
 
-        public Task<MemoryRecord?> GetAsync(string collectionName, string key, CancellationToken cancel = default)
+        public Task<MemoryRecord?> GetAsync(string collectionName, string key, bool withEmbedding = default, CancellationToken cancel = default)
```
adrianwyatt, line 79: [nit] `bool withEmbedding = false`

adrianwyatt (changes requested):

samples/dotnet/kernel-syntax-examples/Example23_ReadOnlyMemoryStore.cs
awharrison-28: @adrianwyatt ah I got feedback earlier suggesting that I set this to default as booleans default to false. Happy to change it back to be more clear.

awharrison-28: done

dotnet/src/SemanticKernel/Planning/SKContextPlanningExtensions.cs
awharrison-28: done

dotnet/src/Connectors/Connectors.Memory.CosmosDB/CosmosMemoryStore.cs
```diff
@@ -119,7 +119,7 @@ public async Task DeleteCollectionAsync(string collectionName, CancellationToken
     }
 
     /// <inheritdoc />
-    public async Task<MemoryRecord?> GetAsync(string collectionName, string key, CancellationToken cancel = default)
+    public async Task<MemoryRecord?> GetAsync(string collectionName, string key, bool withEmbedding = default, CancellationToken cancel = default)
```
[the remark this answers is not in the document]
awharrison-28: yes! I thought I had already done this, but I clearly missed some. Done now

adrianwyatt (approved):

PR #429, opened 2023-04-12
Functionality to register RestApiOperationRunner in kernel config

SergeyMenshykh:
### Motivation and Context

This PR introduces a few changes:
1. An instance of the RestApiOperationRunner needs to be registered through kernel config so that it can be reused by any skill that uses RestApiOperation functionality.
2. RestApiOperationRunner doesn't create HttpClient instance anymore and requires a hosting app to provide the instance instead. SK SDK can't make assumptions about the way hosting apps create/manage HttpClient objects/instances - whether they use HttpClientFactory to create an instance of HttpClient instance per request, or an HttpClient instance is created using the new operator and stored as singleton, or maybe an instance per host?

These changes add more flexibility to SK SDK to support different scenarios a hosting app may require:
- Working with OpenAPI skills that are run against same host. In this case, the same set of Http handlers can be used for all the skills and only one HttpClient instance will be created.
- Working with OpenAPI skills that are run against different hosts. This case may require a set of Http handlers unique to each host and more than one insnance of HttpClient. 

### Description
* Two overloads of a new SetRestApiOperationRunner method are added to the KernalConfiguration class allowing a hosting app to supply either an instance of HttpClient or a callback that will be called when the RestApiOperationRunner sends Http requests. The callback can be used by the hosting app to route the Http requests to a dedicated HttpClient instance based on the host, header, url string parameter, etc...
* The RegisterRestApiFunction is changed to access the instance of the RestApiOperationRunner class from the Kernal config rather than creating the instance itself.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
+++ b/dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
@@ -33,7 +33,6 @@ public static class KernelOpenApiExtensions
     /// <param name="skillName">Skill name.</param>
     /// <param name="url">Url to in which to retrieve the OpenAPI definition.</param>
     /// <param name="httpClient">Optional HttpClient to use for the request.</param>
-    /// <param name="authCallback">Optional callback for adding auth data to the API requests.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>A list of all the semantic functions representing the skill.</returns>
     public static async Task<IDictionary<string, ISKFunction>> ImportOpenApiSkillFromUrlAsync(
@@ -41,7 +40,6 @@ public static class KernelOpenApiExtensions
         string skillName,
         Uri url,
         HttpClient? httpClient = null,
-        AuthenticateRequestAsyncCallback? authCallback = null,
         CancellationToken cancellationToken = default)
     {
         Verify.ValidSkillName(skillName);
@@ -74,7 +72,7 @@ public static class KernelOpenApiExtensions
                 throw new MissingManifestResourceException($"Unable to load OpenApi skill from url '{url}'.");
             }
 
-            return await kernel.RegisterOpenApiSkillAsync(stream, skillName, authCallback, cancellationToken).ConfigureAwait(false);
+            return await kernel.RegisterOpenApiSkillAsync(stream, skillName, cancellationToken).ConfigureAwait(false);
         }
         finally
         {
@@ -90,13 +88,11 @@ public static class KernelOpenApiExtensions
     /// </summary>
     /// <param name="kernel">Semantic Kernel instance.</param>
     /// <param name="skillName">Skill name.</param>
-    /// <param name="authCallback">Optional callback for adding auth data to the API requests.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>A list of all the semantic functions representing the skill.</returns>
     public static Task<IDictionary<string, ISKFunction>> ImportOpenApiSkillFromResourceAsync(
         this IKernel kernel,
         string skillName,
-        AuthenticateRequestAsyncCallback? authCallback = null,
         CancellationToken cancellationToken = default)
     {
         Verify.ValidSkillName(skillName);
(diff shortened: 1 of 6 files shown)


samples/dotnet/kernel-syntax-examples/Example21_ChatGPTPlugins.cs
```diff
@@ -38,7 +44,12 @@ private static async Task RunChatGptPluginAsync()
 
         //--------------- Example of using Klarna ChatGPT plugin ------------------------
 
-        //var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
+        //using var httpClient = new HttpClient();
```
dluc, line 47: remove commented code

dotnet/src/SemanticKernel/KernelConfig.cs
```diff
@@ -87,6 +91,11 @@ public sealed class KernelConfig
     /// <returns>IEnumerable of all embedding service Ids in the kernel configuration.</returns>
     public IEnumerable<string> AllTextEmbeddingGenerationServiceIds => this.TextEmbeddingGenerationServices.Keys;
 
+    /// <summary>
+    /// The REST API operation runner.
+    /// </summary>
+    internal IRestApiOperationRunner? RestApiOperationRunner { get; private set; }
```
dluc, line 97: this should be private and at the bottom of the file

dotnet/src/SemanticKernel/KernelConfig.cs
```diff
@@ -273,6 +282,35 @@ public KernelConfig SetDefaultTextEmbeddingGenerationService(string serviceId)
...
+    /// <summary>
+    /// Set REST API operation runner.
+    /// </summary>
+    /// <remarks>
+    /// This method should be used when the same handlers(logging, resiliency, etc), used by HttpClient, are applicable to all hosts REST API operations are run against.
+    /// </remarks>
+    /// <param name="httpClient">An instance of the HttpClient class.</param>
+    /// <param name="authCallback">Optional callback for adding auth data to the API requests.</param>
+    /// <returns>The updated kernel configuration.</returns>
+    public KernelConfig SetRestApiOperationRunner(HttpClient httpClient, AuthenticateRequestAsyncCallback? authCallback = null)
```
dluc, line 294: could you refactor this code to use the same approach used for AI services? instead of passing instances, which are expensive, register a factory.

dotnet/src/SemanticKernel/KernelConfig.cs
dluc, line 97: If we keep adding objects/factories into KernelConfig, we're adding more tech debt about dependency injection, increasing the work we'll have to do to clean up. Why not introduce a generic service collection in the kernel instead?

dluc (changes requested):

samples/dotnet/kernel-syntax-examples/Example21_ChatGPTPlugins.cs
SergeyMenshykh: It's not the code that is not required any more, it's an example commenting out which will give you full, 2E2 example of working with a ChatGPT plugin skill.

dluc: could you add a comment explaining what you just said here :-)

dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
```diff
@@ -237,24 +227,27 @@ public static class KernelOpenApiExtensions
...
         RestApiOperation operation,
-        AuthenticateRequestAsyncCallback? authCallback = null,
         CancellationToken cancellationToken = default)
     {
         var restOperationParameters = operation.GetParameters();
 
+        var runner = kernel.Config.RestApiOperationRunner;
+        if (runner == null)
+        {
+            var message = $"No {nameof(RestApiOperationRunner)} is register. Please use one of {nameof(KernelConfig)}.{nameof(KernelConfig.SetRestApiOperationRunner)} methods to register it.";
```
lemillermicrosoft, line 243: nit:
```suggestion
            var message = $"No {nameof(RestApiOperationRunner)} is registered. Please use one of {nameof(KernelConfig)}.{nameof(KernelConfig.SetRestApiOperationRunner)} methods to register one.";
```

SergeyMenshykh:
This PR is paused until a required infrastructure is in place to accommodate these changes so that they are aligned design wise with the other similar concepts.

PR #432, opened 2023-04-12
Copilot Chat Persistent Chat History

TaoChenOSU:
### Motivation and Context

The Copilot chat app currently doesn't support multiple isolated chat sessions and persist chat history once the Webapp is terminated.

We would like to support multiple chat sessions so that users can create chats for different topics. We would also like to have persistent chat history so that users can return to their chats any time after they close the App.

### Description

1. Create isolated chat sessions. 
2. Create models to save chat sessions and chat history.
3. Use repository pattern as a data access pattern for saving and retrieving chat sessions and chat history. The underlying storage for the data can be easily swapped (The implementation is currently using InMemory storage).
4. Have the front end automatically load chat history once the user is authenticated.

### Contribution Checklist


samples/apps/copilot-chat-app/WebApp/src/components/chat/ChatHistoryItem.tsx
```diff
@@ -97,14 +97,16 @@ export const ChatHistoryItem: React.FC<ChatHistoryItemProps> = (props) => {
             time;
     }
 
-    const isMe = message.sender === account?.homeAccountId;
-    const member = chat.getAudienceMemberForId(message.sender, selectedId, conversations[selectedId].audience);
+    // TODO: Ask Tao to use UserId (immutable through course of profile)
+    // standardize to one
+    const isMe = message.userName === account?.name || message.userId === account?.homeAccountId;
```
teresaqhoang, line 102: This should be limited to the userId since it's a unique GUID and immutable through the user profile lifespan

samples/apps/copilot-chat-app/WebApp/src/components/chat/chat-list/ChatList.tsx
```diff
@@ -51,10 +53,7 @@ export const ChatList: FC = () => {
                         Conversations
                     </Text>
                     <div>
-                        {/* TODO: Allow users to filter conversations by name
-                        <Button
-                        icon={<Filter20Regular />}
-                        appearance="transparent" /> */}
+                        <Button icon={<Filter20Regular />} appearance="transparent" />
```
teresaqhoang, line 56: This (and line 47) was removed in a previous PR since we don't have filter support yet, make sure your branch is updated with main

samples/apps/copilot-chat-app/WebApp/src/libs/useChat.ts
```diff
@@ -25,54 +31,151 @@ export const useChat = () => {
...
-            botProfilePicture: botProfilePictures.at(botProfilePictureIndex) ?? '/assets/bot-icon-1.png',
-        };
-        dispatch(addConversation(newChat));
-        dispatch(setSelectedConversation(name));
-        return name;
+            await sk.invokeAsync(ask, 'ChatMemorySkill', 'CreateChat').then(async (result: IAskResult) => {
+                const newChatId = result.value;
+                const initialBotMessage = getVariableValue(result.variables, 'initialBotMessage');
+
+                // TODO: hook up chat object once Tao makes change to return Chat object
```
teresaqhoang, line 74: Is this still required?

teresaqhoang (changes requested):

samples/apps/copilot-chat-app/WebApp/src/components/chat/ChatWindow.tsx
```diff
@@ -65,27 +76,43 @@ const useClasses = makeStyles({
...
+                } catch (e: any) {
+                    alert('[EditChat] Unable to retrieve chat. Details:\n' + e);
+                }
+            }
         }
         setIsEditing(!isEditing);
     };
 
     const onTitleChange = (_ev: React.ChangeEvent<HTMLInputElement>, data: InputOnChangeData) => {
+        console.log(data.value);
```
craigomatic, line 108: Might want to drop this, unless it helps the developer get familiar with something they are likely to want to change

samples/apps/copilot-chat-app/WebApp/src/components/chat/ChatHistoryItem.tsx
craigomatic: it might make sense to return a viewmodel with an additional property that has calculated a message.IsMe boolean server-side vs. client-side

samples/apps/copilot-chat-app/webapi/Skills/ChatMemory.cs
```diff
@@ -0,0 +1,136 @@
...
+{
+    /// <summary>
+    /// Chat ID that is persistent and unique.
+    /// </summary>
+    [JsonPropertyName("id")]
+    [JsonPropertyOrder(1)]
+    public string Id { get; set; }
+
+    /// <summary>
+    /// Chat ID that is persistent and unique.
```
craigomatic, line 23: Looks like a copy/paste error, Chat ID vs. User ID

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -177,135 +178,92 @@ public async Task<SKContext> ChatAsync(string message, SKContext context)
...
-    /// <param name="context">Contains the memory object.</param>
-    private async Task<MemoryQueryResult?> GetLatestMemoryAsync(SKContext context)
+    /// <param name="response"></param>
+    /// <param name="chatId">The chat ID</param>
+    private async Task SaveNewResponseAsync(string response, string chatId)
     {
-        var allMemories = this.GetAllMemoriesAsync(context);
-
-        return await allMemories.FirstOrDefaultAsync();
+        await this.SaveNewMessageAsync(response, "bot", "bot", chatId);
```
craigomatic, line 266: You could drop async from the method signature and just return this.SaveNewMessaegAsync

craigomatic (changes requested):
A few minor items to clean up, I think the larger one to take care of would be the user identity for the chat messages - managing this in the backend feels a little more sensible given you should have the bearer token that you can extract the given users' OID from (although I don't remember seeing this, so maybe you do not?).

samples/apps/copilot-chat-app/WebApp/src/libs/useChat.ts
TaoChenOSU: I think what we have now is good. Will remove the TODO.

samples/apps/copilot-chat-app/WebApp/src/components/chat/chat-list/ChatList.tsx
TaoChenOSU: Will rebase

samples/apps/copilot-chat-app/WebApp/src/components/chat/ChatWindow.tsx
TaoChenOSU: Good catch!

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
TaoChenOSU: This is no longer needed.

PR #455, opened 2023-04-14
Add ability to read outlook emails to EmailSkill

sw47:
### Motivation and Context

Today, there is no ability to get outlook email contents in SK.

Ability to get outlook email contents in SK.

Ability to get specific emails from one's outlook inbox and perform actions (i.e. additional skills)

### Description

Add the ability to read outlook email contents to SK Skill.
Currently, this PR is in Draft state - please see the comments for questions on design/intention.

I will add full UT once those questions have been resolved.

### Contribution Checklist


dotnet/src/SemanticKernel.Skills/Skills.MsGraph/EmailSkill.cs
```diff
@@ -79,4 +108,35 @@ public async Task SendEmailAsync(string content, SKContext context)
...
+
+        IUserMessagesCollectionPage emailsPage = await this._connector.GetMessagesAsync(
+            topClause: string.IsNullOrWhiteSpace(top) ? null : int.Parse(top),
+            skipClause: string.IsNullOrWhiteSpace(skip) ? null : int.Parse(skip),
+            filterClause: filter,
+            orderByClause: orderBy,
+            selectClause: select
+        );
+
+        // Serialize all email content either by their body or subject returned by the page and return as a string as skill output
```
sw47, line 139: I would like to return an object collection (Email) from the skill but the only examples I have seen are all string's returned by the Skill.  I'm not sure if that is by design the the contract for Skills can ultimately only return string to represent raw text. 

For example I see the OrgHierarchy connector: https://github.com/microsoft/semantic-kernel/blob/ed9f38f364ad04e27a11491e2eaa3829de4560d7/dotnet/src/SemanticKernel.Skills/Skills.MsGraph/Connectors/OrganizationHierarchyConnector.cs#L40 return a IEnumerable<string> but no corresponding Skill for that function: https://github.com/microsoft/semantic-kernel/blob/ed9f38f364ad04e27a11491e2eaa3829de4560d7/dotnet/src/SemanticKernel.Skills/Skills.MsGraph/OrganizationHierarchySkill.cs#L13

 
 
Would be great to get some clarity here.

@craigomatic for your input, please feel free to tag others.

samples/dotnet/graph-api-skills/Program.cs
```diff
@@ -179,6 +180,49 @@ public static async Task Main()
...
+        getEmailMessagesVariable.Set(EmailSkill.Parameters.Select, "subject,from,receivedDateTime");
+        getEmailMessagesVariable.Set(EmailSkill.Parameters.OrderBy, "receivedDateTime DESC");
+        getEmailMessagesVariable.Set(EmailSkill.Parameters.Top, "10");
+
+        // Parse returned emails
+        SKContext emailsResult = await sk.RunAsync(getEmailMessagesVariable, outlook["GetEmailMessagesAsync"]);
+        string[] emailsSubjectOrBodyContent = emailsResult.Result.Split('\0');
+        foreach (var email in emailsSubjectOrBodyContent)
+        {
+            logger.LogInformation($"Email Subject or Body: {email}");
```
sw47, line 196: My use case is for each email, run an additional classification skill on email attributes (e.g. email subject, email body, etc.) and then possibly either generate a auto-respond email or further actions (e.g. add reminder, etc.)

I think the consumer of this skill would want full access to a concrete Email object that encapsulates the aforementioned attributes but I understand/think the intention SKContext to to generically encapsulate the info I need in the most generic way (i.e. string) so I'm not sure what is the best way to go about my problem. Would appreciate some tips :) @craigomatic and other folks

MovGP0:
I recommend using the `int.TryParse()` methods and consider the locale settings of the user using the `CultureInfo` class. The vast majority of people are not from the U.S.A.

adrianwyatt:
https://github.com/microsoft/semantic-kernel/pull/801

PR #458, opened 2023-04-15
Allow to use chat models for semantic functions

dluc:
### Motivation and Context

Chat models are (currently) less expensive and can be used like text completion models also for semantic functions and prompt completion.

### Description

Extend OpenAI chat clients to simulate text completion over the chat protocol. This behavior is enabled by default, but can also be disabled passing a boolean parameter when registering an AI service.

### DIFF
--- a/samples/dotnet/kernel-syntax-examples/Example26_SemanticFunctionsUsingChatGPT.cs
+++ b/samples/dotnet/kernel-syntax-examples/Example26_SemanticFunctionsUsingChatGPT.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel;
+using RepoUtils;
+
+/**
+ * This example shows how to use GPT3.5 Chat model for prompts and semantic functions.
+ */
+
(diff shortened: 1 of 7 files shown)


shawncal (approved):

samples/dotnet/kernel-syntax-examples/Example26_SemanticFunctionsUsingChatGPT.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel;
+using RepoUtils;
+
+/**
+ * This example shows how to use GPT3.5 Chat model for prompts and semantic functions.
```
NTaylorMullen, line 9: GPT-4 + all future models are currently planned to only be chat-completions too aren't they?

dluc: I wouldn't take that for granted. Currently there's a lot of attention for chat scenarios, and I wouldn't be surprised if this shifts back to text once explorations move from UI to scalability, e.g. complex semantic memory

PR #460, opened 2023-04-15
Update READMEs and notebooks with latest nugets

dluc:
Update READMEs and notebooks with latest nugets


dluc:
I will update with the exact versions after the release

shawncal (dismissed):

PR #464, opened 2023-04-16
Introduce new meta package to install core packages

dluc:
Change **Microsoft.SemanticKernel** to be a metapackage that installs 2 packages:
* Microsoft.SemanticKernel.Core: core code (previously named Microsoft.SemanticKernel) + abstractions
* Microsoft.SemanticKernel.Connectors.AI.OpenAI: OpenAI clients, depending on Azure OpenAI SDK and Azure Core

Benefits:
* Users installing Microsoft.SemanticKernel have all the code like before to use OpenAI
* Users can install the core kernel without OpenAI if they need to, e.g. when working with other AI providers
* We can use the metapackage to include more useful dependencies if needed, without having to put all in the same package

### DIFF
--- a/dotnet/src/Connectors/Connectors.Memory.Qdrant/Connectors.Memory.Qdrant.csproj
+++ b/dotnet/src/Connectors/Connectors.Memory.Qdrant/Connectors.Memory.Qdrant.csproj
@@ -14,6 +14,7 @@
       <!-- NuGet Package Settings -->
         <PackageId>Microsoft.SemanticKernel.Connectors.Memory.Qdrant</PackageId>
         <Title>Semantic Kernel - Qdrant Connector</Title>
+        <Description>Qdrant connector for Semantic Kernel skills and semantic memory</Description>
     </PropertyGroup>
 
     <ItemGroup>
--- a/dotnet/src/SemanticKernel.MetaPackage/SemanticKernel.MetaPackage.csproj
+++ b/dotnet/src/SemanticKernel.MetaPackage/SemanticKernel.MetaPackage.csproj
@@ -1,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RepoRoot>$([System.IO.Path]::GetDirectoryName($([MSBuild]::GetPathOfFileAbove('.gitignore', '$(MSBuildThisFileDirectory)'))))</RepoRoot>
+  </PropertyGroup>
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.SemanticKernel</AssemblyName>
+    <RootNamespace>Microsoft.SemanticKernel</RootNamespace>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- NuGet Package Settings -->
+    <PackageId>Microsoft.SemanticKernel</PackageId>
+    <Title>Semantic Kernel</Title>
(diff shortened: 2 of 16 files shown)


dluc:
Note: for some reason "content files" of implicitly referenced packages (e.g. OpenAI connector) are not copied into the bin folder, breaking the GPT3 tokenizer.

dotnet/src/Connectors/Connectors.Memory.Qdrant/Connectors.Memory.Qdrant.csproj
```diff
@@ -14,6 +14,7 @@
       <!-- NuGet Package Settings -->
         <PackageId>Microsoft.SemanticKernel.Connectors.Memory.Qdrant</PackageId>
         <Title>Semantic Kernel - Qdrant Connector</Title>
+      <Description>Qdrant connector for Semantic Kernel skills and semantic memory</Description>
```
shawncal, line 17: Funny tabbing

dotnet/src/SemanticKernel.MetaPackage/SemanticKernel.MetaPackage.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RepoRoot>$([System.IO.Path]::GetDirectoryName($([MSBuild]::GetPathOfFileAbove('.gitignore', '$(MSBuildThisFileDirectory)'))))</RepoRoot>
+  </PropertyGroup>
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.SemanticKernel.MetaPackage</AssemblyName>
+    <RootNamespace>Microsoft.SemanticKernel.MetaPackage</RootNamespace>
```
shawncal, line 10: I don't think we actually call this "MetaPackage" (here or in file paths)... shouldn't this just be "Microsoft.SemanticKernel" and the other be under SK.Core?

dotnet/src/SemanticKernel/SemanticKernel.csproj
```diff
@@ -6,7 +6,7 @@
   <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
 
   <PropertyGroup>
-    <AssemblyName>Microsoft.SemanticKernel</AssemblyName>
+    <AssemblyName>Microsoft.SemanticKernel.Core</AssemblyName>
```
shawncal, line 9: I think all of this should move to SemanticKernel.Core/ folder and rename the csproj too. These should match the nugets.

dluc: can we move the files separately, to keep the PR small?

dotnet/src/Connectors/Connectors.Memory.Qdrant/Connectors.Memory.Qdrant.csproj
dluc: fixed

dotnet/src/SemanticKernel.MetaPackage/SemanticKernel.MetaPackage.csproj
dluc: checking... last time I tried the build was failing...

dluc: done

dotnet/src/SemanticKernel/SemanticKernel.csproj
shawncal: yup, sounds good.

shawncal (approved):

PR #485, opened 2023-04-17
Copilot chat: Extract semantic memory

TaoChenOSU:
### Motivation and Context

We need to demonstrate the ability of the Copilot chat app to extract and retrieve semantic contexts that are generated from chat history and save as embeddings.

### Description

1. Extract semantic memories.
2. Retrieve semantic memories.

### Contribution Checklist


adrianwyatt (approved):

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -94,15 +104,29 @@ public async Task<string> ExtractUserMemoriesAsync(SKContext context)
...
 
+        // Search for relevant memories.
+        List<MemoryQueryResult> relevantMemories = new List<MemoryQueryResult>();
+        foreach (var memoryName in SystemPromptDefaults.MemoryMap.Keys)
+        {
+            var results = context.Memory.SearchAsync(
+                SemanticMemoryExtractor.MemoryCollectionName(chatId, memoryName),
+                latestMessage.ToString(),
+                limit: 1000,
+                minRelevanceScore: 0.8);
```
glahaye, line 115: Should that 0.8 be (eventually, not necessarily today) configurable?

TaoChenOSU: Yes, eventually, I think we are thinking about moving the prompt configurations to a config file too. We can make this param configurable when we do that. 

We should be fine for now.

adrianwyatt (approved):

PR #491, opened 2023-04-17
Fix planner RememberFunctions method

awharrison-28:
### Motivation and Context
This PR fixes the text that gets embedded for remembering planner functions and clarifies the parameters being passed to SaveInformationAsync()

### Description
 - Add a new method to generate a string for embedding a function, based on its name and description, and use it instead of the manual string.
- text to embed is the function embedding string
- Use the function's fully qualified name as the key for saving it to memory, and use the function's description as the description for the memory entry, to avoid ambiguity and duplication.
- Revert default planner prompt to non-conditional support, include configuration to enable conditional support.
- Update test configuration to be a better example of how to handle relevancy.

### Contribution Checklist


dotnet/src/SemanticKernel/Planning/SKContextPlanningExtensions.cs
```diff
@@ -132,15 +132,17 @@ internal static async Task RememberFunctionsAsync(SKContext context, List<Functi
         foreach (var function in availableFunctions)
         {
             var functionName = function.ToFullyQualifiedName();
-            var key = string.IsNullOrEmpty(function.Description) ? functionName : function.Description;
+            var key = functionName;
+            var description = string.IsNullOrEmpty(function.Description) ? functionName : function.Description;
+            var textToEmbed = function.ToManualString();
```
lemillermicrosoft, line 137: Thanks Abby! I think this is fine for now, although I'm locally considering whether the ManualString is good for this purpose or if just the FunctionName+Description would be better -- I can iterate on this separately, just saying it out loud for those following along. Also considering whether we should include skill name in the key to avoid collisions.

lemillermicrosoft (approved):

lemillermicrosoft (changes requested):
Actually, can you remove the `skip` from the one test in dotnet\src\SemanticKernel.IntegrationTests\CoreSkills\PlannerSkillTests.cs

dotnet/src/SemanticKernel/CoreSkills/SemanticFunctionConstants.cs
```diff
@@ -61,49 +77,54 @@ internal static class SemanticFunctionConstants
...
+    <else>
         <function.LanguageHelpers.TranslateTo input=""$SUMMARIZED_INPUT"" translate_to_language=""Japanese"" setContextVariable=""TRANSLATED_TEXT"" />
         <function.EmailConnector.LookupContactEmail input=""Martin"" setContextVariable=""CONTACT_RESULT"" />
         <function.EmailConnector.EmailTo input=""$TRANSLATED_TEXT"" recipient=""$CONTACT_RESULT""/>
     </else>
   </if>
 </plan><!-- END -->
+[END EXAMPLE PLANS]
+
+Create an XML plan step by step, to satisfy the goal given.
```
lemillermicrosoft, line 105: @RogerBarreto I'd like to update these rules a bit for the short term to try and get back to a stable place. Let me know if you have thoughts/concerns.

lemillermicrosoft (approved):

dotnet/src/SemanticKernel.IntegrationTests/CoreSkills/PlannerSkillTests.cs
```diff
@@ -167,7 +169,8 @@ public async Task CreatePlanShouldHaveIfElseConditionalStatementsAndBeAbleToExec
     }
 
     [Theory]
-    [InlineData("Start with a X number equals to the current minutes of the clock and remove 20 from this number until it becomes 0. After that tell me a math style joke where the input is X number + \"bananas\"",
+    [InlineData(
```
RogerBarreto, line 172: I would add a test with the conditional set to false to see if we get a plan without <if> tags using the same goal.

RogerBarreto (approved):
LGTM

dotnet/src/SemanticKernel.IntegrationTests/CoreSkills/PlannerSkillTests.cs
lemillermicrosoft: added.

PR #499, opened 2023-04-18
Update CopilotChat to use only Chat APIs for completions, by default.

adrianwyatt:
### Motivation and Context
Chat completion (gpt-3.5-turbo/gpt-4) models are the future, let's optimize for those.

### Description
Updated CopilotChat to use Chat completions instead of Text completions.


glahaye (approved):
LGTM

PR #507, opened 2023-04-18
Copilot chat: Interactively ask permissions

TaoChenOSU:
### Motivation and Context

The Copilot Chat App will enter an error state when a user doesn't not have permissions to the scopes the app requests.

### Description

1. Interactively ask users permissions if users don't have the requested once when acquiring tokens.
2. Prevent the app from entering the Chat state if errors occur in LoadingChats state.

### Contribution Checklist

### DIFF
--- a/samples/apps/copilot-chat-app/WebApp/src/App.tsx
+++ b/samples/apps/copilot-chat-app/WebApp/src/App.tsx
@@ -60,9 +60,13 @@ const App: FC = () => {
     useEffect(() => {
         if (isAuthenticated && account && appState === AppState.LoadingChats) {            
             // Load all chats from memory
-            chat.loadChats().then(() => {
-                setAppState(AppState.Chat);
-            });
+            async function loadChats() {
+                if (await chat.loadChats()) {
+                    setAppState(AppState.Chat);
+                }
+            }
+
+            loadChats();
         }
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [instance, inProgress, isAuthenticated, appState]);
--- a/samples/apps/copilot-chat-app/WebApp/src/libs/auth/AuthHelper.ts
+++ b/samples/apps/copilot-chat-app/WebApp/src/libs/auth/AuthHelper.ts
@@ -85,7 +85,7 @@ const getSKaaSAccessToken = async (instance: IPublicClientApplication) => {
             return token.accessToken;
         }
     }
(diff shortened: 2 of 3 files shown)


samples/apps/copilot-chat-app/WebApp/src/libs/auth/AuthHelper.ts
```diff
@@ -75,9 +76,16 @@ const logoutAsync = async (instance: IPublicClientApplication) => {
...
+        const token = await instance.acquireTokenSilent({ scopes: Constants.msal.skScopes, account: instance.getAllAccounts()[0] });
         return token.accessToken;
-    });
+    } catch (ex: any) {
+        if (ex instanceof InteractionRequiredAuthError) {
+            const token = await instance.acquireTokenPopup({ scopes: Constants.msal.skScopes, account: instance.getAllAccounts()[0] });
+            return token.accessToken;
+        }
+    }
+    throw new Error('Unable to get SKaaS access token');
```
adrianwyatt, line 88: If this message is seen by the user, let's make it a little more friendly: "Failed to get access token for web service."

adrianwyatt (changes requested):

glahaye (dismissed):

samples/apps/copilot-chat-app/WebApp/src/App.tsx
```diff
@@ -60,9 +60,13 @@ const App: FC = () => {
     useEffect(() => {
         if (isAuthenticated && account && appState === AppState.LoadingChats) {            
             // Load all chats from memory
-            chat.loadChats().then(() => {
-                setAppState(AppState.Chat);
-            });
+            async function loadChats() {
+                if (await chat.loadChats()) {
```
Vivihung, line 64: Ideally, we should addAlert() to set the alert state of the app so the UI will render error banners to show users failure on critical path. But I'm okay if we want to address it next time considering the bug is painful to users.

Vivihung (dismissed):

samples/apps/copilot-chat-app/WebApp/src/App.tsx
TaoChenOSU: The alert state is populated inside the loadChats call.

glahaye (approved):

adrianwyatt (approved):

PR #511, opened 2023-04-18
Python: Use overloads for run_async method

awharrison-28:
### Motivation and Context
To make the Python SK more pythonic, combined all kernel.run_* methods into a single method with optional parameters. 

### Description
- Removed `kernel.run_on_vars_async()`
- Removed `kernel.run_on_str_async()`
- Added `SKContext`, `ContextVariables`, and string params to `kernel.run_async()`
- Moved `run_on_vars_async()` logic to `run_async()` method
- Arranged method params so that required params come before optional params. 
- Updated tests and notebooks where applicable. 

### Contribution Checklist


alexchaomander (approved):
Definitely makes it more Pythonic! LGTM!

python/semantic_kernel/kernel.py
```diff
@@ -106,21 +106,34 @@ def register_semantic_function(
...
+            input_str: Optional[str] = None,
     ) -> SKContext:
-        context = SKContext(
-            input_vars,
-            self._memory,
-            self._skill_collection.read_only_skill_collection,
-            self._log,
-        )
+        
+        if input_context is None:
```
dluc, line 117: I think we should handle the case where all of the 3 are not None
* copy input_str into input_vars
* copy input_vars into input_context
* use context

awharrison-28: right now in this case, I'm just logging warnings. Do you think it makes sense to overwrite the vars of the passed context?

awharrison-28: Updated to overwrite existing context if input_str or input_vars are not none

dluc (approved):

PR #516, opened 2023-04-19
Add error propagation for invalid completion configuration

adrianwyatt:
### Motivation and Context
When the Completion config is invalid, we end up repeating the user's message back to them instead of showing an error.

### Description
- Added propagation of a failed SK invoke back to the top-level controller call.
- Enabled AI exception detail messages

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
+++ b/samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
@@ -85,15 +85,9 @@ public class SemanticKernelController : ControllerBase
         SKContext result = await kernel.RunAsync(contextVariables, function!);
         if (result.ErrorOccurred)
         {
-            // TODO latest NuGets don't have the Detail property on AIException
-            //if (result.LastException is AIException aiException && aiException.Detail is not null)
-            //{
-            //    return this.BadRequest(string.Concat(aiException.Message, " - Detail: " + aiException.Detail));
-            //}
-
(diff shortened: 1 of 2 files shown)


glahaye (approved):

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -85,15 +85,9 @@ public SemanticKernelController(IServiceProvider serviceProvider, IConfiguration
...
-            //if (result.LastException is AIException aiException && aiException.Detail is not null)
-            //{
-            //    return this.BadRequest(string.Concat(aiException.Message, " - Detail: " + aiException.Detail));
-            //}
-
-            if (result.LastException is AIException aiException)
+            if (result.LastException is AIException aiException && aiException.Detail is not null)
             {
-                return this.BadRequest(aiException.Message);
+                return this.BadRequest(string.Concat(aiException.Message, " - Detail: " + aiException.Detail));
```
hathind-ms, line 90: I think exception message is enough, exception details have redundant message and adds noise. We can say "check api logs for more details".

adrianwyatt: Totally agree, but in this case we don't get the source of the problem without displaying the details. We only get Something like "BadRequest => blah blah 404". With the details it shows that we have misconfigured the connection to Completion.

hathind-ms (approved):

PR #519, opened 2023-04-19
Fix Microsoft.SemanticKernel.Core xml doc file name

stephentoub:
The current nuget package lacks IntelliSense descriptions for APIs because the documentation xml file's name doesn't match the corresponding assembly name.  This is just fixing it to match.


dluc (approved):

dluc:
@shawncal FYI

PR #520, opened 2023-04-19
Vectorize CosineSimilarityImplementation with Vector<T>

stephentoub:
### Motivation and Context

Accelerate the CosineSimilarityImplementation for floats and doubles using SIMD.

### Description

Employ `Vector<T>` to accelerate CosineSimilarityImplementation.  If the hardware supports acceleration, a vector's worth of data is processed at a time until there isn't a vector's worth remaining.  Then it falls back to the scalar path that's also used if hardware acceleration isn't available.  It's a fairly simplistic implementation that maintains a clean parallel with the scalar code and achieves a reasonable speedup.  As with anything that impacts the order in which floating-point operations are performed, this can impact the precise resulting answer.

### Contribution Checklist

On my machine:

```C#
private float[] _values1, _values2;

[Params(10, 100, 1_000, 10_000)]

[Params(false, true)]
public bool Vectorized { get; set; }

[GlobalSetup]
public void Setup()
{
    _values1 = Enumerable.Range(0, Length).Select(_ => Random.Shared.NextSingle()).ToArray();
    _values2 = Enumerable.Range(0, Length).Select(_ => Random.Shared.NextSingle()).ToArray();
}

[Benchmark] public double CosineSimilarity() => CosineSimilarityImplementation(_values1, _values2, Vectorized);
```

|           Method | Length | Vectorized |         Mean |
|----------------- |------- |----------- |-------------:|
| CosineSimilarity |     10 |      False |     17.80 ns |
| CosineSimilarity |     10 |       True |     12.97 ns |
| CosineSimilarity |    100 |      False |    117.07 ns |
| CosineSimilarity |    100 |       True |     52.06 ns |
| CosineSimilarity |   1000 |      False |  1,078.35 ns |
| CosineSimilarity |   1000 |       True |    384.26 ns |
| CosineSimilarity |  10000 |      False | 10,702.32 ns |
| CosineSimilarity |  10000 |       True |  3,799.04 ns |

cc: @tannergooding

### DIFF
--- a/dotnet/src/SemanticKernel/AI/Embeddings/VectorOperations/CosineSimilarityOperation.cs
+++ b/dotnet/src/SemanticKernel/AI/Embeddings/VectorOperations/CosineSimilarityOperation.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
+using System.Numerics;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.SemanticKernel.AI.Embeddings.VectorOperations;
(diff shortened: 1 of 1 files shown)


umeshma (approved):
Nice work Stephen! Thank you!

PR #528, opened 2023-04-19
Vectorize DivideByInPlaceImplementation with Vector<T>

stephentoub:
### Motivation and Context

Accelerate the DivideByInPlaceImplementation for floats and doubles using SIMD.

### Description

Employ `Vector<T>` to accelerate DivideByInPlaceImplementation. If the hardware supports acceleration, a vector's worth of data is processed at a time until there isn't a vector's worth remaining. Then it falls back to the scalar path that's also used if hardware acceleration isn't available.

### Contribution Checklist

On my machine:
```C#
private float[] _values, _copy;
private float _divisor;

[Params(10, 100, 1_000, 10_000)]

[Params(false, true)]
public bool Vectorized { get; set; }

[GlobalSetup]
public void Setup()
{
    _copy = Enumerable.Range(0, Length).Select(_ => Random.Shared.NextSingle()).ToArray();
    _values = new float[_copy.Length];
    _divisor = Random.Shared.NextSingle();
}

[Benchmark]
public void DivideByInPlace()
{
    Array.Copy(_copy, _values, _copy.Length);
    DivideByInPlaceImplementation(_values, _divisor, Vectorized);
}
```

|          Method | Length | Vectorized |        Mean |
|---------------- |------- |----------- |------------:|
| DivideByInPlace |     10 |      False |    11.29 ns |
| DivideByInPlace |     10 |       True |    10.96 ns |
| DivideByInPlace |    100 |      False |    76.45 ns |
| DivideByInPlace |    100 |       True |    20.25 ns |
| DivideByInPlace |   1000 |      False |   743.17 ns |
| DivideByInPlace |   1000 |       True |   198.14 ns |
| DivideByInPlace |  10000 |      False | 7,922.85 ns |
| DivideByInPlace |  10000 |       True | 2,345.91 ns |

cc: @tannergooding


umeshma (approved):
Pointer math looks ok this evening. Nice Stephen!

Issue #105, opened 2023-03-18
Important Error Message from Azure OpenAI service is ignored


masonzhang:
**Describe the bug**
In OpenAIClientAbstract.ExecutePostRequestAsync, when I sent a large prompt, there is error message in response:

{
  "error": {
    "message": "This model's maximum context length is 4097 tokens, however you requested 7326 tokens (3230 in your prompt; 4096 for the completion). Please reduce your prompt; or completion length.",
    "type": "invalid_request_error",
    "param": null,
    "code": null
  }
}

In the ExecutePostRequestAsync, you throw new AIException(AIException.ErrorCodes.InvalidRequest,$"The request is not valid, HTTP status: {response.StatusCode:G}"). The important information from error.message is missed. I can only get this message when I debug into the code.

Actually the error message should be post to outer scope.

**To Reproduce**
1. Give a very large ASK during your sample code.

**Expected behavior**
the error message from OpenAI serviceshould be post to outer scope, otherwise users will never be able to find the root cause.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows
 - IDE: Visual Studio
 - NuGet Package Version: 0.8.11.1-preview

**Additional context**
N/A

alexchaomander:
This is a great callout! Thanks for bringing this up. We'll add a backlog item to get this fixed!

dluc:
100%, I had a similar thought when I saw the issues in the PR to bring in sample 4. Error messages are great but still too opaque

evchaki:
@masonzhang , we are tracking this with the project team to add better error handling.

PR #529, opened 2023-04-19
Fix to log more details about failed function execution.

SergeyMenshykh:
### Motivation and Context

This change is required to allow SK SDK client code/hosting app to get more details about SK function execution failures.

Open issue: [Important Error Message from Azure OpenAI service is ignored](https://github.com/microsoft/semantic-kernel/issues/105)

### Description
This change does two things:
- Passes a logger instance as an argument of Kernel.FromSemanticConfig method so that the method can write logs at all.
- Adds AIException details to the warning message logged by SKFunction when handling the function execution failure.

### Contribution Checklist

- [] I didn't break anyone :smile:


dotnet/src/SemanticKernel/Orchestration/SKFunction.cs
```diff
@@ -112,11 +113,16 @@ public CompleteRequestSettings RequestSettings
                 string completion = await client.CompleteAsync(prompt, requestSettings, context.CancellationToken);
                 context.Variables.Update(completion);
             }
+            catch (AIException ex)
+            {
+                const string message = "Something went wrong while rendering the semantic function or while executing the text completion. Function: {0}.{1}. Error: {2}. Details: {3}";
+                log?.LogWarning(ex, message, skillName, functionName, ex.Message, ex.Detail);
```
shawncal, line 119: LogError?

dotnet/src/SemanticKernel/Orchestration/SKFunction.cs
```diff
@@ -112,11 +113,16 @@ public CompleteRequestSettings RequestSettings
...
+                context.Fail(ex.Message, ex);
```
shawncal, line 120: Should context.Fail() not do all the same things? It knows what skill/function was called, and has the exception. Does it have access to the logger instance?

dotnet/src/SemanticKernel/Orchestration/SKFunction.cs
SergeyMenshykh: Yep, make sense.

dotnet/src/SemanticKernel/Orchestration/SKFunction.cs
SergeyMenshykh: I wanted to do that as well but then I thought that SK SDK already has the mechanism that can be used for troubleshooting purposes - **logger**, and decided not to add unnecessary details into the context to keep it "light" until the point we have a strong reason to do so. Let me know what you think.

SergeyMenshykh: _Does it have access to the logger instance?_ - Yes, the SK context has access to the logger, but it's not used by it. Instead, the other classes that has access to the context instance use its Log property to access the logger.
 _Should context.Fail() not do all the same things?_ - The current pattern used in SK SDK is that every class that catches exceptions logs them itself, and calls context.Fail to indicate that the last operation failed and save the failure details. We can of cause change it, but should we? It kind of make sense to allow the classes themselves to log the errors because they have more information about the failure and can provide additional, more meaningful, failure details than if the logging was implemented in the context class. Moreover, if the context.Fail method loggs errors then the same error will be logged twice unless we refactor the classes.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -187,7 +187,7 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
             try
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                context = await f.InvokeAsync(context);
+                context = await f.InvokeAsync(context, log: this._log);
```
lemillermicrosoft, line 190: Is this required? `this._log` is already assigned to `context` above.

lemillermicrosoft (approved):

lemillermicrosoft (approved):
LGTM

MovGP0:
- instead of making the `ILogger log` nullable, consider using the [NullLogger](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.abstractions.nulllogger) instead.
- you should check [ILogger.IsEnabled(LogLevel)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger.isenabled) before interpolating the message string to improve performance.

SergeyMenshykh:
> * instead of making the `ILogger log` nullable, consider using the [NullLogger](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.abstractions.nulllogger) instead.
> * you should check [ILogger.IsEnabled(LogLevel)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger.isenabled) before interpolating the message string to improve performance.

Hi Johann, thank for your feedback. We might consider it on a next iteration of error handling/logging improvements to have it applied to all simal places and not only to the code touched in this PR.

PR #530, opened 2023-04-19
Vectorize DotProductImplementation with Vector<T>

stephentoub:
### Motivation and Context

Accelerate the DotProductImplementation for floats and doubles using SIMD.

### Description

Employ `Vector<T>` to accelerate DotProductImplementation. If the hardware supports acceleration, a vector's worth of data is processed at a time until there isn't a vector's worth remaining. Then it falls back to the scalar path that's also used if hardware acceleration isn't available.

Note that this removes the manual loop unrolling for lengths 2 and 4.  That would only still be useful on machines that don't have hardware acceleration support, which would be very old machines. The length 2 implementation also had a bug that prevented it from being used at all... it's up-front guard check was doing % 4 instead of % 2, so if you did call DotProduct with a length that was a multiple of 2 but not of 4 such that it fell back to using this implementation, it would immediately throw an exception.  I added a test that would fail with the old implementation.

### Contribution Checklist

On my machine:

```C#
private double[] _values1, _values2;

[Params(2, 4, 16, 130, 1_024)]

[GlobalSetup]
public void Setup()
{
    _values1 = Enumerable.Range(0, Length).Select(_ => Random.Shared.NextDouble()).ToArray();
    _values2 = Enumerable.Range(0, Length).Select(_ => Random.Shared.NextDouble()).ToArray();
}

[Benchmark]
public double Old() => DotProductImplementation1(_values1, _values2);

[Benchmark]
public double New() => DotProductImplementation2(_values1, _values2);
```

Floats:

| Method | Length |         Mean |
|------- |------- |-------------:|
|    Old |      2 |     6.688 ns |
|    New |      2 |     4.683 ns |
|    Old |      4 |     8.324 ns |
|    New |      4 |     5.594 ns |
|    Old |     16 |    13.444 ns |
|    New |     16 |     7.326 ns |
|    Old |    130 |    64.490 ns |
|    New |    130 |    27.029 ns |
|    Old |   1024 |   432.598 ns |
|    New |   1024 |   192.221 ns |
|    Old |  10240 | 4,717.146 ns |
|    New |  10240 | 1,927.729 ns |

Doubles:

| Method | Length |         Mean |
|------- |------- |-------------:|
|    Old |      2 |     6.832 ns |
|    New |      2 |     4.724 ns |
|    Old |      4 |     7.298 ns |
|    New |      4 |     4.205 ns |
|    Old |     16 |    12.143 ns |
|    New |     16 |     7.881 ns |
|    Old |    130 |    59.685 ns |
|    New |    130 |    34.040 ns |
|    Old |   1024 |   269.562 ns |
|    New |   1024 |   245.630 ns |
|    Old |  10240 | 3,080.644 ns |
|    New |  10240 | 2,816.317 ns |

cc: @tannergooding


dotnet/src/SemanticKernel/AI/Embeddings/VectorOperations/DotProductOperation.cs
```diff
@@ -75,237 +76,76 @@ private static unsafe double DotProductImplementation(ReadOnlySpan<double> x, Re
...
-        {
-            return DotProduct_Len4(x, y);
-        }
-
-        if (x.Length % 2 == 0)
-        {
-            // Twice as fast
-            return DotProduct_Len2(x, y);
-        }
+                    dotSum += Vector.Dot(*(Vector<double>*)px, *(Vector<double>*)py); // Dot product
```
umeshma, line 92: Since the input is ReadOnlySpan<float>, this looks fishy:
*(Vector<double>*)px, 
Should it be *(Vector<float>) ?

dotnet/src/SemanticKernel/AI/Embeddings/VectorOperations/DotProductOperation.cs
```diff
@@ -75,237 +76,76 @@ private static unsafe double DotProductImplementation(ReadOnlySpan<double> x, Re
...
-        {
-            return DotProduct_Len2(x, y);
-        }
+            double dotSum = 0;
 
-        // Vanilla Dot Product
-        fixed (double* pxBuffer = x)
-        {
-            fixed (double* pyBuffer = y)
+            if (Vector.IsHardwareAccelerated &&
```
umeshma, line 86: I like this impl! Simple, but will much faster...

dotnet/src/SemanticKernel/AI/Embeddings/VectorOperations/DotProductOperation.cs
```diff
@@ -75,237 +76,76 @@ private static unsafe double DotProductImplementation(ReadOnlySpan<double> x, Re
...
-                float* px = pxBuffer;
-                float* pxMax = px + x.Length;
-                float* py = pyBuffer;
-                while (px < pxMax)
-                {
-                    // Dot product
-                    dotSum += *px * *py;
-                    ++px;
-                    ++py;
+                    px += Vector<double>.Count;
```
umeshma, line 94: This should be Vector< float> I figure...

dotnet/src/SemanticKernel/AI/Embeddings/VectorOperations/DotProductOperation.cs
```diff
@@ -75,237 +76,76 @@ private static unsafe double DotProductImplementation(ReadOnlySpan<double> x, Re
...
+                    py += Vector<double>.Count;
```
umeshma, line 95: Same as above. Pointer increment will work right but should be Vector< float >

lemillermicrosoft (approved):
LGTM

PR #531, opened 2023-04-19
Tweak HttpSkill for improved performance

stephentoub:
### Motivation and Context

Suggested changes to HttpSkill based on code reviewing its implementation.

### Description

- If no HttpClient is provided, use a shared HttpClientHandler singleton so that connections may be pooled under the covers.  Otherwise, every instantiation of the skill would create and teardown a connection for the request.  If there's only ever a single instance of the skill created, it's effectively equivalent.  In the future this could also be changed to use HttpClientFactory and allow it to manage singletons.
- Use GetStringAsync for the GetAsync function: this lets the implementation optimize the operation a bit more than what the consumer can do.
- Use a shared SendRequestAsync to deduplicate the Put/Post/DeleteAsync implementations.
- Use HttpCompletionOption.ResponseHeadersRead to allow the implementation of ReadAsStringAsync to more efficiently create the returned string rather than first buffering the whole response as a byte[].
- Ensure the response message is disposed.
- Add a remark to the HttpSkill ctor accepting an HttpClient that it assumes ownership of that instance such that a caller should expect it to be disposed of when the skill is disposed.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/CoreSkills/HttpSkill.cs
+++ b/dotnet/src/SemanticKernel/CoreSkills/HttpSkill.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.SemanticKernel.Orchestration;
 using Microsoft.SemanticKernel.SkillDefinition;
@@ -48,10 +49,11 @@ public class HttpSkill : IDisposable
     /// Sends an HTTP GET request to the specified URI and returns the response body as a string.
     /// </summary>
     /// <param name="uri">URI of the request</param>
+    /// <param name="context">The context for the operation.</param>
     /// <returns>The response body as a string.</returns>
     [SKFunction("Makes a GET request to a uri")]
-    public Task<string> GetAsync(string uri) =>
-        this._client.GetStringAsync(uri);
+    public Task<string> GetAsync(string uri, SKContext context) =>
+        this.SendRequestAsync(uri, HttpMethod.Get, cancellationToken: context.CancellationToken);
 
     /// <summary>
     /// Sends an HTTP POST request to the specified URI and returns the response body as a string.
@@ -62,7 +64,7 @@ public class HttpSkill : IDisposable
     [SKFunction("Makes a POST request to a uri")]
     [SKFunctionContextParameter(Name = "body", Description = "The body of the request")]
     public Task<string> PostAsync(string uri, SKContext context) =>
-        this.SendRequestAsync(uri, HttpMethod.Post, new StringContent(context["body"]));
+        this.SendRequestAsync(uri, HttpMethod.Post, new StringContent(context["body"]), context.CancellationToken);
 
     /// <summary>
     /// Sends an HTTP PUT request to the specified URI and returns the response body as a string.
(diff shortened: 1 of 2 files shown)


dotnet/src/SemanticKernel/CoreSkills/HttpSkill.cs
```diff
@@ -83,25 +72,28 @@ public async Task<string> PostAsync(string uri, SKContext context)
...
+        this.SendRequestAsync(uri, HttpMethod.Put, new StringContent(context["body"]));
 
     /// <summary>
     /// Sends an HTTP DELETE request to the specified URI and returns the response body as a string.
     /// </summary>
     /// <param name="uri">URI of the request</param>
     /// <returns>The response body as a string.</returns>
     [SKFunction("Makes a DELETE request to a uri")]
-    public async Task<string> DeleteAsync(string uri)
+    public Task<string> DeleteAsync(string uri) =>
```
shawncal, line 84: Should we skip reading/returning body content for HttpMethod.Delete?

dotnet/src/SemanticKernel/CoreSkills/HttpSkill.cs
```diff
@@ -67,13 +61,8 @@ public async Task<string> GetAsync(string uri)
...
     [SKFunction("Makes a POST request to a uri")]
     [SKFunctionContextParameter(Name = "body", Description = "The body of the request")]
-    public async Task<string> PostAsync(string uri, SKContext context)
-    {
-        using var httpContent = new StringContent(context["body"]);
-        var response = await this._client.PostAsync(uri, httpContent);
-        var content = response.Content;
-        return await content.ReadAsStringAsync();
-    }
+    public Task<string> PostAsync(string uri, SKContext context) =>
```
shawncal, line 64: While we're updating, could you please plumb through the CancellationToken (accessible from SKContext.CancellationToken) to the downstream async calls? Same for Get, Put, Delete.

stephentoub: Happy to.  I actually had it on my list of questions / feedback how cancellation was intended to be handled... I hadn't seen yet it was part of the context.

dotnet/src/SemanticKernel/CoreSkills/HttpSkill.cs
stephentoub: If we don't expect responses, it'd make sense to simplify.  It really depends on the usage you expect; deletes can of course have a response body.

stephentoub:
> use a shared HttpClientHandler singleton

In looking through more code since putting up this PR, I see multiple places we're creating individual handlers.  After this PR, I'd like to look at consolidating most / all of those.

dotnet/src/SemanticKernel/CoreSkills/HttpSkill.cs
shawncal: Fine to leave it if it's possible. As long as an empty response body won't crash. Thanks!

shawncal:
> > use a shared HttpClientHandler singleton
> 
> In looking through more code since putting up this PR, I see multiple places we're creating individual handlers. After this PR, I'd like to look at consolidating most / all of those.

Yup, good call. We've got a conversation ongoing about a few options for how to distribute Handlers (or HttpClientFactory instance) around the kernel. We'll try to close on that design this week and implement shortly thereafter.

dotnet/src/SemanticKernel/CoreSkills/HttpSkill.cs
stephentoub: > As long as an empty response body won't crash.

Nope, it'll just result in an empty string.

shawncal (approved):

PR #553, opened 2023-04-20
Adding ConfigureAwait(false) to SDK projects

shawncal:
### Motivation and Context
It's recommended that .ConfigureAwait(false) be added to all awaits in SDK-style projects, to prevent deadlocks in calling apps.

### Description
- This change re-enables the .editorconfig rules that check for, and enforce, a ConfigureAwait(true or false), in projects under the dotnet/ folder. 
- Sample apps (under samples/) are not affected
- Unit test projects under dotnet/ are excluded via <NoWarn> tags in the csproj

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -73,7 +73,7 @@ public abstract class ClientBase
         }
 
         Response<Completions>? response = await RunRequestAsync<Response<Completions>?>(
-            async () => await this.Client.GetCompletionsAsync(this.ModelId, options, cancellationToken));
+            async () => await this.Client.GetCompletionsAsync(this.ModelId, options, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
```
stephentoub, line 76: Looks like a search/replace mistake... you only need one ConfigureAwait, not two :)

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -99,7 +99,7 @@ public abstract class ClientBase
             var options = new EmbeddingsOptions(text);
 
             Response<Embeddings>? response = await RunRequestAsync<Response<Embeddings>?>(
-                async () => await this.Client.GetEmbeddingsAsync(this.ModelId, options, cancellationToken));
+                async () => await this.Client.GetEmbeddingsAsync(this.ModelId, options, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
```
stephentoub, line 102: Oops, I misread the code. This is right.

stephentoub:
One important question to answer here...

Developers can supply delegates that will end up being invoked by the library. Is there any expectation that those delegates will be invoked back on the original SynchronizationContext that created them? eg Might someone reasonably expect that they can supply a delegate that directly updates controls on their UI?

MovGP0:
LGTM

💡You could use [Fody.ConfigureAwait](https://github.com/Fody/ConfigureAwait) to make that implicit, since it should be the default for new code anyway.

MovGP0:
> Developers can supply delegates that will end up being invoked by the library.

Yes that is an issue here. With the `ConfigureAwait(false)` the code will potentially return in a different thread, such that the delegate will also be called from a different thread; and the awaited value might also happen on a different thread. Client libraries, which are not written in a thread-save manner, will likely be effected by this change.

My personal opinion is that this would be an issue with the client that should be fixed; rather than an issue that should be blamed on the library.

PR #555, opened 2023-04-20
Avoid fully buffering HTTP response streams

stephentoub:
### Motivation and Context

Avoid potentially significant memory consumption by buffering downloaded files in memory.

### Description

In situations where we're reading from a response stream, e.g. copying it to a file, it's generally best to just go straight from the network to the destination... but the default behavior of HttpClient.GetAsync is to buffer the entire response payload.  We can avoid that by specifying HttpCompletionOption.ResponseHeadersRead instead of the default GetAsync behavior of HttpCompletionOption.ResponseContentRead.

### Contribution Checklist


dotnet/src/SemanticKernel.Skills/Skills.Web/WebFileDownloadSkill.cs
```diff
@@ -69,14 +69,14 @@ public async Task DownloadToFileAsync(string source, SKContext context)
...
-        HttpResponseMessage response = await this._httpClient.GetAsync(new Uri(source), context.CancellationToken);
+        using HttpResponseMessage response = await this._httpClient.GetAsync(new Uri(source), HttpCompletionOption.ResponseHeadersRead, context.CancellationToken);
         response.EnsureSuccessStatusCode();
         this._logger.LogDebug("Response received: {0}", response.StatusCode);
 
         using Stream webStream = await response.Content.ReadAsStreamAsync();
         using FileStream outputFileStream = new(Environment.ExpandEnvironmentVariables(filePath), FileMode.Create);
 
-        await webStream.CopyToAsync(outputFileStream, (int)webStream.Length, cancellationToken: context.CancellationToken);
+        await webStream.CopyToAsync(outputFileStream, 81920, cancellationToken: context.CancellationToken);
```
shawncal, line 79: Can you please add a comment or set a constant explaining this magic number? I found an explainer here, so I'm not entirely in the dark:
https://learn.microsoft.com/en-us/windows/uwp/debug-test-perf/optimize-file-access

stephentoub: Sure. It's the same constant value used by Stream when a size isn't supplied. Unfortunately when targeting netstandard2.0, in order to supply a cancellation token you also need to provide the size... that's not needed when targeting net core, where there's an overload that takes just a token.

lemillermicrosoft (approved):

Issue #508, opened 2023-04-18
WebSearch Results using the Bing Connector has a hard code value of count=1. We need to to return more than 1 result or should be able to pass count of result requested


Tbablani:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]
 - IDE: [e.g. Visual Studio, VS Code]
 - NuGet Package Version [e.g. 0.1.0]

**Additional context**

evchaki:
@Tbablani , thanks for the idea, we can take a look.

Tbablani:
Hi All

I tried to change the interface and create a method which will accept a count. SK solution complied properly but when I was trying to test it giving error when trying to initiate the kernel

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

shawncal:
Fixed with https://github.com/microsoft/semantic-kernel/pull/556

Tbablani:
For this fix , do you have the latest nuget package
________________________________

danf25:
I have the same problem.

Thank you for your time.
Danf

PR #556, opened 2023-04-20
fix issue 508

feiyun0112:
# Motivation and Context
fix issue #508 

# Description
WebSearch Results using the Bing Connector has a hard code value of count=1. 

We need to return more than 1 result or should be able to pass count of result requested

# Contribution Checklist
 The PR follows SK Contribution Guidelines (https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md)
 The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with dotnet format
 I didn't break anyone 😄


dluc (dismissed):
a few things before merging:
* provide PR title and description, pls see PR template
* fix code style (see rider/resharper)
* make sure this follow the same naming convention and approach used for "recall"

dotnet/src/SemanticKernel.Skills/Skills.Web/WebSearchEngineSkill.cs
```diff
@@ -20,14 +40,35 @@ public WebSearchEngineSkill(IWebSearchEngineConnector connector)
 
     [SKFunction("Perform a web search.")]
     [SKFunctionInput(Description = "Text to search for")]
+    [SKFunctionContextParameter(Name = Parameters.Count, Description = "Number of results")]
```
adrianwyatt, line 43: Let's add default values of count = 1 and offset = 0, to support the default case. This will help the Kernel and language models to understand what it means to not set these parameters.

dotnet/src/SemanticKernel.Skills/Skills.Web/WebSearchEngineSkill.cs
```diff
@@ -20,14 +40,35 @@ public WebSearchEngineSkill(IWebSearchEngineConnector connector)
...
+        if (!results.Any())
         {
             context.Fail("Failed to get a response from the web search engine.");
         }
 
-        return result;
+        string resultString;
+        if (countInt == 1)
+        {
+            var result = results.FirstOrDefault();
```
adrianwyatt, line 64: these lines can be simplified to `resultString = results.FirstOrDefault() ?? string.Empty;`

adrianwyatt (changes requested):

MovGP0:
Instead of 
```csharp
string resultString;
if (countInt == 1)
{
    resultString = results.FirstOrDefault() ?? string.Empty;
}
else
{
    resultString = JsonSerializer.Serialize(results);
}

return resultString;
```

you could do something like

```csharp
return countInt == 1
    ? results.FirstOrDefault() ?? string.Empty
    : JsonSerializer.Serialize(results);
```

adrianwyatt:
LGTM - please resolve conflicts and we'll bring it in.

adrianwyatt (dismissed):

shawncal:
Merged from main, updated Google connector too, as common interface changes require it.

shawncal (dismissed):

shawncal (approved):

Tbablani:
Do you have new nuget package for this fix. I downloaded the code locally and trying to run it from a console app adding local references and it is giving me build code.

Tbablani:
trying to test this through  a console app. As latest Nuget is not available for this issue.
Adding local project references to the DLL. It is not working . Getting errors when trying to run it.

PR #557, opened 2023-04-20
The wrong variable was assigned

feiyun0112:
[``](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs#L102)

```
var limit = context.Variables.ContainsKey(LimitParam) ? context[LimitParam] : DefaultLimit;     
if (string.IsNullOrWhiteSpace(limit)) { relevance = DefaultLimit; }
```


awharrison-28 (approved):
Thank you for fixing!

PR #523, opened 2023-04-19
Remove PlannerSkill and add SequentialPlanner

lemillermicrosoft:
### Motivation and Context

This pull request contains a series of changes that refactor and improve the planning module and tests in the SemanticKernel project. The planning module is responsible for creating and executing plans based on natural language prompts and available skills. The changes aim to simplify the code, enhance the functionality, and increase the test coverage and quality of the planning module.

### Description

- Add a new SequentialPlanner class that uses semantic function to generate plans from natural language prompts
- Add a new SequentialPlanParser class that parses the XML output of the semantic function and converts it into a Plan object
- **Remove PlannerSkill**
- Rename PlannerConfig class that contains common configuration options for planner instances, such as the maximum number of tokens, the relevancy threshold, and the excluded or included functions
- Refactor the Plan class to make it more consistent and robust, and add support for serialization, named outputs, and nested plans
- Add unit tests for the SequentialPlanner, SequentialPlanParser, and Plan classes and refactor the existing planning tests to use mock functions and skills.
- Fix a failing web skill test by skipping it due to inconsistent Bing search results

### Contribution Checklist

- [X] The code builds clean without any errors or warnings
- [X] The PR follows SK Contribution Guidelines (https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md)
- [X] The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`
- [X] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:
- [x] [Integration Tests passing](https://github.com/lemillermicrosoft/semantic-kernel/actions/runs/4760463551)

### Whats next
- Custom Example support
- Serialization/Deserialization improvements
- Other Planners

PR #560, opened 2023-04-20
Fix replacing for variables

Sy0xFF:
### Motivation and Context
We had a problem with `SKFunctionInput` when these are arrays. For example, the output was `[$variable1,$variable2]`. These were not replaced with the actual values.

### Description
Replaced the logic to only get one variable with a regex solution to replace all variables in the inner text. Now all values will be replaced correctly.

### Contribution Checklist

- [x] All unit tests pass, and I have added new tests where possible (excluding unrelated already broken unit tests)

### DIFF
--- a/dotnet/src/SemanticKernel/Planning/FunctionFlowRunner.cs
+++ b/dotnet/src/SemanticKernel/Planning/FunctionFlowRunner.cs
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Xml;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


lemillermicrosoft:
Thanks for the PR @Sy0xFF -- we're in the process of refactoring things here, see https://github.com/microsoft/semantic-kernel/pull/523. I'll incorporate this fix with those changes.

dotnet/src/SemanticKernel/Planning/FunctionFlowRunner.cs
```diff
@@ -269,35 +269,31 @@ private async Task<string> ProcessNodeListAsync(XmlNodeList nodeList, string fun
                             foreach (XmlAttribute attr in o2.Attributes)
                             {
                                 context.Log.LogTrace("{0}: processing attribute {1}", parentNodeName, attr.ToString());
-                                bool innerTextStartWithSign = attr.InnerText.StartsWith("$", StringComparison.Ordinal);
+                                var variablesRegex = new Regex("\\$(?<variableName>[0-9A-Za-z_]+)");
```
dluc, line 272: why would this be better than StartsWith?

dluc (changes requested):
before merging I would like to understand the rationale about replacing `StartsWith` with a regular expression

dluc:
as per @lemillermicrosoft the fix will be integrated into #523

PR #567, opened 2023-04-21
Some behavioral / perf fixes for Embedding<T>

stephentoub:
### Motivation and Context

Clean up some `Embedding<T>` behaviors / perf issues.

### Description

- IsSupported was more expensive than it needs to be: if it's changed to just compare the types directly rather than searching a list, the JIT can turn the entire operation into a JIT-time constant.
- The explicit cast to `ReadOnlySpan<T>` was cloning the array. It doesn't need to do that.
- `default(Embedding<T>)` results in a struct that fails on a bunch of operations with NullReferenceExceptions if you try to use it.  We've learned from experience with `ImmutableArray<T>` this is suboptimal.  I've made it functional and behave equivalent to Empty... in fact, Empty is now just `default`.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/Embedding.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/Embedding.cs
@@ -2,7 +2,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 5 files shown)


dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/Embedding.cs
```diff
@@ -46,38 +46,47 @@ public Embedding(IEnumerable<TEmbedding> vector)
...
 
+    private static void ThrowNotSupportedEmbedding() =>
+        throw new NotSupportedException($"Embeddings do not support type '{typeof(TEmbedding).Name}'. Supported types include: [ Single, Double ]");
+
     /// <summary>
-    /// Gets the vector as a <see cref="ReadOnlyCollection{T}"/>
+    /// Gets the vector as an <see cref="IEnumerable{TEmbedding}"/>
     /// </summary>
     [JsonPropertyName("vector")]
-    public IEnumerable<TEmbedding> Vector => this._vector.AsEnumerable();
```
stephentoub, line 61: I just realized the intent here with the AsEnumerable may have been to return am enumerable that wraps the array so that the array itself isn't exposed. But AsEnumerable os just a cast, so it was a nop here. If the goal is to return a wrapper, Select can be used.

MovGP0:
I don't like code in the form of `private readonly TEmbedding[]? _vector;` as seen in `Embeddings.cs`. 

Enumerable objects should return an empty enumerable object like `Array.Empty<T>` when not initialized instead of the `null` value.

dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/Embedding.cs
shawncal: I recall this going through a few iterations in the interest of having it be visible to the json serializer without a full copy. Whatever the case -- looks good.

shawncal (approved):

PR #570, opened 2023-04-21
Throw ArgumentNullExceptions for null arguments

stephentoub:
### Motivation and Context

Use ArgumentNullException as devs expect, and avoid exception-related work when not throwing.

### Description

.NET developers expect that passing null erroneously to a method will result in an ArgumentNullException, but currently it's resulting in a custom ValidationException.

There are also places where string interpolation is being used to create the error message; the work and allocation associated with that interpolation is going to happen regardless of whether the exception is thrown or not.

### Contribution Checklist


MovGP0:
LGTM. 

💡 A code weaver like [Fody.NullGuard](https://github.com/Fody/NullGuard) or a [Roslyn code generator](https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md) might be helpful in replacing all those runtime `null` checks.

dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs
```diff
@@ -15,18 +15,18 @@ internal static class Verify
     private static readonly Regex s_asciiLettersDigitsUnderscoresRegex = new("^[0-9A-Za-z_]*$");
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
stephentoub: > Thoughts on having a ValidateNotNull for the previous behavior?

We could... but most of the places that were using it for things other than actual arguments were doing things like this:
```C#
                config = PromptTemplateConfig.FromJson(File.ReadAllText(configPath));
                Verify.NotNull(config, $"Invalid prompt template configuration, unable to parse {configPath}");
```
That means it's going to be allocating a whole new string for the message even in the success case where config is non-null.  By providing a helper like this, I worry it'll encourage that kind of misuse.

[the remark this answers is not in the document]
lemillermicrosoft: Sounds good to me, thanks Stephen!

dotnet/src/SemanticKernel/Planning/FunctionFlowRunner.cs
```diff
@@ -259,8 +259,10 @@ private async Task<string> ProcessNodeListAsync(XmlNodeList nodeList, string fun
 
                     if (processFunctions && !string.IsNullOrEmpty(functionName))
                     {
-                        Verify.NotNull(functionName, nameof(functionName));
-                        Verify.NotNull(skillFunction, nameof(skillFunction));
+                        if (skillFunction is null)
```
lemillermicrosoft, line 262: Someone will need to resolve conflict with https://github.com/microsoft/semantic-kernel/pull/523 -- making note for self.

lemillermicrosoft (dismissed):

dotnet/src/SemanticKernel/Planning/FunctionFlowRunner.cs
lemillermicrosoft: I beat you :)

See `Plan.cs`, `SequentialPlanParser.cs`, and `SequentialPlanner.cs` for applying changes.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -40,7 +40,7 @@ public abstract class ClientBase
         CompleteRequestSettings requestSettings,
         CancellationToken cancellationToken = default)
     {
-        Verify.NotNull(requestSettings, "Completion settings cannot be empty");
+        Verify.NotNull(requestSettings);
```
dluc, line 43: Wouldn't you agree that custom error messages would help developers understanding the context, e.g. as opposed to generic exceptions and line numbers?

stephentoub: My changes include using CallerArgumentExpression with NotNull helper. The exception message in this case will be:
```
Value cannot be null. (Parameter 'requestSettings')
```
I think that's more valuable, telling the developer exactly which parameter is the problem. With the custom message:
```
Completion settings cannot be empty
```
the developer needs to know that a) empty means null and b) "completion settings" is referring to the requestSettings parameter.

lemillermicrosoft (dismissed):

stephentoub:
> looks like there are some conflicts that need to be resolved before merging.

Done. Thanks.

shawncal (approved):

PR #572, opened 2023-04-21
Bump C# language version to 11

stephentoub:
### Motivation and Context

Use the latest C# language version.

### Description

Some places already had 11 set, others were still using 10.  This makes it consistently 11 everywhere.  C# 11 is currently the latest, with C# 12 in preview until November.

### Contribution Checklist


MovGP0:
LGTM

dotnet/common.props
lemillermicrosoft, line 1: Is `common.props` used by anything? I'm wondering if we can remove it.

dotnet/src/SemanticKernel.Skills/Skills.UnitTests/Skills.UnitTests.csproj
lemillermicrosoft, line 1: There's a handful of projects that don't specify any version, like `dotnet\src\SemanticKernel.IntegrationTests\SemanticKernel.IntegrationTests.csproj` -- should we add to those as well?

dotnet/common.props
stephentoub: Doesn't look like it's referenced anywhere

dotnet/src/SemanticKernel.Skills/Skills.UnitTests/Skills.UnitTests.csproj
stephentoub: > should we add to those as well?

It'd be better IMHO to have a single location where all of these common settings are picked up from.  Presumably that's the intent of the root Directory.props?

dotnet/common.props
lemillermicrosoft: cc @shawncal might have more historical knowledge

dotnet/src/SemanticKernel.Skills/Skills.UnitTests/Skills.UnitTests.csproj
lemillermicrosoft: Agreed. I'm not sure if there's a reason these UT projects and IntegrationTest, etc. don't reference the Directory.props -- and I'm again finding myself wondering why the SemanticKernel.csproj doesn't either -- does dotnet pick it up automatically? Either way, should we just remove the `<langversion> from the csproj all together? Also cc @shawncal

lemillermicrosoft: https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#explore-customization-options

Found my own answer, so yeah, maybe we can just remove the `LangVersion` in the csprojs in this change.

lemillermicrosoft (dismissed):

dotnet/common.props
shawncal: I think you're right. I'll resolve the thread here, and fix it in an upcoming PR.

shawncal:
@stephentoub Change looks good -- some files were moved in another change, and I cannot resolve the conflicts from here. Could you please sync up?

shawncal (approved):

PR #587, opened 2023-04-21
add bool to enable speech input

amsacha:
### Motivation and Context
Hide the speech input if there is no or an invalid  speech resource key
Additionally also prevent an invalid key and/or the lack of a key for a speech resource from generating an error when it's not present. 

### Contribution Checklist


glahaye (approved):

Vivihung (dismissed):
LGTM. Thank you for the quick follow-up!

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -58,14 +58,15 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
     const [recognizer, setRecognizer] = React.useState<speechSdk.SpeechRecognizer>();
     const [isListening, setIsListening] = React.useState(false);
     const speechService = useSKSpeechService(process.env.REACT_APP_BACKEND_URI as string);
+    const bEnableSpeech = process.env.ENABLE_SPEECH_INPUT;
```
hathind-ms, line 61: rename to "enableSpeech"

samples/apps/copilot-chat-app/webapp/env.example
```diff
@@ -1,2 +1,3 @@
 REACT_APP_BACKEND_URI=https://localhost:40443/
-REACT_APP_AAD_CLIENT_ID=
\ No newline at end of file
+REACT_APP_AAD_CLIENT_ID=
+ENABLE_SPEECH_INPUT=0
```
adrianwyatt, line 3: Would it be difficult to change the parameter to "true" or "false"?
```
ENABLE_SPEECH_INPUT=true
```

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
amsacha: renamed to REACT_APP_ENABLE_SPEECH, the value isnt picked up correctly without that prefix.

adrianwyatt (approved):

Vivihung (approved):

adrianwyatt (dismissed):

samples/apps/copilot-chat-app/README.md
```diff
@@ -55,6 +55,8 @@ First, let’s set up and verify the back-end API server is running.
        ```bash
        dotnet user-secrets set "AzureSpeech:Key" "MY_AZURE_SPEECH_KEY" 
        ```
+     - In the .env file that you create below, set REACT_APP_ENABLE_SPEECH to true.  
+       ```REACT_APP_ENABLE_SPEECH=true```
```
dluc, line 59: I thought we were going to autodetect this

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -58,14 +58,16 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
     const [recognizer, setRecognizer] = React.useState<speechSdk.SpeechRecognizer>();
     const [isListening, setIsListening] = React.useState(false);
     const speechService = useSKSpeechService(process.env.REACT_APP_BACKEND_URI as string);
-
+    
+    const bEnableSpeech = (process.env.REACT_APP_ENABLE_SPEECH as string) === 'true' ? true : false;
+    
     React.useEffect(() => {
-        if (recognizer) return;
+        if (!bEnableSpeech || recognizer) return;
```
dluc, line 65: typo

dluc (changes requested):
in the last conversations we had about Azure Speech we said we would automatically turn it on/off, without asking the user to do extra work

samples/apps/copilot-chat-app/README.md
amsacha: Well we were auto-detecting it prior to this by only creating the input symbol if we found a valid speech resource. But when there isn't a key it would throw an error, that doesnt affect the rest of the app but does look bad for a reference app. So we decided to just hide it behind a bool for now.

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
amsacha: What's the typo?

amsacha: Dont see a type, closing

hathind-ms (dismissed):

samples/apps/copilot-chat-app/README.md
dluc: frontend => request to our backend, asking for speech config

    if (MY_AZURE_SPEECH_KEY is set)
        backend uses AAD token to get a Speech token
        return 200 + Config
    else:
        return 404 + "NOT AVAILABLE"

samples/apps/copilot-chat-app/webapp/env.example
```diff
@@ -1,2 +1,3 @@
 REACT_APP_BACKEND_URI=https://localhost:40443/
-REACT_APP_AAD_CLIENT_ID=
\ No newline at end of file
+REACT_APP_AAD_CLIENT_ID=
+REACT_APP_ENABLE_SPEECH=false
```
dluc, line 3: we can remove this, the backend will return 200 when True, 404 when False

Vivihung (changes requested):
Like the new approach which doesn't require any extra steps to set a flag! Have some recommendations in comments.

Vivihung:
> in the last conversations we had about Azure Speech we said we would automatically turn it on/off, without asking the user to do extra work

Sorry, that was my recommendation without knowing the past conversation. Thanks for the push and I do like the latest change which has less asks on the user side.

Vivihung (dismissed):
Thanks for making the fix better and reducing the efforts on the user side!

dluc:
thanks @amsacha <3

dluc (approved):

PR #389, opened 2023-04-10
Plan object execution capabilities

shawncal:
### Motivation and Context

This is the next PR in improving Planning capabilities in the Kernel. It builds upon previous https://github.com/microsoft/semantic-kernel/pull/255 to enable execution of plans and improves the ability to create plans.

### Description

      - Rename the Plan class to SKPlan and add support for invoking functions as part of the plan execution.
      - Add a new class for executing plans, and remove the unused PlanRunner class.
      - Add unit and integration tests for the planning functionality and a test email skill.
      - Fix some minor issues and add documentation and comments to the planning classes and tests.
      - Update the VS Code tasks and editor config files to improve the development experience.

### Contribution Checklist

- [X] The code builds clean without any errors or warnings
- [X] The PR follows SK Contribution Guidelines (https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md)
- [X] The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`
- [X] All unit tests pass, and I have added new tests where possible
- [X] I didn't break anyone :smile:

### Whas Next
- Update PlannerSkill to parse NL XML results to Plan object for plan creation and rely on native Kernel Support for executing Plan objects.
- Support for other Plan strategies

PR #597, opened 2023-04-22
Python: Basic JSON-based Planner implementation 

alexchaomander:
### Motivation and Context
Many users and customers have expressed interest in having planning capabilities in Python. This PR seeks to provide a basic version of the planner so that people can learn and play around with the concepts and can be comfortable with implementing their own versions. 

A Jupyter notebook is included. 

### Description
Most of the logic of the plan can be found under the `PROMPT` in `basic_planner.py`.

There is a known limitation that this currently doesn't explicit parse out arguments that are defined in the semantic function prompts, so when creating the [AVAILABLE FUNCTIONS] string, it just uses the name of the function and the description.

The execution of the plan also happens from start to finish. There is currently no support for piecewise execution.


python/semantic_kernel/core_skills/planning/basic_planner.py
```diff
@@ -0,0 +1,203 @@
...
+    Usage:
+        kernel.import_skill(BasicPlanner(), "planner");
+
+    Examples:
+
+    {{planner.create_plan}}
+    {{planner.execute_plan}}
+    """
+
+    @sk_function(
```
lemillermicrosoft, line 116: See https://github.com/microsoft/semantic-kernel/pull/523 -- we're moving away from Planner's being skills and instead are a top level concept in SK.

python/semantic_kernel/core_skills/planning/plan.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any
+
+
+class Plan:
```
lemillermicrosoft, line 4: See the latest `Plan` object in dotnet.

https://github.com/microsoft/semantic-kernel/pull/389/files#diff-128f2b23a8593a275761d05e910b08ee6e369f7323487d1d13aed49cdc3e3dbf

samples/notebooks/python/05-using-the-planner.ipynb
lemillermicrosoft, line 1: This is half a question for myself, but shouldn't we update notebooks after we spin nugets with the changes?

lemillermicrosoft: ((Random idea @shawncal - that's why it'd be nice to have a repo that is purely sample code hitting nuget packages, and then the main development repo can stay in sync with what's checked in so it's easy to snap to the sample branch when we spin nugets.))

python/semantic_kernel/core_skills/planning/basic_planner.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Create the [AVAILABLE FUNCTIONS] section of the prompt
+        available_functions_string = ""
+        for name in list(all_functions_dict.keys()):
+            available_functions_string += name + "\n"
+            available_functions_string += (
+                "description: " + all_functions_dict[name] + "\n"
+            )
+            available_functions_string += "args:\n"
+            # TODO: FIGURE OUT HOW TO GET ARGS FROM FUNCTIONS
```
lemillermicrosoft, line 158: `FunctionView` I believe is your solution, looks like it is now available in Python too.

python/semantic_kernel/core_skills/planning/basic_planner.py
alexchaomander: Good to know! I wasn't able to register this planner as a skill anyway.

alexchaomander: @dluc if it's not going to be a skill, should I put this somewhere else? Maybe `python/semantic_kernel/planning`?

dluc (changes requested):
design in progress

python/semantic_kernel/core_skills/planning/basic_planner.py
lemillermicrosoft: For what it's worth, the dotnet stuff will be in nugetpackages like `Microsoft.SemanticKernel.Planning.SequentialPlanner` and the code is in `dotnet/src/Extensions/Planning.SequentialPlanner/` for example.

samples/notebooks/python/05-using-the-planner.ipynb
```diff
@@ -1,203 +1,299 @@
...
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\n",
+      "Hi there,\n",
+      "\n",
+      "Are you looking for a romantic activity to do with your special someone? Here are some ideas that might be fun: \n",
```
mkarle, line 256: nit: This example doesn't really sound like Shakespeare to me. Also, it's a weird start to an email to your significant other.  I say this because I wasn't convinced the planner was working properly, so I ran it with some print statements to make sure each function was running.

python/semantic_kernel/planning/__init__.py
awharrison-28, line 1: We can address this later, but I don't think this __init__ file is needed

python/semantic_kernel/planning/basic_planner.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # Create the semantic function for the planner with the given prompt
+        planner = kernel.create_semantic_function(
+            prompt, max_tokens=1000, temperature=0.8
+        )
+
+        available_functions_string = self._create_available_functions_string(kernel)
+
+        # Create the context for the planner
+        context = ContextVariables()
```
awharrison-28, line 172: nit: it would be cleaner if the plan object could provide the ContextVariables() instead of having to put them together manually. 

If you don't want to address here, we could perhaps add it to the backlog?

python/semantic_kernel/planning/basic_planner.py
```diff
@@ -0,0 +1,208 @@
...
+    async def execute_plan_async(self, plan: Plan, kernel: Kernel) -> str:
+        """
+        Given a plan, execute each of the functions within the plan
+        from start to finish and output the result.
+        """
+        generated_plan = json.loads(plan.generated_plan.result)
+
+        context = ContextVariables()
+        context["input"] = generated_plan["input"]
+        subtasks = generated_plan["subtasks"]
```
awharrison-28, line 188: again, could be part of the Plan() initialization, but doesn't have to be addressed in this PR

samples/skills/WriterSkill/Translate/config.json
```diff
@@ -3,7 +3,7 @@
   "type": "completion",
   "description": "Translate the input into a language of your choice",
   "completion": {
-    "max_tokens": 256,
+    "max_tokens": 2000,
```
awharrison-28, line 6: Why 2000 over 500, 1000 etc? I'm not aginst it, but it does make the skill less flexible to smaller models.

alexchaomander: For the example that I have, the translation of the output requires more tokens. Also, since this is `max_tokens`, it should still be okay for even simple translations right?

python/semantic_kernel/planning/basic_planner.py
alexchaomander: Let's clean it up in a future PR!

python/semantic_kernel/planning/basic_planner.py
alexchaomander: Let's do this in a future PR!

samples/skills/WriterSkill/Translate/config.json
alexchaomander: Also had to rename the TranslateV2 skill because the Planner could get confused on which skill to use. Also TranslateV2 was mostly about improving English so renamed it to EnglishImprover

python/semantic_kernel/planning/__init__.py
alexchaomander: removed

python/semantic_kernel/planning/basic_planner.py
```diff
@@ -0,0 +1,208 @@
...
+description: Write in the style of author Edgar Allen Poe
+args:
+- input: the input to write about
+
+WriterSkill.EmailTo
+description: Write an email to a recipient
+args:
+- input: the input to write about
+- recipient: the recipient's email address.
+
```
mkarle, line 76:
```suggestion

WriterSkill.Translate
description: translate the input to another language
args:
- input: the text to translate
- language: the language to translate to

```

python/semantic_kernel/planning/basic_planner.py
```diff
@@ -0,0 +1,207 @@
...
+E-mail these ideas to my significant other"
+
+[OUTPUT]
+    {
+        "input": "Valentine's Day Date Ideas",
+        "subtasks": [
+            {"function": "WriterSkill.Brainstorm"},
+            {"function": "ShakespeareSkill.shakespeare"},
+            {"function": "WriterSkill.EmailTo", "args": {"recipient": "significant_other"}},
+            {"function": "translate", "args": {"language": "French"}}
```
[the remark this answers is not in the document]
mkarle: The example can include it, but the Available Functions list for this example doesn't include the Translate skill.

mkarle (approved):

python/semantic_kernel/planning/basic_planner.py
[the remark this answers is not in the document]
alexchaomander: Added!

mkarle (approved):

PR #599, opened 2023-04-22
Adding Multiple Skill Directories Simplification + Integration Tests readme changes

RogerBarreto:
### Motivation and Context - Improvement

It is a common scenario having to import multiple skills. 

Adding this as a `param` array to the extension will reduce number of calls and improve cleanliness.  

### Description

As part of this PR I also spotted missing in the Readme the recommendation on using UserSecrets for Integration Tests.

Changed the readme and example config files to use davinci-003 models as our current standard for integration tests.

### Contribution Checklist


dotnet/src/SemanticKernel/KernelExtensions/ImportSemanticSkillFromDirectory.cs
```diff
@@ -53,48 +53,51 @@ public static class ImportSemanticSkillFromDirectoryExtension
...
-            // Prepare lambda wrapping AI logic
-            var functionConfig = new SemanticFunctionConfig(config, template);
+                kernel.Log.LogTrace("Config {0}: {1}", functionName, config.ToJson());
 
-            kernel.Log.LogTrace("Registering function {0}.{1} loaded from {2}", skillDirectoryName, functionName, dir);
-            skill[functionName] = kernel.RegisterSemanticFunction(skillDirectoryName, functionName, functionConfig);
+                // Load prompt template
+                var template = new PromptTemplate(File.ReadAllText(promptPath), config, kernel.PromptTemplateEngine);
+
+                // Prepare lambda wrapping AI logic
```
shawncal, line 95: ? lamba

dotnet/src/SemanticKernel/KernelExtensions/ImportSemanticSkillFromDirectory.cs
```diff
@@ -53,48 +53,51 @@ public static class ImportSemanticSkillFromDirectoryExtension
     /// </summary>
```
shawncal, line 53: This is a clever adaptation, but I wonder how intuitive the (original) call was. 

From the caller's perspective why do we have to specify a parent directory (that effectively means nothing) plus a subdirectory? Why not just provide a list of directories you want to load from?

The current approach forces all callers to have something like a "skills" folder, which doesn't really make any sense. If you just have one or two skills, and each of those is in a folder in the root, we should be able to just register those by passing their folder names directly. (Again, I know this predates this PR, but it's a good opportunity to fix it).

My suggestions:
 - keep this method signature, as I see the utility.
 - refactor out the part inside the while loop into another method for readability
 - make it so that, you can pass in a folder name to the "parentDirectory" arg, and that recursively searches for skills regardless throughout its descendants. The user could pass in the skill directory itself, a higher-level "skills" folder, or even the whole app -- we should discover the available skills under each.
 - the current "skillDirectoryNames" becomes more of a string filter, rather than looking for those directories specifically -- allows you to optionally specify which of the discovered skills to import. As skills are discovered in the folder descendants, it checks to see if their names are in the filtered list (or if filter list is empty) before registering.

shawncal (approved):

shawncal (changes requested):
Changes requested. Follow up with shawncal if needed.

dotnet/src/SemanticKernel/KernelExtensions/ImportSemanticSkillFromDirectory.cs
RogerBarreto: I think the suggestion is great but I opened this PR to be small and fast, not introducing a big change or greater improvement to the way this method behaves.

samples/dotnet/kernel-syntax-examples/Example12_Planning.cs
```diff
@@ -145,17 +147,18 @@ private static async Task MemorySampleAsync()
...
-        kernel.ImportSemanticSkillFromDirectory(folder, "CalendarSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "ChatSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "ChildrensBookSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "ClassificationSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "CodingSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "FunSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "IntentDetectionSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "MiscSkill");
-        kernel.ImportSemanticSkillFromDirectory(folder, "QASkill");
+        kernel.ImportSemanticSkillFromDirectory(folder,
```
lemillermicrosoft, line 150: Things like `ImportSampleSkills` in dotnet\src\SemanticKernel.IntegrationTests\TestHelpers.cs could be improved, too.

RogerBarreto: @lemillermicrosoft Good call. Will add those changes to the PR.

dotnet/src/SemanticKernel/KernelExtensions/ImportSemanticSkillFromDirectory.cs
```diff
@@ -53,48 +53,50 @@ public static class ImportSemanticSkillFromDirectoryExtension
...
 
-        Verify.ValidSkillName(skillDirectoryName);
-        var skillDir = Path.Combine(parentDirectory, skillDirectoryName);
-        Verify.DirectoryExists(skillDir);
-
         var skill = new Dictionary<string, ISKFunction>();
 
-        string[] directories = Directory.GetDirectories(skillDir);
-        foreach (string dir in directories)
+        foreach (string skillDirectoryName in skillDirectoryNames)
```
lemillermicrosoft, line 66: nit: wondering if it's worth doing this in parallel.

lemillermicrosoft (approved):

dotnet/src/SemanticKernel/KernelExtensions/ImportSemanticSkillFromDirectory.cs
lemillermicrosoft: Especially after #603

RogerBarreto: I'm good on merging this one and working on it later as Shawn suggested.

shawncal (dismissed):

shawncal:
@RogerBarreto Looks like we've got a build break after merge. Can you take a look please? Once that's resolved, should be good to go.

shawncal (approved):

PR #610, opened 2023-04-24
Pinecone Client and Memory Store

Kevdome3000:
### Motivation and Context
    SK currently does not have a Pinecone Connector
    Allows users to use Pinecone as a Memory Store
    Memory

### Description

An implementation of IMemoryStore for Pinecone Vector database.

### Contribution Checklist

I still need to add unit tests and example usage but wanted to get the conversation started and garner feedback before finalizing


Kevdome3000:
> @Kevdome3000 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

shawncal (commented):
Great work, thank you! I'll take a look at this with fresh eyes tomorrow, and we'll try to get it in this week.

Kevdome3000:
After playing with the a few of the sample apps and getting a better understanding of how the IMemoryStore interface is being utilized, I think a refactor makes sense.

For those familiar with Pinecone, namespaces are really powerful and for those who aren't, this is from their documentation:

_Pinecone allows you to partition the vectors in an index into namespaces. Queries and other operations are then limited to one namespace, so different requests can search different subsets of your index.

For example, you might want to define a namespace for indexing articles by content, and another for indexing articles by title. For a complete example, see: [Semantic Text Search (Example)](https://docs.pinecone.io/docs/semantic-text-search/).

Every index is made up of one or more namespaces. Every vector exists in exactly one namespace._

Imagine a namespace per Kernel or SemanticTextMemory that can be cleared after a chat session while keeping other semantic data that exists in a Pinecone Index in tact. That's the ideal implementation in my opinion.

This PR in it's current state doesn't allow for that due to the fact that a user can also have multiple Pinecone Indexes, so the "collection"/"collectionName" parameters used throughout the IMemoryStore/ ISemanticTextMemory interface becomes ambiguous.

The easiest solution is to agree that a user is restricted to a single index per PineconeMemoryStore/ PineconeClient instance so that indexName no longer needs to be passed as a parameter so that collectionName/collection can be synonymous with namespace.

I hope that makes sense.

If on board let me know. I can turn that around quickly

dotnet/src/Connectors/Connectors.Memory.Pinecone/Model/IndexMetric.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <summary>
+    /// Enum Dotproduct for value: dotproduct
+    /// </summary>
+    [EnumMember(Value = "dotproduct")]
+    Dotproduct = 3
+
+}
+
+public class IndexMetricJsonConverter : JsonConverter<IndexMetric>
```
dmytrostruk, line 37: There are three new converters:
- PodTypeJsonConverter
- IndexMetricJsonConverter
- IndexStateJsonConverter

Looks like the implementation is the same in all converters, they check for value in `EnumMember` attribute. A couple of options here:
- Check if standard `JsonStringEnumConverter` covers `EnumMember` attribute, if yes, use it.
- Check for already existing converters in `System.Text.Json` which cover `EnumMember`. If exist, use it.
- Create own converter to cover `EnumMember` attribute but make it generic for all enums.

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeClient.cs
```diff
@@ -0,0 +1,758 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.SemanticKernel.Connectors.Memory.Pinecone.Http;
+using Microsoft.SemanticKernel.Connectors.Memory.Pinecone.Http.ApiSchema;
+using Microsoft.SemanticKernel.Connectors.Memory.Pinecone.Model;
+
+// ReSharper disable UnusedVariable
```
dmytrostruk, line 16: Not sure we need to disable this rule 😃

dmytrostruk (changes requested):
Thanks for contribution, this is amazing work! I left initial comments to address, let's resolve them first, then I will do the second round of review with manual testing.

dmytrostruk (commented):
Here is the outcome of second round of review 😃 . 
Let's resolve the comments first and then we will review again. The amount of new code is quite big, so that's why we need to do that in a couple of rounds.

In case you are working on some comment, please make sure that you resolve the comment in all places of new code and not only in the place where the comment exists. Thanks a lot!

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeClient.cs
```diff
@@ -0,0 +1,759 @@
...
+
+            if (data == null)
+            {
+                this._logger.LogWarning("Unable to deserialize Upsert response");
+                continue;
+            }
+
+            totalUpserted += data.UpsertedCount;
+            this._logger.LogInformation("Upserted batch {0} with {1} vectors", totalBatches, data.UpsertedCount);
+            await Task.Delay(TimeSpan.FromMilliseconds(1000), cancellationToken).ConfigureAwait(true);
```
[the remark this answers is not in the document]
Kevdome3000: not required, just a precaution

Kevdome3000:
> After playing with a few of the sample apps and getting a better understanding of how the IMemoryStore interface is being utilized, I think a refactor makes sense.
> 
> For those familiar with Pinecone, namespaces are really powerful and for those who aren't, this is from their documentation:
> 
> > Pinecone allows you to partition the vectors in an index into namespaces. Queries and other operations are then limited to one namespace, so different requests can search different subsets of your index.
> > For example, you might want to define a namespace for indexing articles by content, and another for indexing articles by title. For a complete example, see: [Semantic Text Search (Example)](https://docs.pinecone.io/docs/semantic-text-search/).
> > Every index is made up of one or more namespaces. Every vector exists in exactly one namespace.
> 
> Imagine a namespace per Kernel or SemanticTextMemory that can be cleared after a chat session while keeping other semantic data that exists in a Pinecone Index in tact. That's the ideal implementation in my opinion.
> 
> This PR in it's current state doesn't allow for that due to the fact that a user can also have multiple Pinecone Indexes, so the "collection"/"collectionName" parameters used throughout the IMemoryStore/ ISemanticTextMemory interface becomes ambiguous.
> 
> The easiest solution is to agree that a user is restricted to a single index per PineconeMemoryStore/ PineconeClient instance so that the name of the index no longer needs to be passed as a parameter and the collectionName/collection can be synonymous with namespace.
> 
> I hope that makes sense.
> 
> If on board let me know. I can turn that around quickly

I'd like to address this if all issues are resolved

shawncal:
Hey @Kevdome3000, did you intentionally close this PR, or was that just GitHub being helpful? 

If you need help with a merge or something, don't hesitate to reach out.

Kevdome3000:
Oh no! That was not intentional. How can we recover?

> 
> Be lucky,
> 
> 
> 
> 
> Kevin Bass
> Email: ***@***.***
> Cell: (310)292-9911
> Skype: Kevdome
> 
> 
> ------------------------------------------------------------------
> *CONFIDENTIALITY NOTICE*
> 
> This e-mail transmission may contain confidential information that is
> legally privileged. Do not read this e-mail if you are not the intended
> receipt. This e-mail transmission, and any documents, files or previous
> e-mail messages attached to it may contain confidential information that
> is legally privileged. If you are not the intended recipient, or a person
> responsible for delivering it to the intended recipient, you are hereby
> notified that any use, disclosure, copying, or distribution of any of the
> information contained in or attached to this transmission is STRICTLY
> PROHIBITED. If you have received this transmission in error, please notify
> sender immediately by reply e-mail. Kindly destroy the original
> transmission and its attachments without reading or saving any of them.
> Thank you.
> 

Sent via Superhuman iOS ( ***@***.*** )

dmytrostruk:
Thanks @Kevdome3000 !
Let's switch to the new PR then. We are close to merging this to `main`.

PR #612, opened 2023-04-24
Clean up exposure and handling of CancellationToken

stephentoub:
### Motivation and Context

Clean up exposure and handling of CancellationToken in public APIs.

### Description

- In public API, they should always be the last parameter (unless they can't be, e.g. due to a params array), named "cancellationToken", and be non-nullable.
- Ensure they're forwarded appropriately from one API to the next. There may be some more missing cases, but I fixed all the ones I found.
- Make the XML docs consistent.
- Remove unnecessary cancellation-related analyzer suppressions

One issue I noticed I haven't addressed here:
- Some APIs take both an SKContext and a CancellationToken separately.  If both are supplied, one of the tokens (either the explicitly passed one or the one from the SKContext) might be ignored.  If they're not the same token and they're both cancelable, a combined token should really be passed on to the target so that the target ends up monitoring both.

### Contribution Checklist


dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -186,7 +186,7 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
             try
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                context = await f.InvokeAsync(context).ConfigureAwait(false);
+                context = await f.InvokeAsync(context, cancellationToken: cancellationToken).ConfigureAwait(false);
```
lemillermicrosoft, line 189: We shouldn't need to make this change since the token is in the `context`, right? Just want to make sure users don't have to change behavior if they are already calling `InvokeAsync` themselves with `context`. Fine with the change here though. Does make me wonder if we should then do similar with `this._log` though.

stephentoub: If it's the same token, yes.  This gets at the issue I mentioned in the description above:
> One issue I noticed I haven't addressed here: Some APIs take both an SKContext and a CancellationToken separately. If both are supplied, one of the tokens (either the explicitly passed one or the one from the SKContext) might be ignored. If they're not the same token and they're both cancelable, a combined token should really be passed on to the target so that the target ends up monitoring both.

The API is designed in such a way that we're potentially now taking in two different tokens and we might end up ignoring one of them.  Ideally that wouldn't be possible by construction, e.g. if SKContext is the way we want to pass around a contained token, then a) APIs taking SKContext shouldn't also take a CancellationToken and b) it should be possible to easily create a new SKContext that contains a CancellationToken the dev wants to use.  Alternatively, we need to ensure that if there are two different tokens, that all of them which are cancelable are respected (I haven't done that here).

dluc: By design we should assume that the context has the right cancellation token (managed by the kernel) and that there's no need to pass another one.

dluc (dismissed):
to be reviewed offline

dluc:
Todo: for ISKFunction and Plan it would be nice having the cancellation token as an optional parameter, without ever forcing the user to pass one. There are a lot of changes so I'll have to check the code locally. As a guideline, if the context is available, I would assume it has a valid cancellation token. If the context is created locally though, that's false so we need to check for that scenario. If the context is null and the user didn't pass a cancellation token I think we should log a warning (or an error).

dotnet/src/SemanticKernel/Orchestration/SKFunction.cs
```diff
@@ -159,37 +159,35 @@ public FunctionView Describe()
...
         }
 
         context.Variables.Update(input);
 
-        return this.InvokeAsync(context, settings, log, cancel);
+        return this.InvokeAsync(context, settings, log, cancellationToken);
     }
 
     /// <inheritdoc/>
     public Task<SKContext> InvokeAsync(
```
shawncal, line 181: There's some funny, misleading stuff going on in these two InvokeAsync() methods (predates this PR).

If the user passes all the args, 2 of them will be ignored unless one (context) is also null. If the user passes a default (empty) SKContext, thinking nothing of it except that they're filling in the blanks, why would they think their logger or cancellation token would be ignored?

Looks to me like these should each be split into overloads: 
- One with only SKContext (non-nullable), settings, and (for method above) input. 
- One with no log and cancellationToken (and input), and no SKContext.

@stephentoub  if you agree, I'll file a task and have someone make the change.

stephentoub: > @stephentoub if you agree

Yup.  I hadn't noticed it for the logger, but this is the crux of the issue I was commenting on for the cancellation token.  There should only be one way to pass in a given piece of information; if there are two, either one gets ignored (which is bad) or both should be used (which is complicated).

So basically if an SKContext is accepted, that same method shouldn't also take anything contained in the SKContext, e.g. the cancellation token.

dotnet/src/SemanticKernel/Kernel.cs
dluc: considering that `context` is created locally using the `cancellationToken` passed in to the function, I would use `context.CancellationToken` to avoid the ambiguity.

dotnet/src/SemanticKernel/Orchestration/Plan.cs
```diff
@@ -283,21 +283,21 @@ public FunctionView Describe()
...
-        CancellationToken? cancel = null)
+        CancellationToken cancellationToken = default)
     {
-        context ??= new SKContext(new ContextVariables(), null!, null, log ?? NullLogger.Instance, cancel ?? CancellationToken.None);
+        context ??= new SKContext(new ContextVariables(), null!, null, log ?? NullLogger.Instance, cancellationToken);
 
         if (this.Function is not null)
         {
-            var result = await this.Function.InvokeAsync(context, settings, log, cancel).ConfigureAwait(false);
+            var result = await this.Function.InvokeAsync(context, settings, log, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
stephentoub: @dluc, this is the same issue discussed in https://github.com/microsoft/semantic-kernel/pull/612#discussion_r1179600065 and https://github.com/microsoft/semantic-kernel/pull/612#discussion_r1175389299.  This is a public method on a public type; just ignoring the `log` and the `cancellationToken` argument if `context` is not null is very unexpected.  If we're going to have multiple ways to pass something in, both should be respected if they're not the same; otherwise, we shouldn't have multiple ways to pass something in, e.g. we should remove `log` and `cancellationToken` from any methods that also take a `context`.

dotnet/src/SemanticKernel/Orchestration/SKFunction.cs
shawncal: Thanks! I'll take care of these InvokeAsync overloads

shawncal (approved):

Issue #616, opened 2023-04-24
Operators on ScoredValue produce inconsistent answers


stephentoub:
`ScoredValue<T>` contains both a `T` and a `double`.  Its `==` and `!=` operators factor in equality of both the `T` and the `double`, but its `<`, `<=`, `>=`, and `>` operators only factor in the `double`.  That means, for example, `<=` might return true in situations where `==` returns false, if the `T` in the two instances being compared isn't equal.

dluc:
@awharrison-28 FYI

awharrison-28:
@dmytrostruk

dmytrostruk:
Thanks, I added this issue to our backlog.

PR #617, opened 2023-04-24
Tweak Score and ScoredValue

stephentoub:
### Motivation and Context

A bit of cleanup for the `Score` and `ScoredValue` types.

### Description

- Make them both readonly, removing the setters from properties
- Give Score a public ctor
- Make the Min static properties internal (it doesn't really make sense on ScoredValue, and it's strange to have a Min but no Max)
- Make ScoredValue's implicit operators for casting to T or double to instead be explicit; implicit operators shouldn't be lossy, but by definition throwing away half of the state is lossy
- Remove unnecessary null checks

I also opened https://github.com/microsoft/semantic-kernel/issues/616.

And, do we need `Score` at all?  It's really just giving a unit of measure to a `double`.  Could we delete it and just let `ScoredValue<T>` wrap `T` and a `double`?

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/Memory/Collections/Score.cs
+++ b/dotnet/src/SemanticKernel/Memory/Collections/Score.cs
@@ -8,15 +8,20 @@ namespace Microsoft.SemanticKernel.Memory.Collections;
 /// <summary>
 /// Structure for storing score value.
 /// </summary>
-public struct Score : IComparable<Score>, IEquatable<Score>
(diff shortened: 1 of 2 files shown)


dluc:
> implicit operators shouldn't be lossy

I love this guideline, thanks

dluc:
> And, do we need Score at all? It's really just giving a unit of measure to a double. Could we delete it and just let ScoredValue<T> wrap T and a double?

@awharrison-28 thoughts?

PR #632, opened 2023-04-25
Kernel Config Services Refactor

shawncal:
### Motivation and Context
The current services registration is hard-coded to support only 4 interface types (ITextCompletion, ITextEmedding...). This is not going to be extensible in the long run, as new model types will come onboard, and even within a category (i.e. image generation) not all model providers will have the same requirements, interfaces.

Furthermore, there are additional service types we'd like to manage centrally (HttpClients, Logging, etc.) making them accessible to APIs throughout the kernel -- this change allows for general services management.

### Description
- Implement a simple named service registry
- KernelConfig "is a" ServiceRegistry.
- Modeled after the System.Extensions.DependencyInjection types, splitting the ServiceCollection (registration) and ServiceProvider (getting instances). The code is split such that these are in the same class, but implemented in partial classes. If we want to make KernelBuilder where registrations happen, and KernelConfig a read-only source of registered services (ie KernelBuilder.Build() produces an immutable ServiceProvider), this will be a simple split.
- Note: considered using System.Extensions.DependencyInjection. This does not provide the named service support we need, and it brings in another library that's much heavier than we need.
- Added lots of UTs, updated callers.

### Contribution Checklist


dluc (changes requested):
We've been talking about the need to fix this for a while, and use ServiceCollection, providing a solution that fits naturally with ASP.NET apps. Why the different/custom approach?

dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/ITextEmbeddingService.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.AI.Embeddings;
+
+/// <summary>
+/// Represents a generator of text embeddings of type <c>float</c>.
+/// </summary>
+public interface ITextEmbeddingService : IEmbeddingGenerationService<string, float>
```
dluc, line 8: this is missing "Generation". Overall I really dislike the long names and having the "Service" suffix everywhere, it makes the API verbose and pedantic. If we touch this again we should use the opportunity to simplify.
Using ServiceCollection it would look something like `Kernel.Services.Add<ITextCompletion, ...>(...)`

dluc: looking at some examples:
* ILogger vs ILoggingService
* ITextRendering vs ITextRenderingService
* IOpenApiDocumentParser vs IOpenApiDocumentParserService

we could use
* `ITextEmbeddingGenerator` instead of `ITextEmbeddingGenerationService`
* `ITextGenerator` instead of `ITextCompletion` / `ITextCompletionService`
* `IChatMessageGenerator` instead `IChatCompletion` / `IChatCompletionService`
* etc.

SergeyMenshykh: My 5 cents:

I think the Service suffix does a very good job of denoting that the interfaces/classes having it are related to the infrastructure layer and has nothing to do with domain layer - Kernel. That, BTW, allow quickly identify and prevent cases of domain logic spilling into the services during development or on PR level.

So, the verbosity is not necessarily a bad thing as long at service a dedicated purpose and has a meaning.

shawncal: I'd be happy with *Generator suffix, as it fits the pattern nicely.

Rationale:
ILogg**er** is a service that generates Logs
IOpenApiDocumentPars**er** is a service that parses documents
A TextCompletion is the product of a service -- not the producer. It's literally the object you get back. If I sit down and start intellisensing with an ITextCompletion object, I'm certainly not going to expect it to be a conduit for hitting a service... I'd expect it to have string value, tokens used, etc.

shawncal:
> We've been talking about the need to fix this for a while, and use ServiceCollection, providing a solution that fits naturally with ASP.NET apps. Why the different/custom approach?

Called out in the PR notes -- 

System.Extensions.DependencyInjection does not provide a facility for **named** services, which we use extensively. If we were to adopt it, we'd effectively just be registering Dictionaries<string, T> as services, and then doing lookups on those anyway -- it doesn't it's another dependency nuget package that everyone using SK would need to pull in, and we'd have to make sure we're not conflicting versions with the host app, and it doesn't really gain us any utility.

It would however invite some unintentional practices -- developers could start using the many extension methods and packages that support IServiceCollection, filling SK full of things they expect us to consume internally -- which we won't do.

As @SergeyMenshykh points out, the patterns of Dependecy Injection and IServiceCollection belong to the host app. They should call those as they register AI Providers, instantiate skills objects, etc. with SK.

dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/ITextEmbeddingService.cs
shawncal: ITextGenerator isn't sufficiently descriptive or accurate either. ITextCompletionGenerator is fine, but is longer than *Service.

If we don't like *Service, we could use:

ITextEmbeddingGenerator
IImageGenerator 
ITextCompletionGenerator or ITextCompletionProvider
IChatCompletionGenerator or IChatCompletionProvider

shawncal: Reverted. We can cover services naming in another PR.

dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/TextEmbeddingServiceExtensions.cs
```diff
@@ -0,0 +1,135 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.SemanticKernel.AI.Embeddings;
+using Microsoft.SemanticKernel.Services;
+
+// Use base namespace for better discoverability and to avoid conflicts with other extensions.
+namespace Microsoft.SemanticKernel;
```
shawncal, line 9: I initially had these extension classes all in their "proper" namespaces, but it made the extension undiscoverable... I had to add additional explicit usings to dozens of files, and intellisense was no help at all.

@dluc I know your IDE will squawk if these files are in these folders with mismatched namespaces. Is there a suppression we can apply? Alternatively, we could put all of these classes in one huge file in the project root, but I think it's kind of nice having them hang out with their relatives.

dotnet/src/SemanticKernel/Orchestration/SKFunction.cs
```diff
@@ -209,7 +209,7 @@ public ISKFunction SetAIService(Func<ITextCompletion> serviceFactory)
     {
         Verify.NotNull(serviceFactory, "AI LLM service factory is empty");
         this.VerifyIsSemantic();
-        this._aiService = serviceFactory.Invoke();
+        this._aiService = new Lazy<ITextCompletion>(serviceFactory);
```
shawncal, line 212: @dluc Please take a closer look here... this was setting as a func() so that it could be delay-loaded, but then it was invoking it immediately anyway, upon semantic function registration. One effect was that, if you registered functions before AI connectors, it would throw.

Note: even with the "Lazy" approach, if you change your default completions service after or between calls to the function, the change will not be reflected. Another approach would be to store this._aiServiceFactory here and only construct/dispose upon each call to the function.

shawncal: Update: I see from ItFailsIfTextCompletionServiceConfigIsNotSet that it's expected to throw if you register a function before a completion service. Do we want to stick with that (even though you could feasibly remove the completion service right after registering the function)?

dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/TextEmbeddingServiceExtensions.cs
dluc: for extensions methods I've been suppressing, there's a few examples in the repo

dluc:
We should not create another DI just because we dislike what .NET offers, or because we want to isolate SK from the rest of the applications where it is used. The .NET DI solution is mature, well-documented, and familiar to developers. It also supports named binding extensions if needed, leveraging third party containers, and as far as I know, we only use named binding for AI in testing scenarios. 
SK should integrate seamlessly with existing ASP.NET apps. If developers configure their DI to select a specific logger, HTTP client factory, or any other component, SK should adhere to that configuration instead of requiring developers to configure dependencies twice.

dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/TextEmbeddingServiceExtensions.cs
shawncal: Thanks for the pointers.

SergeyMenshykh:
> We should not create another DI just because we dislike what .NET offers, or because we want to isolate SK from the rest of the applications where it is used. The .NET DI solution is mature, well-documented, and familiar to developers. It also supports named binding extensions if needed, leveraging third party containers, and as far as I know, we only use named binding for AI in testing scenarios. SK should integrate seamlessly with existing ASP.NET apps. If developers configure their DI to select a specific logger, HTTP client factory, or any other component, SK should adhere to that configuration instead of requiring developers to configure dependencies twice.

A few rules that would allow us to avoid a lot of problems and have a very flexible SDK:
- SK SDK should not make any assumptions about hosting app/client code. 
- SK SDK should not expect hosting app/client code to use or not to use DI container.
- SK SDK should not expect hosting app/client code to use or not to use HttpClientFactory/HttpClient.
- SK SDK should have integration points that demark SK SDK boundaries.
- SK SDK should have integration/extension points to allow hosting app/client code to plug required dependencies.

dluc:
> SK SDK should have integration points that demark SK SDK boundaries.

This would make for a bad developer experience, increasing the amount of configuration.

> SK SDK should not expect hosting app/client code to use or not to use HttpClientFactory/HttpClient.

HttpClientFactory is a mature solution that we should leverage. It's also a well known for resilient architectures. Expecting people to use it is a fair assumption, leads uses into proven solutions, and allows us to leverage work already done by MSFT.

> SK SDK should have integration points that demark SK SDK boundaries.

Why? SK should integrate seamlessly into standard .NET solutions, ie drop-in and it works without any effort other than providing some API keys.


* Ideally SK should integrate seamlessly into .NET apps without changing or adding any code.
* SK dependency configuration should follow the existing configuration of the hosting apps, particularly when these hosting apps use the standard .NET dependency injection. In other words, developers shouldn't have to configure dependencies twice, once for ASP.NET and once for SK. 

The problems we should solve are:
* The current DI is limited to AI connectors. This is code inherited from initial prototypes focused just on switching models, rejecting the idea of DI in the first place. Since then the code has gone through a few iterations to make it easier to fit into the standard .NET DI. The AI side of the code is almost there. There is still a lot of work in other sides of the SK.
* SK DI (AI, HTTP, Logger) doesn't integrate with hosting apps, asking developers to configure DI twice. The most evident part is HTTP, with several users asking to inject a custom HTTP client, disable SSL certs verification, etc. The second most problematic part is the number of HTTP clients being created, with the risk of running out of sockets. And then logging, where SK logging doesn't automatically follow the logging rules of the hosting app. And unit tests, which often cannot mock dependencies, leading for example to tests depending on code under tests, flaky tests, lack of coverage.

PR #635, opened 2023-04-25
Move authority out from Constants.ts to the .env (env.example)

craigomatic:
### Motivation and Context
Small PR to shift the AAD authority out of the Constants.ts file to the .env alongside the AAD client id. Developers may wish to change the authority from /common to the MSA endpoint /9188040d-6c67-4c5b-b112-36a304b66dad if their AAD app is for consumer MSA logins or to their own tenant for other reasons.

### Description
Change only the location of the default authority to the env.example file. The expectation is that the developer will rename the env.example file to .env

### Contribution Checklist


hathind-ms (approved):

samples/apps/copilot-chat-app/webapp/env.example
```diff
@@ -1,2 +1,3 @@
 REACT_APP_BACKEND_URI=https://localhost:40443/
+REACT_APP_AAD_AUTHORITY=https://login.microsoftonline.com/common
```
adrianwyatt, line 2: Since this is currently bound to AAD, convention would be to have this just be the tenant in place of the the whole authority URI.
```
REACT_APP_AAD_TENANT=common
```

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapp/env.example
craigomatic: I was following the msal.js convention: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/msal-react-samples/react-router-sample/.env.development

They typically refer to authority and include the entire string:

image

adrianwyatt: Cool - sounds good. I remember now that MSAL.js supports non-AAD while MSAL.NET supports only AAD, so that's probably where the friction is. Your changes looks good.

adrianwyatt (approved):

PR #641, opened 2023-04-25
Open API GitHub Plugin Skill + Frontend Auth (MSAL + Basic / PAT)

teresaqhoang:
### Motivation and Context

This PR adds:

- Swagger files detailing List and Get Pull Request(s) functions. 
- Example 22 c on how to leverage imported OpenAPI GitHub Skill
- Basic (+ User PAT) and Msal Authentication flows in Copilot chat for connector plug-ins, sending auth data if user enables plug-in
- Parsing headers in SemanticKernelController to import respective OpenAPI Skill if auth information is provided

### Description

image
image
image
image
image

### Contribution Checklist


dotnet/src/Skills/Skills.OpenAPI/Skills/GitHubSkill/openapi.json
amsacha, line 1: File uploaded twice

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -102,4 +109,34 @@ public SemanticKernelController(IServiceProvider serviceProvider, IConfiguration
...
     }
+
+    private async Task RegisterOpenApiSkills(PluginAuthHeaders pluginAuthHeaders, Kernel kernel)
+    {
+        if (pluginAuthHeaders.GithubAuthentication != null)
+        {
+            var authenticationProvider = new BearerAuthenticationProvider(() => { return Task.FromResult(pluginAuthHeaders.GithubAuthentication); });
+            this._logger.LogInformation("Registering GitHub Skill, token {0}", pluginAuthHeaders.GithubAuthentication);
+
+            var skill = await kernel.ImportOpenApiSkillFromResourceAsync(SkillResourceNames.GitHub,
```
amsacha, line 120: Can we switch to using either 
 kernel.ImportOpenApiSkillFromFile
or kernel.ImportOpenApiSkillFromUrlAsync instead?
Importing from a file or URL is more straight forward and requires fewer modifications to the code files. I also envision developers just being able to point to a URL where there schema exists and taking it from there especially if they are working with 3rd party schemas.
Using ImportOpenApiSkillFromFile is more appropriate here than the URL just because we want to showcase a limited set of operations that we have verified work.
This should also make it so that you dont have to touch files like SkillResourceName.cs

dotnet/src/Skills/Skills.OpenAPI/Skills/SkillResourceNames.cs
```diff
@@ -11,4 +11,9 @@ public static class SkillResourceNames
     /// Azure KeyVault skill name.
     /// </summary>
     public const string AzureKeyVault = "AzureKeyVaultSkill";
+
```
amsacha, line 14: dont need to modify this file if you use either
kernel.ImportOpenApiSkillFromFile
or kernel.ImportOpenApiSkillFromUrlAsync .
More details below

samples/apps/copilot-chat-app/webapp/yarn.lock
amsacha, line 1: should this be in the PR?

samples/apps/copilot-chat-app/webapp/src/redux/features/plugins/PluginsState.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const enum Plugins {
+    MsGraph = 'Microsoft Graph',
+    Jira = 'Jira',
+    GitHub = 'GitHub',
+}
+
+export const enum AuthHeaderTags {
+    MsGraph = 'graph',
+    Jira = 'gira',
```
amsacha, line 12: typo, Jira = 'jira'

samples/dotnet/kernel-syntax-examples/Program.cs
amsacha, line 1: To have all the openAPI skills show up one after another, consider renaming to Example22_x_OpenApiSkill_Github

dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Skills/GitHubSkill/openapi.json
```diff
@@ -0,0 +1,5935 @@
+{
```
adrianwyatt, line 1: Is there a URL we could get this from instead of checking in our own copy?

dotnet/src/Skills/Skills.OpenAPI/Skills.OpenAPI.csproj
```diff
@@ -17,7 +17,9 @@
 
   <ItemGroup>
     <None Remove="Skills\AzureKeyVaultSkill\openapi.json" />
+    <None Remove="Skills\GitHubSkill\openapi.json" />
     <EmbeddedResource Include="Skills\AzureKeyVaultSkill\openapi.json" />
+    <EmbeddedResource Include="Skills\GitHubSkill\openapi.json" />
```
adrianwyatt, line 22: Let's not perpetuate a pattern of bringing in OpenAPI definitionas to the OpenAPISkill project. We should be able to keep all of this within CopilotChat.

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace SemanticKernel.Service.Plugins;
+
+public class PluginAuthHeaders
```
adrianwyatt, line 5: XMLDoc for the class and properties.

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -102,4 +109,34 @@ public SemanticKernelController(IServiceProvider serviceProvider, IConfiguration
...
+             authenticationProvider.AuthenticateRequestAsync);
+        }
+
+        // if (pluginAuthHeaders.GraphAuthentication != null)
```
adrianwyatt, line 124: I appreciate the guidance, but let's keep these dead code sections out of the codebase until we bring in each plugin.

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -102,4 +109,34 @@ public SemanticKernelController(IServiceProvider serviceProvider, IConfiguration
 
         return this.Ok(new AskResult { Value = result.Result, Variables = result.Variables.Select(v => new KeyValuePair<string, string>(v.Key, v.Value)) });
     }
+
+    private async Task RegisterOpenApiSkills(PluginAuthHeaders pluginAuthHeaders, Kernel kernel)
+    {
+        if (pluginAuthHeaders.GithubAuthentication != null)
+        {
+            var authenticationProvider = new BearerAuthenticationProvider(() => { return Task.FromResult(pluginAuthHeaders.GithubAuthentication); });
+            this._logger.LogInformation("Registering GitHub Skill, token {0}", pluginAuthHeaders.GithubAuthentication);
```
adrianwyatt, line 118: Does this log the token or password? We should make sure and never log credentials.

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -40,6 +43,7 @@ public SemanticKernelController(IServiceProvider serviceProvider, IConfiguration
     /// <param name="chatRepository">Storage repository to store chat sessions</param>
     /// <param name="chatMessageRepository">Storage repository to store chat messages</param>
     /// <param name="ask">Prompt along with its parameters</param>
+    /// <param name="pluginAuthHeaders"></param>
```
glahaye, line 46: XML Doc - Maybe:

```suggestion
    /// <param name="pluginAuthHeaders">Auth headers for plugins to use</param>
```

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    [FromHeader(Name = "x-sk-copilot-graph-authorization")]
```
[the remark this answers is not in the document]
glahaye: I sidestep this problem by just calling my things "something-something-auth" :)

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
```diff
@@ -0,0 +1,15 @@
...
+    public string? GraphAuthentication { get; set; }
+
+    [FromHeader(Name = "x-sk-copilot-jira-authorization")]
+    public string? JiraAuthentication { get; set; }
+
+    [FromHeader(Name = "x-sk-copilot-github-authorization")]
```
glahaye, line 13: Not critical for now, but consider having a generic way to extract auth headers, like adding all headers that follow the "x-sk-copilot-XXX-authorization" format to a dictionary.  This way, when we get new plug-ins, we don't have to modify this object. This would involved the creation of a method to be invoked in the relevant controllers.

samples/dotnet/kernel-syntax-examples/Example27_OpenApiGithubSkill.cs
```diff
@@ -0,0 +1,96 @@
...
+        var pullRequests = JArray.Parse((string)resultJson!["content"]);
+
+        if (pullRequests != null && pullRequests.First != null)
+        {
+            var number = pullRequests.First["number"];
+            return number?.ToString() ?? "-1";
+        }
+        else
+        {
+            return "-1";
```
adrianwyatt, line 74: Why not return string.empty?

adrianwyatt (changes requested):

dotnet/src/Skills/Skills.OpenAPI/Skills/GitHubSkill/openapi.json
teresaqhoang: Removed both from Skills.OpenAPI, uploaded to under copilot-chat skills

dotnet/src/SemanticKernel.Abstractions/Connectors/WebApi/Rest/RestApiOperationRunner.cs
```diff
@@ -84,6 +84,8 @@ public RestApiOperationRunner(HttpClient httpClient, AuthenticateRequestAsyncCal
             requestMessage.Content = payload;
         }
 
+        requestMessage.Headers.Add("User-Agent", "semantic-kernel");
```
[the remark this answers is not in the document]
teresaqhoang: Discussed offline -- it's super cumbersome to extract the correct package version, so will leave as is

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: Done

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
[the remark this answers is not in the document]
teresaqhoang: Simplified to just auth

samples/apps/copilot-chat-app/webapp/yarn.lock
teresaqhoang: Yeah I added the buffer package to Base64 encode the Jira Auth header

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
teresaqhoang: For now, it's a better to keep the PluginAuthHeaders class to keep all the plug-ins well defined. I think your proposal is great for discovery, but we still need to know exactly what plug-ins to expect in order to register them properly since each can require different Authentication methods and we have to instantiate the right one.  i.e., Jira uses basic and Graph uses Bearer 

Not priority right now, will note to revisit in the future (i.e., have a header send in which skills need to be registered to run the query)

dotnet/src/SemanticKernel.Skills/Skills.OpenAPI/Skills/GitHubSkill/openapi.json
teresaqhoang: afaik no, GH just has one massive specification for all of their APIs so I had to reduce this manually: https://github.com/github/rest-api-description/tree/main/descriptions

dotnet/src/Skills/Skills.OpenAPI/Skills.OpenAPI.csproj
teresaqhoang: Removed this and moved spec to CopilotChat

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: Removed

samples/apps/copilot-chat-app/webapp/src/redux/features/plugins/PluginsState.ts
```diff
@@ -0,0 +1,80 @@
+import GithubAvatar from '../../../assets/github-icon.png';
+import JiraAvatar from '../../../assets/jira-icon.png';
```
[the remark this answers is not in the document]
teresaqhoang: Done

samples/apps/copilot-chat-app/webapp/src/redux/features/plugins/PluginsState.ts
```diff
@@ -0,0 +1,80 @@
...
+    MsGraph = 'Microsoft Graph',
+    Jira = 'Jira',
+    GitHub = 'GitHub',
+}
+
+export const enum AuthHeaderTags {
+    MsGraph = 'graph',
+    Jira = 'gira',
+    GitHub = 'github',
+    Ado = 'ado',
```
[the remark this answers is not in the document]
teresaqhoang: Removed

samples/dotnet/kernel-syntax-examples/Example27_OpenApiGithubSkill.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    public static async Task RunAsync()
+    {
+        var authenticationProvider = new BearerAuthenticationProvider(() => getToken());
+        Console.WriteLine("Example27_OpenApiGitHubSkill");
+        var firstPRNumber = await ListPullRequestsFromGitHubAsync(authenticationProvider);
+        await GetPullRequestFromGitHubAsync(authenticationProvider, firstPRNumber);
+    }
+
+    private static Task<string> getToken()
```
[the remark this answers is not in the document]
teresaqhoang: Done

samples/apps/copilot-chat-app/webapp/src/components/open-api-plugins/BasicAuthPluginButton.tsx
```diff
@@ -0,0 +1,159 @@
+import {
```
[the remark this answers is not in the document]
teresaqhoang: Completely changed the UX so no longer relevant

dotnet/src/SemanticKernel.Abstractions/Connectors/WebApi/Rest/RestApiOperationRunner.cs
```diff
@@ -84,6 +84,8 @@ public RestApiOperationRunner(HttpClient httpClient, AuthenticateRequestAsyncCal
             requestMessage.Content = payload;
         }
 
+        requestMessage.Headers.Add("User-Agent", "Semantic-Kernel");
```
dluc, line 87: what about "Microsoft Semantic Kernel"?

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -54,6 +57,7 @@ public SemanticKernelController(IServiceProvider serviceProvider, IConfiguration
         [FromServices] ChatSessionRepository chatRepository,
         [FromServices] ChatMessageRepository chatMessageRepository,
         [FromBody] Ask ask,
+        [FromHeader] PluginAuthHeaders pluginAuthHeaders,
```
dluc, line 60: what is "plugin" in this context?

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace SemanticKernel.Service.Plugins;
+
+/// /// <summary>
+/// Represents the authentication headers imported Open API Plugin Skills.
```
dluc, line 6: OpenAPI or OpenAI?

dluc (dismissed):
there's some confusion about mixing plugins and skills terminology, open AI and open API. I'd like to understand what story we're telling with this sample, e.g. if it's ok to talk about all of these concepts, and how we clarify differences between SK skills, Open AI plugins, OpenAPI swaggers. Just writing it here makes my head spin...

please also fix code style, e.g. separating blocks with empty lines

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
teresaqhoang: Open API -- these headers contain the authentication information to connect to skills imported from Swagger specs

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: Open API Skills (imported from Swagger specs)

samples/apps/copilot-chat-app/webapi/Model/PluginAuthHeaders.cs
teresaqhoang: Renamed it to make it more clear

dotnet/src/SemanticKernel.Abstractions/Connectors/WebApi/Rest/RestApiOperationRunner.cs
teresaqhoang: done

dotnet/src/Skills/Skills.OpenAPI/Skills/SkillResourceNames.cs
teresaqhoang: Done

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: Renamed it to make it more clear

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -7,9 +7,11 @@
 using Microsoft.SemanticKernel;
 using Microsoft.SemanticKernel.AI;
 using Microsoft.SemanticKernel.Orchestration;
+using Microsoft.SemanticKernel.Skills.OpenAPI.Authentication;
 using SemanticKernel.Service.Config;
 using SemanticKernel.Service.Model;
 using SemanticKernel.Service.Skills;
+using SemanticKernel.Service.Skills.OpenAPI.Authentication;
```
amsacha, line 14: Do we need both SemanticKernel.Service.Skills.OpenAPI.Authentication and Microsoft.SemanticKernel.Skills.OpenAPI.Authentication ?

samples/apps/copilot-chat-app/webapp/src/libs/services/BaseService.ts
```diff
@@ -12,16 +14,25 @@ export class BaseService {
...
             Authorization: `Bearer ${accessToken}`,
             'Content-Type': 'application/json',
         });
 
-        //TODO: not sure what scenario this helps with?
-        if (connectorAccessToken) headers.append(`sk-copilot-connector-access-token`, connectorAccessToken);
+        // For each enabled plugin, pass its auth information as a customer header
+        // to the backend so the server can authenticate to the plugin
+        if (enabledPlugins && enabledPlugins.length > 0) {
+            // eslint-disable-next-line
```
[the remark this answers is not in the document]
teresaqhoang: changed to for loop

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: The former is the local copilot chat namespace -- decided to keep the GitHub skill confined to copilot chat

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: It was logging the token but I removed the auth header value log altogether

dotnet/src/SemanticKernel.Abstractions/Connectors/WebApi/Rest/RestApiOperationRunner.cs
```diff
@@ -84,6 +84,8 @@ public RestApiOperationRunner(HttpClient httpClient, AuthenticateRequestAsyncCal
             requestMessage.Content = payload;
         }
 
+        requestMessage.Headers.Add("User-Agent", "Microsoft-Semantic-Kernel");
```
adrianwyatt, line 87: Could you add a comment on why we want to attach this header for *all* REST api operations?

samples/apps/copilot-chat-app/webapi/Skills/GitHubOpenApiSkill/openapi.json
```diff
@@ -0,0 +1,5935 @@
+{
```
adrianwyatt, line 1: [nit] Could we put a README next to this file the has a link to where we got it from and if/how it's different from the original.

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -119,4 +125,17 @@ public class SemanticKernelController : ControllerBase
 
         return this.Ok(new AskResult { Value = result.Result, Variables = result.Variables.Select(v => new KeyValuePair<string, string>(v.Key, v.Value)) });
     }
+    private async Task RegisterOpenApiSkillsAsync(OpenApiSkillsAuthHeaders openApiSkillsAuthHeaders, IKernel kernel)
```
adrianwyatt, line 128: The planner registers skills for itself through a factory: https://github.com/microsoft/semantic-kernel/blob/main/samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs#L95

With your pattern here we may need to rearchitect how we create the planner. Let's chat tomorrow about it!

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Skills/GitHubOpenApiSkill/openapi.json
amsacha: Grouping all the open api schemas we add (so far just github and jira) under a single folder, Skills/OpenApiSkills/... , might be cleaner. That way we can just have a single bigger readme for them rather than multiple 2-3 line readmes

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: RegisterPlannerSkillsAsync now calls RegisterOpenApiSkillsAsync

samples/apps/copilot-chat-app/webapi/Skills/GitHubOpenApiSkill/openapi.json
teresaqhoang: Added new folder /Skills/OpenApiSkills/ with README

dotnet/src/SemanticKernel.Abstractions/Connectors/WebApi/Rest/RestApiOperationRunner.cs
teresaqhoang: Made it optional -- RestApiOperationRunner now accepts the header value as an optional parameter of the class constructor allowing a client code to configure it if needed. The RestApiOperationRunner does not contain any SK SDK specific logic so far so this change is to keep it that way keep it that way.

Now this header is passed in KernelOpenApiExtensions with comment detailing requirement

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -128,13 +132,32 @@ public class SemanticKernelController : ControllerBase
         return this.Ok(new AskResult { Value = result.Result, Variables = result.Variables.Select(v => new KeyValuePair<string, string>(v.Key, v.Value)) });
     }
 
+    /// <summary>
+    /// Register authenticated OpenAPI skills with the planner's kernel.
+    /// If the request includes an auth header for an OpenAPI skill, register the skill with the kernel.
+    /// Else, don't register the skill as it'll fail on auth.
+    /// </summary>
+    private async Task RegisterOpenApiSkillsAsync(CopilotChatPlanner planner, OpenApiSkillsAuthHeaders openApiSkillsAuthHeaders)
```
glahaye, line 140: This is to register the Github skill, or the OpenAPI skill?

Or is it registering the Github skill as an OpenAPI skill?

amsacha: Github also uses the open api schema; We will be adding jira here as well; This is basically where you register skills with the kernel with a call like ImportOpenApiSkillFromFileAsync(...)
I think it will be less confusing when we also have jira in here.

teresaqhoang: @glahaye we can hop on a call to clarify this more, but there's a couple of overloaded references here. 

- We have the actual OpenAPI SK Skill that takes an Open API / Swagger spec and parses through it to make a new skill and functions out of the API definitions. 
- These imported skills are referred as OpenAPI Skills, i.e., GitHub OpenAPI Skills. To help differentiate it from the skill defined in (1), we started referring to these as Plugins, so GitHub Plug-in, which is the frontend terminology.

In short, the OpenAPI SK Skill parses and imports from OpenAPI swagger to make Plug-in (OpenAPI Skill). 

You might also hear the plugins referenced as connectors but that got way too confusing with the actual connector skills, so we introduced the 'plug-in' concept

glahaye: Got it!

dotnet/src/SemanticKernel.Abstractions/Connectors/WebApi/Rest/RestApiOperationRunner.cs
```diff
@@ -84,7 +91,10 @@ public RestApiOperationRunner(HttpClient httpClient, AuthenticateRequestAsyncCal
             requestMessage.Content = payload;
         }
 
-        requestMessage.Headers.Add("User-Agent", "Microsoft-Semantic-Kernel");
+        if (this._userAgent != null)
```
adrianwyatt, line 94: For strings, let's check "string.IsNullOrWhitespace" unless you want to support empty strings here.

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
```diff
@@ -252,7 +252,10 @@ async Task<SKContext> ExecuteAsync(SKContext context)
         {
             try
             {
-                var runner = new RestApiOperationRunner(new HttpClient(), authCallback);
+                // User Agent may be a required request header fields for some Rest APIs,
+                // but this detail isn't specified in OpenAPI specs, so including for all Rest APIs imported
+                var userAgent = "Microsoft-Semantic-Kernel";
```
adrianwyatt, line 257: Would be helpful for other applications to be able to override this value? Consider making this a parameter with a default value.

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -133,17 +131,40 @@ public class SemanticKernelController : ControllerBase
...
         return this.Ok(new AskResult { Value = result.Result, Variables = result.Variables.Select(v => new KeyValuePair<string, string>(v.Key, v.Value)) });
     }
-    private async Task RegisterOpenApiSkillsAsync(OpenApiSkillsAuthHeaders openApiSkillsAuthHeaders, IKernel kernel)
+
+    /// <summary>
+    /// Register authenticated OpenAPI skills with the planner's kernel.
+    /// If the request includes an auth header for an OpenAPI skill, register the skill with the kernel.
+    /// Else, don't register the skill as it'll fail on auth.
+    /// </summary>
+    private async Task RegisterOpenApiSkillsAsync(CopilotChatPlanner planner, OpenApiSkillsAuthHeaders openApiSkillsAuthHeaders)
```
adrianwyatt, line 140: [nit] Could we inline the contents of this function with "RegisterPlannerSkillsAsync" - it'd be good to keep all of this together for the first offering.

adrianwyatt (changes requested):

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
teresaqhoang: Done

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -82,7 +86,7 @@ public class SemanticKernelController : ControllerBase
         // Register skills with the planner if enabled.
         if (plannerOptions.Value.Enabled)
         {
-            await this.RegisterPlannerSkillsAsync(planner, plannerOptions.Value);
+            await this.RegisterPlannerSkillsAsync(planner, plannerOptions.Value, openApiSkillsAuthHeaders);
```
Vivihung, line 89: Just curious and would like to learn more. Why we need to set request header?

samples/apps/copilot-chat-app/webapp/src/redux/features/plugins/pluginsSlice.ts
```diff
@@ -0,0 +1,58 @@
...
+            var plugin: Plugin;
+            var authData = action.payload.accessToken;
+
+            switch (action.payload.plugin) {
+                case Plugins.MsGraph:
+                    plugin = state.MsGraph;
+                    break;
+                case Plugins.Jira:
+                    plugin = state.Jira;
+                    const encodedData = Buffer.from(
```
Vivihung, line 21: Curious to learn what Buffer helps here. If it's just base64 encoding, can't we use the native [btoa()](https://developer.mozilla.org/en-US/docs/Web/API/btoa) API in JavaScript?

Vivihung (commented):
I took a pass on the changes in Copilot Chats. LGTM and have a couple curious questions. I might not be helpful in changes outside of Copilot Chats.

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
```diff
@@ -238,21 +238,26 @@ public static class KernelOpenApiExtensions
...
     /// <param name="cancellationToken">The cancellation token.</param>
+    /// <param name="userAgent">Optional override for request-header field containing information about the user agent originating the request</param>
     /// <returns>An instance of <see cref="SKFunction"/> class.</returns>
     private static ISKFunction RegisterRestApiFunction(
         this IKernel kernel,
         string skillName,
         RestApiOperation operation,
         AuthenticateRequestAsyncCallback? authCallback = null,
-        CancellationToken cancellationToken = default)
+        CancellationToken cancellationToken = default,
```
amsacha, line 248: As the warnings state, make this cancellation token the last param.

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
amsacha: Calls to some open apis (github and jira) will fail without some form of oauth. The headers at least tell us there is some auth combo that the user is trying. Since the openAPI plugins will fail without Oauth it doesnt make sense to expose them as skills to the planner/kernel since it wont be able to use them in its planning anyway.

samples/apps/copilot-chat-app/webapp/src/libs/connectors/useConnectors.ts
```diff
@@ -61,21 +41,72 @@ export const useConnectors = () => {
...
+        );
     };
 
     /**
      * Helper function to invoke SK skills
-     * using ADO token
+     * with GitHub token.
+     */
+    const invokeSkillWithGitHubToken = async () => {
+        // TODO: For testing, change as needed
```
amsacha, line 58: Remove TODO and rephrase it as an example "This is an example of invoking openAPi skills, follow this format:
...
describe the input and variable params and what the map to on the kernel
...
"

samples/apps/copilot-chat-app/webapp/src/redux/features/plugins/pluginsSlice.ts
amsacha: Can we also just pass the `${action.payload.username}:${action.payload.accessToken}` as just a string and not convert to base64 here. We can do that in the back end with the BasicAuthenticationProvider. That will make sure we are converting the string to a UTF8 encoded byte stream before converting to base64. It is also how we are telling users to perform Basic Authentication in the dotnet examples we have.

amsacha (changes requested):

samples/apps/copilot-chat-app/webapp/src/redux/features/plugins/pluginsSlice.ts
teresaqhoang: image
btoa has been deprecated and not recommend for use in new applications

teresaqhoang: On your note @amsacha, I'll leave you to change the header as you see fit with the Jira integration. This will help keep the PR scope limited correctly.

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
```diff
@@ -214,7 +214,7 @@ public static class KernelOpenApiExtensions
             try
             {
                 kernel.Log.LogTrace("Registering Rest function {0}.{1}", skillName, operation.Id);
-                var function = kernel.RegisterRestApiFunction(skillName, operation, authCallback, cancellationToken);
+                var function = kernel.RegisterRestApiFunction(skillName, operation, authCallback, null, cancellationToken);
```
adrianwyatt, line 217: [nit] null parameter value doesn't give the reader much of hint of what is being null'd out. I would recommend using parameter names (e.g. `userAgent: null`).

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
teresaqhoang: @Vivihung + we need to isolate separate auth headers for each plug-ins so the server knows how to map the auth data to the respective plug-ins. If the auth data isn't available, the skill won't be registered

adrianwyatt (approved):

amsacha (approved):

PR #643, opened 2023-04-25
Python SK: Use ChatGPT for Semantic Functions

awharrison-28:
### Motivation and Context
Users want to use ChatGPT as the service for their semantic functions. This feature exists with the C# SK. ChatGPT is perfectly capable of doing this. Additionally added more end-to-end tests with the intention of being able to run them as integration tests for the Python SK. 

### Description
- `OpenAIChatCompletion` now additionally inherits from `TextCompletionClientBase`. This extends to `AzureOpenAIChatCompletion`.
- If a chat service is also a `TextCompletionClientBase`, it will be added to both `text_services` and `chat_services`.
- Created 4 end-to-end tests that target the combination of Azure, OpenAI, text service, and chat service. Abstracted common behavior for these tests. 
- Added asserts to these end-to-end tests. These should be run as integration tests, but that will be handled in a separate PR. 

### Contribution Checklist

### DIFF
--- a/FEATURE_MATRIX.md
+++ b/FEATURE_MATRIX.md
@@ -5,7 +5,7 @@
 |---|---|---|---|
 | TextGeneration                    | ✅ | ✅ | Example: Text-Davinci-003 |
 | TextEmbeddings                    | ✅ | ✅ | Example: Text-Embeddings-Ada-002 |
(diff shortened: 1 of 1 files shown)


alexchaomander (approved):
Looks good to me! Very important feature! In the future, we should offer guidance (both for C# and Python) on when to use ChatCompletion vs TextCompletion.

PR #648, opened 2023-04-25
Jira plugin open api

amsacha:
### Motivation and Context
Provide an example implementation of connecting the Semantic Kernel to external plugins following the open api schema. 

### Description
Jira Open API schema usage example
Renaming the azure key vault example, so all open api skill examples are grouped together 
Prevent RestApiOperationRunner.cs from wrapping a json response into a json object.

### Contribution Checklist


samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        //Run operation via the semantic kernel
+        var result = await kernel.RunAsync(contextVariables, jiraSkills["AddComment"]);
+
+        Console.WriteLine("\n\n\n");
+        var formattedContent = JsonConvert.SerializeObject(JsonConvert.DeserializeObject(result.Result), Formatting.Indented);
+        Console.WriteLine("AddComment jiraSkills response: \n{0}", formattedContent);
+    }
+
+    private static Task<string> AuthenticateWithBasicAPITokenAsync()
```
gitri-ms, line 75: super nit: make Api camel case to match other naming instances (e.g. OpenApiSkill, RestApiOperationRunner)

samples/dotnet/kernel-syntax-examples/Skills/JiraSkill/apiDefinition.swagger.json
```diff
@@ -0,0 +1,2826 @@
+{
+    "swagger": "2.0",
```
gitri-ms, line 2: How is this file different from openapi.json? Is this being referenced by the example?

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,80 @@
...
+        var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
+        var contextVariables = new ContextVariables();
+
+        string s_serverURL = "https://<jiraProject>.atlassian.net/rest/api/latest/";
+        contextVariables.Set("server-url", s_serverURL);
+
+        IDictionary<string, ISKFunction> jiraSkills;
+        var tokenProvider = new BasicAuthenticationProvider(AuthenticateWithBasicAPITokenAsync);
+
+        bool bUseLocalFile = true;
```
gitri-ms, line 31: Do we want to check in this if/else statement? If it's just for our own testing, I'm inclined not to. If you do check it in, I recommend adding a comment about what flipping the bool does/why someone running the example might want to.

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,81 @@
...
+using RepoUtils;
+
+public static class Example22_b_OpenApiSkill_Jira
+{
+    public static async Task RunAsync()
+    {
+        var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
+        var contextVariables = new ContextVariables();
+
+        string s_serverURL = "https://<jiraProject>.atlassian.net/rest/api/latest/";
```
adrianwyatt, line 25: Our code style omits type prefixes. 
```suggestion
        string serverUrl= "https://<jiraProject>.atlassian.net/rest/api/latest/";
```

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,81 @@
...
+        contextVariables.Set("server-url", s_serverURL);
+
+        IDictionary<string, ISKFunction> jiraSkills;
+        var tokenProvider = new BasicAuthenticationProvider(AuthenticateWithBasicApiTokenAsync);
+
+        // The bool bUseLocalFile can be used to toggle the ingestion method for the openapi schema between a file path and a URL
+        bool bUseLocalFile = true;
+        if (bUseLocalFile)
```
adrianwyatt, line 33: The scenario we're trying to show with Jira is being able to pull a Power Platform Connector directly. Could we simplify this to just using the api definition URL and remove the checked-in openapi.json?

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,81 @@
...
+            jiraSkills = await kernel.ImportOpenApiSkillFromFileAsync("jiraSkills", apiSkillFile, tokenProvider.AuthenticateRequestAsync);
+        }
+        else
+        {
+            var apiSkillRawFileURL = new Uri("https://raw.githubusercontent.com/microsoft/PowerPlatformConnectors/dev/certified-connectors/JIRA/apiDefinition.swagger.json");
+            jiraSkills = await kernel.ImportOpenApiSkillFromUrlAsync("jiraSkills", apiSkillRawFileURL, null, tokenProvider.AuthenticateRequestAsync);
+        }
+
+
+        await GetIssueSkillAsync(kernel, jiraSkills, contextVariables);
```
adrianwyatt, line 45: Since this is a syntax example, let's inline these functions so readers can see everything in one sequence.

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,81 @@
...
+    public static async Task RunAsync()
+    {
+        var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
+        var contextVariables = new ContextVariables();
+
+        string s_serverURL = "https://<jiraProject>.atlassian.net/rest/api/latest/";
+        contextVariables.Set("server-url", s_serverURL);
+
+        IDictionary<string, ISKFunction> jiraSkills;
+        var tokenProvider = new BasicAuthenticationProvider(AuthenticateWithBasicApiTokenAsync);
```
adrianwyatt, line 29: Could we inline the callback for readability? We want to make it easy for developers to see the whole flow at once.

adrianwyatt (changes requested):

samples/dotnet/kernel-syntax-examples/Skills/JiraSkill/apiDefinition.swagger.json
amsacha: its the same file. ill remove it

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
amsacha: I feel like its worth showing both approaches?
But the other concern with the full open api definition for jira is that not all the operations work. like createIssue doesnt follow the api schema and therefore doesnt work.

samples/dotnet/kernel-syntax-examples/Example22_a_OpenApiSkill_AzureKeyVault.cs
```diff
@@ -9,7 +9,7 @@
 using RepoUtils;
 
 // ReSharper disable once InconsistentNaming
-public static class Example22_OpenApiSkill
+public static class Example22_a_OpenApiSkill_AzureKeyVault
```
dluc, line 12: the name seems off, we don't use "a" and "b"...

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,75 @@
...
+using RepoUtils;
+
+public static class Example22_b_OpenApiSkill_Jira
+{
+    public static async Task RunAsync()
+    {
+        var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
+        var contextVariables = new ContextVariables();
+
+        string serverUrl = "https://<jiraProject>.atlassian.net/rest/api/latest/";
```
dluc, line 25: if one runs this example (or all the examples) considering that this URL is incomplete, a placeholder, the code will just throw.

shouldn't there be some instructions? We should document the code, explain what is going on, provide setup steps, etc.

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,75 @@
...
+        IDictionary<string, ISKFunction> jiraSkills;
+        var tokenProvider = new BasicAuthenticationProvider(() =>
+        {
+            string s = Env.Var("MY_EMAIL_ADDRESS") + ":" + Env.Var("JIRA_API_KEY");
+            return Task.FromResult(s);
+        });
+
+        // The bool bUseLocalFile can be used to toggle the ingestion method for the openapi schema between a file path and a URL
+        bool bUseLocalFile = true;
+        if (bUseLocalFile)
```
dluc, line 37: why are we using Hungarian notation...?

dluc (changes requested):
pls see comments about naming and docs

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
amsacha: @adrianwyatt, thoughts?

samples/dotnet/kernel-syntax-examples/Example22_a_OpenApiSkill_AzureKeyVault.cs
amsacha: This is to keep all the OpenApi Skills next to one another in the list. I dont think we should rename all the examples after them cause we might add more example variations. 
Could change it to Example_22_OpenApi.... but then it feels like a typo because Example_22_OpenApi.... shows up twice

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
amsacha: This is how the other samples are written up as well, including the other open api plugin example , azure key vault. The placeholder name should be self explanatory enough. I can call it out explicitly in a comment. 
From OpenApiSkill.cs : contextVariables.Set("server-url", "https://<keyvault-name>.vault.azure.net");

samples/dotnet/kernel-syntax-examples/Example22_a_OpenApiSkill_AzureKeyVault.cs
amsacha: @dluc thoughts?

teresaqhoang: We could just keep them all Example22 with the specific postfix 
Example22_OpenApiSkill_AzureKeyVault
Example22_OpenApiSkill_Jira
Example22_OpenApiSkill_GitHub

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,76 @@
...
+        contextVariables.Set("server-url", serverUrl);
+
+        IDictionary<string, ISKFunction> jiraSkills;
+        var tokenProvider = new BasicAuthenticationProvider(() =>
+        {
+            string s = Env.Var("MY_EMAIL_ADDRESS") + ":" + Env.Var("JIRA_API_KEY");
+            return Task.FromResult(s);
+        });
+
+        // The bool bUseLocalFile can be used to toggle the ingestion method for the openapi schema between a file path and a URL
```
teresaqhoang, line 36:
```suggestion
        // The bool useLocalFile can be used to toggle the ingestion method for the openapi schema between a file path and a URL
```

samples/dotnet/kernel-syntax-examples/Example22_b_OpenApiSkill_Jira.cs
adrianwyatt: Ah - okay. If we're modifying the openapi JSON, than checking it in and loading it from disk makes sense. Let's also put a README next to the jira openAPI json explaining why we have a separate copy and why.

samples/dotnet/kernel-syntax-examples/Example22_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Orchestration;
+using Microsoft.SemanticKernel.SkillDefinition;
+using Microsoft.SemanticKernel.Skills.OpenAPI.Authentication;
+using Newtonsoft.Json;
+using RepoUtils;
+
+public static class Example22_OpenApiSkill_Jira
```
dluc, line 13: pls check https://github.com/microsoft/semantic-kernel/blob/main/samples/dotnet/kernel-syntax-examples/Example25_AADAuth.cs and https://github.com/microsoft/semantic-kernel/blob/main/samples/dotnet/kernel-syntax-examples/Example24_Tokenizer.cs. 

Could you add more information at the top explaining what the example is meant to show? Don't worry about being too verbose, people take these examples as learning docs

samples/dotnet/kernel-syntax-examples/Example22_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,71 @@
...
+            var result = await kernel.RunAsync(contextVariables, jiraSkills["GetIssue"]);
+
+            Console.WriteLine("\n\n\n");
+            var formattedContent = JsonConvert.SerializeObject(JsonConvert.DeserializeObject(result.Result), Formatting.Indented);
+            Console.WriteLine("GetIssue jiraSkills response: \n{0}", formattedContent);
+        }
+
+        // AddComment Skill
+        {
+            //Set Properties for the AddComment operation in the openAPI.swagger.json
```
dluc, line 59: nit: add a space between `//` and comments

samples/dotnet/kernel-syntax-examples/Skills/JiraSkill/README.md
```diff
@@ -0,0 +1,5 @@
+# Jira Open API Schema
```
dluc, line 1: nit: add empty line after headers

samples/dotnet/kernel-syntax-examples/Skills/JiraSkill/README.md
```diff
@@ -0,0 +1,5 @@
...
+We have our own curated version of the jira open api schema because the one available online
```
dluc, line 2: I believe "Jira" has a capital "J"

dluc: same for the spelling of "open api"

samples/dotnet/kernel-syntax-examples/Skills/JiraSkill/README.md
```diff
@@ -0,0 +1,5 @@
+# Jira Open API Schema
+We have our own curated version of the jira open api schema because the one available online 
+at https://raw.githubusercontent.com/microsoft/PowerPlatformConnectors/dev/certified-connectors/JIRA/apiDefinition.swagger.json, 
+doesn't follow OpenAPI specification for all of its operations. For example CreateIssueV2, its body param does not describe properties
```
dluc, line 4: did you report it to their repo?

amsacha: No I have not. The PowerPlatform Connector repo uses an older version of the swagger.json, version 2. Atlassian has a version 3 here: https://developer.atlassian.com/cloud/jira/platform/swagger-v3.v3.json. And version 3 has fixed this but its a more dynamic building of the schema for operations. I don't believe we support that in the kernel yet (I think Sergey sergey was talking about how to pass information like this more generically as jsons rather than as specific Context Variables which support limited data types).

samples/dotnet/kernel-syntax-examples/Program.cs
```diff
@@ -71,7 +71,10 @@ public static async Task Main()
         await Example21_ChatGptPlugins.RunAsync();
         Console.WriteLine("== DONE ==");
 
-        await Example22_OpenApiSkill.RunAsync();
+        await Example22a_OpenApiSkill_AzureKeyVault.RunAsync();
```
gitri-ms, line 74: remove 'a'

samples/dotnet/kernel-syntax-examples/Program.cs
```diff
@@ -71,7 +71,10 @@ public static async Task Main()
...
+        Console.WriteLine("== DONE ==");
+
+        await Example22b_OpenApiSkill_Jira.RunAsync();
```
gitri-ms, line 77: remove 'b'

samples/dotnet/kernel-syntax-examples/Example22a_OpenApiSkill_AzureKeyVault.cs
```diff
@@ -9,7 +9,7 @@
 using RepoUtils;
 
 // ReSharper disable once InconsistentNaming
-public static class Example22_OpenApiSkill
+public static class Example22_OpenApiSkill_AzureKeyVault
```
gitri-ms, line 12: File name should be changed to match class name (remove 'a')

samples/dotnet/kernel-syntax-examples/Example22b_OpenApiSkill_Jira.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.SemanticKernel.Skills.OpenAPI.Authentication;
+using Newtonsoft.Json;
+using RepoUtils;
+
+/// <summary>
+/// This sample shows how to connect the Semantic Kernel to Jira as an Open Api plugin based on the Open Api schema.
+/// This format of registering the skill and its operations, and subsequently executing those operations can be applied
+/// to an Open Api plugin that follows the Open Api Schema.
+/// </summary>
+public static class Example22_OpenApiSkill_Jira
```
gitri-ms, line 18: Filename should be changed to match class name (remove 'b')

gitri-ms (changes requested):

adrianwyatt (approved):

cchighman:
Curious whether it makes sense to pull the openapi spec from https://developer.atlassian.com/cloud/jira/platform/swagger-v3.v3.json versus including the spec as part of the skill since this create a tight coupling between the version included with the skill and the current version available -- implying I assume that updates must be made here if we want to keep up to date.  There's also the risk of a breaking change to Jira's API, which is less likely to happen but does happen, that might render this skill unusable for a customer in production until fix were made.

cchighman:
> Curious whether it makes sense to pull the openapi spec from https://developer.atlassian.com/cloud/jira/platform/swagger-v3.v3.json versus including the spec as part of the skill since this create a tight coupling between the version included with the skill and the current version available -- implying I assume that updates must be made here if we want to keep up to date. There's also the risk of a breaking change to Jira's API, which is less likely to happen but does happen, that might render this skill unusable for a customer in production until fix were made.

Perhaps it could be cached, for efficiency.  There is an implicit assumption here that you would always be pulling the latest API as opposed to a specific version so maybe that link above wouldnt work.  Maybe https://developer.atlassian.com/cloud/jira/platform/swagger.json is meant to always point to the latest.

amsacha:
> 

We don't support version 3 in the semantic kernel, so we cant use the latest swagger anyway

dluc (approved):

PR #679, opened 2023-04-26
refactor(AIServiceOptions.cs): remove [NotEmptyOrWhitespace] attribute from Endpoint property validation because it throws error when using OpenAI

Kevdome3000:
### Description

In the appsettings.json the comments say the Endpoint property will be ignored if using OpenAI but it's not. Removing the attribute fixes that

### Contribution Checklist


adrianwyatt:
Thank you for the PR! We just put out another fix for this: https://github.com/microsoft/semantic-kernel/pull/687.

PR #682, opened 2023-04-26
Copilot chat: refactor chat history

TaoChenOSU:
### Motivation and Context

The chat history skill handles the creation of new chat sessions, retrieval of chat sessions and messages, and editing of chat titles. It was unnecessary and improper to use a skill to handle non-AI related task. 

### Description

Replace the chat history skill with a controller that provides APIs to handle the above-mentioned tasks.

### Contribution Checklist


samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,193 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ChatHistoryController"/> class.
+    /// </summary>
+    /// <param name="logger">The logger.</param>
+    /// <param name="chatSessionRepository">The chat session repository.</param>
+    /// <param name="chatMessageRepository">The chat message repository.</param>
+    /// <param name="promptSettings">The prompt settings.</param>
+    public ChatHistoryController(
+        ILogger<SemanticKernelController> logger,
```
adrianwyatt, line 30:
```suggestion
        ILogger<ChatHistoryController> logger,
```

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,193 @@
...
+        this._chatMessageRepository = chatMessageRepository;
+        this._promptSettings = promptSettings;
+    }
+
+    /// <summary>
+    /// Create a new chat session and populate the session with the initial bot message.
+    /// </summary>
+    /// <param name="chatParameters">Object that contains the parameters to create a new chat.</param>
+    /// <returns>The HTTP action result.</returns>
+    [Authorize]
```
adrianwyatt, line 46: Could we put `[Authorize]` on the class and avoid putting it on each API?

samples/apps/copilot-chat-app/webapi/Storage/ChatMessageRepository.cs
```diff
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
-using SemanticKernel.Service.Skills;
+using SemanticKernel.Service.Model;
```
adrianwyatt, line 3:
```suggestion
using SemanticKernel.Service.Storage;
```

samples/apps/copilot-chat-app/webapi/Storage/ChatSessionRepository.cs
```diff
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
-using SemanticKernel.Service.Skills;
+using SemanticKernel.Service.Model;
```
adrianwyatt, line 3:
```suggestion
using SemanticKernel.Service.Storage;
```

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Storage/ChatMessageRepository.cs
TaoChenOSU: Are you suggesting moving the classes to under Storage/?

adrianwyatt: It looks like they are under Storage. ? 
image

samples/apps/copilot-chat-app/webapi/Storage/ChatSessionRepository.cs
adrianwyatt: I see now that this is a using statement and not a namespace - my mistake.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,193 @@
...
+        return this.Ok(chat);
+    }
+
+    /// <summary>
+    /// Get all chat sessions associated with a user. Return an empty list if no chats are found.
+    /// </summary>
+    /// <param name="userId">The user id.</param>
+    [Authorize]
+    [HttpGet]
+    [Route("chat/getAllChats/{userId:regex([[0-9]]+\\.[[0-9]]+)}")]
```
Vivihung, line 94: if we want to make sure the user id is guid, use `{userId:guid}` should be enough.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,193 @@
...
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> GetAllChatsAsync(string userId)
```
Vivihung, line 98: `(Guid userId)` instead of string?

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,193 @@
...
+    [Authorize]
+    [HttpPost]
+    [Route("chat/edit")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> EditChatAsync([FromBody] ChatSession chatParameters)
+    {
+        var chatId = chatParameters.Id;
+        Console.WriteLine(chatId);
```
Vivihung, line 159: Feel like this can be removed. Besides, to my knowledge, we should use ILogger for logs.

samples/apps/copilot-chat-app/webapp/src/libs/models/ChatMessage.ts
```diff
@@ -23,3 +23,5 @@ export interface ChatMessage {
     authorRole: AuthorRoles;
     debug?: string;
 }
+
+export type ChatMessages = IChatMessage[];
```
Vivihung, line 27: I rarely see folks make type for the plural form (aka list/array) of an object. IMO, having the list/array inline (`IChatMessage[]`) is much clear than `ChatMessages`. I would love to learn the motivation behind this plural type.

samples/apps/copilot-chat-app/webapp/src/libs/services/BaseService.ts
```diff
@@ -32,7 +32,11 @@ export class BaseService {
             });
 
             if (!response.ok) {
-                throw Object.assign(new Error(response.statusText + ' => ' + (await response.text())));
+                const responseText = await response.text();
+                let errorMessage = response.status + ': ' + response.statusText;
+                errorMessage += responseText ? ' => ' + responseText : '';
```
Vivihung, line 37: Had hard time to read the string formatting logic here. Could we use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)?

Vivihung (changes requested):
Thanks for the refactory! Left some questions.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
TaoChenOSU: The user id is not a guid. It's actually two guid string separated by a period.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
TaoChenOSU: Same as above

samples/apps/copilot-chat-app/webapp/src/libs/models/ChatMessage.ts
TaoChenOSU: We use lists of IChatMessage[] way more often than just a single IChatMessage. I think having an alias can make things easier. I am open to reversing this change.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
Vivihung: uh. I see. thanks.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
Vivihung: Do we need the console log?

samples/apps/copilot-chat-app/webapp/src/libs/models/ChatMessage.ts
Vivihung: I would appreciate it if we don't create the pattern of creating alias for plural type based on usage. Although type alias and the type doesn't impact developer exp, from readability perspective, it might be confusing.

Vivihung (approved):
Thanks for taking feedbacks.

Vivihung (dismissed):

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,189 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using SemanticKernel.Service.Model;
+using SemanticKernel.Service.Skills;
+using SemanticKernel.Service.Storage;
+
+namespace SemanticKernel.Service.Controllers;
+
+/// <summary>
+/// Controller for chat history.
```
adrianwyatt, line 12: Could you expand this description to include the purpose and context? Something like "Chat history controller. This controller handles storing and recalling chat messages for client display and to be included in short-term context of AI prompts."

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,189 @@
...
+    /// Create a new chat session and populate the session with the initial bot message.
+    /// </summary>
+    /// <param name="chatParameters">Object that contains the parameters to create a new chat.</param>
+    /// <returns>The HTTP action result.</returns>
+    [HttpPost]
+    [Route("chat/create")]
+    [ProducesResponseType(StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> CreateChatAsync(
```
adrianwyatt, line 52: Let's try to be super consistent with our terms. What's the term we want to use here - "Chat" or "ChatSession"? It feels like "chat session" is clearer, so consider "CreateChatSessionAsync". Same comment for all methods in this class.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        return this.Ok(chat);
+    }
+
+    /// <summary>
+    /// Get all chat sessions associated with a user. Return an empty list if no chats are found.
+    /// </summary>
+    /// <param name="userId">The user id.</param>
+    [HttpGet]
+    [Route("chat/getAllChats/{userId:regex(([[a-z0-9]]+-)+[[a-z0-9]]+\\.([[a-z0-9]]+-)+[[a-z0-9]]+)}")]
```
adrianwyatt, line 94: Could you add a natural language description of the regex in the summary or add a "remarks" tag with it?

adrianwyatt: From gpt-3.5.turbo: 
```
This regex matches a string that starts and ends with one or more alphanumeric characters separated by hyphens, and has one or more groups of hyphen-separated alphanumeric characters in-between, followed by a dot, and then another series of alphanumeric characters separated by hyphens. 

For example, this regex would match the following strings: 

- my-string-example.com
- hello-world-123-example.net
- 1a-2b-3c.xyz

And it would not match these strings:

- my_string_example.com (contains underscore instead of hyphen)
- hello-world-.net (ends with a hyphen)
- -starts-with-hyphen.com (starts with a hyphen)
```

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapp/src/libs/models/Bot.ts
```diff
@@ -1,11 +1,11 @@
...
-import { ChatMessage } from './ChatMessage';
+import { IChatMessage } from './ChatMessage';
 
 export interface Bot {
     Schema: { Name: string; Version: number };
     Configurations: { EmbeddingAIService: string; EmbeddingDeploymentOrModelId: string };
     ChatTitle: string;
-    ChatHistory: ChatMessage[];
+    ChatHistory: IChatMessage[];
     Embeddings: any[]; // TODO: type this
```
amsacha, line 10: Is this still a valid todo?

TaoChenOSU: Yes

adrianwyatt (approved):

PR #696, opened 2023-04-27
Introducing trust checks in the kernel

rracanicci:
### Motivation and Context

Prompt injection attacks can fool the LLMs into doing not intended actions. This happens when untrusted input is injected into the prompt, fooling the LLM into doing something that is not intended.

To help identifying and handling such cases, we can start by tracking the trust of: (i) input variables; (ii) prompts; and (iii) function outputs. This way, if a potential sensitive function tries to run with untrusted content of any kind, the desired action can be taken (e.g. prevent the function from running and throwing an exception).

### Description

The main idea behind this PR is to update the context to keep track of the trust state of its variables. So if the kernel identifies that an user defined sensitive function is trying to run with untrusted input, an action can be taken. Currently, the trust is defined as a boolean flag.

The main changes are:

- Update `ContextVariables` to store not only a single string, but an abstraction of a string called `TrustAwareString` that also has a boolean flag about the trust of its content
- Update the kernel to add a service for trust events. This service is currently named `TrustService`, for which there is an interface definition: `ITrustService`. The user can implement this interface and set an instance of it in the kernel and/or for a specific function. 
    - For now the interface includes two events that gets called to validate the input context and rendered prompt (semantic functions). They should return if the context/prompt is currently trusted or not. This can also be used for validation, sanitization and other tasks.
    - There is a default implementation called `DefaultTrustService` that when configured in the kernel, throws a `UntrustedContentException`
- Update semantic and native functions to include the following information:
    - `bool IsSensitive`: when set to true, configures the function to be sensitive (e.g. should not run with untrusted input) [default false]
- Implemented unit tests

The trust checks are also chained across multiple function calls. For example, imagine that we have 4 functions running like below:

```mermaid
flowchart LR
   A["Function A"] --> B["Function B"] == output becomes untrusted ==> C["Function C"]  --> D["Function D\n(SENSITIVE)"] 
```

If the `DefaultTrustService` is used, function D will not execute, this happens because its input comes from Function C, that then comes from Function B, which is untrusted.

### Further ideas (not implemented)

The idea of replacing the string usage within the Kernel with the `TrustAwareString` (that includes trust information) could be extended to other places in the kernel, making it easier to track provenance of the inputs and string operations. This has not been implemented in this PR because this would result in a bigger number of changes.

### What's currently missing

- Currently the planner has not been updated to also track trust. This can be tackled as part of another PR if desired.
- Currently, if any of the variables stored in the context is untrusted, the entire context becomes untrusted, regardless if the variable has been used or not. This has been implemented this way for now to simplify the PR, since updating the template engine to track trust through the `TrustAwareString` would result in a bigger number of changes.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
@@ -60,9 +60,21 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, TrustAwa
     /// <param name="content">The new input value, for the next function in the pipeline, or as a result for the user
     /// if the pipeline reached the end.</param>
     /// <returns>The current instance</returns>
-    public ContextVariables Update(string content)
+    public ContextVariables Update(string? content)
     {
         return this.Update(TrustAwareString.Trusted(content));
+    }
+
+    /// <summary>
+    /// Updates the main input text with the new value after a function is complete.
+    /// This will keep the trust state of the current input set.
+    /// </summary>
+    /// <param name="content">The new input value, for the next function in the pipeline, or as a result for the user
+    /// if the pipeline reached the end.</param>
+    /// <returns>The current instance</returns>
+    public ContextVariables UpdateKeepingTrustState(string? content)
+    {
+        return this.Update(new TrustAwareString(content, this.Input.IsTrusted));
     }
 
     /// <summary>
@@ -206,6 +218,14 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, TrustAwa
             // Note: we don't use an internal setter for better multi-threading
             this._variables[item.Key] = TrustAwareString.Untrusted(item.Value.Value);
         }
+    }
+
+    /// <summary>
+    /// Make the input variable untrusted.
+    /// </summary>
+    public void UntrustInput()
+    {
+        this.Update(TrustAwareString.Untrusted(this.Input.Value));
     }
 
     /// <summary>
--- a/dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
@@ -5,7 +5,9 @@ using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using Microsoft.SemanticKernel.Diagnostics;
+using Microsoft.SemanticKernel.Security;
 
 namespace Microsoft.SemanticKernel.Orchestration;
 
@@ -15,32 +17,52 @@ namespace Microsoft.SemanticKernel.Orchestration;
 /// </summary>
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 [DebuggerTypeProxy(typeof(ContextVariables.TypeProxy))]
-public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
+public sealed class ContextVariables : IEnumerable<KeyValuePair<string, TrustAwareString>>
 {
     /// <summary>
     /// In the simplest scenario, the data is an input string, stored here.
     /// </summary>
-    public string Input => this._variables[MainKey];
+    public TrustAwareString Input => this._variables[MainKey];
 
     /// <summary>
     /// Constructor for context variables.
     /// </summary>
-    /// <param name="content">Optional value for the main variable of the context.</param>
-    public ContextVariables(string? content = null)
+    /// <param name="trustAwareContent">Optional value for the main variable of the context including trust information.</param>
+    public ContextVariables(TrustAwareString? trustAwareContent = null)
     {
-        this._variables[MainKey] = content ?? string.Empty;
+        this._variables[MainKey] = trustAwareContent ?? TrustAwareString.Empty;
+    }
+
+    /// <summary>
+    /// Constructor for context variables.
+    /// By default the content will be trusted.
+    /// </summary>
+    /// <param name="content">Optional value for the main variable of the context.</param>
+    public ContextVariables(string? content) : this(TrustAwareString.Trusted(content)) { }
+
+    /// <summary>
+    /// Updates the main input text with the new value after a function is complete.
+    /// The string includes trust information and will overwrite the trust state of the input.
+    /// </summary>
+    /// <param name="trustAwareContent">The new input value, for the next function in the pipeline, or as a result for the user
+    /// if the pipeline reached the end.</param>
+    /// <returns>The current instance</returns>
+    public ContextVariables Update(TrustAwareString? trustAwareContent)
+    {
+        this._variables[MainKey] = trustAwareContent ?? TrustAwareString.Empty;
+        return this;
     }
 
     /// <summary>
     /// Updates the main input text with the new value after a function is complete.
+    /// By default the content will be trusted.
     /// </summary>
     /// <param name="content">The new input value, for the next function in the pipeline, or as a result for the user
     /// if the pipeline reached the end.</param>
     /// <returns>The current instance</returns>
-    public ContextVariables Update(string? content)
+    public ContextVariables Update(string content)
     {
-        this._variables[MainKey] = content ?? string.Empty;
-        return this;
+        return this.Update(TrustAwareString.Trusted(content));
     }
 
     /// <summary>
@@ -57,7 +79,7 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
             // If requested, discard old data and keep only the new one.
             if (!merge) { this._variables.Clear(); }
 
-            foreach (KeyValuePair<string, string> varData in newData._variables)
+            foreach (KeyValuePair<string, TrustAwareString> varData in newData._variables)
             {
                 this._variables[varData.Key] = varData.Value;
             }
@@ -70,16 +92,16 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     /// This method allows to store additional data in the context variables, e.g. variables needed by functions in the
     /// pipeline. These "variables" are visible also to semantic functions using the "{{varName}}" syntax, allowing
     /// to inject more information into prompt templates.
+    /// The string value includes trust information and will overwrite the trust information already stored for the variable.
     /// </summary>
     /// <param name="name">Variable name</param>
-    /// <param name="value">Value to store. If the value is NULL the variable is deleted.</param>
-    /// TODO: support for more complex data types, and plan for rendering these values into prompt templates.
-    public void Set(string name, string? value)
+    /// <param name="trustAwareValue">Value to store. If the value is NULL the variable is deleted.</param>
+    public void Set(string name, TrustAwareString? trustAwareValue)
     {
         Verify.NotNullOrWhiteSpace(name);
-        if (value != null)
+        if (trustAwareValue != null)
         {
-            this._variables[name] = value;
+            this._variables[name] = trustAwareValue;
         }
         else
         {
@@ -88,18 +110,51 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     }
 
     /// <summary>
+    /// This method allows to store additional data in the context variables, e.g. variables needed by functions in the
+    /// pipeline. These "variables" are visible also to semantic functions using the "{{varName}}" syntax, allowing
+    /// to inject more information into prompt templates.
+    /// By default the variables' value will be trusted.
+    /// </summary>
+    /// <param name="name">Variable name</param>
+    /// <param name="value">Value to store</param>
+    /// TODO: support for more complex data types, and plan for rendering these values into prompt templates.
+    public void Set(string name, string value)
+    {
+        this.Set(name, TrustAwareString.Trusted(value));
+    }
+
+    /// <summary>
+    /// Fetch a variable value and if its content is trusted from the context variables.
+    /// </summary>
+    /// <param name="name">Variable name</param>
+    /// <param name="trustAwareValue">Variable value as a string with trust information</param>
+    /// <returns>Whether the value exists in the context variables</returns>
+    public bool Get(string name, out TrustAwareString trustAwareValue)
+    {
+        if (this._variables.TryGetValue(name, out TrustAwareString result))
+        {
+            trustAwareValue = result;
+            return true;
+        }
+
+        trustAwareValue = TrustAwareString.Empty;
+
+        return false;
+    }
+
+    /// <summary>
     /// Fetch a variable value from the context variables.
     /// </summary>
     /// <param name="name">Variable name</param>
     /// <param name="value">Value</param>
     /// <returns>Whether the value exists in the context variables</returns>
-    /// TODO: provide additional method that returns the value without using 'out'.
     public bool Get(string name, out string value)
     {
-        if (this._variables.TryGetValue(name, out value!)) { return true; }
+        var exists = this.Get(name, out TrustAwareString trustAwareValue);
 
-        value = string.Empty;
-        return false;
+        value = trustAwareValue.Value;
+
+        return exists;
     }
 
     /// <summary>
@@ -109,8 +164,15 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     /// <returns>The value of the variable.</returns>
     public string this[string name]
     {
-        get => this._variables[name];
-        set => this._variables[name] = value;
+        get => this._variables[name].Value;
+        set
+        {
+            // This will be trusted by default for now.
+            // TODO: we could plan to replace string usages in the kernel
+            // with TrustAwareString, so here "value" could directly be a trust aware string
+            // including trust information
+            this._variables[name] = TrustAwareString.Trusted(value);
+        }
     }
 
     /// <summary>
@@ -121,6 +183,29 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     public bool ContainsKey(string key)
     {
         return this._variables.ContainsKey(key);
+    }
+
+    /// <summary>
+    /// True if all the stored variables have trusted content.
+    /// </summary>
+    /// <returns></returns>
+    public bool IsAllTrusted()
+    {
+        return this._variables.Values.All(v => v.IsTrusted);
+    }
+
+    /// <summary>
+    /// Make all the variables stored in the context untrusted.
+    /// </summary>
+    public void UntrustAll()
+    {
+        // Create a copy of the variables map iterator with ToList to avoid
+        // iterating in the map while updating it
+        foreach (var item in this._variables.ToList())
+        {
+            // Note: we don't use an internal setter for better multi-threading
+            this._variables[item.Key] = TrustAwareString.Untrusted(item.Value.Value);
+        }
     }
 
     /// <summary>
@@ -136,7 +221,7 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     /// Get an enumerator that iterates through the context variables.
     /// </summary>
     /// <returns>An enumerator that iterates through the context variables</returns>
-    public IEnumerator<KeyValuePair<string, string>> GetEnumerator()
+    public IEnumerator<KeyValuePair<string, TrustAwareString>> GetEnumerator()
     {
         return this._variables.GetEnumerator();
     }
@@ -153,9 +238,9 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     public ContextVariables Clone()
     {
         var clone = new ContextVariables();
-        foreach (KeyValuePair<string, string> x in this._variables)
+        foreach (KeyValuePair<string, TrustAwareString> x in this._variables)
         {
-            clone[x.Key] = x.Value;
+            clone.Set(x.Key, x.Value);
         }
 
         return clone;
@@ -165,7 +250,7 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
 
     [DebuggerBrowsable(DebuggerBrowsableState.Never)]
     internal string DebuggerDisplay =>
-        this._variables.TryGetValue(MainKey, out string input) && !string.IsNullOrEmpty(input)
+        this._variables.TryGetValue(MainKey, out var input) && !string.IsNullOrEmpty(input.Value)
             ? $"Variables = {this._variables.Count}, Input = {input}"
             : $"Variables = {this._variables.Count}";
 
@@ -174,7 +259,7 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     /// <summary>
     /// Important: names are case insensitive
     /// </summary>
-    private readonly ConcurrentDictionary<string, string> _variables = new(StringComparer.OrdinalIgnoreCase);
+    private readonly ConcurrentDictionary<string, TrustAwareString> _variables = new(StringComparer.OrdinalIgnoreCase);
 
     private sealed class TypeProxy
     {
@@ -183,7 +268,7 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
         public TypeProxy(ContextVariables variables) => this._variables = variables;
 
         [DebuggerBrowsable(DebuggerBrowsableState.RootHidden)]
-        public KeyValuePair<string, string>[] Items => this._variables._variables.ToArray();
+        public KeyValuePair<string, TrustAwareString>[] Items => this._variables._variables.ToArray();
     }
 
     #endregion
--- a/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.SemanticKernel.Memory;
+using Microsoft.SemanticKernel.Security;
 using Microsoft.SemanticKernel.SkillDefinition;
 
 namespace Microsoft.SemanticKernel.Orchestration;
@@ -21,6 +22,11 @@ public sealed class SKContext
     /// </summary>
     /// <returns>Processed input, aka result</returns>
     public string Result => this.Variables.ToString();
+
+    /// <summary>
+    /// Whether all the context variables are trusted or not.
+    /// </summary>
+    public bool IsTrusted => this.Variables.IsAllTrusted();
 
     /// <summary>
     /// Whether an error occurred while executing functions in the pipeline.
@@ -129,6 +135,14 @@ public sealed class SKContext
     }
 
     /// <summary>
+    /// Make all the variables stored in the context untrusted.
+    /// </summary>
+    public void UntrustAll()
+    {
+        this.Variables.UntrustAll();
+    }
+
+    /// <summary>
     /// Print the processed input, aka the current data after any processing occurred.
     /// If an error occurred, prints the last exception message instead.
     /// </summary>
@@ -184,4 +198,33 @@ public sealed class SKContext
             return display;
         }
     }
+
+    #region internal
+
+    /// <summary>
+    /// Update the current result with a new string result and trust information.
+    /// - If the current result is already untrusted, it will be kept untrusted.
+    /// - If the provided stringResult is null, the previous result will be kept and
+    /// only trust information will be updated.
+    /// </summary>
+    /// <param name="stringResult">New result value</param>
+    /// <param name="isResultTrusted">Whether the new result is trusted or not</param>
+    internal void UpdateResult(string? stringResult, bool isResultTrusted)
+    {
+        // Keep previous result if one is not provided
+        string newResult = stringResult ?? this.Result;
+
+        if (isResultTrusted)
+        {
+            // Keeps previous trust state
+            this.Variables.Update(new TrustAwareString(newResult, this.Variables.Input.IsTrusted));
+        }
+        else
+        {
+            // Force the result to be untrusted
+            this.Variables.Update(TrustAwareString.Untrusted(newResult));
+        }
+    }
+
+    #endregion
 }
--- a/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
@@ -6,7 +6,6 @@ using System.Threading;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.SemanticKernel.Memory;
-using Microsoft.SemanticKernel.Security;
 using Microsoft.SemanticKernel.SkillDefinition;
 
 namespace Microsoft.SemanticKernel.Orchestration;
@@ -143,6 +142,14 @@ public sealed class SKContext
     }
 
     /// <summary>
+    /// Make the result untrusted.
+    /// </summary>
+    public void UntrustResult()
+    {
+        this.Variables.UntrustInput();
+    }
+
+    /// <summary>
     /// Print the processed input, aka the current data after any processing occurred.
     /// If an error occurred, prints the last exception message instead.
     /// </summary>
@@ -198,33 +205,4 @@ public sealed class SKContext
             return display;
         }
     }
-
-    #region internal
-
-    /// <summary>
-    /// Update the current result with a new string result and trust information.
-    /// - If the current result is already untrusted, it will be kept untrusted.
-    /// - If the provided stringResult is null, the previous result will be kept and
-    /// only trust information will be updated.
-    /// </summary>
-    /// <param name="stringResult">New result value</param>
-    /// <param name="isResultTrusted">Whether the new result is trusted or not</param>
-    internal void UpdateResult(string? stringResult, bool isResultTrusted)
-    {
-        // Keep previous result if one is not provided
-        string newResult = stringResult ?? this.Result;
-
-        if (isResultTrusted)
-        {
-            // Keeps previous trust state
-            this.Variables.Update(new TrustAwareString(newResult, this.Variables.Input.IsTrusted));
-        }
-        else
-        {
-            // Force the result to be untrusted
-            this.Variables.Update(TrustAwareString.Untrusted(newResult));
-        }
-    }
-
-    #endregion
 }
--- a/dotnet/src/SemanticKernel.Abstractions/Security/ITrustService.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Security/ITrustService.cs
@@ -29,7 +29,7 @@ public interface ITrustService
     /// - This is called for semantic functions before rendering the prompt template.
     /// - This is called for native functions before calling the native function implementation.
     ///
-    /// If the return is false, this means the context will be tagged as untrusted.
+    /// If the return is false, this means the result of the function will be tagged as untrusted.
     ///
     /// The implementation might depend on the application needs. A simple sample implementation
     /// could be accomplished by analyzing the variables in the context, if they are all trusted, then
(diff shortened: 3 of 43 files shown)


shawncal:
@rracanicci Very interesting pattern. Thanks for a careful and thorough PR! I'll take a look.

timlaverty (changes requested):
Two topics on this we should discuss before merging:
1) For design - should the RAI feature set be coupled with SK or factored separately so other orchestrators or components could work with it?  I have thought the RAI work we do would be factored separately and this is tightly integrated with SK implementation.  I could argue it either way.
2) Tactically assuming we go forward with merging - should trust for variables/prompts/function outputs be binary or a sliding scale?

I'd like @dluc to also review this for overall design.

rracanicci:
> Two topics on this we should discuss before merging:
> 
> 1. For design - should the RAI feature set be coupled with SK or factored separately so other orchestrators or components could work with it?  I have thought the RAI work we do would be factored separately and this is tightly integrated with SK implementation.  I could argue it either way.
> 2. Tactically assuming we go forward with merging - should trust for variables/prompts/function outputs be binary or a sliding scale?
> 
> I'd like @dluc to also review this for overall design.



> Two topics on this we should discuss before merging:
> 
> 1. For design - should the RAI feature set be coupled with SK or factored separately so other orchestrators or components could work with it?  I have thought the RAI work we do would be factored separately and this is tightly integrated with SK implementation.  I could argue it either way.
> 2. Tactically assuming we go forward with merging - should trust for variables/prompts/function outputs be binary or a sliding scale?
> 
> I'd like @dluc to also review this for overall design.

Hello @timlaverty, thank you very much for your response, we are glad to have these discussions.

*For Question 1* - What do you mean here Tim? Are you talking about just reorganizing the code, or creating other levels of abstractions to control trust? We are happy to do both for sure. We could even extract the trust flow from the `SKFunction` and inject it in the Kernel as external dependency using DI.

*For Question 2* - This makes a lot of sense, we initially created this as a boolean so we can now have these conversations. I think it could be binary, categorial, or even something else, depending on the use case. We could also abstract this away as an interface to make it more customizable.

I am really happy to do all the suggested changes, I just want to make sure I understand them so I can iterate on the PR.

Thank you very much again :)

StephenHodgson:
+1 for interest. This has been something that has concerned me looking though the pattern and the implementation of this library.

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptTemplateConfig.cs
```diff
@@ -152,6 +152,32 @@ public class InputConfig
...
+    /// </summary>
+    [JsonPropertyName("force_output_to_be_untrusted")]
+    [JsonPropertyOrder(8)]
+    public bool ForceOutputToBeUntrusted { get; set; } = false;
+
+    /// <summary>
+    /// Whether the function is set to be sensitive (default false).
+    /// Sensitive functions should not be allowed to run with untrusted input.
+    /// </summary>
+    [JsonPropertyName("is_sensitive")]
```
dluc, line 177: I don't know if we can apply this flag to a prompt. Prompts can be sent to multiple backends, e.g. OpenAI, Azure OpenAI, local models, etc. and these have different trust characteristics

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
```diff
@@ -36,6 +37,18 @@ public interface ISKFunction
     /// </summary>
     bool IsSemantic { get; }
 
+    /// <summary>
+    /// If true, the output of the function will always considered to be untrusted
+    /// regardless of the input (default false).
+    /// </summary>
+    public bool ForceOutputToBeUntrusted { get; set; }
```
dluc, line 44: similar comment here, I think the trust characteristics depends on the source and not the function, so I'm concerned about the approach and scalability.  For instance if we have a "FetchWebPage" function, it can be used to fetch information from Intranet, a personal forum, over HTTP without SSL, etc. and in each case the trust would be different

dotnet/src/SemanticKernel.Abstractions/Security/SensitiveString.cs
```diff
@@ -0,0 +1,128 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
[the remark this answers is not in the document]
rracanicci: For sure, I have updated the `SensitiveString` class to remove the unused code and I've also cleaned the related tests.

The main idea behind the implementation was to add trust information for each variable stored in the `ContextVariables` class, including the main input. This way, `SKContext` has also been updated to expose a `IsTrusted` boolean that does a lookup in the context's variables trust. It is also allowed to be updated, directly linking to the trust of the main input (INPUT).

```c#
    public bool IsTrusted
    {
        get
        {
            return this.Variables.IsAllTrusted();
        }
        internal set
        {
            this.Variables.SensitiveInput = this.Variables.SensitiveInput.UpdateIsTrusted(value);
        }
    }
```

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptTemplateConfig.cs
rracanicci: Absolutely, the `IsSensitive` flag might not make sense directly linked to the prompt. I've just updated the code and moved this flag to the `SemanticFunctionConfig`. For native functions, it is available as part of the `SKFunctionAttribute` attribute.

Those can be changed in runtime in the current implementation if desired by the user, meaning this flag can be turned on and off by doing `functionInstance.IsSensitive = true|false`.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
rracanicci: 100% agree with you! As a matter of fact, this was just a flag added for debugging during development, I've just removed it now from the code.

Also, I have updated the code to decouple the trust checks using a `ITrustHandler` that is provided to the `SKFunction` at runtime (there is a default example implementation that throws an exception for sensitive functions with untrusted content). The handler provides a method:

```c#
public interface ITrustHandler
{
    bool ValidateInput(ISKFunction func, SKContext context, string? prompt);
}
```

Such method gets called to validate the input context before:

- Executing a completion for a semantic function;
- Calling a native function;

It should return if the current context is considered to be trusted or not. This allows the dynamic behavior you described above (`FetchWebPage` function example). Another way to accomplish this would be to return the context in the native function and directly update its trust before returning, since it should be properly propagated.

Furthermore, the `ValidateInput` method should allow things like sanitization to take place, or any other custom behavior desired.

dotnet/src/SemanticKernel.Abstractions/Security/DefaultTrustHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.SemanticKernel.SkillDefinition;
+
+namespace Microsoft.SemanticKernel.Security;
+
+/// <summary>
+/// Default implementation of the trust handler.
+/// When set, throws an exception to stop execution when sensitive functions try
+/// to run with untrusted content.
+/// </summary>
+public class DefaultTrustHandler : ITrustHandler
```
dluc, line 13: `ITrustService`

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -20,6 +20,21 @@ public sealed class SKContext
...
+    /// </summary>
+    public bool IsTrusted
+    {
+        get
+        {
+            return this.Variables.IsAllTrusted();
+        }
+        internal set
+        {
+            this.Variables.SensitiveInput = this.Variables.SensitiveInput.UpdateIsTrusted(value);
```
dluc, line 34: flag all variables as not trusted (don't allow to change to True, only to False)

dotnet/src/SemanticKernel.Abstractions/Security/SensitiveString.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System.Linq;
+using System.Text.RegularExpressions;
+
```
dluc, line 5:
```suggestion
```

these can be removed

dotnet/src/SemanticKernel.Abstractions/Security/SensitiveString.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /// <summary>
+    /// Creates a copy of the string, keeping the string content but updating the trust flag value.
+    /// </summary>
+    /// <param name="isTrusted">New trust value</param>
+    /// <returns>The new updated sensitive string</returns>
+    public SensitiveString UpdateIsTrusted(bool isTrusted)
+    {
+        return new SensitiveString(this.Value, isTrusted);
```
dluc, line 47: is there a security reason for creating a copy? if so, could you expand the documentation? otherwise you can add a private setter

dotnet/src/SemanticKernel.Abstractions/Security/DefaultTrustService.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// <summary>
+    /// Returns true if the context is already trusted and DefaultTrusted is true.
+    /// Otherwise, returns false.
+    /// </summary>
+    /// <param name="func">Instance of the function being called</param>
+    /// <param name="context">The current execution context</param>
+    /// <returns>Should return true if the context is to be considered trusted, or false otherwise.</returns>
+    /// <exception cref="UntrustedContentException">Raised when the context is unstrusted and the function is sensitive.</exception>
```
dluc, line 37:
```suggestion
    /// <exception cref="UntrustedContentException">Raised when the context is untrusted and the function is sensitive.</exception>
```
typo

dotnet/src/SemanticKernel.Abstractions/Security/DefaultTrustService.cs
```diff
@@ -0,0 +1,67 @@
...
+    public Task<bool> ValidateInputAsync(ISKFunction func, SKContext context)
+    {
+        if (func.IsSensitive && !context.IsTrusted)
+        {
+            throw new UntrustedContentException(
+                UntrustedContentException.ErrorCodes.SensitiveFunctionWithUntrustedContent,
+                 $"Could not run {func.SkillName}.{func.Name}, the function is sensitive and the input untrusted"
+            );
+        }
+        return Task.FromResult(context.IsTrusted && this._defaultTrusted);
```
dluc, line 47:
```suggestion
        if (func.IsSensitive && !context.IsTrusted)
        {
            throw new UntrustedContentException(
                UntrustedContentException.ErrorCodes.SensitiveFunctionWithUntrustedContent,
                $"Could not run {func.SkillName}.{func.Name}, the function is sensitive and the input untrusted"
            );
        }

        return Task.FromResult(context.IsTrusted && this._defaultTrusted);
```

indentation

dotnet/src/SemanticKernel.Abstractions/Security/DefaultTrustService.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <summary>
+    /// Returns the prompt wrapped in a SensitiveString. The SensitiveString will
+    /// be trusted if the context is already trusted and DefaultTrusted is true.
+    /// </summary>
+    /// <param name="func">Instance of the function being called</param>
+    /// <param name="context">The current execution context</param>
+    /// <param name="prompt">The current rendered prompt to be used with the completion client.</param>
+    /// <returns>Should return a SensitiveString representing the final prompt to be used with the completion client.
+    /// The SensitiveString includes trust information.</returns>
+    /// <exception cref="UntrustedContentException">Raised when the context is unstrusted and the function is sensitive.</exception>
```
dluc, line 59:
```suggestion
    /// <exception cref="UntrustedContentException">Raised when the context is untrusted and the function is sensitive.</exception>
```
typo

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -120,6 +167,26 @@ public bool ContainsKey(string key)
...
+    }
+
+    /// <summary>
+    /// Make all the variables stored in the context untrusted.
+    /// </summary>
+    public void MakeAllUntrusted()
+    {
+        foreach (var item in this._variables.ToList())
+        {
+            this._variables[item.Key] = item.Value.UpdateIsTrusted(false);
```
dluc, line 186:
```suggestion
            this._variables[item.Key] = new SensitiveString(item.Value.Value, false);
```

dotnet/src/SemanticKernel.Abstractions/Security/SensitiveString.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
```
dluc, line 48:
```suggestion
```

we can now remove this, only used by ContextVariables

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -70,32 +82,63 @@ public ContextVariables Update(ContextVariables newData, bool merge = true)
...
+
+        if (this._variables.TryGetValue(name, out result!))
+        {
+            value = result.Value;
+            return true;
+        }
 
         value = string.Empty;
+
         return false;
```
dluc, line 142:
```suggestion
        return this.Get(name, out value, out _);
```

dotnet/src/SemanticKernel.Abstractions/Security/ITrustService.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    /// Called to validate the input context before:
+    ///
+    /// - Semantic Functions: rendering the prompt used for the text completion client.
+    /// - Native Functions: calling the native function.
+    /// </summary>
+    /// <param name="func">Instance of the function being called</param>
+    /// <param name="context">The current execution context</param>
+    /// <returns>Should return true if the context is to be considered trusted, or false otherwise.</returns>
+    Task<bool> ValidateInputAsync(ISKFunction func, SKContext context);
```
dluc, line 23: rename to `ValidateContextAsync`

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -523,6 +588,26 @@ private void EnsureContextHasSkills(SKContext context)
         context.Skills ??= this._skillCollection;
     }
 
+    private async Task<bool> ValidateInputAsync(SKContext context)
```
dluc, line 591: rename to ValidateContextAsync

dotnet/src/SemanticKernel.Abstractions/Security/ITrustService.cs
rracanicci: Updated :)

craigomatic (dismissed):
LGTM

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -137,6 +140,19 @@ public KernelBuilder WithConfiguration(KernelConfig config)
         return this;
     }
 
+    /// <summary>
+    /// Use the given default trust service with the kernel to be build.
```
dluc, line 144:
```suggestion
    /// Use the given default trust service with the kernel to be built.
```

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -120,6 +152,28 @@ public bool ContainsKey(string key)
...
+    /// <summary>
+    /// Make all the variables stored in the context untrusted.
+    /// </summary>
+    public void MakeAllUntrusted()
+    {
+        // Create a copy of the variables map iterator with ToList to avoid
+        // iterating in the map while updating it
+        foreach (var item in this._variables.ToList())
+        {
+            this._variables[item.Key] = new SensitiveString(item.Value.Value, false);
```
dluc, line 173:
```suggestion
            // Note: we don't use an internal setter for better multi-threading
            this._variables[item.Key] = new SensitiveString(item.Value.Value, false);
```

dotnet/src/SemanticKernel.Abstractions/Security/UntrustedContentException.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using Microsoft.SemanticKernel.Diagnostics;
+
+namespace Microsoft.SemanticKernel.Security;
+
+#pragma warning disable RCS1194 // Implement standard exception constructors
+
+/// <summary>
+/// Untrusted content exception.
```
dluc, line 11:
```suggestion
/// Untrusted content exception, used to warn about:
- untrusted content in the context passed to any function
- unstrusted prompts when using semantic functions
```

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
```diff
@@ -36,6 +37,17 @@ public interface ISKFunction
     /// </summary>
     bool IsSemantic { get; }
 
+    /// <summary>
+    /// Whether the function is set to be sensitive (default false).
```
dluc, line 41:
```suggestion
    /// Whether the function is set to be sensitive (default false).
    /// When a function is sensitive, the default trust service will throw an exception
    /// if the function is invoked passing in some untrusted input (or context, or prompt).
```

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -77,6 +78,14 @@ public sealed class Plan : ISKFunction
     [JsonIgnore]
     public bool IsSemantic { get; private set; }
 
+    /// <inheritdoc/>
+    [JsonIgnore]
```
dluc, line 82: note for the future: this might be useful to include in the serialization

dotnet/src/SemanticKernel.Abstractions/Security/SensitiveString.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.Security;
+
+/// <summary>
+/// A string wrapper that carries trust information.
+/// All field are readonly.
+/// </summary>
+public class SensitiveString
```
dluc, line 9: rename to `TrustAwareString`

dluc (dismissed):

dotnet/src/SemanticKernel.Abstractions/Security/SensitiveString.cs
rracanicci: Just pushed this changed :)

dluc (dismissed):

dluc (approved):

dluc:
Great work @rracanicci - LGTM

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -17,26 +19,28 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
...
     /// <summary>
     /// Constructor for context variables.
     /// </summary>
     /// <param name="content">Optional value for the main variable of the context.</param>
-    public ContextVariables(string? content = null)
+    /// <param name="isTrusted">Whether the content is trusted or not (default true).</param>
+    public ContextVariables(string? content = null, bool isTrusted = true)
     {
-        this._variables[MainKey] = content ?? string.Empty;
+        this._variables[MainKey] = new TrustAwareString(content ?? string.Empty, isTrusted);
```
shawncal, line 31: (don't block on this):
For a future PR: we can either just allow null to flow through to the ctor, and/or create a TrustAwareString.Empty static instance (which should probably just be isTrusted=true, cause it's empty.)

dluc (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
dluc: yeah, there's a `TrustAwareString.Empty` but we'll have to check how nullability propagates through the code

dotnet/src/SemanticKernel.Abstractions/Security/DefaultTrustService.cs
```diff
@@ -0,0 +1,107 @@
...
+namespace Microsoft.SemanticKernel.Security;
+
+/// <summary>
+/// Default implementation of the trust service.
+///
+/// This is just a simple example implementation that will be used by default if no other is provided.
+///
+/// When set, throws an exception to stop execution when sensitive functions try to run with untrusted content.
+/// </summary>
+public class DefaultTrustService : ITrustService
```
shawncal, line 16: I think you can privatize the constructor here, and have 2 static instances with public static getters

```
public class TrustService 
{
   public static TrustService DefaultTrusted { get; } => new TrustService(true);
   public static TrustService DefaultUntrusted  { get; } => new TrustService(false);
}
```
Also: I prefer having the class name providing the defaults be the same as the interface, but without the "I" prefix (**TrustService.Default**)... makes it easy to guess, reach for when intellisensing. This is seen frequently throughout dotnet (for example, IEnumerable -> Enumerable.Empty<>() )

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -155,4 +168,26 @@ public SKContext Clone()
             LastException = this.LastException
         };
     }
+
+    #region internal
+
+    /// <summary>
+    /// Update the current result with a new string result and trust information.
+    /// - If the current result is already untrusted, it will be kept untrusted.
+    /// - If the provided stringResult is null, the previous result will be kept and
```
shawncal, line 177: @dluc This one's on us... do you think this would be expected behavior? With ContextVars, setting as "null" actually deletes the value. I would think that an explicit call to UpdateResult(null) would be my only/obvious way to clear out the result and prevent it from propagating to the next function.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -155,4 +168,26 @@ public SKContext Clone()
...
+    /// </summary>
+    /// <param name="stringResult">New result value</param>
+    /// <param name="isResultTrusted">Whether the new result is trusted or not</param>
+    internal void UpdateResult(string? stringResult, bool isResultTrusted)
+    {
+        this.Variables.UpdateWithTrustCheck(
+            // Keep previous result if one is not provided
+            stringResult ?? this.Result,
+            // Use previous trust information to update
+            isTrusted: isResultTrusted
```
shawncal, line 188: Edge case: if stringResult and this.Result are both null/empty, and result ends up null, do we still want to consider that "untrusted" for the rest of the pipeline?

dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Create a new trust aware string.
+    /// </summary>
+    /// <param name="value">The raw string value</param>
+    /// <param name="isTrusted">Whether the raw string value is trusted or not</param>
+    public TrustAwareString(string value, bool isTrusted = true)
+    {
+        this.Value = value;
+        this.IsTrusted = isTrusted;
+    }
+
```
shawncal, line 36: I'm usually weary of this... not usually a fan of "magic tricks" in code... but this is a perfect place to use implicit operators. It'll make this much easier, transparent, to consume for those that don't care about trust state. Much like the [StringValues](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringvalues?view=dotnet-plat-ext-7.0) struct, which similarly can be treated just like a string unless you want to tap into its advanced (collection) features.

```suggestion
public static implicit operator string(TrustAwareString s) => s.ToString();
public static implicit operator TrustAwareString(string s) => new TrustAwareString(s);
```

shawncal (commented):
Looks really good! Mostly nits and naming which you can consider or file away for later.

Important ones to consider: struct vs class, and implicit operators.

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -85,29 +90,56 @@ public void Set(string name, string? value)
...
-    /// Fetch a variable value from the context variables.
+    /// Fetch a variable value and if its content is trusted from the context variables.
     /// </summary>
     /// <param name="name">Variable name</param>
     /// <param name="value">Value</param>
+    /// <param name="isTrusted">Whether the variable value is trusted or not</param>
     /// <returns>Whether the value exists in the context variables</returns>
-    /// TODO: provide additional method that returns the value without using 'out'.
-    public bool Get(string name, out string value)
+    public bool Get(string name, out string value, out bool isTrusted)
```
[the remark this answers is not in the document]
shawncal: Also, eh em... TryGet @dluc ;)

shawncal:
@rracanicci @dluc I left a bunch of comments but I don't think any should block this merge. Make a pass through, and @dluc can merge when ready. If there are outstanding things to address, we can file as Issues and fix or close them separately. Great work, great thought leadership -- looking forward to this in main!

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -137,6 +140,19 @@ public KernelBuilder WithConfiguration(KernelConfig config)
...
     }
 
+    /// <summary>
+    /// Use the given default trust service with the kernel to be built.
+    /// Functions directly created through the kernel will use this trust service.
+    /// If null, the created functions will rely on the DefaultTrustService implementation.
+    /// </summary>
+    /// <param name="defaultTrustService">Default trust service to use.</param>
+    /// <returns>Updated kernel builder including the given service.</returns>
+    public KernelBuilder WithDefaultTrustService(ITrustService? defaultTrustService)
```
shawncal, line 150: nit: I'd just call this WithTrustService(..) to differentiate it from THE `DefaultTrustService`.

Similar to the other With* things (WithLogger, WithMemory), it doesn't prevent caller from overriding with a different instance to a specific function call later on; I don't think this one needs the "Default*" name treatment.

dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.Security;
+
+/// <summary>
+/// A string wrapper that carries trust information.
+/// All field are readonly.
+/// </summary>
+public class TrustAwareString
```
shawncal, line 9: Note for follow-up PR:
TrustAwareString: IEquatable<TrustAwareString>, IEquatable<string>

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -120,6 +152,29 @@ public bool ContainsKey(string key)
...
+    /// Make all the variables stored in the context untrusted.
+    /// </summary>
+    public void MakeAllUntrusted()
+    {
+        // Create a copy of the variables map iterator with ToList to avoid
+        // iterating in the map while updating it
+        foreach (var item in this._variables.ToList())
+        {
+            // Note: we don't use an internal setter for better multi-threading
+            this._variables[item.Key] = new TrustAwareString(item.Value.Value, false);
```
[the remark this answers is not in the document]
rracanicci: In this particular case, I am leaning towards keeping it immutable, what do you think? Here is my line of thinking:

The concept of trust might be relative to the context and situation, so the same string might be considered trusted in a particular scenario, but untrusted in another. So making it immutable would ensure one context will not be messing with the other (specially because there is a `.Clone()` on it) if they ever end up pointing to same object reference.

Furthermore, let's imagine we have two references to the same `TrustAwareString` being used by two different contexts and threads (A and B). In this case, if thread A modifies its reference to the `TrustAwareString` instance we might be creating a race condition with thread B, and also, if the update in place works, we might be changing the trust state or the content of a variable being used by another context and not be fully aware of it.

Regarding making it a struct, it is a really nice idea! I love it. I guess in the case, the struct would hold a reference to the encapsulated string (8 bytes in a 64 bits machine) plus a reference to the boolean (most likely 1 byte). For memory alignment optimization the compiler might decide to make it 16 bytes total anyway, but that would be really small and might fit well on the stack or inline somewhere. So the question would be how many variables we are willing to target in the `ContextVariables` map? If this number is not that big, would be nice pointing to it by value and not by refence.

I will update the code to add the missing methods that need to be implemented, such as `GetHashCode`, `Equals` and the ones in `IEquatable`, so if we decide to make it a struct, would be just a matter of updating the class definition :).

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -135,7 +190,11 @@ public override string ToString()
     /// <returns>An enumerator that iterates through the context variables</returns>
     public IEnumerator<KeyValuePair<string, string>> GetEnumerator()
     {
-        return this._variables.GetEnumerator();
+        // For now, return an iterator that does not directly expose the trust aware string
+        foreach (KeyValuePair<string, TrustAwareString> kv in this._variables)
+        {
+            yield return new KeyValuePair<string, string>(kv.Key, kv.Value.Value);
+        }
```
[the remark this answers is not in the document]
rracanicci: For sure, I totally missed the `ToDictionary` method on `Linq`. Let's update the code with the suggested change :)

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -150,22 +209,37 @@ IEnumerator IEnumerable.GetEnumerator()
...
 
+    /// <summary>
+    /// Updates the main input text with the new value after a function is complete.
+    /// Retain the trust boolean, e.g. don't allow to switch from Not Trusted to Trusted.
+    /// </summary>
+    /// <param name="content">The new input value, for the next function in the pipeline, or as a result for the user
+    /// if the pipeline reached the end.</param>
+    /// <param name="isTrusted">Whether the content is trusted or not (default true).</param>
+    /// <returns>The current instance</returns>
+    internal ContextVariables UpdateWithTrustCheck(string? content, bool isTrusted)
```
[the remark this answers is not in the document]
rracanicci: I really like your second suggestion of having a:

- `public ContextVariables Update(string? content)`: updates only the content and keeps the current trust state
-  `public ContextVariables UpdateUntrusted(string? content)`: updates the content and sets `IsTrusted = false`

I will update the code to this if you guys agree :)

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -126,6 +131,14 @@ public ISKFunction Func(string skillName, string functionName)
         this.CancellationToken = cancellationToken;
     }
 
+    /// <summary>
+    /// Make all the variables stored in the context untrusted.
+    /// </summary>
+    public void MakeAllUntrusted()
```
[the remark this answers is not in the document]
rracanicci: No problem at all, suggestions are always welcomed :). I will update this

dotnet/src/SemanticKernel.Abstractions/Security/DefaultTrustService.cs
rracanicci: This is a really nice comment. I am all in for this update, what do you think @dluc?

dotnet/src/SemanticKernel/KernelBuilder.cs
rracanicci: For sure, makes sense :)

But I guess we would have to rename attribute in the `ISKFunction` and `IKernel` to something like `TrustServiceInstance`, to avoid name conflicts with the `TrustService` (default) class name.

dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
rracanicci: Oh yes, specially if we update it to be a struct.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
rracanicci: This function is actually internal and I believe will need to behave this way for the logic to work. This is used in the `SKFunction` flow for both semantic and native functions. Its goal is only to update the context with the function result and make sure that if the previous value was untrusted, keep it as untrusted for safety (the only one that would able to turn untrusted into trusted would be the user in the `TrustService`, but not directly the kernel).

I believe this needs to behave this way (if string is null, do not update the context's input content) because on the native function flows, there are a few delegate types that directly manipulate and return the context, so the kernel would not want to change whatever it is the user provided (except the actual trust flag, what also came from the user with the TrustService). The kernel only does update the string content if the delegate type returns a string.

If you think it is clearer, we could move this logic to something internal in the `SKFunction` directly.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
rracanicci: I would believe so, as I mentioned in some of my other comments above, once the context became untrusted (or a variable), I would say the kernel should make sure it stays this way. The power to turn something untrusted into trusted would only come from the user through the `ITrustService` implementation.

rracanicci:
> @rracanicci @dluc I left a bunch of comments but I don't think any should block this merge. Make a pass through, and @dluc can merge when ready. If there are outstanding things to address, we can file as Issues and fix or close them separately. Great work, great thought leadership -- looking forward to this in main!

Thank you very much for your thoughtful comments @shawncal! I really appreciate the detailed review and comments.

I have addressed most of what you have proposed (the main one missing is the implicit conversion from `TrustAwareString` to `string`). The changes are in a [separate branch on my fork](https://github.com/rracanicci/semantic-kernel/tree/feature/trust-checks-updates). I haven't directly pushed it here in case you guys want to tackle them in a separate PR. 

I will keep working iterating on it :)

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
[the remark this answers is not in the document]
shawncal: One of the additional arguments for struct is that we can mark it as a 
`public readonly struct ...`
...which enforces the immutability!

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -194,19 +242,7 @@ public FunctionView Describe()
...
-            InvokeSemanticAsync(context, settings) :
-            this._function(null, settings, context);
-
-        async Task<SKContext> InvokeSemanticAsync(SKContext context, CompleteRequestSettings? settings)
-        {
-            var resultContext = await this._function(this._aiService?.Value, settings ?? this._aiRequestSettings, context).ConfigureAwait(false);
-            context.Variables.Update(resultContext.Variables);
-            return context;
-        }
+        // ???
```
dluc, line 245: @rracanicci we need to rewrite this from scratch because the underlying implementation has changed /cc @stephentoub

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -217,14 +253,7 @@ async Task<SKContext> InvokeSemanticAsync(SKContext context, CompleteRequestSett
...
     {
-        SKContext context = new(
-            new ContextVariables(input),
-            memory: memory,
-            skills: this._skillCollection,
-            logger: logger,
-            cancellationToken: cancellationToken);
-
-        return this.InvokeAsync(context, settings);
+        // ???
```
dluc, line 256: same here

dluc (approved):

PR #705, opened 2023-04-27
Plan,TextMemorySkill improvements and custom planner examples

lemillermicrosoft:
### Motivation and Context
The motivation of the PR is to demonstrate custom planning capabilities, specifically for the markup returned by `ContextQuery` function. 

In creating this example, some gaps were found in the Plan object and TextMemorySkill. This PR proposes some changes to those objects.

#### Plan
- Plan now updates its state with the result of each step, appending the result to the previous result if the step has more than one named output. This allows the plan to keep track of the context and the intermediate results of each step, and to pass them to the next step as input.

#### Example Custom|Markup Planner
- A new example for using a custom planner to create and execute a plan using the Markup Skill, which can process XML markup created by the ContextQuery skill, which can generate natural language queries from a context. 
- A new example of using a custom planner to generate plans from XML markup, using a simple markup language that supports lookup, echo, and concat functions, as well as nested elements.
- Some minor fixes and improvements to the plan execution logic, the plan parser, the plan invocation logic, and the PlanToString method.

#### TextMemorySkill improvements (for the example).
- TextMemorySkill now has a constructor that allows customizing the default parameters for the Recall function, such as the collection, relevance, and limit. By default, getting a single memory for ContextQuery was not interesting. This change makes it easier to demonstrate more interesting capabilities.

### Contribution Checklist

- [x] Integration Tests pass https://github.com/lemillermicrosoft/semantic-kernel/actions/runs/4831594273

### DIFF
--- a/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
+++ b/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
@@ -66,25 +66,25 @@ internal static class SequentialPlanParser
             var plan = new Plan(goal);
 
             // loop through solution node and add to Steps
-            foreach (XmlNode o in solution)
+            foreach (XmlNode solutionNode in solution)
             {
-                var parentNodeName = o.Name;
+                var parentNodeName = solutionNode.Name;
 
-                foreach (XmlNode o2 in o.ChildNodes)
+                foreach (XmlNode childNode in solutionNode.ChildNodes)
                 {
-                    if (o2.Name == "#text")
+                    if (childNode.Name == "#text")
                     {
-                        if (o2.Value != null)
+                        if (childNode.Value != null)
                         {
-                            plan.AddSteps(new Plan(o2.Value.Trim()));
+                            plan.AddSteps(new Plan(childNode.Value.Trim()));
                         }
 
                         continue;
                     }
 
-                    if (o2.Name.StartsWith(FunctionTag, StringComparison.OrdinalIgnoreCase))
+                    if (childNode.Name.StartsWith(FunctionTag, StringComparison.OrdinalIgnoreCase))
                     {
-                        var skillFunctionName = o2.Name.Split(s_functionTagArray, StringSplitOptions.None)?[1] ?? string.Empty;
+                        var skillFunctionName = childNode.Name.Split(s_functionTagArray, StringSplitOptions.None)?[1] ?? string.Empty;
                         GetSkillFunctionNames(skillFunctionName, out var skillName, out var functionName);
 
                         if (!string.IsNullOrEmpty(functionName) && context.IsFunctionRegistered(skillName, functionName, out var skillFunction))
@@ -101,24 +101,24 @@ internal static class SequentialPlanParser
                                 functionVariables.Set(p.Name, p.DefaultValue);
                             }
 
-                            if (o2.Attributes is not null)
+                            if (childNode.Attributes is not null)
                             {
-                                foreach (XmlAttribute attr in o2.Attributes)
+                                foreach (XmlAttribute attr in childNode.Attributes)
                                 {
                                     context.Log.LogTrace("{0}: processing attribute {1}", parentNodeName, attr.ToString());
                                     if (attr.Name.Equals(SetContextVariableTag, StringComparison.OrdinalIgnoreCase))
                                     {
                                         functionOutputs.Add(attr.InnerText);
-                                        continue;
                                     }
                                     else if (attr.Name.Equals(AppendToResultTag, StringComparison.OrdinalIgnoreCase))
                                     {
                                         functionOutputs.Add(attr.InnerText);
                                         functionResults.Add(attr.InnerText);
-                                        continue;
                                     }
-
-                                    functionVariables.Set(attr.Name, attr.InnerText);
+                                    else
+                                    {
+                                        functionVariables.Set(attr.Name, attr.InnerText);
+                                    }
                                 }
                             }
 
@@ -135,13 +135,13 @@ internal static class SequentialPlanParser
                         else
                         {
                             context.Log.LogTrace("{0}: appending function node {1}", parentNodeName, skillFunctionName);
-                            plan.AddSteps(new Plan(o2.InnerText));
+                            plan.AddSteps(new Plan(childNode.InnerText));
                         }
 
                         continue;
                     }
 
-                    plan.AddSteps(new Plan(o2.InnerText));
+                    plan.AddSteps(new Plan(childNode.InnerText));
                 }
             }
 
--- a/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
+++ b/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
+using System.Collections.Generic;
 using System.Xml;
 using Microsoft.Extensions.Logging;
 using Microsoft.SemanticKernel.Orchestration;
@@ -91,8 +92,8 @@ internal static class SequentialPlanParser
                             var planStep = new Plan(skillFunction);
 
                             var functionVariables = new ContextVariables();
-                            var functionOutputs = new ContextVariables();
-                            var functionResults = new ContextVariables();
+                            var functionOutputs = new List<string>();
+                            var functionResults = new List<string>();
 
                             var view = skillFunction.Describe();
                             foreach (var p in view.Parameters)
@@ -107,12 +108,14 @@ internal static class SequentialPlanParser
                                     context.Log.LogTrace("{0}: processing attribute {1}", parentNodeName, attr.ToString());
                                     if (attr.Name.Equals(SetContextVariableTag, StringComparison.OrdinalIgnoreCase))
                                     {
-                                        functionOutputs.Set(attr.InnerText, string.Empty);
+                                        functionOutputs.Add(attr.InnerText);
                                         continue;
                                     }
                                     else if (attr.Name.Equals(AppendToResultTag, StringComparison.OrdinalIgnoreCase))
                                     {
-                                        functionResults.Set(attr.InnerText, string.Empty);
+                                        functionOutputs.Add(attr.InnerText);
+                                        functionResults.Add(attr.InnerText);
+                                        continue;
                                     }
 
                                     functionVariables.Set(attr.Name, attr.InnerText);
@@ -120,9 +123,13 @@ internal static class SequentialPlanParser
                             }
 
                             // Plan properties
-                            planStep.NamedOutputs = functionOutputs;
-                            planStep.NamedParameters = functionVariables;
-                            planStep.NamedResults = functionResults;
+                            planStep.Outputs = functionOutputs;
+                            planStep.Parameters = functionVariables;
+                            foreach (var result in functionResults)
+                            {
+                                plan.Outputs.Add(result);
+                            }
+
                             plan.AddSteps(planStep);
                         }
                         else
--- a/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
+++ b/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
@@ -92,6 +92,7 @@ internal static class SequentialPlanParser
 
                             var functionVariables = new ContextVariables();
                             var functionOutputs = new ContextVariables();
+                            var functionResults = new ContextVariables();
 
                             var view = skillFunction.Describe();
                             foreach (var p in view.Parameters)
@@ -104,11 +105,14 @@ internal static class SequentialPlanParser
                                 foreach (XmlAttribute attr in o2.Attributes)
                                 {
                                     context.Log.LogTrace("{0}: processing attribute {1}", parentNodeName, attr.ToString());
-                                    if (attr.Name.Equals(SetContextVariableTag, StringComparison.OrdinalIgnoreCase)
-                                        || attr.Name.Equals(AppendToResultTag, StringComparison.OrdinalIgnoreCase))
+                                    if (attr.Name.Equals(SetContextVariableTag, StringComparison.OrdinalIgnoreCase))
                                     {
                                         functionOutputs.Set(attr.InnerText, string.Empty);
                                         continue;
+                                    }
+                                    else if (attr.Name.Equals(AppendToResultTag, StringComparison.OrdinalIgnoreCase))
+                                    {
+                                        functionResults.Set(attr.InnerText, string.Empty);
                                     }
 
                                     functionVariables.Set(attr.Name, attr.InnerText);
@@ -118,6 +122,7 @@ internal static class SequentialPlanParser
                             // Plan properties
                             planStep.NamedOutputs = functionOutputs;
                             planStep.NamedParameters = functionVariables;
+                            planStep.NamedResults = functionResults;
                             plan.AddSteps(planStep);
                         }
                         else
--- a/dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
+++ b/dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
@@ -49,9 +49,9 @@ public class TextMemorySkill
     /// <summary>
     /// Creates a new instance of the TextMemorySkill
     /// </summary>
-    /// <param name="collection">The default collection for Recall</param>
-    /// <param name="relevance">The default relevance value for Recall</param>
-    /// <param name="limit">The default limit for Recall</param>
+    /// <param name="collection">The default collection for Recall. Memories collection to search.</param>
+    /// <param name="relevance">The default relevance value for Recall. The relevance score, from 0.0 to 1.0, where 1.0 means perfect match.</param>
+    /// <param name="limit">The default limit for Recall. The maximum number of relevant memories to recall.</param>
     public TextMemorySkill(string collection = DefaultCollection, string relevance = DefaultRelevance, string limit = DefaultLimit)
     {
         this._collection = collection;
--- a/dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
+++ b/dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
@@ -47,6 +47,19 @@ public class TextMemorySkill
     private const string DefaultLimit = "1";
 
     /// <summary>
+    /// Creates a new instance of the TextMemorySkill
+    /// </summary>
+    /// <param name="collection">The default collection for Recall</param>
+    /// <param name="relevance">The default relevance value for Recall</param>
+    /// <param name="limit">The default limit for Recall </param>
+    public TextMemorySkill(string collection = DefaultCollection, string relevance = DefaultRelevance, string limit = DefaultLimit)
+    {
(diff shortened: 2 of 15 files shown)


dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
```diff
@@ -46,6 +46,19 @@ public class TextMemorySkill
     private const string DefaultRelevance = "0.75";
```
awharrison-28, line 46: I like the constructor change here. I think with this, we should make DefaultRelevance = "0.0", 0.75 is very ADA specific

dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
```diff
@@ -92,21 +105,22 @@ public async Task<string> RetrieveAsync(SKContext context)
...
-        var limit = context.Variables.ContainsKey(LimitParam) ? context[LimitParam] : DefaultLimit;
+        var limit = context.Variables.ContainsKey(LimitParam) ? context[LimitParam] : this._limit;
         if (string.IsNullOrWhiteSpace(limit)) { limit = DefaultLimit; }
 
         context.Log.LogTrace("Searching memories in collection '{0}', relevance '{1}'", collection, relevance);
 
         // TODO: support locales, e.g. "0.7" and "0,7" must both work
-        int limitInt = int.Parse(limit, CultureInfo.InvariantCulture);
+        var limitInt = int.Parse(limit, CultureInfo.InvariantCulture);
+        var relevanceScore = float.Parse(relevance, CultureInfo.InvariantCulture);
```
awharrison-28, line 121: nit: call this relevanceThreshold? Or minimumRelevanc?

samples/dotnet/kernel-syntax-examples/Example29_CustomPlanner.cs
```diff
@@ -0,0 +1,384 @@
...
+
+    public static implicit operator XmlNode(XmlNodeInfo info)
+    {
+        return info.Node;
+    }
+}
+#pragma warning restore CA1815
+
+#pragma warning disable CA1711
+public static class XmlEx
```
dluc, line 293: why are we still using XML and not leveraging the plan data structure?

dluc (commented):
after all the effort put into abstracting away from XML and having a plan data structure, we show an example with 200 lines of code dealing with XML. Does it mean that the plan data structure doesn't provide what we needed? Why can't we leverage all the work done instead of writing another parser?

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -43,12 +43,19 @@ public sealed class Plan : ISKFunction
     public ContextVariables NamedParameters { get; set; } = new();
 
     /// <summary>
-    /// Named outputs for the function
+    /// Named outputs for the function -- these are used within a plan.
     /// </summary>
     [JsonPropertyName("named_outputs")]
     [JsonConverter(typeof(ContextVariablesConverter))]
     public ContextVariables NamedOutputs { get; set; } = new();
```
shawncal, line 50: I see a definite point of confusion between NamedOutputs and NamedResults. I think NamedOutputs could be changed, since it feels to me like "Plan.NamedOutputs" should be the outputs of the plan.

Maybe PlanVariables?

samples/dotnet/kernel-syntax-examples/Example29_CustomPlanner.cs
```diff
@@ -0,0 +1,384 @@
...
+    private static string PlanToString(Plan originalPlan, string indent = " ")
+    {
+        var sb = new StringBuilder($"{indent}Goal: {originalPlan.Description}\n\n{indent}Steps:\n");
+
+        foreach (var step in originalPlan.Steps)
+        {
+            sb.Append(step.Steps.Count switch
+            {
+                0 =>
+                    $"{indent}{indent}- {string.Join(".", step.SkillName, step.Name)} {string.Join(" ", step.NamedParameters.Select(p => $"{p.Key}='{p.Value}'"))}{(step.NamedOutputs.Where(s => s.Key.ToUpper(System.Globalization.CultureInfo.CurrentCulture) != "INPUT").Select(p => $"{p.Key}").FirstOrDefault() is var namedOutputs ? $" => {namedOutputs}" : "")}",
```
shawncal, line 173: Can you please just break this out into a private helper function, or a mutli-line body? Break the magic up into variables and then fortmat the string, or use StringBuilder.

samples/dotnet/kernel-syntax-examples/Example29_CustomPlanner.cs
shawncal: This literally is a "XmlMarkup" Plan example, showing how you might create a Plan (data structure) out of your own XML. I would think XML parsing would be a required component of such an example.

dotnet/src/SemanticKernel/Planning/Plan.cs
lemillermicrosoft: Yeah @RogerBarreto and I had discussions on this last week that we need to continue. I wanted to push this change as a step in a better direction, but will pump the breaks for now.

The problem we want to solve is:
- How to identify that a plan/ISKFunction has an output that needs to be saved to the parent Plan.State
- How to identify that M<=N steps in a Plan have an output that is part of the parent output (Plan.State.ToString()).

lemillermicrosoft: NamedOutputs and NamedResults are both outputs of the plan. Right now, functions have a single output. In the future that could change where a function has multiple outputs that should be preserved in the state, the main issue is knowing which (single/multiple) output is something that should be part of the default result.

lemillermicrosoft:
> after all the effort put into abstracting away from XML and having a plan data structure, we show an example with 200 lines of code dealing with XML. Does it mean that the plan data structure doesn't provide what we needed? Why can't we leverage all the work done instead of writing another parser?

Great idea that we should build a planner or example that uses the serialization syntax for plan creation.

As Shawn noted, the goal of this PR is to create a planner for existing semantic function `ContextQuery`.

samples/dotnet/kernel-syntax-examples/Example12_SequentialPlanner.cs
```diff
@@ -130,12 +131,21 @@ private static async Task BookSamplesAsync()
...
+        foreach (var step in originalPlan.Steps)
+        {
+            sb.Append(step.Steps.Count switch
+            {
+                0 =>
+                    $"{indent}{indent}- {string.Join(".", step.SkillName, step.Name)} {string.Join(" ", step.NamedParameters.Select(p => $"{p.Key}='{p.Value}'"))}{(step.NamedOutputs.Where(s => s.Key.ToUpper(System.Globalization.CultureInfo.CurrentCulture) != "INPUT").Select(p => $"{p.Key}").FirstOrDefault() is var namedOutputs ? $" => {namedOutputs}" : "")}",
+                _ => PlanToString(step, indent + indent)
+            });
+        }
+
```
lemillermicrosoft, line 147: note:

lemillermicrosoft: If I use both ActionPlanner and SequentialPlanner I must do the following. This is because ActionPlanner is a skill as well.

> var planner = new SequentialPlanner(kernel, new SequentialPlannerConfig() { ExcludedSkills = new[] { "this" } });

dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
```diff
@@ -43,9 +43,22 @@ public class TextMemorySkill
     public const string LimitParam = "limit";
 
     private const string DefaultCollection = "generic";
-    private const string DefaultRelevance = "0.75";
+    private const string DefaultRelevance = "0.0";
```
dluc, line 46: this seems a pretty big change, how so?

dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
```diff
@@ -43,9 +43,22 @@ public class TextMemorySkill
...
     private const string DefaultLimit = "1";
 
+    /// <summary>
+    /// Creates a new instance of the TextMemorySkill
+    /// </summary>
+    /// <param name="collection">The default collection for Recall</param>
+    /// <param name="relevance">The default relevance value for Recall</param>
+    /// <param name="limit">The default limit for Recall</param>
+    public TextMemorySkill(string collection = DefaultCollection, string relevance = DefaultRelevance, string limit = DefaultLimit)
```
dluc, line 55: users will need some help with the relevance param, e.g. what's the range? what's a good value? the function should probably validate it too

samples/skills/QASkill/ContextQuery/skprompt.txt
```diff
@@ -17,27 +17,27 @@ essay: longer answers. You can have sub-elements such as fact and fiction
 TODAY is {{time.Date}}
 FIRST NAME: {{$firstname}}
 LAST NAME: {{$lastname}}
-CITYl {{$city}}
+CITY: {{$city}}
 STATE: {{$state}}
 COUNTRY: {{$country}}
-{{recall $input}}
+{{memory.recall $input}}
```
dluc, line 23: in all our demos we've always assumed there's a global "recall" function, which is a nice to have. Why this change?

samples/dotnet/kernel-syntax-examples/Example12_SequentialPlanner.cs
dluc: is there a way not to load the actionplanner skill in the kernel?

lemillermicrosoft: Yes. But if you try to use both, this will happen.

dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
lemillermicrosoft: Guidance was from @awharrison-28 in above comments. The current value is very much dependent on the ADA models, which may not be the default for users.

dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
lemillermicrosoft: Added comments from below to help here. There hadn't been validation elsewhere -- if they give a value greater than 1.0 it will fail to find memories. I'll lean on existing flexibility there for now.

samples/skills/QASkill/ContextQuery/skprompt.txt
lemillermicrosoft: I liked the consistency with `time` above but will revert.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -325,7 +346,7 @@ public FunctionView Describe()
```
shawncal, line 346: AddVariablesToContext(this.State, functionContext);
   - I *think* this can be moved above the loop. Looks like this has to only be called once before the first run, then the UpdateContextWithOutputs takes care of the remainder.

shawncal (approved):

dotnet/src/SemanticKernel/Planning/Plan.cs
shawncal: Not a blocker. Filed a task to check this out.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -252,18 +257,34 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context)
...
+                if (resultAppended)
                 {
                     continue;
                 }
 
-                this.State.Set(item.Key, result.Result.Trim());
+                this.State.Get(DefaultResultKey, out var currentPlanResult);
+                this.State.Set(DefaultResultKey, string.Join("\n", currentPlanResult.Trim(), resultValue));
+                resultAppended = true;
+                continue;
```
shawncal, line 274: `continue;` not needed

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -252,18 +257,34 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context)
...
 
-            // Update state with named outputs (if any)
-            foreach (var item in step.NamedOutputs)
+            // Update Plan Result in State with matching outputs (if any)
+            bool resultAppended = false;
+            foreach (var item in this.Outputs.Intersect(step.Outputs))
             {
-                // ignore the input key
-                if (item.Key.ToUpperInvariant() == "INPUT")
+                if (resultAppended)
```
shawncal, line 266: once `resultAppended = true;` all the rest of the iterations will this this and `continue`, right?
Instead, just call `break;`?

But then it looks like resultAppended is always, unconditionally set to true at the end of the first iteration so maybe not a loop at all?

shawncal (changes requested):
Pulled branch, tried it out. All works. I noticed a couple code things pointed out, worthy of another look.

dotnet/src/SemanticKernel/Planning/Plan.cs
lemillermicrosoft: Yeah sounds good, will want to make sure I test this change well, for follow up.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -133,21 +131,27 @@ public Plan(ISKFunction function)
...
         IReadOnlyList<Plan> steps)
     {
         this.Name = name;
         this.SkillName = skillName;
         this.Description = description;
         this.NextStepIndex = nextStepIndex;
         this.State = state;
-        this.NamedParameters = namedParameters;
-        this.NamedOutputs = namedOutputs;
+        this.Parameters = parameters;
```
RogerBarreto, line 153: `Outputs` is a much better name but I see you kept the `Parameters` instead of using `Inputs`.

My suggestion just to make it easier for the audience is where I see `Outputs` I would expect an `Inputs`.

`InputParameters` and `OutputParameters` would be my ask if you want to keep the `Parameters`

samples/dotnet/kernel-syntax-examples/RepoUtils/PlanExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        string goalHeader = $"{indent}Goal: {originalPlan.Description}\n\n{indent}Steps:\n";
+
+        string stepItems = string.Join("\n", originalPlan.Steps.Select(step =>
+        {
+            if (step.Steps.Count == 0)
+            {
+                string skillName = step.SkillName;
+                string stepName = step.Name;
+
+                string namedParams = string.Join(" ", step.Parameters.Select(param => $"{param.Key}='{param.Value}'"));
```
RogerBarreto, line 21: update the `named` prefixes in those variables.

samples/dotnet/kernel-syntax-examples/Example31_CustomPlanner.cs
```diff
@@ -0,0 +1,364 @@
...
+using Microsoft.SemanticKernel.Orchestration;
+using Microsoft.SemanticKernel.Planning;
+using Microsoft.SemanticKernel.SkillDefinition;
+using Microsoft.SemanticKernel.Skills.Web;
+using Microsoft.SemanticKernel.Skills.Web.Bing;
+using RepoUtils;
+
+// ReSharper disable CommentTypo
+// ReSharper disable once InconsistentNaming
+internal static class Example31_CustomPlanner
```
RogerBarreto, line 20: I think this example is awesome, but became too complex to fully understand.

I would split it into:
1. Creating a Custom Planner from a Custom Markup and Invoking it.
2. (Add new sample) Running Plans within Skills (Plan calling other Plans) 
3. Mixing the two above (Current example) with some reutilization.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -252,18 +257,26 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context)
             #region Update State
 
             // Update state with result
-            this.State.Update(result.Result.Trim());
+            this.State.Update(resultValue);
+
+            // Update Plan Result in State with matching outputs (if any)
+            if (this.Outputs.Intersect(step.Outputs).Any())
```
RogerBarreto, line 263: I would suggest moving away of having a DefaultResultKey, since all the information is already in the plan.

Potentially I would add a `Results` property where you get the collection of results for the plan and you can actually do whatever you want with it, including Join.

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
+using System.Collections.Generic;
 using System.Xml;
 using Microsoft.Extensions.Logging;
 using Microsoft.SemanticKernel.Orchestration;
```
RogerBarreto, line 7: In the `ToPlanFromXml` change the variable names `o`, `o2` to actually represent what they are.

RogerBarreto (changes requested):

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -423,6 +436,32 @@ private static void AddVariablesToContext(ContextVariables vars, SKContext conte
...
+    /// <returns>The updated context.</returns>
+    private SKContext UpdateContextWithOutputs(SKContext context)
+    {
+        var resultString = this.State.Get(DefaultResultKey, out var result) ? result : this.State.ToString();
+        context.Variables.Update(resultString);
+
+        // copy previous step's variables to the next step
+        foreach (var item in this._steps[this.NextStepIndex - 1].Outputs)
+        {
+            if (this.State.Get(item, out var val))
```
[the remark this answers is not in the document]
lemillermicrosoft: I'd like to not touch `ContextVariables` class as part of this change. The pattern is `TryGet` it's just named `Get`.

dotnet/src/SemanticKernel/Planning/Plan.cs
lemillermicrosoft: Let's follow up separately on this one. I kept with `Parameters` since that's the naming on `SKFunction`

dotnet/src/SemanticKernel/Planning/Plan.cs
lemillermicrosoft: I'd really like to sticking with the State object and SKContext -- until those patterns change/evolve (SKRequest/SKResponse maybe?). Since Input/Result is overloaded in SKContext, having this special key is useful right now.

rogerio-barreto (approved):

RogerBarreto (approved):

samples/dotnet/kernel-syntax-examples/RepoUtils/PlanExtensions.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Linq;
+using Microsoft.SemanticKernel.Planning;
+
+namespace RepoUtils;
+
+internal static class PlanExtensions
+{
+    internal static string ToPlanString(this Plan originalPlan, string indent = " ")
```
shawncal, line 10: Feature request (separate PR): please make sure Plan.ToString() calls this.

PR #708, opened 2023-04-28
CopilotChat: Update yarn.lock

adrianwyatt:
Updating to proper format and content.


Vivihung (approved):
thanks for the quick fix

PR #712, opened 2023-04-28
CopilotChat: Refactor planner to support more dynamic skill imports.

adrianwyatt:
### Motivation and Context
In preparation for using skills in the planner that require auth headers.

### Description
- Created CopilotChatPlanner as a light wrapper on SequentialPlanner
- Moved planner skill registration to the controller


amsacha (approved):

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
```diff
@@ -69,8 +69,16 @@ internal static IServiceCollection AddSemanticKernelServices(this IServiceCollec
                 serviceProvider.GetRequiredService<IOptionsSnapshot<AIServiceOptions>>().Get(AIServiceOptions.EmbeddingPropertyName)
                     .ToTextEmbeddingsService(serviceProvider.GetRequiredService<ILogger<AIServiceOptions>>())));
 
-        // Add the planner factory.
-        services.AddPlannerFactory();
+        // Add the planner.
+        services.AddScoped<CopilotChatPlanner>(sp =>
+        {
+            // Create a kernel for the planner with the same contexts as the chat's kernel except with no skills.
+            // This allows the planner to use only the skills that are available at call time.
```
teresaqhoang, line 76: This is not entirely intuitive so adding an additional detail that points to where the skills are registered in SemanticKernelController would be helpful

PR #713, opened 2023-04-28
Adding option to use private PAT token on GitHub Q&A sample

### LINKED ISSUE #444: Private repos on the GitHub Q&A sample
I would want to use the GitHub Q&A sample with private repositories. 

## Details

I have a local prototype of this feature, I would like to put a PR, but in the contributing guidelines talks about "Get agreement from the team and the community that your proposed change is a good one."

### Screenshots of the prototype

Adding a PAT Token option in the REACT app:

image

Adding the PAT Token to the GitHubSkill:

image

Allow the Download skill to modify headers:

image

mahomedalid:
### Motivation and Context

1. Why is this change required?  What problem does it solve? What scenario does it contribute to?

Most integrations of OpenAPI and Semantic Kernel would have the motivation to integrate it with private information or non-public data. GitHub Q&A is an amazing sample, and be able to try it with private repositories it is a very straightforward way to solve scenarios where customers and users want to try it with private data. 

Fixes #444 

### Description

* Added a new Input in the React App, this is a password type of input. I decided to not include this pat token as part of the verification if it is a different repo.
* In the GitHub skill I added the PatToken parameter using the existing context variables pattern. I saw this as straightforward since the PatToken may be used for other GitHub skill features.
* WebFileDownloadSkill is modified to receive any additional headers to the request. I see that WebFileDownloadSkill could have had 3 ways to modify the headers of the request: 1) modify the constructor and class to receive and/or make HttpClient public, 2) Add the headers as part of the context variables received, 3) Override the download methods available to have one that specifically receive custom headers. I decided to use 3 to have the minimal impact in existing code (and tests), although I could see 2 could also be a good approach. Then the GitHub skill detects that an authorization header is necessary, and calls the specific overrided method.

### Pending

* Add Unit and/or integration tests.

### Contribution Checklist


samples/dotnet/github-skills/GitHubSkill.cs
```diff
@@ -37,6 +36,11 @@ public static class Parameters
         /// </summary>
         public const string SearchPattern = "searchPattern";
 
+        /// <summary>
+        /// Personal access token for private repositories.
+        /// </summary>
+        public const string PatToken = "patToken";
```
shawncal, line 42: I don't think this is actionable with the current architecture, since the front-end app doesn't have direct access to the Kernel, but flagging anyway...

In general: I don't think we should be putting passwords into ContextVariables. Instead, the Skill should be constructed with that password, or set as an env/config/app secret.

ContextVariables will get passed in plain text to the AI in various prompts, and be visible to all other functions, etc.

In this application, I also don't think the GitHub logic needs to be in a Skill at all, since the AI is not calling it. 
SK team: should we remove the GitHubSkill and just put this in a helper function? @hathind-ms @SergeyMenshykh @dluc @adrianwyatt

mahomedalid: I agree, although I was not sure what is a sample (to guide an eventual implementation) and what is for demo, there are other security issues out there, the most visible is that the OpenAPI key is passed in plain text from the front-end to the backend (although I saw signs of this probably being migrated to env vars).

FWIW The front-end does not have direct access to the Kernel, but it does through function apps (the kernel http service). This code actually works, I have tested this with private repo.

shawncal: Hey @mahomedalid sorry for the delay. I agree, and I did not intend to block this PR for the above issue. I was just pointing it out to peers, as it's a thing we should look out for: passing private keys as ContextVariables.

Though not strictly the case, SKFunctions are intended to be available for the AI to call via Planners, and ContextVariables are how the data moves between them. At times, the ContextVariables can be sent in plain text to the AI, and they are available to all other functions in the Plan.

The issue you've (accidentally) uncovered above is just good for us to know about so we can help communicate best practices.

This is a great addition to the demo, and we'd like to merge it in. Once you do, we'll have somebody look at reworking how that skill gets its keys. If you don't mind, could you please sync with main, resolve conflicts, and run the `build.cmd` command under the repo root? The integration test may fail if you don't have keys set up, but the unit tests should all pass, and dotnet format will fix the errors showing up in the checks below.

Thanks again!

mahomedalid: @shawncal yeah, don't worry, I understand. Solved conflicts, run build.sh, I saw some integration tests failing as you mentioned, otherwise looks good.

```

Failed!  - Failed:    24, Passed:    27, Skipped:     6, Total:    57, Duration: 1 s - IntegrationTests.dll (net6.0)

Passed!  - Failed:     0, Passed:   632, Skipped:     0, Total:   632, Duration: 1 s - SemanticKernel.UnitTests.dll (net6.0)

Passed!  - Failed:     0, Passed:    90, Skipped:     0, Total:    90, Duration: 4 s - SemanticKernel.Connectors.UnitTests.dll (net6.0)

Passed!  - Failed:     0, Passed:    72, Skipped:     0, Total:    72, Duration: 380 ms - SemanticKernel.Skills.UnitTests.dll (net6.0)
```

evchaki:
@craigomatic  can you take a look at this PR?

dotnet/src/Skills/Skills.Web/WebFileDownloadSkill.cs
```diff
@@ -37,6 +37,19 @@ public WebFileDownloadSkill(ILogger<WebFileDownloadSkill>? logger = null)
         this._httpClient = new HttpClient(this._httpClientHandler);
     }
 
+    public async Task DownloadToFileAsync(string source, IDictionary<string, string> headers, SKContext context)
+    {
+        this._httpClient.DefaultRequestHeaders.Clear();
```
glahaye, line 42: Just wondering...  Are there any headers in there by default before calling this method?  If so, we should save that collection of headers and re-apply it at the end of this method.

1983Thai: TS•Pro

mahomedalid: Sorry, I was two weeks on vacation. That is a good suggestion, I don't think are there any headers since the httpclient is initialized in the constructor, but maybe to prevent future contamination.

mahomedalid: @glahaye on a second thought, because _httpClient is private readonly, the class has total control over the object, including the headers. In the same note, when I was doing this PR I thought a cleaner implementation would be that this class receives the httpClient in the constructor (to avoid dependency injection) or with some kind of DI pattern (including maybe accessors), that way, we don't even need to modify this class, and the caller (ex. GitHubSkill) would need to just add the headers for authorization or others. It would have required to modify substantially the current implementation, so I avoided to do that thinking would be better to have a short and sweet PR instead of a refactor one.

I will close this thread since it would be better to refactor this class, but that was out of the scope of the purpose of this PR, if you feel strongly about it, I can close this.

samples/dotnet/github-skills/GitHubSkill.cs
```diff
@@ -125,9 +130,30 @@ public async Task SummarizeRepositoryAsync(string source, SKContext context)
...
-            context1.Variables.Set(FilePathParamName, filePath);
-            await this._downloadSkill.DownloadToFileAsync($"{repositoryUri}/archive/refs/heads/{repositoryBranch}.zip", context1);
+            var downloadSkillContext = new SKContext(logger: context.Log);
+            downloadSkillContext.Variables.Set(FilePathParamName, filePath);
+
+            var repoBundle = $"{repositoryUri}/archive/refs/heads/{repositoryBranch}.zip";
+
+            this._logger.LogDebug("Downloading {RepoBundle}", repoBundle);
+
+            if (context.Variables.Get(PatTokenParamName, out string patToken))
```
craigomatic, line 140:
```suggestion
            if (context.Variables.TryGetValue(PatTokenParamName, out string? patToken))
```

I realise we have been sitting on this PR for some time and these APIs were likely different when you did the work, lets see if we can get this cleaned up and into the repo asap

samples/dotnet/github-skills/GitHubSkill.cs
craigomatic, line 1: Because we have nuget package references in the samples (this might also have changed between when you first did the work and now), it might be better to update the package reference to point to the Skills.Web project so that the code will build. 

We can later come back in and restore the reference to an updated nuget package with the HTTP headers support once the nuget package has been updated.

samples/apps/github-qna-webapp-react/src/components/GitHubRepoSelection.tsx
```diff
@@ -98,6 +100,20 @@ const GitHubProjectSelection: FC<IData> = ({ uri, keyConfig, prevProject, prevBr
                     placeholder="https://github.com/microsoft/semantic-kernel"
                 />
             </div>
+            <Label>
+                <strong>GitHub PAT Token (optional)</strong>
```
craigomatic, line 104: nit on the wording, perhaps one of: _GitHub PAT (optional)_ or _GitHub Personal Access Token (optional)_. 

Might be worth adding an open in new window link to explain what the PAT is and how to get one, ie: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens

craigomatic (changes requested):
Thank you again for the work here, apologies it's taken an eternity to bring this in. 

I will help get this over the finish line if you are ok with the suggestions :)

samples/dotnet/KernelHttpServer/Extensions.cs
```diff
@@ -160,7 +160,7 @@ internal static void RegisterNativeSkills(this IKernel kernel, IEnumerable<strin
         if (ShouldLoad(nameof(GitHubSkill), skillsToLoad))
         {
             var downloadSkill = new WebFileDownloadSkill();
```
mahomedalid, line 162: @adrianwyatt  this (downloadSkill variable) is not necessary anymore.

samples/dotnet/github-skills/GitHubSkill.cs
```diff
@@ -176,6 +163,25 @@ public async Task SummarizeRepositoryAsync(string source, SKContext context)
         }
     }
 
+    private async Task DownloadToFileAsync(string uri, IDictionary<string, string> headers, string filePath, CancellationToken cancellationToken)
```
mahomedalid, line 166: @adrianwyatt is this a pattern that we want to follow? (Reverting the single-responsibility principle, and duplicating code  instead of reusing it from webskill), or is this only temporary? I am ok, with either decision, just want to make sure it is intentional.

samples/apps/github-qna-webapp-react/src/components/GitHubRepoSelection.tsx
mahomedalid: Adding link. Thanks @adrianwyatt  for the rewording.

samples/dotnet/KernelHttpServer/Extensions.cs
mahomedalid: Removed. cc @craigomatic

samples/dotnet/github-skills/GitHubSkill.cs
mahomedalid: As a sample project we are not worried about duplication of code at this point.

samples/dotnet/github-skills/GitHubSkill.cs
craigomatic: ^No longer valid, resolving

dotnet/src/Skills/Skills.Web/WebFileDownloadSkill.cs
```diff
@@ -46,6 +46,19 @@ public WebFileDownloadSkill(HttpClient httpClient, ILogger<WebFileDownloadSkill>
         this._logger = logger ?? NullLogger<WebFileDownloadSkill>.Instance;
     }
 
+    public async Task DownloadToFileAsync(string source, IDictionary<string, string> headers, SKContext context)
+    {
+        this._httpClient.DefaultRequestHeaders.Clear();
+
+        foreach (var header in headers)
+        {
+            this._httpClient.DefaultRequestHeaders.Add(header.Key, header.Value);
```
SergeyMenshykh, line 55: This class can accept an instance of an HTTP client supplied by a hosting application. The application can use the same instance for other skills that need to work over HTTP and send requests. Changing the headers collection on the same instance may impact the other skills using it. Please add headers to the HttpRequestMessage rather than to the HttpClient instance. For example: https://github.com/microsoft/semantic-kernel/blob/28181d7e556d503cc34b4ada2c83a1a556f7a1aa/dotnet/src/Connectors/Connectors.Memory.Weaviate/WeaviateMemoryStore.cs#LL552C13-L552C13

mahomedalid: I think this was outdated code that is no longer used after some changes that were pushed, will remove. cc  @craigomatic  @adrianwyatt

mahomedalid:
@adrianwyatt  @craigomatic  I saw a change made from the call to an older github api to the latest one. The latest one changes two things: now it redirects to a temporary (5min) private link (302), the redirect request will also need the authorization headers, and it requires extra headers (X-GitHub-Api-Version, User-Agent, Accept), which were not there. Updated, and should be back to working state now.

craigomatic (changes requested):
Validated that this works using a fine-grained token with Contents = read only on a private repo, ie: 
image

If you can make the 1 change discussed offline I'm ok bringing this one in.

craigomatic (approved):

PR #701, opened 2023-04-27
Make Exception<T> base class non-generic and clean up derived types

stephentoub:
### Motivation and Context

Fixes https://github.com/microsoft/semantic-kernel/issues/573

### Description

- Make the base class non-generic and rename it to SKException
- Remove unnecessary private ctors from derived types
- Clean up how messages are generated

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows SK Contribution Guidelines (https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md)
- [x] The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`
- [x] All unit tests pass, and I have added new tests where possible
- [ ] I didn't break anyone :smile:

PR #596, opened 2023-04-22
Rename Skills and IntegrationTests folders

dluc:
### Motivation and Context

Make the files a bit easier to browse in filesystem and VS, having consistency between virtual folders and directories, and consistency with the connectors folder.

### Description

* Rename `dotnet/src/Microsoft.Skills` to `dotnet/src/Skills`
* Rename `dotnet/src/Microsoft.IntegrationTests` to `dotnet/src/IntegrationTests`

PR #717, opened 2023-04-28
🛠️Fix a bug in AIException constructor

lemillermicrosoft:
The AIException constructor that takes an errorCode, a message, and a detail was incorrectly passing null as the detail parameter to the base constructor. This commit fixes this bug by passing the detail parameter correctly. This ensures that the detail information is preserved and available for error handling.

### Motivation and Context
Bug Introduced: #701
Bug Hidden by: #596 

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/AI/AIException.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/AI/AIException.cs
@@ -49,7 +49,7 @@ public class AIException : SKException
     /// <param name="message">A string that describes the error.</param>
     /// <param name="detail">A string that provides additional details about the error.</param>
     public AIException(ErrorCodes errorCode, string? message, string? detail)
-        : this(errorCode, message, detail: null, innerException: null)
+        : this(errorCode, message, detail, innerException: null)
     {
     }
 


glahaye (approved):

dotnet/src/SemanticKernel.Abstractions/AI/AIException.cs
```diff
@@ -49,7 +49,7 @@ public AIException(ErrorCodes errorCode, string? message, Exception? innerExcept
     /// <param name="message">A string that describes the error.</param>
     /// <param name="detail">A string that provides additional details about the error.</param>
     public AIException(ErrorCodes errorCode, string? message, string? detail)
-        : this(errorCode, message, detail: null, innerException: null)
```
stephentoub, line 52: Sorry I missed this.

Seems like we should augment the tests for all of these.

stephentoub: Since I last touched them, I'll add tests after this merges.

lemillermicrosoft: We all missed it, no sweat :) Thanks for offering to add more though, a good idea.

PR #718, opened 2023-04-28
Show how to render prompts

dluc:
### Motivation and Context

New code example showing how to render prompts, e.g. in cases where one wants to leverage semantic functions but execute them with a custom flow, e.g. passing prompts to GPT Turbo 3.5.

The example shows the basics of how one could add a chat UI into an app, selecting some piece of content, and asking AI to extract data. In this case, instead of calling a semantic function, the code calls a Chat model directly, to leverage chat model protections against prompt injection.

### Description

* New syntax example, explaining how to use `PromptTemplateEngine`
* Make one chat service param optional, to simplify the code


samples/dotnet/kernel-syntax-examples/Resources/EmbeddedResource.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System.IO;
+using System.Reflection;
+using RepoUtils;
+
+namespace Resources;
+
+internal static class EmbeddedResource
+{
+    private static readonly string? s_namespace = typeof(EmbeddedResource).Namespace;
```
glahaye, line 11: Nit:  I don't think this can ever be null so you don't need to but the '?' after "string".  If the editors give you a hard time for doing that, you can then put a '!' after "Namespace" to get rid of warnings.

samples/dotnet/kernel-syntax-examples/Resources/EmbeddedResource.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    internal static string Read(string name)
+    {
+        var assembly = typeof(EmbeddedResource).GetTypeInfo().Assembly;
+        if (assembly == null) { throw new ConfigurationException($"[{s_namespace}] {name} assembly not found"); }
```
glahaye, line 16: Is this even a possible situation??

samples/dotnet/kernel-syntax-examples/Resources/EmbeddedResource.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.IO;
+using System.Reflection;
+using RepoUtils;
+
+namespace Resources;
+
+internal static class EmbeddedResource
```
glahaye, line 9: This class does some funky stuff and could totally use a comment or two and how and and why it does.

samples/dotnet/kernel-syntax-examples/Resources/EmbeddedResource.cs
dluc: it's a default implementation I copy&paste where needed, it allows scenarios where it can be null. I rather not customize the code for each scenario

samples/dotnet/kernel-syntax-examples/Resources/EmbeddedResource.cs
dluc: see other comment, also the signature of `.Assembly`. .NET allows that to be null, so I'm making sure it's handled

samples/dotnet/kernel-syntax-examples/Resources/EmbeddedResource.cs
dluc: sure I can add comment, but it's nothing special (or funky) :-) It's just loading resources from the assembly.

lemillermicrosoft (approved):

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/IChatCompletion.cs
```diff
@@ -16,7 +16,7 @@ public interface IChatCompletion
     /// <returns>Generated chat message in string format</returns>
     public Task<string> GenerateMessageAsync(
         ChatHistory chat,
-        ChatRequestSettings requestSettings,
+        ChatRequestSettings? requestSettings = null,
```
shawncal, line 19: Do we expect all chat-style models to have the same (or very similar) properties as OpenAI? i.e. Is this specific settings object required in the IChatCompletion interface, or does that effectively make it an IOpenAIChatCompletion?

dluc: we'll revisit the interface once we have more chat models. We have a similar need with image generation taking different params

Issue #694, opened 2023-04-27
Inconsistent handling in QdrantVectorDbClient results in copilot-sample-app to fail when qdrant is used for Memories


cchighman:
**Describe the bug**
When MemoriesStore is set to qdrant, copilot-sample-app is non-functional due to handling of ISemanticTextMemory.SearchAsync when existing results are not found.  SearchAsync returns an IAsyncEnumerable object but will throw if a result is not found in the underlying source in all places where _response.EnsureSuccessStatusCode()_ is not being implicitly caught.

Currently, for copilot-sample-app, it is impacted due to the Qdrant Skill as seen in this list below where this response method is used.
![image](https://user-images.githubusercontent.com/659214/234844230-9aca825e-3c96-4984-b6ca-80c7d315e84e.png)

There is an inconsistency across _QdrantVectorDbClient_ source for how _response.EnsureSuccessStatusCode()_ is handled when performing a search.  Here is an example where it's being caught and tracing "not found" which I believe is the ideal way to handle this provided a collection is returned.  **GOOD**
![image](https://user-images.githubusercontent.com/659214/234844895-5931b914-d409-4703-869d-7726458f3c3e.png)

_SearchAsync()_ for Qdrant uses _FindNearestInCollectionAsync()_ to return back a collection.  In the example below, there isn't the same try/catch handling and just below, you'll see a check for _Any()_.  The code inside that block is unreachable as if there were not any results returned, _response.EnsureSuccessStatusCode()_ would throw.    **BAD**
![image](https://user-images.githubusercontent.com/659214/234845440-942ee5a4-f5ed-41fc-a9f0-c3bd9fabd5cb.png)

For the interim, I changed the semantics around how _SearchAsync()_ is used in copilot-sample-app similar to below where its catching when _SearchAsync()_ throws having a 404 not found.  
![image](https://user-images.githubusercontent.com/659214/234847648-a5ae1857-f801-4caf-91b6-cf4f555e5322.png)

The current handling in copilot-sample-app assumes _SearchAsync()_ will not throw.  Because it does throw, it bubbles itself all the way up and fails the completion which resulted in this skill to be executed.  I believe this handling below is correct so long as we correctly ensure that _SearchAsync()_  will not throw if results are not found like the other operations in that class.
![image](https://user-images.githubusercontent.com/659214/234854637-17c5abd5-132e-45a0-9010-5849c775b91f.png)

[...]

PR #719, opened 2023-04-28
Resolve completion failures from an unexpected exception thrown in the Qdrant Connector

cchighman:
### Motivation and Context
An unexpected exception is being thrown when attempting to find nearest vectors and no vectors exist.  This results in the current completion activity to fail.  An example of this behavior can be observed in the copilot-sample-app by changing the MemoriesStore type in appsettings from "volatile" to "qdrant".  All subsequent chat completions will fail.

There are seven other API requests in this same class which all use roughly the same conventions and handling of the HTTP request / response.  _FindNearestInCollectionAsync_ is the only one of the eight total which doesn't explicitly check or catch to prevent an exception.
  
For more details and context, see the write-up outlined in the linked issue:  #694 

### Description
Align HTTP response behavior in FindNearestInCollectionAsync to match the consistent behavior uniformly observed for HTTP request/response handling.  This amounts to simply breaking when the response status code is NotFound.

### Contribution Checklist


awharrison-28 (approved):

awharrison-28:
I believe this is already resolved by https://github.com/microsoft/semantic-kernel/pull/698

cchighman:
> I believe this is already resolved by #698
My only concern with patching this in the MemoryStore is the choice to leave this API as the only API that's not doing status code checking / catching in the DbClient.  There's also a pretty heavy workaround there in order to do the checks in the MemoryStore. due to an issue highlighted in the mentioned commit.   If there are changes in the future to DbClient that leverage the NearestVectors API, it would be suseptible to the same issue unless that same methodology introduced in MemoryStore is replicated in other places.  All other APIs in DBClient are expected to return an empty collection or null IAsyncEnumerator in the event of a 404 when calling the API.  This would be the only API that we're expecting to throw instead.  I would suggest to resolve with this change for consistency.

dotnet/src/Connectors/Connectors.Memory.Qdrant/QdrantVectorDbClient.cs
cchighman, line 1: See Lines 72, 122, 177, 216, 259, 383 for examples - in each case, an explicit try/catch is encapsulating response.EnsureSuccessStatusCode() or an explicit check on status code is being performed.  Because we check for status code conditons in the DBClient, exposing these validations to the MemoryStore seems like an inconsistent place to expose this checking.

cchighman:
There's also a nuance introduced with the commit in #698 thats not too obvious.  In the memory store unit tests, it's mocking scenarios that are expecting the NearestVectors API in DBClient to return an empty collection but this scenario can never occur as it stands now.  The change in #698 is modifying the behavior to expect receiving an exception from the DbClient but this case is not covered by the tests.  This is another reason why I believe catching this in the DBClient is the best way to resolve this as it then makes the current unit test in place for this scenario valid.

cchighman:
> There's also a nuance introduced with the commit in #698 thats not too obvious. In the memory store unit tests, it's mocking scenarios that are expecting the NearestVectors API in DBClient to return an empty collection but this scenario can never occur as it stands now. The change in #698 is modifying the behavior to expect receiving an exception from the DbClient but this case is not covered by the tests. This is another reason why I believe catching this in the DBClient is the best way to resolve this as it then makes the current unit test in place for this scenario valid.

The test in question is located in QdrantMemoryStoreTests3.cs with method signature:  _"public async Task ItReturnsEmptyTupleIfNearestMatchNotFoundAsync()"_

In this test, we do the following:
```

       // Arrange
        var mockQdrantClient = new Mock<IQdrantVectorDbClient>();
        mockQdrantClient
            .Setup<IAsyncEnumerable<(QdrantVectorRecord, double)>>(x => x.FindNearestInCollectionAsync(
                It.IsAny<string>(),
                It.IsAny<IEnumerable<float>>(),
                It.IsAny<double>(),
                It.IsAny<int>(),
                It.IsAny<bool>(),
                null,
                It.IsAny<CancellationToken>()))
            .Returns(AsyncEnumerable.Empty<(QdrantVectorRecord, double)>());
```

**Current state:**  FindNearestInCollectionAsync throws an exception and this execution context is unreachable.
**State from #698:**  The above state remains as we are now expecting to throw an exception and handle this in MemoryStore as opposed to DbClient, without any test that confirming these assumptions, and against the current assumptions from the tests.
**State from this PR:**  FindNearestInCollectionsAsync no longer throws when 404 is encountered and handles implicitly in line with other methods.  It now returns the empty collection as expected by the unit tests.

awharrison-28:
Thank you for your detailed response and rationale - I do prefer this solution. 

@adrianwyatt can you review as well.

adrianwyatt:
> Thank you for your detailed response and rationale - I do prefer this solution.
> 
> @adrianwyatt can you review as well.

Same feelings - much appreciated.

adrianwyatt (approved):

PR #741, opened 2023-04-29
Fix typos and code style

dluc:
* Fix filename containing a space (`ApiKeyAuthenticationHandler .cs`)
* Fix typos
* Fix code style
* Add missing rule for local constants
* Add commented code to automate code style checks. These don't work with .NET Standard 2.0, so we'll need some extra research.
* Fix one incorrect ILogger naming
* Add some TODOs about unused code
* Fix xmldocs


dotnet/src/SemanticKernel.UnitTests/AI/AIExceptionTests.cs
```diff
@@ -40,14 +40,14 @@ public void ItRoundtripsArgsToErrorCodeMessageCtor()
     public void ItRoundtripsArgsToErrorCodeMessageExceptionCtor()
     {
         // Arrange
-        const string Message = "this is a test";
+        const string message = "this is a test";
```
stephentoub, line 43: Do you have a style guide somewhere that says what style rules are used in the repo?  I'm used to .NET repos where consts are always PascalCased, regardless of whether they're field-level or local-level.

dluc: I'm currently using the resharper rules in the repo. For constants currently it allows: all caps and camelCase. PascalCase is used in other scenarios. 

We've been talking about automating these checks for people who do not use Rider or Resharper, hopefully that will be in the next sprint. The resharper checks can also be run from the command line. On my side I rely on Rider to identify the issues, e.g. here's a screenshot:

image

I think the ideal would be to automate these rules with editorconfig.

stephentoub: For reference, here's our style guide for dotnet/runtime:
https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

dluc: > I'm used to .NET repos where consts are always PascalCased, regardless of whether they're field-level or local-level.

we can make the switch if that's more common to .NET, I think we only need a good solution to automate across different platforms. Looks like currently VS/editorconfig is not enforcing any convention for constants.

dluc: checking if this would work:

```
dotnet_naming_symbols.constant_vars.applicable_kinds = local
dotnet_naming_symbols.constant_vars.applicable_accessibilities = *
dotnet_naming_symbols.constant_vars.required_modifiers = const

dotnet_naming_rule.constant_vars_should_be_pascal_case.severity = error
dotnet_naming_rule.constant_vars_should_be_pascal_case.symbols = constant_vars
dotnet_naming_rule.constant_vars_should_be_pascal_case.style = pascal_case_style
```

stephentoub: Whatever style you want to use is fine. It'd just be good if it were documented and automated, as you say :)

dluc: from a quick look at ASP.NET looks like local const are camelCase, e.g.

https://github.com/dotnet/aspnetcore/blob/main/src/Servers/Kestrel/tools/CodeGenerator/HttpUtilities/HttpUtilities.cs#L247

https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/Formatters/TextOutputFormatter.cs#L130

stephentoub: Yes, different repos have different styles.  We can choose whatever style we want for this repo, it's just desirable to have that style documented / validated via automation.

.editorconfig
```diff
@@ -161,9 +170,13 @@ dotnet_naming_symbols.any_async_methods.required_modifiers = async
...
 # Rules
 
-dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = error
-dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols = constant_fields
 dotnet_naming_rule.constant_fields_should_be_pascal_case.style = pascal_case_style
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = error
+
+dotnet_naming_rule.local_constant_should_be_uppercase.symbols = local_constant
+dotnet_naming_rule.local_constant_should_be_uppercase.style = uppercase_with_underscore_separator
```
shawncal, line 178: All screaming caps and underscores? Not C# style, and certainly not something we want to enforce in a OSS project, where this will make dotnet developers squirm.

I think should be PascalCase in const class members (public or private). For consts declared in methods like these, it really doesn't matter whether it's pascal or camel case, but if it has to be enforced at all, it should be consistent (PascalCase as @stephentoub says).

shawncal (changes requested):

.editorconfig
dluc: with PascalCase isn't it harder to distinguish constants from Properties?

dluc: changed to PascalCase. I find the convention quite confusing, plus it's one more reason to enforce `this.`

adrianwyatt (approved):

PR #742, opened 2023-04-29
Fix potential multiple enumerations

dluc:
There are more of these in unit tests and samples, where I think it's not a big deal. The PR fixes code in the SDK (one false positive simply resolved using ICollection) and code that could be copied into other projects.


samples/apps/copilot-chat-app/importdocument/Program.cs
```diff
@@ -67,14 +68,15 @@ public static void Main(string[] args)
...
+            IEnumerable<IAccount>? accounts = await app.GetAccountsAsync();
+            IEnumerable<IAccount> accountList = accounts.ToList();
+            if (!accountList.Any())
             {
                 Console.WriteLine("Error: No accounts found");
                 return null;
             }
 
-            return accounts.First().HomeAccountId.Identifier;
+            return accountList.First().HomeAccountId.Identifier;
```
stephentoub, line 79: This actually makes it more expensive, as it's forcing a full enumeration of the accounts in order to turn it into a `List<T>`, allocating a `List<T>` and enough space to hold the results, etc.  In contrast, at worst previously `Any` would allocate an enumerator and just look at the first element, but if the `IEnumerable<T>` returned from `GetAccountsAsync` was actually an `ICollection<T>`, then `Any` will just read its Count property.  Similarly, `First` won't enumerate more than a single item.

Since it looks like the implementation doesn't allow nor account for null IAccounts, you can do this whole thing with just a single FirstOrDefault call:
```C#
IEnumerable<IAccount> accounts = await app.GetAccountsAsync();
IAccount? first = accounts.FirstOrDefault();
if (first is null)
{
    Console.WriteLine("Error: No accounts found");
    return null;
}

return first.HomeAccountId.Identifier;
```

samples/apps/copilot-chat-app/webapi/Storage/ChatMessageRepository.cs
```diff
@@ -28,7 +28,8 @@ public Task<IEnumerable<ChatMessage>> FindByChatIdAsync(string chatId)
 
     public async Task<ChatMessage> FindLastByChatIdAsync(string chatId)
     {
-        var messages = await this.FindByChatIdAsync(chatId);
+        var chatMessages = await this.FindByChatIdAsync(chatId);
```
stephentoub, line 31: Same here.  This can be:
```C#
var chatMessages = await this.FindByChatIdAsync(chatId);
var first = chatMessages.OrderByDescending(e => e.Timestamp).FirstOrDefault();
if (first is null)
{
    throw new KeyNotFoundException($"No messages found for chat {chatId}.");
}
return first;
```

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -136,16 +136,17 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
         }
-        else if (startIdx + count > chatMessages.Count() || count == -1)
+        else if (startIdx + count > messages.Count() || count == -1)
         {
-            count = chatMessages.Count() - startIdx;
+            count = messages.Count() - startIdx;
         }
 
-        chatMessages = chatMessages.OrderByDescending(m => m.Timestamp).Skip(startIdx).Take(count);
+        chatMessages = messages.OrderByDescending(m => m.Timestamp).Skip(startIdx).Take(count);
```
stephentoub, line 149: I'm not sure why all of this work is being done.  Skip and Take are both tolerant of a larger number being passed in than actually exists, so count and startIdx don't need to be validated or sized to be used.  It can just be:
```C#
chatMessages = chatMessages.OrderByDescending(m => m.Timestamp).Skip(startIdx);
if (count >= 0)
{
    chatMessages.Take(count);
}
return this.Ok(chatMessages);
```

mayanpakal (commented):
blue moon water sun

samples/apps/copilot-chat-app/importdocument/Program.cs
dluc: good catch, thanks :-)

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
dluc: `chatMessages = chatMessages.Take(count);` right?

samples/apps/copilot-chat-app/webapi/Storage/ChatMessageRepository.cs
dluc: what about using MaxBy?

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
dluc: fixed

samples/apps/copilot-chat-app/webapi/Storage/ChatMessageRepository.cs
stephentoub: You'd need to be targeting .NET 6+.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
stephentoub: yes

samples/apps/copilot-chat-app/webapi/Storage/ChatMessageRepository.cs
stephentoub: Ah, ok, I had it in my head the whole repo was targeting netstandard2.0.

adrianwyatt (approved):

Issue #527, opened 2023-04-19
Concerns with Connector Classes: Improper Use of HttpClient and the Risk of Port/Socket Exhaustion


alexandrereyes:
I'm analyzing the connector classes and I'm worried about the use of HttpClient. It seems that they are not following the guidelines outlined in https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines. This could potentially cause port/socket exhaustion problems and lead to the application shutting down.

dluc:
hi @alexandrereyes thanks for looking into this, yes we're aware of the issue and we plan on start using IHttpClientFactory shortly. The work fits into a bigger story that covers refactoring the early prototype, introducing ServiceCollection and the usual .NET DI, which will then allow to support this easily.

matthewbolanos:
The HttpClient has been updated across the kernel; please let us know if this issue still persists.

Issue #133, opened 2023-03-21
Allow Custom httpClientHandler


vincilee2:
## Context
I deployed my own LLM models on Azure OpenAI, the endpoint is a self-signed.

## Issues Description
When I sent complete request to my self-signed endpoint, semantic kernel will return an error message:
```
Error: The SSL connection could not be established, see inner exception.
```
I tried to trace the inner exception at debugging mode, got:
```
The remote certificate is invalid because of errors in the certificate chain: RevocationStatusUnKnown"
```

## Related Code
In [OpenAIClientAbstract.cs](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel/AI/OpenAI/Clients/OpenAIClientAbstract.cs):
```
    internal OpenAIClientAbstract(ILogger? log = null, IDelegatingHandlerFactory? handlerFactory = null)
    {
        this.Log = log ?? this.Log;
        this._handlerFactory = handlerFactory ?? new DefaultHttpRetryHandlerFactory();

        this._httpClientHandler = new() { CheckCertificateRevocationList = true };
        this._retryHandler = this._handlerFactory.Create(this.Log);
        this._retryHandler.InnerHandler = this._httpClientHandler;

        this.HTTPClient = new HttpClient(this._retryHandler);
        this.HTTPClient.DefaultRequestHeaders.Add("User-Agent", HTTPUseragent);
    }
```

```this._httpClientHandler = new() { CheckCertificateRevocationList = true };``` The default httpClientHandler is set to always check certificate RevocationList.

## Feature Request
Allow pass customized httpClientHandler either from _handlerFactory or from kernelBuilder.

clintonimaroo:
you can try passing a customized httpClientHandler to the OpenAIClientAbstract class either from _handlerFactory or from kernelBuilder.

vincilee2:
I was wondering if pass _handlerFactory would work since 
this._httpClientHandler = new() { CheckCertificateRevocationList = true };
this._retryHandler = this._handlerFactory.Create(this.Log);
this._retryHandler.InnerHandler = this._httpClientHandler;

dluc:
@lemillermicrosoft suggestions? we're improving the backend to allow this sort of scenarios, but maybe there's a workaround now without changing the code

lemillermicrosoft:
[...]

PR #743, opened 2023-04-30
Allow passing custom HttpClient

carlos-the-ai:
### Motivation and Context

My company network needs more custom set up in HttpClient to make calls, but SK OpenAI code only allows me to pass a hander delegate factory. I cannot use SK without more settings in HttpClient and I want to pass in my own instances anyway so I can manage them from my application.

I think I am not the only one to need this:
 - https://github.com/microsoft/semantic-kernel/issues/527
 - https://github.com/microsoft/semantic-kernel/issues/133

See https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines.

### Description

Changes included:
 - The OpenAI and AzureOpenAI  connectors take optional HttpClient? in the constructor methods instead of IDelegatingHandlerFactory
   - Note first approach was adding two new constructor methods to each type so you could choose. But two constructors could not be there together with optional HttpClient? in one and optional IDelegatingHandlerFactory? in the other because it made calls ambiguous. Then I tried making HttpClient not optional, but that made the parameter order different because in some there are also other optional parameters like openai `string? organization = null`. So the not optional HttpClient would have to come BEFORE optional openai organization in one constructor, but the optional IDelegatingHandlerFactory comes AFTER optional openai organization in another constructor. In the end it seemed better to just take HttpClient and do the custom client creation in the extension methods. I hope this is okay I know it is a bigger change than I planned.
   - IDelegatingHandlerFactory is not being used as a factory in these Open AI connector classes anyway. They use it one to create an HttpClient instance in the constructor method anyway and never call it again so it's just as easy to pass in the HttpClient you want.
 - Updated the extension methods in `KernelConfigOpenAIExtensions` so that they can optionally pass in a custom HttpClient and ILogger. If null then it uses the same defaults as before.
 - There was some mix use of `Logger log` and `ILogger logger` in different places so on the parts I changed I set these all to "logger" because it makes it more clear that it is a object and not a "log" method. I can undo if you don't like this.
 - Changed "credentials" to "credential" to match this is one `TokenCredenial` not a credential collection. I can undo if you don't like this.
 - Made the default HttpClientHandler static and shared for new HttpClients with `disposeHandler: false` https://github.com/dotnet/runtime/issues/16255
 - Updated a sample Program
 - I can add more tests if required

### Contribution Checklist

  - Most tests pass but it skips some because I don't have keys for all these model types so I hope your github build will run this and double check.
  - I changed the constructor method args in the OpenAI code, and the will break your users that are creating these open ai classes using their constructors directly. If the users are using the extension methods (AddAzureOpenAI) then they should not have problems.
  - If you want I can put back all of the constructor methods so they can be created with HttpClient or IDelegatingHandlerFactory but with parameters orders different as explained
  - I don't think those classes need to be aware of IDelegatingHandlerFactory or have it included in the constructor methods, since the extension method or calling code can always convert to HttpClient before calling.


carlos-the-ai:

adrianwyatt:
LGTM - waiting for build break fixes.

lemillermicrosoft (commented):
It might be useful to add a `kernel-syntax-examples` to demonstrate supplying a custom httpclient.

shawncal:
> IDelegatingHandlerFactory is not being used as a factory in these Open AI connector classes anyway. They use it one to create an HttpClient instance in the constructor method anyway and never call it again so it's just as easy to pass in the HttpClient you want.

Yup, you are right. I think we can actually take this further and remove the `IDelegatingHandlerFactory` interface altogether. That will take more work and planning. This change is a huge step in the right direction, and we should close on it before doing more.

> It might be useful to add a `kernel-syntax-examples` to demonstrate supplying a custom httpclient.

@lemillermicrosoft I agree with the above. I'll take a note to do that when we finish the change -- if we are able to remove IDelegatingHandlerFactory as noted above, we can likely update "Example08_RetryHandler". There is an example in "KernelBuilder".

PR #747, opened 2023-05-01
Azure Cognitive Search connector for Semantic Memory

dluc:
### Motivation and Context

Azure Cognitive Search is a managed service on Azure that allows to store data and search it using natural language.

### Description

Add a new Semantic Memory Connector leveraging Azure Cognitive Search. Update the syntax example to show how it seamlessly provides `ISemanticTextMemory` API without the need to use embeddings.


dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        // Create the index if it doesn't exist
+        if (existingIndex == null || existingIndex.Value == null)
+        {
+            await this.CreateIndexAsync(indexName, cancellationToken).ConfigureAwait(false);
+        }
+
+        // Search an available client from the local cache
+        if (!this._clientsByIndex.TryGetValue(indexName, out SearchClient? client) || client == null)
```
stephentoub, line 205: In what situation will `client == null`?

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -0,0 +1,316 @@
...
+        {
+            client = this._adminClient.GetSearchClient(indexName);
+            this._clientsByIndex[indexName] = client;
+        }
+
+        return client;
```
stephentoub, line 211: Two concurrent calls to GetSearchClientAsync could end up getting a different SearchClient instance.  Is that ok?  With small tweaks here you could ensure the same instance is returned regardless of such race conditions.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -0,0 +1,316 @@
...
+        if (indexName.Length > 128)
+        {
+            throw new AzureCognitiveSearchMemoryException("The collection name is too long, it cannot exceed 128 chars");
+        }
+
+#pragma warning disable CA1308 // The service expects a lowercase string
+        indexName = indexName.ToLowerInvariant();
+#pragma warning restore CA1308
+
+        return Regex.Replace(indexName.Trim(), "[\\s|\\\\|/|.|_|:]", "-");
```
stephentoub, line 296: The Regex static methods are easy to use but have hidden costs.  It'd be better to move it to a field and then use the instance Replace method, ala https://github.com/microsoft/semantic-kernel/pull/728.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryException.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <param name="message">Exception message.</param>
+    /// <param name="innerException">Inner exception.</param>
+    public AzureCognitiveSearchMemoryException(string? message, Exception? innerException) : base(message, innerException)
+    {
+    }
+
+    private AzureCognitiveSearchMemoryException()
+    {
+        // Do not use, error message is required
+    }
```
stephentoub, line 32: This should either be public or deleted.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
dluc: it would be fine, but I'll double check for improvements

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryException.cs
dluc: IIRC the build throws warnings if it's not defined. We make it private to stop people from raising exceptions without providing useful information.

stephentoub: That's because there's an analyzer enabled that's checking that condition; its actual intent is to ensure that such ctors exist and are public. If weet don't want to expose such ctors, the analyzer should be disabled, or if we don't want to disable it everywhere, it should instead be suppressed here. Adding such a private ctor is an expebsive way to suppress the analyzer :-)

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -0,0 +1,316 @@
...
+
+    /// <inheritdoc />
+    public Task<string> SaveInformationAsync(
+        string collection,
+        string text,
+        string id,
+        string? description = null,
+        string? additionalMetadata = null,
+        CancellationToken cancel = default)
+    {
```
adrianwyatt, line 57: Do we want to do any parameter validation?

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -0,0 +1,316 @@
...
+        CancellationToken cancellationToken = default)
+    {
+        var client = await this.GetSearchClientAsync(indexName, cancellationToken).ConfigureAwait(false);
+
+        Response<IndexDocumentsResult>? result = await client.MergeOrUploadDocumentsAsync(
+            new List<AzureCognitiveSearchRecord> { record },
+            new IndexDocumentsOptions { ThrowOnAnyError = true },
+            cancellationToken).ConfigureAwait(false);
+
+        if (result == null)
```
adrianwyatt, line 259: Should we check `result.Value.Results.Count > 0` as well?

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -0,0 +1,316 @@
...
+        if (indexName.Length > 128)
+        {
+            throw new AzureCognitiveSearchMemoryException("The collection name is too long, it cannot exceed 128 chars");
+        }
+
+#pragma warning disable CA1308 // The service expects a lowercase string
+        indexName = indexName.ToLowerInvariant();
+#pragma warning restore CA1308
+
+        return Regex.Replace(indexName.Trim(), "[\\s|\\\\|/|.|_|:]", "-");
```
adrianwyatt, line 296: Could you add a comment describing the intent of the regex? For example: replace any whitespace, backslash, forward slash, period, underscore or colon with a dash.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchRecord.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Azure.Search.Documents.Indexes;
+using Azure.Search.Documents.Indexes.Models;
+
+namespace Microsoft.SemanticKernel.Connectors.Memory.AzureCognitiveSearch;
+
+public class AzureCognitiveSearchRecord
```
adrianwyatt, line 8: XMLDoc throughout would be lovely.

adrianwyatt (changes requested):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /// <inheritdoc />
+    public Task<string> SaveInformationAsync(
+        string collection,
+        string text,
+        string id,
+        string? description = null,
+        string? additionalMetadata = null,
+        CancellationToken cancel = default)
```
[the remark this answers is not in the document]
dluc: yeah that's after the other PR merge. I had these fixed and then reverted and now I'm fixing again lol

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
dluc: it's very unlikely, but I added the check

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchRecord.cs
dluc: I added to all the public methods and to the non-obvious ones. Let me know if I missed some :-)

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
dluc: I can integrate the optimization later, I wouldn't block the PR on this. Could you share the code?

hathind-ms (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
adrianwyatt: I see - we're just passing it down the line.

adrianwyatt (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryException.cs
dluc: solved

PR #749, opened 2023-05-01
Remove ValidationException

stephentoub:
### Motivation and Context

Remove custom ValidationException in favor of using built-in .NET exception types typically used for the same purposes.

### Description

The conditions ValidationException was used for validating are better suited to existing .NET exception types that .NET developers expect in those situations.

- NotNullOrWhiteSpace is represented with ArgumentException.
- Directory not found is represented with DirectoryNotFoundException
- Invalid state of an object (e.g. a field not being initialized) is represented with InvalidOperationException

Argument validation exceptions are also usage exceptions typically not intended to be caught.  The same would go for the ValidationExceptions that were serving the same purpose, so having a separate type identity for those exceptions isn't desirable.

### Contribution Checklist

cc: @shawncal, @dluc


adrianwyatt (dismissed):

dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs
```diff
@@ -53,28 +55,29 @@ internal static void ValidFunctionName([NotNull] string? functionName)
...
+            throw new ArgumentException(textParamName, message);
         }
     }
 
     internal static void DirectoryExists(string path)
     {
         if (!Directory.Exists(path))
         {
-            ThrowValidationException(ValidationException.ErrorCodes.DirectoryNotFound, $"Directory not found: {path}");
+            throw new DirectoryNotFoundException($"Directory '{path}' could not be found.");
```
shawncal, line 80: nit: I think we could just pass "path" to DirectoryNotFoundException and do the format in the exception class

stephentoub: DirectoryNotFoundException is a built-in .NET exception.  Its ctor just takes a message; it doesn't have one that takes a path and wraps that in a message.

dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs
```diff
@@ -24,18 +26,18 @@ internal static void NotNull([NotNull] object? obj, [CallerArgumentExpression(na
     }
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    internal static void NotEmpty([NotNull] string? str, string? message = null, [CallerArgumentExpression(nameof(str))] string? paramName = null)
+    internal static void NotNullOrWhiteSpace([NotNull] string? str, [CallerArgumentExpression(nameof(str))] string? paramName = null)
```
shawncal, line 29: OMG thank you! This has always been annoying.

dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs
shawncal: ah, gotcha!

shawncal (approved):

PR #754, opened 2023-05-01
Add property model to CompleteRequestSettings.cs

DavidLuong98:
Summary:

This pull request adds a new "Model" property to the CompleteRequestSettings class, which allows clients to specify the name of the GPT model to use for text generation.

Details:

The CompleteRequestSettings class previously included properties for controlling various aspects of the text generation process, such as temperature, top-p, and penalty values. However, it did not include a property for specifying the GPT model to use for generating text. I'd like to specify whether i'm using `text-davinci-003` or `gpt-3.5-turbo`

This pull request addresses this issue by adding a new Model property to the CompleteRequestSettings class. By default, this property is set to`text-davinci-003`, but clients can set it to the name of any available GPT model.

Additionally, the FromCompletionConfig method has been updated to set the Model property based on the config.Model value.

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/CompleteRequestSettings.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/CompleteRequestSettings.cs
@@ -48,6 +48,11 @@ public class CompleteRequestSettings
     public IList<string> StopSequences { get; set; } = Array.Empty<string>();
 
     /// <summary>
+    /// The name of the model to use for the completion.
+    /// </summary>
+    public string Model { get; set; } = "text-davinci-003";
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


dluc:
hi @DavidLuong98 - in SK we consider each model a "service", e.g. if you want to use text-davinci-002, text-davinci-003 and gpt turbo 3.5 you'll write something like this:

```csharp
kernel.Config.AddOpenAITextCompletionService("text-davinci-002", Env.Var("OPENAI_API_KEY"));

kernel.Config.AddOpenAITextCompletionService("text-davinci-003", Env.Var("OPENAI_API_KEY"));

kernel.Config.AddOpenAITextCompletionService("gpt-turbo-35", Env.Var("OPENAI_API_KEY"));
```

PR #755, opened 2023-05-01
Clean up CA1032 code analyzer rule/usage

dluc:
Minor clean up of the analyzer rule CA1032 for exceptions. Aligning on the same approach.

Almost all (if not all) exceptions in the SDK don't have a parameterless constructor, to force passing useful error messages. The PR align all the exceptions to follow the same approach.

We might need to revisit this rule in future if serialization and exception routing becomes a requirement, allowing parameterless ctors only for serialization, and blocking PRs that introduce throwing exceptions without useful info.


lemillermicrosoft:
Can we just disable the rule in our editorconfig and remove all the #pragma disable directives?

dluc:
@lemillermicrosoft yeah I was thinking the same but I felt like I shouldn't touch the global rules. @shawncal thoughts?

stephentoub:
> We might need to revisit this rule in future if serialization and exception routing becomes a requirement, allowing parameterless ctors only for serialization

Binary formatter is being deprecated:
https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md

PR #765, opened 2023-05-02
CopilotChat: Migrate to ActionPlanner

adrianwyatt:
### Motivation and Context
The SequentialPlanner needs GPT-4 to operate with relative consistency and correctness. Using ActionPlanner simplifies the planning feature set but also allows us to use gpt-3.5-turbo with single step plans.

### Description
- Updated CopilotChat planner feature flag to enabled=true
- Updated CopilotChatPlanner to use ActionPlanner (single step planner)
- Added an AIService configuration for CopilotChat's planner to disaggregate the completion model with the planner model.
- CopilotChat won't invoke plans with no steps.
- Removed core and semantic skills from CopilotChat planner manuals
- ActionPlanner: Added support for non-string parameters
- ActionPlanner: Added JSON property sanitation to 'rationale' values (replace double quotes with single and remove newlines)
- ActionPlanner: fixed JSON bug when no relevant function is found (extra close curly brace)


dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanResponse.cs
```diff
@@ -24,7 +24,7 @@ public class PlanData
         /// <summary>
         /// Parameter values
         /// </summary>
-        public Dictionary<string, string> Parameters { get; set; } = new();
+        public Dictionary<string, object> Parameters { get; set; } = new();
```
dluc, line 27: is the change to support numbers/bools?

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -98,25 +98,25 @@ public async Task<Plan> CreatePlanAsync(string goal)
...
         {
             var parts = planData.Plan.Function.Split('.');
-            function = this._context.Skills!.GetFunction(parts[0], parts[1]);
+            plan.AddSteps(this._context.Skills!.GetFunction(parts[0], parts[1]));
         }
-        else
+        else if (!string.IsNullOrWhiteSpace(planData.Plan.Function))
         {
-            function = this._context.Skills!.GetFunction(planData.Plan.Function);
+            plan.AddSteps(this._context.Skills!.GetFunction(planData.Plan.Function));
```
dluc, line 109: actually I think we should store the function in the root. There was a bug in the plan object forcing us to use AddStep. Now we can use:

```
var plan = new Plan(function)
```

(no need to store the goal)

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -230,7 +231,14 @@ private void PopulateList(StringBuilder list, IDictionary<string, List<FunctionV
...
             {
                 // Function description
-                list.AppendLine($"// {AddPeriod(func.Description)}");
+                if (func.Description != null)
+                {
+                    list.AppendLine($"// {AddPeriod(func.Description)}");
+                }
+                else
+                {
+                    list.AppendLine($"// This is a description of {func.SkillName}.{func.Name}.");
```
dluc, line 240: you know, we could consider throwing an error here. Using the planner with functions without a description is bound to go badly, and I think it would help leading developers in a direction where functions must be described to be planner-compatible.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -249,4 +257,16 @@ private static string AddPeriod(string x)
     {
         return x.EndsWith(".", StringComparison.Ordinal) ? x : $"{x}.";
     }
+
+    /// <summary>
+    /// Replace all double quotes and remove all new lines in resultString up to the function property.
```
dluc, line 262: could you add some information about why this is needed (examples might help)?

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
dluc: if you like, we need a Plan ctor that allows to pass a goal and a single function, storing the function in the root, ie something like

```csharp
public Plan(ISKFunction function, string goal, string? functionName = null, string? skillName = null)
{
    this.SetFunction(function);
    this.Description = goal;
    this.Name = functionName ?? goal;
    this.SkillName = skillName ?? this.GetType().FullName;
}
```

/cc @lemillermicrosoft

dluc (commented):
Can we sync about the changes to the planner before merging?

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
cchighman, line 1: I get this comment isn't necessarily in scope for the intent of this PR so just mentioning this overall in terms of design. 
 There's quite a bit of manual string operations happening here.  I'm wondering if there might be value making a serializer class that handled these operations implicitly and deserialized into POCOs.  This would provide the ability to organize these various operations in an isolated space.   Ideally, you would be able to lock in all string semantics into operations that could occur in say a JSON object so that your complexity is reduced down to that standard.  Arbitrary string parsing opens the door for quite a bit of edge cases and without any tests, seems like it might be a bit brittle.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
lemillermicrosoft: We have `public Plan(ISKFunction function)` and `public Plan(string goal, params ISKFunction[] steps)`. Maybe we can use one of those. I'd like to think more about a single function being the root of a plan and the result of overwriting that functions description with a goal.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanResponse.cs
adrianwyatt: Correct.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
adrianwyatt: This is an unfortunate reality of OpenAPI definitions.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
adrianwyatt: I brought in a change to have the AI model generate the total JSON rather than a fragment, which pushes the onus of JSON formatting to the AI.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
dluc: we receive a lot of feedback about the planner not doing the right thing, we should at least log a warning. Also we can save some tokens here making the comment shorter, e.g.
```
// Function {func.SkillName}.{func.Name}
```

dotnet/src/Extensions/Planning.ActionPlanner/skprompt.txt
```diff
@@ -8,6 +8,4 @@ For each function the list includes details about the input parameters.
 - List of functions:
 {{this.ListOfFunctions}}
 - End list of functions.
-Goal: {{ $input }}
-{"plan":{
```
dluc, line 12: I would undo this change:
* the JSON was semi open on purpose to reduce the chance of models introducing things that are not json
* less tokens to generate (speed concern)

dluc (dismissed):
Some comments to address before merging, in particular the change to the prompt

dotnet/src/Extensions/Planning.ActionPlanner/skprompt.txt
adrianwyatt: I was getting more instances of invalid JSON without it - let's chat about this today.

adrianwyatt: Per our offline discussion, we'll do it this way for now and start bringing up test cases for other models.

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
```diff
@@ -70,24 +70,27 @@ internal static IServiceCollection AddSemanticKernelServices(this IServiceCollec
         // Add the planner.
         services.AddScoped<CopilotChatPlanner>(sp =>
         {
-            // Create a kernel for the planner with the same contexts as the chat's kernel except with no skills.
+            // Create a kernel for the planner with the same contexts as the chat's kernel except with no skills and its own completion backend.
```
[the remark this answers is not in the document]
adrianwyatt: Simple planners can do well with lower cost models and more comprehensive planners (i.e. SequentialPlanner) require expensive models (GPT-4). This allows us to encapsulate the right model for the job. :)

samples/apps/copilot-chat-app/webapi/Skills/OpenApiSkills/KlarnaSkill/openapi.json
```diff
@@ -0,0 +1,120 @@
+{
+  "openapi": "3.0.1",
+  "info": {
+    "version": "v0",
+    "title": "Open AI Klarna product Api"
+  },
+  "servers": [
+    {
+      "url": "https://www.klarna.com/us/shopping"
```
amsacha, line 9: nit: potentially not an issue or a concern. 
But it might make sense to have the user pass in a URL or country name for klarna from the options on https://www.klarna.com/ instead of limiting it to the us which might be slow or blocked for a small number of users.

amsacha (dismissed):

hathind-ms (dismissed):

hathind-ms (approved):

PR #767, opened 2023-05-02
Update QdrantMemoryStore.cs

TerrenceAltman:
Do deleting collection when it exists, seems '!' is redundant


dotnet/src/Connectors/Connectors.Memory.Qdrant/QdrantMemoryStore.cs
```diff
@@ -72,7 +72,7 @@ public IAsyncEnumerable<string> GetCollectionsAsync(CancellationToken cancellati
     /// <inheritdoc/>
     public async Task DeleteCollectionAsync(string collectionName, CancellationToken cancellationToken = default)
     {
-        if (!await this._qdrantClient.DoesCollectionExistAsync(collectionName, cancellationToken).ConfigureAwait(false))
+        if (await this._qdrantClient.DoesCollectionExistAsync(collectionName, cancellationToken).ConfigureAwait(false))
```
cchighman, line 75: I wonder if this check is actually redundant altogether.  In public async Task DeleteCollectionAsync(string collectionName, CancellationToken cancellationToken = default),  we have the following code:

```
 // Deletion is idempotent, ignore error
        if (response.StatusCode == HttpStatusCode.NotFound)
        {
            return;
        }

```
Checking if the collection exists results in two calls to the database regardless of whether it exists whereas without this check, we always have a single call.  If it does in fact not exist, it will return a 404 anyway and the DbClient is already checking for this.    

My thoughts are that you could safely remove this check altogether.  

Just realized we don't seem to surface back the result to the caller unless an exception is thrown.  I imagine we'd want a consistent pattern for return results back to a caller so that future skills or integrations can uniformly rely on those responses for making decisions.

adrianwyatt:
@awharrison-28 could you take a look?

dotnet/src/Connectors/Connectors.Memory.Qdrant/QdrantMemoryStore.cs
tawalke: Great point. I think reducing the trip to the vDb is a great call.

dluc: one thing we've done with ACS is assuming the collection exists. Then, if the service throws an error because the collection doesn't exist, create the collection and retry.

awharrison-28 (changes requested):
The discussion here is good. At a glance, it does seem like a redundant call. Unit tests need to be updated to reflect the change in expected calls.

lemillermicrosoft:
This has been fixed, closing this PR. Thanks @TerrenceAltman for helping to get traction on the issue.

PR #770, opened 2023-05-02
Pinecone dataconnector

Kevdome3000:
### Contribution Checklist


Kevdome3000:
[@shawncal](https://github.com/shawncal) [@dmytrostruk](https://github.com/dmytrostruk)

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeMemoryStore.cs
```diff
@@ -0,0 +1,1078 @@
...
+        {
+            await this._pineconeClient.DeleteIndexAsync(collectionName, cancel).ConfigureAwait(false);
+        }
+    }
+
+    /// <inheritdoc/>
+    /// <param name="collectionName"> in the case of Pinecone, collectionName is synonymous with indexName </param>
+    /// <param name="record"></param>
+    /// <param name="cancel"></param>
+    public async Task<string> UpsertAsync(string collectionName, MemoryRecord record, CancellationToken cancel = default)
```
dmytrostruk, line 163: Method `UpsertAsync` is the same as `UpsertToNamespaceAsync` except the usage of `indexNamespace` parameter. So I think we can define `UpsertAsync` method like this:

```csharp
public async Task<string> UpsertAsync(string collectionName, MemoryRecord record, CancellationToken cancel = default) 
{
    return await this.UpsertToNamespaceAsync(collectionName, string.Empty, record, cancellationToken).ConfigureAwait(false);
}
```

dmytrostruk: **Note:** same applies for:
- `UpsertBatchAsync` and `UpsertBatchToNamespaceAsync`
- `GetAsync` and `GetFromNamespaceAsync`
- `GetBatchAsync` and `GetBatchFromNamespaceAsync`
- `RemoveAsync` and `RemoveFromNamespaceAsync`
- `RemoveBatchAsync` and `RemoveBatchFromNamespaceAsync`
- `GetNearestMatchesAsync` and `GetNearestMatchesFromNamespaceAsync`
- `GetNearestMatchAsync` and `GetNearestMatchFromNamespaceAsync`

Maybe there is no need to keep two methods to support one additional parameter, in case the implementation is the same.

You can just keep one version of it, and add `inputNamespace` parameter as optional with default value `string.Empty`.

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeUtils.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// </summary>
+    /// <param name="documents"></param>
+    /// <returns> A stream of documents with valid metadata. </returns>
+    /// <remarks>
+    ///  If the metadata size is too large, the document will be split into multiple documents.
+    ///  The text field will be removed from the metadata and the document ID will be set to {document ID}-{split counter}
+    ///  The text field will be added to the metadata of the split documents. The split documents will be returned in the
+    ///  same order as the original document.
+    /// </remarks>
+    internal static async IAsyncEnumerable<PineconeDocument> EnsureValidMetadataAsync(
```
dmytrostruk, line 71: Let's add unit tests for this method. We should also verify the performance of this method in case of big text size.

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeUtils.cs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /// <summary>
+    ///  Utility method to split a stream of documents into batches of a given size.
+    /// </summary>
+    /// <param name="data"></param>
+    /// <param name="batchSize"></param>
+    /// <returns></returns>
+    /// <remarks>
+    ///  The recommended batch size as per Pinecone documentation is 100.
```
dmytrostruk, line 135: Let's leave here link to the Pinecone documentation about recommended batch size.

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeUtils.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// <summary>
+    ///  Utility method to convert a dictionary of filters to the format expected by Pinecone.
+    /// </summary>
+    /// <param name="filter"></param>
+    /// <returns></returns>
+    /// <remarks>
+    ///  This is necessary because Pinecone has a different format for filters using the MongoDB query language
+    ///  operators. This method converts the filters to the format expected by Pinecone.
+    /// </remarks>
+    internal static Dictionary<string, object> ConvertFilterToPineconeFilter(Dictionary<string, object> filter)
```
dmytrostruk, line 194: Please add unit tests for this method.

samples/dotnet/kernel-syntax-examples/KernelSyntaxExamples.csproj
```diff
@@ -29,6 +29,10 @@
     <ProjectReference Include="..\..\..\dotnet\src\Connectors\Connectors.Memory.AzureCognitiveSearch\Connectors.Memory.AzureCognitiveSearch.csproj" />
     <ProjectReference Include="..\..\..\dotnet\src\Extensions\Planning.ActionPlanner\Planning.ActionPlanner.csproj" />
     <ProjectReference Include="..\..\..\dotnet\src\Extensions\Planning.SequentialPlanner\Planning.SequentialPlanner.csproj" />
+    <ProjectReference Include="..\..\..\dotnet\src\Connectors\Connectors.Memory.Pinecone\Connectors.Memory.Pinecone.csproj" />
+    <ProjectReference Include="..\..\..\dotnet\src\Connectors\Connectors.Memory.Pinecone\Connectors.Memory.Pinecone.csproj" />
```
dmytrostruk, line 33: Adding Pinecone project reference twice here:
```suggestion

```

samples/dotnet/kernel-syntax-examples/Program.cs
```diff
@@ -97,5 +97,8 @@ public static async Task Main()
 
         await Example30_ChatWithPrompts.RunAsync();
         Console.WriteLine("== DONE ==");
+        
+        await Example29_Pinecone.RunAsync();
+        Console.WriteLine("== DONE ==");
```
dmytrostruk, line 102: Let's put this example before `Example30` 😃

dmytrostruk (changes requested):
First of all, thanks for the awesome job! A lot of new code was added and it's going to be very useful for Semantic Kernel users!

Before merging it to `main`, a couple of things need to be resolved:

- [ ] Resolve all PR comments.
- [ ] Current PR is missing `Pinecone.csproj` in solution file, please add it and verify that solution builds.
- [ ] Resolve warning `Call System.IDisposable.Dispose on object created by 'batch.ToNamespace(indexNamespace).Build()' before all references to it are out of scope	| dotnet\src\Connectors\Connectors.Memory.Pinecone\PineconeClient.cs | Line 249`
- [ ] Resolve all warnings `... change parameter name cancel to cancellationToken in order to match the identifier as it has been declared in ...`. In short, please use `cancellationToken` parameter name instead of `cancel` in all places.
- [ ] Ensure that example in `Example29_Pinecone.cs` file works without any issues.
- [ ] Ensure that unit tests are passed.

As soon as this list is completed, I think the PR will be ready for merge to `main`. Thanks again!

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeMemoryStore.cs
```diff
@@ -0,0 +1,1078 @@
...
+    {
+        if (!await this.DoesCollectionExistAsync(collectionName, cancel).ConfigureAwait(false))
+        {
+            IndexDefinition indexDefinition = IndexDefinition.Create(collectionName);
+            await this._pineconeClient.CreateIndexAsync(indexDefinition, cancel).ConfigureAwait(false);
+        }
+    }
+
+    /// <inheritdoc />
+    /// <returns> a list of index names </returns>
```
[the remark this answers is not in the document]
Kevdome3000: I did this for clarity because Pinecone also has "Collections" but they aren't the same as what "Collections" is meant to mean from the IMemoryStore interface. I think it's important for developers to know that they are getting back names of indexes not collection names.

This speaks to my larger issue that the nomenclature doesn't line up all that well. As a developer, I am likely to use a single pinecone index and multiple namespaces within that index, so when I want a list of "Collection" names, in my case that would mean a list of namespace names. 

It's an issue that will have to be addressed sooner or later, assuming other VectorDB's schemas also don't align perfectly with the IMemoryStore interface.

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeMemoryStore.cs
```diff
@@ -0,0 +1,1078 @@
...
+    /// <returns> a new instance of the <see cref="PineconeMemoryStore"/> class </returns>
+    /// <remarks>
+    ///  This is the preferred method of creating a new instance of the <see cref="PineconeMemoryStore"/>
+    ///  class because it ensures that the index exists and is ready. I think it makes sense to have this
+    ///  method be static because it is a factory method that returns a new instance of the class.
+    ///  If the index does not exist, it will be created. If the index exists, it will be connected to.
+    ///  If it is a new index, the method will block until it is ready.
+    ///  If the index exists but is not ready, it will be connected to and the method will block until it is ready.
+    /// </remarks>
+    public static async Task<PineconeMemoryStore?> InitializeAsync(
```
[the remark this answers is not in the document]
Kevdome3000: I would agree except that currently a PineconeClient for .Net does not exist, so that would mean a user would always have to create an index via Pinecone GUI and never be able to create one on the fly. I got this method from the official Pinecone python client

A common use case I've seen being used with Pinecone else where is creating an index at the beginning of a chat session and then deleting it afterwards.

I added a Constructor that takes a PineconeEnvironment and an api key string as an alternative means of creating a PineconeMemoryStore instance without initializing a new index.

Kevdome3000:
> First of all, thanks for the awesome job! A lot of new code was added and it's going to be very useful for Semantic Kernel users!
> 
> Before merging it to `main`, a couple of things need to be resolved:
> 
> * [ ]  Resolve all PR comments.
> * [ ]  Current PR is missing `Pinecone.csproj` in solution file, please add it and verify that solution builds.
> * [ ]  Resolve warning `Call System.IDisposable.Dispose on object created by 'batch.ToNamespace(indexNamespace).Build()' before all references to it are out of scope	| dotnet\src\Connectors\Connectors.Memory.Pinecone\PineconeClient.cs | Line 249`
> * [ ]  Resolve all warnings `... change parameter name cancel to cancellationToken in order to match the identifier as it has been declared in ...`. In short, please use `cancellationToken` parameter name instead of `cancel` in all places.
> * [ ]  Ensure that example in `Example29_Pinecone.cs` file works without any issues.
> * [ ]  Ensure that unit tests are passed.
> 
> As soon as this list is completed, I think the PR will be ready for merge to `main`. Thanks again!

These have been resolved except I haven't be able to confirm Example31_Pinecone.cs works because I'm getting a 401 from my OpenAI APiKey all of a sudden. You should be able to run it on your end

dmytrostruk (changes requested):
I tested manually and there are a couple of things here:

1. `PineconeEnvironment` is currently an enumeration with predefined values (`us-west1-gcp`, `us-west4-gcp` etc.).
In my free account the environment is different - `us-west1-gcp-free`, so I'm not able to select the valid environment programmatically with current implementation:
image

**Suggestion:** in case if Pinecone environment list can be flexible, maybe it's better to convert that property to `string`. It will also help us to avoid situations when Pinecone will add a new geo/region/environment to their list and we will have to update our enumeration manually. Same with `PodType`, please ensure that this property needs to be `enum` instead of plain string. In case Pod type list is potentially extensible in the future, it is probably better to convert it to string as well.

2. I tried to run current example and it doesn't work, on index creation step I'm receiving `500 Internal Server Error` without error message. Here is the body which we are sending:
```json
{
  "name": "pinecone-test",
  "metric": "cosine",
  "pod_type": "p1.x1",
  "dimension": 0,
  "pods": 0,
  "replicas": 0
}
```

I took this body and tried to send the same parameters in Pinecone playground and got the same problem.
I believe the reason is in `dimension` and/or `pods` and/or `replicas`, it looks like they can't be zero.

In `PineconeMemoryStore` we initialize the index object in the way, that the parameters above are zero:
image

So it looks like a bug.

Please, ensure that the example you provided works without any errors. We need to make sure that the solution we provide is working. If needed, please add more tests to verify that it's working as expected. Thank you!

tawalke:
> I tested manually and there are a couple of things here:
> 
> 1. `PineconeEnvironment` is currently an enumeration with predefined values (`us-west1-gcp`, `us-west4-gcp` etc.).
>    In my free account the environment is different - `us-west1-gcp-free`, so I'm not able to select the valid environment programmatically with current implementation:
>    image
> 
> **Suggestion:** in case if Pinecone environment list can be flexible, maybe it's better to convert that property to `string`. It will also help us to avoid situations when Pinecone will add a new geo/region/environment to their list and we will have to update our enumeration manually. Same with `PodType`, please ensure that this property needs to be `enum` instead of plain string. In case Pod type list is potentially extensible in the future, it is probably better to convert it to string as well.
> 
> 2. I tried to run current example and it doesn't work, on index creation step I'm receiving `500 Internal Server Error` without error message. Here is the body which we are sending:
> 
> ```json
> {
>   "name": "pinecone-test",
>   "metric": "cosine",
>   "pod_type": "p1.x1",
>   "dimension": 0,
>   "pods": 0,
>   "replicas": 0
> }
> ```
> 
> I took this body and tried to send the same parameters in Pinecone playground and got the same problem. I believe the reason is in `dimension` and/or `pods` and/or `replicas`, it looks like they can't be zero.
> 
> In `PineconeMemoryStore` we initialize the index object in the way, that the parameters above are zero: image
> 
> So it looks like a bug.
> 
> Please, ensure that the example you provided works without any errors. We need to make sure that the solution we provide is working. If needed, please add more tests to verify that it's working as expected. Thank you!

On the free account you can only have 1x1 pod. So once an index is created with 1 pod and 1 replica. All subsequent indexed must be 1 pod, 0 replicas.

Kevdome3000:
> > These have been resolved except I haven't be able to confirm Example31_Pinecone.cs works because I'm getting a 401 from my OpenAI APiKey all of a sudden. You should be able to run it on your end
> 
> Hey, I think it's not your API key, the reason is that we updated interface for adding backends and now it doesn't accept service label, so during your testing the key was passed as a wrong parameter. I fixed the example in this commit: [1d32b4d](https://github.com/microsoft/semantic-kernel/commit/1d32b4d914f8ec6650794b8793175a8c77e08f06)
> 
> Please try again and let me know if that works for you.

is this the fix you are refering to?
`
 IKernel kernel = Kernel.Builder
            .WithLogger(ConsoleLogger.Log)
            .Configure(c =>
            {
                c.AddOpenAITextCompletionService("text-davinci-003", Env.Var("OPENAI_API_KEY"));
                c.AddOpenAITextEmbeddingGenerationService("text-embedding-ada-002", Env.Var("OPENAI_API_KEY"));
            })
            .WithMemoryStorage(memoryStore)
            .Build();
`

if so, I am still getting a 401 error and confirmed it's not my api key by calling the open ai api direct to generate an embedding.

I even tried something simple like this and still get the same error.

`
Console.WriteLine( Env.Var( "OPENAI_API_KEY"));
        var embeddingGenerator = new OpenAITextEmbeddingGeneration("text-embedding-ada-002", Env.Var("OPENAI_API_KEY"));
        
        var embedding = await embeddingGenerator.GenerateEmbeddingAsync("My favorite color is orange");

        Console.WriteLine( $"Embedding: {string.Join(", ", embedding)}");
`

here's the error:
Unhandled exception. Microsoft.SemanticKernel.AI.AIException: Access denied: The request is not authorized, HTTP status: 401
   at Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk.ClientBase.RunRequestAsync[T](Func`1 request) in semantic-kernel/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs:line 259
   at Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk.ClientBase.InternalGenerateTextEmbeddingsAsync(IList`1 data, CancellationToken cancellationToken) in semantic-kernel/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs:line 101
   at Microsoft.SemanticKernel.AI.Embeddings.EmbeddingGenerationExtensions.GenerateEmbeddingAsync[TValue,TEmbedding](IEmbeddingGeneration`2 generator, TValue value, CancellationToken cancellationToken) in semantic-kernel/dotnet/src/SemanticKernel.Abstractions/AI/Embeddings/IEmbeddingGeneration.cs:line 47
   at Example32_Pinecone.RunAsync() in semantic-kernel/samples/dotnet/kernel-syntax-examples/Example32_Pinecone.cs:line 29

Kevdome3000:
> 

here is the list of environments from Pinecone:
https://docs.pinecone.io/docs/projects

looks like the us-west1-gcp-free environment was just added so I'll add an enum for it. My view is that keeping it an enum will alleviate a lot of formatting/ misspelling issues making it more developer friendly, and adding another option is simple should pinecone add more environments.

I just created another free account with a p2.x1 (fastest , and the one I prefer) and was given the environment us-west4-gcp , so it's unclear why they have us-west1-gcp-free.

as for creating index with the following body:
```
{
  "name": "pinecone-test",
  "metric": "cosine",
  "pod_type": "p1.x1",
  "dimension": 0,
  "pods": 0,
  "replicas": 0
}
```

Dimension and pods can never be zero. I just added default values for them

Kevdome3000:
Can someone please try the PineconeExample_32 in the KernalSyntaxExamples? I'm still blocked

dmytrostruk:
> looks like the us-west1-gcp-free environment was just added so I'll add an enum for it. My view is that keeping it an enum will alleviate a lot of formatting/ misspelling issues making it more developer friendly, and adding another option is simple should pinecone add more environments.

You didn't add `us-west1-gcp-free` environment in this place:
image

That's what I'm talking about, in order to add new environment as enum value, we should also remember to add it in these two methods. So we either should avoid these two methods and keep `PineconeEnvironment` enum as one source of truth of Pinecone environments, or make it just string. Also, in new code I don't see the usage of these enum values. In `Example32_Pinecone`, we read environment from `Env.Var` as string, for Pinecone API we also use string value:
image

The way I see it, since Pinecone environment is used only to construct API base URL and those environments can be extended in the future, I'm not sure we need to provide enumeration for Semantic Kernel users. If users want to maintain Pinecone environments in specific way, they can always add this enum on their side, but pass string value to our implementation.

dmytrostruk:
Also, looks like there is no Asia environment in `PineconeEnvironment` enumeration:
image

So it's not clear why we list only some of them and not all of them. Again, as I suggested previously, `string` type for Pinecone environment should work fine.

dmytrostruk:
> Can someone please try the PineconeExample_32 in the KernalSyntaxExamples? I'm still blocked

Just tested it. This time I was able to create an index, but I received `TimeoutException` here:
image

My index is still in `Initializing` state in Pinecone, already 10 minutes. We can't block the execution and wait for `Ready `index state for so long, even for 1 minute. It looks like it is better to implement `PineconeMemoryStore` only in the way, when the index already exists. Otherwise, throw an error immediately.

dmytrostruk:
> if so, I am still getting a 401 error and confirmed it's not my api key by calling the open ai api direct to generate an embedding.

I was able to generate an embedding with my OpenAI key, within `Example32_Pinecone`. Please ensure you set valid key in `Env.Var`. But I received `TimeoutException` in `PineconeMemoryStore`, see comment above.

Kevdome3000:
> Also, looks like there is no Asia environment in `PineconeEnvironment` enumeration: image
> 
> So it's not clear why we list only some of them and not all of them. Again, as I suggested previously, `string` type for Pinecone environment should work fine.

Fair point... I'm goin to remove the enum completely

dmytrostruk (changes requested):
**Current status and feedback:**
I was able to save information to Pinecone vectors and then perform a search within `Example32_Pinecone.cs`, which is great! I think we are almost on the finish line 😃🎉  

Remaining issues (sorry for long text, but it's important to describe it in detail):

- [ ] Creating an index in Pinecone is asynchronous operation, and we receive `201 Created`, index state is `Initializing` and we should wait until the state will be `Ready`. When I tried the integration for the first time, my index initialization failed (state was `InitializationFailed`, looks like internal problem in Pinecone). This flow doesn't fit in current `IMemoryStore` implementation, where we require to create an index/collection immediately. 
**Suggestion:** for now we can remove logic for creating an index. In case it doesn't exist in user's account we can throw an exception and specify that index should be created manually before usage (in similar way as table in SQL database, which we usually create only once and then re-use it).

- [ ] Each index (a.k.a. collection) in Pinecone has its own host, which we should use to make API calls. In current implementation, we have `EnsurePineconeClientReadyAsync` in `PineconeMemoryStore` which fetches index host before doing vector operations. Ideally, `PineconeMemoryStore` shoudn't know about specifics of Pinecone API, it should use `PineconeClient` to make API operations without ensuring that the host is ready and if not, then fetching index definition and then performing an operations.
**Suggestion:** Move all logic for "reconnection" (i.e. fetching index host when it's not there) to `PineconeClient` and make it internal. In the same way how we would implement OAuth protocol to fetch access token before making actual request.

- [ ] Caching: in `PineconeClient` we have member `PineconeIndex? _indexDescription` which works as a cache, it stores index information, including host value, which is required for making API calls. If we call `kernel.Memory.SaveInformationAsync` three times with three different index/collection names, `_indexDescription` member will be re-assigned, so it keeps information only about last index.
**Suggestion:** Do we actually need to fetch index host at all?
Index host template is following: `https://{index_name}-{project_id}.svc.{environment}.pinecone.io`, which means that it's constant once you created index, mapping between index name and index host is 1:1.
Instead of `indexName` user can pass `indexHost` and avoid extra calls to fetch that `{project-id}` parameter, which will never change. In case we still want to pass `indexName`, so the user will pass just `pinecone-test` instead of long `pinecone-test-d6e9d39.svc.us-west1-gcp-free.pinecone.io`, then we can make a call and fetch index host, but for caching introduce `ConcurrentDictionary<string, string>`, where `key` will be index name and `value` will be index host.

Kevdome3000:
> @Kevdome3000 please let me know what you think about my last comment above. In case you need help, I can join you and assist in implementation. Thanks for all the hard work and contribution, we are very close to merging!



> **Current status and feedback:** I was able to save information to Pinecone vectors and then perform a search within `Example32_Pinecone.cs`, which is great! I think we are almost on the finish line 😃🎉
> 
> Remaining issues (sorry for long text, but it's important to describe it in detail):
> 
> * [ ]  Creating an index in Pinecone is asynchronous operation, and we receive `201 Created`, index state is `Initializing` and we should wait until the state will be `Ready`. When I tried the integration for the first time, my index initialization failed (state was `InitializationFailed`, looks like internal problem in Pinecone). This flow doesn't fit in current `IMemoryStore` implementation, where we require to create an index/collection immediately.
>   **Suggestion:** for now we can remove logic for creating an index. In case it doesn't exist in user's account we can throw an exception and specify that index should be created manually before usage (in similar way as table in SQL database, which we usually create only once and then re-use it).
> * [ ]  Each index (a.k.a. collection) in Pinecone has its own host, which we should use to make API calls. In current implementation, we have `EnsurePineconeClientReadyAsync` in `PineconeMemoryStore` which fetches index host before doing vector operations. Ideally, `PineconeMemoryStore` shoudn't know about specifics of Pinecone API, it should use `PineconeClient` to make API operations without ensuring that the host is ready and if not, then fetching index definition and then performing an operations.
>   **Suggestion:** Move all logic for "reconnection" (i.e. fetching index host when it's not there) to `PineconeClient` and make it internal. In the same way how we would implement OAuth protocol to fetch access token before making actual request.
> * [ ]  Caching: in `PineconeClient` we have member `PineconeIndex? _indexDescription` which works as a cache, it stores index information, including host value, which is required for making API calls. If we call `kernel.Memory.SaveInformationAsync` three times with three different index/collection names, `_indexDescription` member will be re-assigned, so it keeps information only about last index.
>   **Suggestion:** Do we actually need to fetch index host at all?
>   Index host template is following: `https://{index_name}-{project_id}.svc.{environment}.pinecone.io`, which means that it's constant once you created index, relationship between index name and index host is 1:1.
>   Instead of `indexName` user can pass `indexHost` and avoid extra calls to fetch that `{project-id}` parameter, which will never change.
>   In case we still want to pass `indexName`, so the user will pass just `pinecone-test` instead of long `pinecone-test-d6e9d39.svc.us-west1-gcp-free.pinecone.io`, then we can make a call and fetch index host, but for caching introduce `ConcurrentDictionary<string, string>`, where `key` will be index name and `value` will be index host.

I'm on board with all of the feedback.
My plan was to do as you suggested and move IntitalizeAsync and EnsurePineconeClientReadyAsync into Pinecone client and merge with WaitForIndexState. This way a user can do whatever IndexOperation they may need to do and then when ready create a PineconeMemoryStore instance with the client.

On the point about fetching the index host, I agree that it would be ideal to simply pass the index name and use a Index host template, even Pinecone documentation suggests to use https://index_name-{indexName}.svc.environment.pinecone.io but i was never able to get that to work. I hadn't thought about combining the project id largely because I don't know where to find it. If you can get that to work then great!

Kevdome3000:
> > @Kevdome3000 please let me know what you think about my last comment above. In case you need help, I can join you and assist in implementation. Thanks for all the hard work and contribution, we are very close to merging!
> 
> > **Current status and feedback:** I was able to save information to Pinecone vectors and then perform a search within `Example32_Pinecone.cs`, which is great! I think we are almost on the finish line 😃🎉
> > Remaining issues (sorry for long text, but it's important to describe it in detail):
> > 
> > * [x]   Creating an index in Pinecone is asynchronous operation, and we receive `201 Created`, index state is `Initializing` and we should wait until the state will be `Ready`. When I tried the integration for the first time, my index initialization failed (state was `InitializationFailed`, looks like internal problem in Pinecone). This flow doesn't fit in current `IMemoryStore` implementation, where we require to create an index/collection immediately.
> >   **Suggestion:** for now we can remove logic for creating an index. In case it doesn't exist in user's account we can throw an exception and specify that index should be created manually before usage (in similar way as table in SQL database, which we usually create only once and then re-use it).
> > * [x]   Each index (a.k.a. collection) in Pinecone has its own host, which we should use to make API calls. In current implementation, we have `EnsurePineconeClientReadyAsync` in `PineconeMemoryStore` which fetches index host before doing vector operations. Ideally, `PineconeMemoryStore` shoudn't know about specifics of Pinecone API, it should use `PineconeClient` to make API operations without ensuring that the host is ready and if not, then fetching index definition and then performing an operations.
> >   **Suggestion:** Move all logic for "reconnection" (i.e. fetching index host when it's not there) to `PineconeClient` and make it internal. In the same way how we would implement OAuth protocol to fetch access token before making actual request.
> > * [ ]   Caching: in `PineconeClient` we have member `PineconeIndex? _indexDescription` which works as a cache, it stores index information, including host value, which is required for making API calls. If we call `kernel.Memory.SaveInformationAsync` three times with three different index/collection names, `_indexDescription` member will be re-assigned, so it keeps information only about last index.
> >   **Suggestion:** Do we actually need to fetch index host at all?
> >   Index host template is following: `https://{index_name}-{project_id}.svc.{environment}.pinecone.io`, which means that it's constant once you created index, relationship between index name and index host is 1:1.
> >   Instead of `indexName` user can pass `indexHost` and avoid extra calls to fetch that `{project-id}` parameter, which will never change.
> >   In case we still want to pass `indexName`, so the user will pass just `pinecone-test` instead of long `pinecone-test-d6e9d39.svc.us-west1-gcp-free.pinecone.io`, then we can make a call and fetch index host, but for caching introduce `ConcurrentDictionary<string, string>`, where `key` will be index name and `value` will be index host.
> 
> I'm on board with all of the feedback. My plan was to do as you suggested and move IntitalizeAsync and EnsurePineconeClientReadyAsync into Pinecone client and merge with WaitForIndexState. This way a user can do whatever IndexOperation they may need to do and then when ready create a PineconeMemoryStore instance with the client.
> 
> On the point about fetching the index host, I agree that it would be ideal to simply pass the index name and use a Index host template, even Pinecone documentation suggests to use https://index_name-{indexName}.svc.environment.pinecone.io but i was never able to get that to work. I hadn't thought about combining the project id largely because I don't know where to find it. If you can get that to work then great!

My latest commit should resolve these issues except the question around Caching

dmytrostruk:
@Kevdome3000 I added caching and removed unused methods. Please see my commit here: https://github.com/microsoft/semantic-kernel/pull/770/commits/2dac6b96d96fd9a7ac6c673791e9412b5a91189d
Please let me know if it looks good to you, and if yes, then we are ready to merge it to main.

Kevdome3000:
> @Kevdome3000 I added caching and removed unused methods. Please see my commit here: [2dac6b9](https://github.com/microsoft/semantic-kernel/commit/2dac6b96d96fd9a7ac6c673791e9412b5a91189d) Please let me know if it looks good to you, and if yes, then we are ready to merge it to main.

looks good and thanks for the assist!

dmytrostruk (approved):

dmytrostruk:
@shawncal Could you please take a look and check if we ready to merge this to `main`?

shawncal:
> @shawncal Shawn Callegari FTE Could you please take a look and check if we ready to merge this to `main`?

Certainly. Just updated the branch.

dmytrostruk (dismissed):

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeMemoryStore.cs
```diff
@@ -0,0 +1,711 @@
...
+    /// To make operations within index, its state should be <see cref="IndexState.Ready"/>.
+    /// </remarks>
+    public async Task CreateCollectionAsync(string collectionName, CancellationToken cancellationToken = default)
+    {
+        if (!await this.DoesCollectionExistAsync(collectionName, cancellationToken).ConfigureAwait(false))
+        {
+            throw new PineconeMemoryException(
+                PineconeMemoryException.ErrorCodes.IndexNotReady,
+                $"Index creation is not supported within memory store. " +
+                $"It should be created manually or using {nameof(IPineconeClient.CreateIndexAsync)}. " +
```
awharrison-28, line 71: Is there a reason for not calling CreateIndexAsync within CreateCollectionAsync?

awharrison-28: Is this related to Indexes vs namespaces?

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeMemoryStore.cs
```diff
@@ -0,0 +1,711 @@
...
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        await foreach (MemoryRecord? record in this.GetWithDocumentIdBatchAsync(indexName, new[] { documentId }, limit, indexNamespace, withEmbedding, cancellationToken).ConfigureAwait(false))
+        {
+            yield return record;
+        }
+    }
+
+    /// <summary>
+    /// Get a MemoryRecord from the Pinecone Vector database by a group of pointIds.
```
awharrison-28, line 337: Is the term pointId a relic from the QdrantMemoryStore?

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeUtils.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// <summary>
+    ///  The maximum size of the metadata associated with each vector.
+    ///  See https://docs.pinecone.io/docs/metadata-filtering#supported-metadata-size
+    /// </summary>
+    public const int MaxMetadataSize = 40 * 1024;
+
+    /// <summary>
+    ///  The default dimension for Pinecone vectors. Equivalent to text-embeddings-ada-002 dimension.
+    /// </summary>
+    public const int DefaultDimension = 1536;
```
awharrison-28, line 27: I'd prefer to have this as a required constructor to avoid affinity to ADA

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeUtils.cs
```diff
@@ -0,0 +1,261 @@
...
+
+    public const string DefaultIndexName = "sk-index";
```
awharrison-28, line 29: Since index is synonymous with collection, is this necessary if CreateCollectionAsync is changed to call CreateIndexAsync?

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeMemoryStore.cs
dmytrostruk: @awharrison-28 During `.SaveInformationAsync` in `IMemoryStore` we expect to create collection if it doesn't exist and then use it immediately. 
Unfortunately, index creation in Pinecone is asynchronous, meaning we are receiving `201 Created` response and then we should wait until the index will be provisioned on Pinecone side.

First version of this integration had a logic to wait and poll index provisioning status with some delay. During my testing I noticed that this process could take more than 10 minutes, so this scenario doesn't fit in current `IMemoryStore` implementation.

So, I proposed to remove index creation logic as for now, until Pinecone will provide a way to create an index synchronously, or we redefine `IMemoryStore` to support such cases.

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeMemoryStore.cs
dmytrostruk: Looks like that, I will update it.
```suggestion
    /// Get a MemoryRecord from the Pinecone Vector database by a group of documentIds.
```

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeUtils.cs
dmytrostruk: @awharrison-28 Responded about this in first comment, please let me know if that makes sense.

dotnet/src/Connectors/Connectors.Memory.Pinecone/PineconeUtils.cs
Kevdome3000: MaxMetaDataSize is a constraint set by Pinecone.

DefaultDimension is only used in the creation of a Default IndexDefinition:

```
  public static IndexDefinition Default(string? name = default)
    {
        string indexName = name ?? PineconeUtils.DefaultIndexName;

        return Create(indexName)
            .WithDimension(PineconeUtils.DefaultDimension)
            .WithMetric(PineconeUtils.DefaultIndexMetric)
            .NumberOfPods(1)
            .NumberOfReplicas(1)
            .WithPodType(PineconeUtils.DefaultPodType)
            .WithMetadataIndex(MetadataIndexConfig.Default);
    }
```

I can move all Default IndexDefinition values to the IndexDefinition class file if you prefer.

Kevdome3000: DefaultDimension is only used in the creation of a Default IndexDefinition:

```
  public static IndexDefinition Default(string? name = default)
    {
        string indexName = name ?? PineconeUtils.DefaultIndexName;

        return Create(indexName)
            .WithDimension(PineconeUtils.DefaultDimension)
            .WithMetric(PineconeUtils.DefaultIndexMetric)
            .NumberOfPods(1)
            .NumberOfReplicas(1)
            .WithPodType(PineconeUtils.DefaultPodType)
            .WithMetadataIndex(MetadataIndexConfig.Default);
    }
```

I can move all Default IndexDefinition values to the IndexDefinition class file if you prefer.

lemillermicrosoft (commented):
Consider updating `FEATURE_MATRIX.md` to include these changes in the `Connectors and Skill Libraries` section.

shawncal:
Hey @Kevdome3000 looks like this is nearly ready to go! You'll notice we checked in some new dotnet analyzers today that broke the formatting checks, but that should be easy to fix (most will be automatic). If it's alright with you, I'd like to retarget your PR to a feature branch, "feature-pinecone", so that I can more easily make those style updates for you and run a battery of tests, then merge to main shortly thereafter. Sound good?

Kevdome3000:
> Hey @Kevdome3000 looks like this is nearly ready to go! You'll notice we checked in some new dotnet analyzers today that broke the formatting checks, but that should be easy to fix (most will be automatic). If it's alright with you, I'd like to retarget your PR to a feature branch, "feature-pinecone", so that I can more easily make those style updates for you and run a battery of tests, then merge to main shortly thereafter. Sound good?

Indeed, let me know if you need anything else from

PR #773, opened 2023-05-02
Copilot chat: Avoid duplication in document memories

TaoChenOSU:
### Motivation and Context

Importing documents in Copilot chat will not check for duplication. This will potentially create issues where the same documents are uploaded multiple times.

### Description

1. Enable the document import controller to check for duplication before saving a new memory to the document memory.
2. Allow developers to configure the threshold of duplication (similarity score) in the settings.
3. Move a couple hard-coded similarity score settings to settings.

### Contribution Checklist

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
+++ b/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
@@ -49,8 +49,7 @@ public class DocumentMemoryOptions
 
     /// <summary>
     /// Similarity threshold to avoid document memory duplication when importing documents to memory.
-    /// The higher the value, the more similar document memories will be allowed.
-    /// The lower the value, the more unique document memories will be allowed.
+    /// The higher the value, the more document memories can be similar without being eliminated as duplicates.
     /// </summary>
     [Range(0.0, 1.0)]
     public double DeduplicationSimilarityThreshold { get; set; } = 0.8;
--- a/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
+++ b/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
@@ -48,7 +48,7 @@ public class DocumentMemoryOptions
     public int FileSizeLimit { get; set; } = 1000000;
 
     /// <summary>
-    /// Similarity threshold for avoid document memory duplication when importing documents to memory.
+    /// Similarity threshold to avoid document memory duplication when importing documents to memory.
     /// </summary>
     [Range(0.0, 1.0)]
     public double DeduplicationSimilarityThreshold { get; set; } = 0.8;
--- a/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
+++ b/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
@@ -49,6 +49,8 @@ public class DocumentMemoryOptions
 
     /// <summary>
     /// Similarity threshold to avoid document memory duplication when importing documents to memory.
+    /// The higher the value, the more similar document memories will be allowed.
+    /// The lower the value, the more unique document memories will be allowed.
     /// </summary>
     [Range(0.0, 1.0)]
     public double DeduplicationSimilarityThreshold { get; set; } = 0.8;
--- a/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
+++ b/samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
@@ -46,4 +46,10 @@ public class DocumentMemoryOptions
     /// </summary>
     [Range(0, int.MaxValue)]
     public int FileSizeLimit { get; set; } = 1000000;
+
+    /// <summary>
+    /// Similarity threshold for avoid document memory duplication when importing documents to memory.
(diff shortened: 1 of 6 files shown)


samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
```diff
@@ -46,4 +46,10 @@ public class DocumentMemoryOptions
     /// </summary>
     [Range(0, int.MaxValue)]
     public int FileSizeLimit { get; set; } = 1000000;
+
+    /// <summary>
+    /// Similarity threshold to avoid document memory duplication when importing documents to memory.
+    /// </summary>
+    [Range(0.0, 1.0)]
+    public double DeduplicationSimilarityThreshold { get; set; } = 0.8;
```
adrianwyatt, line 54: Could you add a comment on the logic of the default value? I naively feel its low for dedupe.

samples/apps/copilot-chat-app/webapi/appsettings.json
```diff
@@ -129,13 +129,16 @@
...
+  //   another document that is already in memory.
   //
   "DocumentMemory": {
     "GlobalDocumentCollectionName": "global-documents",
     "UserDocumentCollectionNamePrefix": "user-documents-",
     "DocumentLineSplitMaxTokens": 30,
     "DocumentParagraphSplitMaxLines": 100,
-    "FileSizeLimit": 1000000
+    "FileSizeLimit": 1000000,
+    "DeduplicationSimilarityThreshold": 0.8
```
adrianwyatt, line 141: @glahaye  - do we need to run all appsettings into the deployment logic, or do default  values work as well?

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
TaoChenOSU: It's actually the opposite. The lower the value, the stronger the deduplication will be.
I have added a comments in both the source and the setting Json.

samples/apps/copilot-chat-app/webapi/appsettings.json
glahaye: Right now, there is nothing related to DocumentMemory in the deployment templates.  So, all the config we get is from the appsettings.json document.

These values can be overridden online through the Azure Portal too.

glahaye: So, nothing to change in templates :)

samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
```diff
@@ -46,4 +46,12 @@ public class DocumentMemoryOptions
     /// </summary>
     [Range(0, int.MaxValue)]
     public int FileSizeLimit { get; set; } = 1000000;
+
+    /// <summary>
+    /// Similarity threshold to avoid document memory duplication when importing documents to memory.
+    /// The higher the value, the more similar document memories will be allowed.
```
glahaye, line 52: That comment makes my head spin.

Is it:
The higher the value, the more documents have to be similar to be eliminated as duplicates
OR
The higher the value, the more documents can be similar without being eliminated as duplicates

samples/apps/copilot-chat-app/webapi/Controllers/DocumentImportController.cs
```diff
@@ -181,11 +181,21 @@ private async Task ParseDocumentContentToMemoryAsync(IKernel kernel, string cont
...
                 collection: targetCollectionName,
-                text: paragraph,
-                id: Guid.NewGuid().ToString(),
-                description: $"Document: {documentName}");
+                query: paragraph,
+                limit: 1,
+                minRelevanceScore: this._options.DeduplicationSimilarityThreshold
+            ).ToEnumerable();
+
+            if (!memories.Any())
```
glahaye, line 191: Don't you need to away before using memories?

Is SearchAsync returning an IAsyncEnumerable?

samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
TaoChenOSU: The second one

glahaye: Yeah, maybe replace the two lines of the comment above (and in the settings) with the second line above.

samples/apps/copilot-chat-app/webapi/Controllers/DocumentImportController.cs
TaoChenOSU: `ToEnumerable` is applied on the IAsyncEnumerable instance.

samples/apps/copilot-chat-app/webapi/Config/DocumentMemoryOptions.cs
glahaye: To me, this single line "The higher the value, the more documents can be similar without being eliminated as duplicates" is way easier to understand than the 2 that are currently there.  Unless you have a reason not to, just use that one line

adrianwyatt (dismissed):

glahaye (approved):

Issue #727, opened 2023-04-28
Clean up APIs that take both SKContext and ILogger/CancellationToken


stephentoub:
Some APIs, e.g. Plan.InvokeAsync, accept optional SKContext, ILogger, and CancellationToken instances, e.g.
```C#
public async Task<SKContext> InvokeAsync(SKContext? context = null, CompleteRequestSettings? settings = null, ILogger? log = null, CancellationToken cancellationToken = default);
```
but `SKContext` includes both an `ILogger` and a `CancellationToken`.  That means it's possible to pass in multiple loggers and multiple cancellation tokens, and as currently implemented, in some cases if the SKContext is supplied, the separately provided logger and token will be ignored, even if they're different from the ones inside of the SKContext.  That's a reliability and data loss problem if a developer was expecting their separately-supplied logger and token to be respected.

We can fix this by either:
1) Ensuring that if the supplied logger / token are non-default and different from the ones in the SKContext that they're also respected.  This could mean, for example, creating and passing along a new SKContext that includes a combined token / logger.
2) Avoiding this trap by design by ensuring you can only pass an SKContext or a logger/cancellationToken, but not both.

I'd suggest we do (2).

cc: @shawncal

stephentoub:
@shawncal, @dluc, if you agree, once https://github.com/microsoft/semantic-kernel/pull/612 is merged I'd be happy to do this.

PR #777, opened 2023-05-02
SKFunction Invoke Parameter Duplication

shawncal:
### Motivation and Context
Each of the SKFunction.Invoke calls have multiple ways of passing the same data. It's not clear that, if one way is specified, the other will be ignored.

https://github.com/microsoft/semantic-kernel/issues/727

### Description
This cleans things up such that you either pass a full `SKContext`, which has everything included (CancellationToken, Logger, input), or you pass each of these as separate args, and an SKContext is created for you.

### Contribution Checklist


shawncal:
@stephentoub PR addressing Issue https://github.com/microsoft/semantic-kernel/issues/727

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -23,9 +23,9 @@ public sealed class ContextVariables : IEnumerable<KeyValuePair<string, string>>
     /// Constructor for context variables.
     /// </summary>
     /// <param name="content">Optional value for the main variable of the context.</param>
-    public ContextVariables(string content = "")
+    public ContextVariables(string? content = null)
     {
-        this._variables[MainKey] = content;
+        this._variables[MainKey] = content ?? string.Empty;
```
stephentoub, line 28: Is it important that the MainKey always has a value set for it?  If we didn't do this, _variables could be lazily-initialized only when variables were actually added.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -279,30 +280,35 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context)
...
-        context ??= new SKContext(new ContextVariables(), null!, null, log ?? NullLogger.Instance, cancellationToken);
+        this.State.Update(input);
 
-        context.Variables.Update(input);
+        SKContext context = new(
+            this.State,
+            NullMemory.Instance,
+            null,
+            log ?? NullLogger.Instance,
+            cancellationToken);
```
stephentoub, line 301: The SKContext ctor isn't validating any of these arguments.  Should it be?

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -155,44 +155,29 @@ public FunctionView Describe()
...
-        }
-
-        return this.IsSemantic
-            ? this.InvokeSemanticAsync(context, settings)
-            : this.InvokeNativeAsync(context);
+        SKContext context = new(
+            new ContextVariables(input),
+            NullMemory.Instance,
+            this._skillCollection,
+            log ?? NullLogger.Instance,
```
stephentoub, line 177: Do we expect end-user code to create SKContext instances?  I'm wondering if we should make it easier, e.g. allow the logger to be null and have the ctor pick a default logger if null was passed, or even just allow it to store null.

stephentoub (dismissed):

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
shawncal: yup, good call

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
shawncal: done! update coming

dluc (commented):
The changes to SKFunction will break scenarios where we pass the context only for the internal helpers, and want to pass the input as a string.

shawncal:
> The changes to SKFunction will break scenarios where we pass the context only for the internal helpers, and want to pass the input as a string param.

Yeah I had the same consideration. This is another one in the theme of methods have multiple paths to provide the same arg, and it's unclear which will be used. If I set a context[Input] and then call a Invoke with a empty input (and a valid context), should the input be empty or what was there in context?

I think this makes the API nice and clean:
If you already have a context, go ahead and update it however you want, and pass that in.
If you don't, give us the pieces (input, vars, logger, cancel) -- most are optional -- and we'll build the context for you.
Make sense? I've got an updated PR to send out, and then I'll check in with you on the latest.

Flipping to draft until I have time to update.

stephentoub:
> If you don't, give us the pieces (input, vars, logger, cancel) -- most are optional -- and we'll build the context for you.

It'd also be good for the SKContext.Skills collection to be non-nullable.  If no collection is supplied to the ctor, it can just initialize the Skills to an empty collection.  Then downstream consumption doesn't need to check whether the collection is null or not.

shawncal:
> > If you don't, give us the pieces (input, vars, logger, cancel) -- most are optional -- and we'll build the context for you.
> 
> It'd also be good for the SKContext.Skills collection to be non-nullable. If no collection is supplied to the ctor, it can just initialize the Skills to an empty collection. Then downstream consumption doesn't need to check whether the collection is null or not.

Yup, good call! In fact, I've got something like this in my update... Because the IROSkillCollection interface is in Abstractions and the implementation is in SK.Core, I had to create a NullReadOnlySkillCollection in Abstractions. Just need to merge with that big skill collection change, and I'll get the update sent out.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
```diff
@@ -71,15 +71,6 @@ public async Task CanExecutePlanWithContextAsync()
         // Assert
         Assert.NotNull(result);
         Assert.Equal("Some input", result.Result);
-
```
lemillermicrosoft, line 74:
```suggestion

        plan = new Plan(goal);

        // Act
        context.Variables.Update("other input");
        result = await plan.InvokeAsync(context);

        // Assert
        Assert.NotNull(result);
        Assert.Equal("other input", result.Result);
```

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
```diff
@@ -560,21 +551,20 @@ public async Task CanExecutePlanWithOneStepAndStateAsync()
...
         var mockFunction = new Mock<ISKFunction>();
-        mockFunction.Setup(x => x.InvokeAsync(It.IsAny<SKContext>(), null, null, default))
-            .Callback<SKContext, CompleteRequestSettings, ILogger, CancellationToken?>((c, s, l, ct) =>
+        mockFunction.Setup(x => x.InvokeAsync(It.IsAny<SKContext>(), null))
+            .Callback<SKContext, CompleteRequestSettings>((c, s) =>
                 returnContext.Variables.Update("Here is a poem about " + c.Variables.Input))
             .Returns(() => Task.FromResult(returnContext));
 
         var plan = new Plan(mockFunction.Object);
-        plan.State.Set("input", "Cleopatra");
```
lemillermicrosoft, line 569: I'd hope this one still works. Thoughts?

dotnet/src/SemanticKernel/Planning/Plan.cs
shawncal: Updated with defaults for all values

PR #781, opened 2023-05-03
CopilotChat: Enable MsGraph ToDos in CopilotChat

adrianwyatt:
### Motivation and Context
Enable MsGraph skills in CopilotChat.

### Description
- Added TaskListSkill to CopilotChat when MsGraph authentication is available
- Reduced the webapp's scopes to just User.Read and Tasks.ReadWrite
- Added a "GetDefaultTasks" function with filter for not getting completed tasks.
- Fixed importing Klarna from ChatGPT plugin URL.
- Removed unused methods from the OneDrive connector
- Removed unused semantic skills from CopilotChat


glahaye (dismissed):

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -394,7 +394,12 @@ private bool TryExtractJsonFromPlanResult(string openApiSkillResponse, out strin
         catch (JsonException)
         {
             // Expected if not valid JSON.
-            this._logger.LogDebug("Unable to extract JSON from planner response, it is likely not JSON formatted");
+            this._logger.LogDebug("Unable to extract JSON from planner response, it is likely not from an OpenAPI skill.");
+        }
+        catch (InvalidOperationException)
+        {
+            // Expected if valid JSON but not from an OpenAPI-based skill.
+            this._logger.LogDebug("Unable to extract JSON from planner response, it is likely already proper JSON.");
```
teresaqhoang, line 402: If it's already proper json, shouldn't you return the openApiSkillResponse instead of string.Empty?

adrianwyatt: I'll update the comment. We don't technically know if it's JSON, we just know we didn't get the format we expected. We want to inform the caller that we failed to verify the output and that our out parameter should not be used.

glahaye (dismissed):

glahaye (approved):

samples/dotnet/kernel-syntax-examples/Example21_ChatGPTPlugins.cs
```diff
@@ -40,7 +43,7 @@ private static async Task RunChatGptPluginAsync()
 
         //var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
```
amsacha, line 44: Why is this commented out? There are other examples in kernel-syntax-examples that will fail if the user doesnt replace place holders with their appropriate values. These examples are also not part of any tests that could fail

adrianwyatt: I'm not sure - let's take a look at this outside of this PR.

samples/dotnet/kernel-syntax-examples/Example21_ChatGPTPlugins.cs
```diff
@@ -17,14 +18,16 @@ public static async Task RunAsync()
     private static async Task RunChatGptPluginAsync()
     {
         var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
+        using HttpClient importHttpClient = new HttpClient();
+        importHttpClient.DefaultRequestHeaders.Add("User-Agent", "Microsoft.SemanticKernel");
```
gitri-ms, line 22: Should be consistent in the value we use for user agent. Looks like elsewhere in this PR we are using "Microsoft-Semantic-Kernel"

adrianwyatt: Good catch!

PR #786, opened 2023-05-03
Extending list of retryable codes

SergeyMenshykh:

This change is required to improve SK SDK resilience while working with REST API by extending list of HTTP status codes DefaultHttpRetryHandler retries on.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Reliability/HttpRetryConfig.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Reliability/HttpRetryConfig.cs
@@ -59,7 +59,11 @@ public sealed class HttpRetryConfig
         (HttpStatusCode)HttpStatusCodeType.RequestTimeout,
         (HttpStatusCode)HttpStatusCodeType.ServiceUnavailable,
         (HttpStatusCode)HttpStatusCodeType.GatewayTimeout,
-        (HttpStatusCode)HttpStatusCodeType.TooManyRequests
+        (HttpStatusCode)HttpStatusCodeType.TooManyRequests,
+        (HttpStatusCode)HttpStatusCodeType.InternalServerError,
+        (HttpStatusCode)HttpStatusCodeType.NotImplemented,
+        (HttpStatusCode)HttpStatusCodeType.BadGateway,
+        (HttpStatusCode)HttpStatusCodeType.InsufficientStorage
     };
 
     /// <summary>


dotnet/src/SemanticKernel.Abstractions/Reliability/HttpRetryConfig.cs
```diff
@@ -59,7 +59,11 @@ public int MaxRetryCount
         (HttpStatusCode)HttpStatusCodeType.RequestTimeout,
         (HttpStatusCode)HttpStatusCodeType.ServiceUnavailable,
         (HttpStatusCode)HttpStatusCodeType.GatewayTimeout,
-        (HttpStatusCode)HttpStatusCodeType.TooManyRequests
+        (HttpStatusCode)HttpStatusCodeType.TooManyRequests,
+        (HttpStatusCode)HttpStatusCodeType.InternalServerError,
+        (HttpStatusCode)HttpStatusCodeType.NotImplemented,
```
lemillermicrosoft, line 64: Thoughts on NotImplemented and InsufficientStorage? Those feel less likely to be transient issues. Curious if you have some scenarios where these have been helpful.

adrianwyatt (approved):

dotnet/src/SemanticKernel.Abstractions/Reliability/HttpRetryConfig.cs
```diff
@@ -59,7 +59,9 @@ public int MaxRetryCount
         (HttpStatusCode)HttpStatusCodeType.RequestTimeout,
         (HttpStatusCode)HttpStatusCodeType.ServiceUnavailable,
         (HttpStatusCode)HttpStatusCodeType.GatewayTimeout,
-        (HttpStatusCode)HttpStatusCodeType.TooManyRequests
+        (HttpStatusCode)HttpStatusCodeType.TooManyRequests,
+        (HttpStatusCode)HttpStatusCodeType.InternalServerError,
```
dluc, line 63: 500 is not universally retriable or not, it depends on the scenario.

dluc: 500 is typically an uncontrolled exception, returned by services that are suffering an unexpected outage. Retrying by default will hurt several services trying to recover and having at the same time to cope with extra traffic caused by clients in a retry loop.

adrianwyatt: Correct - we should not retry on "500". 502,503, and 504 are retry-able: https://learn.microsoft.com/en-us/dotnet/architecture/cloud-native/application-resiliency-patterns

adrianwyatt, line 63: Let's remove "InternalServerError"

adrianwyatt (changes requested):

dotnet/src/SemanticKernel.Abstractions/Reliability/HttpRetryConfig.cs
SergeyMenshykh: No problem, 500 can be easily excluded from the default config and included, if needed.

RogerBarreto: Should we consider:

425(Too Early)? 
507 InsufficientStorage?

RogerBarreto:
Should we consider:

425(Too Early)?
507 InsufficientStorage?

adrianwyatt (approved):

SergeyMenshykh:
> Should we consider:
> 
> 425(Too Early)? 507 InsufficientStorage?

Not sure about 425 one because it's related to a specific scenario when a client wants to send data in advance of the secure connection being set up to optimize the time required for TLS to create a secure connection.

There're no clear guidelines about the 507 one. Some sources say that it's a temporary issue implying that it can be retried the other ones say that it should not be retried in a scope of the same action?

Taking into account that it's always possible to add any additional status code when configuring Kernel, I would not add these codes now, and wait until they are requested be a default ones.

adrianwyatt:
> Should we consider:
> 
> 425(Too Early)? 507 InsufficientStorage?

Good callout - we'll use this PR to align with MS practices, but I would welcome an issue for 425/507

PR #787, opened 2023-05-03
Method name is added to exception message.

SergeyMenshykh:

This PR adds function/method name into exception message that is thrown if a native function, marked with SKFunctionInputAttribute, doesn't have an input parameter of string type. The method name can be useful for debugging and troubleshooting.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
+++ b/dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
@@ -513,7 +513,7 @@ public sealed class SKFunction : ISKFunction, IDisposable
 
         if (!result.HasSkFunctionAttribute || skFunctionAttribute == null)
         {
-            log?.LogTrace("Method {0} doesn't have SKFunctionAttribute", result.Name);
+            log?.LogTrace("Method '{0}' doesn't have SKFunctionAttribute", result.Name);
             return result;
         }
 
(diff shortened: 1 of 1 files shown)


dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -554,9 +554,8 @@ private static MethodDetails GetMethodDetails(MethodInfo methodSignature, object
         else if (skMainParam != null)
         {
             // The developer used [SKFunctionInput] on a function that doesn't support a string input
```
shawncal, line 556: (preexisting problem) I don't believe this exception is valid -- if the user doesn't specify a string, they can still get the "input" from SKContext. This should only be thrown if the function has neither a SKContext arg nor a string arg.

SergeyMenshykh: There could be functions having no parameters like those from TimeSkill class. Is it something we plan to keep as it's or do we plan to update them to accept SKContext even so they don't need it?

dluc: supporting functions without a context allows some flexibility which I believe is nice to have

dluc: @shawncal `SKFunctionInputAttribute` is used for `string input`. It's true that one can pass the input via the context, but the `SKFunctionInputAttribute` attribute is specific to the string param

shawncal (approved):

PR #788, opened 2023-05-03
Some clean up in OpenAIClientBase

stephentoub:
### Motivation and Context

Fix some possible functional issues as well as remove unnecessary code.

### Description

- The type accepts an HttpClient instance that it'll use.  It shouldn't then mutate that instance, as it might be shared with other consumers.  But it's adding to its DefaultRequestHeaders collection, which will impact all other users of that instance (DefaultRequestHeaders also isn't thread-safe and should only ever be added to while it's not being used for issuing requests).  I've removed the interaction with DefaultRequestHeaders and instead added the relevant headers to each message being sent.
- The base is exposing a lot of protected surface area, but it's not actually extensible outside of the assembly.  I've changed all of its internal and protected members to instead be "private protected" or just "private", reflecting their actual usage and ensuring the surface area doesn't show up in public documentation (which protected members would).
- As it's not actually extensible publicly, and as no derived types are finalizable, it needn't implement the full IDisposable pattern nor use GC.SuppressFinalize.
- Both the base and the derived image generation client had duplicated APIs for url vs base64 image generation.  I've consolidated that via arguments to just have one copy of the code. Note that the base64 image generation was private and unused, so I've deleted that unused member, but the functionality can be exposed easily if desired based on the shared routines now in place.
- Removed some unnecessary null checking for things that'll never be null.
- Ensured the HttpClient response message is disposed.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/OpenAIImageGeneration.cs
+++ b/dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/OpenAIImageGeneration.cs
@@ -1,12 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
-using System.Linq;
+using System;
+using System.Diagnostics;
 using System.Net.Http;
-using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
-using Microsoft.SemanticKernel.AI;
 using Microsoft.SemanticKernel.AI.ImageGeneration;
 using Microsoft.SemanticKernel.Connectors.AI.OpenAI.CustomClient;
 using Microsoft.SemanticKernel.Diagnostics;
@@ -16,8 +15,20 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.ImageGeneration;
 
 public class OpenAIImageGeneration : OpenAIClientBase, IImageGeneration
 {
-    // 3P OpenAI REST API endpoint
+    /// <summary>
+    /// OpenAI REST API endpoint
+    /// </summary>
     private const string OpenAIEndpoint = "https://api.openai.com/v1/images/generations";
+
+    /// <summary>
+    /// Optional value for the OpenAI-Organization header.
+    /// </summary>
+    private readonly string? _organizationHeaderValue;
+
+    /// <summary>
+    /// Value for the authorization header.
+    /// </summary>
+    private readonly string _authorizationHeaderValue;
 
     /// <summary>
     /// Create a new instance of OpenAI image generation service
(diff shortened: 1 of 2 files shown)


lemillermicrosoft (approved):
LGTM

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/OpenAIImageGeneration.cs
```diff
@@ -34,50 +45,55 @@ public class OpenAIImageGeneration : OpenAIClientBase, IImageGeneration
...
+    /// <summary>Adds headers to use for OpenAI HTTP requests.</summary>
+    private protected override void AddRequestHeaders(HttpRequestMessage request)
+    {
+        base.AddRequestHeaders(request);
+
+        request.Headers.Add("Authorization", this._authorizationHeaderValue);
+        if (!string.IsNullOrEmpty(this._organizationHeaderValue))
         {
-            this.HTTPClient.DefaultRequestHeaders.Add("OpenAI-Organization", organization);
+            request.Headers.Add("OpenAI-Organization", this._organizationHeaderValue);
```
dluc, line 60: wouldn't it be faster doing it once in the ctor, using DefaultRequestHeaders, as opposed to checking if the value is not empty on each request?

stephentoub: As is noted in the PR description, we can't use DefaultRequestHeaders.  If we're accepting an HttpClient from a caller, adding to DefaultRequestHeaders will end up setting that header for any other requests they make on that same HttpClient, which makes other use of that instance a bug.  It's also not thread-safe to modify DefaultRequestHeaders while it might be in use, so if someone is making calls on that HttpClient while this instance is being constructed and mutating DefaultRequestHeaders, that's a bug.  From a perf perspective, it's also a wash, as all of the headers need to be copied from DefaultRequestHeaders to these request headers, with various checks and enumerations and whatnot along the way.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/OpenAIImageGeneration.cs
```diff
@@ -34,50 +45,55 @@ public class OpenAIImageGeneration : OpenAIClientBase, IImageGeneration
...
 
-        return this.GenerateImageUrlAsync(description, width, height, cancellationToken);
+        return this.GenerateImageAsync(description, width, height, 1, "url", x => x.Url, cancellationToken);
     }
 
-    private async Task<string> GenerateImageUrlAsync(string description, int width, int height, CancellationToken cancellationToken = default)
+    private async Task<string> GenerateImageAsync(
+        string description,
+        int width, int height,
+        int count, string format, Func<ImageGenerationResponse.Image, string> extractResponse,
```
shawncal, line 79: if we're passing in the `count`, we should also modify the return type to Task<IEnumerable<string>>, change name to `GenerateImagesAsync`, and return the whole list.

stephentoub: good point, I'll put it back to a const 1 for now

shawncal (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/OpenAIImageGeneration.cs
shawncal: I agree, I think this has been answered. Resolving!

PR #792, opened 2023-05-03
GitHub OpenAPI skill integration with planner

teresaqhoang:
### Motivation and Context

This PR enables GitHub integration with Copilot Planner. 

### Description

Server-side
- Added C# classes for GitHub schema objects
- Adds tokenLimit for Planner response content. Currently, it's set at 75% of the remaining limit after memories response and document context have already been allocated.
- Added `OptimizeOpenApiSkillJson` method that is called on Planner completion. This method:
     - removes all new line characters (huge token suck) from API response
     - truncates the API response by object until it's under the relatedInformationTokenLimit. 
     - If the request can't be truncated, error message added to bot context
     - For GitHub skills specifically, the API response is deserialized into `PullRequest` or `PullRequest[]` objects to filter noisy properties.

Web app side
- If plugins require additional api requirements, these are added to the ask variables (rather than headers in http request) -- this allows the kernel/planner to directly consume these without any additional handling server side
- `repo` and `owner` added as additional GitHub api requirements

image

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
+++ b/samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
@@ -3,6 +3,7 @@
 using System.Globalization;
 using System.Text.Json;
 using System.Text.Json.Nodes;
+using System.Text.RegularExpressions;
 using Microsoft.SemanticKernel;
 using Microsoft.SemanticKernel.AI.TextCompletion;
 using Microsoft.SemanticKernel.Memory;
@@ -11,6 +12,7 @@ using Microsoft.SemanticKernel.Planning;
 using Microsoft.SemanticKernel.SkillDefinition;
 using SemanticKernel.Service.Config;
 using SemanticKernel.Service.Model;
+using SemanticKernel.Service.Skills.OpenApiSkills;
 using SemanticKernel.Service.Storage;
 
 namespace SemanticKernel.Service.Skills;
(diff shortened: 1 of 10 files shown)


samples/apps/copilot-chat-app/webapi/Skills/OpenApiSkills/GitHubSkill/Model/Label.cs
```diff
@@ -0,0 +1,40 @@
+using System.Text.Json.Serialization;
+
+namespace SemanticKernel.Service.Skills.OpenApiSkills;
+
+/// <summary>
+/// Represents a label.
```
glahaye, line 6:
```suggestion
/// Represents a pull request label.
```

glahaye (dismissed):

samples/apps/copilot-chat-app/webapi/Skills/PromptSettings.cs
```diff
@@ -34,6 +34,13 @@ public PromptSettings(PromptsConfig promptsConfig)
     /// </summary>
     internal double DocumentContextWeight { get; } = 0.3;
 
+    /// <summary>
+    /// Weight of information returned from planner (i.e., responses from OpenAPI skills).
+    /// Percentage calculated from remaining token limit after memories response and document context have already been allocated.
+    /// Contextual prompt excludes all the system commands.
+    /// </summary>
+    internal double RelatedInformationContextWeight { get; } = 0.75;
```
amsacha, line 42: The comments state that this is calculated from the remaining token limit, but in practice this is being hard coded, can you explain why in the comments

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -404,6 +409,83 @@ private bool TryExtractJsonFromOpenApiPlanResult(string openApiSkillResponse, ou
...
+    {
+        int jsonTokenLimit = (int)(tokenLimit * 0.8);
+
+        // Remove all new line characters + leading and trailing white space
+        jsonContent = Regex.Replace(jsonContent.Trim(), @"[\n\r]", string.Empty);
+        var document = JsonDocument.Parse(jsonContent);
+        string lastSkillInvoked = plan.Steps[^1].SkillName;
+
+        // Check if the last skill invoked was GitHubSkill and deserialize the JSON content accordingly
+        if (string.Equals(lastSkillInvoked, "GitHubSkill", StringComparison.Ordinal))
```
amsacha, line 426: I think our long term approach should be to ingest the operation response in parts. This is beyond the scope of things we are doing but worth discussing with Sergey. The swagger schema defines what the response for a particular operation will be. This response can be very big but if we can split the response schema into multiple parts we can acquire them individually, summarize them and combine them all together. That would prevent us from having to spin up a new class to determine how to parse the json.

All of that being said, can we not achieve the same result by slimming down the actual swagger.json response for the operation?

samples/apps/copilot-chat-app/webapi/Skills/PromptSettings.cs
teresaqhoang: Good catch! I forgot to change that during testing, ChatSkill leverages the PromptSettings variable now

samples/apps/copilot-chat-app/webapi/Skills/OpenApiSkills/GitHubSkill/Model/PullRequest.cs
```diff
@@ -0,0 +1,115 @@
...
+    [JsonPropertyName("number")]
+    public int Number { get; set; }
+
+    /// <summary>
+    /// Gets or sets the state of the pull request
+    /// </summary>
+    [JsonPropertyName("state")]
+    public string State { get; set; }
+
+    // Whether the pull request is locked
```
adrianwyatt, line 34: Missing summary open tag.

samples/apps/copilot-chat-app/webapi/Skills/OpenApiSkills/GitHubSkill/Model/Repo.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Initializes a new instance of the <see cref="Repo"/>.
+    /// </summary>
+    /// <param name="name">The name of the repository, e.g. "dotnet/runtime".</param>
+    /// <param name="fullName">The full name of the repository, e.g. "Microsoft/dotnet/runtime".</param>
+    public Repo(string name, string fullName)
+    {
+        this.Name = name;
+        this.FullName = fullName;
+    }
+
```
adrianwyatt, line 33: [nit] extra line

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
teresaqhoang: I don't want to modify the schema anymore than limiting it to APIs we want to leverage. Operations should come as defined from the publisher

Rest APIs are meant for publishers to communicate data in a standardized way, and once the person requesting that information gets the response, they can manipulate that data as they see fit for their application. In our case, leaving the properties in the GitHub schema indicates that these properties are still available if needed, but since we don't want them for our application, we cut them out on deserialization (kinda like model vs view state).

Long term, Sergey has work planned to parse and expose the response schema using the RestApiOperation object the kernel creates. What we do with that schema is still open for discussion (some ideas is using the AI to semantically identifying properties related to user intent and extracting only those fields from the response). Let's take this conversation offline and schedule a meeting to discuss all our ideas together.

PR #797, opened 2023-05-03
OpenAPI Skill server-url configurations

teresaqhoang:
### Motivation and Context

**1. Add default parameter values for available functions on Plan creation (ActionPlanner)** 

- This commit changes the `PopulateList` function to include the default value of each parameter in the generated string. This makes the list of available functions more informative and helpful for LLM to know the default behavior of a function. The default value is shown in parentheses after the parameter description.

- This is helpful in parsing the correct server-url for OpenAPI functions (before, the planner was suggesting arbitrary values that were more-than-intermittently wrong).

**2. Add support to override the server-url from user input**

- User input is passed as customer variable in Ask context when SK is invoked in the format {openApiSkillName}-server-url
- This can then be parsed in the `SemanticKernelController` and passed to the Import method
i.e., 
```
if (openApiSkillsAuthHeaders.JiraAuthentication != null)
        {
            this._logger.LogInformation("Registering Jira Skill");
            var authenticationProvider = new BasicAuthenticationProvider(() => { return Task.FromResult(openApiSkillsAuthHeaders.JiraAuthentication); });
            var hasServerUrlOverride = variables.Get("jira-server-url", out string serverUrlOverride);

            await planner.Kernel.ImportOpenApiSkillFromFileAsync(
                skillName: "JiraSkill",
                filePath: Path.Combine(System.IO.Directory.GetCurrentDirectory(), @"Skills/OpenApiSkills/JiraSkill/openapi.json"),
                authCallback: authenticationProvider.AuthenticateRequestAsync,
                serverUrlOverride: hasServerUrlOverride ? serverUrlOverride : null
                );
        }
```

3. Adds missing user-agent parameter to all ImportOpenApiSkill methods for consistency


amsacha (approved):

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -118,6 +118,13 @@ public async Task<Plan> CreatePlanAsync(string goal)
             plan = new Plan(goal);
         }
 
+        // Planner should not suggest server-url as a parameter.
+        // Server URL should be be parsed from OpenAPI spec or provided by user
+        if (planData.Plan.Parameters.ContainsKey("server-url"))
```
lemillermicrosoft, line 123: I think the better fix would be to update the functions for specifying functions to include their default values. See `defaultValue` usage in dotnet\src\Extensions\Planning.SequentialPlanner\FunctionViewExtensions.cs

adrianwyatt: Things like Jira won't have a default url directly associated with the specification. We need to be able to provide this at runtime.

adrianwyatt, line 123: There is no need to do a "ContainsKey" check if all you're doing is removing. Dictionary.Remove will not fail if the key does not exist.

adrianwyatt: @teresaqhoang  - will this change prevent proper usage of default values?

adrianwyatt (changes requested):

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
teresaqhoang: No it won't, the default value lives in the plan's kernel's context. This is just removing the server-url parameter from the plan's suggested parameters collection. The following lines 129-141 update the planner's kernel's context with the plan's suggested parameters (which is why server-url was being overridden before).

adrianwyatt:
It feels a little odd to have this logic in the ActionPlanner itself. Is it possible to move it to the CopilotChatPlanner shim (genuinely curious)?

lemillermicrosoft (changes requested):
Agree with Adrian, this kind of business logic should not be in the Planner. I believe the appropriate fix is to supply the default value in the manual. See #647

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
lemillermicrosoft: I would do that on plan consumption or creation. Who supplies the server-url? Can it be part of the planning prompt? Or set the value in the context you give to the plan invocation to override the plan parameters.

adrianwyatt (dismissed):

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
amsacha: Also this isnt fool proof, if the planner doesnt formulate the ask with the server-url as an input, it can fail much earlier, with a "Plan parsing error, invalid JSON". This happened a bunch for jira because the openapi spec cant provide a usable default value for the server url. And atleast right now the server url we ask users to enter is not being passed along to the backend. I had to hard code the variable to get consistent results.

teresaqhoang: The "Plan parsing error, invalid JSON" error is due to a different bug in ActivePlanner -- the planner tries to return 2 plans so the resulting JSON is invalid. @amsacha if you come across this error again, can you confirm? You can do this by putting a breakpoint right after the CreatePlanAsync function is called in ChatSkill and inspecting the plan (line 214 in the screenshot)
Screenshot 2023-05-03 at 8 05 19 PM

I don't think hardcoding the value is the best method, it's passed in the context variables now, so maybe we can add a wrapper in Plan.cs when evaluating namedParameters?

dotnet/src/Extensions/Planning.ActionPlanner/FunctionViewExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Microsoft.SemanticKernel.Planning.Action;
+
+internal static class FunctionViewExtensions
+{
+    /// <summary>
+    /// Create a manual-friendly string for a function.
+    /// </summary>
+    /// <param name="function">The function to create a manual-friendly string for.</param>
+    /// <returns>A manual-friendly string for a function.</returns>
+    internal static string ToManualString(this FunctionView function)
```
lemillermicrosoft, line 15: Since we're duplicating this code, it might be worth preserving the previous formatting as much as possible (while also using `DefaultValue` to fix the bug) -- I know @dluc wanted to promote lower token usage for ActionPlanner. Curoius your thoughts, too.

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
```diff
@@ -49,6 +50,9 @@ public static class KernelOpenApiExtensions
         AuthenticateRequestAsyncCallback? authCallback = null,
         string? userAgent = "Microsoft-Semantic-Kernel",
         HttpRetryConfig? retryConfiguration = null,
+#pragma warning disable CA1054 // URI-like parameters should not be strings
```
lemillermicrosoft, line 53: Who is going to call with this parameter? I don't see it being used.

dotnet/src/Extensions/Planning.ActionPlanner/FunctionViewExtensions.cs
teresaqhoang: Added default value to existing `PopulateList` function

adrianwyatt (dismissed):

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelOpenApiExtensions.cs
```diff
@@ -49,6 +50,9 @@ public static class KernelOpenApiExtensions
...
+        string? serverUrlOverride = null,
```
lemillermicrosoft, line 54: Let's get rid of the pragma and use `Uri` type. It's done elsewhere. If there is a reason it can't be done, let's add that reasoning here.

adrianwyatt (approved):

dotnet/src/Skills/Skills.UnitTests/OpenAPI/OpenApiDocumentParserV31Tests.cs
```diff
@@ -92,7 +92,7 @@ public async Task ItCanParsePutOperationMetadataSuccessfullyAsync()
         var putOperation = operations.Single(o => o.Id == "SetSecret");
         Assert.NotNull(putOperation);
         Assert.Equal("Sets a secret in a specified key vault.", putOperation.Description);
```
lemillermicrosoft, line 94: nit: tests for the optional parameter in `GetParameters` would be good, feel free to resolve for follow up though.

lemillermicrosoft (approved):

PR #798, opened 2023-05-04
Fix code style

dluc:


samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -20,6 +18,7 @@
 using SemanticKernel.Service.Model;
 using SemanticKernel.Service.Skills;
 using SemanticKernel.Service.Storage;
+using Directory = System.IO.Directory;
```
stephentoub, line 21: Why do this rather than `using System.IO;`?

dluc: less code/readability (btw, I don't do this manually, it's an automation code style fix in R#)

stephentoub: `using System.IO;` is less code than `using Directory = System.IO.Directory;` :)

dluc: sorry, I see what you mean, there's an ambiguous ref, the same Directory class in MsGraph lib

stephentoub: > another Directory class in MsGraph lib

😦

hathind-ms (approved):

PR #799, opened 2023-05-04
Roll back default language version from 11 to 10

adrianwyatt:
### Motivation and Context
C# 11 requires .NET 7+ and we would like to keep the requirement at .NET 6 for now.

- includes a minor cosmetic cleanup for the user-agent name in example 21
- removed unused common.props


stephentoub:
> C# 11 requires .NET 7+ and we would like to keep the requirement at .NET 6 for now.

The .NET 7 SDK. You can still run on .NET 6.  Why do you need to stay on the .NET 6 SDK?

shawncal:
> > C# 11 requires .NET 7+ and we would like to keep the requirement at .NET 6 for now.
> 
> The .NET 7 SDK. You can still run on .NET 6. Why do you need to stay on the .NET 6 SDK?

I think this is because we have all these tutorials for sample apps that say to clone the repo, build a few projects, and go. Those projects aren't all .net 7 yet (we were trying to maximize compatibility). The long-term fix here is to ensure all those samples are pulling from our nuget feed and not using project references (...but then they also don't automatically inherit our refactoring changes like renaming things everywhere), or to just update all the samples to 7.0 and make that a requirement in the getting started docs.

I'm ok with a temporary reversion here while the SDK is still under development and names are changing. I'll capture a task to update samples and update SDK lang version.

@stephentoub Does that sound okay?

adrianwyatt:
> > C# 11 requires .NET 7+ and we would like to keep the requirement at .NET 6 for now.
> 
> The .NET 7 SDK. You can still run on .NET 6. Why do you need to stay on the .NET 6 SDK?

We will need to update documentation and other assets to state having .NET 6 as a requirement. I don't think we are averse to it, we just want to be intentional about it.

shawncal:
@adrianwyatt Am I right about the reason for the reversion above? Where did this surface?

stephentoub:
> Does that sound okay?

Yup.  You get to decide what .NET SDK you require, and I can understand wanting to stick with .NET 6 for now (e.g. I could imagine a valid answer being "we want to advertise that we only require an LTS version").  I just want to make sure we're encouraging staying relatively up-to-date, and there are nice features in C# 11 for this domain, e.g. raw string literals.

adrianwyatt:
> @adrianwyatt Am I right about the reason for the reversion above? Where did this surface?

Yup!

samples/dotnet/kernel-syntax-examples/Example21_ChatGPTPlugins.cs
```diff
@@ -19,7 +19,7 @@ private static async Task RunChatGptPluginAsync()
     {
         var kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
         using HttpClient importHttpClient = new HttpClient();
-        importHttpClient.DefaultRequestHeaders.Add("User-Agent", "Microsoft.SemanticKernel");
+        importHttpClient.DefaultRequestHeaders.Add("User-Agent", "Microsoft-Semantic-Kernel");
```
shawncal, line 22: Reviewers note: spoke with @adrianwyatt about this. Technically not related to the lang version change, but confirmed the reason. I checked other uses of the User-Agent throughout the repo and confirmed this is just making it match what we have elsewhere (see https://github.com/microsoft/semantic-kernel/blob/05c586b9d8c040e33b47630c969495b2d7facd8b/dotnet/src/Connectors/Connectors.AI.OpenAI/CustomClient/OpenAIClientBase.cs#L189). Approved.

shawncal (approved):

shawncal (approved):

PR #805, opened 2023-05-04
fix implementation to return embedding

hsm207:
### Motivation and Context

   because calling the memory record's embedding property throws `RecursionError: maximum recursion depth exceeded`
    calling the memory record's embedding property throws `RecursionError: maximum recursion depth exceeded`
4. What scenario does it contribute to?
    people calling the embedding attribute of a memory memory record
5. If it fixes an open issue, please link to the issue here.
    NA
### Description

return the attribute instead of the property

### Contribution Checklist


shawncal (approved):
Good find! Thank you

PR #813, opened 2023-05-04
Added timestamp property to chat session model

dbarkol:
Adding a timestamp property to the ChatSession object.

This will help sort sessions for users who have multiple chats.


TaoChenOSU:
Hello! 
We have internally noticed this problem and was planning to address that!
Thank you so much for your PR!

I have 2 questions regarding the `Timestamp` property in the `ChatSession` class:
1. It's not used anywhere in this PR. How is this property going to be used?
2. It's never updated once a `ChatSession` object is created. We may want to sort the chat sessions by the most recent activities. Is there another PR coming that will update this property?

Just a thought: Can we use the timestamps of the most recent chat messages to sort the chat sessions?

samples/apps/copilot-chat-app/webapi/Model/ChatSession.cs
```diff
@@ -28,10 +28,17 @@ public class ChatSession : IStorageEntity
     [JsonPropertyName("title")]
     public string Title { get; set; }
 
+    /// <summary>
+    /// Timestamp of the message.
+    /// </summary>
+    [JsonPropertyName("timestamp")]
+    public DateTimeOffset Timestamp { get; set; }
```
TaoChenOSU, line 35: This new property is not being used anywhere.  Unclear how this is going to be used.

TaoChenOSU (changes requested):

samples/apps/copilot-chat-app/webapi/Model/ChatSession.cs
dbarkol: We plan on using the timestamp property in a different UI/application. Users will see a list of chat sessions and pick one that they would like to use. 

The WebAPI project is fantastic, and we'd like to use it in multiple applications. The hope is that it stays decoupled from the existing UI sample, as it is today, since it provides a lot of great code and good patterns.

TaoChenOSU: Glad to hear you are creating another application with the WebAPI!

Some feedback:
If you need to have a time property in the chat sessions, we suggest renaming the property to `CreatedOn` if it's going to be read only. If you are planning to modify the timestamp, consider renaming it to `LastModified` and include the logic that will update the property.

dbarkol: That makes a lot of sense. :)

Timestamp was chosen to be more consistent with the ChatMessage object but CreatedOn is more explicit and seems like a better option. We can make the update if you think it is a reasonable PR. Seems like it would only help.

TaoChenOSU: Yes, totally agree on keeping it consistent with the `ChatMessage`. I didn't want this PR to include the changes to `ChatMessage` because it will involve changes to the WebApp too. But if you are inclined, feel free to make changes to `ChatMessage` too (and the WebApp).

TaoChenOSU:
Will merged this PR and create a subsequence PR to rename this variable to `CreatedOn`.

TaoChenOSU (approved):

adrianwyatt (approved):

PR #814, opened 2023-05-04
gRPC skills functionality

SergeyMenshykh:
### Motivation and Context

This PR adds MVP functionality for gRPC skills by introducing a set of classes for importing, parsing and executing gRPC operations defined in the .proto files. This functionality may be useful in scenarios when skills are represented as methods that are hosted on a different machine and need to be called remotely.

### Description
The PR adds the following classes:
- ProtoDocumentParser - iterates over a all methods of all services declared in .proto file and creates an instance of GrpcOperation model class for each of them.
- GrpcOperation - model class representing gRPC operation.
- GrpcOperationRunner - accepts an instance of the GrpcOperation class and a collection of arguments. Dynamically creates an instance of gRPC method request and calls the method. Receives, transforms and returns the method call result.
- GrpcOperationExtensions - extension methods for GrpcOperation to extend the class with SK SDK domain related functionality.
- KernelGrpcExtensions - declares a few register and import methods for importing the gRPC skills from a folder or a file and registering them in SK Kernel.

### Contribution Checklist


dotnet/Directory.Packages.props
```diff
@@ -10,26 +10,25 @@
     <PackageVersion Include="System.Linq.Async" Version="[6.0.1, )" />
     <PackageVersion Include="System.Text.Json" Version="[6.0.0, )" />
     <PackageVersion Include="Azure.AI.OpenAI" Version="[1.0.0-beta.5, )" />
-
+    <PackageVersion Include="Grpc.Net.Client" Version="2.52.0" />
```
lemillermicrosoft, line 13: Nit: Under the  section?

dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj
```diff
@@ -28,6 +28,7 @@
     <InternalsVisibleTo Include="Microsoft.SemanticKernel.Planning.ActionPlanner" PublicKey="$(StongNamePublicKey)"/>
     <InternalsVisibleTo Include="Microsoft.SemanticKernel.Planning.SequentialPlanner" PublicKey="$(StongNamePublicKey)"/>
     <InternalsVisibleTo Include="Microsoft.SemanticKernel.Skills.OpenAPI" PublicKey="$(StongNamePublicKey)"/>
+	<InternalsVisibleTo Include="Microsoft.SemanticKernel.Skills.Grpc" PublicKey="$(StongNamePublicKey)"/>
```
lemillermicrosoft, line 31: nit: spacing

dotnet/src/SemanticKernel/SemanticKernel.csproj
```diff
@@ -20,10 +20,11 @@ Install this package manually only if you are selecting individual Semantic Kern
   </PropertyGroup>
 
   <ItemGroup>
-    <InternalsVisibleTo Include="Microsoft.SemanticKernel.Planning.ActionPlanner" PublicKey="$(StongNamePublicKey)"/>
-    <InternalsVisibleTo Include="Microsoft.SemanticKernel.Planning.SequentialPlanner" PublicKey="$(StongNamePublicKey)"/>
+	<InternalsVisibleTo Include="Microsoft.SemanticKernel.Planning.ActionPlanner" PublicKey="$(StongNamePublicKey)"/>
```
lemillermicrosoft, line 23: nit: spacing

dotnet/src/Skills/Skills.Grpc/Extensions/GrpcOperationExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.Collections.Generic;
+using Microsoft.SemanticKernel.SkillDefinition;
+using Microsoft.SemanticKernel.Skills.Grpc.Model;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.SemanticKernel.Skills.Grpc.Extensions;
+
+/// <summary>
+/// Class for extensions methods for the <see cref="GrpcOperationExtensions"/> class.
```
lemillermicrosoft, line 11:
```suggestion
/// Class for extensions methods for the <see cref="GrpcOperation"/> class.
```

dotnet/src/Skills/Skills.Grpc/Extensions/KernelGrpcExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+            {
+                kernel.Log.LogWarning(ex, "Something went wrong while rendering the gRPC function. Function: {0}.{1}. Error: {2}", skillName, operation.Name,
+                    ex.Message);
+                context.Fail(ex.Message, ex);
+            }
+
+            return context;
+        }
+
+        // TODO: to be fixed later
```
lemillermicrosoft, line 173: can `using` here resolve this issue?

lemillermicrosoft (commented):
nitpick pass

dotnet/src/Skills/Skills.Grpc/Extensions/GrpcOperationExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+/// </summary>
+internal static class GrpcOperationExtensions
+{
+    /// <summary>
+    /// Returns list of gRPC operation parameters.
+    /// </summary>
+    /// <returns>The list of parameters.</returns>
+    public static IReadOnlyList<ParameterView> GetParameters(this GrpcOperation operation)
```
RogerBarreto, line 19: Any special reason for creating an extension method and not adding this to the model itself?

RogerBarreto: The method itself is called getParameters, but it is adding... ?

dotnet/src/Skills/Skills.Grpc/Extensions/KernelGrpcExtensions.cs
SergeyMenshykh: Unfortunately, it won't. If we put using here the instance of the SKFunction will be disposed as soon as execution control leaves this registration method way before the function got executed. It's not what we want.

At the moment, we don't need to do anything here because the instance will be registered in Kernel's skillsCollection which we plan to make disposable so that all functions registered in the collection will be disposed automatically when Kernel.Dispose method is called.

An alternative approach would be to move responsibility of managed resources creation out of SK SDK to a hosting application. As result, SK SDK would not need to dispose them because it did not create them and does not own them. That would simplify SK SDK code and would give more control over the resources lifetime to the hosting app which could choose to declare them as singletons and retain them for the entire app lifetime or have them as transient ones - create an instance per request.

dotnet/src/Skills/Skills.Grpc/Extensions/GrpcOperationExtensions.cs
SergeyMenshykh: > Any special reason for creating an extension method and not adding this to the model itself?

The idea here is to keep gRPC and SK SDK domains separately and not mix them. It usually simpler to maintain because code is less rigid and changes in one domain does not impacting the other one(s).

SergeyMenshykh: Yep, it's intentional. To execute a gRPC operation, the address, and payload parameters are required. However, GrpcOperation model class doesn't have them because they are dynamic ones and not described in the .proto files. So, as result we need to create them to be able to resolve them later, from SKContext, before the operation execution.

lemillermicrosoft (approved):

shawncal (approved):

dotnet/Directory.Packages.props
```diff
@@ -38,8 +34,10 @@
     <PackageVersion Include="Microsoft.Identity.Client.Extensions.Msal" Version="[2.28.0, )" />
     <PackageVersion Include="Microsoft.OpenApi" Version="[1.6.3, )" />
     <PackageVersion Include="Microsoft.OpenApi.Readers" Version="[1.6.3, )" />
-    <PackageVersion Include="Newtonsoft.Json" Version="[13.0.3, )" />
-
+    <PackageVersion Include="Newtonsoft.Json" Version="[11.0.1, )" />
```
dluc, line 37: @SergeyMenshykh ⚠️

dluc, line 37: ⚠️

PR #826, opened 2023-05-05
Add support for custom context in plan execution

lemillermicrosoft:
This commit fixes a bug that allows users to pass a custom SKContext object to a plan invocation, which can override or augment the default context variables. This enables more flexibility and control over the plan execution and the output. The commit also adds a unit test to verify the functionality of the new feature.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/Planning/Plan.cs
+++ b/dotnet/src/SemanticKernel/Planning/Plan.cs
@@ -508,7 +508,7 @@ public sealed class Plan : ISKFunction
         var functionParameters = step.Describe();
         foreach (var param in functionParameters.Parameters)
         {
-            if (param.Name.Equals("INPUT", StringComparison.OrdinalIgnoreCase))
+            if (param.Name.Equals(ContextVariables.MainKey, StringComparison.OrdinalIgnoreCase))
             {
                 continue;
             }
--- a/dotnet/src/SemanticKernel/Planning/Plan.cs
+++ b/dotnet/src/SemanticKernel/Planning/Plan.cs
@@ -470,21 +470,49 @@ public sealed class Plan : ISKFunction
     /// <returns>The context variables for the next step in the plan.</returns>
     private ContextVariables GetNextStepVariables(ContextVariables variables, Plan step)
     {
-        // If the current step is passing to another plan, we set the default input to an empty string.
-        // Otherwise, we use the description from the current plan as the default input.
-        // We then set the input to the value from the SKContext, or the input from the Plan.State, or the default input.
-        var defaultInput = step.Steps.Count > 0 ? string.Empty : this.Description ?? string.Empty;
-        var planInput = string.IsNullOrEmpty(variables.Input) ? this.State.Input : variables.Input;
-        var stepInput = string.IsNullOrEmpty(planInput) ? defaultInput : planInput;
-        var stepVariables = new ContextVariables(stepInput);
+        // Priority for Input
+        // - Parameters (expand from variables if needed)
+        // - SKContext.Variables
+        // - Plan.State
+        // - Empty if sending to another plan
+        // - Plan.Description
+
+        var input = string.Empty;
+        if (!string.IsNullOrEmpty(step.Parameters.Input))
+        {
+            input = this.ExpandFromVariables(variables, step.Parameters.Input);
+        }
+        else if (!string.IsNullOrEmpty(variables.Input))
+        {
+            input = variables.Input;
+        }
+        else if (!string.IsNullOrEmpty(this.State.Input))
+        {
+            input = this.State.Input;
+        }
+        else if (step.Steps.Count > 0)
+        {
+            input = string.Empty;
+        }
+        else if (!string.IsNullOrEmpty(this.Description))
+        {
+            input = this.Description;
+        }
+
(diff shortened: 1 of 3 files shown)


dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -497,6 +497,12 @@ private ContextVariables GetNextStepVariables(ContextVariables variables, Plan s
 
         foreach (var item in step.Parameters)
         {
+            // Don't overwrite variable values that are already set except for INPUT
+            if (!item.Key.Equals("INPUT", StringComparison.OrdinalIgnoreCase) && stepVariables.Get(item.Key, out var val))
```
RogerBarreto, line 501: Suggest making `ContextVariable.MainKey` public and use it across the solution...


[SKFunction.cs](https://github.com/microsoft/semantic-kernel/blob/f1448c9a0759947256af7f4c6b940c9eab06ea91/dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs#L552)


[SKFunctionInputAttribute](https://github.com/microsoft/semantic-kernel/blob/f1448c9a0759947256af7f4c6b940c9eab06ea91/dotnet/src/SemanticKernel.Abstractions/SkillDefinition/SKFunctionInputAttribute.cs#L51)

lemillermicrosoft: Will update contextvariables and plan for now, want to keep this scoped.

lemillermicrosoft: Especially since the casing is different for reasons.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -470,21 +470,49 @@ private SKContext UpdateContextWithOutputs(SKContext context)
...
+        // - Step Parameters (pull from variables or state by a key value)
         var functionParameters = step.Describe();
         foreach (var param in functionParameters.Parameters)
         {
+            if (param.Name.Equals("INPUT", StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
+            }
+
             if (variables.Get(param.Name, out var value) && !string.IsNullOrEmpty(value))
```
RogerBarreto, line 515:
```suggestion
if (!string.IsNullOrEmpty(value) && 
    (variables.Get(param.Name, out var value) 
    || this.State.Get(param.Name, out value)))
{
    stepVariables.Set(param.Name, value);
}

```

RogerBarreto: Same for the others... check for empty/null first.

lemillermicrosoft: `Get` could technically return an empty string -- I don't think it should then be used.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -510,6 +543,10 @@ private ContextVariables GetNextStepVariables(ContextVariables variables, Plan s
             {
                 stepVariables.Set(item.Key, value);
             }
+            else if (!string.IsNullOrEmpty(expandedValue) && !stepVariables.Get(item.Key, out var _))
```
shawncal, line 546: the check on 528 should take care of this part
  !stepVariables.Get(item.Key, out var _)

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -510,6 +543,10 @@ private ContextVariables GetNextStepVariables(ContextVariables variables, Plan s
...
+            {
+                stepVariables.Set(item.Key, expandedValue);
```
shawncal, line 548: ContextVariables.Set will check for null, and remove the arg if it's null. You may not need this last conditional (i.e. it could just be "else") if you're okay letting it remove the entry at item.Key.

dotnet/src/SemanticKernel/Planning/Plan.cs
lemillermicrosoft: Okay, I'm going back on this, thanks. If it's an empty string, something did that on purpose, let's respect it.

dotnet/src/SemanticKernel/Planning/Plan.cs
lemillermicrosoft: I was worried about `string.Empty` values, but I'm letting go of that for now. If someone sets something to an empty string, we'll treat it as valid.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -470,46 +470,83 @@ private SKContext UpdateContextWithOutputs(SKContext context)
...
             {
                 stepVariables.Set(param.Name, value);
             }
         }
 
         foreach (var item in step.Parameters)
         {
-            if (!string.IsNullOrEmpty(item.Value))
+            // Don't overwrite variable values that are already set
+            if (stepVariables.Get(item.Key, out var val))
```
shawncal, line 528: Super nitpicky chage: discarding value so it doesn't accidentally get used for something below

```suggestion
            if (stepVariables.Get(item.Key, out _))
```

shawncal (approved):

PR #829, opened 2023-05-05
IChatCompletion Streaming Support

### LINKED ISSUE #378: Streaming support, implementation of IAsyncEnumberalbe
Because I see that Example App Chat Copilot's response  is too slow, I would like to ask if Streaming will be implemented in the future.

Additional description:
Nugetbao package Azure.AI.OpenAI supports Streaming
Example:
 ``` cs
Response<StreamingChatCompletions> response = await client.GetChatCompletionsStreamingAsync(
                deploymentOrModelName: model,
                chatCompletionsOptions);
using StreamingChatCompletions streamingChatCompletions = response.Value;

await foreach (StreamingChatChoice choice in streamingChatCompletions.GetChoicesStreaming())
{
    await foreach (ChatMessage message in choice.GetMessageStreaming())
    {
        Console.Write(message.Content);
    }
    Console.WriteLine("-------------");
}
```

### PR DESCRIPTION
**Pull request by RogerBarreto**
### Motivation and Context

Get streaming results from chat completion models using the IAsyncEnumerable pattern.

Closes #378 
Resolves #378 

Added samples how to use those as well as how to implement your own custom chat completion model using streaming capability.

### Description

ChatCompletionStream

Co-authored-by: dedalo <dedalo@gmail.com>

### Contribution Checklist

### CODE REVIEW

**Comment by dedalo** [comment]
This works great and helps implement signalr in chat applications wich improves user experience. Looging forward for the merge.

File: samples/dotnet/kernel-syntax-examples/Example17_ChatGPT.cs [hunk 1]
```diff
@@ -10,71 +11,183 @@
 /**
  * The following example shows how to use Semantic Kernel with OpenAI ChatGPT API
  */
-
 // ReSharper disable once InconsistentNaming
-// ReSharper disable CommentTypo
 public static class Example17_ChatGPT
```
dluc, line 15: in the spirit of having basic syntax examples I would move all the streaming examples in a separate example dedicated to streaming

samples/dotnet/kernel-syntax-examples/Example33_CustomChat.cs
```diff
@@ -0,0 +1,171 @@
...
+        foreach (string word in streamedOutput)
+        {
+            await Task.Delay(200, cancellationToken);
+            yield return $"{word} ";
+        }
+    }
+}
+
+// ReSharper disable once InconsistentNaming
+public static class Example33_CustomChat
```
dluc, line 60: what about a more descriptive name?

samples/dotnet/kernel-syntax-examples/Example17_ChatGPT.cs
RogerBarreto: @dluc 
When I'm a end user the first thing I would look is for ChatGPT example and then I would see how many different ways I could do it. That's why I put the streaming in here... moving this to streaming in the end of the line will mostly hide it and enforce the End user to know that there is this `StreamingChat` that do the same but 20 examples further... 

I believe we can actually improve how we present our samples, is getting big.

samples/dotnet/kernel-syntax-examples/Example33_CustomChat.cs
RogerBarreto: @dluc I'm using here almost the same name you used in the `CustomLLM` maybe I could put all those examples in the CustomLLM or rename it as `CustomLLMChat` and `CustomLLM` -> `CustomLLMCompletion`

samples/dotnet/kernel-syntax-examples/Example17_ChatGPT.cs
shawncal: Yeah, I'd rather condense the syntax examples into logical domains, based on likely user lookups. They're becoming too numerous and, with no particular order or clustering, it's becoming difficult to find what you need.

samples/dotnet/kernel-syntax-examples/Example33_CustomChat.cs
shawncal: I think the reasoning is that there could be confusion between bringing your own chat model (your example) and implementing a custom chat solution (copilot chat). I'd agree with either merging into CustomLLM, or renaming to something like CustomChatModel

shawncal (approved):

Issue #791, opened 2023-05-03
Why does SkillCollection differentiate semantic vs native in its APIs?


stephentoub:
SkillCollection duplicates its surface area, once for semantic functions and once for native functions.  Why does SkillCollection care about the distinction at all?  It ends up storing them all in the same dictionary, so it already doesn't support having two different functions with the same name, and in any scenario where a consumer cares to only get a semantic vs native function, they can just check the resulting function's IsSemantic (it's also not clear to me in what situation a consumer would actually care).

cc: @shawncal

evchaki:
@stephentoub thanks for bringing this up, we will take a look

dluc:
It's part of the journey from internal prototypes written a year ago when we first started using "AI functions". We're going through another set of design polish tasks and this was one of the TODO, there's a few PRs coming and a couple in particular touching this area.

PR #848, opened 2023-05-07
Simplify orchestration API

dluc:
### Motivation and Context

Complete some early work from Feb, simplifying the orchestration API, removing redundant code in the function registry.

### Description

Merge registry methods used for native and semantic functions, given that the underlying implementation was refactored and the distinction is not required anymore.


dotnet/src/SemanticKernel/Orchestration/SKContextExtensions.cs
```diff
@@ -15,13 +15,20 @@ internal static class SKContextExtensions
     /// <param name="skillName">The skill name</param>
     /// <param name="functionName">The function name</param>
     /// <param name="registeredFunction">The registered function, if found</param>
-    internal static bool IsFunctionRegistered(this SKContext context, string skillName, string functionName, [NotNullWhen(true)] out ISKFunction? registeredFunction)
+    internal static bool IsFunctionRegistered(this SKContext context,
+        string skillName,
```
stephentoub, line 19: Nit:
```suggestion
        string? skillName,
```
since skillName is allowed to be null per the below checks.

dotnet/src/SemanticKernel/Orchestration/SKContextExtensions.cs
```diff
@@ -15,13 +15,20 @@ internal static class SKContextExtensions
...
-        return context.Skills!.TryGetNativeFunction(skillName, functionName, out registeredFunction) ||
-               context.Skills.TryGetNativeFunction(functionName, out registeredFunction) ||
-               context.Skills.TryGetSemanticFunction(skillName, functionName, out registeredFunction);
+        if (string.IsNullOrEmpty(skillName))
+        {
+            return context.Skills!.TryGetFunction(functionName, out registeredFunction);
+        }
+
+        return context.Skills!.TryGetFunction(skillName, functionName, out registeredFunction) ||
+               context.Skills.TryGetFunction(functionName, out registeredFunction);
```
stephentoub, line 31:
```suggestion
        if (!string.IsNullOrEmpty(skillName) &&
            context.Skills!.TryGetFunction(skillName, functionName, out registeredFunction))
        {
            return true;
        }

        return context.Skills!.TryGetFunction(functionName, out registeredFunction);
```

stephentoub:
Looks like this addresses https://github.com/microsoft/semantic-kernel/issues/791. Thanks.

dotnet/src/SemanticKernel/Orchestration/SKContextExtensions.cs
```diff
@@ -15,13 +15,19 @@ internal static class SKContextExtensions
     /// <param name="skillName">The skill name</param>
     /// <param name="functionName">The function name</param>
     /// <param name="registeredFunction">The registered function, if found</param>
-    internal static bool IsFunctionRegistered(this SKContext context, string skillName, string functionName, [NotNullWhen(true)] out ISKFunction? registeredFunction)
+    internal static bool IsFunctionRegistered(this SKContext context,
+        string? skillName,
+        string functionName,
+        [NotNullWhen(true)] out ISKFunction? registeredFunction)
     {
```
shawncal, line 22: See comments in https://github.com/microsoft/semantic-kernel/pull/849
 - Recommend removing the word "registered" from method and args, since the Register* methods are all now Import* or AddFunction
 - Rename this method to TryGetFunction for consistency with dotnet and SK's other similar methods

dluc: what about renaming both to `IsFunctionAvailable`? the name would provide a more fluent API

stephentoub: > what about renaming both to IsFunctionAvailable? the name would provide a more fluent API

What do you do after such a method returns true?  Isn't the 99% case to then get the function, in which case the simpler and cheaper thing is to just do it all in a single lookup.  If you don't care about the resulting object, the usability downside to `TryGetFunction` is just an additional `, out _` at the call site, but if you do need the object and you have to make a separate call to get it, it means a much larger bit of code needs to be tacked on.

If we're trying to make this feel like a collection of skills, then it should mirror the shape of collections in .NET.  None of them have "IsAvaialable" APIs.  If there are common cases where the queried for function isn't needed and you only need to know whether it's contained or not, we could also expose "ContainsFunction" in addition to "TryGetFunction".

shawncal: Yeah, I agree... I believe, if it returns bool and has out var, then it's a TryGet by convention/expectation. If it just returns a bool (no out var), it can be a Has* or Is*. We could have both methods, but the TryGet* is a superset, and is the better one to have as it's a single lookup for both has+get.

RogerBarreto: IMHO I would have two methods 

1. `GetFunctionOrDefault` or just `GetFunction` (not throwing):
2. `TryGetFunction`

```
bool TryGetFunction (...params, out ISKFunction? x) 
{
    x = GetOrDefault(...params);
    return x is null;
}

ISKFunction? GetOrDefault(...params) 
{ 
    ... 
}
```

dluc: each language has a specific approach, looking at options across multiple languages

C#: bool TryGet(key, out value)
python: T get(key, default) // no bool
Java: T get(key) // null if not exists
TS:  [key] // undefined is not exists

dluc:
ready to go

shawncal (approved):

PR #849, opened 2023-05-07
Simplify IKernel interface

### LINKED ISSUE #726: kernel.CreateSemanticFunction functionName is marked optional, but is not optional
**Describe the bug**
If using the overload for kernel.CreateSemanticFunction that takes a PromptTemplateConfig, the functionName is marked optional, but if you execute the code it will throw an exception that functionName is not optional.

If you use the overload that takes just a promptTemplate (everything else optional), then you omit the functionName and when executing it will work, despite not having a functionName defined.

@shawncal

dluc:
### Motivation and Context

* Move semantic methods out of IKernel, since they can be delivered using more generic code.
* Fix #726

### Description

* `RegisterCustomFunction` -> `ImportFunction`, consistent with `ImportSkill`
* `RegisterSemanticFunction` -> `ImportSemanticFunction` kernel extension

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/IKernel.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/IKernel.cs
@@ -6,7 +6,6 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.SemanticKernel.Memory;
 using Microsoft.SemanticKernel.Orchestration;
-using Microsoft.SemanticKernel.SemanticFunctions;
 using Microsoft.SemanticKernel.SkillDefinition;
 using Microsoft.SemanticKernel.TemplateEngine;
 
@@ -43,26 +42,13 @@ public interface IKernel
     IReadOnlySkillCollection Skills { get; }
 
     /// <summary>
-    /// Build and register a function in the internal skill collection, in a global generic skill.
+    /// Import a set of functions from the given skill. The functions must have the `SKFunction` attribute.
+    /// Once these functions are imported, the prompt templates can use functions to import content at runtime.
     /// </summary>
-    /// <param name="functionName">Name of the semantic function. The name can contain only alphanumeric chars + underscore.</param>
-    /// <param name="functionConfig">Function configuration, e.g. I/O params, AI settings, localization details, etc.</param>
-    /// <returns>A C# function wrapping AI logic, usually defined with natural language</returns>
-    ISKFunction RegisterSemanticFunction(
-        string functionName,
-        SemanticFunctionConfig functionConfig);
-
-    /// <summary>
-    /// Build and register a function in the internal skill collection.
-    /// </summary>
-    /// <param name="skillName">Name of the skill containing the function. The name can contain only alphanumeric chars + underscore.</param>
-    /// <param name="functionName">Name of the semantic function. The name can contain only alphanumeric chars + underscore.</param>
-    /// <param name="functionConfig">Function configuration, e.g. I/O params, AI settings, localization details, etc.</param>
-    /// <returns>A C# function wrapping AI logic, usually defined with natural language</returns>
-    ISKFunction RegisterSemanticFunction(
-        string skillName,
-        string functionName,
-        SemanticFunctionConfig functionConfig);
+    /// <param name="skillInstance">Instance of a class containing functions</param>
+    /// <param name="skillName">Name of the skill for skill collection and prompt templates. If the value is empty functions are registered in the global namespace.</param>
+    /// <returns>A list of all the semantic functions found in the directory, indexed by function name.</returns>
+    IDictionary<string, ISKFunction> ImportSkill(object skillInstance, string? skillName = null);
 
     /// <summary>
     /// Registers a custom function in the internal skill collection.
@@ -70,16 +56,7 @@ public interface IKernel
     /// <param name="skillName">Name of the skill containing the function. The name can contain only alphanumeric chars + underscore.</param>
     /// <param name="customFunction">The custom function to register.</param>
     /// <returns>A C# function wrapping the function execution logic.</returns>
-    ISKFunction RegisterCustomFunction(string skillName, ISKFunction customFunction);
-
-    /// <summary>
-    /// Import a set of functions from the given skill. The functions must have the `SKFunction` attribute.
-    /// Once these functions are imported, the prompt templates can use functions to import content at runtime.
-    /// </summary>
-    /// <param name="skillInstance">Instance of a class containing functions</param>
-    /// <param name="skillName">Name of the skill for skill collection and prompt templates. If the value is empty functions are registered in the global namespace.</param>
-    /// <returns>A list of all the semantic functions found in the directory, indexed by function name.</returns>
-    IDictionary<string, ISKFunction> ImportSkill(object skillInstance, string skillName = "");
+    ISKFunction ImportFunction(string skillName, ISKFunction customFunction);
 
     /// <summary>
     /// Set the semantic memory to use
--- a/dotnet/src/SemanticKernel/Kernel.cs
+++ b/dotnet/src/SemanticKernel/Kernel.cs
@@ -6,7 +6,6 @@ using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
-using Microsoft.SemanticKernel.AI;
 using Microsoft.SemanticKernel.AI.ChatCompletion;
 using Microsoft.SemanticKernel.AI.Embeddings;
 using Microsoft.SemanticKernel.AI.ImageGeneration;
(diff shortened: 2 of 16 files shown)


dluc:
to be rebased and merged after https://github.com/microsoft/semantic-kernel/pull/848

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -43,43 +42,21 @@ public interface IKernel
...
-    /// <param name="skillName">Name of the skill containing the function. The name can contain only alphanumeric chars + underscore.</param>
-    /// <param name="functionName">Name of the semantic function. The name can contain only alphanumeric chars + underscore.</param>
-    /// <param name="functionConfig">Function configuration, e.g. I/O params, AI settings, localization details, etc.</param>
-    /// <returns>A C# function wrapping AI logic, usually defined with natural language</returns>
-    ISKFunction RegisterSemanticFunction(
-        string skillName,
-        string functionName,
-        SemanticFunctionConfig functionConfig);
+    /// <param name="skillInstance">Instance of a class containing functions</param>
+    /// <param name="skillName">Name of the skill for skill collection and prompt templates. If the value is empty functions are registered in the global namespace.</param>
```
stephentoub, line 49: "If the value is empty"... why not if it's null?  Wondering why we have `string skillName = ""` instead of `string? skillName = null`.

dluc: I don't think there's a strong reason. We've been back and forth with nullability checks about twenty times (not kidding) because we keep switching language version/.net version. It's kind of annoying in fact

shawncal: We've had a mix of empty and null default string args throughout the codebase. I've been giving guidance in PRs to use "string? arg = null", as this is a stronger indication that the string was not provided at all, rather than a string was provided that happens to be empty (and, in some cases, is valid).

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -43,43 +42,21 @@ public interface IKernel
...
+    /// <returns>A list of all the semantic functions found in the directory, indexed by function name.</returns>
```
shawncal, line 50: "A list of all the semantic functions"
Is this true (only semantic)?

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -43,43 +42,21 @@ public interface IKernel
...
-
-    /// <summary>
-    /// Import a set of functions from the given skill. The functions must have the `SKFunction` attribute.
-    /// Once these functions are imported, the prompt templates can use functions to import content at runtime.
-    /// </summary>
-    /// <param name="skillInstance">Instance of a class containing functions</param>
-    /// <param name="skillName">Name of the skill for skill collection and prompt templates. If the value is empty functions are registered in the global namespace.</param>
-    /// <returns>A list of all the semantic functions found in the directory, indexed by function name.</returns>
-    IDictionary<string, ISKFunction> ImportSkill(object skillInstance, string skillName = "");
+    ISKFunction ImportFunction(string skillName, ISKFunction customFunction);
```
shawncal, line 59: Recommend removing the "custom" in the arg name and method comments, if it's removed from the method name,

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -112,21 +91,21 @@ public ISKFunction RegisterSemanticFunction(string skillName, string functionNam
...
-            this._skillCollection.AddNativeFunction(f.Value);
+            this._skillCollection.AddFunction(f.Value);
         }
 
         return skill;
     }
 
     /// <inheritdoc/>
-    public ISKFunction RegisterCustomFunction(string skillName, ISKFunction customFunction)
+    public ISKFunction ImportFunction(string skillName, ISKFunction customFunction)
```
shawncal, line 101: same comment: recommend remove "custom"

dotnet/src/SemanticKernel/Orchestration/SKContextExtensions.cs
```diff
@@ -15,13 +15,20 @@ internal static class SKContextExtensions
     /// <param name="skillName">The skill name</param>
     /// <param name="functionName">The function name</param>
     /// <param name="registeredFunction">The registered function, if found</param>
-    internal static bool IsFunctionRegistered(this SKContext context, string skillName, string functionName, [NotNullWhen(true)] out ISKFunction? registeredFunction)
+    internal static bool IsFunctionRegistered(this SKContext context,
```
shawncal, line 18: This returns an object in the out param and bool=true if found...
Can we rename to TryGetFunction, to align with ,net and similar calls throughout the Kernel?

Consider also the use of "registered" in the args and method comments... if we're removing the "Register*" naming in favor of "Import", it might be a good idea to replace it here too for consistency.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -76,25 +74,6 @@ public sealed class Kernel : IKernel, IDisposable
...
-        Verify.ValidFunctionName(functionName);
-
-        ISKFunction function = this.CreateSemanticFunction(skillName, functionName, functionConfig);
-        this._skillCollection.AddSemanticFunction(function);
-
-        return function;
-    }
-
     /// <inheritdoc/>
     public IDictionary<string, ISKFunction> ImportSkill(object skillInstance, string skillName = "")
```
SergeyMenshykh, line 78: Next step would be to move remaining skills related methods - ImportSkill, ImportFunction to the place they naturally belong to SkillsCollection class since they have nothing to do with Kernel code – both, despite declared in Kernel class, just register functions in the Kernel skills collection. It'll simplify Kernel code a little bit.

After that, we can think about splitting SkillsCollection and group all code we have today for working with semantic, native, OpenAPI, gRPC skills, spread across bunch of Kernel extension methods today, to corresponding skill sets – SemanticSkillSet, NativeSkillSet, OpenApiSkillSet, and GrpcSkillSet. The grouping will allow to have better separation of concern since all logic related to particular type of skills will live in the skill set class or it’s extension methods and not in one common SkillsCollection one or Kernel extension methods.

dotnet/src/SemanticKernel/Kernel.cs
RogerBarreto, line 101: Not sure about the terms `Import` / `Add` we clearly are using different ones for ISkillCollection and IKernel. IMO would use one term for all I like `Add`.

dotnet/src/SemanticKernel/Orchestration/SKContextExtensions.cs
RogerBarreto: Or `Added` 
We have three different terms `Import` `Register` `Add`, would be good to stick to one.

dotnet/src/SemanticKernel/SemanticFunctions/KernelSemanticFunctionsExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        return function;
+    }
+
+    private static ISKFunction CreateSemanticFunction(IKernel kernel,
+        string skillName,
+        string functionName,
+        SemanticFunctionConfig functionConfig)
+    {
+        if (!functionConfig.PromptTemplateConfig.Type.Equals("completion", StringComparison.OrdinalIgnoreCase))
```
RogerBarreto, line 59: Should we use a `const` `Enum` or `nameof` instead of `"completion"` literal?

dotnet/src/SemanticKernel.UnitTests/Planning/PlanSerializationTests.cs
```diff
@@ -435,15 +435,16 @@ public async Task CanStepAndSerializeAndDeserializePlanWithStepsAndContextAsync(
             .Returns(() => Task.FromResult(returnContext));
         mockFunction.Setup(x => x.Describe()).Returns(new FunctionView()
```
RogerBarreto, line 436:
```suggestion
        mockFunction.Setup(x => x.Describe()).Returns(new FunctionView
```

RogerBarreto (changes requested):

dotnet/src/SemanticKernel/Orchestration/SKContextExtensions.cs
dluc: `Register` can go away probably. `Add` is for implementation, `Import` is for user space.

dotnet/src/SemanticKernel/Kernel.cs
dluc: see other comment

dotnet/src/SemanticKernel/SemanticFunctions/KernelSemanticFunctionsExtensions.cs
dluc: one of many TODOs, for another PR

dotnet/src/SemanticKernel.UnitTests/Planning/PlanSerializationTests.cs
dluc: could you suggest that to the original author?

PR #856, opened 2023-05-08
Disposing skills

SergeyMenshykh:
### Motivation and Context
This change is required to fix a memory leak related to not disposed resource used by Skills. The issue may manifest itself any time skills that work with managed and unmanaged resources like files, network, etc... are registered in Kernel and the Kernel is disposed but the Skills aren't.

### Description
This PR adds functionality that disposes all Skills registered in Kernel when Kernel.Dispose method is called.

### Contribution Checklist


dotnet/src/SemanticKernel/SkillDefinition/SkillCollection.cs
```diff
@@ -161,6 +161,20 @@ public FunctionsView GetFunctionsView(bool includeSemantic = true, bool includeN
         return result;
     }
 
+    /// <summary>
+    /// Dispose of skills.
+    /// </summary>
+    public void Dispose()
+    {
+        foreach (var pair in this._skillCollection)
```
shawncal, line 169: I'd just enumerate on this._skillCollection.Values

shawncal (changes requested):
Note new build warnings in [Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SKContextExtensionsTests.cs](https://github.com/microsoft/semantic-kernel/pull/856/files#diff-353564ee35d30258e1ffd064120b1c1c284734238de9432c61dc0a51198fe085)

dotnet/src/SemanticKernel/SkillDefinition/SkillCollection.cs
SergeyMenshykh: Thanks, it allowed me to uncover a potential bug. Fixed and covered by unit test.

dotnet/src/SemanticKernel.UnitTests/SkillDefinition/SkillCollectionTests.cs
```diff
@@ -43,4 +44,23 @@ public void ItAllowsToReplaceFunctions()
...
+    public void ItShouldDisposeSkillsFunctions()
+    {
+        // Arrange
+        var disposableFunction = new Mock<ISKFunction>();
+        disposableFunction.SetupGet(f => f.Name).Returns("fake-function-name");
+        disposableFunction.SetupGet(f => f.SkillName).Returns("fake-skill-name");
+        disposableFunction.As<IDisposable>();
+
+        var skillsCollection = new SkillCollection();
+        skillsCollection.AddFunction(disposableFunction.Object);
```
lemillermicrosoft, line 58: What happens if you add it to a different skillsCollection, too?

SergeyMenshykh: Good question Lee. The function will be registered in two collections, and it will become invalid as soon as one of the collections is disposed.

The skills collections are made disposable because, as far as I understand, they can contain functions that use system (managed or unmanaged) resources (such as HttpClient) that need to be released when the functions are no longer used.

I don't think that skills collections should be disposable because they may share the same functions, and disposing one collection will invalidate those functions. It also feels wrong to have a collection be disposable - it's like making a List<HttpClient> disposable just because it contains instances of HttpClient.

The functions should not be disposable either, because two or more functions can use the same resource, and if one function is disposed, then the resource is invalidated and can no longer be used by the other function.

I believe the correct solution to this problem is to delegate the creation of system resources to a hosting application and make it responsible for disposing them since it created and owns them. 

@shawncal @stephentoub any thoughts?

stephentoub: > I believe the correct solution to this problem is to delegate the creation of system resources to a hosting application and make it responsible for disposing them since it created and owns them.

Does SK make it possible for a hosting application to know when all resources are completed?

Also how would a hosting application get at all of the resources?

SergeyMenshykh:
The solution for the issue is going to be completely different. Closing this PR for now.

PR #859, opened 2023-05-08
CopilotChat: Add Azure Cog Search for Semantic Memory

adrianwyatt:
### Motivation and Context
Enable integration of Azure Cognitive Search as a semantic memory source.

### Description
- Added "AzureCognitiveSource" option for "MemoriesStore"
- Refactored DI to provide full SemanticTextMemory instead of just MemoryStore since the Azure Cognitive Search connector does not have support for vectors and therefor is implemented at the SemanticTextMemory layer.


glahaye (approved):

gitri-ms:
Does the SemanticMemory  example in kernel syntax examples need to be updated at all due to this change?

gitri-ms:
Does the Semantic Memory example need to be updated at all due to this change? (/samples/dotnet/kernel-syntax-examples/Example14_SemanticMemory.cs)

adrianwyatt:
> Does the Semantic Memory example need to be updated at all due to this change? (/samples/dotnet/kernel-syntax-examples/Example14_SemanticMemory.cs)

It does not - we are using the already provided ACS connector.

PR #865, opened 2023-05-08
CopilotChat: Add Azure Cog Search for Semantic Memory w/IOptions fix

adrianwyatt:
### Motivation and Context
Re-commit reverted ACS change with fix for IOptions issue

### Description
- Switch ISemanticTextMemory from Singleton to Scoped to allow for proper IOptions loading.


adrianwyatt:
- dupe

PR #867, opened 2023-05-08
CopilotChat: Add Klarna to extensions list

adrianwyatt:
### Motivation and Context
CopilotChat's planner always had Klarna enabled - this allows the user to opt-in.

### Description
- Changed semantics of extensions from Connect/Disconnect to Enable/Disable
- Added Klarna tile to UX 
- Klarna will be enabled when the auth header exists with any value in it

### Screenshots
image
image


samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -144,19 +144,21 @@ public class SemanticKernelController : ControllerBase, IDisposable
...
+        // Klarna Shopping
+        if (openApiSkillsAuthHeaders.KlarnaAuthentication != null)
+        {
+            // Register the Klarna shopping ChatGPT plugin with the planner's kernel.
+            using DefaultHttpRetryHandler retryHandler = new(new HttpRetryConfig(), this._logger)
+            {
+                InnerHandler = new HttpClientHandler() { CheckCertificateRevocationList = true }
+            };
+            using HttpClient importHttpClient = new HttpClient(retryHandler, false);
+            importHttpClient.DefaultRequestHeaders.Add("User-Agent", "Microsoft.CopilotChat");
```
teresaqhoang, line 158: Why do we want to override this for Klarna specifically?

adrianwyatt: We should be providing a proper HttpClient everywhere, but I'm starting here.

adrianwyatt:
Tao is out today and we'd like to get this in today - recreating PR to reset review.

PR #870, opened 2023-05-09
Copilot chat: join chats via chat ids

TaoChenOSU:
### Motivation and Context

The first step to enable a multi-user chat experience is to have a mechanism to allow users to join chat sessions.
With this PR, users will be able to join chat sessions via chat ids shared by others.

### Description

1. A new model called ChatParticipant to manage users in chat sessions.
2. A new ChatParticipantController to provide a service API to join chat sessions via chat ids.
3. Remove user id in the ChatSession model since all users will be managed by ChatParticipant.
4. Front end to retrieve chat ids for sharing.
5. Front end to join chat sessions via chat ids.
image
image
image
image

Note that we still need to have a real-time message relay system to have a complete multi-user chat experience.

### Contribution Checklist


samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
```diff
@@ -95,18 +106,28 @@ public async Task<IActionResult> GetChatSessionByIdAsync(Guid chatId)
...
     [HttpGet]
     [Route("chatSession/getAllChats/{userId:regex(([[a-z0-9]]+-)+[[a-z0-9]]+\\.([[a-z0-9]]+-)+[[a-z0-9]]+)}")]
     [ProducesResponseType(StatusCodes.Status200OK)]
-    [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
     {
-        var chats = await this._chatSessionRepository.FindByUserIdAsync(userId);
-        if (chats == null)
+        var chatParticipants = await this._chatParticipantRepository.FindByUserIdAsync(userId);
```
Vivihung, line 116: I might be biased as how I usually do RBAC with objects (e.g. filter chat sessions that aren't associated with the user). The logic here looks different (different doesn't mean anything bad), could you explain more with comments? Why are we getting a list of users (participants) from a given user id, and find chats sessions per users (participants)?

samples/apps/copilot-chat-app/webapi/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,75 @@
...
+        this._chatParticipantRepository = chatParticipantRepository;
+        this._chatSessionRepository = chatSessionRepository;
+    }
+
+    /// <summary>
+    /// Join a use to a chat session given a chat id and a user id.
+    /// </summary>
+    /// <param name="chatParticipantParam">Contains the user id and chat id.</param>
+    [HttpPost]
+    [Route("chatParticipant/join")]
```
Vivihung, line 46: (nit) At some point, we should think about our api route patterns. For example, I would use "chat/{chatid}/userJoin" as the route. my early thoughts: "{semantic_object}/{id}/{action}".
Please don't be blocked by this comment. (cc @adrianwyatt )

Vivihung (changes requested):
Thanks for the PR. Would like to see some comments in ChatParticipantController.cs.

samples/apps/copilot-chat-app/webapi/Controllers/ChatHistoryController.cs
TaoChenOSU: This is a good point. 
I guess the name `ChatParticipant` causes confusion here. 

User != Participant.

`ChatParticipant` contains the information of a participant in a particular chat session. One user can be multiple `ChatParticipant`s in different chat sessions.

To get all the chat sessions a user has access to, we need to first find all the `ChatParticipant`s that belongs to the user.

Vivihung: got it. recommend adding inline comments to explain might help others understand the thoughts.

Vivihung (dismissed):
thx!

amsacha (approved):

samples/apps/copilot-chat-app/webapi/Controllers/ChatParticipantController.cs
TaoChenOSU: Agreed!

hathind-ms (dismissed):

samples/apps/copilot-chat-app/webapi/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        catch (Exception ex) when (ex is KeyNotFoundException || ex is ArgumentOutOfRangeException)
+        {
+            return this.BadRequest("Chat session does not exist.");
+        }
+
+        // Make sure the user is not already in the chat session.
+        if (await this._chatParticipantRepository.IsUserInChatAsync(userId, chatId))
+        {
+            return this.BadRequest("User is already in the chat session.");
```
adrianwyatt, line 67: I wouldn't consider the user being already in the chat session as a failure. Maybe return "OK" here.

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,75 @@
...
+/// <summary>
+/// Controller for chat invitations.
+/// This controller is responsible for:
+/// 1. Creating invitation links.
+/// 2. Accepting/rejecting invitation links.
+/// 3. Managing participants in a chat session.
+/// </summary>
+[ApiController]
+[Authorize]
+public class ChatParticipantController : ControllerBase
```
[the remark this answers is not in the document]
TaoChenOSU: It used to be `ChatInvitation`. We decided to use `ChatParticipant` instead since it's not really an invitation system. It doesn't have an invitation object that it accepts when the APIs is invoked. Instead, it creates a new participant.

samples/apps/copilot-chat-app/webapi/Controllers/ChatParticipantController.cs
TaoChenOSU: In that case, the front end will not show anything to indicate the user is already in the session. I think it's better to at least tell the users that they are already in the session.

adrianwyatt (approved):

PR #882, opened 2023-05-09
InternalsVisibleTo Reduction & Shared Common Utilities - Part 1

shawncal:
### Motivation and Context
In preparation for strong name signing, we must reduce/remove the extensive usage of **InternalsVisibleTo**, leaking internals between packages.

_Note: use of InternalsVisibleTo to expose certain types to specific unit test projects is acceptable and common practice. We should not be exposing Kernel internals to connector/skill projects._

### Description
Most of the internals usage between packages is for simple attributes and Verify() methods -- things that we don't want as part of our public API surface, but that we also don't want to copy/paste between projects and maintain multiple copies.

With this change, the common files are in an InternalUtilities folder, and are imported as source files into projects that use them.  Each project therefore will build these into its dll, but they will be kept as internals, but there remains only one copy of the code for maintenance.

Note: an alternative we can consider would be to build these into a nuget package (all types public) and import it into each project with visibility=private so that it doesn't show up as part of the SK API surface unless you import this package directly. This would mean that there is only one such lib, but also introduces yet another nuget. This PR does not preclude taking that approach down the road, should this version prove problematic.

The only types made public to support this change are the RestApi* classes. I think it's harmless for these to be public for now, though we may consider putting these (back) into a dedicated project with private imports.

### Contribution Checklist


shawncal:
@SergeyMenshykh Can you please look at the visibility of the RestAPI* classes -- recommendations? Ok to be public, or should we move to another package?

shawncal:
@stephentoub Thoughts on this pattern? Sharing source files (internals only), importing into the various packages that need them, vs. building into a shared nuget and referencing with visibility=false

stephentoub:
> With this change, the common files are in an InternalUtilities folder, and are imported as source files into projects that use them. Each project therefore will build these into its dll, but they will be kept as internals, but there remains only one copy of the code for maintenance.

FWIW, this is what we do as well in dotnet/runtime.

We avoid InternalsVisibleTo like the plague, as well.  There are only a few remaining projects that still use it for test access, and they only do so because they were doing so before they were brought into dotnet/runtime and we haven't taken the time to remove them yet.  In other libraries where we want to test internals, we either use reflection or we have a dedicated unit tests project for that library that also builds the sources into the test project.

shawncal:
> We may consider to move the functionality back to the Skills. OpenAPI project/package and keep it there as internal classes until we have strong reasons to expose them as publicly.

Will do. I'll ask that we do this in a second PR (today) so that we can keep each one scoped to its purpose. Moving the RestApi* still will also touch unit tests, etc.

@SergeyMenshykh If you agree, please consider and, when ready, approve this PR first, and I'll follow up with the RestApi changes shortly thereafter.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -140,6 +142,37 @@ public CompleteRequestSettings RequestSettings
...
+    /// Create a native function instance, wrapping a native object method
+    /// </summary>
+    /// <param name="customFunction">Signature of the method to invoke</param>
+    /// <param name="skillName">SK skill name</param>
+    /// <param name="functionName">SK function name</param>
+    /// <param name="description">SK function description</param>
+    /// <param name="parameters">SK function parameters</param>
+    /// <param name="log">Application logger</param>
+    /// <returns>SK function instance</returns>
+    public static ISKFunction FromCustomMethod(
```
dluc, line 155: we need to refactor this out to an extension method

PR #883, opened 2023-05-09
Plan serialization: add support for indented JSON

dluc:
### Motivation and Context

When debugging and running demos, it's useful looking at a readable version of the plan.

### Description

Add optional indented serialization to Plan class.


dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -166,7 +166,7 @@ public Plan(ISKFunction function)
     /// <remarks>If Context is not supplied, plan will not be able to execute.</remarks>
     public static Plan FromJson(string json, SKContext? context = null)
     {
-        var plan = JsonSerializer.Deserialize<Plan>(json, new JsonSerializerOptions() { IncludeFields = true }) ?? new Plan(string.Empty);
+        var plan = JsonSerializer.Deserialize<Plan>(json, new JsonSerializerOptions { IncludeFields = true }) ?? new Plan(string.Empty);
```
shawncal, line 169: nit: could probably just be `new`, without the JsonSerializerOptions 
...but I don't mind having it show what kind it is explicitly.

shawncal (approved):

PR #888, opened 2023-05-09
Allow to replace functions (fix regression)

dluc:
### Motivation and Context

* Functions stored in a skill can be replaced, fix regression introduced in the last release
* Bump version to 0.14

### DIFF
--- a/dotnet/src/SemanticKernel/SkillDefinition/SkillCollection.cs
+++ b/dotnet/src/SemanticKernel/SkillDefinition/SkillCollection.cs
@@ -38,7 +38,7 @@ public class SkillCollection : ISkillCollection
         Verify.NotNull(functionInstance, "The function is NULL");
 
         ConcurrentDictionary<string, ISKFunction> skill = this._skillCollection.GetOrAdd(functionInstance.SkillName, static _ => new(StringComparer.OrdinalIgnoreCase));
-        skill.TryAdd(functionInstance.Name, functionInstance);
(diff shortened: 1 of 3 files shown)


dotnet/src/SemanticKernel/SkillDefinition/SkillCollection.cs
```diff
@@ -38,7 +38,7 @@ public ISkillCollection AddFunction(ISKFunction functionInstance)
         Verify.NotNull(functionInstance, "The function is NULL");
 
         ConcurrentDictionary<string, ISKFunction> skill = this._skillCollection.GetOrAdd(functionInstance.SkillName, static _ => new(StringComparer.OrdinalIgnoreCase));
-        skill.TryAdd(functionInstance.Name, functionInstance);
+        skill.AddOrUpdate(functionInstance.Name, functionInstance, (_, _) => functionInstance);
```
stephentoub, line 41: If you really want to allow replacement, it can just use the indexer:
```suggestion
        skill[functionInstance.Name] = functionInstance;
```

shawncal (approved):

stephentoub (approved):
This is fine for now.  But SkillCollection as designed currently is a dictionary, and it's unexpected for a dictionary.Add to replace what was already there.  That would typically be the realm of a Set method or an indexer setter.  Generally dictionary.Add would throw an exception if trying to add something that already existed in the dictionary.

PR #891, opened 2023-05-09
Make all protected members of ClientBase / OpenAIClientBase / AzureOpenAIClientBase private protected

stephentoub:
### Motivation and Context

I'm _assuming_ we don't actually intend for this surface area to be exposed and we don't want others deriving from these types.

### Description

As currently exposed, anyone could derive from these base types and see "Internal" methods (e.g. "InternalCompleteTextAsync").  This uses "private protected" to make everything actually internal and disable the ability to externally derive from these types.  For whenever we stand up documentation, this will also keep all of this protected surface area out of the docs.

### Contribution Checklist


shawncal (approved):

shawncal:
Pulled and ran integration tests -- all looks good. Thank you!

PR #893, opened 2023-05-09
Add support for custom delegates as SKFunctions

dluc:
### Motivation and Context

In some scenarios like RPC and OpenAPI it can be useful turning native code into SK Functions allowing to pass signature details (name, description, params) without the use of SK* attributes. For example, loading these details from OpenAPI docs or config files, rather than using reflection.

### Description

Add a new `FromNativeFunction` factory method to `SKFunction` to pass a custom delegate. The custom delegate must use one of the supported signatures, but doesn't need the "SK*[...]" attributes.

Example:

```
async Task<SKContext> MyFunctionAsync(SKContext contextIn)
{
    await Task.Delay(0);
    return contextIn;
}
```

```
ISKFunction function = SKFunction.FromNativeFunction(
            nativeFunction: MyFunctionAsync,
            skillName: "Test",
            functionName: "HelloWorld"
            description: "This is a test");
```

### DIFF
--- a/dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
+++ b/dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
@@ -25,8 +25,6 @@ namespace Microsoft.SemanticKernel.SkillDefinition;
 /// </summary>
 public sealed class SKFunction : ISKFunction, IDisposable
 {
-    public delegate Task<SKContext> CustomFunctionAsync(SKContext context);
-
     /// <inheritdoc/>
     public string Name { get; }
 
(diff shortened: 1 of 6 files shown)


dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -85,6 +83,37 @@ public CompleteRequestSettings RequestSettings
...
+    /// Create a native function instance, wrapping a delegate function
+    /// </summary>
+    /// <param name="nativeFunction">Function to invoke</param>
+    /// <param name="skillName">SK skill name</param>
+    /// <param name="functionName">SK function name</param>
+    /// <param name="description">SK function description</param>
+    /// <param name="parameters">SK function parameters</param>
+    /// <param name="log">Application logger</param>
+    /// <returns>SK function instance</returns>
+    public static ISKFunction FromNativeFunction(
```
shawncal, line 96: Remember, we are dropping the term NativeFunction. 
FromDelegate?

dluc: we can have that conversation offline

dluc:
The PR is ready to be merge. One of the issues fixed here has been addressed also in #900 so the remaining part is fixing how native function signatures are handled reusing the same code used for native skills.

lemillermicrosoft (approved):

adrianwyatt (approved):

PR #895, opened 2023-05-09
Plan viewer and approval for Copilot Chat

teresaqhoang:
### Motivation and Context

This PR enables Plan Viewer and Approval for Copilot chat. If SK plan is suggested on chatbot response generation, the plan will be returned to copilot chat for review before execution.

### Description

Flow

1. If ChatSkill suggests a Plan, that plan object will be returned to the client for user approval before it's run.
2. If the client receives a plan object back, it will provide a view only render of the plan in the chat window, and the user can approve or reject the plan.
3. If the user approves the plan, the User Intent determined from the first run (step 1) of the ChatSkill will be used, the plan will be invoked, and execution will continue as normal.
4. If the user rejects the plan, no action will be taken and chatbot response will be a constant string. 

Tech Details
WebApi
- Added new ProposedPlan class object (needs to be embedded so it can be parsed client side)

WebApp
- Added new Plan model objects

Screenshot 2023-05-09 at 7 06 31 PM

Screenshot 2023-05-09 at 7 09 10 PM

Screenshot 2023-05-09 at 7 09 18 PM


samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -323,44 +335,80 @@ public async Task<SKContext> ChatAsync(string message, SKContext context)
...
-        chatContext.Variables.Set("knowledgeCutoff", this._promptSettings.KnowledgeCutoffDate);
-        chatContext.Variables.Set("audience", userName);
-
-        // Extract user intent and update remaining token count
-        var userIntent = await this.ExtractUserIntentAsync(chatContext);
-        if (chatContext.ErrorOccurred)
+        if (chatContext.Variables.Get("userCancelledPlan", out var userCancelledPlan))
         {
-            return chatContext;
+            response = "Sorry this plan didn't meet your goals. How can I help you today?";
```
teresaqhoang, line 340: Do we want to have a pre-generated response like this or have the bot generate one? This way saves some time in response generation / roundtrips to the kernel

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -370,9 +418,10 @@ public async Task<SKContext> ChatAsync(string message, SKContext context)
         }
 
         // Extract semantic memory
+        // TODO: Ask TAO - on Plan approval response, do we need to exclude plan from semantic memory?
```
teresaqhoang, line 421: @TaoChenOSU

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -60,6 +60,11 @@ public class ChatSkill
     /// </summary>
     private readonly PlannerOptions _plannerOptions;
 
+    /// <summary>
```
teresaqhoang, line 63: @TaoChenOSU this is my work around since the context variables aren't being populated as expected

adrianwyatt:
Currently seeing the error `System.InvalidCastException: Unable to cast object of type 'CustomFunctionAsync' to type 'System.Func`2[Microsoft.SemanticKernel.Orchestration.SKContext,System.Threading.Tasks.Task`1[Microsoft.SemanticKernel.Orchestration.SKContext]]'.` using GitHub or Klarna plugins.

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
amsacha: Pre-generated will save on costing. Plus since the accept/cancel options are buttons, we arent getting more context from the user to drive a conversation. 
Perhaps reword it to "Sorry this plan didn't meet your goals. How shall I change this plan to better meet your requirements?"

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
lemillermicrosoft: Can you let me know more about the issues you are having, especially if you think they are an issue with `Plan`? Thanks!

lemillermicrosoft (approved):
Awesome stuff! I might reach out more about ideas, this will be nice to play with. Thanks!

samples/apps/copilot-chat-app/webapp/src/libs/models/Plan.ts
```diff
@@ -0,0 +1,18 @@
+export type IPlanInput = {
+    // These have to be capitalized to match the server response
+    Key: string;
+    Value: string;
+};
+
+export type IPlanStep = {
+    skill: string;
```
lemillermicrosoft, line 8: I'm curious about this definition, I'd expect things like 'name' instead of 'function'.

lemillermicrosoft: Ah I see the mapping in `sk-utilities`

lemillermicrosoft: Something like

```
export type IPlan = {
    skill: string;
    function: string;
    description: string;
    stepInputs: IPlanInput[];
    steps: IPlan[];
};
```

might be better.

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
amsacha: Is there a reason we are trying to avoid storing the plan in semantic memory?

If a user cancels a plan they may then type up a prompt that asks the bot to change the plan in a certain step or how it approaches something over multiple steps. The LLM probably needs the previous plan to know how to modify it.

samples/apps/copilot-chat-app/webapp/src/libs/semantic-kernel/sk-utilities.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+    return null;
+};
+
+const extractPlanSteps = (plan: any) => {
+    const planInputs: IPlanInput[] = [];
+    for (var input in plan.state) {
+        if (
+            // Omit reserved context variable names
+            plan.state[input].Key !== 'INPUT' &&
```
amsacha, line 31: Can we store the list of reserved keywords somewhere, and have a helper function do the check. I have a feeling as we interact with plans more, adding to context variables is going to happen in more places. It'll be easier to keep a track of it if its all in one place.

adrianwyatt (approved):

samples/apps/copilot-chat-app/webapp/src/libs/models/Plan.ts
teresaqhoang: Sounds good -- I'll hold off on this and you can just merge your changes from your 📝 Refactor plan model and UI to support nested steps commit in after?

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
teresaqhoang: Resolved with Tao offline - will leave as is. It doesn't really matter if the plan is stored in semantic memory or not so we'll leave it to the bot to determine that

samples/apps/copilot-chat-app/webapp/src/libs/models/Plan.ts
lemillermicrosoft: I'm not sure if/when any of that work will be going to `main` so I wouldn't count on it. I've already cherry-picked your UI work (and done this suggested change to support my scenarios) and will probably stay that way for the time being.

lemillermicrosoft (approved):

adrianwyatt (approved):

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
teresaqhoang: The issue we were seeing is that the context variables aren't being propagated down the prompt chain as expected. 

It's not specifically a planner issue, but when I tried setting a new context variable 'proposedPlan' in AcquireExternalInformation method, it's not returned in the chatContext response in the ChatAsync method. 

Another observation to back this is the token limits aren't being passed down, i.e.,
tokenLimit is set in ExtractUserMemoriesAsync, so it should be lower or equal to that value in following methods, but it's not. 

One run yielded:
- tokenLimit = 2829 at the end of ExtractUserMemoriesAsync
- then tokenLimit = 2846 in AcquireExternalInformationAsync (which is called after)

Issue #871, opened 2023-05-09
copilot-chat-app can't build


lone1y-51:
**Describe the bug**
I refer to https://github.com/microsoft/semantic-kernel/blob/main/samples/apps/copilot-chat-app/README.md this article, want to run copilot-chat-app, but can not successfully build project
```
/path/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs(20,90): error CS0103: 当前上下文中不存在名称“obj” [/path/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj]
/path/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs(29,102): error CS0103: 当前上下文中不存在名称“str” [/path/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj]
/path/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs(65,114): error CS0103: 当前上下文中不存在名称“text” [/path/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj]
```

**To Reproduce**
1. cd `/path/semantic-kernel/samples/apps/copilot-chat-app/webapi`
2. run `dotnet build`

**Expected behavior**
build success

**Desktop (please complete the following information):**
 - OS: mac  13.0 (22A380)  apple m1
 - dotnet sdk: 6.0.408

erphao:
the same error!

suxiao0102:
I also encountered the same problem，when run dotnet build 
https://github.com/microsoft/semantic-kernel/blob/main/samples/apps/copilot-chat-app/README.md

centos 7，dotnet 6.0.408 

commit 7ed26eacaa51182e51ff96ff4e04201c97fe450c

/root/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/Diagnostics/Verify.cs(20,90): error CS0103: The name 'obj' does not exist in the current context [/root/semantic-kernel/dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj]
[...]

PR #901, opened 2023-05-10
Removed CallerArgumentExpressionAttribute to fix support for dotnet 6

adrianwyatt:
### Motivation and Context
Build was failing on systems with dotnet 6 installed (and not dotnet 7)

### Description
Removed CallerArgumentExpressionAttribute and its usage.


dluc (approved):

dluc:
THANKS! We need CI on .NET6 only systems (and cross-platform code reviewers)

stephentoub:
This should not fail. It's supported by C# 10 /.NET 6. What exactly is failing?

adrianwyatt:
> This should not fail. It's supported by C# 10 / .NET 6. It even works when targeting .NET Framework. What exactly is failing?
> 
> (Also, if you just remove this it's going to remove all of the parameter names from ArgumentExceptions that get thrown.)

https://github.com/microsoft/semantic-kernel/issues/871 - let's get it working and we can re-implement a dotnet6 compatible version after.

adrianwyatt:
> THANKS! We need CI on .NET6 only systems (and cross-platform code reviewers)

Our PR/CI builds say they are .NET6 but apparently not quite? It's not apparent why PR/CI passes at the moment.

dluc (approved):

stephentoub:
Please don't submit this.

Just change the `nameof(obj)` to `"obj"`.

adrianwyatt:
> Please don't submit this.
> 
> Just change the `nameof(obj)` to `"obj"` (more specifically the name of the first parameter as a string).
> 
> Whoever is hitting this has an outdated C# compiler.

Updated with your suggestion and looks good - thanks!

dluc (approved):

PR #920, opened 2023-05-10
Fix dotnet graph-api-skills sample

MGudgin:
The skillParentDirectory local used to find the SummarizeSkill in the
graph-api-skills sample had one too few levels of directory navigation,
resulting in a DirectoryNotFoundException.

Initialize variable using RepoFiles.SampleSkillsPath from
kernel-syntax-examples project.

Reference kernel-syntax-examples project.

Make RepoFiles class and SampleSkillsPath method public in
kernel-syntax-examples.

Add namespace to graph-api-skills/Program.cs to avoid type collision
with Program class in kernel-syntax-examples project.


samples/dotnet/graph-api-skills/Program.cs
```diff
@@ -137,7 +137,7 @@ public static async Task Main()
             throw new InvalidOperationException("Unable to determine current assembly directory.");
         }
 
-        string skillParentDirectory = Path.GetFullPath(Path.Combine(currentAssemblyDirectory, "../../../../skills"));
+        string skillParentDirectory = Path.GetFullPath(Path.Combine(currentAssemblyDirectory, "../../../../../skills"));
```
dluc, line 140: I think this might be due to the compilation settings and where the code ends up running. What about re-using `RepoUtils.SampleSkillsPath()`? see https://github.com/microsoft/semantic-kernel/blob/main/samples/dotnet/kernel-syntax-examples/RepoUtils/RepoFiles.cs

MGudgin: Thanks, I will try that!

MGudgin: That works. Had to make RepoFiles and the method public, hope that is OK. Also had to add namespace to Program.cs file to avoid type collision.

samples/dotnet/graph-api-skills/MsGraphSkillsExample.csproj
```diff
@@ -29,6 +29,7 @@
     <ProjectReference Include="..\..\..\dotnet\src\Connectors\Connectors.AI.OpenAI\Connectors.AI.OpenAI.csproj" />
     <ProjectReference Include="..\..\..\dotnet\src\Skills\Skills.MsGraph\Skills.MsGraph.csproj" />
     <ProjectReference Include="..\..\..\dotnet\src\SemanticKernel\SemanticKernel.csproj" />
+    <ProjectReference Include="..\kernel-syntax-examples\KernelSyntaxExamples.csproj" />
```
adrianwyatt, line 32: We would like to keep the examples silo'd from each other. I recommend removing this reference and updating the path explicitly.

samples/dotnet/graph-api-skills/Program.cs
```diff
@@ -137,7 +139,7 @@ public static async Task Main()
             throw new InvalidOperationException("Unable to determine current assembly directory.");
         }
 
-        string skillParentDirectory = Path.GetFullPath(Path.Combine(currentAssemblyDirectory, "../../../../skills"));
+        string skillParentDirectory = RepoUtils.RepoFiles.SampleSkillsPath();
```
adrianwyatt, line 142:
```suggestion
        string skillParentDirectory = Path.GetFullPath(Path.Combine(currentAssemblyDirectory, "../../../../../skills"));
```

adrianwyatt (changes requested):

samples/dotnet/graph-api-skills/MsGraphSkillsExample.csproj
MGudgin: I got the opposite feedback from dluc, who suggested I not update the path explicitly, but rather use RepoFiles.SampleSkillsPath. Should I move RepoUtils to some shared assembly?

adrianwyatt: I wouldn't sweat that at the moment. Your change is fine in the current architecture of the repo. :) We'll take a pass at the samples in the near future to clean some of these patterns up, but you're good for now.

adrianwyatt (approved):

Issue #922, opened 2023-05-10
Add relative time concepts to the core TimeSkill component


tim-gill:
I have found that the SequentialPlanner regularly becomes confused when it is asked to deal with relative time concepts like "last wednesday" or "5 days ago". It will often try and make use of the supplied date() skill, adding a parameter to offset it (which is not supported). I propose that we add two new skills - one to find the date of the last instance of a given day name, and another to provide the date x days in the past.

Pull request adding these will be sent soon.

evchaki:
@tim-gill can't wait to see the PR come in!

tim-gill:
It is up at https://github.com/microsoft/semantic-kernel/pull/927 - would love feedback

tim-gill:
The one for day names worries me a little as it is currently locked to English day names. Suggestions on making it more general purpose would be great.

PR #927, opened 2023-05-10
Add new relative time skills for ' ago' and 'last '

tim-gill:
### Motivation and Context

The change enables the planner to make use of time references like "last wednesday" or "three days ago" which currently cause issues. This enables better use of SequentialPlanner when servicing user intent.
Issue: https://github.com/microsoft/semantic-kernel/issues/922

### Description

This adds two new functions to the core TimeSkill class in both .net and python.
1. DaysAgo(str days) - takes an argument representing a number of days and offsets it from the current date to return the resultant date as the answer.
2.  LastMatchingDay(str dayName) - takes the name of a day of the week and works backwards from yesterday to give it the first matching date of that day (e.g. "last wednesday"). Currently it only works with English day names, and this is reflected in the prompt description.

Simple unit tests are added for each function.

### Contribution Checklist


lemillermicrosoft (changes requested):
Great idea! Overall LGTM -- some build warnings in the test file that need to be cleaned up.

tim-gill:
Updated to remove the warnings in my code - there appears to be one still there from a nuget which I did not add.

lemillermicrosoft (approved):

dotnet/src/SemanticKernel.UnitTests/CoreSkills/TimeSkillTests.cs
```diff
@@ -25,4 +28,56 @@ public void ItCanBeImported()
...
+        DateTime expected = DateTime.Now.AddDays(-interval);
+        TimeSkill skill = new TimeSkill();
+        string result = skill.DaysAgo(interval.ToString(CultureInfo.CurrentCulture));
+        DateTime returned = DateTime.Parse(result, CultureInfo.CurrentCulture);
+        Assert.Equal(expected.Day, returned.Day);
+        Assert.Equal(expected.Month, returned.Month);
+        Assert.Equal(expected.Year, returned.Year);
+    }
+
+
```
dluc, line 45:
```suggestion

```

python/semantic_kernel/core_skills/time_skill.py
```diff
@@ -174,6 +176,48 @@ def minute(self) -> str:
...
+             SKContext["input"] = "3"
+             {{time.days_ago $input}} => Sunday, 7 May, 2023
+        """
+        d = datetime.date.today() - datetime.timedelta(days=int(days))
+        return d.strftime("%A, %d %B, %Y")
+
+    @sk_function(
+        description="Get the date of the last day matching the supplied day name in English"
+    )
+    def last_matching_day(self, day_name: str) -> str:
```
dluc, line 201: is this function English only? if so, what's the plan to support localization?

tim-gill: Sadly no plan as yet (I am not a python developer by day). Happy for any suggestions though. For now I have just ensured that the semantic description calls out the need for English in hope that the Planner (if used) can handle translation as needed

shawncal: @dluc Does this address your question (enough to complete this PR), or can you provide an example of what you have in mind for localization support?

lemillermicrosoft (dismissed):

python/semantic_kernel/core_skills/time_skill.py
dluc: @tim-gill consdiering it's English only and that the description for AI could lead to some incorrect plans when using the Planner, I would ask to remove this function from the PR. If you could think about a description that would work well with the planner, otherwise I would suggest having this as a private skill in local projects

tim-gill: @dluc - while I am happy enough to do this, I do want to point out that all of the other python functions around month and day in this file (not my changes) are also locked to English only as they also use the strftime function to generate their output - they also return different formats for the same data ("Sunday, 12 January" and "Sunday, January 12"). Do we also want to remove all of them?

dotnet/src/SemanticKernel/CoreSkills/TimeSkill.cs
```diff
@@ -202,6 +221,37 @@ public string HourNumber()
...
 
+    /// <summary>
+    /// Get the date of the previous day matching the supplied day name
+    /// </summary>
+    /// <example>
+    /// {{time.lastMatchingDay $dayName}} => Sunday, 7 May, 2023
+    /// </example>
+    /// <returns> The date of the last instance of this day name </returns>
+    /// <exception cref="ArgumentOutOfRangeException">dayName is not a recognized name of a day of the week</exception>
+    [SKFunction("Get the date of the last day matching the supplied day name in English")]
```
dluc, line 232: could you try this:
```
[SKFunction("Get the date of the last day matching the supplied week day name in English. Example: Che giorno era 'Martedi' scorso -> lastMatchingDay 'Tuesday' => Tuesday, 16 May, 2023")]
```

dluc (approved):

PR #930, opened 2023-05-11
Signal r integration

amsacha:
### Motivation and Context
SignalR realtime relay service added to support broadcasting to multiple users

### Description
integrated signalR into the frontend and backend; Responses are fed into chat messages directly 

### Contribution Checklist

### DIFF
--- a/samples/apps/copilot-chat-app/webapp/src/components/chat/ChatRoom.tsx
+++ b/samples/apps/copilot-chat-app/webapp/src/components/chat/ChatRoom.tsx
@@ -12,6 +12,8 @@ import { RootState } from '../../redux/app/store';
 import { updateConversation } from '../../redux/features/conversations/conversationsSlice';
 import { ChatHistory } from './ChatHistory';
 import { ChatInput } from './ChatInput';
+import { UserAsk, UserAskResult } from './../../libs/semantic-kernel/skMultiUserChat';
+import { useSKMultiUserChat } from './../../libs/semantic-kernel/useSKMultiUserChat';
 
 const log = debug(Constants.debug.root).extend('chat-room');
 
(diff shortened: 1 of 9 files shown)


samples/apps/copilot-chat-app/webapp/src/libs/semantic-kernel/skMultiUserChat.ts
```diff
@@ -0,0 +1,112 @@
...
+        };
+	}
+
+	// Set up a SignalR connection to the specified hub URL
+	setupSignalRConnectionToChatHub() {
+		const serviceUrl = process.env.REACT_APP_BACKEND_URI as string;
+		const connectionHubUrl = (new URL('/chatHub', serviceUrl)).toString();
+
+		const signalRConnectionOptions = {
+			skipNegotiation: true,
```
amsacha, line 48: Moving conversation here @Vivihung , @adrianwyatt, @glahaye 

I can't establish a connection without skipping negotiations with the server. This is being blocked by the CORS policy of the app. I am not familiar enough with CORS to know if that has some unintended consequences?

Error:
Ensure CORS requests include credentials only when allowed
A cross-origin resource sharing (CORS) request was blocked because it was configured to include credentials but the Access-Control-Allow-Credentials response header of the request or the associated preflight request was not set to true.

amsacha: Changing CORS would look something like this:

    app.UseCors(x => x
        .AllowAnyMethod()
        .AllowAnyHeader()
        .SetIsOriginAllowed(origin => true) // allow any origin
        .AllowCredentials()); // allow credentials
        
But i am not sure what the fallout of that is?

Source: https://www.abrahamberg.com/blog/aspnet-signalr-and-react/ , scroll to Dealing with CORS section

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatRoom.tsx
```diff
@@ -56,6 +58,41 @@ export const ChatRoom: React.FC = () => {
...
+            content: userAskResult.value,
+            authorRole: AuthorRoles.Bot,
+        };    
+        dispatch(updateConversation({ message: messageResult, chatId: selectedId }));
+    }
+
+    React.useEffect(() => {
+        // This code will run once, when the component is mounted
+        console.log('SignalR setup called');
+        const { events } = SKMultiUserChatConnector();
```
amsacha, line 90: Moving conversation to this PR  @teresaqhoang / @TaoChenOSU / @adrianwyatt / @Vivihung 
The call to SKMultiUserChatConnector() should only happen once when the component is mounted. However there seems to be a bug somewhere thats making the component get mounted twice.

PR #931, opened 2023-05-11
Fix python bug preventing use of Azure OpenAI and OpenAI at the same …

mkarle:
### Motivation and Context

Using both OpenAI and Azure OpenAI services would not work due to the way they were configured.  The `openai` module would be configured globally with keys and endpoints for one type of service and then overwritten when adding another service. Instead, we can configure the settings when making the API calls.

Also, some integration tests are failing because of throttling by GPT4.

### Description

- Use the API settings when calling the openai API instead of configuring gloabally
- Increased delay to 3 seconds for some integration tests to avoid rate limit errors
- Relaxed the assertion strings for some integration tests
- Ran pre-commit linting
- Switch some tests to GPT Turbo 3.5


aUsABuisnessman (approved):
i approve

python/tests/integration/completions/e2e_text_completion.py
```diff
@@ -79,7 +80,7 @@ async def summarize_function_test(kernel: sk.Kernel):
     )
     output = str(summary).strip()
     print(f"Summary using input context: '{output}'")
-    assert "humans" in output or "Humans" in output or "preserve" in output
+    assert "human" in output or "Human" in output or "preserve" in output
```
dluc, line 83: I used the plural because "human" is in the input and there is a risk of the test passing if the function returns the input instead of a summary. Any reason for this change?

python/tests/integration/completions/e2e_text_completion.py
```diff
@@ -7,9 +7,10 @@
 logger = logging.getLogger()
 
 
-async def retry(func, retries=15, delay=1):
+async def retry(func, retries=15, delay=3):
     for i in range(retries):
         try:
+            time.sleep(delay)
```
dluc, line 13: there is a sleep at line 22, after an error occurs. Why add another sleep before retrying? 

btw this will cause a sleep also before the 1st attempt

dluc (changes requested):
there's a couple of issues in the tests I think needs addressed before merging

python/tests/integration/completions/e2e_text_completion.py
mkarle: Yeah, there are six calls to it in this file and I could only get it to not reach the rate limit by sleeping between each one.

python/tests/integration/completions/e2e_text_completion.py
mkarle: That makes sense. I was getting responses like 
"Summary using context and additional string: 'Protect humans, obey, self-preserve'
Summary using context, additional variables, and additional string: 'Human safety, obey, self-preservation, birds.'"

lemillermicrosoft:
It'd be nice to see a green pass on your repo fork as well to confirm this fixes things.

python/tests/integration/completions/e2e_text_completion.py
awharrison-28: We should consider skipping the majority of the OpenAI tests like we do for the .NET SK

dluc: @mkarle we can increase the delay after the error, which would be equivalent right? (without the need to wait before the first request)

python/tests/integration/completions/e2e_text_completion.py
```diff
@@ -7,9 +7,10 @@
 logger = logging.getLogger()
 
 
-async def retry(func, retries=15, delay=1):
+async def retry(func, retries=15, delay=3):
```
dluc, line 10:
```suggestion
async def retry(func, retries=15, delay=6):
```

python/tests/integration/completions/e2e_text_completion.py
dluc, line 13:
```suggestion
```

dluc (changes requested):
recommended changes

dluc (approved):

mkarle:
More greens in my fork https://github.com/mkarle/semantic-kernel/actions/runs/4950586657/jobs/8854401012

PR #933, opened 2023-05-11
Fix linter not running

mkarle:
### Motivation and Context

Black wasn't running because it depended on paths-filter which was renamed in a previous commit.


mkarle:
You can see the Python Lint workflow is running again https://github.com/microsoft/semantic-kernel/actions/workflows/lint.yml

dluc (approved):

PR #952, opened 2023-05-11
ITextCompletion Multiple Results (Incl. Streaming)

### LINKED ISSUE #526: Support for chat completion choices parameter
[Open AI API](https://platform.openai.com/docs/api-reference/chat/create#chat/create-n) has parameter `n` for number of choices to be retruned. Is it possible to add this param to the SK? 

image

At the moment class `ChatCompletionRequest` lacks it

```cs
public sealed class ChatCompletionRequest
{
    public string? Model { get; set; }
    public IList<Message> Messages { get; set; } = new List<Message>();
    public double? Temperature { get; set; }
    public double? TopP { get; set; }
    public double? PresencePenalty { get; set; }
    public double? FrequencyPenalty { get; set; }
    public object? Stop { get; set; }
    public int? MaxTokens { get; set; } = 256;
}
```

as well as `GenerateMessageAsync` method returns single string value

```cs
public interface IChatCompletion
{
   public Task<string> GenerateMessageAsync(
        ChatHistory chat,
        ChatRequestSettings requestSettings,
}

```

RogerBarreto:
### Motivation and Context

Potential to get multiple results for the same prompt with and without streaming support.

Resolves #526 
Closes #526 

MultiChatCompletionStream2

### Description

This PR introduces abstractions and the capability to get multiple results for the same prompt using the same request including streaming support.

Proposed:

- Added 2 result abstractions `ITextCompletionResult` and `ITextCompletionStreamingResult`
- Changed current base `ITextCompletion` interface with the multiple results abstraction capability
- Added 2 extensions methods to `ITextCompletion` to keep it retro compatible / avoid major breaking changes
- Added concrete implementation to: Azure OpenAI / Open AI / Hugging Face completion impl.
- Added Samples (with and without streaming) how to use multiple results

Breaking potential:
- Any custom completion implementation will need to implement the new methods in `ITextCompletion` interface with the `ITextCompletionResult`/`ITextCompletionStreamingResult` abstractions. Pretty straightforward and example was added in `CustomLLM` sample.

### Contribution Checklist


dluc (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/TextCompletion/AzureTextCompletion.cs
```diff
@@ -68,5 +68,21 @@ public sealed class AzureTextCompletion : AzureOpenAIClientBase, ITextCompletion
...
+        return this.InternalMultiCompletionStreamAsync(text, requestSettings, cancellationToken);
+    }
+
+    public Task<IReadOnlyList<ITextCompletionResult>> CompleteMultiAsync(
+        string text,
+        CompleteRequestSettings requestSettings,
+        CancellationToken cancellationToken = default)
+    {
+        return this.InternalCompleteTextMultiAsync(text, requestSettings, cancellationToken);
+    }
```
shawncal, line 86: One thought: consider using overloads for CompleteAsync:
- one with a resultsPerPrompt arg that returns a collection, and 
- one without that returns a single instance. Same for CompleteStreamAsync().

Then we don't need the "Multi" methods, and `ResultsPerPrompt` need not be in the CompleteRequestSettings, but rather can be specified on a per-usage basis (i.e. not hard-coded in a prompt's config.json)

RogerBarreto: Yes that's the final intention here, I used other methods just to validate the approach. Will move those internals to be all multi capable with no dupes.

RogerBarreto: I moved forward with the implementation and added the multiple results as the standard interfaces behavior. 

The completionsettings has the property but this is not currently exposed to the json file, reading from a json file currently will default to 1 ResultsPerPrompt.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -266,8 +262,8 @@ private static CompletionsOptions CreateCompletionsOptions(string text, Complete
             FrequencyPenalty = (float?)requestSettings.FrequencyPenalty,
             PresencePenalty = (float?)requestSettings.PresencePenalty,
             Echo = false,
-            ChoicesPerPrompt = 1,
-            GenerationSampleCount = 1,
+            ChoicesPerPrompt = requestSettings.ResultsPerPrompt,
```
dluc, line 265: FYI: Has a valid range of 1 to 128. https://learn.microsoft.com/en-us/dotnet/api/azure.ai.openai.completionsoptions.choicesperprompt?view=azure-dotnet-preview

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -266,8 +262,8 @@ private static CompletionsOptions CreateCompletionsOptions(string text, Complete
...
+            GenerationSampleCount = requestSettings.ResultsPerPrompt,
```
dluc, line 266: FYI: must be greater than ChoicesPerPrompt - https://learn.microsoft.com/en-us/dotnet/api/azure.ai.openai.completionsoptions.generationsamplecount?view=azure-dotnet-preview

Maybe the documentation is wrong, if that's the case I would raise a bug/reach out to Azure OpenAI team

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureChatCompletion.cs
```diff
@@ -76,20 +76,19 @@ public ChatHistory CreateNewChat(string instructions = "")
...
         CompleteRequestSettings requestSettings,
         CancellationToken cancellationToken = default)
     {
-        return this.InternalCompleteTextUsingChatAsync(text, requestSettings, cancellationToken);
+        return this.InternalCompletionStreamAsync(text, requestSettings, cancellationToken);
     }
 
-    /// <inheritdoc/>
-    public IAsyncEnumerable<string> CompleteStreamAsync(string text,
+    public Task<IReadOnlyList<ITextCompletionResult>> GetCompletionsAsync(
```
dluc, line 87: here and above: xmldoc

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/OpenAIChatCompletion.cs
```diff
@@ -59,21 +59,19 @@ public ChatHistory CreateNewChat(string instructions = "")
         return InternalCreateNewChat(instructions);
     }
 
-    /// <inheritdoc/>
-    public Task<string> CompleteAsync(
+    public IAsyncEnumerable<ITextCompletionStreamingResult> GetStreamingCompletionsAsync(
```
dluc, line 62: xmldoc

dotnet/src/Connectors/Connectors.AI.OpenAI/TextCompletion/OpenAITextCompletion.cs
```diff
@@ -50,5 +51,21 @@ public sealed class OpenAITextCompletion : OpenAIClientBase, ITextCompletion
         CancellationToken cancellationToken = default)
     {
         return this.InternalCompletionStreamAsync(text, requestSettings, cancellationToken);
+    }*/
+
+    public IAsyncEnumerable<ITextCompletionStreamingResult> GetStreamingCompletionsAsync(
```
dluc, line 56: ditto, pls check all public methods

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        var completions = await textCompletion.GetCompletionsAsync(text, requestSettings, cancellationToken).ConfigureAwait(false);
+
+        if (completions.Count == 0)
+        {
+            throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "No completion result");
+        }
+
+        if (completions.Count > 1)
+        {
+            throw new NotSupportedException($"{nameof(CompleteAsync)} don't support multiple completions results. Use {nameof(ITextCompletion.GetCompletionsAsync)} instead.");
```
dluc, line 38: I don't understand this part: the client returned multiple completions and the error seems to say the opposite. Also why is this an error, when would it happen? Could you add some comment explaining the scenario?

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        var completionResults = textCompletion.GetStreamingCompletionsAsync(text, requestSettings, cancellationToken).ConfigureAwait(false);
+
+        await foreach (var completionResult in completionResults)
+        {
+            await foreach (var word in completionResult.CompleteStreamAsync(cancellationToken))
+            {
+                yield return word;
+            }
+
+            yield return Environment.NewLine;
```
dluc, line 67: could you explain the logic about the two loops and why we insert a new line? This will also behave differently on Windows (`/r/n`) vs Linux (`/n`), which could be a problem, e.g. with tokenization

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
rogerio-barreto: If any user attemps to use the extension methods providing resultsperprompt > 1 they will get the error. I can just ignore the other results silently or throw, either options is fine for me, but IMHO hidding/losing info is worse than failing.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
rogerio-barreto: I'm not clearly sure why the extra line is needed, I can removed. 

I saw this in the internal implementation of Azure Clients added to their samples, I can remove it since we will have only one choice when output from the CompleteStream extension impl.

Here: [Sample04_Streaming_Chat](https://github.com/Azure/azure-sdk-for-net/blob/3fa8005a1af13f48f7bb38efa0ba272b53351ef3/sdk/openai/Azure.AI.OpenAI/tests/Samples/Sample04_StreamingChat.cs#L43)

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
rogerio-barreto: Good catch, will add this validation. Interestingly enough current SDK impl don't validate that size neither the usage of ChoicesPerPrompt > GenerationSampleCount which gives a bad request.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
rogerio-barreto: Yeah, the greaterthan seems to be wrong, should be GreaterThen Or Equals.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
RogerBarreto: That's a fair point and I also noted that currently the Results implementation actually loops on all the choices found, I changed the behavior on the non streaming to actually be the same. Not adding error implementation I reduce the changes of breaking any code that were using those methods before.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -98,17 +100,24 @@ public HuggingFaceTextCompletion(string apiKey, string model, string endpoint =
     }
 
     /// <inheritdoc/>
-    public async Task<string> CompleteAsync(string text, CompleteRequestSettings requestSettings, CancellationToken cancellationToken = default)
+    public async IAsyncEnumerable<ITextCompletionStreamingResult> GetStreamingCompletionsAsync(
```
shawncal, line 103: I'm not quite sure what to recommend here, but something to think about...
As it is, I don't think this needs to return an IAsyncEnumerable collection of (effectively) IAsyncEnumerables, because it depends on ExecuteGetCompletionsAsync which will always return a full, completed list. In that sense, it could just be an IList<ITextCompletionStreamingResult>.

My only hesitation is that, as an interface method, a future implementer might have to execute these N completions serially, waiting for each one to complete, but then hand back a result as soon as each one is complete, which is an IAsyncEnumerable. Even then, I might argue that they should construct their list of response objects up front, and populate the (inner) IAsyncEnumerables whenever they can.

Can you see a good use case for IAsyncEnum of IAsyncEnums, where the caller would want to receive each result (object) when available, instead of the result contents when available?

shawncal: @stephentoub, if you know of any good precedent here.

The idea is that you request N completions, each of which can stream, all of which could be in parallel. I think you just want the flat list of IAsyncEnums up front, but I might be missing a scenario.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/TextCompletionStreamingResult.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    public TextCompletionStreamingResult(StreamingChoice choice)
+    {
+        this._choice = choice;
+    }
+
+    public async Task<string> CompleteAsync(CancellationToken cancellationToken = default)
+    {
+        var fullMessage = new StringBuilder();
+        await foreach (var message in this._choice.GetTextStreaming(cancellationToken).ConfigureAwait(false))
```
shawncal, line 24: Will calling GetTextStreaming return the same results twice? If someone calls TextCompletionStreamingResult.CompleteAsync() twice, do they get the streaming result once, and the completed result the second time?

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextCompletionResult.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.AI.TextCompletion;
+
+public interface ITextCompletionResult
+{
+    Task<string> CompleteAsync(CancellationToken cancellationToken = default);
```
shawncal, line 10: The method names on these 2 interfaces:
CompleteAsync sounds like this is where you are invoking the work, but I think we're literally just getting a Task or IAsyncEnumerable already in progress, right? If that's the case, then perhaps Get* methods are better here (i.e. GetCompletionAsync(), GetCompletionStreamingAsync) as that sounds more like you're just retrieving the thing instead of invoking a new action on the model.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -98,17 +100,24 @@ public HuggingFaceTextCompletion(string apiKey, string model, string endpoint =
...
-        return await this.ExecuteCompleteRequestAsync(text, cancellationToken).ConfigureAwait(false);
+        foreach (var completion in await this.ExecuteGetCompletionsAsync(text, cancellationToken).ConfigureAwait(false))
+        {
+            yield return completion;
+        }
     }
 
-    public IAsyncEnumerable<string> CompleteStreamAsync(
+    /// <inheritdoc/>
+    public async Task<IReadOnlyList<ITextCompletionResult>> GetCompletionsAsync(
```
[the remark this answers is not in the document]
RogerBarreto: I can't, this works in the other completions because the internal already gives me a IAsyncEnumerable, when using the non async pattern ToAsyncEnumerable actually converts the Task<IReadOnlyList<ItextCompletionStreamingResult>> which breaks the interface.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -146,9 +148,13 @@ private async Task<string> ExecuteCompleteRequestAsync(string text, Cancellation
             var response = await this._httpClient.SendAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
             var body = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 
-            var completionResponse = JsonSerializer.Deserialize<List<TextCompletionResponse>>(body);
+            List<TextCompletionResponse>? completionResponse = JsonSerializer.Deserialize<List<TextCompletionResponse>>(body);
+            if (completionResponse is null)
+            {
+                throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "No completions were found");
```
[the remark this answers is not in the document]
RogerBarreto: Agreed. Changed the message to "Unexpected response from model", added the response as Data to the exception payload if the caller want to investigate.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextCompletionResult.cs
RogerBarreto: Agreed, renamed them accordingly

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/TextCompletionStreamingResult.cs
RogerBarreto: That's right, made a double check test here. 

The result is kept in memory after getting from the model.

So, if someone calls for the same `Result` instance a second time (either Streaming or not) it will get the same response.

ℹ️ Any further call on finished streaming result will be almost instant.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
RogerBarreto: AzureOpenAI client implementation currently returns IAsyncEnums for both the Streaming Results and the Messaging Results.

They use the double foreach approach in their [sample here](https://github.com/Azure/azure-sdk-for-net/blob/1cc96c56cce7891c31963fd059c5debb2ee9f258/sdk/openai/Azure.AI.OpenAI/tests/Samples/Sample04_StreamingChat.cs#L37).

I believe there are better ways to use it for parallel processing like you were saying and I did that in our samples where actually I get the CompletionStreamingResults first in one call and cosume them in parallel in a Task.WhenAll.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -98,17 +99,24 @@ public HuggingFaceTextCompletion(string apiKey, string model, string endpoint =
...
+    public async IAsyncEnumerable<ITextCompletionStreamingResult> GetStreamingCompletionsAsync(
+        string text,
+        CompleteRequestSettings requestSettings,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
-        return await this.ExecuteCompleteRequestAsync(text, cancellationToken).ConfigureAwait(false);
+        foreach (var completion in await this.ExecuteGetCompletionsAsync(text, cancellationToken).ConfigureAwait(false))
+        {
+            yield return completion;
+        }
```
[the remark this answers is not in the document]
stephentoub: Oh, nevermind the first question, I see now that the return type of GetStreamingCompletionsAsync isn't what I thought it was.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -144,11 +145,21 @@ private async Task<string> ExecuteCompleteRequestAsync(string text, Cancellation
...
+            if (completionResponse is null)
+            {
+                throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Unexpected response from model")
+                {
+                    Data = { { "ModelResponse", body } },
+                };
+            }
 
-            return completionResponse.First().Text!;
+            return completionResponse.Select((completion) => new TextCompletionStreamingResult(completion.Text)).ToList();
```
stephentoub, line 162:
```suggestion
            return completionResponse.ConvertAll(c=> new TextCompletionStreamingResult(c.Text));
```

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -144,11 +145,21 @@ private async Task<string> ExecuteCompleteRequestAsync(string text, Cancellation
             };
 
             var response = await this._httpClient.SendAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
```
stephentoub, line 147:
```suggestion
            using var response = await this._httpClient.SendAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
```

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -257,6 +253,13 @@ private static ChatHistory PrepareChatHistory(string text, CompleteRequestSettin
 
     private static CompletionsOptions CreateCompletionsOptions(string text, CompleteRequestSettings requestSettings)
     {
+        if (requestSettings.ResultsPerPrompt < 1 ||
+            requestSettings.ResultsPerPrompt > 128)
+        {
+            // <see cref="CompletionsOptions.ChoicesPerPrompt"/> must be in range between 1 and 128.
+            throw new ArgumentException($"{nameof(requestSettings.ResultsPerPrompt)} invalid. The value must be in range between 1 and 128.");
```
stephentoub, line 260:
```suggestion
            throw new ArgumentException("The value must be in range between 1 and 128, inclusive.", $"{nameof(requestSettings)}.{nameof(requestSettings.ResultsPerPrompt)}");
```

dotnet/src/Connectors/Connectors.UnitTests/HuggingFace/TextCompletion/HuggingFaceTextCompletionTests.cs
```diff
@@ -63,7 +63,7 @@ public void Dispose()
         GC.SuppressFinalize(this);
     }
 
-    protected virtual void Dispose(bool disposing)
+    private void Dispose(bool disposing)
```
stephentoub, line 66: You can now simplify both of these methods down to just:
```C#
public void Dispose() => this._response.Dispose();
```

stephentoub: (though I see now this is in a test, so, not a big deal either way)

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextCompletion.cs
```diff
@@ -12,25 +12,26 @@ namespace Microsoft.SemanticKernel.AI.TextCompletion;
...
-    /// Creates a completion for the prompt and settings.
+    /// Get streaming completion results for the prompt and settings.
     /// </summary>
     /// <param name="text">The prompt to complete.</param>
     /// <param name="requestSettings">Request settings for the completion API</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <returns>Streaming content of the text generated by the remote model</returns>
-    public IAsyncEnumerable<string> CompleteStreamAsync(string text,
+    /// <returns>List of different completion streaming results generated by the remote model</returns>
+    public IAsyncEnumerable<ITextCompletionStreamingResult> GetStreamingCompletionsAsync(
```
stephentoub, line 33: Note that the "public" here and above are ok but not actually needed, since this is an interface.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Streaming content of the text generated by the remote model</returns>
+    public static async IAsyncEnumerable<string> CompleteStreamAsync(this ITextCompletion textCompletion,
+        string text,
+        CompleteRequestSettings requestSettings,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        var completionResults = textCompletion.GetStreamingCompletionsAsync(text, requestSettings, cancellationToken).ConfigureAwait(false);
+
+        await foreach (var completionResult in completionResults)
```
stephentoub, line 56:
```suggestion
        var completionResults = textCompletion.GetStreamingCompletionsAsync(text, requestSettings, cancellationToken);

        await foreach (var completionResult in completionResults.ConfigureAwait(false))
```

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            await foreach (var word in completionResult.GetCompletionStreamingAsync(cancellationToken))
```
stephentoub, line 58:
```suggestion
            await foreach (var word in completionResult.GetCompletionStreamingAsync(cancellationToken).ConfigureAwait(false))
```

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <summary>
+    /// Creates a completion for the prompt and settings.
+    /// </summary>
+    /// <param name="textCompletion">Target interface to extend</param>
+    /// <param name="text">The prompt to complete.</param>
+    /// <param name="requestSettings">Request settings for the completion API</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Streaming content of the text generated by the remote model</returns>
+    public static async IAsyncEnumerable<string> CompleteStreamAsync(this ITextCompletion textCompletion,
```
stephentoub, line 49: Similar naming concern here.

stephentoub (dismissed):

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
RogerBarreto: Agree, this methods are here for retrocompatibility, those names are the current existing names in our ITextCompletion interface. Keeping the same here so it won't break everything.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/CompleteRequestSettings.cs
```diff
@@ -47,6 +47,13 @@ public class CompleteRequestSettings
     /// </summary>
     public IList<string> StopSequences { get; set; } = Array.Empty<string>();
 
+    /// <summary>
+    /// How many completions to generate for each prompt. Default is 1.
+    /// Note: Because this parameter generates many completions, it can quickly consume your token quota.
+    /// Use carefully and ensure that you have reasonable settings for max_tokens and stop. 
+    /// </summary>
+    public int ResultsPerPrompt { get; set; } = 1;
```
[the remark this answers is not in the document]
RogerBarreto: Since this class lives in the abstractions namespaces I didn't want to introduce any validation here because the 1-128 range is a limitation for Azure and OpenAI only. Might worth think about validations abstractions later on.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/TextCompletionStreamingResult.cs
shawncal: Thanks!

shawncal (approved):

shawncal (approved):

PR #962, opened 2023-05-11
Function config extensibility

dmytrostruk:
### Motivation and Context

**NOTE**: This PR contains breaking changes, that are applied to SK C# version only, Python version is not included.

Currently, _config.json_ is coupled to completion settings only, which are specific to OpenAI parameters (e.g. temperature, presence_penalty, frequency_penalty etc).

In order to allow AI extensibility in the future, we need to update _config.json_ schema to allow flexible settings format for different AI models.

We also need to be able to specify _recommended_ services for function, so Semantic Kernel will be able to automatically use one of the recommended services in case if it's registered.

New _config.json_ proposed schema:
```json
{ 
  "description": "Generate a funny joke", 
  "input": { 
    "parameters": [ 
      { 
        "name": "input", 
        "description": "Joke subject", 
        "defaultValue": "" 
      } 
    ] 
  }, 
  "default_settings": { 
    "max_tokens": 256, 
    "temperature": 0.9 
  }, 
  "services": [ 
    { 
      "model_id": "text-davinci-003", 
      "order": 1, 
      "settings": { 
        "max_tokens": 1000, 
        "temperature": 0.9, 
        "top_p": 0.0, 
        "presence_penalty": 0.0, 
        "frequency_penalty": 0.0, 
        "stop_sequences": [ 
          "\n" 
        ] 
      } 
    }, 
    { 
      "model_id": "gpt-3.5-turbo", 
      "order": 2, 
      "settings": { 
        "system_message": "You are assistant to generate funny jokes", 
        "max_tokens": 1000, 
        "temperature": 0.9 
      } 
    }, 
    { 
      "model_id": "dalle", 
      "order": 3, 
      "settings": { 
        "n": 1, 
        "size": "1024x1024" 
      } 
    }, 
    { 
      "model_id": "gpt2", 
      "order": 4 
    } 
  ] 
} 
```

Schema description:
- **description**: as it was previously, describes purpose of the function. 
- **input**: as it was previously, contains information about input parameters. 
- **default_settings**: contains default settings for AI model. Will be used when there is no matching recommended AI service, which is registered in kernel. 
- **services** (optional): contains list of recommended AI services to use when running a function. Contains objects with following properties: 
  - **model_id**: model identifier as described by AI provider, e.g. text-davinci-003. 
  - **order**: service priority order. Kernel will choose recommended service which is registered and has highest priority order. Used to avoid coupling on how (in what order) services are located in JSON file.  
  - **settings**: dynamic object, which can contain any properties, specific to AI model. Will be defined in C# using _System.Text.Json.Nodes.JsonObject_.  

### Description

1. Replaced `CompleteRequestSettings` in AI service interfaces to `JsonObject`.
2. Updated `Kernel` to use recommended service or `default_settings` by default.
3. Added `SKFunctionConfigTests.cs` tests which show usage of new **config.json**.
4. Updated all **config.json** files in samples to follow new format.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextCompletion.cs
```diff
@@ -20,7 +21,7 @@ public interface ITextCompletion
     /// <returns>Text generated by the remote model</returns>
     public Task<string> CompleteAsync(
         string text,
-        CompleteRequestSettings requestSettings,
+        JsonObject requestSettings,
```
dluc, line 24: we looked into this approach 2-3 months ago and agreed not to use generic object/dictionaries, and use type safety, clearly defined types and interfaces

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptTemplateConfig.cs
```diff
@@ -107,69 +54,63 @@ public class InputConfig
...
-    [JsonPropertyOrder(1)]
-    public int Schema { get; set; } = 1;
+    public class ServiceConfig
+    {
+        /// <summary>
+        /// Model ID in AI service provider (e.g. text-davinci-003).
+        /// </summary>
+        [JsonPropertyName("model_id")]
+        [JsonPropertyOrder(1)]
+        public string ModelId { get; set; } = string.Empty;
```
dluc, line 66: How are you mapping this value to Azure, where the API requires a deployment name, which is customized by developers?

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptTemplateConfig.cs
```diff
@@ -107,69 +54,63 @@ public class InputConfig
...
 
-    /// <summary>
-    /// Type, such as "completion", "embeddings", etc.
-    /// </summary>
-    /// <remarks>TODO: use enum</remarks>
-    [JsonPropertyName("type")]
-    [JsonPropertyOrder(2)]
-    public string Type { get; set; } = "completion";
+        /// <summary>
+        /// Service priority order. Kernel will check all registered services and choose one with highest priority order.
```
dluc, line 69: what's the scenario for needing an order? Also, you can use an array to sort models, removing the need to lookup all the values and sort

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
```diff
@@ -39,7 +40,7 @@ public interface ISKFunction
     /// <summary>
     /// AI service settings
     /// </summary>
-    CompleteRequestSettings RequestSettings { get; }
+    JsonObject ServiceSettings { get; }
```
dluc, line 43: same comment here. There are other ways to deal with custom config settings, e.g. using generics

dotnet/src/SemanticKernel/SkillDefinition/InlineFunctionsDefinitionExtension.cs
```diff
@@ -49,19 +51,30 @@ public static class InlineFunctionsDefinitionExtension
...
-                FrequencyPenalty = frequencyPenalty,
-                MaxTokens = maxTokens,
-                StopSequences = stopSequences?.ToList() ?? new List<string>()
+                stopSequencesArray.Add(JsonValue.Create<string>(stopSequence));
             }
+        }
+
+        var serviceSettings = new JsonObject
+        {
+            ["temperature"] = temperature,
```
dluc, line 66: this is one place where we can see the problem with JsonObject:
* Keys are hardcoded
* There is no way to know which keys are required - normally one would just use IntelliSense
* there is no validation, each value is an object

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
dmytrostruk: @dluc Thanks for review! Could you please provide an example how we could achieve the flexibility in settings (i.e. support different set of fields for different model types) using generics? We assume that there are a lot of different model types and settings schema can be changed in the future, and we can't predict what the format would be.

dotnet/src/SemanticKernel/SkillDefinition/InlineFunctionsDefinitionExtension.cs
dmytrostruk: We can still use serialization/deserialization and work with strongly typed models on AI services level (like in `ClientBase.cs` class of OpenAI connector). On AI service level we still perform validation.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextCompletion.cs
dmytrostruk: Assuming the settings should be flexible, I'm not sure it will be possible to define types to cover all cases, AI types and models.

PR #964, opened 2023-05-11
A few examples of using SK SDK in DI/IoC container

SergeyMenshykh:
### Description
This PR adds a few examples showing the way SK SDK can be used by applications that use DI/IoC containers for bootstrapping.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/KernelBuilder.cs
+++ b/dotnet/src/SemanticKernel/KernelBuilder.cs
@@ -44,7 +44,7 @@ public sealed class KernelBuilder
             this._promptTemplateEngine ?? new PromptTemplateEngine(this._logger),
             this._memory,
             this._config,
-            this._logger ?? NullLogger<Kernel>.Instance
+            this._logger
         );
 
         // TODO: decouple this from 'UseMemory' kernel extension
--- a/dotnet/src/SemanticKernel/KernelBuilder.cs
+++ b/dotnet/src/SemanticKernel/KernelBuilder.cs
@@ -24,6 +24,7 @@ public sealed class KernelBuilder
     private ILogger _logger = NullLogger.Instance;
     private IMemoryStore? _memoryStorage = null;
     private IDelegatingHandlerFactory? _httpHandlerFactory = null;
+    private IPromptTemplateEngine? _promptTemplateEngine;
     private readonly AIServiceCollection _aiServices = new();
 
     /// <summary>
@@ -40,7 +41,7 @@ public sealed class KernelBuilder
         var instance = new Kernel(
             new SkillCollection(this._logger),
             this._aiServices.Build(),
-            new PromptTemplateEngine(this._logger),
+            this._promptTemplateEngine ?? new PromptTemplateEngine(this._logger),
             this._memory,
             this._config,
             this._logger ?? NullLogger<Kernel>.Instance
@@ -88,6 +89,18 @@ public sealed class KernelBuilder
     {
         Verify.NotNull(storage);
         this._memoryStorage = storage;
+        return this;
+    }
+
+    /// <summary>
+    /// Add prompt template engine to the kernel to be built.
+    /// </summary>
+    /// <param name="promptTemplateEngine">Prompt template engine to add.</param>
+    /// <returns>Updated kernel builder including the prompt template engine.</returns>
+    public KernelBuilder WithPromptTemplateEngine(IPromptTemplateEngine promptTemplateEngine)
+    {
+        Verify.NotNull(promptTemplateEngine);
+        this._promptTemplateEngine = promptTemplateEngine;
         return this;
     }
 
--- a/dotnet/src/SemanticKernel/Services/AIServiceProvider.cs
+++ b/dotnet/src/SemanticKernel/Services/AIServiceProvider.cs
@@ -4,7 +4,7 @@ using System;
 using System.Collections.Generic;
 
 namespace Microsoft.SemanticKernel.Services;
-internal class AIServiceProvider : NamedServiceProvider<IAIService>, IAIServiceProvider
+public class AIServiceProvider : NamedServiceProvider<IAIService>, IAIServiceProvider
 {
     public AIServiceProvider(Dictionary<Type, Dictionary<string, Func<object>>> services, Dictionary<Type, string> defaultIds)
         : base(services, defaultIds)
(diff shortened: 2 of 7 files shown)


dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -40,7 +41,7 @@ public IKernel Build()
         var instance = new Kernel(
             new SkillCollection(this._logger),
             this._aiServices.Build(),
-            new PromptTemplateEngine(this._logger),
+            this._promptTemplateEngine ?? new PromptTemplateEngine(this._logger),
             this._memory,
             this._config,
             this._logger ?? NullLogger<Kernel>.Instance
```
RogerBarreto, line 47: The `??` logic for logger only exists for the Kernel, shouldn't all the other be the same?

i.e:
```
var buildLogger = this._logger ?? NullLogger<Kernel>.Instance
var instance = new Kernel(
            new SkillCollection(buildLogger),
            this._aiServices.Build(),
            new PromptTemplateEngine(buildLogger),
            this._memory,
            this._config,
            buildLogger)
```

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -91,6 +92,18 @@ public KernelBuilder WithMemoryStorage(IMemoryStore storage)
         return this;
     }
 
+    /// <summary>
+    /// Add prompt template engine to the kernel to be built.
+    /// </summary>
+    /// <param name="promptTemplateEngine">Prompt template engine to add.</param>
+    /// <returns>Updated kernel builder including the prompt template engine.</returns>
+    public KernelBuilder WithPromptTemplateEngine(IPromptTemplateEngine promptTemplateEngine)
```
RogerBarreto, line 100: Can this all be just: AddAIService<IPromptTemplateEngine>(promptTemplateEngine); ?

dotnet/src/SemanticKernel/KernelBuilder.cs
SergeyMenshykh: Got you, it appeared that the ?? logic for logger is not needed because the logger field is initialized by default and the WithLogger method does not allow nulls. So, I updated the code to to remove the check.

dotnet/src/SemanticKernel/KernelBuilder.cs
SergeyMenshykh: You probably meant AddService<IPromptTemplateEngine>(promptTemplateEngine) instead of Add**AI**Service<IPromptTemplateEngine>(promptTemplateEngine), right?

If that is the case, then the answer is yes, we can remove the WithPromptTemplateEngine, WithLogger, and WithMemory methods, but the question we should ask ourselves is whether we should do it. Removing these methods would make it difficult for developers to understand the capabilities of the builder. Without these methods, when developers type **new KernelBuilder().**, the only method that Intellisense will show them is AddService<T>, which is not very informative. On the other hand, with all the With* methods shown by Intellisense, developers can immediately start exploring them in the IDE by looking at their signatures to familiarize themselves with the public API surface of the class. Although this is not a replacement for official documentation, it is a good addition to it.

dluc, line 100: I think we're going in the opposite direction of the plan to simplify the public API, reducing methods, and the plan to use .NET DI. 

Before merging this branch into main we'll need to address this.

dotnet/src/SemanticKernel/Services/AIServiceProvider.cs
```diff
@@ -4,7 +4,7 @@
 using System.Collections.Generic;
 
 namespace Microsoft.SemanticKernel.Services;
-internal class AIServiceProvider : NamedServiceProvider<IAIService>, IAIServiceProvider
+public class AIServiceProvider : NamedServiceProvider<IAIService>, IAIServiceProvider
```
dluc, line 7: similar comment about public API, long term maintainability

dotnet/src/SemanticKernel/KernelBuilder.cs
SergeyMenshykh: Completely agree with you @dluc that the API should be simplified but not by decreasing number of methods but by redesigning SK SDK components so that those methods are moved to the classes they naturally belong to.

dotnet/src/SemanticKernel/Services/AIServiceProvider.cs
SergeyMenshykh: The change can be appreciated by .net developers who may want/need to use SK SDK in an advanced way by creating Kernel manually without KernelBuilder, why not? Anyway, I am open to discussing any concerns regarding this approach.

samples/dotnet/kernel-syntax-examples/Example35_DIContainer.cs
```diff
@@ -0,0 +1,129 @@
...
+{
+    public static async Task RunAsync()
+    {
+        await UseKernelInDIPowerAppAsync();
+
+        await UseKernelInDIPowerApp_AdvancedScenarioAsync();
+    }
+
+    /// <summary>
+    /// This example shows how to registered Kernel in DI container without registering it's dependencies and using KernelBuilder instead.
```
shawncal, line 28: ``
```suggestion
    /// This example shows how to register a Kernel in a DI container using KernelBuilder instead of 
    /// registering its dependencies.
```

samples/dotnet/kernel-syntax-examples/Example35_DIContainer.cs
```diff
@@ -0,0 +1,129 @@
...
+        });
+
+        //Registering class that uses Kernel to execute a skill
+        collection.AddTransient<KernelClient>();
+
+        //Creating a service provider for resolving registered services
+        var serviceProvider = collection.BuildServiceProvider();
+
+        //If an application follows DI guidelines, the following line is unnecessary because DI will inject an instance of the KernelClient class to a class that references it.
+        //DI container guidelines - https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-guidelines#recommendations
```
shawncal, line 56:
```suggestion
        //If an application follows DI guidelines, the following line is unnecessary because DI will inject an instance 
        //of the KernelClient class to a class that references it.
        //DI container guidelines: 
        // https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-guidelines#recommendations
```

shawncal (approved):

shawncal:
merging to feature branch

PR #965, opened 2023-05-12
Move CopilotChat webapi to nuget package dependencies

adrianwyatt:
### Motivation and Context
A step towards allowing CopilotChatApi project to be built without project dependencies. This should help users who want to copy/paste the project for their own projects, without having to take project dependencies on the rest of the repo.

### Description
- Removed all project dependencies in CopilotChatApi and replaced them with their equivalent NuGets.


samples/apps/copilot-chat-app/webapi/CopilotChatApi.csproj
```diff
@@ -14,6 +14,14 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.SemanticKernel" Version="0.13.442.1-preview" />
```
glahaye, line 17: This will force us to have the core team release new nugets whenever we make changes to their code.  If you are comfortable with that, proceed.

I'm just wondering whether we want to do this now or after Build, when there should be less churn...

adrianwyatt: Correct - we've stabilized on our SK dependencies at the moment, so I was hoping we could try this out. We can always switch back to a project reference if we need to. This will also have the side effect of us being able to validate changes to the kernel before we take a dependency on them.

glahaye (approved):

hathind-ms (approved):

chentao1133:
3Q

PR #979, opened 2023-05-12
Default minrelevancescore zero

awharrison-28:
### Motivation and Context
A default minRelevanceThreshold of 0.7 works well with OpenAI's ada embeddings generation, but not necessarily with other models. It also is likely to return no results as the most relevant item could be less than 0.7. To ensure that similarity search always returns at least one result, this limit should default to 0. 

### Description
minRelevanceThreshold for similarity search is 0.0 across the board. 

### Contribution Checklist


lemillermicrosoft (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -134,7 +134,7 @@ public AzureCognitiveSearchMemory(string endpoint, TokenCredential credentials)
         string collection,
         string query,
         int limit = 1,
-        double minRelevanceScore = 0.7,
+        double minRelevanceScore = 0.0,
```
dluc, line 137: this changes the behavior from returning something relevant to returning anything, e.g. potentially irrelevant information that will affect the quality of every component using this. E.g. the planner searching for relevant skills will get irrelevant ones in many more scenarios. I would be more useful having a default behavior that exclude some noise, e.g. in this case 0.5.

awharrison-28: It will always return the 'most relevant' result(s). In some cases, the 'most relevant' result may have a similarity score of <0.5. For example, if you're using sentence-transformers/all-MiniLM-L6-v2, a similarity score of 0.35 is quite useful, but for text-similarity-ada-002, this score would be considered not very similar. Hardcoding a threshold is very likely to return empty, which is not expected when searching for a 'nearest' value.

lemillermicrosoft: FWIW ActionPlanner doesn't use relevancy and for SequentialPlanner it is off by default and users must supply a value here of their choice to enable relevancy filtering.

adrianwyatt (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
dluc: @awharrison-28 "It will always return the 'most relevant' result(s)." sure, but when the most relevant has a similarity of 0.1 it's hardly useful and can lead to hallucination and/or incorrect text generation. This is such a core component that a similar change will propagate in multiple places. I would prefer a safer a approach, at the cost of not returning any data. Users should override the value to take "more risk" (decreasing the default value) rather than taking "less risk" (increasing the threshold).

shawncal:
Folks, I don't have enough context to submit this one. Given the discussion, I'd like to have a quick chat about it. Let's follow up.

shawncal:
@awharrison-28, @dluc, were you able to connect about this change? Shall we proceed, modify, or close?

shawncal:
@awharrison-28 @dluc I'm going to flip this to "Draft" to get it off the queue.
Please follow up - reopen and ping me to merge, or "Close with comment" if we're not going to take this change.

shawncal:
Closing PR. @awharrison-28 feel free to reopen if/when ready to submit.

PR #986, opened 2023-05-14
Copilot chat: ChatController

TaoChenOSU:
### Motivation and Context

We would like developers to be able to deploy SKaaS that is not tightly coupled to Copilot Chat.

### Description

1. Group Chat related components to their own folders.
2. Strip out anything related to the current Chat features out of the SemanticKernelController.
3. Create a ChatController specifically to invoke the ChatSkill.
4. Refactor the WebApp to use the new ChatController. Make the service invocation pattern more consistent.
5. Remove the SKBotAudienceMember and refactor ChatStatus. Note that this component does not currently take any effect. But this component will be useful moving forward with the multi user support.
6. Disable implicit using.

### Contribution Checklist


samples/apps/copilot-chat-app/webapi/Controllers/CopilotChatControllers/ChatController.cs
```diff
@@ -0,0 +1,204 @@
...
+
+/// <summary>
+/// Controller responsible for handling chat messages and responses.
+/// </summary>
+[ApiController]
+public class ChatController : ControllerBase, IDisposable
+{
+    private readonly ILogger<ChatController> _logger;
+    private readonly List<IDisposable> _disposables;
+    private readonly string _chatSkillName = "ChatSkill";
```
glahaye, line 31: You could make the two constants string "const" instead of readonly

samples/apps/copilot-chat-app/webapi/Controllers/SemanticKernelController.cs
```diff
@@ -1,40 +1,24 @@
...
-using Microsoft.SemanticKernel.Skills.OpenAPI.Authentication;
-using SemanticKernel.Service.Config;
 using SemanticKernel.Service.Model;
-using SemanticKernel.Service.Skills;
-using SemanticKernel.Service.Storage;
 
 namespace SemanticKernel.Service.Controllers;
 
 [ApiController]
 public class SemanticKernelController : ControllerBase, IDisposable
```
glahaye, line 14: This class doesn't need to be IDisposable anymore.

Also remove the declaration of _disposables below.

glahaye (dismissed):

glahaye (approved):

samples/apps/copilot-chat-app/webapi/Controllers/CopilotChatControllers/ChatController.cs
```diff
@@ -0,0 +1,204 @@
...
+using Microsoft.SemanticKernel.SkillDefinition;
+using Microsoft.SemanticKernel.Skills.MsGraph;
+using Microsoft.SemanticKernel.Skills.MsGraph.Connectors;
+using Microsoft.SemanticKernel.Skills.MsGraph.Connectors.Client;
+using Microsoft.SemanticKernel.Skills.OpenAPI.Authentication;
+using SemanticKernel.Service.Config;
+using SemanticKernel.Service.Model;
+using SemanticKernel.Service.Skills;
+
+namespace SemanticKernel.Service.Controllers;
```
Vivihung, line 21: If I recall correctly, we are using the namespace naming convention based on folder structure.
So the namespace of ChatController should be `<project namespace>.Controllers.CopilotChatController`.

Another feedback - the folder name might be a bit redundant. Recommend something like:
Controllers/Chat/ChatController

samples/apps/copilot-chat-app/webapp/src/libs/connectors/useConnectors.ts
```diff
@@ -6,11 +6,11 @@ import { AuthHeaderTags } from '../../redux/features/plugins/PluginsState';
 import { AuthHelper } from '../auth/AuthHelper';
 import { TokenHelper } from '../auth/TokenHelper';
 import { IAsk } from '../semantic-kernel/model/Ask';
-import { useSemanticKernel } from '../semantic-kernel/useSemanticKernel';
+import { SemanticKernelService } from '../services/SemanticKernelService';
 
 export const useConnectors = () => {
     const { instance, inProgress } = useMsal();
-    const sk = useSemanticKernel(process.env.REACT_APP_BACKEND_URI as string);
+    const sk = new SemanticKernelService(process.env.REACT_APP_BACKEND_URI as string);
```
Vivihung, line 13: (nit -  can be done in the future) It seems `process.env.REACT_APP_BACKEND_URI` is scattered over the places. Maybe we can add logic in the constructor: If the parameter is undefined, then use process.env.REACT_APP_BACKEND_URI; otherwise take the parameter.

Note: JavaScript doesn't support contructor overload. Hence we would need to manually check the value of parameters.

Vivihung (changes requested):
Thanks for the refactory. I've left some comments

samples/apps/copilot-chat-app/webapi/Controllers/CopilotChatControllers/ChatController.cs
amsacha: So far we've been keeping the namespaces up to the folder structure, and then everything inside the folder fell into the same namespace. We have also inserted .Service into the namespace name everywhere. I think that was nice especially when you had a bunch of really tiny classes spread out over many files. It was granular but not overtly so. 
I think theres enough going on in the controllers and they are distinct enough to have their own namespace but i wouldnt want to apply that to using SemanticKernel.Service.Model; or SemanticKernel.Service.Skills;

+1 on the folder name redundancy.

samples/apps/copilot-chat-app/webapi/Controllers/CopilotChatControllers/ChatController.cs
```diff
@@ -0,0 +1,204 @@
...
+
+/// <summary>
+/// Controller responsible for handling chat messages and responses.
```
amsacha, line 24:
```suggestion
/// Controller responsible for handling chat messages and responses.
/// This controller utilizes the planner to generate well formed step by step plans 
/// and to connect copilot chat to a variety of plugins.
```

samples/apps/copilot-chat-app/webapi/Controllers/CopilotChatControllers/ChatController.cs
TaoChenOSU: We will likely do another pass on this. I agree we do need to clean up the names.

samples/apps/copilot-chat-app/copilotchat/Controllers/BotController.cs
```diff
@@ -20,7 +26,7 @@ public class BotController : ControllerBase
     private readonly ChatSessionRepository _chatRepository;
     private readonly ChatMessageRepository _chatMessageRepository;
     private readonly BotSchemaOptions _botSchemaOptions;
-    private readonly AIServiceOptions _embeddingOptions;
+    private readonly CopilotChatAIServiceOptions _embeddingOptions;
```
glahaye, line 29: The name of the variable doesn't seem to match with its type...

samples/apps/copilot-chat-app/copilotchat/Extensions/ServiceExtensions.cs
```diff
@@ -0,0 +1,183 @@
...
+            .Bind(configuration.GetSection(PlannerOptions.PropertyName))
+            .ValidateOnStart()
+            .PostConfigure(TrimStringProperties);
+
+        services.AddOptions<PromptsOptions>()
+            .Bind(configuration.GetSection(PromptsOptions.PropertyName))
+            .ValidateOnStart()
+            .PostConfigure(TrimStringProperties);
+
+
```
glahaye, line 74: Limit that to one empty lines, not 3. I'm surprised our our formatter doesn't complain about stuff like this.

samples/apps/copilot-chat-app/copilotchat/Models/Ask.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SemanticKernel.CopilotChat.Models;
+
+public class Ask
```
glahaye, line 8: This is re-usable across more than just the copilot chat app.  It would be great to have it defined somewhere where it can be re-used

samples/apps/copilot-chat-app/copilotchat/Models/AskResult.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SemanticKernel.CopilotChat.Models;
+
+public class AskResult
```
glahaye, line 8: Same: re-usable

samples/apps/copilot-chat-app/copilotchat/Options/CopilotChatAIServiceOptions.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.ComponentModel.DataAnnotations;
+
+namespace SemanticKernel.CopilotChat.Config;
+
+/// <summary>
+/// Configuration options for Copilot Chat AI services, such as Azure OpenAI and OpenAI.
+/// </summary>
+public sealed class CopilotChatAIServiceOptions
```
glahaye, line 10: This is generic and re-usable.  Why not simply re-use from WebApi?

samples/apps/copilot-chat-app/copilotchat/Options/NotEmptyOrWhitespaceAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace SemanticKernel.CopilotChat.Config;
+
+/// <summary>
+/// If the string is set, it must not be empty or whitespace.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+internal sealed class NotEmptyOrWhitespaceAttribute : ValidationAttribute
```
glahaye, line 12: More re-usable stuff copied?

samples/apps/copilot-chat-app/copilotchat/Options/RequiredOnPropertyValueAttribute.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Reflection;
+
+namespace SemanticKernel.CopilotChat.Config;
+
+/// <summary>
+/// If the other property is set to the expected value, then this property is required.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+internal sealed class RequiredOnPropertyValueAttribute : ValidationAttribute
```
glahaye, line 13: Again re-usable...

samples/apps/copilot-chat-app/copilotchat/Options/CopilotChatAIServiceOptions.cs
adrianwyatt: I believe that would create a circular dependency.

samples/apps/copilot-chat-app/copilotchat/Storage/ChatMessageRepository.cs
```diff
@@ -30,11 +33,6 @@ public async Task<ChatMessage> FindLastByChatIdAsync(string chatId)
     {
         var chatMessages = await this.FindByChatIdAsync(chatId);
         var first = chatMessages.MaxBy(e => e.Timestamp);
-        if (first is null)
-        {
-            throw new KeyNotFoundException($"No messages found for chat '{chatId}'.");
-        }
-
-        return first;
+        return first is null ? throw new KeyNotFoundException($"No messages found for chat '{chatId}'.") : first;
```
glahaye, line 36: I think this can be made even shorter and more legible:

```suggestion
        return first ?? throw new KeyNotFoundException($"No messages found for chat '{chatId}'.") ;
```

samples/apps/copilot-chat-app/webapi/Options/NotEmptyOrWhitespaceAttribute.cs
```diff
@@ -8,7 +8,7 @@ namespace SemanticKernel.Service.Config;
 /// If the string is set, it must not be empty or whitespace.
 /// </summary>
 [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
-internal sealed class NotEmptyOrWhitespaceAttribute : ValidationAttribute
+public sealed class NotEmptyOrWhitespaceAttribute : ValidationAttribute
```
adrianwyatt, line 11: Is 'public' still necessary?

samples/apps/copilot-chat-app/webapi/Options/RequiredOnPropertyValueAttribute.cs
```diff
@@ -9,7 +9,7 @@ namespace SemanticKernel.Service.Config;
 /// If the other property is set to the expected value, then this property is required.
 /// </summary>
 [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
-internal sealed class RequiredOnPropertyValueAttribute : ValidationAttribute
+public sealed class RequiredOnPropertyValueAttribute : ValidationAttribute
```
adrianwyatt, line 12: Is 'public' still necessary?

samples/apps/copilot-chat-app/webapi/Options/NotEmptyOrWhitespaceAttribute.cs
glahaye, line 11: public for reuse outside the project, right?  Then, we should re-use it outside of the project instead of making a copy!

samples/apps/copilot-chat-app/webapi/Options/RequiredOnPropertyValueAttribute.cs
glahaye, line 12: Same: public -> re-use

samples/apps/copilot-chat-app/webapi/Program.cs
```diff
@@ -27,7 +28,9 @@ public static async Task Main(string[] args)
         builder.Services
             .AddSingleton<ILogger>(sp => sp.GetRequiredService<ILogger<Program>>()) // some services require an un-templated ILogger
             .AddOptions(builder.Configuration)
+            .AddCopilotChatOptions(builder.Configuration)
```
adrianwyatt, line 31: [nit] have these as separate calls from the others. For example:
```
builder.Services
            .AddSingleton<ILogger>(sp => sp.GetRequiredService<ILogger<Program>>()) // some services require an un-templated ILogger
            .AddOptions(builder.Configuration)
            .AddSemanticKernelServices()
            .AddPersistentChatStore();

builder.Services
            .AddCopilotChatOptions(builder.Configuration)
            .AddCopilotChatPlannerServices()
```

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
```diff
@@ -25,11 +24,6 @@ internal static class SemanticKernelExtensions
     /// </summary>
     public delegate Task RegisterSkillsWithKernel(IServiceProvider sp, IKernel kernel);
 
-    /// <summary>
-    /// Delegate to register skills with a planner.
-    /// </summary>
-    public delegate Task RegisterSkillsWithPlanner(IServiceProvider sp, CopilotChatPlanner planner);
```
glahaye, line 31: This looks like it should be left in there!

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
```diff
@@ -58,51 +52,23 @@ internal static IServiceCollection AddSemanticKernelServices(this IServiceCollec
             .AddEmbeddingBackend(serviceProvider.GetRequiredService<IOptionsSnapshot<AIServiceOptions>>()
                 .Get(AIServiceOptions.EmbeddingPropertyName)));
 
-        // Planner (AI plugins) support
-        IOptions<PlannerOptions>? plannerOptions = services.BuildServiceProvider().GetService<IOptions<PlannerOptions>>();
```
glahaye, line 62: Should this really be removed?

samples/apps/copilot-chat-app/webapi/ServiceExtensions.cs
```diff
@@ -42,47 +40,12 @@ internal static IServiceCollection AddOptions(this IServiceCollection services,
...
-            .PostConfigure(TrimStringProperties);
-
-        // Bot schema configuration
-        services.AddOptions<BotSchemaOptions>()
-            .Bind(configuration.GetSection(BotSchemaOptions.PropertyName))
-            .ValidateOnStart()
-            .PostConfigure(TrimStringProperties);
-
-        // Document memory options
-        services.AddOptions<DocumentMemoryOptions>()
```
glahaye, line 70: Should this be really removed?

glahaye (changes requested):
See comments

samples/apps/copilot-chat-app/webapi/WebApi.csproj
```diff
@@ -0,0 +1,42 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
adrianwyatt, line 1: Naming recommendation:
1. Have this project be "SemanticKernelWebApi.csproj"
2. and the copilot chat project be "CopilotChatServices.csproj
3. Put both projects under a "CopilotChat" directory. We'll likely move it, but it would help if they were obviously bundled together.

samples/apps/copilot-chat-app/copilotchat/Storage/ChatMessageRepository.cs
adrianwyatt, line 36:
```suggestion
        return first ?? throw new KeyNotFoundException($"No messages found for chat '{chatId}'.");
```

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/copilotchat/Options/CopilotChatAIServiceOptions.cs
glahaye: Right, but maybe we should extract the re-usable stuff to a lib. Maybe not in this PR, which is already quite big

samples/apps/copilot-chat-app/webapi/Controllers/CopilotChatControllers/ChatController.cs
TaoChenOSU: Planner can be disabled. The skill/function doesn't necessarily use planner.

samples/apps/copilot-chat-app/copilotchat/Models/Ask.cs
TaoChenOSU: The CopilotChatApi cannot make reference to the WebApi as it will introduce circular dependency. We will need to further consider how to create common dependencies.

samples/apps/copilot-chat-app/copilotchat/Options/NotEmptyOrWhitespaceAttribute.cs
TaoChenOSU: Circular dependency

samples/apps/copilot-chat-app/copilotchat/Options/RequiredOnPropertyValueAttribute.cs
TaoChenOSU: Circular dependency.

TaoChenOSU: We need to improve the pattern.

samples/apps/copilot-chat-app/copilotchat/Models/Ask.cs
glahaye: Just thinking out loud here. But what if we had the Copilot Chat reference WebApi instead of the other way around?

You might ask me "How would the WebApi load the Copilot Chat without knowing about it?"

We could use a plugin-style loader and/or something that would use reflection and/or something delegate to load Copilot Chat from WebApi.

This is for a future PR, however.

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
TaoChenOSU: This is not used.

samples/apps/copilot-chat-app/webapi/ServiceExtensions.cs
TaoChenOSU: It's moved the CopilotChat project.

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
glahaye: @adrianwyatt - Wasn't this part of your effort to genericize the loading of skills?

samples/apps/copilot-chat-app/webapi/WebApi.csproj
TaoChenOSU: If the SemanticKernelWebApi lives under a CopilotChat directory, would it still be considered a SKaaS sample?

samples/apps/copilot-chat-app/copilotchat/Models/AskResult.cs
TaoChenOSU: Circular dependency.

TaoChenOSU: We don't necessarily need to use Ask for chat. Will be considered at the 2nd round of refactoring.

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
TaoChenOSU: This is moved to CopilotChatApi

samples/apps/copilot-chat-app/webapi/WebApi.csproj
hathind-ms: Yes, it will be fine

samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
adrianwyatt: It was, but we didn't end up using. Let's remove it.

samples/apps/copilot-chat-app/webapi/WebApi.csproj
TaoChenOSU: Here is the current sample folder tree:
-samples
-- apps
---copilot-chat-app
----copilotchatapi
----webapi
----webapp
---- ...
--- ...
-- ...

Discussed with Adrian. This aligns with what he suggested. Comment resolved.

adrianwyatt (approved):

amsacha (dismissed):

glahaye (dismissed):

Vivihung (dismissed):
I worried the new project structure will confuse developers who are not familiar with the codebase. It's not straightforward to have two api project of the same app. And it's not common to include Controllers in a lib. I will let these big changes in but will work with Adrian offline to see if there's any mitigation. 

Given said that, it's a great effort to separate components apart. Appreciate the huge refactory here.

craigomatic:
> I worried the new project structure will confuse developers who are not familiar with the codebase. It's not straightforward to have two api project of the same app. And it's not common to include Controllers in a lib. I will let these big changes in but will work with Adrian offline to see if there's any mitigation.
> 
> Given said that, it's a great effort to separate components apart. Appreciate the huge refactory here.

I agree with Vivian, it's going to feel strange to have controllers in a project that does not appear to be runnable (I don't see a Program.cs)

Another pattern to consider:

* Everything under your new copilotchat folder becomes 1 or more services (ChatService, DocumentService, etc) and anything aspnet specific disappears (controllers, etc)
* Your services know nothing about the underlying hosting mechanism - it could be a console app, windows forms, aspnet, etc
* They do the work you have in the controllers currently
* We ship another aspnet app that uses the services so there remains a way for devs to deploy copilot chat

And if the objective is to allowing devs to deploy SKaaS with or without copilot chat, does this also imply that we should move the webapi project out of the copilot-chat-app folder?

dotnet/SK-dotnet.sln
```diff
@@ -59,7 +59,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "connectors", "connectors",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Connectors.UnitTests", "src\Connectors\Connectors.UnitTests\Connectors.UnitTests.csproj", "{EB3FC57F-E591-4C88-BCD5-B6A1BC635168}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CopilotChatApi", "..\samples\apps\copilot-chat-app\webapi\CopilotChatApi.csproj", "{CCABF515-2C79-453E-A5A2-69C69B8D172E}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebApi", "..\samples\apps\copilot-chat-app\webapi\WebApi.csproj", "{CCABF515-2C79-453E-A5A2-69C69B8D172E}"
```
adrianwyatt: To keep with the theme of being explicit, how about "CopilotChatWebApi"?

adrianwyatt (approved):

adrianwyatt (approved):

Issue #960, opened 2023-05-11
TextMemorySkill's Recall function blocks threads and causes WPF app to freeze


SenseYang:
**Describe the bug**
I crafted a WPF app and hooked a button with a RelayCommand that will eventually invoke `ISKFunction.InvokeAsync` method. Even without the `await`, it will still blocks the WPF app. And this exact same function, when running with a console app, produced results and finished within several seconds.
I am using `<PackageReference Include="Microsoft.SemanticKernel" Version="0.12.207.1-preview" />`
**To Reproduce**
1. Create a WPF app and add a button to the view, and create a new ISKFunction by calling `IKernel.CreateSemanticFunction`.
2. Configure the button with a RelayCommand.
3. Let the RelayCommand to invoke `ISKFunction.InvokeAsync`
4. Run debug in Visual Studio.

**Expected behavior**
Clicking the button shouldn't block the WPF app.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - IDE: Visual Studio
 - NuGet Package Version: `<PackageReference Include="Microsoft.SemanticKernel" Version="0.12.207.1-preview" />`

**Additional context**
The VS debug output also shows some interesting information:
```
The thread 0x5fb8 has exited with code 0 (0x0).
The thread 0x25f0 has exited with code 0 (0x0).
Exception thrown: 'System.IO.IOException' in System.Net.Sockets.dll
Exception thrown: 'System.IO.IOException' in System.Net.Security.dll
Exception thrown: 'System.IO.IOException' in System.Private.CoreLib.dll
The thread 0x622c has exited with code 0 (0x0).
The thread 0x7b14 has exited with code 0 (0x0).
The thread 0x3868 has exited with code 0 (0x0).
The thread 0x1788 has exited with code 0 (0x0).
```
But the WPF process is just blocked without seeing any exceptions at all.

SenseYang:
Forgot to mention - exact same function, when running with a console app, produced results and finished within several seconds.

runceel:
I have created a plain WPF app that invokes ISKFunction (a semantic function) on .NET 7. Plain means without any packages.
Semantic Kernel version is `0.13.442.1-preview`.

The source code is:

```csharp
using Microsoft.SemanticKernel;
using Microsoft.SemanticKernel.SkillDefinition;
using System;
using System.Windows;

namespace WpfApp2;
/// <summary>
/// Interaction logic for MainWindow.xaml
/// </summary>
[...]

PR #987, opened 2023-05-14
Fix deadlock in TextMemory skill

dluc:
### Motivation and Context

WPF compatibility, see https://github.com/microsoft/semantic-kernel/issues/960

### Description

Refactor blocking code

### DIFF
--- a/dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
+++ b/dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
@@ -1,11 +1,13 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.SemanticKernel.Diagnostics;
+using Microsoft.SemanticKernel.Memory;
 using Microsoft.SemanticKernel.Orchestration;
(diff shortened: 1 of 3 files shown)


dotnet/src/SemanticKernel/CoreSkills/TextMemorySkill.cs
```diff
@@ -119,30 +121,21 @@ public string Recall(string text, SKContext context)
...
-        {
-            var memory = memories.FirstOrDefault();
-            resultString = (memory != null) ? memory.Metadata.Text : string.Empty;
-        }
-        else
-        {
-            resultString = JsonSerializer.Serialize(memories.Select(x => x.Metadata.Text));
-        }
+        // Search memory
+        List<MemoryQueryResult>? memories = await context.Memory
```
stephentoub, line 126: `memories` should never be null.  You can remove the `?` as well as the null check below.

stephentoub (approved):
I see ToEnumerable also being used in:
https://github.com/microsoft/semantic-kernel/blob/4a4e5ffa706cb2c3ef44c21d909fb71045f10037/samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs#L697
Does that need to be fixed as well?

stephentoub (approved):
I see ToEnumerable also being used in:
https://github.com/microsoft/semantic-kernel/blob/4a4e5ffa706cb2c3ef44c21d909fb71045f10037/samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs#L697
Does that need to be fixed as well?

adrianwyatt (approved):

PR #993, opened 2023-05-15
Feature serviceregistry

SergeyMenshykh:

This PR introduces several changes aimed to simplify the kernel and enhancing its extensibility for AI services management. The changes include the following:

1. Kernel simplification and extensibility: The code responsible for AI services management, such as registration and resolution, has been moved out of the Kernel and KernelConfig classes into a new class called AIServiceProvider. This restructuring promotes the principle of Single Responsibility, ensuring that each class has a clear and focused purpose.

2. Hosting apps customization: With the introduction of the AIServiceProvider, hosting applications now have the flexibility to provide their own implementation of the provider if needed. This promotes extensibility, allowing the hosting apps to tailor the AI service management according to their specific requirements.

3. AI services registration improvements: A new method called WithAiService has been added, which enables the registration of any new AI service that is not currently supported.

4. Improved kernel builder interface: The Add* methods for AI services registration have been moved from the KernelConfig class to the KernelBuilder. This change aligns the interface with other KernelBilder methods, such as WithLogger and WithMemory, creating a more cohesive and intuitive kernel builder interface.

### Contribution Checklist


samples/apps/copilot-chat-app/webapi/SemanticKernelExtensions.cs
SergeyMenshykh, line 1: This file has compilation issues because it depends on the new API that is not available through nuget packages yet. The changes will be rolled back, and pushed in a scope of a separate PR after CopilotChatApi project references the new nuget packages.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
```diff
@@ -5,117 +5,114 @@
...
-            httpClient ?? kernel.Config.HttpHandlerFactory.CreateHttpClient(kernel.Log),
-            logger ?? kernel.Log);
-
-        config.AddTextCompletionService(Factory, serviceId);
+        builder.WithAIService<ITextCompletion>(serviceId, (logger) =>
+            new AzureTextCompletion(
+                deploymentName,
+                endpoint,
+                apiKey,
+                httpClient,
```
lemillermicrosoft, line 48: `HttpHandlerFactory` -- what's the plan here?

dotnet/src/Connectors/Connectors.AI.OpenAI/Connectors.AI.OpenAI.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\SemanticKernel.Abstractions\SemanticKernel.Abstractions.csproj" />
+    <ProjectReference Include="..\..\SemanticKernel\SemanticKernel.csproj" />
```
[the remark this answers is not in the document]
shawncal: Yeah, it's required to create extension methods for the KernelBuilder class, which is a concrete class in the Core project. 
I'd like to figure out an abstraction that relaxes this requirement, but I'm conflicted as I don't think we should have an IKernelBuilder interface.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
shawncal: That's on @SergeyMenshykh's plate next. In any case, I expect these params will remain, as one must be able to override any default httpClient on a per-connector basis (i.e. maybe Azure OpenAI requires special access credentials, but OpenAI does not).

SergeyMenshykh: Thanks Lee, I returned HttpHandlerFactory usage because we need default http client. The code might change a little bit or might not in a scope of the HTTP stack alignment user story and the parameters will remail as Shawn said.

dotnet/src/Connectors/Connectors.AI.OpenAI/Connectors.AI.OpenAI.csproj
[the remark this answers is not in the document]
SergeyMenshykh: Agree. Ideally, we should keep Connectors and Kernel projects unaware of each other. 
The fastest way, but not the most elegant one, would be to create  IKernelBuilder interface and put it to the Kernel.Abstraction project. It will allow all the other project add extensions methods to the KernelBuilder witout referencing the Kernel project. 
Alternatively, we could consider having a project/package that knows about both Kernel and 
 Connectors projects and contains all the extension methods.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
lemillermicrosoft: These changes will make `OpenAIHttpRetryPolicyTestAsync` fail (and therefore demo scenarios in the kernel syntax examples and elsewhere). I'm not positive why this test was marked as skip (since it was targeting OpenAI not Azure), but alas. I wonder if this should be staged somewhere before the follow up work is defined.

lemillermicrosoft: #1096 in hopes to re-enable those tests.

SergeyMenshykh: I rolled back the changes a few days ago, so the code doesn't have anything that should break the tests, at least I don't see. I'll reach you to get more details about it and discuss it. 
Regarding the retry for OpenAI, I plan to inject HttpClient instance with common retry handler to the OpeanAI connectors in scope of the Http stack improvements, so, if nothing changes during code review, the retry will be supported when the HTTP story is closed.

lemillermicrosoft (dismissed):
Thanks Sergey. @shawncal will leave to you for final review/merge.

shawncal (approved):

PR #938, opened 2023-05-11
Update System.Text.Json and enable transitive pinning

lemillermicrosoft:
This commit updates the System.Text.Json package version to 7.0.0 and enables the CentralPackageTransitivePinningEnabled property in the Directory.Packages.props file. This property allows the central package management feature to pin the versions of transitive dependencies to the ones specified in the props file, avoiding potential conflicts or inconsistencies. This change improves the reliability and maintainability of the dotnet projects that use the Directory.Packages.props file.

The test ReadFromJsonThrowsWithInvalidJson was passing an invalid JSON string to the JsonSerializer.Deserialize method, which was expected to throw a JsonException. However, the method actually throws an ArgumentNullException in System.Text.Json 7.0.0 when the named property is missing. This commit fixes the test by updating the expected exception type. It also adds a pragma directive to suppress the JSON001 warning for the invalid JSON string.

[Central Package Management | Microsoft Learn](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management)
[Update dependencies | Docs - Microsoft Open Source](https://docs.opensource.microsoft.com/tools/cg/how-to/updating-dependencies/#nuget)

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows SK Contribution Guidelines (https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md)
- [x] The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #1004, opened 2023-05-15
Adding UnitTest tags for dotnet 6,7,8 version support

shawncal:
### Motivation and Context
Unit tests are currently targeting net6.0 and the UTs will only run on machines with net6.0 SDK installed. That is, if a machine has only net7.0 or net8.0, installed, the UT projects will build, but the tests will not run.

### Description
With this change, the UT projects self-identify as test projects with the tag:
`<IsTestProject>true</IsTestProject>`

...and we add the RollForward tag which ensures the UTs build for the highest installed version of the SDK:
`<RollForward>LatestMajor</RollForward>`

This ensures that the tests are runnable on any installation net6.0 or greater, including the new docker-based CI builds, each of which targets only a single SDK version.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.UnitTests/Connectors.UnitTests.csproj
+++ b/dotnet/src/Connectors/Connectors.UnitTests/Connectors.UnitTests.csproj
@@ -4,6 +4,8 @@
     <AssemblyName>SemanticKernel.Connectors.UnitTests</AssemblyName>
     <RootNamespace>SemanticKernel.Connectors.UnitTests</RootNamespace>
     <TargetFramework>net6.0</TargetFramework>
+    <RollForward>LatestMajor</RollForward>
+    <IsTestProject>true</IsTestProject>
     <Nullable>enable</Nullable>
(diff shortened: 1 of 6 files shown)


lemillermicrosoft (approved):

lemillermicrosoft:
I'm guessing the UT failure is similar to the change that was required when we tried transitive pinning.

https://github.com/microsoft/semantic-kernel/pull/938/files#diff-69e3fabad2e4646e9bd7d79e0cc573d286fb811afce4bf9084c84a04e357e27d

shawncal:
> I'm guessing the UT failure is similar to the change that was required when we tried transitive pinning.
> 
> https://github.com/microsoft/semantic-kernel/pull/938/files#diff-69e3fabad2e4646e9bd7d79e0cc573d286fb811afce4bf9084c84a04e357e27d

Yup, it's a dotnet 7 failure. I can repro on net7 build. I'll change to "draft" until I can fix it.

lemillermicrosoft (approved):

PR #1007, opened 2023-05-15
Adding support for JSON responses with embedded results

teresaqhoang:
### Motivation and Context

This PR add an API response optimization case where the key itself represents an embedded answer within a JSON response.

If the JSON object returned contains only one property, the value of that property will be used for truncation. The key of the property is concatenated to the truncated result to maintain context.


samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
```diff
@@ -534,26 +534,32 @@ private string OptimizeOpenApiSkillJson(string jsonContent, int tokenLimit, Plan
...
+                jsonTokenLimit -= Utilities.TokenCount(firstProperty.Name);
+                resultsDescriptor = string.Format(CultureInfo.InvariantCulture, "{0}: ", firstProperty.Name);
+
+                // Extract object to be truncated
+                JsonElement value = firstProperty.Value;
+                document = JsonDocument.Parse(value.GetRawText());
             }
         }
 
+        // START Truncation
```
adrianwyatt, line 561: Could you add more context to this comment (or remove it)?

adrianwyatt, line 561: Could you add more context to this comment (or remove it)?

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/Skills/ChatSkill.cs
teresaqhoang: Removed it

adrianwyatt (approved):

PR #1029, opened 2023-05-17
Copilot log out

teresaqhoang:
### Motivation and Context

This PR adds simple User Settings menu on Avatar click and adds option to "Log Out"
image


adrianwyatt (approved):

amsacha (approved):

samples/apps/copilot-chat-app/webapp/src/components/header/UserSettings.tsx
```diff
@@ -0,0 +1,66 @@
...
+                    <MenuItem className={classes.persona}>
+                        <Persona
+                            name={account?.name}
+                            secondaryText={account?.username}
+                            presence={{ status: 'available' }}
+                            avatar={{ color: 'colorful' }}
+                        />
+                    </MenuItem>
+                    <MenuDivider />
+                    <MenuItem onClick={async () => await AuthHelper.logoutAsync(instance)}>Log Out</MenuItem>
```
TaoChenOSU, line 61: Do we have to clean up the app stores?

teresaqhoang: No, Login and Logout both use redirect, so the app state is reloaded on each redirect

I'll test this more in depth to verify

Issue #1047, opened 2023-05-17
dotnet sample 06-memory-and-embeddings.ipynb assumes model name text-embedding-ada-002


andhesky:
**Describe the bug**
AzureOpenAI requires you to deploy models and you can provide a custom name. If there is no model deployed with the name "text-embedding-ada-002" then the "Manually adding memories" section in the dotnet sample 06-memory-and-embeddings.ipynb will fail with the error

```
Error: Microsoft.SemanticKernel.AI.AIException: InvalidRequest: The request is not valid, HTTP status: 404
at Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk.ClientBase.RunRequestAsync[T](Func`1 request)
at Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk.ClientBase.InternalGenerateTextEmbeddingsAsync(IList`1 data, CancellationToken cancellationToken)
at Microsoft.SemanticKernel.AI.Embeddings.EmbeddingGenerationExtensions.GenerateEmbeddingAsync[TValue,TEmbedding](IEmbeddingGeneration`2 generator, TValue value, CancellationToken cancellationToken)
at Microsoft.SemanticKernel.Memory.SemanticTextMemory.SaveInformationAsync(String collection, String text, String id, String description, String additionalMetadata, CancellationToken cancel)
at Submission#15.<<Initialize>>d__0.MoveNext()
--- End of stack trace from previous location ---
at Microsoft.CodeAnalysis.Scripting.ScriptExecutionState.RunSubmissionsAsync[TResult](ImmutableArray`1 precedingExecutors, Func`2 currentExecutor, StrongBox`1 exceptionHolderOpt, Func`2 catchExceptionOpt, CancellationToken cancellationToken)
```

**To Reproduce**
0. Configure settings as instructed in 0-AI-settings.ipynb. Connect to an Azure Open AI endpoint with an embedding model defined using a name other than "text-embedding-ada-002"
1. Go to '06-memory-and-embeddings.ipynb'
2. Click on 'run for the first three sections'

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - IDE: VS Code
 - NuGet Package Version 0.12.207.1-preview

**Additional context**

craigomatic:
@alexchaomander

alexchaomander:
[...]

PR #1048, opened 2023-05-17
Fix dotnet sample six

andhesky:
### Motivation and Context
 #1047

### Description
Fix issue with sample 06-memory-and-embeddings.ipynb where it assumes the name of the azure model is always text-embedding-ada-002.
Could be cleaned up by including the embedding model in the public LoadFromFile return value, but this would require updating all samples so a new method was added for now LoadFromFileWithEmbeddingModel

Also reverse the order of asking for ApiKey and Model in 0-AI-settings - This means an Azure user can fill out all settings from a single page then move on to another page. It does not impact open ai users

### Contribution Checklist

- [] All unit tests pass, and I have added new tests where possible


samples/notebooks/dotnet/06-memory-and-embeddings.ipynb
```diff
@@ -48,6 +48,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "This sample will use an embedding model. If you are using AzureOpenAI you will need to specify the name of a deployment of an [embedding model](https://learn.microsoft.com/azure/cognitive-services/openai/concepts/models#embeddings-models).\n",
```
dluc, line 51: typo
```suggestion
    "This sample will use an embedding model. If you are using Azure OpenAI you will need to specify the name of a deployment of an [embedding model](https://learn.microsoft.com/azure/cognitive-services/openai/concepts/models#embeddings-models).\n",
```

dluc (commented):
Could you update also the template files under `config` folder? the one for OpenAI can be prepopulated

lemillermicrosoft:
Thanks for your contribution! We're moving fast, and there's already a new package out, that will be updated to in #1089. Please re-open as appropriate if anything was missed.

Issue #970, opened 2023-05-12
IMemoryStore.RemoveAsync


EashShow:
**Describe the bug**
An error will occur when call IMemoryStore.RemoveAsync

**Expected behavior**
Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Microsoft.SemanticKernel.Connectors.Memory.Qdrant.Http.ApiSchema.QdrantResponse'. Path: $ | LineNumber: 0 | BytePositionInLine: 1.

**Desktop (please complete the following information):**
 - OS:  Windows
 - IDE: Visual Studio 2022
 - NuGet Package Version 0.13.442.1-preview

**Additional context**
the vector database is Qdrant

craigomatic:
@tawalke any thoughts here?

EashShow:
Let me fix this?

craigomatic:
Closing as we ended up fixing this over at #1025 - thank you for the PR in any case :)

Sam--Shan:
![image](https://github.com/microsoft/semantic-kernel/assets/10590637/2dabfc96-74c2-43b8-b1cb-e67f2908c708)
Hi @craigomatic, I am still encountering an error with version 0.19.230804.2-preview in NuGet. Has the fix been released for this version?

craigomatic:
> ![image](https://user-images.githubusercontent.com/10590637/259907514-2dabfc96-74c2-43b8-b1cb-e67f2908c708.png) Hi @craigomatic, I am still encountering an error with version 0.19.230804.2-preview in NuGet. Has the fix been released for this version?

Hmm, it was fixed in a prior version, maybe a regression has snuck in. If you are using Qdrant, internally it's doing deserialisation via https://github.com/microsoft/semantic-kernel/blob/4ea1d48dfcbe0d1594351b6d946d60f2b9dd60a6/dotnet/src/Connectors/Connectors.Memory.Qdrant/QdrantVectorDbClient.cs#L203

Would you mind opening a new issue and we will investigate?

PR #1057, opened 2023-05-18
fix the serialize bug when delete a Qdrant point

EashShow:
### Motivation and Context

fixed the bug #970 

### Contribution Checklist


EashShow:

dotnet/src/Connectors/Connectors.Memory.Qdrant/QdrantMemoryStore.cs
```diff
@@ -376,6 +398,12 @@ public async Task RemoveWithPointIdBatchAsync(string collectionName, IEnumerable
         return (record.Item1, record.Item2);
     }
 
+    /// <inheritdoc/>
```
lemillermicrosoft, line 401: This isn't part of the `IMemoryStore` interface.

lemillermicrosoft (changes requested):
This seems to be doing more than fixing  bug, and it's hard to see where any bug fix is happening. The UT added doesn't do any deletion of points. It seems like it's more about adding a new method ListPoints that should likely be part of the IMemoryStore interface and deserves a full proposal.

Poltuu:
@EashShow would you mind splitting this PR in two parts, the part fixing the deserialization bug, and the other one adding a method to the `IQdrantVectorDbClient` interface ? Your fix would be dearly appreciated here 🙏

EashShow:
> @EashShow would you mind splitting this PR in two parts, the part fixing the deserialization bug, and the other one adding a method to the `IQdrantVectorDbClient` interface ? Your fix would be dearly appreciated here 🙏

Yes, no problem

dotnet/src/Connectors/Connectors.Memory.Qdrant/QdrantMemoryStore.cs
EashShow: This is about the #1057

rliberoff:
When can we expect a new version of the NuGet packages?

This is something that we are needing.

shawncal:
The original bug #970 has already been fixed, and what remains here is mostly documentation and additional functionality. @EashShow closing the PR, but if you feel the other changes are valuable, please re-submit with description of added features and new tests.

PR #1060, opened 2023-05-18
Improvements to SKFunction interface

dluc:
* Allow to call functions passing a string. This is a common scenario, particularly with demos, allowing for a shorter syntax. The code logs a warning if the input string exists also in the context and is different.
* Add SKContext.Clone and delete unnecessary/redundant SKFunction.InvokeWithCustomInput method.
* Allow to create a context with a non empty cancellation token. This can be useful to inject an existing token from outside.
* Fix SKFunction invoke, allow to pass memory reference.


dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
```diff
@@ -57,19 +58,23 @@ public interface ISKFunction
...
     /// <param name="input">String input</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <param name="settings">LLM completion settings (for semantic functions only)</param>
+    /// <param name="memory">Semantic memory</param>
     /// <param name="logger">Application logger</param>
-    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The updated context, potentially a new one if context switching is implemented.</returns>
     Task<SKContext> InvokeAsync(
         string? input = null,
+        CancellationToken cancellationToken = default,
```
stephentoub, line 73: Typically CancellationToken comes last in the arguments list.

dotnet/src/SemanticKernel/Orchestration/CancellationTokenExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.SemanticKernel.Orchestration;
+
+#pragma warning disable IDE0130
+// ReSharper disable once CheckNamespace - Using the namespace of CancellationToken
+namespace System.Threading;
+#pragma warning restore IDE0130
+
+public static class CancellationTokenExtensions
```
stephentoub, line 10: If we want to use these helpers internally in SK, that's fine and we can make these internal.  But I don't think SK should be adding such non-SK-related helpers to its public surface area.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -301,22 +302,26 @@ public FunctionView Describe()
     /// <inheritdoc/>
     public Task<SKContext> InvokeAsync(
         string? input = null,
+        CancellationToken cancellationToken = default,
```
stephentoub, line 305: Same note about CT typically coming last.  It's more of an issue when there aren't so many optional parameters, but even so it's relevant.  Devs are used to seeing it last in docs, and there's tooling that expects it to be last, e.g. analyzers that will look at the last parameter to see whether it's a CT and act accordingly.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -199,12 +200,14 @@ public FunctionView Describe()
     /// <inheritdoc/>
     public Task<SKContext> InvokeAsync(
         string? input = null,
+        CancellationToken cancellationToken = default,
```
stephentoub, line 203: Ditto for all of these.  I'll stop commenting on them.

dotnet/src/SemanticKernel/Orchestration/CancellationTokenExtensions.cs
shawncal: @dluc if these are handy helpers that we want to be "internal" but available throughout our projects, you can drop them in this folder where they'll be included as source (not as project references): dotnet\src\InternalUtilities

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -135,4 +135,24 @@ public override string ToString()
     {
         return this.ErrorOccurred ? $"Error: {this.LastErrorDescription}" : this.Result;
     }
+
+    /// <summary>
+    /// Create a clone of the current context, using the same kernel references (memory, skills, logger)
+    /// and a new set variables, so that variables can be modified without affecting the original context.
+    /// </summary>
+    /// <returns>A new context copied from the current one</returns>
+    public SKContext Clone()
```
shawncal, line 144: If it's useful, we could implement [IClonable](https://learn.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-7.0) and use [object.MemberwiseClone](https://learn.microsoft.com/en-us/dotnet/api/system.object.memberwiseclone?view=net-7.0#system-object-memberwiseclone). Same for ContextVariables. <<keep in mind for future change -- don't block this one>>

dluc: thanks - I took a look, the main drawback I guess is that ICloneable.Clone() would require a cast, e.g. require more code. Once we get into distributed workflows, durable functions etc we'll probably look into these and related features.

stephentoub: FWIW, we largely consider `ICloneable` to have been a mistake on our part, and we haven't implemented it on anything new since the early days of .NET.  The primary issue with it is it's underspecified whether it's a shallow or deep clone or somewhere in between, making it difficult to reason about abstractly.

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
```diff
@@ -87,37 +86,51 @@ public static ISKFunction UseFrequencyPenalty(this ISKFunction skFunction, doubl
...
+            {
+                context.Log.LogWarning(
+                    "Function {0}.{1} has been invoked with an explicit input text that is different and overrides the input text defined in the context",
+                    function.SkillName, function.Name);
+            }
         }
 
-        return tmpContext;
+        // Create a copy of the context, to avoid editing the original set of variables
+        SKContext contextClone = context.Clone();
```
shawncal, line 129: Passing a clone of SKContext into the function: this differs from the Invoke* methods in SKFunction and Plan, where the actual context object is passed into the function and updated (mutable). Here, it's always a shallow clone (partially immutable), which introduces some different behavior/expectation of SKContext usage.

dotnet/src/SemanticKernel/Orchestration/CancellationTokenExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// unless they are both not null/default.
+    /// </summary>
+    /// <param name="token">Cancellation token 1</param>
+    /// <param name="secondToken">Cancellation token 2</param>
+    /// <returns>A combination of the two tokens</returns>
+    internal static CombinedCancellationToken CombineWith(
+        this CancellationToken token, CancellationToken secondToken)
+    {
+        return new CombinedCancellationToken(token, secondToken);
+    }
```
shawncal, line 23: I don't think this is used either, and can probably be removed, but if it is, this appears exactly equivalent to:

```suggestion
    internal static CancellationTokenSource CombineWith(
        this CancellationToken token, CancellationToken secondToken)
    {
        return CancellationTokenSource.CreateLinkedTokenSource(token1, token2);
    }
```
...both return an IDisposable object with a `.Token` property.

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
```diff
@@ -87,37 +86,51 @@ public static ISKFunction UseFrequencyPenalty(this ISKFunction skFunction, doubl
...
+    /// <summary>
+    /// Execute a function allowing to pass the main input separately from the rest of the context.
+    /// Note: if the context contains an INPUT key/value, that value is ignored, logging a warning.
+    /// </summary>
+    /// <param name="function">Function to execute</param>
+    /// <param name="input">Main input string</param>
+    /// <param name="context">Execution context, including variables other than input</param>
+    /// <param name="settings">LLM completion settings (for semantic functions only)</param>
+    /// <returns>The result of the function execution</returns>
+    public static Task<SKContext> InvokeAsync(this ISKFunction function,
```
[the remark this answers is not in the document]
dluc: technically yes, but those are quite convoluted ways to pass a string and include the context. In many scenarios there is only one input value, and the context is passed only to share access to Logger, Memory, Skills, Cancellation token. In this case the reasoning is "I'm passing my input, param 1, and access to logger/skills/etc, param 2"

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
```diff
@@ -62,12 +63,14 @@ public interface ISKFunction
...
     /// <param name="input">String input</param>
     /// <param name="settings">LLM completion settings (for semantic functions only)</param>
+    /// <param name="memory">Semantic memory</param>
     /// <param name="logger">Application logger</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The updated context, potentially a new one if context switching is implemented.</returns>
     Task<SKContext> InvokeAsync(
         string? input = null,
         CompleteRequestSettings? settings = null,
+        ISemanticTextMemory? memory = null,
```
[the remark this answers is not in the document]
dluc: in the current design, SKContext contains a Memory object. When calling this method, passing a single input string, one should be allowed to pass the current memory object stored in the kernel/context. This is not forcing to have a single memory, only allowing to carry over the existing memory. Without this, one could create a kernel with a memory instance, call a function passing a string, and the function internally would not find the memory defined earlier, e.g. "recall" would return no data. We had a few of these errors in the past in the notebooks.

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
dluc: This is in line with the other calls, happy to chat about this offline. 

When a function receives in input a context and returns a context, the 2 are not the same. Changing variables in the input-context is different from setting variables in the output-context. In fact, we should make the input-context read-only to avoid similar attempts.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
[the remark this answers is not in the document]
dluc: without this, SKFunction.InvokeAsync is creating a new context that doesn't have the memory configured by the user. For instance:

```
skill = kernel.importSkill(a skill that uses memory)
skill[functionname].invoke("this  content")
```

^^^ this will fail because the skill is designed to inject memory on call, not on ctor

And the change allows to do this:
```
memory1 = memory about me
memory2 = memory about someone

skill = kernel.importSkill(a skill that uses memory)
skill[functionname].invoke("this  content", memory1)
skill[functionname].invoke("this  content", memory2)
```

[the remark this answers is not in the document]
shawncal: gotcha. Makes sense, thank you

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
dluc: pls see `ContextSwitchInSKContextOutTaskSKContext`

```
context = await callable(context).ConfigureAwait(false);
return context;
```

the context returned is not the mutable context sent in, but the context returned by `callable`

dluc: I added a parameter to allow both behaviors, so one can run the function with immutable state, e.g. to compare results. The default behavior is allowing functions to change it (mutable context)

shawncal (approved):

shawncal (approved):

PR #1066, opened 2023-05-18
Python: integration test improvements

awharrison-28:
### Motivation and Context
This PR slims down the number of integration tests running against AOAI and OAI models - reducing time and resources spent running them. Additional justification for slimming down these tests is to reduce the chance of throttling from these endpoints (leads to unstable integration tests).

The scenarios previously being covered are already handled using HF models which are much less expensive to test against and do not run the risk of throttling.

### Description
- added a top-level test `conftest.py` to handle kernel creation and OpenAI model secret handling. I had originally intended `create_kernel` to do more than just create a kernel, but additional setup wasn't needed. I've left the pytest fixture though since other fixtures can call it, and using it in tests can making importing Kernel from semantic_kernel unncessary. 
- added a completions `conftest.py` to set up completions tests. For example, setup_hf_text_completion_function allows for testing text2text_generation and text_generation models using the same test file. 
- Common test code is now in pytest fixtures instead of common methods.
- For a number of the completion tests, I have broken out the asserts to individual tests instead of running one giant test. This makes it easier to identify regressions in specific patterns around invoking skills.
- Added retry logic to conversationSummarySkill
- renamed tests to be more descriptive. 

### Contribution Checklist


alexchaomander (approved):
I like the idea of using more local models to do these tests so that we don't run into issues of the APIs being down or throttling.

python/tests/integration/completions/conftest.py
```diff
@@ -0,0 +1,151 @@
...
+        Whales are fully aquatic, open-ocean animals:
+        they can feed, mate, give birth, suckle and raise their young at sea.
+        Whales range in size from the 2.6 metres (8.5 ft) and 135 kilograms (298 lb)
+        dwarf sperm whale to the 29.9 metres (98 ft) and 190 tonnes (210 short tons) blue whale,
+        which is the largest known animal that has ever lived. The sperm whale is the largest
+        toothed predator on Earth. Several whale species exhibit sexual dimorphism,
+        in that the females are larger than males.
+    """
+    additional_text = """
+        The word "whale" comes from the Old English hwæl, from Proto-Germanic *hwalaz,
```
mkarle, line 51: I love that I can learn things from our PRs.

python/tests/integration/embeddings/test_hf_embedding_service.py
```diff
@@ -20,8 +17,54 @@ async def test_hf_embeddings_with_memories():
...
+
+    # Search for documents
+    query = "What are mammals?"
+    result = await kernel.memory.search_async(
+        "test", query, limit=2, min_relevance_score=0.0
+    )
+    print(f"Query: {query}")
+    print(f"\tAnswer 1: {result[0].text}")
+    print(f"\tAnswer 2: {result[1].text}\n")
+    assert "mammals." in result[0].text
```
mkarle, line 45: Nit: This doesn't really assert that it read the document about whales and dolphins.

PR #1067, opened 2023-05-18
UX: Fix preview wrap in chatlistitem

Vivihung:
### Motivation and Context
The preview text wrap causes the ChatListItem changes its height if the latest message is longer than the previous one. It results in instability of the UI.

image

### Description
* Prevent the preview text from wrapping by setting the display to 'block' and applying 'no-wrap' and ellipsis
* Remove styles that are not in-use.

After the fix:
image

### Contribution Checklist

- ~[ ] The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`~
- ~[ ] All unit tests pass, and I have added new tests where possible~


teresaqhoang (approved):

hathind-ms:
Can we also set min width on chat list item as well?

teresaqhoang:
> Can we also set min width on chat list item as well?

@hathind-ms it's already set in the root styles

adrianwyatt (approved):

PR #1070, opened 2023-05-18
Frontloading chat on UploadBot + fixing chatlist timestamps

teresaqhoang:
### Motivation and Context

This PR fixes a couple bugs. Now.
- On UploadBot, new chats are automatically opened and selected at the top of chat list
- Timestamps in chatlist show date if not same day
- Fixed ChatHistory user rendering 
image

### Description

Other changes
- Setting selectedId on addConversation so dev doesn't have to remember to do this every time
- bot/upload returns 201 Create with new chat session details (previously 202 Accepted) 
- Reverse list of chats on API response instead of in loadChats

### Contribution Checklist


hathind-ms (approved):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
```diff
@@ -135,6 +140,7 @@ public class BotController : ControllerBase
     /// <returns>The serialized Bot object of the chat id.</returns>
     [Authorize]
     [HttpGet]
+    [ActionName("DownloadAsync")]
```
Vivihung, line 143: Do we need to set ActionName while there is Route?

Vivihung (approved):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
```diff
@@ -64,14 +64,14 @@ public class BotController : ControllerBase
     /// <param name="kernel">The Semantic Kernel instance.</param>
     /// <param name="userId">The user id.</param>
     /// <param name="bot">The bot object from the message body</param>
-    /// <returns>The HTTP action result.</returns>
+    /// <returns>The HTTP action result with new chat session object.</returns>
     [Authorize]
     [HttpPost]
     [Route("bot/upload")]
-    [ProducesResponseType(StatusCodes.Status202Accepted)]
+    [ProducesResponseType(StatusCodes.Status201Created)]
```
adrianwyatt, line 71: Yay!

amsacha (dismissed):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
```diff
@@ -124,7 +125,11 @@ public class BotController : ControllerBase
             throw;
         }
```
adrianwyatt, line 126: Should we just remove this catch/throw?

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
teresaqhoang: Yeah originally I was going to pass DownloadAsync into CreatedAtAction in bot/upload so this attribute was required for correct action reference, but since I opted to pass the ChatHistoryController.GetChatSessionByIdAsync action instead, we technically don't need this. 

But I left it there because it was a pain to debug and will be helpful if someone wants to reference this method as an action in the future. Without the attribute explicitly defined like this, the method wasn't being discovered correctly, and led to this error: Error: 500: => System.InvalidOperationException: No route matches the supplied values.

adrianwyatt (approved):

Issue #794, opened 2023-05-03
The data type of the `Id` member in the `ScoredPoint` class may need to be changed to `int`.


AwesomeYuer:
**Describe the bug**
The data type of the `Id` member in the `ScoredPoint` class may need to be changed to `int`.

https://github.com/microsoft/semantic-kernel/blob/562ab5bc63ec4043fe6c797145e326934ce61639/dotnet/src/Connectors/Connectors.Memory.Qdrant/Http/ApiSchema/SearchVectorsResponse.cs#L14

**To Reproduce**
```csharp

    public async Task qdrant_SK_Http_HNSW_index_Cosine_50_ProcessAsync()
    {
        QdrantMemoryStore memoryStore =
                new QdrantMemoryStore
                            (
                                GlobalManager.SelfHostQdrantHttpConnectionString
                                , 6333
                                , vectorSize: 4
                                //, ConsoleLogger.Log
                            );
        IKernel kernel = Kernel.Builder
            //.WithLogger(ConsoleLogger.Log)
            .Configure
            (
                c =>
                {
                    c.AddOpenAITextCompletionService("text-davinci-003", "123");
                    c.AddOpenAITextEmbeddingGenerationService("text-embedding-ada-002", "123");
                }
            )
            .WithMemoryStorage(memoryStore)
            .Build();

        var MemoryCollectionName = "small";
       //bug here!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        var searchResults = kernel
                                .Memory
                                .SearchAsync
                                        (
                                            MemoryCollectionName
                                            , "My favorite color is orange"
                                            , limit: 20
                                            , minRelevanceScore: 0.8
                                            , withEmbeddings: true
                                        );

        await foreach (var item in searchResults)
        {
            //Console.WriteLine(item.Metadata.Text + " : " + item.Relevance);
            _ = item.Relevance;
        }
    }

```

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/1026479/236021740-a0f5fd2b-0886-4ddf-bf77-f540a5da45bf.png)

![image](https://user-images.githubusercontent.com/1026479/236020588-f8d2d49a-4a6f-4906-b4c2-4f48d3471234.png)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]
 - IDE: [e.g. Visual Studio, VS Code]
 - NuGet Package Version [e.g. 0.1.0]

[...]

PR #1075, opened 2023-05-18
Fix bug preventing Qdrant points with numeric id from being consumed

craigomatic:
### Motivation and Context
Qdrant supports both string and int for the id of a ScoredPoint. This minor change allows the serialiser to handle both of these scenarios while retaining string as the type for SK.

Thank you to @AwesomeYuer for the heads up in #794 

### Description
Added a single attribute decorating the attribute that massages ints to string and leaves string as-is.

### Contribution Checklist


tawalke (approved):
I think this is an effective and efficient solution for this!

AwesomeYuer:
glad to hear that

EashShow:
when call the function "kernel.Memory.SaveInformationAsync", the ID is assigned to a GUID,
then call the function "IQdrantVectorDbClient.GetVectorByPayloadIdAsync", an error will occur

Starting in .NET 6, if you apply JsonNumberHandlingAttribute to a property that's a collection of non-number values and attempt to serialize or deserialize the property, an InvalidOperationException is thrown.

PR #1084, opened 2023-05-18
Fix for CopilotChat error: "Unable to load chats. Details: interaction_in_progress: Interaction is currently in progress." 

gitri-ms:
### Motivation and Context

Issue Banner on Co-pilot start up: "Unable to load chats. Details: interaction_in_progress: Interaction is currently in progress." 

This error occurs when another MSAL interactive API call is currently in progress -- e.g., we are requesting a token before the previous token request has completed. Per the [MSAL React documentation](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/docs/errors.md#interaction_in_progress), the solution is to check if another interaction is in progress and wait until it's done. We already do this when acquiring tokens for skills (in `TokenHelper.ts`), so we should use the same code to acquire the token for the SKaaS backend.

### Description

- Pipe MSAL's `inProgress` indicator into all calls to `AuthHelper.getSKaaSAccessToken()`
- Change `AuthHelper.getSKaaSAccessToken()` to call `TokenHelper.getAccessTokenUsingMsal()`, which already handles the `interaction_in_progress` error

### Contribution Checklist


teresaqhoang (approved):

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -63,7 +63,7 @@ interface ChatInputProps {
 export const ChatInput: React.FC<ChatInputProps> = (props) => {
     const { isTyping, onSubmit } = props;
     const classes = useClasses();
-    const { instance, accounts } = useMsal();
+    const { instance, accounts, inProgress } = useMsal();
     const account = useAccount(accounts[0] || {});
```
teresaqhoang, line 67:
```suggestion
    const account = instance.useActiveAccount();
```
nit

hathind-ms (approved):

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
hathind-ms: @teresaqhoang , @gitri-ms - let's file separate bug for this suggestion.

gitri-ms: I agree, but it looks like there are several places we are doing this. Let me do it as a separate PR so I can make sure I don't miss any without holding this one up.

PR #1085, opened 2023-05-18
Fix MIME type console error and enable WebApp production build

teresaqhoang:
### Motivation and Context

This PR fixes the MIME Type error we were seeing in the console and adds details for running a production build.

Currently, we have the WebApp wrapped in React.StrictMode, which enables additional development behaviors, including rendering the App.tsx component twice to find bugs caused by impure rendering. Since StrictMode is development only, serving a production build of the WebApp avoids double rendering the app, so wanted to provide it as an option for developers.

### Description

MIME type error we were observing: 
image

### Contribution Checklist


glahaye (approved):

Vivihung (approved):
Thanks for the changes!

Issue #568, opened 2023-04-21
Suggested tweaks to Embedding types


stephentoub:
Currently, Semantic Kernel contains:
```C#
// *** In Microsoft.SemanticKernel.dll ***
namespace Microsoft.SemanticKernel.AI.Embeddings
{
    public static class SupportedTypes
    {
        public static bool IsSupported<T>();
        public static bool IsSupported(Type type);
        public static IEnumerable<Type> Types { get; }
    }

    public ref struct EmbeddingSpan<TEmbedding>
    {
        public static bool IsSupported { get; }
        ...
    }

    public ref struct EmbeddingReadOnlySpan<TEmbedding>
    {
        public static bool IsSupported { get; }
        ...
    }
    ...
}

// *** In Microsoft.SemanticKernel.Abstractions ***
namespace Microsoft.SemanticKernel.AI.Embeddings
{
    public readonly struct Embedding<TEmbedding>
    {
        ...
        public static bool IsSupported { get; }
    }

    public static class Embedding
    {
        public static readonly Type[] SupportedTypes { get; }
    }
}
```
This is duplicative.  The `Embedding.SupportedTypes` property is also returning an array that can be mutated without requiring any casting.  I propose:
- Remove the `IsSupported` properties from `Embedding<T>`, `EmbeddedSpan<T>`, and `EmbeddedReadOnlySpan<T>`.
- Move the contents of the `SupportedTypes` class to the `Embedding` class, while also renaming the `Types` property to be `SupportedTypes` and deleting the one that's currently on `Embedding`.
- Delete the `SupportedTypes` class.

That would result in:
```C#
// *** In Microsoft.SemanticKernel.dll ***
namespace Microsoft.SemanticKernel.AI.Embeddings
{
    public ref struct EmbeddingSpan<TEmbedding>
    {
        ...
    }

    public ref struct EmbeddingReadOnlySpan<TEmbedding>
    {
        ...
    }
    ...
}

// *** In Microsoft.SemanticKernel.Abstractions ***
namespace Microsoft.SemanticKernel.AI.Embeddings
{
    public readonly struct Embedding<TEmbedding>
    {
        ...
    }

    public static class Embedding
    {
        public static bool IsSupported<TEmbedding>();
        public static bool IsSupported(Type type);
        public static IEnumerable<Type> SupportedTypes { get; }
    }
}
```

cc: @dluc, @shawncal

craigomatic:
@awharrison-28

PR #1093, opened 2023-05-19
Revise Embedding types' IsSupported and add vector ownership ctor

stephentoub:
### Motivation and Context

Fixes https://github.com/microsoft/semantic-kernel/issues/568

### Description

Consolidate IsSupported to just being on the static Embedding type. And add a ctor to Embedding that allows transferring an array rather than copying it, then use that ctor in a bunch of places where we deserialize a vector and then immediately hand it off to an embedding. (I think subsequently we should think about getting rid of this ctor and just having the single ctor store rather than copy the original... I don't think the semi-immutability buys us much here.)

### Contribution Checklist


lemillermicrosoft (approved):
LGTM. Added @awharrison-28 for additional review.

shawncal (approved):

PR #1097, opened 2023-05-19
Suppress CA1852 warning in KernelBuilder.csproj

lemillermicrosoft:
The KernelBuilder.csproj file was generating a CA1852 warning, which
suggests using a const instead of a static readonly field. However, this
warning is not applicable for the KernelBuilder class, which uses
reflection to load the kernel assemblies. Therefore, the warning is
suppressed by adding it to the NoWarn property in the project file. This
makes the code analysis more accurate and avoids false positives.

This warning was impacting dotnet-pr builds against 7.0.

### Contribution Checklist


dluc:
I didn't see the warning and the CI pipelines were green, I'm a bit confused about how these errors are showing up

PR #1117, opened 2023-05-19
IChatCompletion Multiple Results / New Interfaces

RogerBarreto:
### Motivation and Context

Add potential for IChatCompletion with multiple results per prompt.

MultiChatResultPR

### Description

- Adds new Interfaces/Abstractions for Chat Results: `IChatResult`, `IChatStreamingResult`, `IChatMessage`
- Adds new `IChatCompletion` interface methods for Streaming and Non-Streaming results
- Moves old `IChatCompletion` interface methods `IAsyncEnumerable<string> GenerateMessageStreamAsync` and `Task<string> GenerateMessageAsync` as extensions for retro compatibility
- Changed `ChatAsText` to use the new multiple results implementations
- Makes `ChatHistory.ChatMessage` & `ChatHistory.AuthorsRole` types obsolete (discouraging usage of those implementation in favor of new `IChatMessage`
- `ChatHistory` abstraction now extends from `List<IChatMessage>`
- Updates all Samples using the new interfaces. 

### Paves the way to ... (Futures)
- Remove `ChatHistory.ChatMessage` inner class
- Remove `ChatHistory.AuthorsRole` enums 
- Make `ChatHistory` an abstract class or an `IChatHistory` interface.
- Rename `ITextCompletionResult` to `ITextResult`
- Rename `ITextCompletionStreamingResult` to `ITextStreamingResult`

### Contribution Checklist


stephentoub:
> Moves old IChatCompletion interface methods IAsyncEnumerable<string> GenerateMessageStreamAsync and Task<string> GenerateMessageAsync as extensions for retro compatibility

What's our plan for tracking / removing such temporary measures?  Do we actually need them?

RogerBarreto:
> What's our plan for tracking / removing such temporary measures? Do we actually need them?

@stephentoub The plan here is to keep those as extensions while we still use it as part of the `SK internal implementations`. 

Once we moved all the `SK` implementations to use the `Result` approach those extension methods will be flagged as `Obsolete` and removed from the extensions or being kept depending on the community request.

lemillermicrosoft (dismissed):

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatCompletionStreamingResult.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// <inheritdoc/>
+    public async Task<IChatMessage> GetChatMessageAsync(CancellationToken cancellationToken = default)
+    {
+        ChatMessage? chatMessage = null;
+        await foreach (var message in this._choice.GetMessageStreaming(cancellationToken))
+        {
+            // As it loops thru the stream, the return ChatMessage is an updated/concatenated version of the previous ChatMessage
+            chatMessage = message;
```
NTaylorMullen, line 30: This keeps replacing it, not intended I assume?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -316,22 +297,29 @@ private static ChatCompletionsOptions CreateChatCompletionsOptions(ChatRequestSe
...
         }
 
         return options;
     }
 
+    private static void ValidateChatAuthors(string role, out ChatRole validRole)
+    {
+        Verify.NotNullOrWhiteSpace(role);
+
+        validRole = role.ToUpperInvariant() switch
```
NTaylorMullen, line 314: This is an odd way to do an `if` statement with multiple `string.Equals(...., StringComparison.OrdinalIgnoreCase)`

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatRequestSettings.cs
```diff
@@ -42,6 +42,13 @@ public class ChatRequestSettings
     /// </summary>
     public IList<string> StopSequences { get; set; } = Array.Empty<string>();
 
+    /// <summary>
+    /// How many completions to generate for each prompt. Default is 1.
+    /// Note: Because this parameter generates many completions, it can quickly consume your token quota.
+    /// Use carefully and ensure that you have reasonable settings for max_tokens and stop.
+    /// </summary>
+    public int ResultsPerPrompt { get; set; } = 1;
```
NTaylorMullen, line 50: Azure SDK dropped the ball on this one. In my opinion this should just be `N`. It's what all the core APIs reference / document

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/IChatMessage.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.AI.ChatCompletion;
+
+public interface IChatMessage
```
NTaylorMullen, line 5: As a general note for API design in .NET: Utilizing interfaces gives an "I can do" methodology where as things like abstract classes represent an "I am". In addition, once SK goes GA you'll be bound to these interfacers and if you ever want to change them you'll need to make an `IChatMessage2` in order to expand. Alternatively, if you use an abstract class you can simply add new members as virtual.

Gist: Interfaces have a place in .NET but have a lot of flaws. For instance, all of Roslyn and a lot of ASP.NET migrated away from interface representations specifically for this reason.

samples/dotnet/kernel-syntax-examples/Example30_ChatWithPrompts.cs
```diff
@@ -106,10 +107,10 @@ public static async Task RunAsync()
         // The full chat history. Depending on your scenario, you can pass the full chat if useful,
         // or create a new one every time, assuming that the "system message" contains all the
         // information needed.
-        var chatHistory = chatGPT.CreateNewChat(systemMessage);
+        var chatHistory = (OpenAIChatHistory)chatGPT.CreateNewChat(systemMessage);
```
NTaylorMullen, line 110: Why do we need to cast to `OpenAIChatHistory`?

samples/dotnet/kernel-syntax-examples/Example40_MultiChatCompletion.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        Console.WriteLine("======== Azure OpenAI - Multiple Chat Completion ========");
+
+        IKernel kernel = new KernelBuilder().WithLogger(ConsoleLogger.Log).Build();
+        kernel.Config.AddAzureChatCompletionService(
+            Env.Var("AZURE_OPENAI_CHAT_DEPLOYMENT_NAME"),
+            Env.Var("AZURE_OPENAI_ENDPOINT"),
+            Env.Var("AZURE_OPENAI_KEY"));
+
+        IChatCompletion completion = kernel.GetService<IChatCompletion>();
```
NTaylorMullen, line 33: Ya this stinks / breaks down the kernel expectations w/ SK (shouldn't need to pull out the chat completion bit). From a high level I'd imagine services would be valuable purely for skills/connectors etc; however, once you expect an end-user to extract services from the kernel then IMO you're in rough territory / breaking boundaries between what's a service

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatCompletionStreamingResult.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+        // As it leaves the streaming loop, the return ChatMessage is the final version of the ChatMessage to present in a non-streaming behavior
+
+        if (chatMessage is null)
+        {
+            throw new AIException(AIException.ErrorCodes.UnknownError);
```
SergeyMenshykh, line 36: Consider adding a few details describing what happened to simplify debugging/troubleshooting process.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatCompletionStreamingResult.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
+
+internal sealed class ChatStreamingResult : IChatStreamingResult, ITextCompletionStreamingResult
+{
+    private readonly StreamingChatChoice _choice;
+
+    public ChatStreamingResult(StreamingChatChoice choice)
+    {
+        this._choice = choice;
```
SergeyMenshykh, line 20: Should we check the parameter for null? because if it's null neither of the class methods are able to fulfill their contract.

RogerBarreto (commented):
Addressed PR Comments

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatCompletionStreamingResult.cs
RogerBarreto: Good point, added more details to the exception message.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatRequestSettings.cs
RogerBarreto: Not sure what you mean here, but in the docs we make no mention to any upper limit.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -296,16 +270,23 @@ private static CompletionsOptions CreateCompletionsOptions(string text, Complete
         return options;
     }
 
-    private static ChatCompletionsOptions CreateChatCompletionsOptions(ChatRequestSettings requestSettings, ChatHistory chat)
+    private static ChatCompletionsOptions CreateChatCompletionsOptions(ChatRequestSettings requestSettings, IEnumerable<ChatMessageBase> chatHistory)
     {
+        if (requestSettings.ResultsPerPrompt < 1 ||
+            requestSettings.ResultsPerPrompt > 128)
```
shawncal, line 276: tip: new tricks (optional)
```suggestion
        if (requestSettings.ResultsPerPrompt is < 1 or > 128)
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureChatCompletion.cs
```diff
@@ -83,7 +82,7 @@ public ChatHistory CreateNewChat(string? instructions = null)
         CompleteRequestSettings requestSettings,
```
shawncal, line 82: Should CompleteRequestSettings have the same nullability/default rules as ChatRequestSettings?

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureChatCompletion.cs
```diff
@@ -54,21 +53,21 @@ public sealed class AzureChatCompletion : AzureOpenAIClientBase, IChatCompletion
...
 
     /// <inheritdoc/>
-    public IAsyncEnumerable<string> GenerateMessageStreamAsync(
+    public IAsyncEnumerable<IChatStreamingResult> GetStreamingChatCompletionsAsync(
         ChatHistory chat,
         ChatRequestSettings? requestSettings = null,
         CancellationToken cancellationToken = default)
     {
-        return this.InternalGenerateChatMessageStreamAsync(chat, requestSettings ?? new(), cancellationToken);
+        return this.InternalGenerateStreamingChatMessageAsync(chat, requestSettings ?? new(), cancellationToken);
```
shawncal, line 70: The public methods here have a nullable ChatRequestSettings?, but the internal implementations require this arg to be non-null (hence the new() here)... if falling back to a new is valid, I'd rather see the value of requestSettings passed through here (null or not) to the Internal* methods.

Then, only those methods, where the value is actually used, can set it to new() or otherwise ignore it when it's null. Otherwise, we get different fallback behavior across different callers (some will inject new, others will Verify.NotNull() and error out, others may make something else up).

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -203,52 +188,41 @@ await foreach (ChatMessage message in choice.GetMessageStreaming(cancellationTok
...
-
-    /// <summary>
-    /// Creates a completion for the prompt and settings using the chat endpoint
-    /// </summary>
-    /// <param name="text">The prompt to complete.</param>
-    /// <param name="requestSettings">Request settings for the completion API</param>
-    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <returns>Text generated by the remote model</returns>
-    private async Task<string> InternalCompleteTextUsingChatAsync(
+    private protected async Task<IReadOnlyList<ITextCompletionResult>> InternalGetTextCompletionAsChatAsync(
```
shawncal, line 196: I think this should probably be plural `Completions`, as it returns multiple results, and to align with `GetStreamingChatCompletionsAsync`. The public methods appear plural, but internal ones are singular.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -203,52 +188,41 @@ await foreach (ChatMessage message in choice.GetMessageStreaming(cancellationTok
...
         ChatHistory chat = PrepareChatHistory(text, requestSettings, out ChatRequestSettings settings);
 
-        return await this.InternalGenerateChatMessageAsync(chat, settings, cancellationToken).ConfigureAwait(false);
+        return (await this.InternalGenerateChatMessageAsync(chat, settings, cancellationToken).ConfigureAwait(false))
+            .OfType<ITextCompletionResult>()
+            .ToList();
     }
 
-    private IAsyncEnumerable<string> InternalCompleteTextUsingChatStreamAsync(
+    private protected async IAsyncEnumerable<ITextCompletionStreamingResult> InternalGetTextCompletionStreamingAsChatAsync(
```
shawncal, line 208: Naming: I don't thing "AsChat" is quite right here, as that implies that we're casting something to a Chat, but it's actually doing the opposite (chat data "as" text completion result).

Suggestion: Instead of calling the model in these methods, could you just create a helper that takes in an IAsyncEnum and returns the other kind?
`IAsyncEnumerable<ITextCompletionStreamingResult> AsTextCompletionStreamingResult(
      IAsyncEnumerable<IChatCompletionStreamingResult> chatStreamResult
 )`
 
Not sure if this is practical, but it would mean there is only one actual implementation hitting Chat models, and one hitting the Completion models, and then the others are just casting/converting helper methods.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/AuthorRole.cs
```diff
@@ -0,0 +1,95 @@
...
+    public static bool operator ==(AuthorRole left, AuthorRole right)
+    {
+        if (Object.ReferenceEquals(left, null) != Object.ReferenceEquals(right, null))
+        {
+            return false;
+        }
+        else if (Object.ReferenceEquals(left, null))
+        {
+            return true;
+        }
```
shawncal, line 61: I know this is boilerplate, and what I'm about to say matters not one bit... but in the interest of logic puzzles. Is this the same as?:
```suggestion
        if (Object.ReferenceEquals(left, null))
        {
            return Object.ReferenceEquals(right, null);
        }
        
        return left.Equals(right);
```

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/IChatCompletion.cs
```diff
@@ -14,28 +14,28 @@ public interface IChatCompletion : IAIService
...
-    /// <param name="chat">Chat history</param>
-    /// <param name="requestSettings">AI request settings</param>
-    /// <param name="cancellationToken">Async cancellation token</param>
-    /// <returns>Stream the generated chat message in string format</returns>
-    public IAsyncEnumerable<string> GenerateMessageStreamAsync(
+    /// <param name="chat">The chat history context.</param>
+    /// <param name="requestSettings">Request settings for the completion API</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>AsyncEnumerable list of different streaming chat results generated by the remote model</returns>
+    IAsyncEnumerable<IChatStreamingResult> GetStreamingChatCompletionsAsync(
```
shawncal, line 38: I like the new names here!

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/IChatStreamingResult.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Microsoft.SemanticKernel.AI.ChatCompletion;
+
+public interface IChatStreamingResult : IChatResult
+{
+    /// <summary>
+    /// Get the chat message from the streaming result.
+    /// </summary>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Current chat message streaming content</returns>
+    IAsyncEnumerable<ChatMessageBase> GetChatMessageStreamingAsync(CancellationToken cancellationToken = default);
```
shawncal, line 15: Should this not also be in the format "GetStreaming<<things>>Async"?
`GetStreamingChatMessageAsync`

shawncal (commented):
A few comments...

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/AuthorRole.cs
shawncal: (note: the above diff suggestion is missing the final return in the red/removed part)...

 `else
        {
            return left.Equals(right);
        }
    }
    `

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatRequestSettings.cs
NTaylorMullen: Ah I was referring purely to the naming. Aka `N` instead of `ResultsPerPrompt` naming

lemillermicrosoft (dismissed):

lemillermicrosoft (approved):

PR #1119, opened 2023-05-19
CopilotChat email body is the user intent and not the requested body.

gitri-ms:
### Motivation and Context

By design, the planner prioritizes using the provided context variables over whatever is in the plan's state. In this case, the plan's state "input" was the requested body, but it was overridden by the provided context's "input" (user intent). To avoid this, we should call the planner without providing the existing context variables, unless we explicitly want to override what is in the plan's state.

### Description

Pass in `newPlanContext` when invoking the plan, so the planner can have context about memory, skills, etc, but do not include context variables.

### Contribution Checklist


Vivihung (approved):
thanks for the fix!

hathind-ms (approved):

adrianwyatt (approved):

PR #1131, opened 2023-05-20
Add additonal tests

johnoliver:
- Add Example04CombineLLMPromptsAndNativeCodeTest 
- Remove some left over code from previous attempts at immutability in skill/function collections
- Add annotations to allow use of dependency injection

### Contribution Checklist

### DIFF
--- a/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/SKException.java
+++ b/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/SKException.java
@@ -1,4 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 package com.microsoft.semantickernel;
 
-public class SKException extends RuntimeException {}
+public class SKException extends RuntimeException {
+    public SKException(String message) {
+        super(message);
+    }
+}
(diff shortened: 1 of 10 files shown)


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/SKException.java
```diff
@@ -1,4 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 package com.microsoft.semantickernel;
```
markwallace-microsoft, line 3:
```suggestion

/**
 * Provides the base exception from which all Semantic Kernel exceptions derive.
 */
```

java/semantickernel-core/src/main/java/com/microsoft/semantickernel/KernelDefault.java
```diff
@@ -29,29 +35,34 @@
 public class KernelDefault implements Kernel {
```
markwallace-microsoft, line 35: Should be named `DefaultKernel`?

Elsewhere in the Java code class have been named `Default<Foo> implements <Foo>`

PR #1133, opened 2023-05-21
Rewrite how native skill functions are invoked

stephentoub:
### Motivation and Context

Simplify native function invocation and prepare for additional functionality.

### Description

The current approach involves exactly matching the target method against a fixed set of known signatures.  I've revised it to instead be a little more dynamic, such that it instead just looks for valid parameters (e.g. currently up to one string and up to one SKContext) and valid return types (e.g. void, string, context, and tasks of each of those), and any combination of those is valid.  This not only allows the implementation to be simplified and supports combinations that someone could reasonably expect to be valid (e.g. Func(context, string) rather than just Func(string, context)), but it also sets it up to easily allow for additional features to be added, e.g. parameters that match by name against context variables such that any number of parameters could be supported, non-string arguments once those are allowed in context variables, etc.

### Contribution Checklist

cc: @dluc, @shawncal

### DIFF
--- a/dotnet/src/SemanticKernel.UnitTests/SkillDefinition/SKFunctionTests2.cs
+++ b/dotnet/src/SemanticKernel.UnitTests/SkillDefinition/SKFunctionTests2.cs
@@ -17,26 +17,25 @@ public sealed class SKFunctionTests2
     private readonly Mock<IReadOnlySkillCollection> _skills;
 
     private static string s_expected = string.Empty;
-    private static string s_canary = string.Empty;
+    private static string s_actual = string.Empty;
 
     public SKFunctionTests2()
     {
         this._log = new Mock<ILogger>();
         this._skills = new Mock<IReadOnlySkillCollection>();
 
-        s_canary = "";
         s_expected = Guid.NewGuid().ToString("D");
     }
 
     [Fact]
-    public async Task ItSupportsType1Async()
+    public async Task ItSupportsStaticVoidVoidAsync()
     {
         // Arrange
         [SKFunction("Test")]
         [SKFunctionName("Test")]
         static void Test()
         {
-            s_canary = s_expected;
+            s_actual = s_expected;
         }
 
         var context = this.MockContext("");
@@ -48,19 +47,18 @@ public sealed class SKFunctionTests2
 
         // Assert
         Assert.False(result.ErrorOccurred);
-        this.VerifyFunctionTypeMatch(1);
-        Assert.Equal(s_expected, s_canary);
+        Assert.Equal(s_expected, s_actual);
     }
 
     [Fact]
-    public async Task ItSupportsType2Async()
+    public async Task ItSupportsStaticVoidStringAsync()
     {
         // Arrange
         [SKFunction("Test")]
         [SKFunctionName("Test")]
         static string Test()
         {
-            s_canary = s_expected;
+            s_actual = s_expected;
             return s_expected;
         }
 
@@ -73,21 +71,20 @@ public sealed class SKFunctionTests2
 
         // Assert
         Assert.False(result.ErrorOccurred);
-        this.VerifyFunctionTypeMatch(2);
-        Assert.Equal(s_expected, s_canary);
+        Assert.Equal(s_expected, s_actual);
         Assert.Equal(s_expected, result.Result);
         Assert.Equal(s_expected, context.Result);
     }
 
     [Fact]
-    public async Task ItSupportsType3Async()
+    public async Task ItSupportsStaticVoidTaskStringAsync()
     {
         // Arrange
         [SKFunction("Test")]
         [SKFunctionName("Test")]
         static Task<string> Test()
         {
-            s_canary = s_expected;
+            s_actual = s_expected;
             return Task.FromResult(s_expected);
         }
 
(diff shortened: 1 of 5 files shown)


dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -611,158 +409,115 @@ private void EnsureContextHasSkills(SKContext context)
 
         result.Description = skFunctionAttribute?.Description ?? "";
 
-        log?.LogTrace("Method '{0}' found, type `{1}`", result.Name, result.Type.ToString("G"));
+        log?.LogTrace("Method '{0}' found`", result.Name);
```
RogerBarreto, line 412:
```suggestion
        log?.LogTrace("Method '{0}' found", result.Name);
```

dotnet/src/SemanticKernel.UnitTests/SkillDefinition/SKFunctionTests3.cs
```diff
@@ -138,13 +138,8 @@ public void Invalid1(string x, string y)
         {
         }
 
-        [SKFunction("two")]
-        public void Invalid2(SKContext cx, string y)
```
dluc, line 142: nice :-)

dotnet/src/SemanticKernel.UnitTests/SkillDefinition/SKFunctionTests2.cs
```diff
@@ -155,238 +150,268 @@ static string Test(SKContext cx)
...
+        Func<SKContext, string?> method = Test;
+        var function = SKFunction.FromNativeMethod(Method(method), method.Target, log: this._log.Object);
         Assert.NotNull(function);
         SKContext result = await function.InvokeAsync(context);
 
         // Assert
         Assert.False(result.ErrorOccurred);
-        this.VerifyFunctionTypeMatch(5);
-        Assert.Equal(s_expected, s_canary);
+        Assert.Equal(1, invocationCount);
```
dluc, line 183: how come we need to check the count in this and other tests, but not all tests?

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -370,171 +327,6 @@ private void VerifyIsSemantic()
...
-        context.Variables.Update((await callable(this._aiService?.Value, settings, context).ConfigureAwait(false)).Variables);
-        return context;
-    }
-
-    // Run the native function
-    private async Task<SKContext> InvokeNativeAsync(SKContext context)
-    {
-        TraceFunctionTypeCall(this._delegateType, this._log);
-
-        this.EnsureContextHasSkills(context);
```
dluc, line 392: I'll need to check if this is still required

dotnet/src/SemanticKernel.UnitTests/SkillDefinition/SKFunctionTests2.cs
stephentoub: It'd be nice to check in them all, but it's much more complicated to do so in one's that are static and can't close over instance state. It'll be easy to improve once variables can be arbitrary objects.

dotnet/src/SemanticKernel.UnitTests/SkillDefinition/SKFunctionTests3.cs
stephentoub: Locally I have a change I'll put up after this to support arbitrary numbers of parameters, automatically pulling from the context variables.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
stephentoub: I didn't remove the lazy init. There was just a single caller and I moved the line to that location.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -263,20 +269,11 @@ public override string ToString()
...
-        this.ReleaseUnmanagedResources();
-    }
-
     #region private
 
     private static readonly JsonSerializerOptions s_toStringStandardSerialization = new();
     private static readonly JsonSerializerOptions s_toStringIndentedSerialization = new() { WriteIndented = true };
-    private readonly DelegateTypes _delegateType;
-    private readonly Delegate _function;
+    private readonly Func<ITextCompletion?, CompleteRequestSettings?, SKContext, Task<SKContext>> _function;
```
lemillermicrosoft, line 276: I know we're already kind of tied to ITextCompletion, but wondering what this might look like when|if we support IImageGeneration or other IText* variants, as SKFunctions.

stephentoub: This includes ITextCompletion just to be able to unify the semantic and native functions into a single delegate; the native functions ignore the first argument.  This is all a private implementation detail and can evolve to whatever the semantic invocation needs.

lemillermicrosoft (dismissed):

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -245,11 +255,7 @@ public ISKFunction SetAIConfiguration(CompleteRequestSettings settings)
     /// <summary>
     /// Dispose of resources.
     /// </summary>
-    public void Dispose()
-    {
-        this.ReleaseUnmanagedResources();
-        GC.SuppressFinalize(this);
-    }
+    public void Dispose() => (this._aiService as IDisposable)?.Dispose();
```
SergeyMenshykh, line 258: There could be unnecessary initialization of _aiService instance (Lazy<ITextCompletion>) if the being disposed function was created but never invoked.

stephentoub: This apparently changed to be a `Lazy<>` between when I first put this up and now; this code is based on what was there pre-24 hours ago.  I'll update it.

dmytrostruk (approved):

shawncal:
Great stuff! I cloned, played with this, ran all unit tests, integration tests. All looks good.

shawncal (approved):

1983Thai:
Hope to add comments to promote better

PR #1146, opened 2023-05-22
Exceptions thrown by Java API should extend SKException

dsgrieve:
### Motivation and Context
There were two SKException classes and other exception classes in the API that extended RuntimeException. There should, of course, only be one SKException class. The other exception classes that extend RuntimeException should extend SKException, which makes it easier for a caller who wants to catch exceptions thrown by the API. 

### Description
The SKException class in com.microsoft.semantickernel.diagnostics was removed in favor of the one in com.microsoft.semantickernel. 
Other exception classes in the API were changed to extend com.microsoft.semantickernel.SKException instead of RuntimeException. SKException itself is a RuntimeException. 

### Contribution Checklist
- [x] The code follows --the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`-- Java code style follows AOSP style. 


dsgrieve:
I don't know what happened here. Way too many changes.

PR #1149, opened 2023-05-22
CopilotChat: Low-level memory optimizations when uploading bot

adrianwyatt:
### Motivation and Context
High level semantic memory implementations, such as Azure Cognitive Search, do not allow for providing embeddings when storing memories. We wrap the memory store in an optional memory store to allow controllers to pass dependency injection validation and potentially optimize for a lower-level memory implementation (e.g. Qdrant). Lower level memory implementations (i.e., IMemoryStore) allow for reusing embeddings, whereas high level memory implementation (i.e., ISemanticTextMemory) assume embeddings get recalculated on every write.

### Description
- Implemented OptionalIMemoryStore to allow for BotController to pass DI validation and decide on using IMemoryStore or ISemanticTextMemory when uploading a bot.

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
+++ b/samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
@@ -24,6 +24,7 @@ namespace SemanticKernel.Service.CopilotChat.Controllers;
 public class BotController : ControllerBase
 {
     private readonly ILogger<BotController> _logger;
+    private readonly IMemoryStore? _memoryStore;
     private readonly ISemanticTextMemory _semanticMemory;
     private readonly ChatSessionRepository _chatRepository;
     private readonly ChatMessageRepository _chatMessageRepository;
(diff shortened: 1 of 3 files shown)


hathind-ms (dismissed):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
```diff
@@ -34,14 +35,15 @@ public class BotController : ControllerBase
     /// <summary>
     /// The constructor of BotController.
     /// </summary>
-    /// <param name="memoryStore">The memory store (optional).</param>
+    /// <param name="optionalIMemoryStore">Optional memory store.</param>
```
TaoChenOSU, line 38: nit: we should also detail when/why this optional memory store will be used in this controller. Maybe just copy and paste the comments in `SemanticKernelExtensions.cs` to here?

TaoChenOSU (approved):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
TaoChenOSU: Just to make things more obvious because it's little confusing.

glahaye (approved):

TaoChenOSU (approved):

PR #1151, opened 2023-05-22
CopilotChat: Replace planner:enabled with implicit disable when no plugins enabled.

adrianwyatt:
### Motivation and Context
We have optimized planner usage to automatically disable when no plugins are enabled.

### Description
- Remove planner options
- Replace planner:enabled check with check if any plugins are enabled.


samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatController.cs
```diff
@@ -78,10 +75,7 @@ public ChatController(ILogger<ChatController> logger)
         }
 
         // Register plugins that have been enabled
-        if (planner != null && plannerOptions.Value.Enabled)
-        {
-            await this.RegisterPlannerSkillsAsync(planner, plannerOptions.Value, openApiSkillsAuthHeaders, contextVariables);
-        }
+        await this.RegisterPlannerSkillsAsync(planner, openApiSkillsAuthHeaders, contextVariables);
```
amsacha, line 78: nit: Can you add an assert right before this call to fail when the planner is null

amsacha (approved):

glahaye (approved):

PR #1153, opened 2023-05-22
Http client consistency

SergeyMenshykh:

1. Reuse of globally configured policies and handlers: The AzureOpenAI, OpenAI, and custom connectors now use the globally configured policies and handlers provided with the custom HttpClient, eliminating the need for connector-specific implementations. This change improves consistency and allows reuse of globally configured settings.

2. Mandatory HttpClient parameter: HttpClient is now a mandatory parameter for AzureOpenAI, OpenAI, and custom connectors. This removes the responsibility of connectors to create their own HttpClients and manage(dispose) their lifetime, simplifying the code and promoting centralized HttpClient management.

3. To minimize the number of handlers utilized by the SK SDK, all AzureOpenAI, OpenAI, and custom connectors have been configured to reuse the same instance of HttpClientHandler.

### Contribution Checklist


samples/dotnet/kernel-syntax-examples/Example40_HttpUsage.cs
RogerBarreto, line 1: Suggest changing the name to `HttpClientUsage`

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/AzureOpenAIClientBase.cs
```diff
@@ -30,20 +31,19 @@ public abstract class AzureOpenAIClientBase : ClientBase
...
 
         var options = new OpenAIClientOptions();
-
-        if (httpClient != null)
-        {
-            options.Transport = new HttpClientTransport(httpClient);
-        }
+        // The following lines enable the reuse of globally configured SK SDK policies that are configured on the custom HTTPClient.
+        options.Transport = new HttpClientTransport(httpClient);
+        options.RetryPolicy = new NoRetryPolicy();
```
stephentoub, line 46: What is the default retry policy?

dotnet/src/Connectors/Connectors.AI.OpenAI/KernelConfigOpenAIExtensions.cs
```diff
@@ -393,11 +395,27 @@ public static class KernelConfigOpenAIExtensions
...
+    /// <param name="logger">An optional logger.</param>
+    /// <returns>An instance of HttpClient.</returns>
+    private static HttpClient GetHttpClient(IKernel kernel, HttpClient? httpClient, ILogger? logger)
     {
-        var retryHandler = handlerFactory.Create(logger);
-        retryHandler.InnerHandler = new HttpClientHandler { CheckCertificateRevocationList = true };
-        return new HttpClient(retryHandler);
+        if (httpClient == null)
+        {
+            var retryHandler = kernel.Config.HttpHandlerFactory.Create(logger ?? kernel.Log);
```
stephentoub, line 409: Pre-existing your PR, but it seems like we're creating our own variant of IHttpClientFactory here?  What's the reason for that?

dotnet/src/Connectors/Connectors.AI.OpenAI/Policies/NoRetryPolicy.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.Policies;
+/// <summary>
```
stephentoub, line 9:
```suggestion

/// <summary>
```

dotnet/src/Connectors/Connectors.AI.OpenAI/Policies/NoRetryPolicy.cs
```diff
@@ -0,0 +1,25 @@
...
+/// Represents a custom HTTP pipeline no retry policy.
+/// </summary>
+public class NoRetryPolicy : HttpPipelinePolicy
```
stephentoub, line 12: Can this be internal?

stephentoub, line 12: Rather than create a custom policy for this, can uses just use `new RetryPolicy(maxRetries: 0)`?

dotnet/src/Connectors/Connectors.AI.OpenAI/Policies/NoRetryPolicy.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <inheritdoc/>
+    public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
+    {
+        ProcessNext(message, pipeline);
+    }
+
+    /// <inheritdoc/>
+    public override async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
+    {
+        await ProcessNextAsync(message, pipeline).ConfigureAwait(false);
```
stephentoub, line 23:
```suggestion
        return ProcessNextAsync(message, pipeline);
```

samples/dotnet/kernel-syntax-examples/Example40_HttpUsage.cs
SergeyMenshykh: Yep, fixed.

dotnet/src/Connectors/Connectors.AI.OpenAI/Policies/NoRetryPolicy.cs
SergeyMenshykh: Yep, fixed.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/AzureOpenAIClientBase.cs
SergeyMenshykh: There are two mechanisms/properties used by Azure SDK for retry functionality - Retry and RetryPolicy. 

There is not default configuration for the RetryPolicy as far as I can see from source code. 

However, the Retry property assigned a default RetryOptions - https://github.com/Azure/azure-sdk-for-net/blob/1ceb48d8fd964b432a050c6dc024cb76ddbf5cdf/sdk/core/Azure.Core/src/RetryOptions.cs#L12
image

dotnet/src/Connectors/Connectors.AI.OpenAI/KernelConfigOpenAIExtensions.cs
SergeyMenshykh: Yep, looks like it. Later, we may consider extracting it into a separate factory class that can be used by all SK SDK components that need to use HTTP client.

dotnet/src/Connectors/Connectors.AI.OpenAI/KernelConfigOpenAIExtensions.cs
```diff
@@ -393,11 +395,27 @@ public static class KernelConfigOpenAIExtensions
...
+            retryHandler.InnerHandler = DefaultHttpClientHandler;
+            return new HttpClient(retryHandler, false); //We should refrain from disposing the underlying SK default HttpClient handler as it would impact other HTTP clients that utilize the same handler.
```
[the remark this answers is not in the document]
SergeyMenshykh: As discussed offline, neither (1) or (2) matters because at this stage (.net runtime is in the process of cleaning up application resources and disposing the instance of the HttpClientHandler) all the requests should be completed and those that are not completed (app is closing after an HTTP request was sent but before the response was received) could not relevant anymore because application is closing/exiting.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureChatCompletion.cs
```diff
@@ -31,7 +31,7 @@ public sealed class AzureChatCompletion : AzureOpenAIClientBase, IChatCompletion
         string modelId,
         string endpoint,
         string apiKey,
-        HttpClient? httpClient = null,
+        HttpClient httpClient,
```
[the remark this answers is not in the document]
shawncal: I don't love having the HttpClient be required... I suspect most developers won't care or understand why they have to provide their own here. (As you say, most should likely use the "WithOpenAI..." extension methods, but some people are using these directly for access to the new streaming mehods).

The other thing that jumps out at me, is that with `httpClient` being required and `organization` being optional, you kind of split the openAI API args around an infrastructure arg. If we stick with required httpClient, then I'd recommend creating overloads with/without the "string organization" arg before httpClient, instead of giving it a default value.

That said, my preference would be to keep httpClient optional if possible; as Stephen says, share an assembly instance of handlers. And just make sure that, where we're calling it with helpers and from KernelBuilder, we're passing it in with the kernel's settings/defaults.

dotnet/src/Connectors/Connectors.AI.OpenAI/KernelConfigOpenAIExtensions.cs
shawncal: @SergeyMenshykh @stephentoub Would [SocketsHttpHandler ](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler?view=net-7.0) be appropriate to use here (as an assembly-level fallback, if httpClient args remain optional and caller doesn't provide one?
As an alternative to HttpClientFactory, without the DI stuff, it still provides common settings and protection against port exhaustion.

stephentoub: Yes... but not while targeting netstandard2.0.  If/when the assembly multitargets, it could use a singleton HttpClientHandler for netstandard2.0 and a singleton SocketsHttpHandler for net6.0+.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureChatCompletion.cs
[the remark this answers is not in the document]
SergeyMenshykh: Make sense, the changes are rolled back.

dotnet/src/InternalUtilities/Http/NonDisposableHttpClientHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+
+/// <summary>
+/// Represents a singleton implementation of <see cref="HttpClientHandler"/> that is not disposable.
+/// </summary>
+internal sealed class NonDisposableHttpClientHandler : HttpClientHandler
+{
+    /// <summary>
+    /// Lazy-initialized instance of <see cref="NonDisposableHttpClientHandler"/>.
+    /// </summary>
+    private static readonly Lazy<NonDisposableHttpClientHandler> instance = new(() => new NonDisposableHttpClientHandler());
```
stephentoub, line 13: This doesn't need to be Lazy.  It can just be:
```C#
public static NonDisposableHttpClientHandler Instance { get; } = new NonDisposableHttpClientHandler();
```
The only time this type is accessed is when accessing its Instance member, so the runtime will handle appropriately lazily-initializing it before first use.  Further, even if it wasn't lazy, there are very few resources associated with an unused HttpClientHandler, as the expensive connection resources are themselves created lazily when needed.

stephentoub (approved):

dotnet/src/InternalUtilities/Http/NonDisposableHttpClientHandler.cs
SergeyMenshykh: Got it. It was not about optimal resources usage and more about dealing with concurrency - I remember using Lazy was one of the simplest ways to have singleton out of the box without using locks/double ifs, etc. Not sure if it's still relevan and the Lazy class is good candidate for this, though.

stephentoub:
```C#
public static NonDisposableHttpClientHandler Instance { get; } = new NonDisposableHttpClientHandler();
```
is thread-safe.  It'll be constructed once and all threads will see the same published instance.  The runtime handles all necessary synchronization around static cctor invocation.  And in this case, the type will end up being beforefieldinit such that the runtime can employ additional optimizations.

samples/dotnet/kernel-syntax-examples/KernelSyntaxExamples.csproj
```diff
@@ -21,6 +21,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="7.0.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0-preview.4.23259.5" />
+    <PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
```
RogerBarreto, line 24: Just checking that we have a .NET 6 project using a 7.0 package and also a 8.0-preview.

We should be fixed to 6.0 only right?
image

lemillermicrosoft: It looks like all of those packages support net6.0 by default. The version number isn't in reference to the version of .net it supports.

lemillermicrosoft: https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection/8.0.0-preview.4.23259.5#supportedframeworks-body-tab

PR #1172, opened 2023-05-22
Restructuring BaseService to handle FormData as body + adds Import Document is a No Response operation 

teresaqhoang:
### Motivation and Context

Restructuring BaseService to handle FormData as body + adds Import Document is a No Response operation

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,8 @@ Follow the links for more information and instructions about running these apps.
   to get started.
 - [Azure Functions Core Tools](https://learn.microsoft.com/azure/azure-functions/functions-run-local)
   are required to run the kernel as a local web service, used by the sample web apps.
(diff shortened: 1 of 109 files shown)


glahaye (approved):

TaoChenOSU (approved):

TaoChenOSU:
We can let this in now. I think moving forward it is better for us to have the base service know nothing about the other services that derives the base service. We can have the base service to provide multiple methods for different content types and respond types.

hathind-ms (approved):

teresaqhoang:
> We can let this in now. I think moving forward it is better for us to have the base service know nothing about the other services that derives the base service. We can have the base service to provide multiple methods for different content types and respond types.

Yeah I agree, I think we also need a better way to handle operations that expect no response body (whether it's aligning all requests in the backend to return either a 202 Accepted or a 204 No Content)

PR #1175, opened 2023-05-23
Java Sample Code modified to modernized code and easy to understand

yoshioterada:
This is the update pull request of 
https://github.com/microsoft/semantic-kernel/pull/1128

I closed the previous Pull Request #1128.

1 Modified getConfigValue()
1.1: local variable name from apiKey to propertyValue 1.2: Use NIO.2 libraries instead of old FileInputStream 1.3: modified Exception throw code

2. Defined TEXT_TO_SUMMARIZE Instead of local variable definition, I defined static final field

3. As a sample, I wrote non blocking code for Async operation Previously it was wrote as
summarize.invokeAsync(TEXT_TO_SUMMARIZE).block() to subscribe()

4. In order to show the result, I added following code to receive the result. try {
TimeUnit.SECONDS.sleep(10);
} catch (InterruptedException e) {
LOGGER.warn("Interrupted : {}", e.getMessage());
Thread.currentThread().interrupt();
}

5. Code Structure modified: I modified the code structure of both 'InlineFunctionExample.java' and 'InlineFunctionWithPreBuiltSkillExample.java' to be as similar as possible, making them easier to compare.


markwallace-microsoft (commented):
LGTM

java/api-test/api-usage-example/src/main/java/InlineFunctionExample.java
```diff
@@ -20,107 +19,127 @@
...
+                },
+                error -> {
+                    LOGGER.error("Error: {} ", error.getMessage());
+                },
+                () -> {
+                    LOGGER.info("Completed");
+                });
+
+        try {
+            TimeUnit.SECONDS.sleep(10);
```
johnoliver, line 139: The 10 second sleep does make this a bit probabilistic, especially when we have throttles on usage. I would use a count down latch to block externally. i.e

```
CountDownLatch cdl = new CountDownLatch(1);
...
() -> {
    LOGGER.info("Completed");
    cdl.countDown();
})

cdl.await();
```

java/api-test/api-usage-example/src/main/java/InlineFunctionExample.java
```diff
@@ -20,107 +19,127 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(InlineFunctionExample.class);
     private static final String MODEL = "text-davinci-003";
 
-    public static String getToken(String configName) throws IOException {
+    private static String API_KEY = "";
+    private static String ENDPOINT = "";
+
+    static{
+        try {
+            API_KEY = getToken(AZURE_CONF_PROPERTIES);
```
johnoliver, line 34: Just as an FYI we have an update soon reworking reading tokens

java/api-test/api-usage-example/src/main/java/InlineFunctionWithPreBuiltSkillExample.java
```diff
@@ -69,65 +79,70 @@ public static String getEndpoint(String configName) throws IOException {
...
+                },
+                error -> {
+                    LOGGER.error("Error: {} ", error.getMessage());
+                },
+                () -> {
+                    LOGGER.info("Completed");
+                });
+
+        try {
+            TimeUnit.SECONDS.sleep(10);
```
johnoliver, line 142: same as above

johnoliver (approved):

johnoliver (dismissed):

markwallace-microsoft (approved):

PR #1193, opened 2023-05-23
Java Semantic Kernel CI

Luigi96:
### Description
Opening a PR with initial CI changes to build and run tests against Java packages.

### Contribution Checklist


.github/workflows/java-build.yml
```diff
@@ -1,24 +1,61 @@
...
+permissions:
+  contents: read
 
+jobs:
+  java-build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java-versions: [8, 11, 17]
```
brunoborges, line 23: Hey @johnoliver, should @Luigi96 here update the build pipeline so that the library is built using JDK 17, with target for Java 8?

One thing I believe we would have to do though, is run tests with JDK 8, unless there is a guaranteed way to prevent us from using Java 17 APIs (types/methods).

If it is ok to have 8 for building, should we simplify and remove 11? Or do we need all three?

johnoliver: I would prefer to build with 8 and at least one lts>9, (i.e 11, 17). I don't _think_ there is going to be much difference discovered between 11 and 17 build so I am fine with removing 11

.github/workflows/java-build.yml
```diff
@@ -1,24 +1,61 @@
...
         with:
-          java-version: '11'
-          distribution: 'temurin'
+          java-version: ${{ steps.set-jdk.outputs.JDK_VERSION }}
+          distribution: microsoft
           cache: maven
 
       - name: Build with Maven
-        run: cd java && ./mvnw -B -Pbug-check clean install --file pom.xml
+        run: ./mvnw -B -Pbug-check -DskipTests -X -Pcompile-jdk${{ matrix.java-versions }} clean install --file pom.xml
```
johnoliver, line 50: I would remove the -X

.github/workflows/java-integration-tests.yml
```diff
@@ -1,30 +1,70 @@
 name: Run Java Integration Tests and Samples
 
 on:
+  workflow_dispatch:
   push:
     branches: [ "experimental-java", "java-dev" ]
```
johnoliver, line 6: java-dev

.github/workflows/java-integration-tests.yml
```diff
@@ -1,30 +1,70 @@
...
+    paths:
+      - 'java/**'
+      - 'samples/java/**'
```
johnoliver, line 9: for now we dont commit to `samples/java`

java/pom.xml
```diff
@@ -10,14 +10,14 @@
...
 
 
-    <properties>
+    <!-- <properties>
         <jdk.version>1.8</jdk.version>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <maven.compiler.testSource>1.8</maven.compiler.testSource>
         <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
johnoliver, line 19: can keep the sourceEncoding

java/pom.xml
```diff
@@ -131,22 +131,41 @@
     </build>
 
     <profiles>
+         <profile>
+            <id>compile-jdk8</id>
```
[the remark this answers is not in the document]
Luigi96: As discussed outside this PR, I will set JDK17 as default instead.

.github/workflows/java-build.yml
brunoborges: Alright, so after our meeting today, let's move forward with JDK 8 and JDK 17 only, in the build pipelines.

samples/java/semantickernel-samples/pom.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
markwallace-microsoft, line 1: @SergeyMenshykh I think you added this Java sample. Is it related to the Java SK implementation?

samples/java/semantickernel-samples/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
johnoliver, line 2: Unfortunately cannot have changes outside of /java, we can ignore the sample that is at /samples/java for now, no need to keep that up to date

Luigi96: I will remove the file then. Should I remove the `Build semantic-kernel samples` step since the path doesn't exist on the `experimental-java` branch?

johnoliver: building the samples that are in `java/samples` on the experimental-java branch is fine for now we can just ignore the `/samples` directory

samples/java/semantickernel-samples/pom.xml
Luigi96: File removed since we are not going to alter files outside the java directory.

johnoliver (dismissed):

johnoliver (approved):

markwallace-microsoft (approved):

PR #1199, opened 2023-05-24
Java - Add module to allow dependency injection via guice

johnoliver:
### Contribution Checklist


java/semantickernel-guice/src/main/java/com/microsoft/semantickernel/guice/SemanticKernelModule.java
```diff
@@ -0,0 +1,131 @@
...
+        if (withMemory) {
+            bind(SemanticTextMemory.class).toProvider(Providers.of(memory));
+        } else {
+            bind(SemanticTextMemory.class).toInstance(NullMemory.getInstance());
+        }
+
+        bind(Kernel.class).to(DefaultKernel.class);
+        bind(CompletionFunctionFactory.class)
+                .to(CompletionFunctionFactory.CompletionFunctionFactoryImpl.class);
+    }
```
markwallace-microsoft, line 119: I haven't used Guice before but if I understand correctly this Module does the bindings for SK using OpenAI text completion or embedding. So for any other combination of provider(s) I would provide a different Module.

Would this be better as a sample?

At a minimum the name should change e.g. `OpenAIBindingModule` might be better.

johnoliver: Yes this is a fairly opinionated module that is specifically binding those services to an OpenAI implementation, so maybe should be renamed to `semantickernel-openai-guice` to indicate that is specifically binding to OpenAI implementations. Potentially its better as a sample to show _how_ to create a module for dependency injection rather than a reusable component

johnoliver: fixed

1983Thai:
Yeah

markwallace-microsoft (approved):

PR #1209, opened 2023-05-25
Add Azure Dall-E

xbotter:
### Motivation and Context

Implement Azure OpenAI Dall E

### Description
Implement Azure OpenAI Image Generation according to the official [documentation](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/reference#image-generation) and the examples provided by Azure OpenAI Studio, so that it can be used in SK.

Due to the differences between the API of Azure OpenAI Dall-E and OpenAI Dall-E, some modifications have been made to OpenAIClientBase under CustomClient to ensure compatibility.

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,109 @@
...
+            throw new AIException(
+                AIException.ErrorCodes.UnknownError,
+                $"Something went wrong: {e.Message}", e);
+        }
+        finally
+        {
+            response?.Dispose();
+        }
+    }
+    /// <summary>Adds headers to use for Azure OpenAI HTTP requests.</summary>
```
lemillermicrosoft, line 104:
```suggestion

    /// <summary>Adds headers to use for Azure OpenAI HTTP requests.</summary>
```

dotnet/src/Connectors/Connectors.AI.OpenAI/CustomClient/OpenAIClientBase.cs
```diff
@@ -114,112 +114,139 @@ private protected virtual void AddRequestHeaders(HttpRequestMessage request)
...
+        {
+            using (var request = new HttpRequestMessage(method, url))
             {
                 this.AddRequestHeaders(request);
-                request.Content = new StringContent(requestBody, Encoding.UTF8, "application/json");
+                if (content != null)
+                {
+                    request.Content = content;
+                }
                 response = await this._httpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
```
lemillermicrosoft, line 165:
```suggestion

                response = await this._httpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
```

dotnet/src/Connectors/Connectors.AI.OpenAI/CustomClient/OpenAIClientBase.cs
```diff
@@ -114,112 +114,139 @@ private protected virtual void AddRequestHeaders(HttpRequestMessage request)
...
                 AIException.ErrorCodes.UnknownError,
                 $"Something went wrong: {e.Message}", e);
         }
-        finally
-        {
-            response?.Dispose();
-        }
     }
-
     #endregion
```
lemillermicrosoft, line 251:
```suggestion

    #endregion
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,109 @@
...
+    }
+    /// <inheritdoc/>
+    public async Task<string> GenerateImageAsync(string description, int width, int height, CancellationToken cancellationToken = default)
+    {
+        Verify.NotNull(description);
+        if (width != height || (width != 256 && width != 512 && width != 1024))
+        {
+            throw new ArgumentOutOfRangeException(nameof(width), width, "OpenAI can generate only square images of size 256x256, 512x512, or 1024x1024.");
+        }
+        var requestBody = Json.Serialize(new AzureImageGenerationRequest()
```
lemillermicrosoft, line 48:
```suggestion

        var requestBody = Json.Serialize(new AzureImageGenerationRequest()
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            Caption = description,
+            Resolution = $"{width}x{height}"
+        });
+        var result = await this.EnsureImageGenerationAsync(this._endpoint, requestBody, cancellationToken).ConfigureAwait(false);
+        if (result.Result == null)
+        {
+            throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Response not contains result");
+        }
+        return result.Result.ContentUrl;
```
lemillermicrosoft, line 58:
```suggestion

        return result.Result.ContentUrl;
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,109 @@
...
+    private async Task<AzureImageGenerationResponse> EnsureImageGenerationAsync(string url, string? requestBody, CancellationToken cancellationToken = default)
+    {
+        HttpResponseMessage? response = null;
+        try
+        {
+            using (var content = new StringContent(requestBody, Encoding.UTF8, "application/json"))
+            {
+                response = await this.ExecuteRequestAsync(url, HttpMethod.Post, content, cancellationToken).ConfigureAwait(false);
+            }
+            var responseJson = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```
lemillermicrosoft, line 70:
```suggestion

            var responseJson = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,109 @@
...
+            var result = this.JsonDeserialize<AzureImageGenerationResponse>(responseJson);
+            while (result.Status != "Succeeded")
+            {
+                if (!response.Headers.TryGetValues("operation-location", out var locationValues) || string.IsNullOrWhiteSpace(locationValues.FirstOrDefault()))
+                {
+                    throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Not found operation-location");
+                }
+                var operationLocation = locationValues.First();
```
lemillermicrosoft, line 78:
```suggestion

                var operationLocation = locationValues.First();
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,109 @@
...
+                if (response.Headers.TryGetValues("retry-after", out var retryValues))
+                {
+                    var retryValue = retryValues.FirstOrDefault();
+                    if (int.TryParse(retryValue, out var retry))
+                    {
+                        await Task.Delay(TimeSpan.FromSeconds(retry), cancellationToken).ConfigureAwait(false);
+                    }
+                }
+                response = await this.ExecuteRequestAsync(operationLocation, HttpMethod.Get, null, cancellationToken).ConfigureAwait(false);
```
lemillermicrosoft, line 87:
```suggestion

                response = await this.ExecuteRequestAsync(operationLocation, HttpMethod.Get, null, cancellationToken).ConfigureAwait(false);
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,109 @@
...
+                responseJson = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                result = this.JsonDeserialize<AzureImageGenerationResponse>(responseJson);
+            }
+            return result;
```
lemillermicrosoft, line 91:
```suggestion

            return result;
```

lemillermicrosoft (commented):
Tested locally. Batch of formatting comments.

lemillermicrosoft (commented):
Tested locally. Batch of formatting comments.

lemillermicrosoft (dismissed):
LGTM! Thanks for the PR.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,119 @@
...
+        var result = await this.EnsureImageGenerationAsync(this._endpoint, requestBody, cancellationToken).ConfigureAwait(false);
+        if (result.Result == null)
+        {
+            throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Response not contains result");
+        }
+
+        return result.Result.ContentUrl;
+    }
+
+    private async Task<AzureImageGenerationResponse> EnsureImageGenerationAsync(string url, string? requestBody, CancellationToken cancellationToken = default)
```
dmytrostruk, line 66: nit: Can we simplify method name to something like:
- `GenerateImageAsync`
- `GetImageAsync`
- `GetImageGenerationAsync`

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,119 @@
...
+        try
+        {
+            using (var content = new StringContent(requestBody, Encoding.UTF8, "application/json"))
+            {
+                response = await this.ExecuteRequestAsync(url, HttpMethod.Post, content, cancellationToken).ConfigureAwait(false);
+            }
+
+            var responseJson = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+            var result = this.JsonDeserialize<AzureImageGenerationResponse>(responseJson);
+            while (result.Status != "Succeeded")
```
dmytrostruk, line 78: 1. I think we should put this status value to the constant.
2. What will happen if result status will be `Failed` for example, are we going to hit infinite loop? Should we consider to add timeout logic here?

xbotter:
I tried to add some limitations on the maximum number of retries to avoid falling into an infinite loop.

While making improvements, I noticed that there have been some updates to the Azure Dall-E API version, so I also took the opportunity to update it.

Please review again. @dmytrostruk

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    /// <summary>
+    /// Constants of success status
+    /// </summary>
+    private const string SucceededStatus = "SUCCEEDED";
+
+    /// <summary>
+    /// Maximum retry times for the Azure Dall-E API
+    /// </summary>
+    private const int MaxRetryCount = 5;
```
dmytrostruk, line 28: We can add this as optional parameter to constructor and use default value from this constant. We can add the same parameter to extension method in `OpenAIKernelBuilderExtensions`. In this case `max retry count` will be customizable by user.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    /// <summary>
+    /// Default retry wait time in seconds
+    /// </summary>
+    private const int DefaultRetryAfter = 6;
```
dmytrostruk, line 33: We can add this as optional parameter to constructor and use default value from this constant. We can add the same parameter to extension method in `OpenAIKernelBuilderExtensions`. In this case `retry-after` will be customizable by user.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            if (!response.Headers.TryGetValues("operation-location", out var locationValues) || string.IsNullOrWhiteSpace(locationValues.FirstOrDefault()))
+            {
+                throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Not found operation-location");
+            }
+
+            var operationLocation = locationValues.First();
+            while (true)
+            {
+                if (SucceededStatus == result.Status.ToUpperInvariant())
```
dmytrostruk, line 105: nit: I think you can simplify here to avoid maintaining uppercase strings:
```suggestion
                if (result.Status.Equals(SucceededStatus, StringComparison.OrdinalIgnoreCase))
```

dmytrostruk (approved):
LGTM, a couple of small comments.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,153 @@
...
+    /// </summary>
+    /// <param name="endpoint">Azure OpenAI deployment URL, see https://learn.microsoft.com/azure/cognitive-services/openai/quickstart</param>
+    /// <param name="apiKey">Azure OpenAI API key, see https://learn.microsoft.com/azure/cognitive-services/openai/quickstart</param>
+    /// <param name="httpClient">Custom <see cref="HttpClient"/> for HTTP requests.</param>
+    /// <param name="logger">Application logger</param>
+    /// <param name="maxRetryCount">Maximum number of retries for HTTP requests.</param>
+    /// <param name="defaultRetryAfter">Default retry wait time in seconds for HTTP requests.</param>
+    public AzureOpenAIImageGeneration(string endpoint, string apiKey, HttpClient? httpClient = null, ILogger? logger = null, int maxRetryCount = 5, int defaultRetryAfter = 6) : base(httpClient, logger)
+    {
+        this._endpoint = $"{endpoint}openai/images/generations:submit?api-version=2023-06-01-preview";
```
SergeyMenshykh, line 56: There are a few things to think about here:
1. The url can and will change, so we need to be able easily provide it without changing the class. 
2. It's a valid scenario when a custom HttpClient is passed to this constructor and have the BaseAddress property set. We should respect it.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,153 @@
...
+            Count = 1
+        });
+
+        var result = await this.GetImageGenerationAsync(this._endpoint, requestBody, cancellationToken).ConfigureAwait(false);
+        if (result.Result == null)
+        {
+            throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Response not contains result");
+        }
+
+        return result.Result.Images[0].Url;
```
SergeyMenshykh, line 84: What's going to happen if the array is empty?

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    private async Task<AzureImageGenerationResponse> GetImageGenerationAsync(string url, string? requestBody, CancellationToken cancellationToken = default)
```
SergeyMenshykh, line 87: Why the requestBody parameter is optional?

SergeyMenshykh, line 87: This method has multiple responsibilities. It would be good to extract those responsibilities into separate methods with clear names, distinguishing the POST and GET requests.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    await Task.Delay(TimeSpan.FromSeconds(this._defaultRetryAfter), cancellationToken).ConfigureAwait(false);
+                }
+
+                response = await this.ExecuteRequestAsync(operationLocation, HttpMethod.Get, null, cancellationToken).ConfigureAwait(false);
+                responseJson = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                result = this.JsonDeserialize<AzureImageGenerationResponse>(responseJson);
+
+                // increase retry count
+                retryCount++;
```
lemillermicrosoft, line 133: I'm curious if we need to do this retry or not. The kernel supports adding this handling by default or custom httpClients can handle themselves.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            using (var content = new StringContent(requestBody, Encoding.UTF8, "application/json"))
+            {
+                response = await this.ExecuteRequestAsync(url, HttpMethod.Post, content, cancellationToken).ConfigureAwait(false);
+            }
+
+            var responseJson = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+            var result = this.JsonDeserialize<AzureImageGenerationResponse>(responseJson);
+
+            if (!response.Headers.TryGetValues("operation-location", out var locationValues) || string.IsNullOrWhiteSpace(locationValues.FirstOrDefault()))
```
SergeyMenshykh, line 101: What's return in the operation-location header? Could you please put a comment with a link to a documentation describing the header, assuming it exists?

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Not found operation-location");
+            }
+
+            var operationLocation = locationValues.First();
+            while (true)
+            {
+                if (result.Status.Equals(SucceededStatus, StringComparison.OrdinalIgnoreCase))
+                {
+                    return result;
```
SergeyMenshykh, line 111: There's no need to implement the retry functionality in this class because a default one configured during Kernel building will be provided automatically when the class is registered by the WithAzureOpenAIImageGenerationService method. In case the class is created directly, without using this With* method, it's always possible inject an instance of HttpClient with any retry policy.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
xbotter: There are some differences in the retry process here compared to httpClient. 

Azure's Dall E API follows the following process:
First, calling the submit API generates a task, and the response header includes the operation-location indicating the address to query for the task result. 
Then, it is necessary to continuously attempt to check the generated result from the operation-location. 
The ongoing response header also specifies a Retry-After parameter indicating the delay for retrying until the final result state changes to succeeded, which signifies a successful request.

Therefore, some retry steps and limitations have been added here to avoid getting stuck in an infinite loop.

The currently known statuses include notRunning, running, and succeeded. 
So far, no failed cases have been encountered, and it cannot be guaranteed that all tasks will be successful. 
Perhaps it would be advisable to only perform periodic queries during the notRunning and running states?
Any better suggestions are welcome.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
xbotter: 1. Different URLs may have some differences in processing logic, but it could be considered as an optional configuration option for users to modify.
2. This is indeed something that needs to be considered, and I will deal with it later.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
xbotter: Currently, there is no official documentation that provides detailed instructions on how to use this API. 
Based on the call example provided by Azure OpenAI Studio, we can see that the operation-location points to the query address for the generated result, which is used to obtain the final image generation result.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
xbotter: comments here
https://github.com/microsoft/semantic-kernel/pull/1209#discussion_r1220689587

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,153 @@
...
+    private readonly string _apiKey;
+
+    /// <summary>
+    /// Create a new instance of Azure OpenAI image generation service
+    /// </summary>
+    /// <param name="endpoint">Azure OpenAI deployment URL, see https://learn.microsoft.com/azure/cognitive-services/openai/quickstart</param>
+    /// <param name="apiKey">Azure OpenAI API key, see https://learn.microsoft.com/azure/cognitive-services/openai/quickstart</param>
+    /// <param name="httpClient">Custom <see cref="HttpClient"/> for HTTP requests.</param>
+    /// <param name="logger">Application logger</param>
+    /// <param name="maxRetryCount">Maximum number of retries for HTTP requests.</param>
```
SergeyMenshykh, line 52: Please update the comment for the maxRetryCount parameter to clearly indicate that it's related to the maximum number of attempts to made to retrieve an image generation operation result, rather than to resilience-related retries. 
For example: "Maximum number of attempts to retrieve the image generation operation result."

dotnet/src/Connectors/Connectors.AI.OpenAI/CustomClient/OpenAIClientBase.cs
```diff
@@ -114,111 +114,140 @@ private protected virtual void AddRequestHeaders(HttpRequestMessage request)
     /// </summary>
     private readonly HttpClient _httpClient;
 
-    private async Task<T> ExecutePostRequestAsync<T>(string url, string requestBody, CancellationToken cancellationToken = default)
+    private protected async Task<T> ExecutePostRequestAsync<T>(string url, string requestBody, CancellationToken cancellationToken = default)
     {
         HttpResponseMessage? response = null;
+        HttpContent? content = null;
```
SergeyMenshykh, line 120: Neither of the declared variables is used outside the try block/scope. So, please declare and dispose them inside the block using this syntax: using var content = .....

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
xbotter: That's great! I actually need clear documentation right now. I'll give the API a try and make adjustments to the implementation method.

xbotter:
@SergeyMenshykh Thanks for your suggestion. I have updated the implementation method according to the document, and it now appears much clearer.

If there is anything that needs to be adjusted, please let me know.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,198 @@
...
+    /// <param name="logger">Application logger</param>
+    /// <param name="maxRetryCount"> Maximum number of attempts to retrieve the image generation operation result.</param>
+    /// <param name="apiVersion">Azure OpenAI Endpoint ApiVersion</param>
+    public AzureOpenAIImageGeneration(string endpoint, string apiKey, HttpClient? httpClient = null, ILogger? logger = null, int maxRetryCount = 5, string apiVersion = "2023-06-01-preview") : base(httpClient, logger)
+    {
+        this._endpoint = endpoint;
+        this._apiKey = apiKey;
+        this._maxRetryCount = maxRetryCount;
+        this._apiVersion = apiVersion;
+    }
```
SergeyMenshykh, line 68: There are a few considerations regarding the endpoint and custom HTTP client:
1. It should be possible to use endpoint provided through BaseAddress property of the custom HTTP client. The scenario is possible when a hosting application has/creates a dedicated and preconfigured HTTP client per host.
2. At the same time, if the BaseAddress property is not set then the endpoint parameter value should be used instead. That can be convenient if a hosting application needs to target different hosts dinamicaly.
3. If both BaseAddress and endpoint properties are specified the parameter value should override BaseAddress property value. This is useful in cases where a hosting application generally targets the same host in the majority of scenarios, but there may be a rare instance where it needs to communicate with a different host.
4. If neither the BaseAddress nor the endpoint parameter is specified, an exception should be thrown to prevent the construction of the class and avoid an invalid class state.

This approach is implemented in the HugginFace connector - [the class constructor](https://github.com/microsoft/semantic-kernel/blob/4c4670acb2ffb9d16ca406a61e55cd10c96b345d/dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs#LL107C5-L107C5), the [Url selection logic](https://github.com/microsoft/semantic-kernel/blob/4c4670acb2ffb9d16ca406a61e55cd10c96b345d/dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs#L203)

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,198 @@
...
+    /// <param name="endpoint">Azure OpenAI deployment URL, see https://learn.microsoft.com/azure/cognitive-services/openai/quickstart</param>
+    /// <param name="apiKey">Azure OpenAI API key, see https://learn.microsoft.com/azure/cognitive-services/openai/quickstart</param>
+    /// <param name="httpClient">Custom <see cref="HttpClient"/> for HTTP requests.</param>
+    /// <param name="logger">Application logger</param>
+    /// <param name="maxRetryCount"> Maximum number of attempts to retrieve the image generation operation result.</param>
+    /// <param name="apiVersion">Azure OpenAI Endpoint ApiVersion</param>
+    public AzureOpenAIImageGeneration(string endpoint, string apiKey, HttpClient? httpClient = null, ILogger? logger = null, int maxRetryCount = 5, string apiVersion = "2023-06-01-preview") : base(httpClient, logger)
+    {
+        this._endpoint = endpoint;
+        this._apiKey = apiKey;
```
SergeyMenshykh, line 65: Please add a check to verify that the supplied apiKey is valid, meaning it is not null and not empty.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
```diff
@@ -0,0 +1,198 @@
...
+using Microsoft.SemanticKernel.Diagnostics;
+using Microsoft.SemanticKernel.Text;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.ImageGeneration;
+
+/// <summary>
+/// Azure OpenAI Image generation
+/// <see herf="https://learn.microsoft.com/en-us/azure/cognitive-services/openai/reference#image-generation" />
+/// </summary>
+public class AzureOpenAIImageGeneration : OpenAIClientBase, IImageGeneration
```
SergeyMenshykh, line 21: Include unit tests to ensure that happy path scenarios are tested.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
xbotter: Is this okay?
```csharp
public AzureOpenAIImageGeneration(string? endpoint, string apiKey, HttpClient? httpClient = null, ILogger? logger = null, int maxRetryCount = 5, string apiVersion = "2023-06-01-preview") : base(httpClient, logger)
{
    endpoint = !string.IsNullOrEmpty(endpoint) ? endpoint : httpClient?.BaseAddress?.AbsoluteUri;
    Verify.NotNullOrWhiteSpace(endpoint);
```

xbotter: The way of two constructors looks better. Let me work on this.

dotnet/src/Connectors/Connectors.AI.OpenAI/ImageGeneration/AzureOpenAIImageGeneration.cs
xbotter: Done

SergeyMenshykh (approved):

lemillermicrosoft (approved):

raffaeler:
Could someone tell me which version/nuget package will contain this fix?
Is there a nightly feed to get this now?

Thanks!

PR #1224, opened 2023-05-25
Add Java to labeler config

dluc:
* Update rules to label also Java PRs
* Improve python settings
* Simplify .NET rule to catch all kernel changes


.github/labeler.yml
```diff
@@ -1,26 +1,34 @@
...
 # Add 'kernel.core' label to any change within the 'SemanticKernel', 'SemanticKernel.Abstractions', or 'SemanticKernel.MetaPackage' directories
 kernel.core:
   - dotnet/src/SemanticKernel/**/*
   - dotnet/src/SemanticKernel.Abstractions/**/*
   - dotnet/src/SemanticKernel.MetaPackage/**/*
-
-# Add 'python' label to any change within the 'python' directory
-python:
-  - python/**/*
+  - python/semantic_kernel**/*
```
lemillermicrosoft, line 12: I'd keep these separate or introduce python.core labels. I appreciate being able to filter between things that impact skills/connectors/extesions and things that impact the core kernel code.

.github/labeler.yml
```diff
@@ -1,26 +1,34 @@
 # Add 'kernel' label to any change within Connectors, Extensions, Skills, and tests directories
 kernel:
-  - "dotnet/src/Connectors/**/*"
-  - "dotnet/src/Extensions/**/*"
-  - "dotnet/src/Skills/**/*"
-  - "dotnet/src/IntegrationTests/**/*"
-  - "dotnet/src/SemanticKernel.UnitTests/**/*"
+  - dotnet/src/**/*
```
lemillermicrosoft, line 3: Can you test that this will not label changes that touch only  SemanticKernel.Abstractions as `kernel`?

.github/labeler.yml
dluc: we can use "python.core" but then we should rename "kernel.core" to ".NET.core" or something similar

lemillermicrosoft, line 12: Won't this include connectors/skill only changes in python?

lemillermicrosoft (changes requested):
Some test PRs would be nice to confirm behavior.

.github/labeler.yml
dluc: can we rename the labels? e.g. why would we exclude Abstractions from "kernel"?

.github/labeler.yml
dluc: yes, I see that as part of the core. I think it would help renaming the labels, there is a lot of overlap between the definition of "kernel" and "kernel core"

.github/labeler.yml
lemillermicrosoft: I'm open to renaming. I want a way to be able to identify whether a PR touches SemanticKernel or SemanticKernel.Abstractions, whether it touches Skills/Connectors/Extensions, or both.

PR #1229, opened 2023-05-25
AI Connectors - Automatic Add All Models

carlos-the-ai:
### Motivation and Context
Adding all models from Azure OpenAI or OpenAI should be easier to do. Right now, they need to be added line by line.

### Description
Allows for calling just `Kernel.Builder.WithOpenAI(key)` or  `Kernel.Builder.WithAzureOpenAI(endpoint, key)` and all models are added automatically! 

Also supports adding a prefix for service id, in case you add two or more connectors with the same model names. Example:
with prefix "openai:" the service id will be "openai:gpt-3.5-turbo".

Uses REST API instead of the Azure Open AI SDK because the SDK does not seem to have the function to list models. I hope this will come in the future and can replace the REST API calls.

Azure API does not have a property that tells if the model supports "chat" so assume that models starting with "gpt-" support chat. Again, future azure API version should hopefully add this.

OpenAI does list models but does not show what kind of model they are, so right now it just uses the model names they have published. Will need to be updated as new models come out. Like Azure, they might have a better API that lists model type in the future and this can be replaced.

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
```diff
@@ -20,6 +23,149 @@ namespace Microsoft.SemanticKernel;
...
+        HttpClient? httpClient = null,
+        ILogger? logger = null,
+        CancellationToken cancellationToken = default)
+    {
+        // Get list of (available) models on Azure OpenAI instance
+        var models = OpenAIRestClient.GetModelsAsync(apiKey, organization, httpClient, cancellationToken).RunTaskSynchronously();
+
+        foreach (var model in models.Values)
+        {
+            string serviceId = $"{serviceIdPrefix ?? string.Empty}{model.Id}";
```
lemillermicrosoft, line 126: Perhaps the fallback could be `openai:, similar above.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
```diff
@@ -20,6 +23,149 @@ namespace Microsoft.SemanticKernel;
...
+    /// <param name="builder">The <see cref="KernelBuilder"/> instance</param>
+    /// <param name="apiKey">OpenAI API key, see https://learn.microsoft.com/azure/cognitive-services/openai/quickstart</param>
+    /// <param name="organization">OpenAI organization Id</param>
+    /// <param name="serviceIdPrefix">Optional srting to prepend onto the model name when registering the service. Used to disambiguate, when multiple connectors
+    /// may have the same model names. Example: specifying prefix "openai:" will result in model registrations of the form "openai:gpt-3.5-turbo".</param>
+    /// <param name="httpClient">Custom <see cref="HttpClient"/> for HTTP requests.</param>
+    /// <param name="logger">Application logger</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Self instance</returns>
+    public static KernelBuilder WithOpenAI(this KernelBuilder builder,
```
lemillermicrosoft, line 113: Might be worth supporting an optional string to identify a model as the default.

lemillermicrosoft (commented):
Cool idea! Few initial comments.

lemillermicrosoft:
Love the idea @carlos-the-ai -- since it's a little stale, we're going to close this for now. Please re-open as appropriate.

PR #1234, opened 2023-05-25
DotNet Samples use NuGet packages

shawncal:
### Motivation and Context
- standalone sample projects -- if you copy a project out and run it, it should work on its own, without relative includes and references to other files/projects from the repo
- version lock -- ensures a sample will work with the currently-released NuGet packages, reducing confusion for those using these projects as references
- separation of interests -- samples are written & updated independent of the Kernel core, which will make it easier to apply a branching strategy, or even separate the repos, as the project grows

### Description
- Sample projects get their SK packages via public nuget channel, rather than project references.
  - Excluded: KernelSyntaxExamples maintains its direct ProjectReference to the Kernel, as it changes along with the Kernel core. In a future change, we'll likely move this project under the /dotnet folder so that it's closer to the (dotnet) Kernel itself.
- Sample projects no longer have relative path includes to common props files, and are self-contained
- New Directory.Build.* files are added to the samples/dotnet folder -- these are not required for the projects, but they do apply the code analyzers across all the sample projects, ensuring code quality in the repo.

### Contribution Checklist


shawncal:
The dotnet-format failures and all the failed builds (warnings raised as errors) should be fixed by this PR: https://github.com/microsoft/semantic-kernel/pull/1233
Reviewers: please review and merge that one first.

SergeyMenshykh (dismissed):

samples/dotnet/Directory.Build.props
```diff
@@ -0,0 +1,23 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <LangVersion>10</LangVersion>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <IsPackable>disable</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
```
glahaye, line 17: So you want your PDB's to be in Windows format when we are in release and in multi-platform format when we are in debug???

I'm not sure I understand the intent here...

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation

shawncal: Good question. This was copied from the other project, so there's no change here... but that doesn't necessarily mean it was correct before, or should be ported. I'll read the docs.

shawncal: Here's my understanding... the reason we have DebugType=full (and any debugging symbols at all in 'Release') may be that the dotnet analyzers use them, and we've turned on dotnet format to run in Release builds only.

Some of the analyzers cannot use the "portable" format, hence the difference from 'Debug' builds:
> Portable PDBs can be read on any operating system, but there are a number of places where they aren't supported yet. Here are a few –
> ...
> - C# Code analysis (aka FxCop), note that this doesn't apply to Roslyn Analyzer.

This is consistent with the SK dotnet sdk: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/Directory.Build.props

And the CopilotChatWebApi: https://github.com/microsoft/semantic-kernel/blob/main/samples/apps/copilot-chat-app/webapi/CopilotChatWebApi.csproj

samples/dotnet/KernelHttpServer/KernelHttpServer.csproj
```diff
@@ -14,23 +7,21 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>disable</ImplicitUsings>
     <IsPackable>false</IsPackable>
+    <LangVersion>10</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
```
glahaye, line 14: Wouldn't these version numbers belong more in the Directory.Packages.props file?

samples/dotnet/Directory.Build.props
glahaye: I must confess I'm not sure what to think about this. But don't let this block this PR. I mean, it worked before so it should work now too. I'm just a bit confused about this. Then again, I'm confused about many things in this world.

samples/dotnet/KernelHttpServer/KernelHttpServer.csproj
shawncal: Normally, yes. In the /samples/ though, we want to allow developers to copy a sample folder out of the repo and run it on its own, with all dependencies self-contained or available via nuget. In the Directories.* files under samples, we're only really adding analyzers, which are optional and help standardize formatting across the repo (but someone copying the project out and experimenting with it on its own, probably doesn't need all of our analyzers). 

LangVersion, on the other hand, if not specified, will actually break the build, which is why it's specified directly in each project. On that note, I will remove these props from the common props so that they're not accidentally omitted from future samples.

glahaye (dismissed):

samples/dotnet/Directory.Build.props
shawncal: Update: I've been testing various dotnet format violations without the above tags on release, and it doesn't seem to make any difference. That leads to second theory, which is that these may have been included to ensure the symbols made it into (released) nuget packages -- which of course does not apply to samples. I've removed that whole chunk from both this project and copilot chat, and I'll experiment further with sdk projects (ensure that .snupkg files are generated properly).

glahaye (approved):

1983Thai:
9faa7e935007d8628035c95f009d51ee35b13ab2

PR #1241, opened 2023-05-26
Make ContextVariables an actual dictionary

stephentoub:
### Motivation and Context

Make ContextVariables feel more like the dictionary it is.

### Description

ContextVariables waddles like a dictionary and quacks like a dictionary but for some reason isn't a dictionary.  This makes it one.  For now I added the rest of the IDictionary surface area explicitly such that it won't show up on the public surface area of ContextVariables, but we should revisit some members, like Count.  This also renames Get to TryGetValue, as that's the expected name in .NET for a method with its semantics, and it fixes the nullable behavior of the out param when it returns false.

It was also possible previously for someone to Set("input", null), in which case the Input property would return null even though the rest of the type is obviously trying to avoid that.  I changed Input to handle it.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
@@ -4,10 +4,16 @@ using System;
 using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
-using System.Linq;
+using System.Diagnostics.CodeAnalysis;
 using Microsoft.SemanticKernel.Diagnostics;
 using Microsoft.SemanticKernel.Security;
+
+#pragma warning disable CA1710 // ContextVariables doesn't end in Dictionary or Collection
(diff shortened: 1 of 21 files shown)


dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -88,22 +95,41 @@ public void Set(string name, string? value)
...
 
     /// <summary>
-    /// Fetch a variable value from the context variables.
+    /// Create a copy of the current instance with a copy of the internal data
     /// </summary>
-    /// <param name="name">Variable name</param>
-    /// <param name="value">Value</param>
-    /// <returns>Whether the value exists in the context variables</returns>
-    /// TODO: provide additional method that returns the value without using 'out'.
-    public bool Get(string name, out string value)
```
shawncal, line 97: Glad to see the alignment here with IDictionary!
I'm not sure how many existing callers will be using this vs the ["indexer"]... could you please throw an [Obsolete] attribute on this method to guide callers to the new TryGetValue method?

shawncal (approved):

shawncal (approved):

PR #1242, opened 2023-05-26
Python: Added examples of using ChatCompletion models for skill building in Jupyter Notebooks

alexchaomander:
### Motivation and Context
Many people run into the issue of wanting to use ChatCompletion models like gpt-35-turbo or gpt4 for building Semantic Skills but aren't aware that they need to update the API to do so. This adds some examples in the Jupyter notebooks to illustrate the concept.


samples/notebooks/python/03-semantic-function-inline.ipynb
```diff
@@ -150,27 +150,88 @@
...
+   "source": [
+    "# Using ChatCompletion for Semantic Skills"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "29b59b28",
+   "metadata": {},
+   "source": [
+    "You can also use chat completion models (like `gpt-35-turbo` and `gpt4`) for creating skills. Normally you would have to tweak the API to accomodate for a sysem and user role, but SK abstracts that away for you by using `kernel.add_chat_service` and `AzureChatCompletion` or `OpenAIChatCompletion`"
```
awharrison-28, line 166:
```suggestion
    "You can also use chat completion models (like `gpt-35-turbo` and `gpt4`) for creating skills. Normally you would have to tweak the API to accommodate for a sysem and user role, but SK abstracts that away for you by using `kernel.add_chat_service` and `AzureChatCompletion` or `OpenAIChatCompletion`"
```

awharrison-28 (approved):
Small, non-blocking nit: There's an empty box at the end of Notebook 3

awharrison-28 (approved):

PR #1243, opened 2023-05-26
Add connector unit test base

JohnMasen:
### Motivation and Context

  1. Why is this change required? _The unit test project for memory connectors lack of a unified test collection. This makes unit tests for memory connectors hard to create._
  2. What problem does it solve? _Provides a unified test class base. User can easily create memory connector unit tests for their own connector._
  3. What scenario does it contribute to? _Memory connector unit test._
  4. If it fixes an open issue, please link to the issue here. _N/A_

### Description
This PR addes a base class for memory connector unit tests. 
### Simple implementation:
```c#
public class MyTestClass : MemoryStorageTestBase
{
    protected override Task<IMemoryStore> CreateStoreAsync()
    {
        return Task.FromResult(new MyMemoryStore());
    }
}
```

Or you can add your own cleanup code like this:

```C#
public class MyTestClass : MemoryStorageTestBase
{
    protected override Task<IMemoryStore> CreateStoreAsync()
    {
        //your init code
        myMemoryStore=...
        //return the result
        return Task.FromResult<IMemoryStore>(myMemoryStore);
    }
    protected override Task CloseStoreAsync(IMemoryStore memoryStore)
    {
        //your clean up code here
        memoryStore.Dispose();
    }
}

```

### Contribution Checklist


dotnet/src/Connectors/Connectors.UnitTests/Memory/MemoryStorageTestBase.cs
```diff
@@ -0,0 +1,696 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.AI.Embeddings;
+using Microsoft.SemanticKernel.Memory;
+using Xunit;
+
+namespace SemanticKernel.Connectors.UnitTests.Memory;
+public abstract class MemoryStorageTestBase
```
dmytrostruk, line 14: Extra space needed :)
```suggestion

public abstract class MemoryStorageTestBase
```

dotnet/src/Connectors/Connectors.UnitTests/Memory/MemoryStorageTestBase.cs
```diff
@@ -0,0 +1,696 @@
...
+        return $"Collection_{Guid.NewGuid()}";
+    }
+
+    /// <summary>
+    /// Create <see cref="IMemoryStore"/> instance for unit tests
+    /// </summary>
+    /// <remarks>
+    /// you can override <see cref="CloseStoreAsync"/> in case you need extra clean up control
+    /// </remarks>
+    /// <returns><see cref="IMemoryStore"/> The instance for tests</returns>
```
dmytrostruk, line 27:
```suggestion
    /// <summary>
    /// Creates <see cref="IMemoryStore"/> instance for unit tests
    /// </summary>
    /// <returns>The instance of <see cref="IMemoryStore"/> for tests</returns>
```

dotnet/src/Connectors/Connectors.UnitTests/Memory/MemoryStorageTestBase.cs
```diff
@@ -0,0 +1,696 @@
...
+    protected abstract Task<IMemoryStore> CreateStoreAsync();
+
+    /// <summary>
+    /// Overide to add cleanup code for your <see cref="IMemoryStore"/> instance after unit test
+    /// </summary>
+    /// <param name="memoryStore"></param>
+    /// <returns></returns>
```
dmytrostruk, line 34:
```suggestion
    /// <summary>
    /// Cleans resources after unit-test execution.
    /// </summary>
    /// <param name="memoryStore">Instance of <see cref="IMemoryStore"/></param>
```

dotnet/src/Connectors/Connectors.UnitTests/Memory/MemoryStorageTestBase.cs
```diff
@@ -0,0 +1,696 @@
...
+    {
+        return Task.CompletedTask;
+    }
+
+    protected virtual async Task WithStorageAsync(Func<IMemoryStore, Task> factAsync)
+    {
+        var store = await this.CreateStoreAsync();
+        await factAsync(store);
+        await this.CloseStoreAsync(store);
+    }
```
dmytrostruk, line 45: Is this method actually needed? I think it complicates actual unit test, since unit test body will always be nested block of `.WithStorageAsync` method. Also, in xUnit it is possible to use class constructor for resource creation and implement `IDisposable` for resource clean-up. Which means that for storage creation/clean-up we can use dotnet native approach.

dmytrostruk (changes requested):
@JohnMasen thank you for contribution and sorry for long response. 

I left a couple of comments, but in general we need to think what advantages this class will give us for current test architecture. 

For example, this class unifies Memory logic into one format, which is great, but on the other hand, it's really hard to align all Memory implementations into one format, since vector DBs have very different APIs. In fact, we are not really sure that `IMemoryStore` interface will cover all cases, so we may introduce new interfaces in the future (or other approaches) to make it more extensible. In this case, unit tests for Memory implementations will be different as well.

But, even if this class aligns to our current Memory tests (we have unit tests for DuckDB, Pinecone, Qdrant, Sqlite), it would be nice to see this class in action and use it to test our Memory implementations which we have today (you can add this in current PR). In case it's possible and a lot of logic will be extracted from current Memory tests to this abstract base class and re-used - that would be great!

Thanks again!

dotnet/src/Connectors/Connectors.UnitTests/Memory/MemoryStorageTestBase.cs
JohnMasen: Hi @dmytrostruk :
The reason I use the method is to avoid use a protected field in the class. to init with class constructor, I have to define an internal field like "protected IMemoryStore MyStore;" and init it in the constructor in the inhertited instance.

JohnMasen:
@dmytrostruk 
This code was created based on very early version when I was working on an improvement on CosmosDB connector. I agree IMemoryStore may not be enough for further evolution. Actully I'm testing on an improvement of the memory system on my own branch.
The new design separates Store and Embedding into 2 interfaces. the prototype uses In-memory CPU/GPU embedding search and SQLite/SQL Server for data storage. It also modified the store interface a bit to allow nested implementation. We can add an in-memory cache "memroy store" at first and chain other store (example:SQLite,Azure SQL database, Azure Blob) as inner implementation.
in my tests: 10K records takes less than 100ms in GPU embedding search what ever the storage system is.

dmytrostruk:
@JohnMasen Thank you for your pull request and contribution. While I am closing it for now, please feel free to re-open it if you have additional changes or address the PR comments.

PR #1248, opened 2023-05-26
Support chat filtering in Copilot 

teresaqhoang:
### Motivation and Context

This PR adds support for filtering chat list by conversation name
chatfiltering

### Description

Also minor change to full reset plugin state when enabled -- this resets all the auth requirements as well on disable

### Contribution Checklist


samples/apps/copilot-chat-app/webapp/src/components/chat/chat-list/ChatList.tsx
```diff
@@ -55,22 +66,74 @@ const useClasses = makeStyles({
...
+        fontSize: '20px',
+    },
 });
 
 export const ChatList: FC = () => {
     const classes = useClasses();
-    const { conversations, selectedId } = useAppSelector((state: RootState) => state.conversations);
+    const chat = useChat();
+
+    const [isFiltering, setIsFiltering] = useState(false);
```
Vivihung, line 82: Is `isFiltering== true` when `searchString !== ""` ? Do we need two local states?

samples/apps/copilot-chat-app/webapp/src/libs/useChat.ts
```diff
@@ -202,6 +207,31 @@ export const useChat = () => {
...
+                filteredConversations: filteredChats,
+            }),
+        );
+    };
+
+    const clearConversationsFilter = async () => {
+        dispatch(
+            setConversations({
+                conversations: conversations,
+                filteredConversations: conversations,
```
Vivihung, line 230: Should filteredConversations be a local component state instead of presents in app state? Also I'm curious the difference between displayedConversations vs filteredConversations.

Vivihung (commented):
Thanks! Left some questions inline.

samples/apps/copilot-chat-app/webapp/src/components/chat/chat-list/ChatList.tsx
teresaqhoang: Yeah, filtering is a different UI state, so the search string can be empty when the User first clicks the Filter icon. In this case, all conversations are shown
image

But I've removed searchString from state. I realized that doesn't need to be a controlled variable since we're controlling the filtering state, so the value will be controlled by input component mount / unmount

1983Thai: https://github.com/microsoft/semantic-kernel/pull/1248#pullrequestreview-1446913040

samples/apps/copilot-chat-app/webapp/src/libs/useChat.ts
1983Thai: https://github.com/microsoft/semantic-kernel/pull/1248#pullrequestreview-1446913040

teresaqhoang: Great recommendation! I completely overengineered this. I moved the filteredCoversations to local component state in ChatList

hathind-ms (approved):

PR #1251, opened 2023-05-27
A few minor suggested tweaks to TrustAwareString

stephentoub:
### Motivation and Context

A few recommended tweaks after glancing at the new type.

### Description

- Remove the static Empty property.  Static Empty properties in .NET are expected to be singletons, whereas this was allocating a new one on every access.
- Renamed Trusted/Untrusted to be CreateTrusted/Untrusted.  Methods typically begin with verbs. (Separately I'm not sure what benefit these static methods provide over just using the ctor.)
- Sealed the type. Presumably we don't want to deal with the possibility of someone providing a custom TrustAwareString-derived type and passing that into the innards of the system, especially with it able to provide its own interface implementations (e.g. it could replace what equality means via the interface).

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
@@ -5,29 +5,28 @@ using System;
 namespace Microsoft.SemanticKernel.Security;
 
 /// <summary>
-/// A string wrapper that carries trust information.
-/// All field are readonly.
+/// Provides an immutable string that carries trust information.
 /// </summary>
-public class TrustAwareString : IEquatable<TrustAwareString>
+public sealed class TrustAwareString : IEquatable<TrustAwareString>
 {
     /// <summary>
-    /// Create a new empty trust aware string (default trusted).
+    /// Gets a trusted empty string.
     /// </summary>
-    public static TrustAwareString Empty => new(string.Empty, true);
+    public static TrustAwareString Empty { get; } = new(string.Empty, isTrusted: true);
 
     /// <summary>
     /// Create a new trusted string.
     /// </summary>
     /// <param name="value">The raw string value</param>
     /// <returns>TrustAwareString</returns>
-    public static TrustAwareString Trusted(string? value) => new(value, true);
+    public static TrustAwareString CreateTrusted(string? value) => new(value, isTrusted: true);
 
     /// <summary>
     /// Create a new untrusted string.
     /// </summary>
     /// <param name="value">The raw string value</param>
     /// <returns>TrustAwareString</returns>
-    public static TrustAwareString Untrusted(string? value) => new(value, false);
+    public static TrustAwareString CreateUntrusted(string? value) => new(value, isTrusted: false);
 
     /// <summary>
     /// The raw string value.
--- a/dotnet/src/IntegrationTests/Security/TrustServiceTests.cs
+++ b/dotnet/src/IntegrationTests/Security/TrustServiceTests.cs
@@ -147,7 +147,7 @@ public sealed class TrustServiceTests
         );
 
         // Make this untrusted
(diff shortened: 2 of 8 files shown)


dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
```diff
@@ -8,26 +8,21 @@ namespace Microsoft.SemanticKernel.Security;
 /// A string wrapper that carries trust information.
 /// All field are readonly.
 /// </summary>
-public class TrustAwareString : IEquatable<TrustAwareString>
+public sealed class TrustAwareString : IEquatable<TrustAwareString>
 {
-    /// <summary>
-    /// Create a new empty trust aware string (default trusted).
-    /// </summary>
-    public static TrustAwareString Empty => new(string.Empty, true);
```
shawncal, line 16: I believe the intent here was a static constant:
`public static TrustAwareString Empty = new(string.Empty, true);`
...which I do thing is a good idea to have as an option. Thoughts?

dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
```diff
@@ -8,26 +8,21 @@ namespace Microsoft.SemanticKernel.Security;
 /// A string wrapper that carries trust information.
 /// All field are readonly.
 /// </summary>
-public class TrustAwareString : IEquatable<TrustAwareString>
+public sealed class TrustAwareString : IEquatable<TrustAwareString>
```
shawncal, line 11: My recommendation on the main review was to make this 
`public readonly struct TrustAwareString`
...to enforce the immutability, and for consistency with other string-containing, string-like, stringy things like
- (StringValues)[https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringvalues]
- (StringSegment)[https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringsegment]
- (StringTokenizer)[https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringtokenizer]

I'll need to go look up where that conversation went in the original PR, and whether there was a good reason for sticking with classes, but would also like to get your advice @stephentoub.

shawncal:
@rracanicci Some suggestions FYI. Please feel free to review, comment

dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
stephentoub: Yup, as long as the instance is immutable, a singleton is fine.

dotnet/src/SemanticKernel.Abstractions/Security/TrustAwareString.cs
stephentoub: Ideally it would be a readonly struct. My hesitation is if this is actually being used for security decisions. It'll depend on how we expect these to be used. As a multi field struct, it's subject to tearing, so if for example you had a static field containing ("hello", true) and and as you were reading that struct someone else in the process wrote ("world", false) to it at exactly the wrong time, you could end up reading ("hello", false). Now, there's no security boundary here, so no matter what's done here other code in the process could use unsafe code to manipulate these values whether a class or a struct; it's just about issues arising from accidental usage. If putting these things in a shared location that's read/written without additional synchronization isn't expected, we should make it a struct. Note that ConcurrentDictionary _does_ use appropriate synchronization.

shawncal: Closing comment to approve, merge this PR. I'd like to address the struct vs class thing very soon, plus look at the feasibility and general trustworthiness of a "TrustAware<T>" wrapper.

shawncal (approved):

PR #1252, opened 2023-05-27
Update 00-getting-started.ipynb

rodgarcia:
Fixing the "Object of type "SKFunctionBase" is not callable" error in the example. Using invoke method does the trick.

I'm using Python 3.10.4


1983Thai:
Hope to add comments to promote better

PR #1292, opened 2023-05-31
Add persist to Chroma on upsert or delete

### LINKED ISSUE #1289: Chroma memory resets when store is initialized
**Describe the bug**

**To Reproduce**
```python
agent_name = "OpenAI"
memories_dir = os.path.join(os.getcwd(), "agents", agent_name, "memories")
chroma_client = ChromaMemoryStore(
    persist_directory=memories_dir,
    client_settings=Settings(
        chroma_db_impl="chromadb.db.duckdb.PersistentDuckDB",  # also tried 'duckdb+parquet' but it did not work at all.
        persist_directory=memories_dir,
        anonymized_telemetry=False,
    ),
)

# After initializing ChromaMemoryStore, all memories for that are wiped rather than looking at persist dir.

memories_exist = await chroma_client.does_collection_exist_async("memories")
print(f"Memories Exist? {memories_exist}")
if not memories_exist:
    print("Creating memories collection")
    await chroma_client.create_collection_async(collection_name="memories")
    memories = await chroma_client.get_collection_async(collection_name="memories")
else:
    memories = await chroma_client.get_collection_async(collection_name="memories")
    print("Retrieved memories collection") # <- This never happens on the first attempt
```

I also tested adding a memory before trying to reinitialize and came up with the same thing.  As soon as I initialize `ChromaMemoryStore`, it says this:

```
2023-05-31 09:55:59,483 | INFO | loaded in 0 embeddings
2023-05-31 09:55:59,484 | INFO | loaded in 0 collections
2023-05-31 09:55:59,484 | INFO | Persisting DB to disk, putting it in the save folder: /home/josh/josh/Repos/AGiXT/agixt/agents/OpenAI/memories
```

**Expected behavior**
It should be initializing the connection with the persisted data instead of only saying it is persisting.

**Screenshots**
On first go around, it acts like it is creating something that never existed.
image

image
If I run the `await chroma_client.does_collection_exist_async("memories")` line, I can see it turn to true after running the 3 blocks, but if I run the one with the `ChromaMemoryStore`, it doesn't pull persisted data.

**Desktop (please complete the following information):**
 - OS: Linux (Pop!_OS 22.04)
 - IDE: VSCode
 - NuGet Package Version [e.g. 0.1.0]

**Additional context**
Maybe I'm doing this wrong, happy to be corrected if anyone has an idea.

Add a memory:
```python
from hashlib import sha256
from datetime import datetime
from semantic_kernel.memory.memory_record import MemoryRecord

description = "Test memory"
external_source_name = "Test memory"
content = "Test memory"

record = MemoryRecord(
    is_reference=False,
    id=sha256((content + datetime.now().isoformat()).encode()).hexdigest(),
    text=content,
    timestamp=datetime.now().isoformat(),
    description=description,
    external_source_name=external_source_name,  # URL or File path
    embedding=await embedder(content),
)

await chroma_client.upsert_async(
    collection_name="memories",
    record=record,
)
```

Find the memory
```python
results = await chroma_client.get_nearest_matches_async(
    collection_name="memories",
    embedding=await embedder(content),
    limit=1,
    min_relevance_score=0.1,
)
context = []
for memory, score in results:
    context.append(memory._text)
print(context)
```

My embedder function is just:
```python
OpenAITextEmbedding(
            model_id="text-embedding-ada-002",
            api_key=self.AGENT_CONFIG["settings"]["OPENAI_API_KEY"],
            log=logging,
        ).generate_embeddings_async
```

### LINKED ISSUE #1298: .NET Allow ISKFunction to return IAsyncEnumerable
> [!IMPORTANT]  
> Labeled **High** because it will not require a breaking change, but it's very important to complete by v1.0.0

Is there any way currently to allow `ISKFunction` to return an `IAsyncEnumerable`? If not is there any plan to?

Josh-XT:
- Add persist to Chroma on upsert or delete
- Fixes #1289 

### Description
Added `self._client.persist()` in the `upsert_async` and `delete_collection_async`.

### Contribution Checklist


awharrison-28 (approved):

awharrison-28 (changes requested):
There are some use cases where non-persistence could be desired, and I would like to keep the option open. I'd like to see a 'persistence' option in the __init__ code so when someone creates a ChromaMemoryStore, it looks like 

`memory = ChromaMemoryStore(persist=True)`

Then these changes could be wrapped under an if. It's a little more processing, but it's an atomic check. 

@shawncal thoughts?

Josh-XT:
> There are some use cases where non-persistence could be desired, and I would like to keep the option open. I'd like to see a 'persistence' option in the **init** code so when someone creates a ChromaMemoryStore, it looks like
> 
> `memory = ChromaMemoryStore(persist=True)`
> 
> Then these changes could be wrapped under an if. It's a little more processing, but it's an atomic check.
> 
> @shawncal thoughts?

Alternatively, could we put this based on the `persist_directory` flag being `not None`?  No reason to persist if it is None, no reason I can think of to not persist if it is set.

Josh-XT:
Updated if you want to take a look to see if that is better.

awharrison-28:
> > There are some use cases where non-persistence could be desired, and I would like to keep the option open. I'd like to see a 'persistence' option in the **init** code so when someone creates a ChromaMemoryStore, it looks like
> > `memory = ChromaMemoryStore(persist=True)`
> > Then these changes could be wrapped under an if. It's a little more processing, but it's an atomic check.
> > @shawncal thoughts?
> 
> Alternatively, could we put this based on the `persist_directory` flag being `not None`? No reason to persist if it is None, no reason I can think of to not persist if it is set.

By default the persist_directory is .chroma/ I believe, so I'm a little hesitant to base it on `not None`

awharrison-28 (approved):
I'm fine with the current changes, but I'd like additional input from @shawncal, @lemillermicrosoft, or @mkarle before merging.

Josh-XT:
> > > There are some use cases where non-persistence could be desired, and I would like to keep the option open. I'd like to see a 'persistence' option in the **init** code so when someone creates a ChromaMemoryStore, it looks like
> > > `memory = ChromaMemoryStore(persist=True)`
> > > Then these changes could be wrapped under an if. It's a little more processing, but it's an atomic check.
> > > @shawncal thoughts?
> > 
> > 
> > Alternatively, could we put this based on the `persist_directory` flag being `not None`? No reason to persist if it is None, no reason I can think of to not persist if it is set.
> 
> By default the persist_directory is .chroma/ I believe, so I'm a little hesitant to base it on `not None`

#L29

> > > There are some use cases where non-persistence could be desired, and I would like to keep the option open. I'd like to see a 'persistence' option in the **init** code so when someone creates a ChromaMemoryStore, it looks like
> > > `memory = ChromaMemoryStore(persist=True)`
> > > Then these changes could be wrapped under an if. It's a little more processing, but it's an atomic check.
> > > @shawncal thoughts?
> > 
> > 
> > Alternatively, could we put this based on the `persist_directory` flag being `not None`? No reason to persist if it is None, no reason I can think of to not persist if it is set.
> 
> By default the persist_directory is .chroma/ I believe, so I'm a little hesitant to base it on `not None`

In `__init__` we currently have `persist_directory: Optional[str] = None`.  Chroma has a default internally if it is not set, but Semantic Kernel does not appear to.

PR #1297, opened 2023-06-01
Upgrade System.Text.Json and Microsoft.Extensions.Logging versions to 7.0

stephentoub:
### Motivation and Context

Pick up bug fixes and features from these libraries in v7.

### Contribution Checklist

### DIFF
--- a/dotnet/Directory.Packages.props
+++ b/dotnet/Directory.Packages.props
@@ -10,12 +10,12 @@
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="[7.0.0, )" />
     <PackageVersion Include="Pgvector" Version="0.1.3" />
     <PackageVersion Include="System.Linq.Async" Version="[6.0.1, )" />
-    <PackageVersion Include="System.Text.Json" Version="[6.0.0, )" />
+    <PackageVersion Include="System.Text.Json" Version="[7.0.2, )" />
     <PackageVersion Include="Azure.AI.OpenAI" Version="[1.0.0-beta.5, )" />
     <!-- Microsoft.Extensions.Logging -->
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="[6.0.0, )" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="[6.0.0, )" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="[6.0.0, )" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="[7.0.0, )" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="[7.0.0, )" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="[7.0.0, )" />
     <!-- Microsoft.Extensions.Configuration -->
     <PackageVersion Include="Microsoft.Extensions.Configuration" Version="7.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="7.0.4" />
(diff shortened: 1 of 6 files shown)


shawncal (approved):

kirk-marple:
@stephentoub Just an FYI, I tried updating to recent version of SemanticKernel, and this new dependency on System.Text.Json 7.x.x now causes downstream problems because we have to stay on .NET 6 because Azure Durable Entities isn't supported in .NET 7.

Not sure if System.Text.Json 7.x.x is a true requirement for SK, but note, there are compatibility issues with using .NET 7 packages in a .NET 6 env.

stephentoub:
> have to stay on .NET 6 because Azure Durable Entities isn't supported in .NET 7

I'm not clear on the problem; you can use the System.Text.Json 7.0 nuget package on .NET 6. Can you elaborate on what breaks?

Woland2k:
We also get a breaking change due to our project targeting .net 6 and now there is a dependency on Microsoft.Extensions.Logging.Abstractions, Version=7.0.0.0

'
Could not load file or assembly 'Microsoft.Extensions.Logging.Abstractions, Version=7.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
'

stephentoub:
Can you share more details on why it's breaking? You can use the v7 versions of these nugets from .NET 6, from a netstandard20 library, from .NET Framework, etc. Are you pinning the v6 versions?  What's preventing them from moving to latest?  They have the same numbering scheme as .NET and are released on the same cadence, but the v7 versions of these packages are not tied to .NET 7.
cc: @shawncal

Woland2k:
Our base platform has a direct dependency on Microsoft.Extensions.Logging.Abstractions 6.0.0.0. AI Module is an extension added to the platform. The problem with .NET 7 is that it is not LTE release and our next major update will be to .NET 8.

stephentoub:
Thanks for the information. We're discussing.

I'm still not clear on where the "breaking" aspect is coming from, from a technical perspective.  Is it that you're pinning the v6 dependency so the v7 one required by SK isn't getting deployed, hence it not being found at run time?  And you're pinning it because you don't want the support policy for that one component impacted?

stephentoub:
Also, note that even if the core assembly had its dependency version dropped back down, many of the connectors depend on components that in turn rely on 7.0 versions of the nugets.  For example, the Redis connector depends on NRedisStack which depends on the v7 package of System.Text.Json.  The Postgres connector depends on the pgvector nuget which depends on npgsql which depends on the v7 nugets of several Microsoft.Extensions.*/System.* nugets. Etc.

kirk-marple:
> > have to stay on .NET 6 because Azure Durable Entities isn't supported in .NET 7
> 
> I'm not clear on the problem; you can use the System.Text.Json 7.0 nuget package on .NET 6 (you can use it on .NET Framework for that matter). Can you elaborate on what breaks?

I don't know the technical reason for it, but if I upgrade any packages to the 7.x version, I can't load my Azure Functions.  It fails on startup in the in-proc function.  It complains in Serilog about something missing (sorry, I'd have to go back and repro to be more specific; it was either a missing method or assembly).

This has been a problem for us since the .NET 7.x assemblies came out, and we've been unable to take any Nuget packages that have a dependency on anything .NET 7.x.

Whenever we get an Azure Functions failure to load after upgrading Nuget packages, I've had to look thru the assets.json file and figure out what's pulling in something 7.x and revert the change.

I believe this System.Text.Json problem is just an example of this problem; I've had it happen on other Microsoft.* assemblies before, I think.

Separately I've wanted to get on .NET 7, but until Azure Durable Entities supports it, we can't move up.

stephentoub:
@kirk-marple, thanks.  Any chance you have a minimal repro you can share?  We'd like to get to the bottom of it.

kirk-marple:
> @kirk-marple, thanks. Any chance you have a minimal repro you can share? We'd like to get to the bottom of it.

Sure, I'll try over the next week; we just shipped our private preview, and have a little breathing room now.

I'd like to use Semantic Kernel in the future, so it would be good to get this sorted out for us too.

PR #1302, opened 2023-06-01
Copilot Chat: Make message input box input persistent per chat within the UI 

TaoChenOSU:
### Motivation and Context

Bug: When you enter something into the message input box in Copilot Chat but do not send it, and you switch to another chat session, the message will appear in the other chat session's message input box. 
Expected: You type a message in one of the chat sessions, the message should only appear in the input box of that chat session. And when you switch to another session, the input box should become empty or show whatever is already in there previously. 

### Description

1. Remove the feature that enables users to go back to the previous message by pressing the up arrow key.
2. Fix the bug described above to enable the expected behavior.

### Contribution Checklist

### DIFF
--- a/samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
+++ b/samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
@@ -13,6 +13,7 @@ import { DocumentImportService } from '../../libs/services/DocumentImportService
 import { useAppDispatch, useAppSelector } from '../../redux/app/hooks';
 import { RootState } from '../../redux/app/store';
 import { addAlert } from '../../redux/features/app/appSlice';
+import { editConversationInput } from '../../redux/features/conversations/conversationsSlice';
 import { SpeechService } from './../../libs/services/SpeechService';
 import { TypingIndicatorRenderer } from './typing-indicator/TypingIndicatorRenderer';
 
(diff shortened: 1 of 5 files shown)


samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -168,19 +172,15 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
...
                     onKeyDown={(event) => {
                         if (event.key === 'Enter' && !event.shiftKey) {
                             event.preventDefault();
                             handleSubmit(value);
-                            return;
-                        } else if (value === '' && previousValue !== '' && event.key === 'ArrowUp') {
-                            event.preventDefault();
-                            setValue(previousValue);
-                            return;
                         }
```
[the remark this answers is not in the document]
teresaqhoang: Just easy access to inputs instead of copy and paste lol but definitely more overhead than it's worth

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -168,19 +172,15 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
                             setValue(chatInput.value);
                         }
                     }}
-                    onChange={(_event, data) => setValue(data.value)}
+                    onChange={(_event, data) => {
+                        setValue(data.value);
+                        dispatch(editConversationInput({ id: selectedId, newInput: data.value }));
```
teresaqhoang, line 177: Might be a good idea to implement a debounce here so it's not dispatching an action on every keydown (updating the UI can be expensive). 

Declare ` const keyStrokeTimeout = useRef(-1);` before the return statement 

```suggestion
      window.clearTimeout(keyStrokeTimeout.current);
      keyStrokeTimeout.current = window.setTimeout(() => {
            dispatch(editConversationInput({ id: selectedId, newInput: data.value }));
      }, 300);
```

teresaqhoang: Actually it doesn't matter since the conversation.input property isn't tracked by this component so it doesn't have an effect on component state / re-rendering

teresaqhoang (approved):

teresaqhoang (approved):

adrianwyatt (approved):

PR #1307, opened 2023-06-01
CopilotChat: Add Files tab

dehoward:
### Motivation and Context

this PR adds a Files tab to Copilot Chat that allows users to view the documents that have been uploaded to the Bot. in the future we will ideally have more supported file types on this list with Word/Office support being next on my list.

### Preview
image

### Description

changes:
- adds a `MemorySource` model that represents the uploaded document
- changes view in `ChatWindow` from `ChatRoom` to the new `ChatResourceList` depending on the selected tab
- changes `UpdateAsync()` to `UpsertAsync()` in `IStorageContext`. previously, `CosmosDbContext` was
using upsert while the other two were updating, whereas now we upsert
across all three. this is used to prevent duplicate file upload.

### Contribution Checklist


samples/apps/copilot-chat-app/webapi/CopilotChat/Models/DocumentImportForm.cs
```diff
@@ -42,4 +42,9 @@ public enum DocumentScopes
     /// Will be use to validate if the user has access to the chat session.
     /// </summary>
     public string UserId { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Name of the user who sent this file.
+    /// </summary>
+    public string UserName { get; set; } = string.Empty;
```
Vivihung, line 49: To my understanding, UserName is PII. The team might have concerns about storing this info without proper data handling policy implemented. 
A potential way to workaround this is to call Graph API to get user info. However, if the user is an MSA account, we haven't found a way to get user infor from MSA identity provider. You can talk to @TaoChenOSU to get his learnings from multi-user chat.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -171,11 +176,29 @@ public async Task<IActionResult> EditChatSessionAsync([FromBody] ChatSession cha
...
+        await this._chatSessionRepository.UpsertAsync(chat);
 
         return this.Ok(chat);
     }
 
+    /// <summary>
+    /// Service API to get a list of imported documents.
+    /// </summary>
+    [Authorize]
+    [Route("chatSession/{chatId:guid}/documents")]
```
Vivihung, line 188: `sources` will be a better route than `documents`, IMO. (Sorry, I know it is my code :) )

Vivihung (commented):
Thanks for the PR! Leave some comments inline. (btw, I'm biased to my code so we should get feedback from others as well.)

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/DocumentImportForm.cs
adrianwyatt: Correct - we should not be storing usernames or email addresses directly.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -28,23 +29,27 @@ public class ChatHistoryController : ControllerBase
     private readonly ChatSessionRepository _chatSessionRepository;
     private readonly ChatMessageRepository _chatMessageRepository;
     private readonly PromptsOptions _promptOptions;
+    private readonly ChatMemorySourceRepository _chatMemorySourceRepository;
```
adrianwyatt, line 32: [nit] This could be simplified to `_memorySourceRepository` or even `_sourceRepository`.

adrianwyatt: Same [nit] for the classname.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -110,7 +113,18 @@ private enum SupportedFileType
...
+                documentImportForm.ChatId.ToString(),
+                formFile.FileName,
+                documentImportForm.UserName,
+                SourceType.File,
+                existingMemorySource?.Id,
+                null);
+
+            await this._chatMemorySourceRepository.UpsertAsync(newMemorySource);
+
+            await this.ParseDocumentContentToMemoryAsync(kernel, fileContent, documentImportForm, newMemorySource.Id);
```
adrianwyatt, line 127: Consider saving the memory source to the sourceRepo only after the ParseDocument has passed/completed, incase the parsing fails.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -110,7 +113,18 @@ private enum SupportedFileType
...
         }
         catch (Exception ex) when (ex is ArgumentOutOfRangeException)
```
adrianwyatt, line 129: this could be simplifie to `catch (ArgumentOutOfRangeException ex)`

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/DocumentImportForm.cs
adrianwyatt: I recommend letting this be a runtime translation from the userId through the security provider, such as AAD. We don't want to store identifiable information if we can help it.

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/MemorySource.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Text.Json.Serialization;
+using SemanticKernel.Service.CopilotChat.Storage;
+
+namespace SemanticKernel.Service.CopilotChat.Models;
+
+public enum SourceType
```
adrianwyatt, line 9: [nit] Could you stick this inside the `MemorySource` class since that's the primary consumer?

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/MemorySource.cs
```diff
@@ -0,0 +1,79 @@
...
+    public DateTimeOffset UpdatedOn { get; set; }
+
+    /// <summary>
+    /// Empty constructor for serialization.
+    /// </summary>
+    public MemorySource()
+    {
+    }
+
+    public MemorySource(string chatId, string name, string sharedBy, SourceType type, string? id, Uri? hyperlink)
```
adrianwyatt, line 69: Do we need these constructors with all the default values you've provided?

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/DocumentImportForm.cs
dehoward: yes, this actually came up in the multi-user chat discussion earlier today. I decided to go this route since we are already doing the [same for ChatMessage](https://github.com/microsoft/semantic-kernel/blob/1a49468f5306060cf821d24f3a594d4819e86111/samples/apps/copilot-chat-app/webapi/CopilotChat/Models/ChatMessage.cs#LL53C5-L53C41). 
@TaoChenOSU is currently showing GUIDs for the user's name and is considering removing the column for now I believe.

I think that if we're not comfortable with saving the username here, which makes sense since we want to move away from this pattern, I can remove the column as well.

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -113,6 +115,7 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
                 SetDocumentImporting(true);
                 await documentImportService.importDocumentAsync(
                     account!.homeAccountId!,
+                    account!.username,
```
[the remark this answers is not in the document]
dehoward: ah you're right, I thought this was in line with what was used for `ChatMessage` but I see that a [fallback](https://github.com/microsoft/semantic-kernel/blob/1a49468f5306060cf821d24f3a594d4819e86111/samples/apps/copilot-chat-app/webapp/src/libs/useChat.ts#LL110C28-L110C63) is used (`value: account?.name ?? account?.username!`). I will align the two with a helper

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
dehoward: should I change the others as well you think? wondering if we aim for consistency or correctness

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: fair point. I had used this approach to avoid having loose memory in the kernel if the save to storage failed for some reason... I could add a fallback here that deletes the upserted document (as long as it didn't already exist) if the parsing fails

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/MemorySource.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// The user name who shared the source.
+    /// </summary>
+    [JsonPropertyName("sharedBy")]
+    public string SharedBy { get; set; } = string.Empty;
+
+    /// <summary>
+    /// When the source is updated in the bot.
+    /// </summary>
+    [JsonPropertyName("updatedOn")]
+    public DateTimeOffset UpdatedOn { get; set; }
```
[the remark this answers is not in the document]
teresaqhoang: I have a slight preference towards `CreatedOn` because imo it doesn't make sense to have an `UpdateOn` property without a property that tracks the initial creation timestamp and we don't really want / need to track both values. Also, regardless of whether the source already exists in memory, we're creating a new MemorySource object every time a document is imported and overriding the stored value with this new object.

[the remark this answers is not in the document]
teresaqhoang: However, I don't think the default behavior should be to override an existing memory source if one already exists. I'll talk more to this in a comment on the DocumentImportController file

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/DocumentImportForm.cs
dehoward: this is what it looks like without the username, just for now at least:
image

after this PR I can take a look at fetching using the user's ID if no one is doing so yet

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
dehoward: I changed it 😄

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/MemorySource.cs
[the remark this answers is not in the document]
dehoward: > Member

right this makes sense, regardless of if we override existing ones or not, we're still technically creating a new source every time so `createdOn` does make sense

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -110,7 +113,18 @@ private enum SupportedFileType
                     return this.BadRequest($"Unsupported file type: {fileType}");
             }
 
-            await this.ParseDocumentContentToMemoryAsync(kernel, fileContent, documentImportForm);
+            var existingMemorySource = (await this._chatMemorySourceRepository.FindByNameAsync(formFile.FileName)).FirstOrDefault();
```
teresaqhoang: Is the name the unique identifier for each source?

imo I think we should treat each uploaded document as its own artifact, especially since we're saving the content of each document to memory and the bot can still reference information that may have been lost between 2 versions in its response. 

Also thinking about the multiuser scenario @TaoChenOSU's about to light up, two users could upload 2 separate documents with the same name and they could be different files

teresaqhoang: + we're not offering any kind of version control for these sources so I don't think we should allow overrides

dehoward: yes, the name is currently the unique identifier. there was a to-do item from the handoff with Vivien to support upsert instead of create, and I took the same approach that Teams does in that you can only have one document with the same name. 

I think your point is definitely a valid one tho, especially given the multiuser case and that each document can have different data. I can make the change to support allowing duplicate naming hrere

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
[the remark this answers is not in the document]
teresaqhoang: Yep, we need this fallback because name isn't a required property for Personal Microsoft accounts, so when it's null, it causes copilot to break down. Another option is to default this to 'Unknown' rather than falling back to email but I think email is a better option because it's primary purpose is to identify the user in the chat. 

There's a couple of instances where this fallback happens:
https://github.com/search?q=repo%3Amicrosoft%2Fsemantic-kernel%20account%3F.name%20%3F%3F%20account%3F.username&type=code

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const useClasses = makeStyles({
+    root: {
+        ...shorthands.margin(tokens.spacingVerticalM, tokens.spacingHorizontalM),
+    },
+    table: {
+        backgroundColor: tokens.colorNeutralBackground1,
+    },
+    tableHeader: {
+        fontWeight: '600',
```
teresaqhoang, line 28: nit
```suggestion
        fontWeight: tokens.fontSizeBase600,
```

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }, [chatId]);
+
+    const items = resources.map((item) => ({
+        name: {
+            label: item.name,
+            icon: getFileIconByFileExtension(item.name),
+            url: item.hyperlink,
+        },
+        updatedOn: {
+            label: new Date(item.updatedOn).toLocaleTimeString([], {
```
teresaqhoang: If the timestamp isn't today, only show the date instead of time

where date = new Date(item.updatedOn);

```suggestion
            label: date.toDateString() !== new Date().toDateString() ? date.toLocaleDateString([], {
            month: 'numeric',
            day: 'numeric',
        }) : date.toLocaleTimeString([], {
        hour: '2-digit',
        minute: '2-digit',
    })
```

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
```diff
@@ -0,0 +1,104 @@
...
+            url: item.hyperlink,
+        },
+        createdOn: {
+            label: new Date(item.createdOn).toLocaleTimeString([], {
+                hour: '2-digit',
+                minute: '2-digit',
+            }),
+            timestamp: item.createdOn,
+        },
+        sharedBy: { label: item.sharedBy, status: 'available' },
```
teresaqhoang, line 61: Persona isn't needed here, just need a label with text only

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import {
```
teresaqhoang, line 3: In general, the design with Ruokan's Figma
https://www.figma.com/file/9U23g6fAARTvJMSB2drvHU/SK_Bot_Interactions?type=design&node-id=917-62001&t=KUz1fPRA392S1ON1-0

(Unless you're referencing another design)

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
```diff
@@ -80,6 +86,19 @@ const useClasses = makeStyles({
...
+    },
+    contentOuter: {
+        height: '100%',
+        display: 'flex',
+        flexDirection: 'row',
+        justifyContent: 'center',
+    },
+    contentInner: {
+        width: '100%',
+    },
```
teresaqhoang, line 101: Did you add content/contentOuter/contentInner back in explicitly? I removed them in another PR because they were unnecessary and the styles should be handled within the children components, but if you need them, it's fine, just want to make sure it's not from an outdated merge

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
```diff
@@ -111,6 +130,11 @@ export const ChatWindow: React.FC = () => {
         setIsEditing(!isEditing);
     };
 
+    const [selectedValue, setSelectedValue] = React.useState<TabValue>('chat');
```
teresaqhoang, line 133: nit: change spelling to align with object being selected
```suggestion
    const [selectedTab, setSelectedTab] = React.useState<TabValue>('chat');
```

teresaqhoang (changes requested):

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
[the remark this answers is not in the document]
dehoward: I changed this to be the `userId` and for now this column isn't shown. I noted in a different comment that I think @TaoChenOSU is taking the same approach for multiuser for now until we have a unified way of retrieving the name via ID.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: added a fallback. can you take another look>

teresaqhoang:
Awesome first PR!!! Left some slight commments

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/DocumentImportForm.cs
Vivihung: Good call

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
dehoward: not needed anymore since I removed this column for now until we can properly fetch for the name instead of storing it

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
dehoward: correct, here's a screenshot of the design:
image

looks like sorting isn't currently implemented so I can add that in.

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
dehoward: not intentional. I will remove these.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -25,26 +26,30 @@ namespace SemanticKernel.Service.CopilotChat.Controllers;
...
     private readonly PromptsOptions _promptOptions;
+    private readonly ChatMemorySourceRepository _sourceRepository;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="ChatHistoryController"/> class.
     /// </summary>
     /// <param name="logger">The logger.</param>
     /// <param name="chatSessionRepository">The chat session repository.</param>
     /// <param name="chatMessageRepository">The chat message repository.</param>
+    /// <param name="chatMemorySourceRepository">The chat memory resource repository.</param>
```
TaoChenOSU, line 40: The parameter names in the XMLDoc do not match the actual parameter names. I am surprised that the IDE didn't catch this.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -164,18 +169,36 @@ public async Task<IActionResult> EditChatSessionAsync([FromBody] ChatSession cha
...
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<ActionResult<IEnumerable<MemorySource>>> GetSourcesAsync(
+        [FromServices] IKernel kernel,
+        Guid chatId)
+    {
+        this._logger.LogInformation("Get imported sources of chat session {0}", chatId);
+
+        return this.Ok(await this._sourceRepository.FindByChatIdAsync(chatId.ToString()));
```
TaoChenOSU, line 199: Will this return 400 or 404? Should the API check if the chat exists in the source repository first?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -110,9 +114,26 @@ private enum SupportedFileType
                     return this.BadRequest($"Unsupported file type: {fileType}");
             }
 
-            await this.ParseDocumentContentToMemoryAsync(kernel, fileContent, documentImportForm);
+            var memorySource = new MemorySource(
+                documentImportForm.ChatId.ToString(),
```
TaoChenOSU, line 118: Documents uploaded to the global scope will have an empty ChatID. Looks like you are not handling those documents in the `GetSourcesAsync` API.

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
dehoward: done. created a helper fn to help us render this across the app

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
dehoward: neither, the list will just be empty since we're returning an array here. we could check for chat existence, but I'm not sure it's needed since it will be an extra call and it seems like an edge case the user won't typically get into?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: updated to allow duplicates.

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatResourceList.tsx
dehoward: done

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
dehoward: good catch, fixed

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: this is fine since `ChatId` defaults to `string.Empty` in this case and the files can be retrieved with `GetSourcesAsync(kernel, "")`, which we can say is a special case.

we currently don't have any UI/designs for showing this

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
TaoChenOSU: I believe it's still valuable to return a 404. Sometimes developers will restart the service without refreshing the frontend while using volatile storage, which will result in sending a request with an outdated chat id.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
TaoChenOSU: That's ok (but it's probably better to make that clear in the comments).

TaoChenOSU: In addition, global document will potentially be included in the bot responses. If we don't make it clear that the source tab doesn't display global documents, it will likely cause confusion someday.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
dehoward: good point, thanks for bringing up this case. added 404 logic (and fixed it in other routes)

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: ah I see, yeah you're right in that it could be confusing. I think we could include some small text here that notes this, what do you think? I will ping Ruokan as well to get her suggestion

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -110,9 +114,26 @@ private enum SupportedFileType
...
+            await this._sourceRepository.UpsertAsync(memorySource);
+
+            try
+            {
+                await this.ParseDocumentContentToMemoryAsync(kernel, fileContent, documentImportForm, memorySource.Id);
+            }
+            catch (Exception exception)
+            {
+                await this._sourceRepository.DeleteAsync(memorySource);
+                throw exception;
```
adrianwyatt, line 133: This will reset the callstack - I recommend just calling throw.
```suggestion
                throw;
```

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -110,9 +114,26 @@ private enum SupportedFileType
...
+                MemorySourceType.File,
+                null);
+
+            await this._sourceRepository.UpsertAsync(memorySource);
+
+            try
+            {
+                await this.ParseDocumentContentToMemoryAsync(kernel, fileContent, documentImportForm, memorySource.Id);
+            }
+            catch (Exception exception)
```
adrianwyatt, line 130:
```suggestion
            catch (Exception ex) when (!ex.IsCriticalException())
```

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
adrianwyatt: I would recommend # 1 for simplicity. We can expand later if need be.

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: agreed. Ruokan is creating a quick mockup for this and in the meantime I've added a column for what I think the text _could_ be
image

dehoward: incorporating Ruokan's feedback, "Scope" has been changed to "Access" and also "All chats" to "Global" from the above screenshot.

adrianwyatt (approved):

teresaqhoang (dismissed):

TaoChenOSU (approved):

teresaqhoang (approved):

adrianwyatt (approved):

PR #1308, opened 2023-06-01
CopilotChat: Remove unused /invoke API from webapi service

adrianwyatt:
### Motivation and Context
The /invoke API in webapi is no longer being used by the webapp.

### Description
- Removed the SemanticKernelController from webapp
- Removed unused `invokeSkillWithMsalToken` and `invokeSkillWithGraphToken` from `useConnectors.ts`


DavidParks8 (dismissed):

samples/apps/copilot-chat-app/webapp/src/libs/connectors/useConnectors.ts
```diff
@@ -1,55 +1,10 @@
-import { useMsal } from '@azure/msal-react';
-import { Constants } from '../../Constants';
 import { useAppSelector } from '../../redux/app/hooks';
```
teresaqhoang, line 1: I say move `getEnabledPlugins` to useChat (the only place it's called) and get rid of useConnectors.ts altogether

I created this file before we had a clear story with Plugins and Planner so there's no need for it anymore since we're not invoking any Plugins directly

Also the word Connectors in general needs to be scrubbed from the front end

teresaqhoang (changes requested):

teresaqhoang:
We can also get rid of SemanticKernelService.ts in WebApp

teresaqhoang (approved):

PR #1313, opened 2023-06-02
Fix incorrect attribute causing the Qdrant connector to break

craigomatic:
### Motivation and Context
This PR fixes a regression introduced in #1075 and includes unit tests to correctly validate that both string and integer property ids are supported.

### Description

The changes are based on the code suggested by AwesomeYuer over in #794 - the JsonNumberHandling attribute behaved in an unexpected way and was removed.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.Memory.Qdrant/Http/ApiSchema/SearchVectorsResponse.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.Qdrant/Http/ApiSchema/SearchVectorsResponse.cs
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+using System;
 using System.Collections.Generic;
+using System.Globalization;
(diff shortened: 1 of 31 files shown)


dotnet/src/Connectors/Connectors.Memory.Qdrant/Http/ApiSchema/SearchVectorsResponse.cs
```diff
@@ -55,3 +58,33 @@ private SearchVectorsResponse()
     #endregion
 }
 #pragma warning restore CA1812 // Avoid uninstantiated internal classes
+public class NumberToStringConverter : JsonConverter<string>
```
dmytrostruk, line 61: Should we make it internal? Also, it may be worth to put it in a separate file.

dmytrostruk (approved):

adrianwyatt (approved):

PR #1318, opened 2023-06-02
Copilot Chat: Use MSAL active account instead of accounts[0]

gitri-ms:
### Motivation and Context

We are currently using `accounts[0]` to access the user's MSAL account information and retrieve tokens in the Copilot Chat frontend. This works fine if only one account is logged in. However, if the user is signed in with multiple accounts, this list is not guaranteed to be in the same order each time (per MSAL documentation [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/accounts.md#notes:~:text=getAllAccounts()%20is%20not%20ordered%20and%20is%20not%20guaranteed%20to%20be%20in%20the%20same%20order%20across%20multiple%20calls)). We need to ensure we are using the currently active account for consistent behavior.

### Description

- Set the active account once the user has logged in using `handleRedirectPromise`
- Use `instance.getActiveAccount` everywhere we need to access the account

### Contribution Checklist


samples/apps/copilot-chat-app/webapp/src/libs/auth/AuthHelper.ts
```diff
@@ -54,14 +54,14 @@ const logoutRequest: EndSessionRequest = {
...
 
 const ssoSilentRequest = async (msalInstance: IPublicClientApplication) => {
-    await msalInstance.ssoSilent({ account: msalInstance.getAllAccounts()[0], scopes: Constants.msal.semanticKernelScopes });
+    await msalInstance.ssoSilent({ account: msalInstance.getActiveAccount() || undefined, scopes: Constants.msal.semanticKernelScopes });
 };
 
 const loginAsync = async (instance: IPublicClientApplication) => {
     if (Constants.msal.method === 'redirect') {
-        await instance.loginRedirect({ account: instance.getAllAccounts()[0], scopes: Constants.msal.semanticKernelScopes });
+        await instance.loginRedirect({ account: instance.getActiveAccount() || undefined, scopes: Constants.msal.semanticKernelScopes });
```
teresaqhoang, line 62: I don't think it makes sense to pass account here or in loginPopup since we're not setting the active account until the redirect completes. 

Let's remove account from these calls and bypass the user hint altogether on log in (but keep it for the token sso) https://learn.microsoft.com/en-us/azure/active-directory/develop/msal-js-sso#without-user-hint

samples/apps/copilot-chat-app/webapp/src/index.tsx
```diff
@@ -33,6 +33,12 @@ document.addEventListener('DOMContentLoaded', async () => {
         var msalInstance: IPublicClientApplication | null = null;
         if (validEnvFile) {
             msalInstance = new PublicClientApplication(AuthHelper.msalConfig);
+
+            msalInstance.handleRedirectPromise().then((response) => {
+                if (response) {
```
teresaqhoang, line 38: What's the behavior if the log-in fails? Is it still showing a banner alert?

teresaqhoang (changes requested):

samples/apps/copilot-chat-app/webapp/src/index.tsx
gitri-ms: If login fails, user stays on MSFT account login page. If they manually redirect to Copilot Chat, they will still see the Copilot Chat login view.

alliscode (approved):

teresaqhoang (approved):

PR #1323, opened 2023-06-02
Python: Add additional_metadata field to MemoryRecord and address TODOs in ChromaMemoryStore

### LINKED ISSUE #1271: Add metadata as a field in the MemoryRecord class in Python
Add metadata as a field in the python memory record class.
This would improve parity between the implementations.

The field would be a dict[str, str | int | bool | list[str]] this would allow for a flexible usage of the field and would mesh well with Pinecone's db record [metadata](https://docs.pinecone.io/docs/metadata-filtering) implementation and could be stored in other dbs as Json or JSONb 

Furthermore, the text, description and other such fields on the python memory record class could be moved to this dict as well similar to the [.net Memory Record class](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecord.cs)

awharrison-28:
### Motivation and Context
Closes #1271 and addresses several TODOs in ChromaMemoryStore

### Description
- added field `additional_metadata` to `MemoryRecord` and all downstream methods/classes (see `SemanticTextMemoryBase`, `SemanticTextMemory`, `MemoryQueryResult`, and `ChromaMemoryStore`)
- addressed ChromaMemoryStore TODO 'what is key used for?' For chromadb, key and id are effectively the same value, but the two fields are differentiated for memory stores like Qdrant which require that key be a uuid. I updated the logic so that this is more apparent. Note that this results in no change in end-user functionality.
- Updated ChromaMemoryStore TODO 'get_async say embedding is optional but Record constructor requires it.` MemoryRecord constructor no longer requires an embedding. Logic is updated and a test added to show that embedding retrieval is functionally optional. 

### Contribution Checklist


alexchaomander (dismissed):
LGTM

mkarle (approved):

Issue #1270, opened 2023-05-30
Python: Add Postgres with pgvector as a Memory Store for Python


cschadewitz:
I would like to add Postgres with pgvector as a Memory Store to the Python implementation to give parity for the addition to .net in #945

cschadewitz:
[WIP](https://github.com/microsoft/semantic-kernel/compare/main...cschadewitz:semantic-kernel:python_postgres_memory_store)

Taking a different approach from the .net implementation:
collections as postgres schemas as opposed to collection being another column in the table
currently the metadata column is just storing the text field, however, I plan to change it to jsonb and store all additional existing fields (minus key and embedding) and other metadata if #1271 was given the go ahead

1983Thai:
https://github.com/google/omaha/commit/306056c976ba29182019839fdf41410321c3765c

alexchaomander:
Thanks for helping bring this in to Python!

cschadewitz:
Absolutely, I am very excited to help get this up to par with the .net implementation. I do have a few questions though relating to implementation:
1. What is the difference/purpose of the key and id fields in the memory record, is one intended to be an app level identifier while the other is the p key in the db?
2. Is the external source intended to be the original source of the record or source from which it was last pulled?
3. What is the purpose of the different record types, reference vs local?
4. Currently, on both this [WIP](https://github.com/microsoft/semantic-kernel/compare/main...cschadewitz:semantic-kernel:python_postgres_memory_store) and the PR that I am helping with #769, the embedding dimensions are set at collection creation time (which doesnt mesh with the current MemStoreBase signature), so I could add it to the constructor but wanted to ensure that there wouldnt be a need to have the same instance work with collections containing embeddings dimensions of different sizes.
5. With regards to the use of pg schemas to delineate collections, open to opinions on that since I know it doesnt match the C# implementation.

Just trying to make sure I understand the expectations so that your team's reviewing experience is as easy as possible. Let me know any thoughts or concerns @alexchaomander

PR #1338, opened 2023-06-05
Improve postgres memory store

JadynWong:
### Motivation and Context

The metadata uses the jsonb column type to make it easier to filter later. (https://github.com/microsoft/semantic-kernel/issues/1270#issuecomment-1567970396)

For CopilotChat, use PostgresMemoryStore as SemanticTextMemory. It's better to use it as a single instance. Kernel and SemanticTextMemory both implement the Dispose pattern and are registered in the DI container with Scoped lifecycle. This means that whenever the container scoped ends (i.e., when the request is completed), PostgresMemoryStore will be disposed.

Sometimes, the same database may have other tables, and the developer may want to reuse NpgsqlDataSource globally. In such cases, it will be more appropriate for the developer to control when NpgsqlDataSource is disposed.

### Description

- Change the `metadata` from text column type to jsonb column type.
- PostgresMemoryStore no longer implements `IDisposable` mode. This is a break change, but its impact is minimal and the developer only needs to modify `ConnectAsync`.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
@@ -67,19 +67,32 @@ internal sealed class Database
     /// <returns></returns>
     public async Task CreateTableAsync(NpgsqlConnection conn, int vectorSize, CancellationToken cancellationToken = default)
     {
-        await this.CreatePgVectorExtensionAsync(conn, cancellationToken).ConfigureAwait(false);
-
         using NpgsqlCommand cmd = conn.CreateCommand();
 #pragma warning disable CA2100 // Review SQL queries for security vulnerabilities
         cmd.CommandText = $@"
             CREATE TABLE IF NOT EXISTS {TableName} (
                 collection TEXT,
                 key TEXT,
-                metadata TEXT,
+                metadata JSONB,
                 embedding vector({vectorSize}),
                 timestamp BIGINT,
                 PRIMARY KEY(collection, key))";
 #pragma warning restore CA2100 // Review SQL queries for security vulnerabilities
+        await cmd.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
+
+        await this.MigrateAsync(conn, cancellationToken).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Migrate column types from older versions.
+    /// </summary>
+    /// <param name="conn">An opened <see cref="NpgsqlConnection"/> instance.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns></returns>
+    private async Task MigrateAsync(NpgsqlConnection conn, CancellationToken cancellationToken = default)
+    {
+        using NpgsqlCommand cmd = conn.CreateCommand();
+        cmd.CommandText = $"ALTER TABLE {TableName} ALTER COLUMN metadata TYPE JSONB USING metadata::JSONB";
         await cmd.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
     }
 
@@ -107,16 +120,12 @@ internal sealed class Database
     /// <returns></returns>
     public async Task CreateCollectionAsync(NpgsqlConnection conn, string collectionName, CancellationToken cancellationToken = default)
     {
-        if (await this.DoesCollectionExistsAsync(conn, collectionName, cancellationToken).ConfigureAwait(false))
-        {
-            // Collection already exists
-            return;
-        }
-
         using NpgsqlCommand cmd = conn.CreateCommand();
         cmd.CommandText = $@"
             INSERT INTO {TableName} (collection, key)
-            VALUES(@collection, '')";
+            VALUES(@collection, '')
+            ON CONFLICT (collection, key)
+            DO NOTHING";
         cmd.Parameters.AddWithValue("@collection", collectionName);
         await cmd.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
     }
@@ -143,7 +152,7 @@ internal sealed class Database
             DO UPDATE SET metadata=@metadata, embedding=@embedding, timestamp=@timestamp";
         cmd.Parameters.AddWithValue("@collection", collectionName);
         cmd.Parameters.AddWithValue("@key", key);
-        cmd.Parameters.AddWithValue("@metadata", metadata ?? string.Empty);
+        cmd.Parameters.AddWithValue("@metadata", NpgsqlTypes.NpgsqlDbType.Jsonb, metadata ?? (object)DBNull.Value);
         cmd.Parameters.AddWithValue("@embedding", embedding ?? (object)DBNull.Value);
         cmd.Parameters.AddWithValue("@timestamp", timestamp ?? (object)DBNull.Value);
         await cmd.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
--- a/dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresMemoryStore.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresMemoryStore.cs
@@ -10,29 +10,24 @@ using Microsoft.SemanticKernel.AI.Embeddings;
 using Microsoft.SemanticKernel.Memory;
 using Npgsql;
 using Pgvector;
-using Pgvector.Npgsql;
 
 namespace Microsoft.SemanticKernel.Connectors.Memory.Postgres;
 
 /// <summary>
 /// An implementation of <see cref="IMemoryStore"/> backed by a Postgres database with pgvector extension.
 /// </summary>
-public class PostgresMemoryStore : IMemoryStore, IDisposable
+public sealed class PostgresMemoryStore : IMemoryStore
 {
     /// <summary>
     /// Connect a Postgres database
     /// </summary>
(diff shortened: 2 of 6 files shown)


cschadewitz:
@JadynWong curious about your thoughts on changing the implementation of collections to be postgres schemas as opposed to the existing collection column implementation

JadynWong:
> @JadynWong curious about your thoughts on changing the implementation of collections to be postgres schemas as opposed to the existing collection column implementation

Using schemes to differentiate data from different collections will make it clearer. In fact, my initial idea was to use separate tables to store each collection. However, I have some concerns about the length limitations and naming rules for PostgreSQL schemes.

cschadewitz:
I hadn't thought of that as a problem, 63 bytes of alpha-num and '_' seem sufficient, though you are right it is more restrictive than the text column currently in use. I would still push for schemas as it would logically separate the embeddings of different collections and reduce/delay the need for lists when indexing due to only having the rows belonging to a single collection. See [pgvector & indexing](https://github.com/pgvector/pgvector#indexing) for more details on lists and recall.

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
```diff
@@ -67,20 +67,33 @@ public async Task CreatePgVectorExtensionAsync(NpgsqlConnection conn, Cancellati
...
                 key TEXT,
-                metadata TEXT,
+                metadata JSONB,
                 embedding vector({vectorSize}),
                 timestamp BIGINT,
                 PRIMARY KEY(collection, key))";
 #pragma warning restore CA2100 // Review SQL queries for security vulnerabilities
         await cmd.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
+
+        await this.MigrateAsync(conn, cancellationToken).ConfigureAwait(false);
```
SergeyMenshykh, line 83: What is the plan to discontinue the column type migration operation?

JadynWong: I plan to remove the migration operation after the next two releases.

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
dmytrostruk, line 1: What is the reason of storing all collections in one Postgres table `sk_memory_table`? It means that we force developers to align to the same structure - one table, and collections are just basically primary key in that table. 

For example, I have my Postgres DB where I have tables like `Contacts`, `Locations` etc. It's not possible to use this DB with Semantic Kernel, because it will internally create `sk_memory_table` and operate only with it.

Is it possible to do the mapping between SK `collection` to Postgres `table`? So, when I want to create SK collection `Contacts` it will create Postgres table "Contacts` and store all my contacts there. Because now it looks like all the information will be located just in one table, I'm not sure it's a good practice. 

But maybe there is some reason specific for vector search, that's why I want to double-check it with you :)

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
```diff
@@ -107,16 +132,12 @@ public async Task CreateIndexAsync(NpgsqlConnection conn, CancellationToken canc
...
-        {
-            // Collection already exists
-            return;
-        }
-
         using NpgsqlCommand cmd = conn.CreateCommand();
         cmd.CommandText = $@"
             INSERT INTO {TableName} (collection, key)
-            VALUES(@collection, '')";
+            VALUES(@collection, '')
```
dmytrostruk, line 138: What is the purpose of adding collection with empty key? Isn't it just an empty row?

dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresMemoryStore.cs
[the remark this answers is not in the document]
JadynWong: This is indeed a problem and I will fix it.
This code was originally changed from Sqlite and I think both Sqlite and DuckDB need this change. I would like to fix the same bug in other PRs for both Sqlite and DuckDB.

https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.Memory.Sqlite/SqliteMemoryStore.cs#L91-L104


https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.Memory.DuckDB/DuckDBMemoryStore.cs#L112-L124

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
JadynWong: Because `collection` and `key` are joint primary keys, they cannot be null.

dmytrostruk: Correct, but since this implementation uses only one table (which is mentioned in my previous comment), what is the point in adding an empty row? If I use `UpsertAsync` without `CreateCollectionAsync` first, my upsert operation will succeed, is that correct?

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
JadynWong: As the previous review replied, than the memorystore implementation initially referenced the sqlite implementation. Also I did consider using a different table to store the collection. But as in the previous reply to @cschadewitz, I'm not sure how to translate the collection name entered by the user into the appropriate table name. It might be possible to use regular substring for names that don't match the rules, and interception for the right length, but I'm still concerned about the substitution or interception causing different collections to become duplicates. 
For example, "test 1", "test@1", they are all replaced with "test_1". 
Also I'm not sure if the collections may be in non-English languages and if they are supported by postgres. 

I think this needs to be a clear rule and known by the user.

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
JadynWong: Yes, records can be upsert even if the collection does not exist. Also referring to Sqlite 😅, I think it is designed to fetch already created collections when the user only creates a collection and does not insert any records.

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
dmytrostruk: I think that user should be responsible for table/collection naming. For example, if we call `kernel.Memory.SaveInformationAsync("Animals", id: "cat1", text: "british short hair")`, then Postgres table "Animals" should be created.

In case collection name doesn't follow naming rules of Postgres, then `PostgresMemoryStore` should throw an exception with detailed error description why collection was not created. We don't need to modify collection name provided by user.

If the reason of using one table with multiple collections is only because of Postgres table naming rules, then it's probably a wrong decision :) 

Making a table-collection mapping will resolve a lot of issues. The major one - not storing every memory record in one table.

JadynWong: This is clear, and I can make this change, but it is not compatible with the previous data structure. Do we need to make this change in this PR?

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
dmytrostruk: We will improve Sqlite if needed as well :)
That's why I posted my question about table-collection mapping. In case we create new table per collection, it will resolve such problems for us.

Current implementation is fine for demos and examples, but I'm not sure it will be possible to use it in production.

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
dmytrostruk: If you feel that this PR is still applicable taking into account `PostgresMemoryStore` redesign, then we can merge this one and start working on new approach. In case it's not compatible, then maybe it worth to start working on new approach without this PR? Please let me know what you think about it.

JadynWong: I think the new design structure is hardly compatible with the existing data. I will re-implement it in the new PR. This PR can be closed. 

Also I would like to ask if the new design is a huge disruptive update, can the developers know that it will not be compatible with the old version?

dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresMemoryStore.cs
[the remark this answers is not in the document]
JadynWong: If Sqlite and DuckDB will also use the new design, is a new PR needed to fix them?

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
dmytrostruk: > I think the new design structure is hardly compatible with the existing data. I will re-implement it in the new PR.

Thanks a lot! Please let me know if any help is needed. I would be happy to review new PR with new implementation. Feel free to assign it to me when it's ready.


> Also I would like to ask if the new design is a huge disruptive update, can the developers know that it will not be compatible with the old version?

We can mention it in Release Notes. Taking into account, that instead of `sk_memory_table` the new tables will be used, I don't think it's disruptive update. Users should be able to migrate their data from `sk_memory_table` to the new tables if needed. We can also provide migration script as optional solution, but don't include it automatically when using new approach.

dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresMemoryStore.cs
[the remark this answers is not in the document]
dmytrostruk: I think Sqlite and DuckDB redesign is another scope of work, please skip it for now. It should be done in separate PRs.

JadynWong:
Based on the discussion, postgres memory store will be reimplemented on new PR.

dmytrostruk:
@JadynWong thanks a lot for contribution!

PR #1340, opened 2023-06-05
Copilot Chat deployment: Add sk probe endpoint to deployment output

thegovind:
### Motivation and Context
After [deploying Copilot Chat to Azure](https://learn.microsoft.com/en-us/semantic-kernel/deploy/deploy-to-azure), the default behavior of users is to access the [`endpoint` url](https://github.com/microsoft/semantic-kernel/blob/main/samples/apps/copilot-chat-app/webapi/DeploymentTemplates/sk-new.bicep#LL60C1-L60C1) from the Deployment Output. However, this URL results in 404 (since the user has to append `/probe`) to the URI. It would be great to have that `probe` URI as part of the Deployment output, in case users miss the [instruction ](https://learn.microsoft.com/en-us/semantic-kernel/deploy/deploy-to-azure#verifying-the-deployment)

image

### Description
Add `skProbe` as an output to Deployment outputs.

cc - @glahaye 

### Contribution Checklist

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
+++ b/samples/apps/copilot-chat-app/webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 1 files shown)


adrianwyatt:
> I like this change. Can you also re-build the ARM JSON files from the bicep files so that they match (they are used by the Deploy to Azure buttons):
> 
> az bicep build -f .\sk-existing-azureopenai.bicep az bicep build -f .\sk-existing-openai.bicep az bicep build -f .\sk-new.bicep

I updated the probe endpoint to /healthz and rebuilt the ARM jsons.

adrianwyatt (approved):

PR #1345, opened 2023-06-05
Copilot Chat: Multi-user support with SignalR

TaoChenOSU:
### Motivation and Context

Brings in support for multi-user chat experience with AI to Copilot Chat. This will further show case what we can build with SK and the power of Copilot in multi-user settings.

### Description

1. Create ChatParticipant model to manage chat users.
2. Create FE and BE logic to allow users to join chats.
3. Use SignalR to achieve real time chatting experiences.
4. Documents permissions will remain in Chat level and Global level. 
5. Plans can only be approved by the user who initiates the plans.
6. Permission for bot export/import will stay the same (i.e. anyone who is part of the chat can download a bot and anyone who has the Json can upload a bot).
7. Modify the prompts.

Note: we still need to work on prompt engineering to further improve the user experience with AI.

image
image
image
image
image

### Contribution Checklist


samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -123,6 +123,14 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
...
                 );
                 dispatch(addAlert({ message: 'Document uploaded successfully', type: AlertType.Success }));
+                
+                // Broadcast file uploaded alert to other users
+                const docUploadAlert = {
+                    id: getSelectedChatID(),
+                    fileOwner: account?.name as string,
+                    fileName: documentFile.name as string,
+                };
+                dispatch(updateFileUploadedFromUser(docUploadAlert));
```
TaoChenOSU, line 133: It's better to do this from the backend. Will change.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatController.cs
```diff
@@ -61,7 +66,8 @@ public ChatController(ILogger<ChatController> logger)
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     public async Task<IActionResult> ChatAsync(
         [FromServices] IKernel kernel,
-        [FromServices] CopilotChatPlanner planner,
+        [FromServices] IHubContext<MessageRelayHub> messageRelayHubContext,
+        [FromServices] CopilotChatPlanner? planner,
```
adrianwyatt, line 70: I didn't think we could have an optional parameter coming from services. Do we need it to be nullable?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -35,42 +36,48 @@ public class ChatHistoryController : ControllerBase
...
         this._chatMessageRepository = chatMessageRepository;
+        this._chatParticipantRepository = chatParticipantRepository;
         this._promptOptions = promptsOptions.Value;
     }
 
     /// <summary>
     /// Create a new chat session and populate the session with the initial bot message.
+    /// The regex pattern that is used to match the user id will match the following format:
+    ///    - 2 period separated groups of one or more hyphen-delimited alphanumeric strings.
+    /// The pattern matches two GUIDs in canonical textual representation separated by a period.
```
adrianwyatt, line 59: Technically this matches any dash-delimited number of characters. Maybe add an example: "a1-2b-3c.x4-5y-6z"
```suggestion
    /// The pattern in meant to match two GUID-like strings in canonical textual representation separated by a period (e.g., a1-2b-3c.x4-5y-6z)
```

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -35,42 +36,48 @@ public class ChatHistoryController : ControllerBase
...
-    public async Task<IActionResult> CreateChatSessionAsync(
-        [FromBody] ChatSession chatParameters)
+    public async Task<IActionResult> CreateChatSessionAsync([FromRoute] string userId, [FromBody] ChatSession chatParameter)
     {
-        var userId = chatParameters.UserId;
-        var title = chatParameters.Title;
-
-        var newChat = new ChatSession(userId, title);
+        // Create a new chat session
+        var newChat = new ChatSession(chatParameter.Title);
```
adrianwyatt, line 71: Why are we accepting a `ChatSession` in the body and throwing it all out except the title? Could we just have a title in the route instead? Or a simple JSON with the title in the body?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -140,7 +161,7 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
     {
         // TODO: the code mixes strings and Guid without being explicit about the serialization format
         var chatMessages = await this._chatMessageRepository.FindByChatIdAsync(chatId.ToString());
-        if (chatMessages == null)
+        if (!chatMessages.Any())
         {
             return this.NotFound($"No messages found for chat id '{chatId}'.");
```
adrianwyatt, line 166: Should we just return an empty OK here, or do we consider a chat session without messages a failure?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.Logging;
+using SemanticKernel.Service.CopilotChat.Hubs;
+using SemanticKernel.Service.CopilotChat.Models;
+using SemanticKernel.Service.CopilotChat.Storage;
+
+namespace SemanticKernel.Service.CopilotChat.Controllers;
+
+/// <summary>
+/// Controller for chat invitations.
```
adrianwyatt, line 18:
```suggestion
/// Controller for managing invitations and participants in a chat session.
```

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// <param name="messageRelayHubContext">Message Hub that performs the real time relay service.</param>
+    /// <param name="chatParticipantParam">Contains the user id and chat id.</param>
+    [HttpPost]
+    [Route("chatParticipant/join")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    public async Task<IActionResult> JoinChatAsync(
+        [FromServices] IHubContext<MessageRelayHub> messageRelayHubContext,
+        [FromBody] ChatParticipant chatParticipantParam)
+    {
```
adrianwyatt, line 61: How do we know a user was invited? It looks like if they know the chatID, we just accept them?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,114 @@
...
+    [HttpGet]
+    [Route("chatParticipant/getAllParticipants/{chatId:guid}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    public async Task<IActionResult> GetAllParticipantsAsync(Guid chatId)
+    {
+        // Make sure the chat session exists.
+        try
+        {
+            await this._chatSessionRepository.FindByIdAsync(chatId.ToString());
```
adrianwyatt, line 104: This try/catch seems like a common pattern. Could we add a `ChatSesstionRepository.TryGetById` API to simplify this to a n `if` statement?

samples/apps/copilot-chat-app/webapi/CopilotChat/Hubs/MessageRelayHub.cs
```diff
@@ -0,0 +1,72 @@
...
+        this._logger = logger;
+    }
+
+    /// <summary>
+    /// Adds the user to the groups that they are a member of.
+    /// Groups are identified by the chat ID.
+    /// TODO: Retrieve the user ID from the claims and call this method
+    /// from the OnConnectedAsync method instead of the frontend.
+    /// </summary>
+    /// <param name="chatId"></param>
```
adrianwyatt, line 34: XMLDoc needs content

samples/apps/copilot-chat-app/webapi/CopilotChat/Storage/ChatParticipantRepository.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+    }
+
+    /// <summary>
+    /// Finds chat participants by user id.
+    /// A user can be part of multiple chats, thus a user can have multiple chat participants.
+    /// </summary>
+    /// <param name="userId">The user id.</param>
+    /// <returns>A list of chat participants of the same user id in different chat sessions.</returns>
+    public Task<IEnumerable<ChatParticipant>> FindByUserIdAsync(string userId)
```
adrianwyatt, line 30: Consider allowing empty return sets to be valid here. Or a TryFind pattern..

adrianwyatt: Same for `FindByChatIdAsync`

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
```diff
@@ -32,16 +35,34 @@ export const conversationsSlice: Slice<ConversationsState> = createSlice({
...
+        addUserToConversation: (state: ConversationsState, action: PayloadAction<{ user: IChatUser, chatId: string }>) => {
+            const { user, chatId } = action.payload;
+            state.conversations[chatId].users.push(user);
+        },
+        /*
+        * updateConversationFromUser() and updateConversationFromServer() both update the conversations state.
+        * However they are for different purposes. The former action is for updating the conversation from the
+        * user and will be captured by the SignalR middleware and the payload will be broadcasted to all clients
+        * in the same group.
+        * The updateConversationFromUser() action is triggered by the SignalR middleware when a response is received.
```
teresaqhoang, line 47: Still unclear why both actions are required, the middleware should be able to dispatch the same action

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/ChatState.ts
```diff
@@ -11,4 +11,11 @@ export interface ChatState {
     botProfilePicture: string;
     lastUpdatedTimestamp?: number;
     input: string;
+    isBotTyping: boolean;
```
teresaqhoang, line 14: Why are we treating the bot separately from the user? 

Bot is a user and can be identified by the UserRole

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
```diff
@@ -32,16 +35,34 @@ export const conversationsSlice: Slice<ConversationsState> = createSlice({
...
             state.selectedId = newId;
         },
-        updateConversation: (
+        addUserToConversation: (state: ConversationsState, action: PayloadAction<{ user: IChatUser, chatId: string }>) => {
+            const { user, chatId } = action.payload;
+            state.conversations[chatId].users.push(user);
+        },
+        /*
+        * updateConversationFromUser() and updateConversationFromServer() both update the conversations state.
+        * However they are for different purposes. The former action is for updating the conversation from the
```
teresaqhoang, line 44: You mention the former action here (updateConversationFromUser) and then updateConversationFromUser in line 47, are both lines meant to describe the same action?

samples/apps/copilot-chat-app/webapp/src/redux/features/message-relay/signalRMiddleware.ts
```diff
@@ -0,0 +1,173 @@
...
+import { IChatUser } from "../../../libs/models/ChatUser";
+import { IAskResult } from "../../../libs/semantic-kernel/model/AskResult";
+import { addAlert } from "../app/appSlice";
+import { AuthorRoles, ChatMessageState, IChatMessage } from './../../../libs/models/ChatMessage';
+import { isPlan } from './../../../libs/utils/PlanUtils';
+import { getSelectedChatID } from './../../app/store';
+import { FileUploadedAlert } from './../conversations/ChatState';
+
+// These have to match the callback names used in the backend
+const receiveMessageFromServerCallbackName = "ReceiveMessage" as string;
```
teresaqhoang, line 14: I think this would make more sense as an enum rather than separate constant strings

i.e., 
```
const enum ServerCallbackNames {
     ReceiveMessage = 'RecieveMesage'
    ....
}
```

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationJoinDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+        marginBottom: "20px",
+    },
+});
+
+interface InvitationJoinDialogProps {
+    onCancel: () => void;
+    onJoinComplete: () => void;
+}
+
+export const InvitationJoinDialog: React.FC<InvitationJoinDialogProps> = (props) => {
```
teresaqhoang, line 27: nit: remove line 28

```suggestion
export const InvitationJoinDialog: React.FC<InvitationJoinDialogProps> = ({ onCancel, onJoinComplete }) => {
```

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatController.cs
teresaqhoang: I'd remove it as optional if you're going to force it as a non-null in line 84

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
teresaqhoang: New chats should always have at least the initial bot message

samples/apps/copilot-chat-app/webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -136,6 +136,50 @@ public async Task<string> ExtractUserIntentAsync(SKContext context)
...
     }
 
+    /// <summary>
+    /// Extract the list of participants from the conversation history.
+    /// Note that only those who have spoken will be included.
+    /// </summary>
+    [SKFunction("Extract audience list")]
+    [SKFunctionName("ExtractAudience")]
+    [SKFunctionContextParameter(Name = "chatId", Description = "Chat ID to extract history from")]
+    public async Task<string> ExtractAudienceAsync(SKContext context)
```
teresaqhoang, line 146: Why is this necessary? How does it change the behavior if you were to just reference the users list tied to each conversation?

teresaqhoang (changes requested):
ReadMEs also need updated

And we should bug bash this as a team before it goes in

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
TaoChenOSU: Making it explicit that a chat should have at least one message.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
TaoChenOSU: Yes, that is the current design.

samples/apps/copilot-chat-app/webapi/CopilotChat/Storage/ChatParticipantRepository.cs
TaoChenOSU: I believe empty sets are valid return values already.

samples/apps/copilot-chat-app/webapp/src/App.tsx
```diff
@@ -90,6 +91,14 @@ const App: FC = () => {
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [instance, inProgress, isAuthenticated, appState]);
 
+    useEffect(() => {
+        if (isAuthenticated && account) {
+            dispatch(setLoggedInUserId(account.homeAccountId))
+        } else {
+            dispatch(setLoggedInUserId(''))
```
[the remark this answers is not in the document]
TaoChenOSU: My original thinking was that this would only be triggered when the user logs out. 
Since the useEffect hook is attached to isAuthenticated, at what point will isAuthenticated changes if the user is not logging out?

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationCreateDialog.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface InvitationCreateDialogProps {
+    onCancel: () => void;
+    chatId: string;
+}
+
+export const InvitationCreateDialog: React.FC<InvitationCreateDialogProps> = (props) => {
+    const { onCancel, chatId } = props;
+    const [isIdCopied, setIsIdCopied] = React.useState<boolean>(false);
+
+    const classes = useStyles();
```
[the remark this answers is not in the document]
TaoChenOSU: Would it make more sense to let the users decide what to put in their clipboards?

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationJoinDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const { onCancel, onJoinComplete } = props;
+    const { instance, accounts, inProgress } = useMsal();
+    const account = useAccount(accounts[0] || {});
```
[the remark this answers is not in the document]
TaoChenOSU: Shouldn't it be getActiveAccount()?

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
TaoChenOSU: I will do my best to illustrate with the sequences below:
Scenario 1: user sends a message
middleware captures the message -> update the conversation state -> broadcast the message to other users

Scenario 2: bot responds to a message
signalR receieves the event -> dispatches the response to update the conversation state

We have to create two actions for these two scenarios otherwise each bot response will create two messages in the conversation state.

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/ChatState.ts
TaoChenOSU: Two reasons:
1. The bot is currently not registered as a ChatUser in the WebApp.
2. The typing signal broadcast as a boolean. We can in the future create a new model called UserState and broadcast that instead.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
```diff
@@ -111,8 +116,9 @@ public class BotController : ControllerBase
         // Upload chat history into chat repository and embeddings into memory.
 
         // 1. Create a new chat and get the chat id.
-        newChat = new ChatSession(userId, chatTitle);
+        newChat = new ChatSession(chatTitle);
         await this._chatRepository.CreateAsync(newChat);
+        await this._chatParticipantRepository.CreateAsync(new ChatParticipant(userId, newChat.Id));
```
dehoward, line 121: out of curiosity, if this fails will the created chat be empty/dangling?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -28,54 +28,62 @@ public class ChatHistoryController : ControllerBase
...
+    /// The pattern in meant to match two GUID-like strings in canonical textual representation
+    /// separated by a period (e.g., a1-2b-3c.x4-5y-6z)
     /// </summary>
-    /// <param name="chatParameters">Object that contains the parameters to create a new chat.</param>
+    /// <param name="userId">The user ID.</param>
+    /// <param name="chatParameter">Contains the title of the chat.</param>
     /// <returns>The HTTP action result.</returns>
     [HttpPost]
-    [Route("chatSession/create")]
+    [Route("chatSession/create/{userId:regex(([[a-z0-9]]+-)+[[a-z0-9]]+\\.([[a-z0-9]]+-)+[[a-z0-9]]+)}")]
```
dehoward, line 71: can we save this in a constant above and use it in both of the routes that have this regex?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
dehoward: +1, my vote is for the 2nd option (title in body) to be closer to REST.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatHistoryController.cs
TaoChenOSU: The regex is removed and replaced by a string from the body.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/BotController.cs
TaoChenOSU: Yes, you are right! Good catch! I will fix this

samples/apps/copilot-chat-app/webapi/appsettings.json
```diff
@@ -159,17 +160,20 @@
...
+    "SystemResponse": "Either return [silence] or provide a response to the last message. If you provide a response do not provide a list of possible responses or completions, just a single response. ONLY PROVIDE A RESPONSE IF the last message WAS ADDRESSED TO THE 'BOT' OR 'COPILOT'. If it appears the last message was not for you, send [silence] as the bot response.",
     "InitialBotMessage": "Hello, nice to meet you! How can I help you today?",
     "KnowledgeCutoffDate": "Saturday, January 1, 2022",
 
+    "SystemAudience": "Below is a chat history between an intelligent AI bot named Copilot with one or more participants.",
+    "SystemAudienceContinuation": "Using the provided chat history, generate a list of names of the participants of this chat. Do not include 'bot' or 'copilot'.The output should be a single rewritten sentence containing only a comma separated list of names. DO NOT offer additional commentary. DO NOT FABRICATE INFORMATION.\nParticipants:",
```
teresaqhoang, line 169: How important is it to know the audience of the chat history? This adds another round trip and seems expensive in terms of token costs since each trip passes the chat history again.

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationJoinDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const [errorOccurred, setErrorOccurred] = React.useState<boolean>(false);
+    const [errorMessage, setErrorMessage] = React.useState<string>("");
+
+    const classes = useStyles();
+
+    const handleSubmit = async (ev: React.FormEvent<HTMLFormElement>) => { 
+        ev.preventDefault();
+        setErrorOccurred(false);
+
+        const chatId = ev.currentTarget.elements.namedItem("chat-id-input") as HTMLInputElement;
```
teresaqhoang, line 40: Would change this to track a string state value, querying DOM elements can get expensive

samples/apps/copilot-chat-app/webapp/src/components/chat/prompt-details/PromptDetails.tsx
```diff
@@ -32,8 +32,8 @@ export const PromptDetails: React.FC<IPromptDetailsProps> = ({ message }) => {
                 <DialogBody>
                     <DialogTitle>Prompt</DialogTitle>
                     <DialogContent>
-                        {(message.prompt === undefined || message.prompt === '')
-                            ? 'Empty'
+                        {(message.prompt === undefined || message.prompt === '' || message.prompt === null)
+                            ? 'No prompt available. The response is either a plan proposal or an initial greeting message.'
```
teresaqhoang, line 36:
```suggestion
                            ? 'No prompt available. The response is either a plan proposal or auto-generated response.'
```

To encompass responses like "No, cancel" on plan rejection

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
teresaqhoang: I see, if all messages come back anyway (even if you're the user that sent them), can't you skip updating the conversation on sendMessage altogether and rely on the conversation state to be updated once the message has been broadcasted to all chat participants?

samples/apps/copilot-chat-app/webapi/appsettings.json
TaoChenOSU: It's quite important to allow the bot to know the audience. With the new user schema you are working on, we should be able to remove this completely in the future.

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationJoinDialog.tsx
TaoChenOSU: I believe this is only traversing the form, which has limited controls.

samples/apps/copilot-chat-app/webapp/src/components/chat/prompt-details/PromptDetails.tsx
TaoChenOSU: I would change it to "hard-coded response". "auto-generated response" sounds like what we are already doing with the AI, lol

dehoward (changes requested):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,114 @@
...
+        this._chatSessionRepository = chatSessionRepository;
+    }
+
+    /// <summary>
+    /// Join a use to a chat session given a chat id and a user id.
+    /// </summary>
+    /// <param name="messageRelayHubContext">Message Hub that performs the real time relay service.</param>
+    /// <param name="chatParticipantParam">Contains the user id and chat id.</param>
+    [HttpPost]
+    [Route("chatParticipant/join")]
```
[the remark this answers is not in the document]
TaoChenOSU: We have a separate task to standardize the route pattern. We can come up with the appropriate route in that other task all together.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
```diff
@@ -0,0 +1,114 @@
...
+    public async Task<IActionResult> GetAllParticipantsAsync(Guid chatId)
+    {
+        // Make sure the chat session exists.
+        try
+        {
+            await this._chatSessionRepository.FindByIdAsync(chatId.ToString());
+        }
+        catch (Exception ex) when (ex is KeyNotFoundException || ex is ArgumentOutOfRangeException)
+        {
+            return this.BadRequest("Chat session does not exist.");
```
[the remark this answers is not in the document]
TaoChenOSU: We have a separate task to standardize the route pattern. We can come up with the appropriate route in that other task all together.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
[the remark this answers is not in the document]
TaoChenOSU: https://github.com/orgs/microsoft/projects/852?pane=issue&itemId=29522574

samples/apps/copilot-chat-app/webapi/CopilotChat/Storage/ChatParticipantRepository.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <summary>
+    /// Checks if a user is in a chat session.
+    /// </summary>
+    /// <param name="userId">The user id.</param>
+    /// <param name="chatId">The chat id.</param>
+    /// <returns>True if the user is in the chat session, false otherwise.</returns>
+    public Task<bool> IsUserInChatAsync(string userId, string chatId)
+    {
+        return base.StorageContext.QueryEntitiesAsync(e => e.UserId == userId && e.ChatId == chatId).
+            ContinueWith(t => t.Result.Any(), TaskScheduler.Default);
```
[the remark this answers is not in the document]
TaoChenOSU: Yes, we can use await instead.

samples/apps/copilot-chat-app/webapp/src/components/chat/prompt-details/PromptDetails.tsx
```diff
@@ -32,8 +32,8 @@ export const PromptDetails: React.FC<IPromptDetailsProps> = ({ message }) => {
                 <DialogBody>
                     <DialogTitle>Prompt</DialogTitle>
                     <DialogContent>
-                        {(message.prompt === undefined || message.prompt === '')
-                            ? 'Empty'
+                        {(message.prompt === undefined || message.prompt === '' || message.prompt === null)
```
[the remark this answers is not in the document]
TaoChenOSU: You mean `!message.prompt`?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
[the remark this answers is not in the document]
dehoward: awesome, thanks for linking

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
```diff
@@ -51,17 +72,42 @@ export const conversationsSlice: Slice<ConversationsState> = createSlice({
...
+            }
+        },
+        updateUserIsTypingFromServer: (state: ConversationsState, action: PayloadAction<{ userId: string; chatId: string; isTyping: boolean }>) => {
+            const { userId, chatId, isTyping } = action.payload;
+            const conversation = state.conversations[chatId];
+            const user = conversation.users.find(u => u.id === userId);
+            if (user) {
+                user.isTyping = isTyping;
+            }
+        },
```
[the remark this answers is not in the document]
TaoChenOSU: No, the first one is from the webapp and will be captured by the middleware. The second one will get invoked by the webapi to notify other clients.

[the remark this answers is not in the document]
TaoChenOSU: Move the duplicated code to a function.

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatParticipantController.cs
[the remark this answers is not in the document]
dehoward: in this case the ID is valid (since it passes the GUID validation) but it doesn't map to a session. seems like the same case as `GetChatMessagesAsync` and `GetSourcesAsync` in `ChatHistoryController`

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationCreateDialog.tsx
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import { Button, DialogActions, DialogBody, DialogContent, DialogSurface, DialogTitle, Label, makeStyles, tokens } from "@fluentui/react-components";
+import React, { useCallback, useEffect } from "react";
+
+const useStyles = makeStyles({
+    content: {
+        display: "flex",
```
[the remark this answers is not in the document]
TaoChenOSU: Leaving this open for further discussion. But this should not block merging.

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationCreateDialog.tsx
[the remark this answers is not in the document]
TaoChenOSU: Leaving this open for further discussion. But this should not block merging.

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationCreateDialog.tsx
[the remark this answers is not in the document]
dehoward: I think we can resolve this. this can be addressed as part of the ESLint work that I'm doing

samples/apps/copilot-chat-app/webapp/src/components/chat/prompt-details/PromptDetails.tsx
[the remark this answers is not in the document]
dehoward: yup :smile:

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
```diff
@@ -58,17 +77,34 @@ export const conversationsSlice: Slice<ConversationsState> = createSlice({
...
             frontLoadChat(state, id);
         },
+        updateUserIsTyping: (state: ConversationsState, action: PayloadAction<{ userId: string; chatId: string; isTyping: boolean }>) => {
+            const { userId, chatId, isTyping } = action.payload;
+            updateUserTypingState(state, userId, chatId, isTyping);
+        },
+        updateUserIsTypingFromServer: (state: ConversationsState, action: PayloadAction<{ userId: string; chatId: string; isTyping: boolean }>) => {
+            const { userId, chatId, isTyping } = action.payload;
+            updateUserTypingState(state, userId, chatId, isTyping);
+        },
```
dehoward, line 87: do we need both of these functions? I understand that they are used in separate places, but they have the same signature and do the same thing...

amsacha: Yes! This is so we can have different function handles for signalR callbacks; e.g updating state when its generated only from the backend (so broadcast to everyone) or when state is updated by a user (so broadcast to everyone but that user).

TaoChenOSU: They are called by different actors. 
`updateUserIsTyping` is called by the webapp itself when the user starts typing. And this function call will be captured by the redux middleware to broadcast the state.
`updateUserIsTypingFromServer` is called when the app receives the state broadcasted by the server.

If you use only one function, we will create an infinite loop because `updateUsersTyping` will trigger a broadcast event.

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
TaoChenOSU: Yeah, so the getResponse function will not return a message rn. The chat will get populated when the server broadcasts the response back, including the client who sent the message.

dehoward (approved):

amsacha (dismissed):

samples/apps/copilot-chat-app/webapp/src/components/chat/invitation-dialog/InvitationCreateDialog.tsx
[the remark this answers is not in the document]
hathind-ms: We can resolve this as a separate item

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
teresaqhoang: Understood now, but can you add a note in the README to explain this better? Was also super confusing to me until this last message

teresaqhoang (changes requested):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatController.cs
```diff
@@ -89,6 +95,13 @@ public ChatController(ILogger<ChatController> logger)
             return this.NotFound($"Failed to find {ChatSkillName}/{ChatFunctionName} on server");
         }
 
+        // Broadcast bot typing state to all users
+        if (ask.Variables.Where(v => v.Key == "chatId").Any())
+        {
+            var chatId = ask.Variables.Where(v => v.Key == "chatId").First().Value;
```
teresaqhoang, line 101: Can there be multiple chatIds in the context variables? How do you know the first one is right?

TaoChenOSU: We should catch these kinds of errors if there are duplicating variables in the Ask.

TaoChenOSU: Can be done in another PR

teresaqhoang (dismissed):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/ChatController.cs
teresaqhoang: documentation updates can also be done in another PR, just make sure you're tracking an issue for these things

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/conversationsSlice.ts
TaoChenOSU: Done

teresaqhoang (approved):

amsacha (approved):

adrianwyatt (approved):

PR #1346, opened 2023-06-05
Adding local ToListAsync method in Copilot Chat

teresaqhoang:
### Motivation and Context

The `System.Linq.Async` package reference was removed from the core libraries so that everyone using SK wasn't forced to use it as well. This adds a helper `ToListAsync` method to enable async conversion of `IAsyncEnumerable<T>` to `List<T>` object.

### Description

Was seeing the following error when using local project references, so this is an effort to future proof any SK nuget updates

`'IAsyncEnumerable<MemoryQueryResult>' does not contain a definition for 'ToListAsync' and no accessible extension method 'ToListAsync' accepting a first argument of type 'IAsyncEnumerable<MemoryQueryResult>' could be found (are you missing a using directive or an assembly reference?)  `

### Contribution Checklist

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/CopilotChatWebApi.csproj
+++ b/samples/apps/copilot-chat-app/webapi/CopilotChatWebApi.csproj
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.Identity.Web" Version="2.11.0" />
     <PackageReference Include="PdfPig" Version="0.1.8-alpha-20230423-3898f" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.5.0" />
+    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
   </ItemGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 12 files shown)


TaoChenOSU (approved):

samples/apps/copilot-chat-app/webapi/CopilotChatWebApi.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.Identity.Web" Version="2.11.0" />
     <PackageReference Include="PdfPig" Version="0.1.8-alpha-20230423-3898f" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.5.0" />
+    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
```
adrianwyatt, line 26: since it's non-obvious as to why this is here, could you add a comment in the project file explaining?

adrianwyatt (changes requested):

stephentoub:
If all you need is ToListAsync, it's also straightforward to write one's self:
```C#
internal static List<T> ToListAsync<T>(this IAsyncEnumerable<T> source)
{
    var result = new List<T>();
    await foreach (var item in source.ConfigureAwait(false)) result.Add(item);
    return result;
}
```

The `System.Linq.Async` package reference was removed from the core libraries so that everyone using SK wasn't forced to use it as well.

teresaqhoang:
> If all you need is ToListAsync, it's also straightforward to write one's self:
> 
> ```cs
> internal static async Task<List<T>> ToListAsync<T>(this IAsyncEnumerable<T> source)
> {
>     var result = new List<T>();
>     await foreach (var item in source.ConfigureAwait(false)) result.Add(item);
>     return result;
> }
> ```
> 
> The `System.Linq.Async` package reference was removed from the core libraries so that everyone using SK wasn't forced to use it as well.

Ahhh I didn't know the package was removed intentionally. Removed here also and added a local helper method.

adrianwyatt (approved):

PR #1349, opened 2023-06-05
Weaviate: Fix to be compatible with python 3.8

hsm207:
### Description

use run_in_executor instead of to_thread to run the weaviate client ops in async because  to_thread is not available in 3.8

### Contribution Checklist


python/tests/integration/connectors/memory/test_weaviate_memory_store.py
```diff
@@ -67,10 +69,25 @@ def documents():
 
 @pytest.fixture
 def memory_store():
+    max_attempts = 5  # the number of retry attempts
+    delay = 5 * 60  # delay in seconds between each attempt
```
awharrison-28, line 73: Is the 60s number arbitrary? I'm a little concerned that this could result in a wait of 25 minutes before ultimately discovering a failing test. Could 30 seconds adequate?

awharrison-28 (changes requested):
Small nit: can you create a sub folder so that the path for weaviate looks like:

`semantic_kernel/connectors/memory/weaviate/weaviate_memory_store.py`

python/tests/integration/connectors/memory/test_weaviate_memory_store.py
hsm207: 30s worked on my machine. Let's see if is adequate for the CI too.

fyi, the reason for the flakiness is weaviate/weaviate-python-client#349

hsm207:
> Small nit: can you create a sub folder so that the path for weaviate looks like:
> 
> `semantic_kernel/connectors/memory/weaviate/weaviate_memory_store.py`

done!

dluc (approved):

PR #1354, opened 2023-06-06
Python: postgres memory store

### LINKED ISSUE #1270: Python: Add Postgres with pgvector as a Memory Store for Python

cschadewitz:
### Description
Implemention of the MemoryStore to use a Postgresql (with pgvector extension) to resolve #1270 

Taking a different approach from the .net implementation:
Collections as postgres schemas as opposed to collection being another column in the table, willing to change to match but would love to have a discussion comparing the options. 

### Contribution Checklist


cschadewitz:

cschadewitz:
Working on change to conform to new Memory Record field

python/tests/integration/connectors/memory/test_postgres.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+@pytest.mark.asyncio
+async def test_get_collections_async(get_postgres_config):
+    connection_string = get_postgres_config
+    memory = PostgresMemoryStore(connection_string, 1, 5)
+
+    await memory.create_collection_async("test_collection", 2)
+    # await memory.create_collection_async("test_collection-b")
+    # await memory.create_collection_async("test_collection-c")
```
dmytrostruk, line 110: Remove unused code:
```suggestion
```

python/tests/integration/connectors/memory/test_postgres.py
```diff
@@ -0,0 +1,250 @@
...
+    result = await memory.get_collections_async()
+    assert len(result) > 1
```
dmytrostruk, line 112: I would not rely on amount of collections in test database, but verify if `test_collection` is in `result`:
```suggestion
    assert "test_collection" in result
```

python/tests/integration/connectors/memory/test_postgres.py
```diff
@@ -0,0 +1,250 @@
...
+
+    await memory.create_collection_async("test_collection", 2)
+    await memory.upsert_async("test_collection", memory_record1)
+    result = await memory.get_async(
+        "test_collection", memory_record1._id, with_embedding=True
+    )
+    assert result is not None
+    assert result._id == memory_record1._id
+    assert result._text == memory_record1._text
+    assert result.embedding is not None
```
dmytrostruk, line 149: Should we also verify how embedding vector was stored by checking actual vector values?

python/tests/integration/connectors/memory/test_postgres.py
```diff
@@ -0,0 +1,250 @@
...
+    connection_string = get_postgres_config
+    memory = PostgresMemoryStore(connection_string, 1, 5)
+
+    await memory.create_collection_async("test_collection", 2)
+    await memory.upsert_async("test_collection", memory_record1)
+    await memory.remove_async("test_collection", memory_record1._id)
+    with pytest.raises(KeyError):
+        _ = await memory.get_async(
+            "test_collection", memory_record1._id, with_embedding=True
+        )
```
dmytrostruk, line 184: Same as for collections, I would add extra `get` operation just to double-check that deletion actually happened:
```suggestion
    await memory.create_collection_async("test_collection", 2)
    await memory.upsert_async("test_collection", memory_record1)
    
    result = await memory.get_async(
        "test_collection", memory_record1._id, with_embedding=True
    )
    assert result is not None
    
    await memory.remove_async("test_collection", memory_record1._id)
    with pytest.raises(KeyError):
        _ = await memory.get_async(
            "test_collection", memory_record1._id, with_embedding=True
        )
```

dmytrostruk (commented):
@cschadewitz Thank you for contribution! I left initial comments. As soon as they are resolved, we will test it locally and if everything works as expected - merge it to main! 😃

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,335 @@
...
+
+# On Windows, Psycopg is not compatible with the default ProactorEventLoop.
+# Please use a different loop, for instance the SelectorEventLoop.
+# https://www.psycopg.org/psycopg3/docs/advanced/async.html
+if os.name == "nt":
+    import asyncio
+
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+
+MEMORY_TABLE_NAME = "sk_memory_table"
```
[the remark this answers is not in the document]
cschadewitz: Makes sense, I had originally set this up to allow for delineation of the collections to be at the schema level. Since the direction is to go with table level delineation, I will make the necessary changes accordingly.

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,335 @@
...
+    async def get_collections_async(self) -> List[str]:
+        with self._connection_pool.connection() as conn:
+            with conn.cursor() as cur:
+                return await self.__get_collections_async(cur)
+
+    async def delete_collection_async(self, collection_name: str) -> None:
+        with self._connection_pool.connection() as conn:
+            with conn.cursor() as cur:
+                cur.execute(
+                    f"DROP SCHEMA IF EXISTS {collection_name} CASCADE",
```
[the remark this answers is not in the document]
cschadewitz: Will implement full parameterization

python/semantic_kernel/connectors/memory/pinecone/pinecone_memory_store.py
cschadewitz, line 1: fixed param ordering in doc string

cschadewitz:
@dmytrostruk All PR comments so far should be resolved, I have also fixed one issue in the doc string from the pinecone mem store and have added the dependencies to the .toml file. 

Let me know if anything else comes up.

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,524 @@
...
+        if dimension_num > MAX_DIMENSIONALITY:
+            raise ValueError(
+                f"Dimensionality of {dimension_num} exceeds "
+                + f"the maximum allowed value of {MAX_DIMENSIONALITY}."
+            )
+        if dimension_num <= 0:
+            raise ValueError("Dimensionality must be a positive integer. ")
+
+        with self._connection_pool.connection() as conn:
+            conn.execute("CREATE EXTENSION IF NOT EXISTS vector")
```
JadynWong, line 94: The C# implementation does not execute this statement anymore, because it only needs to be executed once after the database is created, and most importantly it seems that not all hosting can execute this statement, for example [Azure Cosmos DB for PostgreSQL](https://learn.microsoft.com/en-us/azure/cosmos-db/postgresql/howto-use-pgvector).

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,524 @@
...
+            with conn.cursor() as cur:
+                cur.execute(
+                    SQL(
+                        """
+                        CREATE TABLE IF NOT EXISTS {scm}.{tbl} (
+                            key TEXT PRIMARY KEY,
+                            embedding vector({dim}),
+                            metadata JSONB,
+                            timestamp TIMESTAMP
```
JadynWong, line 103: C# uses BIGINT type to unix to store timestamp, I am considering whether to use TIMESTAMP to store it. The postgres timestamp requires that it be stored in UTC time, and when you query it again, it returns the utc time, with no way of knowing the actual time zone when it was stored. How do you think about it?

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
cschadewitz: Interesting, I didnt know that there were some hosting options with different statements for this. I can certainly remove it. Thanks for pointing this out! I have seen some of the work you have done/are doing against the C# side for the Postgres memory store. I'll have to take a look at what you currently have in the pipeline. Thanks @JadynWong

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
cschadewitz: This was something that I saw when I was looking at the most recent version of the C# implementation, so I am really glad that you brought this up! You are correct based on the docs, Postgres has the concept of both timestamp without time zone and timestamp with time zone, and this does not affect the storage as both are ultimately stored as 8-byte integers in UTC. 

[Time Stamps](https://www.postgresql.org/docs/current/datatype-datetime.html#:~:text=8.5.1.3.%C2%A0Time%20Stamps)

This would mean that we would need to be contextually aware of the timezone of the client (which sounds like we are in the C# implementation) and sprinkle in the `timezone(zone, timestamp)` or 'AT TIME ZONE' conversion functions to push the conversion work onto the db.

[Time Stamp Functions](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-ZONECONVERT)

Ultimately, I think that we should go with `timestamp with time zone`and use the built-in functions where appropriate. Do you agree?

cschadewitz: Initial stab at handling this is in the conde now.

cschadewitz: I have changed the `MemoryRecord `implementation to use `datetime` instead of `str` for the `timestamp`. With this in place I have changed the `PostresqlMemoryStore` and tests to account for this. 

Additionally, I have made the necessary changes to the `ChromaMemoryStore` and `PineconeMemoryStore` and tests to work with this new structure but am holding off on pushing this since it is outside of the scope of this PR. I am happy to include it if requested. 

The only memory store that I havent created an update for is the `WeaviateMemoryStore`, since I dont have it installed at this time.

dmytrostruk: As a best practice, I think datetimes in database should be stored in UTC and converted on client side to appropriate timezone. In other words, it's a client responsibility to handle datetime conversion.

Semantic Kernel, as SDK, shouldn't be aware of timezones at all. That means, if customer uses UTC to store datetimes in database, then SK puts a record with timestamp in UTC. If customer decided to use local timezone to store information - then SK puts a record with timestamp in local timezone. In other words, we should pass timezone in the same format as we received from SDK user and we should not perform any conversions.

As an example, in case there will be some default logic, when we put some default timestamp (for example `DateTime.UtcNow`) - the timezone should always be in UTC. Although even in this case I would probably set default timestamp value as `null`, but it depends on scenario.

@cschadewitz I see some datetime conversion in `postgres_memory_store.py`, so I think we need to remove it, since it's SDK consumers responsibility. Please let me know if that makes sense.

We will probably need to make some changes in C# kernel to make timestamp field as timezone-agnostic, but it's definitely a scope of separate PR.

python/tests/integration/connectors/memory/test_postgres.py
```diff
@@ -0,0 +1,266 @@
...
+
+    await memory.create_collection_async("test_collection")
+    result = await memory.does_collection_exist_async("test_collection")
+    assert result is not None
+
+
+@pytest.mark.asyncio
+async def test_get_collections_async(get_postgres_config):
+    connection_string = get_postgres_config
+    memory = PostgresMemoryStore(connection_string, 2, 1, 5)
```
dmytrostruk, line 108: nit: We can rename `get_postgres_config` to `connection_string` and pass it directly to `PostgresMemoryStore`. Same applies to all other tests.

dmytrostruk (commented):
@cschadewitz Thank you for all the hard work! I left some comments, but some of them can be definitely resolved in scope of follow-up PR. I think the most important one which should be resolved before merging to main is regarding timezones, please find the details under my comment. Thanks again!

Thanks again!

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
cschadewitz: That is fine with me if we are passing the responsibility off to the consumer of the SDK.

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,564 @@
...
+        """
+        if dimension_num is None:
+            dimension_num = self._default_dimensionality
+        if dimension_num > MAX_DIMENSIONALITY:
+            raise ValueError(
+                f"Dimensionality of {dimension_num} exceeds "
+                + f"the maximum allowed value of {MAX_DIMENSIONALITY}."
+            )
+        if dimension_num <= 0:
+            raise ValueError("Dimensionality must be a positive integer. ")
```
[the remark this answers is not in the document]
cschadewitz: Yeah this one is kind of tough for me since most of the memory store backends require a dimensionality at time of collection creation. I understand that the `MemoryStoreBase` doesn't have it as a parameter and as such the Kernel will not pass anything for it but I wonder if that should be the case. 

I wonder if there is a use case in which a consumer would interface with 2 embedding services (that return embeddings with different dimensionality). As the memory store stands now, they would need two instances of a memory store to support both dimensionalities, which I dont know if the Kernel even supports.

Food for thought.

Regardless, I can move the check into a single method but will maintain the parameter on the `create_collection_async` method for the time being

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,564 @@
...
+            if time.localtime(t).tm_isdst and time.daylight:
+                float_offset = -time.altzone / 60 / 60
+            else:
+                float_offset = -time.timezone / 60 / 60
+
+            self._timezone_offset = (
+                str(float_offset).replace(".", ":")[:5]
+                if float_offset < 0
+                else str(float_offset).replace(".", ":")[:4]
+            )
```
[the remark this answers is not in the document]
cschadewitz: Since Postgres does not store the timezone with a timestamp, I am going to remove all timezone related work and assume that the consumer is always passing in UTC time (or really whatever timezone they want) and are handling all necessary conversions

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,564 @@
...
+        with self._connection_pool.connection() as conn:
+            with conn.cursor() as cur:
+                if not await self.__does_collection_exist_async(cur, collection_name):
+                    raise Exception(f"Collection '{collection_name}' does not exist")
+                cur.execute(
+                    SQL(
+                        """
+                    SELECT key, embedding, metadata, timestamp AT TIME ZONE INTERVAL {tz} AS timestamp
+                    FROM {scm}.{tbl}
+                    WHERE key = %s"""
```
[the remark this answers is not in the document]
cschadewitz: fixed

python/tests/integration/connectors/memory/test_postgres.py
cschadewitz: fixed

python/semantic_kernel/connectors/memory/postgres/postgres_memory_store.py
```diff
@@ -0,0 +1,564 @@
...
+            collection_name {str} -- The name of the collection to upsert the record into.\n
+            record {MemoryRecord} -- The record to upsert.
+
+        Returns:
+            str -- The unique database key of the record. In Pinecone, this is the record ID.
+        """
+        with self._connection_pool.connection() as conn:
+            with conn.cursor() as cur:
+                if not await self.__does_collection_exist_async(cur, collection_name):
+                    raise Exception(f"Collection '{collection_name}' does not exist")
```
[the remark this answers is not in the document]
cschadewitz: Sounds fair, I will look at doing this in a subsequent pr

awharrison-28:
Hi @cschadewitz - I'm setting up infra to be able to run the integration tests in CI. As soon as that's set up, we'll get this in asap.

awharrison-28 (approved):

robertmujica:
Great work @cschadewitz , wondering if there is any pre-release i can use to test it ?

cschadewitz:
> Great work @cschadewitz , wondering if there is any pre-release i can use to test it ?

Funny you should ask, I made a request for python to get nightly/pre-release builds during yesterdays community hour. Here is another community member that want pre-release pip packages @alexchaomander . 
Thanks!

PR #1357, opened 2023-06-07
Added a text completion connector to oobabooga text-generation-webui project

jsboige:
### Motivation and Context
[Oobabooga's text-generation-webui](https://github.com/oobabooga/text-generation-webui) is the most popular open-source platform to host LLMs. It has a web API that supports blocking and streaming requests.
Having a connector to oobabooga brings semantic-kernel to the global "local Llamas" community.

### Description
This PR adds to the solution a project similar to HuggingFace connectors project, and an additional integration test also similar to HuggingFace connector's
The code for the connector was based on the existing HuggingFace's, with a couple improvements (e.g. using web sockets for streaming API)

### Contribution Checklist

As for HuggingFace, the integration tests are meant to be tested manually together with oobabooga installed with a model selected and a running API. Both the blocking and streaming APIs were tested succesfully, though parameters mappings from sk's CompleteRequestSettings  to oobabooga's TextCompletionRequest was done very roughly and would need adjusting since some units scales don't match, and oobabooga's API has a lot more parameters to feed, but I suppose it works well enough for an first version.  


jsboige:
@microsoft-github-policy-service agree [company="{My Intelligence Agency}"]

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OoobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,232 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
dmytrostruk, line 1: Filename doesn't match class name. I think there is extra "o" at the beginning of filename 😃

dotnet/src/IntegrationTests/Connectors/Oobabooga/OobaboogaTextCompletionTests.cs
```diff
@@ -0,0 +1,72 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
dmytrostruk, line 1: Could you please update this [README.md](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/IntegrationTests/README.md) and add instructions how to run tests locally? We have a plan to do the same for HuggingFace later. 

Also, since these are integration tests which require local setup and can be verified only manually, it would be great if we will add unit tests for this functionality. You can use [HuggingFace unit tests](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.UnitTests/HuggingFace/TextCompletion/HuggingFaceTextCompletionTests.cs) as an example.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OoobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,232 @@
...
+    private readonly HttpClientHandler? _httpClientHandler;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="OobaboogaTextCompletion"/> class.
+    /// </summary>
+    /// <param name="endpoint">Endpoint for service API call.</param>
+    /// <param name="blockingPort">The port for blocking requests.</param>
+    /// <param name="streamingPort">The port for streaming requests.</param>
+    /// <param name="httpClientHandler">Instance of <see cref="HttpClientHandler"/> to setup specific scenarios.</param>
+    public OobaboogaTextCompletion(Uri endpoint, int blockingPort, int streamingPort, HttpClientHandler httpClientHandler)
```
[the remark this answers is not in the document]
jsboige: See #1357

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OoobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,232 @@
...
+    {
+        Verify.NotNull(endpoint);
+
+        this._endpoint = endpoint;
+        this._blockingPort = blockingPort;
+        this._streamingPort = streamingPort;
+
+        this._httpClient = new(httpClientHandler);
+        this._httpClient.DefaultRequestHeaders.Add("User-Agent", HttpUserAgent);
+        this._webSocket = new ClientWebSocket();
```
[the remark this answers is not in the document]
jsboige: See #1357. 

Note that websockets work differently, so User-Agent will be permanently overriden here unlike the blocking solution, which only affects the local request. I guess this isn't much of a an issue, since websocketclient shouldn't really be shared with other usages, and the ctor parameter simply allows to add more options to it, but it still feels a bit inconfortable to affect the input client.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OoobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,232 @@
...
+                {
+                    throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Unexpected response from model")
+                    {
+                        Data = { { "ModelResponse", body } },
+                    };
+                }
+
+                switch (responseObject.Event)
+                {
+                    case "text_stream":
```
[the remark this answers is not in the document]
jsboige: See #1357

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OoobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,232 @@
...
+                        yield return new TextCompletionStreamingResult(responseObject.Text);
+                        break;
+                    case "stream_end":
```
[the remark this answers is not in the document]
jsboige: See [#33d3d70dffd2515ae48e090aef32b61ab598ab26](https://github.com/microsoft/semantic-kernel/pull/1357/commits/33d3d70dffd2515ae48e090aef32b61ab598ab26)

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OoobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        // Prepare the request using the provided parameters.
+        return new TextCompletionRequest()
+        {
+            Prompt = text,
+            MaxNewTokens = requestSettings.MaxTokens,
+            Temperature = requestSettings.Temperature,
+            TopP = requestSettings.TopP,
+            RepetitionPenalty = 0.5f + requestSettings.PresencePenalty,
```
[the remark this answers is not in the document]
jsboige: See [#1f54bcc13285571513d0f0cadd9781f80cb3d290](https://github.com/microsoft/semantic-kernel/pull/1357/commits/1f54bcc13285571513d0f0cadd9781f80cb3d290)
As I wrote in my initial PR comment, parameter matching would probably need more attention since the scales don't match.
This one matches a -2:+2 scale to an exponential 0:+inf one. The new formula should be OK except for -2 which will map to 0 and generate an exception on the Oobabooga side, which expects a strictly positive number. 
I thought about adding double.Epsilon by default to avoid that, but then neutral 1 isn't exactly 1 anymore even if very close. I'm not sure which solution is best.
More generally, there seem to be different parameters and different interpretation of those parameters accross the LLM landscape while this isn't propertly standardised. 
Obviously, oobabooga supporting everything tries to accommodate that in a ad-hoc manner, and there isn't probably an ideal answer to those kinds of questions yet. 
It's probably not too important for now though, since semantic-kernel only supports a basic handful of those parameters.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OoobaboogaTextCompletion.cs
jsboige: See [#cf742c4f2d9df1fa48f1410e34670d625e66995f](https://github.com/microsoft/semantic-kernel/pull/1357/commits/cf742c4f2d9df1fa48f1410e34670d625e66995f)

dotnet/src/IntegrationTests/Connectors/Oobabooga/OobaboogaTextCompletionTests.cs
jsboige: For instructions, see [#6d1a4c5840ae2fb7c24a58fdbcc6311e2610bc00](https://github.com/microsoft/semantic-kernel/pull/1357/commits/6d1a4c5840ae2fb7c24a58fdbcc6311e2610bc00). 
I kept them minimal, since I didn't know how much details your wanted to add to that Readme, but overall oobabooga install is pretty straight forward (installing all the accelerators and optimizers is another matter, and may require to play with custom wheels, cuda compilation switching to wsl2, depending on your GPUs hardware, environment and target models, but I'm pretty sure this is temporary). Anyway if you have issues, let me know. If you have a decent GPU, you can go a long way with the default install, and the gradio web ui has tons of parameters that you can hot swap (in the flavor of stable-diffusion web ui as officially stated).

jsboige: Unit tests were hard.
First and foremost, why isn't System.Linq.Async Nuget package used throughout the solution? It seems to me the current situation with half baked replacements is suboptimal, and requires pragmas to get through.
Then my main issue was that you can't get those mocks as easily with ClientWebsocket as you do with Httpclient (handler is internal). I ended up introducing an actual test websocket server that registers on a local port. It works well locally, but hopefully this won't break depending on the test running environment.
Eventually, I believe this is a useful addition since that test server will allow more complex tests than simple mocks would have allowed.

I also updated the integration tests, but you might want to change or complete them according to your needs. 
Hopefully all of this is enough for you to get started and to take over though...

jsboige: Also I guess there is a bit of overlap/redundancy with HuggingFace connector now, and ultimately, you'd want to abstract and mutualise some of the code I introduced.

dmytrostruk: Thanks @jsboige , this is awesome! In case `ClientWebsocket` is not mockable, let's remove unit tests for it and leave that responsibility for integration tests. Let's use `HttpClient` only for unit tests.

jsboige: > Thanks @jsboige , this is awesome! In case `ClientWebsocket` is not mockable, let's remove unit tests for it and leave that responsibility for integration tests. Let's use `HttpClient` only for unit tests.

Well, I ended up mocking the websocket server, which works great on my environment, so unless there is anything preventing it to run on the test server, I'd say we're good to go and that is still a good addition, which will allows for subtle fine-grained unit test since we have complete control on the way the server handles the incoming byte packets.

For instance, [that test](https://github.com/microsoft/semantic-kernel/blob/7025f2442b2cbe0ae9ab8bb1dc70d9525a19ce9c/dotnet/src/Connectors/Connectors.UnitTests/Oobabooga/TextCompletion/OobaboogaTextCompletionTests.cs#L125) aggregates the incoming chunks on the server side and checks that the request was properly submitted, whereas [this one](https://github.com/microsoft/semantic-kernel/blob/7025f2442b2cbe0ae9ab8bb1dc70d9525a19ce9c/dotnet/src/Connectors/Connectors.UnitTests/Oobabooga/TextCompletion/OobaboogaTextCompletionTests.cs#L152) has the server send back the expected response chunks and checks that the client properly handles it.

dmytrostruk (approved):
Tested this connector locally using integration tests - everything works as expected.

dmytrostruk:
Please ignore Spell Check for now - we are going to fix that separately. PR still can be merged.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,215 @@
...
+using Microsoft.SemanticKernel.AI.TextCompletion;
+using Microsoft.SemanticKernel.Diagnostics;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.Oobabooga.TextCompletion;
+
+/// <summary>
+/// Oobabooga text completion service API.
+/// Adapted from https://github.com/oobabooga/text-generation-webui/tree/main/api-examples
+/// </summary>
+public sealed class OobaboogaTextCompletion : ITextCompletion, IDisposable
```
SergeyMenshykh, line 23: We are transitioning to non-disposable connectors by ensuring that connector classes do not internally create disposable resources. Instead, if a disposable resource is needed, it should be supplied through the class constructor, indicating that the responsibility for disposal lies outside the connector. This approach enables reusing the same connector across multiple instances of the kernel and helps avoid potential issues caused by frequent creation and disposal of resources. One example is the port exhaustion due to a large number of short-lived (created and disposed per request) HTTP clients. For more information, please refer to the following link: [HttpClient Guidelines - Pooled Connections](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#pooled-connections).

To address this specific case, please modify the class so that it no longer implements IDisposable. Instead, pass an instance of the WebSocket class as a parameter to the class.

jsboige: OK, I'll update the connector to use the fallback non disposable http client if no websocket is provided

jsboige: Guys, I will need your help to finish this up. I have no idea what's going on. 
I implemented your requested change, I ended up using a transient websocket client when no persistent one is supplied, which wasn't much of an issue (whether this is the appropriate solution can be discussed though).

But then suddently, the streaming integration test stopped working for me, without much reason for that. Oobabooga's Python websocket server would complain that the connection is closed. I tried various stuff and ended up installing Oobabooga on the same machine (I was using a remote server without any issue so far), but it didn't make it.
Upon investigating, I realized that it was the client now returning after the first (empty) packet without even closing the connection properly, whereas it used to receive 3 packets, 1st one empty, 2nd with actual content, and 3rd one to signal the end of the message.

I spent hours trying to fix that. The GetStreamingCompletionsAsync method gets out of the while(true) loop after the first packet is received, with the socket still open and no message received to end the connection.
It's as if suddenly Async won't work the same anymore.
I tried everything without any success (note that I also tried reverting any change, but it just won't come back), and for what it's worth, ChatGPT 4.0 is just as puzzled as I am.

Well maybe you won't even have the issue and something weird happened with my environment, or maybe it did work miraculously beforehand , I'm not sure what is preferable, but I'll be grateful if you can have a look.
Cheers

dmytrostruk: @jsboige I think the problem is on our side, we merged some bad changes to streaming logic in this PR: https://github.com/microsoft/semantic-kernel/pull/1117

We have new PR open which will resolve that issue: https://github.com/microsoft/semantic-kernel/pull/1391.

As soon as it's merged, please pull this change from `main` and test again. I will ping you here as soon as fix is merged. Sorry for inconvenience.

dmytrostruk:
@jsboige I think the fix for streaming was merged to the main branch, can you pull latest changes from `main` and check if your functionality works as expected? Thanks!

jsboige:
> @jsboige I think the fix for streaming was merged to the main branch, can you pull latest changes from `main` and check if your functionality works as expected? Thanks!

@dmytrostruk I saw the merge and started working on the issue again. It seems it did not fix the issue, so I refactored and extended the unit test capabilities with a new multi-message mock web socket server, and corresponding multi-string wrapped oobabooga-specific inherited version so that the issue can be pin-pointed as part of the unit tests rather than the integration tests. I just commited the whole thing, so now the issue was brought back into the unit tests where we have control on the server, which should be more comfortable to debug.

jsboige:
> @jsboige I think the fix for streaming was merged to the main branch, can you pull latest changes from `main` and check if your functionality works as expected? Thanks!

@dmytrostruk OK I got it. The issue was with the CompleteStreamAsync extension method's added "yield break" which would result in returning after 1st message. Now both Unit tests and Integration tests work for me.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
jsboige: See the following comments. The new bug was identified and fixed in [#1bf79f0545bfd71f6291541c022e73d7ae3a78b2](https://github.com/microsoft/semantic-kernel/pull/1357/commits/1bf79f0545bfd71f6291541c022e73d7ae3a78b2)

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,257 @@
...
+
+                if (clientWebSocket.State != WebSocketState.Open)
+                {
+                    break;
+                }
+            }
+        }
+        finally
+        {
+            transientClientWebSocket?.Dispose();
```
SergeyMenshykh, line 155: I'm not sure if creating an instance of ClientWebSocket class and disposing it per method call is good idea because it may lead to port exhaustion issue under high load similar to what may happen if HttpClient instances are created and disposed per request. More details - [link](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#pooled-connections).
@stephentoub what do you think?

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,257 @@
...
+            ClientWebSocket? clientWebSocket;
+            if (this._webSocket is null)
+            {
+                transientClientWebSocket = new();
+                this.SetWebSocketOptions(transientClientWebSocket);
+                clientWebSocket = transientClientWebSocket;
+            }
+            else
+            {
+                clientWebSocket = this._webSocket;
```
SergeyMenshykh, line 94: What's going to happen if the GetStreamingCompletionsAsync method is called more than once in parallel and the first call already sent a prompt to the server and is waiting/reading completion/response from it while the second call sends a new prompt to the same server through the same instance of ClientWebSocket - _webSocket ?

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/TextCompletionResult.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.AI.TextCompletion;
+using Microsoft.SemanticKernel.Orchestration;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.Oobabooga.TextCompletion;
+
+internal sealed class TextCompletionResult : ITextCompletionResult
```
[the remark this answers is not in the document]
jsboige: I did add summary comments to those internal result classes, but that's about it, because there is actually no comment in any of the existing classes implementing the same interfaces. I don't want to set the standard higher than it is for paid contributors. Hopefully you're OK with that.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
jsboige: I'll provide a global answer in your next message, since the 2 issues are pretty much related.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
jsboige: OK so let's summarize the situation.
Websocket clients have those distinctive features:

- They cannot be replaced using a HttpClient since they use a distinct protocol. An HttpListener can be used on the server side though as in the mock server, but that's about it.
- Unlike HttpClient, they don't have a non disposable version, and their handler isn't available, at least until incoming versions.
- As you noted, they're not meant for concurrent access.

The initial version of the connector in this PR had a single websocket client properly disposed.
Then the option was introduced to passe the websocket client as part of the constructor.
Then it was requested that the connector be non disposable, and since there is no non-disposable version of websocket client I introduced those transient version, while keeping the option to provide one externally in the ctor.

Now both your remark make sense, but they put us in a tricky situation: We should avoid single use websocket clients to prevent port exhaustion in the extreme case, but we should also avoir reusing websocket clients in concurrent scenarios to avoid unintended behavior, added to the constraint that we must support both clients supplied externally and instantiated internally, and that we cannot dispose the latter with the class.

That was quite a challenge but I tried to respect those 4 hard constraints, with some kind of convoluted pooling: Incoming client calls signal concurrency by pushing and poping a flag to a concurrent stack. When the stack gets empty means there is no concurrent call anymore. As long as the stack is not empty, used clients get recycled in another concurrent stack, and when there is no concurrency, last call flushes and disposes the existing websocket clients before leaving.
As for externally provided websocket clients, I replaced the original ctor paramter with a factory, leaving the responsibility to deal with pooling to the consumer, with the corresponding comment warning about hazardous concurrent use and port exhaustion.

I also fixed the mock websocket server and added new unit tests, to check websocket server concurrency, and to test concurrent streaming calls with multiple websocket response messages, both in the persistent and the transient cases.

Hopefully this will be enough to merge that PR, since as long as it stays there, I have to deal with the numerous changes to the exposed surface, which has been exhausting. Also, my previous comments addressed a bug, which I spent hours chasing. Can you acknowledge that my fix is sound?

jsboige:
> @jsboige Could you please review the latest comments? I believe we are on the finish line to merge it to main 😃

@dmytrostruk I hope we're there too. I have been trying to accommodate various constraints since I opened that PR. Now I understand that this project is moving fast, but this is quite an inconfortable position. Hopefully you guys can make a decision quick enough so that I don't have to refactor everything by the time it gets accepted.

About the previous integration test async failing, did you see my previous comments and fix?

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/TextCompletionResult.cs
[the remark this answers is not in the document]
dmytrostruk: No need to document `internal` classes, methods and properties (at least for now). Just the `public` ones require documentation. In case all public classes, methods and properties are documented, feel free to resolve this comment, thanks.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
```diff
@@ -0,0 +1,296 @@
...
+            using var response = await this._httpClient.SendAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
+            response.EnsureSuccessStatusCode();
+
+            var body = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+            TextCompletionResponse? completionResponse = JsonSerializer.Deserialize<TextCompletionResponse>(body);
+
+            if (completionResponse is null)
+            {
+                throw new AIException(AIException.ErrorCodes.InvalidResponseContent, "Unexpected response from model")
```
RogerBarreto, line 236: When implementing the exception, please wrap it in a dedicated exception where you expose the details in a property.

Please see:
[dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs](https://github.com/microsoft/semantic-kernel/blob/1efe78aab4472460b71895efc85b65506a6f177b/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs#L62C23-L62C53) and [dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIInvalidResponseException{T}.cs#L9](https://github.com/microsoft/semantic-kernel/blob/1efe78aab4472460b71895efc85b65506a6f177b/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIInvalidResponseException%7BT%7D.cs#L9)

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -54,7 +54,6 @@ await foreach (var word in completionResult.GetCompletionStreamingAsync(cancella
             {
                 yield return word;
             }
-            yield break;
```
RogerBarreto, line 57: This extension method is intended for single results only.

See the comment on the method.
`This extension does not support multiple prompt results (Only the first will be returned)`

Without the `break` this will yield all the multiple responses as one big text making the result indistinguishable and confusing.

RogerBarreto (changes requested):

jsboige:
> Hi @jsboige !
> 
> > About the previous integration test async failing, did you see my previous comments and fix?
> 
> After internal communication with team, I believe this fix cannot be pushed, please see latest comment from @RogerBarreto .
> 
> > I won't have any more time until Tuesday to finish this up
> 
> No worries at all, please take your time! We will prioritize this PR as soon as it's ready. Also, please let us know if any help is needed.
> 
> Thanks a lot!

OK I'll account for those last 2 comments, not relying on the extension method and adding a dedicated exception.

In the mean time, can you have a look at my last attempt at consolidating the pooling mechanism? I believe the latest version is more robust and the last 2 stress unit tests added yield pretty good results:
The first one does not limit concurrency. It makes 1000 calls to the connector with a 1ms delay between each asynchronous task startup, each of which result in 4 websocket messages in response. There is a ramp up period where new websocket clients are created while the first ones are being processed by the http server. Then as the first calls finish, websocket clients get recycled and no more are created.
In the second test, a semaphore ensures that no more than 20 concurrent calls can be made to the connector at a time. Accordingly, there is no warmup and no need for a small delay. The connector is instantly hammered with 10000 simultaneous calls. the websocket client creation then maxes out at the semaphore size, subsequent calls all use recycled websocket clients.
The mock websocket server was also consolidated to handle such a high load (4*10000 websocket response messages in 0.5sec on my environment), and everything seems to work as intended.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
jsboige: @SergeyMenshykh Please check [this test](https://github.com/microsoft/semantic-kernel/pull/1357/files#diff-a76773b8dde48313e8b84fcfeac12a63334d1ceea3e5d7010882c2a108373973R243) and [that one](https://github.com/microsoft/semantic-kernel/pull/1357/files#diff-a76773b8dde48313e8b84fcfeac12a63334d1ceea3e5d7010882c2a108373973R257) where a lot of concurrent calls are made. Distinct websockets are used for each request, but they get recycled to prevent resource exhaustion as per your previous comment.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaTextCompletion.cs
jsboige: @RogerBarreto the new exception type is [there](https://github.com/microsoft/semantic-kernel/pull/1357/files#diff-99b9245cb239afb1fe54f86541c7c96cf34b0f3aa52d7c87fc08d41c42119af4R8).

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
jsboige: OK I believe I got it right this time. 
I restored the yield break while keeping the asynchronous streaming behavior, which was hard to get right.

- I used Channels to enabled on demand streaming iteration in the Result class in a simple implementation,
- I added yet another asynchronous Task to process websockets in order to be able to yield return the first incomplete StreamingResult after the first websocket message chunk is received, while preventing the "yield break" from this extension method to freeze subsquent websocket processing, which was leading to a deadlock where the iterator consumer would wait for the "stream_end" websocket message to terminate and the websocket processing would keep frozen waiting for the iterator to end to resume websocket processing. 
That dead lock took me sometime to understand and properly fix.

That is to say: that yield break puts a hard constraint on the way we can implement streaming completion. I wonder if a simple "break;" would have been gentler.
In any case, I hope my code can help other move forward with supporting websocket servers.

jsboige:
> Hi @jsboige !
> 
> > About the previous integration test async failing, did you see my previous comments and fix?
> 
> After internal communication with team, I believe this fix cannot be pushed, please see latest comment from @RogerBarreto .
> 
> > I won't have any more time until Tuesday to finish this up
> 
> No worries at all, please take your time! We will prioritize this PR as soon as it's ready. Also, please let us know if any help is needed.
> 
> Thanks a lot!

Hi @dmytrostruk 
I believe I'm done with consolidating that PR. Honestly that was quite the ride, and I didn't expect to spend so much time buried in websocket and async coding issues, which I knew practically nothing about. Hopefully others won't have to go through the same trouble.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
jsboige: @RogerBarreto 
I did some more work, with comments about them below. Maybe we can resolve that earlier conversation.

jsboige:
@dmytrostruk , @RogerBarreto, 
 Hi guys, as you may see, I'm strill struggling trying to figure out what's going on with the TextCompletionStreamingResult alternatives.
 After some more tinkering and refactoring, it would seem that Channels do actually support concurrent asynchronous broadcasts of completion streams, which does not fit my understanding of how they work. 
 Conversely, I'm now struggling to get BroadcastBlocks to work with a rare racing condition.
 Accordingly, I added logging and many unit tests to investigate those issues, and there will be some cleanup whether we decide for a single All-use-cases best implementation or not.
 
 In the mean time, I would really appreciate your feedback on those complicated issues.

dmytrostruk:
@jsboige Thank you for the updates. I will review this PR and reply shortly.

jsboige:
> @jsboige Thank you for the updates. I will review this PR and reply shortly.

Hi @dmytrostruk, do you have any news concerning this PR? I'll be happy to contribute some more, but since this has been open for quite some time now, and it does add a new project to the solution, it would be nice if we can wrap it up and move on.

shawncal:
@jsboige Thanks for all the hard work! We'll help take this home.
Note: Our PR process requires that all changes be synced up with main before merge, and that's obviously a moving target. Normally we'd merge the latest changes for you before merging your PR, but it looks like you have the "Allow edits by maintainers" box turned off, so we can't merge into your PR branch.

I've merged the whole PR into a feature branch in the sk repo, where we'll easily be able to address any conflicts and @dmytrostruk will bring it into main. If there are any further changes you'd like to push in the meantime, please add them to the "feature/oobabooga" branch.

Thanks again!

jsboige:
Hi @shawncal , thanks for moving on with merging this PR.
About the "Allow edits by maintainers" feature, it appears that there is [a bug](https://github.com/orgs/community/discussions/5634) preventing to enable that feature when the fork was made under an organization instead of a user, which is the case with this fork. Accordingly, I created a new Fork under my username and will start a new PR to that new branch from that new fork to help with finishing this up.

iamianM:
Any chance this could be done for the python package please?

PR #1366, opened 2023-06-07
Add JSON helpers to ContextVariables

anthonypuppo:
### Motivation and Context

JSON is an incredibly popular object format and is commonly used when dealing with complex object types. More often then not, prompts will attempt to get the LLM to produce well-formed JSON for easy parsing. This change adds helper methods to make working with JSON much easier in the context of the string dependent ContextVariables class (and in turn, SKContext).

### Description
Add helper methods for setting, updating and getting values from ContextVariables as strongly typed JSON values.

### Contribution Checklist

- [✔️] The code builds clean without any errors or warnings
- [✔️] The PR follows SK Contribution Guidelines (https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md)
- [✔️] The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`
- [✔️] All unit tests pass, and I have added new tests where possible
- [✔️] I didn't break anyone :smile:


anthonypuppo:
> @anthonypuppo please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

stephentoub:
Thanks. I'm sympathetic to the premise here, but FWIW, I have a few concerns with the approach. 

First, this won't be very friendly with Native AOT. Any use of this API directly or indirectly will result in warnings and require special steps on the consumer to ensure it actually works at run time by preventing the trigger from removing APIs required for the serialization/deserialization of the T.

Second, https://github.com/microsoft/semantic-kernel/pull/1195 is adding support for SKFunctions to convert arguments and return values between Ts and strings based on TypeConverter, which is the same mechanism used for example by ASP.NET for converting between text-based request/response data and handler arguments/returns. This would create a competing conversation story in the same set of types.

Third, this doesn't achieve anything a caller couldn't do on their own. They're effectively 1-2 line helpers that combine calling Serialize/Deserialize and set/get on the dictionary.

And fourth, while I certainly encourage devs to use System.Text.Json, this bakes JsonSerializer-related types into public API on core types in SK.

anthonypuppo:
@stephentoub 

1) Not familiar with the intricacies of native AOT enough to comment here. Your point seems valid.

2) Is the goal of SK to be used only on web platforms? While I love the idea of re-using already proven mechanisms and methodologies, this only makes sense to native ASP.NET developers. I think this style will get confusing quickly for individuals not familiar with ASP.NET and not translate well to other languages/frameworks. Just my opinion. If the TypeConverter mechanism is already agreed upon as the way forward then I agree this addition is competing and should not be added.

3) Agree, but I could argue the string overloads "are effectively 1-2 line helpers that the caller could do on their own with the TrustAwareString object" with this same logic though, as the caller can just read the Value property of the returned object. This is a silly example, but you get the point.

4) Agree.

With all that being said, I'm perfectly fine with this being closed, as I think the cons you listed outweigh the benefits a few short helper methods. Maybe similar "helpers" could be added to an extensions package going forward?

stephentoub:
(All that said, this is my $.02.  @dluc or @shawncal would need to weigh in.)

shawncal:
@anthonypuppo Thanks for the contribution, and apologies for the delay.
What you've brought up here is super valid -- we'll be aiming to make the variables going in and out of SK-compatible functions as easy and intuitive as possible. We have some bigger changes afoot, in terms of how ContextVariables/SKContext are handled and automatic mapping of primitive types and objects. Writing/result results as json is also on the docket. I'll pin this PR and use it to gut check whether the new changes solve what you're doing here. If not, we may want to bring these back as extension methods.

I'm going to close for now, but rest assured we've got better variable handling coming soon. Thanks again!

@stephentoub as FYI

PR #1367, opened 2023-06-07
Make MaxTokens nullable

### LINKED ISSUE #1362: Make MaxTokens optional
Some LLM's, such as ChatGPT, default to using the remaining context window if the maximum amount of tokens is not included in the request. For such models it would be beneficial to not pass the max tokens argument every time if the intention is to get the most elaborate response possible. This is of course possible by calculating remaining tokens, but is a pain (and performance hit) to constantly calculate the remaining tokens (assuming you even get it right).

anthonypuppo:
### Motivation and Context

Some LLM API's have the "max tokens" parameter as optional (OpenAI, Azure OpenAI, etc.) and will default to the remaining tokens left in the models context window. This PR changes the default value of max tokens from an explicitly defined amount to null. Note that skill specific max token amounts are unchanged, this PR aims to just change the defaults for library consumers right now.
Fixes #1362 

### Description

Change max tokens to a nullable int type (`int` to `int?`).

### Contribution Checklist

- [✔️] The code builds clean without any errors or warnings
- [✔️] The PR follows SK Contribution Guidelines (https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md)
- [✔️] The code follows the .NET coding conventions (https://learn.microsoft.com/dotnet/csharp/fundamentals/coding-style/coding-conventions) verified with `dotnet format`
- [✔️] All unit tests pass, and I have added new tests where possible
- [✔️] I didn't break anyone :smile:


dotnet/src/SemanticKernel/SkillDefinition/InlineFunctionsDefinitionExtension.cs
```diff
@@ -43,7 +43,7 @@ public static class InlineFunctionsDefinitionExtension
         string? functionName = null,
         string? skillName = null,
         string? description = null,
-        int maxTokens = 256,
+        int? maxTokens = null,
```
lemillermicrosoft, line 46: I think we should leave defined defaults here and above but still allow users to explicitly set them to null. Thoughts?

anthonypuppo: My main reasoning for defaulting the base methods to null was because, in my opinion, anyone using an LLM that requires an explicit token count passed is already counting tokens and looking to set this value. If not, the API should handle setting a sensible default. Passing no value is more "future proof" as things evolve.

If an explicit value is to be kept, why is 256 (or any other count for that matter) a good default? Both OpenAI and Microsoft OpenAI completion API's default to 16. I'd argue this behavior is already strange in that the alignment there doesn't match up. We're also looking at tokens through how OpenAI tokenizes, what if a model releases where 256 is past the model limits or even not enough tokens to account for a full word? The developer should be fully aware of what LLM they are using and know whether or not they should be passing a token limit or letting the API default to what it needs.

shawncal (approved):

PR #1391, opened 2023-06-09
TextCompletion Fixes + Interface simplification

RogerBarreto:
### Motivation and Context

Previous interfaces where too verbose and could be confusing. 
Removed the extra `Completion` verbosity from the interfaces as marked the old ones as Obsolete.

Small bug identified when using the TextCompletion Streaming (awaiting indefinitely)

### Description

- Renamed interface `ITextCompletionResult` to `ITextResult`
- Renamed interface `ITextCompletionStreamingResult` to `ITextStreamingResult`
- BugFix TextCompletion Streaming  (was not behaving as expected - awaiting indefinitely for ITextStreamingResult) 
- Added remarks on the `Complete` Extension methods regarding no support for multiple results.
- Added remarks on the `CompleteStreaming` Extension methods regarding no support for multiple results.
- Added remarks to extensions methods that don't support multiple results
  - IChatCompletion.GenerateMessageStreamAsync
  - IChatCompletion.GenerateMessageAsync
  - ITextCompletion.CompleteStreamAsync
  - ITextCompletion.CompleteAsync

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextResult.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Orchestration;
+
+namespace Microsoft.SemanticKernel.AI.TextCompletion;
+
+public interface ITextResult
```
lemillermicrosoft, line 9: nit: Xml docs

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextResult.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Gets the model result data.
```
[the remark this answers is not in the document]
RogerBarreto: After quick investigation Azure SDK don't support Insert and Edit models.

`Insert` completions actually are Text Completions using an extra parameter for suffixes.

`Edit` completions if implemented would need it's dedicated interfaces and implementation. 
Not aware of a big ask from the community to support `Edit` models, if this becomes big ask we work on a new feature PR for this.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/ITextResult.cs
RogerBarreto: Done, added on other files as well.

lemillermicrosoft (approved):

lemillermicrosoft (changes requested):
As discussed, since this might break anyone who was using these interfaces directly, let's bump the version number.

lemillermicrosoft (approved):

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatCompletionExtensions.cs
```diff
@@ -56,6 +59,7 @@ await foreach (var chatMessageStream in chatCompletionResult.GetStreamingChatMes
         {
             var chatMessage = await chatResult.GetChatMessageAsync(cancellationToken).ConfigureAwait(false);
             messageContent.Append(chatMessage.Content);
+            break;
```
shawncal, line 62: If this is always just one result, we should get rid of the foreach loop and the StringBuilder

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -39,20 +39,22 @@ public static class TextCompletionExtensions
...
         string text,
         CompleteRequestSettings requestSettings,
         [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
         var completionResults = textCompletion.GetStreamingCompletionsAsync(text, requestSettings, cancellationToken);
-        var firstResult = await completionResults.FirstOrDefaultAsync(cancellationToken).ConfigureAwait(false);
-        if (firstResult is not null)
+
+        await foreach (var completionResult in completionResults)
         {
```
shawncal, line 52: Wasn't this more clear before? The new loop doesn't do any looping if it exits at the end of the first iteration

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -367,13 +367,16 @@ private struct MethodDetails
...
+    private static async Task<string> GetCompletionsResultContentAsync(IReadOnlyList<ITextResult> completions, CancellationToken cancellationToken = default)
     {
         StringBuilder completionResult = new();
 
-        foreach (ITextCompletionResult result in completions)
+        foreach (ITextResult result in completions)
         {
             completionResult.Append(await result.GetCompletionAsync(cancellationToken).ConfigureAwait(false));
+
+            // To avoid any unexpected behavior we only take the first completion result (when running from the Kernel)
```
shawncal, line 378: Why have a List arg, if we are only returning the contents of the first item? I'd say the method should be more clear -- `Task<string> GetCompletion*ResultContentAsync(ITextResult completion, ...)`
or
`Task<IReadOnlyList<string>> GetCompletionsResultContentAsync(IReadOnlyList<ITextResult> completions`

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
RogerBarreto: When I used the FirstOrDefaultAsync a bug that I cannot understand is happening where it freezes the interation of the message loop. 

Keeping the foreach statement don't surface the bug works as a quick fix as all works as expected.

I can dive more into the problem in another PR, as this might need some more detailed investigation on how the AsyncEnumerable internals work.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
RogerBarreto: This method initially was combining the multiple results in one and I removed this behavior and instead of removing the whole method I kept it because in near term will be touching this need to return multiple results in the same prompt.

But you are right here, instead of keeping this method I can just remove it and rewrite it again in a later time when we fully support multiple responses as part of RunAsync impl.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
RogerBarreto: @stephentoub Thanks for the help, as it was understood using the `FirstOrDefaultAsync` kills the StreamingCompletions object which actually handles the streaming process leaving the inner `foreach` loop in the limbo.

The approach to resolve it would be to either not changing the foreach and keeping the choice yield alive while consuming the inner message/text or making ITextResult IDisposable and handling the disposing down the line.

For the sake of simplicity, I'll keep the outer yield so .net runtime can hold the resource for the time all the inner messages of the first choice are consumed.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatCompletionExtensions.cs
RogerBarreto: Agreed. Done.

shawncal (approved):

PR #1396, opened 2023-06-09
Java: Ensure Java formatting when editing on VS Code / IntelliJ

brunoborges:
This PR enables a Run On Save on VS Code that will format the modified Java code according to the project settings (Google Format with AOSP style).


.vscode/extensions.json
```diff
@@ -7,6 +7,8 @@
     "ms-dotnettools.dotnet-interactive-vscode",
     "esbenp.prettier-vscode",
     "dbaeumer.vscode-eslint",
-    "ms-semantic-kernel.semantic-kernel"
+    "ms-semantic-kernel.semantic-kernel",
+    "emeraldwalk.RunOnSave",
+    "ms-java.vscode-java-pack",
   ]
```
markwallace-microsoft, line 13: Please split into two PR's.

1. `.vscode` changes and PR targets main. Next time main is merged into the `experimental-java` branch we will get those changes.
2. Changes for `java` folder which target the `experimental-java` branch

markwallace-microsoft (changes requested):

johnoliver:
Is the only way to get formatting on vs code to have a jar checked in to the repo?

d3r3kk:
> Is the only way to get formatting on vs code to have a jar checked in to the repo?

Not likely. The plugin/extensions for VSCode likely pull in whatever dependencies you need instead. I'd take a look at `.vscode/extensions.json` for dictating the recommended extensions for this project.

brunoborges:
This was the only true cross-platform way to get Google Java Format with AOSP style to work on VS Code.

brunoborges:
Commit [47cf4d0](https://github.com/microsoft/semantic-kernel/pull/1396/commits/47cf4d0c8ec0d7fd7de24489168bc346d82dc768) moves the formatting plugin to the general profile of the Maven project.

This way, no matter when/how the Maven project is built (command-line, terminal, etc), the code will be formatted.

But the VS Code setting is still required as it ensures new code comes with the right formatting, and solves the problem of when a developer adds a Java class without running a build (unlikely, but still...)

.vscode/settings.json
```diff
@@ -3,15 +3,14 @@
   "css.lint.validProperties": [
```
markwallace-microsoft, line 3: Changes to this file are a potential source of merge conflicts when updating the `experimental-java` branch. Can you create a seperate PR for these changes and target the `main` branch and we will get the change merged to `main` and then the `experimental-java` branch after that.

markwallace-microsoft (changes requested):

.vscode/settings.json
brunoborges: See [#1552](https://github.com/microsoft/semantic-kernel/pull/1552)

PR #1400, opened 2023-06-10
CopilotChat: Fix prompt and improve creation for Document messages 

dehoward:
### Motivation and Context

Fixes the prompt for Document messages to be a natural language string vs the JSON messaged that is stored.
Also changes the import flow to save messages in the same call vs going back to the server again. 

### Description

- updates `getFormattedString()` in `ChatMessage` to return a human readable for Document messages
- updates `DocumentImportController` to create the `ChatMessage` which removes the need to go to the `ChatSkill` after the `importDocument` call returns. 

### Before

image

### After
image

### Contribution Checklist


samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -42,10 +44,29 @@ private enum SupportedFileType
         Pdf,
     };
 
+    /// <summary>
+    /// Value of `Content` for a `ChatMessage` of type `ChatMessageType.Document`.
+    /// </summary>
+    public struct DocumentMessageContent
```
TaoChenOSU, line 50: Show we define this in the ChatMessage model instead?

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -125,6 +149,18 @@ private enum SupportedFileType
 
             await this._sourceRepository.UpsertAsync(memorySource);
 
+            // Create chat message that represents document upload
+            chatMessage = new ChatMessage(
+            memorySource.SharedBy,
```
TaoChenOSU, line 154: nit: tab

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/ChatMessage.cs
```diff
@@ -150,7 +151,14 @@ public static ChatMessage CreateBotResponseMessage(string chatId, string content
     /// <returns>A formatted string</returns>
     public string ToFormattedString()
     {
-        return $"[{this.Timestamp.ToString("G", CultureInfo.CurrentCulture)}] {this.UserName}: {this.Content}";
+        var content = this.Content;
+        if (this.Type == ChatMessageType.Document)
+        {
+            var documentDetails = JsonSerializer.Deserialize<DocumentMessageContent>(content);
```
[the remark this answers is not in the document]
dehoward: I don't think so? not unless I make define another field to specifically store a `DocumentMessageContent`, which I wanted to avoid.

otherwise, we need to serialize the struct when saving it to `Content` and then de-serialize it when we use `Content` in `ToFormattedString()`. am I missing something here you think?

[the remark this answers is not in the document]
TaoChenOSU: I was thinking about storing a DocumenetMessageContent too. It can be null when the message is not a document message. Since you wanted to avoid that, I think we are all good!

TaoChenOSU (approved):

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -120,14 +120,15 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
...
+                const message = await documentImportService.importDocumentAsync(
                     account!.homeAccountId!,
+                    (account!.name ?? account!.username) as string,
                     selectedId,
                     documentFile,
                     await AuthHelper.getSKaaSAccessToken(instance, inProgress),
                 );
 
-                handleSubmit(JSON.stringify(document), ChatMessageType.Document);
+                dispatch(updateConversation({ message, chatId: selectedId }));
```
teresaqhoang, line 131: Do we plan to support document import from anywhere else in the future (i.e., the Files tab)? If so, I'd move this to the useChat hook in one method that handles both importDocument and updateConversation 

(or create a separate useSources hook)

dehoward: moved to `useChat`

TaoChenOSU (dismissed):

teresaqhoang (dismissed):

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -42,10 +45,48 @@ private enum SupportedFileType
         Pdf,
     };
 
+    /// <summary>
+    /// Value of `Content` for a `ChatMessage` of type `ChatMessageType.Document`.
+    /// </summary>
+    public class DocumentMessageContent
```
adrianwyatt, line 51: Let's continue to put public classes in their own files.

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/ChatMessage.cs
```diff
@@ -150,7 +151,14 @@ public static ChatMessage CreateBotResponseMessage(string chatId, string content
     /// <returns>A formatted string</returns>
     public string ToFormattedString()
     {
-        return $"[{this.Timestamp.ToString("G", CultureInfo.CurrentCulture)}] {this.UserName}: {this.Content}";
+        var content = this.Content;
```
adrianwyatt, line 154: [nit] Remove this and just use 'this.Content' everywhere.

adrianwyatt (changes requested):

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/ChatMessage.cs
dehoward: synced offline, leaving as-is as this is by design

samples/apps/copilot-chat-app/webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: fixed

adrianwyatt (approved):

PR #1401, opened 2023-06-10
Java: Nullable to Optional

brunoborges:


brunoborges:
@johnoliver any thoughts on this one?

dsgrieve:
So... I'm not a big fan of this one. The litmus test for Optional is whether the API is meant to be used in a stream/fluent way. Even then, you can do without Optional. If you look at the changes you have here, you see a lot of the `Optional<Blah> blah = whatever(); if (blah.isPresent()) doSomething()` anti-pattern. If we do move forward with this, then the implementation should be more fluent. I think you'd be hard-pressed to change NativeSKFunction#getArgumentValue into a fluent style (you could, but you'd need to break it down into smaller functions).

brunoborges:
Thanks @dsgrieve for the review. I am ready to give up on this proposal as the benefits are definitely not greater than the disadvantages (as experienced by the introduction of bugs).

PR #1413, opened 2023-06-12
CopilotChat: Add drag and drop for file upload

dehoward:
### Motivation and Context

gives users the ability to drag files into the textbox for upload.

### Description

- adds additional styling to the `textarea` in `ChatInput` when a file is being dragged over the element
- uses the existing `importDocument` handler to import the dropped file
> there is a slight lag from when the file is dropped vs when the upload starts that seems to be a result of when the browser calls `onDrop` for the `textarea`

### Preview
gif

### Contribution Checklist


samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
```diff
@@ -160,16 +170,41 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
...
+                        className: isDraggingOver
+                            ? mergeClasses(classes.dragAndDrop, classes.textarea)
+                            : classes.textarea,
+                    }}
                     className={classes.input}
-                    value={value}
+                    value={isDraggingOver ? 'Drop your files here' : value}
+                    onDragEnter={handleDragEnter}
+                    onDragOver={handleDragEnter}
+                    onDragLeave={handleDragLeave}
```
teresaqhoang, line 206: nit: `onDragEnter` and `onDragOver` should be properties on the ChatRoom component so the user knows where to drop the file upon entering the view

teresaqhoang (changes requested):

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatInput.tsx
dehoward: good idea, moved

hathind-ms (approved):

teresaqhoang (approved):

PR #1416, opened 2023-06-12
Python: Fix windows runtime error: adapt asyncio event loop policy for windows python 3.10 and above

imfuli:
… windows

### Motivation and Context

To fix a runtime error on windows for python version 3.10, 3.11 and above

On windows, python version above 3.10.1, there will be a runtime error caused by asyncio event loop. add adapting code for windows when python version is above 3.10.1 to avoid the error.

### Description

Without this fix, the example script will runinto a runtime error, shows as below:

'''
Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x00000289E4B2C550>
Traceback (most recent call last):
  File "C:\Program Files\Python3.10\lib\asyncio\proactor_events.py", line 116, in __del__
    self.close()
  File "C:\Program Files\Python3.10\lib\asyncio\proactor_events.py", line 108, in close
    self._loop.call_soon(self._call_connection_lost, None)
  File "C:\Program Files\Python3.10\lib\asyncio\base_events.py", line 750, in call_soon
    self._check_closed()
  File "C:\Program Files\Python3.10\lib\asyncio\base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
'''

The change only handle windows platform and python version above 3.10.1.

Refer link:

https://docs.python.org/3/library/asyncio-policy.html

"Note In Python versions 3.10.9, 3.11.1 and 3.12 the [get_event_loop()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop) method of the default asyncio policy emits a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) if there is no running event loop and no current loop is set. In some future Python release this will become an error."

Note: per testing, this error also occurred on python 3.10 below 3.10.9

### Contribution Checklist


mkarle (changes requested):
So I've noticed this problem occurring too. However, I see it in versions >= 3.8 but <= 3.10.1.  I ran the unit tests with 3.8 and get the same Runtime error but with 3.10.11, I do not.

imfuli (commented):
updated per Mark's comment on the python version to be adapted.

imfuli:
@mkarle Hello Mark, I've updated the fix, please go ahead the review? thanks.

mkarle (approved):

mkarle (approved):
Needed to lower the version to 3.8.0

mkarle (approved):

PR #1452, opened 2023-06-13
Fix ESLint in samples/apps

dehoward:
### Motivation and Context

correctly using ESLint in our sample apps will help us
- follow best practices, 
- cut down on bugs that the linter can catch, and
- enforce a unified style

### Description

- adds `.env.local` to each app to disable ESLint in all environments
  - temporary until the violations are fixed in the app
  - this seemed to be the easiest way to disable ESLint without folks having to do anything, open to suggestions here.
  - this is needed because now the build picks up on our config file and will fail since there are currently violations
- adds a `package.json` to `samples/apps` for the shared dependencies needed for running `eslint`
- adds the `npm run lint` script and `eslint` package to apps where not already present
- removes unneeded ESLint config/dependencies from CopilotChat
- renames `.eslingrc.js` to `.eslintrc.js`

### Note
**this doesn't start enforcing ESLint anywhere or fix any violations**, and rather just fixes the config so that we can begin to correct the errors.

there are quite a number of violations just in CopilotChat alone:
image

`eslint --fix` can fix about half of these, albeit we should review these changes carefully if used.

### Contribution Checklist


.github/workflows/node-pr.yml
```diff
@@ -39,7 +39,9 @@ jobs:
           shopt -s globstar # enable globstar option to use ** for recursive matching
           for lockfile in samples/**/yarn.lock; do # loop over all yarn.lock files
             dir=$(dirname "$lockfile") # get the directory of the lock file
-            echo "Running yarn install and yarn build for $dir"
-            yarn --cwd "$dir" install --frozen-lockfile # install dependencies
-            yarn --cwd "$dir" build # run build script
+            if [[ "$dir" != "samples/apps" ]]; then # exclude samples/apps directory
```
TaoChenOSU, line 42: We don't have a samples/apps/yarn.lock. Will $dir = "samples/apps"?

teresaqhoang: Does running `yarn install` in the current working directory also install the dependencies defined in parent directories? Or would we have to run `yarn install` in both directories to install all the dependencies?

teresaqhoang (dismissed):

.github/workflows/node-pr.yml
dehoward: it's currently the latter but it would be nice to have the former. I'll look into adding the workspace

dehoward: I looked into Yarn workspaces but it doesn't seem that adding one would enable the behavior you mentioned, and rather would allow us to have dependencies that need to exist in each app, be resolved from the root directory.
for us this isn't necessary, as the dependencies only need to exist in the root directory, since ESLint is smart enough on its own to resolve its dependencies from the directory that the `.eslintrc.js` lives in.

but regardless, at least from what I gathered in my limited research, `yarn install` would still need to be ran in both the root and subfolders.

TaoChenOSU: Should we update the ReadMes to reflect this change in the setup step?

dehoward: definitely, but not right now I don't think. I'm still making sure we'll have the correct setup so that we won't conflict with the default `react-scripts` ESLint config. right now, no changes in the setup are necessary since ESLint is disabled in the build.

TaoChenOSU (approved):

TaoChenOSU (approved):

teresaqhoang (approved):

Issue #1455, opened 2023-06-13
Python Notebook Sample 08 missing AzureTextCompletion references


mathyousee:
**Describe the bug**

The [08-native-function-inline](https://github.com/microsoft/semantic-kernel/blob/main/samples/notebooks/python/08-native-function-inline.ipynb) sample, the second cell does not import `AzureTextCompletion`. If the _useAzureOpenAI_ variable is set to `True`, the cell fails to complete.

Additionally, in the same notebook the first cell in the *Context Variables* section does not include the switch for AzureOpenAI.

**To Reproduce**
1. Open [samples\notebooks\python\08-native-function-inline.ipynb](https://github.com/microsoft/semantic-kernel/blob/main/samples/notebooks/python/08-native-function-inline.ipynb)
2. In the second cell, set the _useAzureOpenAI_ variable to `True`
3. Run the second cell

**Expected behavior**
The cell should import the appropriate components and function properly for the Azure OpenAI option.

PR #1456, opened 2023-06-13
Python: added reference to AzureTextCompletion to notebook

mathyousee:
### Motivation and Context
This is addressing issue#1455 https://github.com/microsoft/semantic-kernel/issues/1455

### Description
In the second cell, I added the missing reference to Azure Text Completion.

In the _Context Variables_ section, I updated the first cell to include an option to run using Azure OpenAI (was missed previously). This follows the same format as used in other cells.


alexchaomander:
Thanks for the fix!

alexchaomander (approved):

PR #1465, opened 2023-06-13
Add chat system prompt option for text completion

lemillermicrosoft:
This change adds a new property to the CompleteRequestSettings class that allows specifying a custom system prompt to use when generating text completions using a chat model. This can be useful for setting the tone and context of the conversation. The default value is "Assistant is a large language model." which is the same as the OpenAI API.

The change also modifies the ClientBase class to pass the chat system prompt to the InternalCreateNewChat method, which creates a new instance of the OpenAIChatHistory class. This way, the chat history will include the system prompt as the first message.

### Contribution Checklist


lemillermicrosoft:
Greatly helps scenarios like #1468

shawncal (approved):

Issue #1220, opened 2023-05-25
.Net: Qdrant (and other memory stores?) - support rigid filtering


piotrek-appstream:
I have structured and unstructured data that I need to search using both vector similarity and structured data filtering.
Qdrant supports [payload filtering](https://qdrant.tech/documentation/filtering/) alongside vector similarity (as far as I know, Pinecone does as well), but the `IMemoryStore` interface does not currently support it.

What do you think would be the best way to address this issue?

- Support rigid filtering through the `IMemoryStore` interface. I'm willing to work on this if it's decided that this is the approach to take.
- Create my own class that inherits from `QdrantVectorDbClient` (or just copy it) and extends the base class with methods that support rigid filtering. I would then use this extended class explicitly in a skill.
- Is there another recommended approach?

Thank you in advance.
P.S. This library is really cool! 👌

evchaki:
@piotrek-appstream , great idea!  @awharrison-28  and @tawalke  does this look like the right approach?

lukasz-appstream:
I agree that there is value in integrating semantic search capabilities with filtering functionalities for the structured segment of data. However, simply extending `IMemoryStore` may not be sufficient. Currently, it is not exposed either through `SkContext` or `SemanticKernel`, correct?

I was thinking about a solution that would allow us to use it seamlessly within the skill, just through `context.Memory`. What are your thoughts on the following options:

 - Extending `ISemanticTextMemory` with the following method:
 ```    
 public IAsyncEnumerable<MemoryQueryResult> SearchStructuredAsync(
        string collection,
        string query,
        int limit = 1,
        double minRelevanceScore = 0.7,
        bool withEmbeddings = false,
        IFilterExpression? filters = default, // some kind of expression to filter by payload fields
        CancellationToken cancellationToken = default)
```
This way, in the skill, we can simply use `context.Memory.SearchStructuredAsync`.
        
 - Introducing a new interface, `ISemanticStructuredMemory`, specifically designed for partially structured memories. It could be configured using `.WithStructuredMemoryStorage(...)` and utilized in the skill as `context.StructuredMemory`. I understand this option may require more effort, but it would allow us to tailor methods like `SaveInformation` or `SaveReference` in the new interface to better suit partially structured data.

lukasz-appstream:
@evchaki we have a PR ready for that, first with Qdrant, but designed in a way to be extendible for other memory stores without limiting their capabilities (source specific generic filter argument).

[...]

PR #1481, opened 2023-06-14
DotNet: Support for rigid filtering in SemanticTextMemory and Qdrant

piotrek-appstream:
### Motivation and Context

The motivation behind this pull request is to enhance the functionality of SK by adding support for rigid filtering in Qdrant. Currently, SK only supports looking up vectors based on vector similarity. However, Qdrant also provides the capability to apply rigid filters during vector retrieval, which allows for more fine-grained control over the search results.

By adding support for rigid filtering functionality, developers using SK will be able to leverage this additional feature of Qdrant. This change addresses the limitation of SK in terms of accessing and utilizing the full potential of Qdrant's search capabilities.

Open issue: https://github.com/microsoft/semantic-kernel/issues/1220

### Description

This pull request introduces the following changes to the SK SDK:

#### Kernel building
- `MemoryConfiguration.UseMemory` methods set `ISemanticTextMemory<TFilter>` as kernel's memory if the registered memory store supports filtering (since there's no DI in the kernel building process, reflection is used to instantiate the `SemanticTextMemory<TFilter>` - if the constructor changes tests will fail).

#### Under the hood
- Two new interfaces: `IMemoryStore<TFilter>` and `ISemanticTextMemory<TFilter>` exposing methods that allow searching for vectors with metadata/payload filters where `TFilter` is a database vendor specific filter object.
- `QdrantMemoryStore` implementing `IMemoryStore<QdrantFilter>`.

#### Usage
- `IKernel.GetFilterableMemory<TFilter>()` method to obtain an instance of `ISemanticTextMemory<TFilter>` if such memory has been registered (returns `null` otherwise).
 

### Contribution Checklist


goyanx:
Hi, interested to use this :)
How would you code this in the filter? Is it supported?

example

 "filter": {
        "should": [
            {
                "key": "city",
                "match": {
                    "value": "London"
                }
            }
        ]
    }

piotrek-appstream:
> Hi, interested to use this :) How would you code this in the filter? Is it supported?
> 
> example ` "filter": { "should": [ { "key": "city", "match": { "value": "London" } } ] }`

Right now only `must` is supported, but in your case `"filter": { "must": [ { "key": "city", "match": { "value": "London" } } ] }` would yield the same results, so you could construct the filter like this:
```
var filter = new QdrantFilter();
filter.Must(new MatchCondition
{
  Key = "city",
  Match = new Match { Value = "London" }
});
```
and then you could use this in a skill like this:
```
public async Task<Result> SearchSkill(string input, SKContext context)
{
  var memory = context.GetFilterableMemory<QdrantFilter>();
  var results = memory.SearchAsync(
    collectionName, // qdrant collection name
    query, // text query for vector similarity
    filter); // rigid filters
}
```
Sorry for the late reply

shawncal:
@piotrek-appstream Thanks for the contribution! I see what you're enabling here -- it's something that our users have asked for in various forms, and really valuable.

We're going to be doing some refactoring on the Kernel.Core and Kernel.Abstractions packages in the couple weeks ahead, to slim down and simplify the SDK, and to put things into discrete, composable packages. As part of this, the memory store registration methods on the IKernel will be removed, and the memory classes moved to their own package. This should ensure better extensibility, and allow us to revise memory faster and on a different timeline than the kernel.core itself.

Given the significant changes ahead, I'm going to flip this to draft for now. After the packaging changes are complete, we can revisit this to see what we bring forth into the memory namespace.

Thanks again!

piotrek-appstream:
@shawncal Sounds like a good direction. Thanks for the update 👍

lukasz-appstream:
@shawncal Hello Sir, any news here? Should start working on resolving conflicts? If not, is there a timeline or a milestone for rearranging the Kernel.Abstractions and Kernel.Core packages?

Thanks!

glorious-beard:
@shawncal - is there a design doc previewing what changes the refactoring brings in terms of the API? does that refactoring fundamentally change the semantic text memory API, or is it just making implementation of filtering a little more straightforward?  if it's the latter, why not at least accept this PR in the interim since you're at a minimum another several weeks out from completing the refactoring, per @evchaki on #2485?

shawncal:
Closing, as this is coming with https://github.com/microsoft/semantic-memory

lukasz-appstream:
@shawncal Just a remark - in our approach we use filter object that is designed for a specific client.

This allows using filters like:
- 'contains' for list field
- 'greater/lower' for numeric field
- even operations on coordinate field
- and more

when the underlying memory storage supports it.

If we only allow tags - it limits functionality a lot and it would be often very useful when using SK with, for example, map search.

IMHO the only thing we gain from this simplistic filtering is easy memory storage switching which isn't something people do often.

Is there a plan (and an idea how) to improve the very basic "requiredTags" filtering?

Thanks,
Lukasz

PR #1496, opened 2023-06-14
Python: Filter only json content from basic planner execution input

### LINKED ISSUE #1469: Python: basic planner generated plan sometimes contains generated text that is not json de-serializable.
The python basic planner sometimes generates text that is not json de-serializable. This manifests as additional text beyond the generated plan. 

Ex:

```
{
    "input": "Valentine's Day Date Ideas",
    "subtasks": [
        {"function": "WriterSkill.Brainstorm"},
        {"function": "WriterSkill.Translate", "args": {"language": "French"}},
        {"function": "TextSkill.uppercase"}
    ]
}
additional text....
```

Some proposals for resolving this issue are:

1. Add stop sequences to ensure that once the json is rendered, additional text generation is halted. This has a downside where it is not guaranteed that the new text will appear in a certain form or pattern.
2. Created a regex to extract the json from the generated text. This is a little more wasteful of tokens, but better guarantees that the json plan will be captured without any additional information. This will also require that the kernel be dependent on the `regex` pip package as `re` does not support recursive matching.

awharrison-28:
### Motivation and Context
Closes #1469 

There are scenarios where the basic_planner prompt results in a valid plan surrounded by additional text. This renders the plan unable to be json deserialized. This PR addresses this scenario. Note that it does NOT address malformed json.

### Description
- added a regex in basic_planner.execute_plan_async that grabs only json blocks. See screenshot below of it ignoring additional, non-json text. 
- Added dependency on `regex` pip package (`re` package does not support regex recursion)

image

### Contribution Checklist


shawncal (approved):

python/semantic_kernel/planning/basic_planner.py
```diff
@@ -187,7 +189,13 @@ async def execute_plan_async(self, plan: Plan, kernel: Kernel) -> str:
         Given a plan, execute each of the functions within the plan
         from start to finish and output the result.
         """
-        generated_plan = json.loads(plan.generated_plan.result)
+
+        # Filter out good JSON from the result in case additional text is present
+        json_regex = r"\{(?:[^{}]|(?R))*\}"
```
dluc, line 194: couple of notes:
* is the regex correct? I'm getting parse errors, e.g. at `?R`
* could you add a comment explaining the regex?
* maybe add unit tests?
* is the regex safe (redos)?

awharrison-28: Here's a breakdown of the regex pattern \{(?:[^{}]|(?R))*\}:

- `\{`: This part of the pattern matches the opening curly brace {.
- `(?:`: This part is the beginning of a non-capturing group. Non-capturing groups allow you to apply quantifiers to part of your regex, without capturing the matched text.
- `[^{}]`: This part of the pattern is a character class that matches any character except the curly braces { and }.
- `|`: This is the alternation (OR) operator, which means the pattern can match either the left side (any character except curly braces) or the right side (the next part of the pattern).
- `(?R)`: This part of the pattern is a regex recursion. It refers to the entire regular expression pattern, allowing the pattern to be applied recursively. This is useful for matching nested structures like JSON.
- `)*`: This part of the pattern is a quantifier that means "match zero or more of the preceding pattern." In this case, the preceding pattern is the non-capturing group that contains the character class and the recursion.
- `\}`: This part of the pattern matches the closing curly brace }.

Overall, the regex pattern can be described as follows:
Match an opening curly brace.
Match any number of characters that are not curly braces or recursively apply the entire regex pattern (to match nested JSON objects).
Match a closing curly brace.

PR #1507, opened 2023-06-15
Fix variable expansion in plan steps

### LINKED ISSUE #1487: SequentialPlanner + OpenAPI Plugin: Variable keys are blanked out in payload
**Describe the bug**
Variable names that start with a $ in an OpenAPI document specifically in the payload have their names set to an empty string before being submitted as an HttpRequest. This causes the HttpRequest to fail. 

**To Reproduce**
1. Import an OpenAPI plugin that contains variable names that start with $ such as '$top'
2. Create a plan using a sequential planner
3. Execute the plan
4. Http call fails

**Expected behavior**
The variable names should not be set to empty string

**Screenshots**

- Here you can see the variable is set in the plan step parameters
Variable is set
- Here you can see the variable name has been blanked out. This is inside of the ExecuteAsync method in the private method RegisterRestApiFunction of the KernelOpenApiExtensions
Variable name removed

**Desktop (please complete the following information):**
 - OS: Windows
 - IDE: Visual Studio 2022
 - 0.15.230609.2-preview

**Additional context**

- I started with the OpenAPI doc found [here](https://github.com/MicrosoftDocs/vsts-rest-api-specs/blob/master/specification/search/7.1/search.json) for Dev Ops search. It uses properties such as $top and $orderby in the code search http call

lemillermicrosoft:
### Motivation and Context
Fixes #1487 

This pull request adds support for expanding variables with a $ in them in plan parameters, such as JSON payloads. Previously, such variables would be replaced with an empty string or left as is, which could cause unexpected behavior or errors. This change allows the plan to use the values of the variables from the context or the plan state, and only leaves the variable as is if it is not found in either. This feature improves the flexibility and robustness of the plan execution and allows for more complex scenarios involving dynamic payloads.

### Description
Details:
- Add a unit test to verify that variables with a $ in them are expanded correctly in plan parameters
- Modify the PlanVariableExpansionTests to reflect the new behavior of leaving unknown variables as is instead of replacing them with an empty string
- Modify the Plan.ExpandFromVariables method to use Distinct() to avoid duplicate replacements, and to only replace variables that are found in the context or the plan state

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/Planning/Plan.cs
+++ b/dotnet/src/SemanticKernel/Planning/Plan.cs
@@ -402,13 +402,14 @@ public sealed class Plan : ISKFunction
     {
         var result = input;
         var matches = s_variablesRegex.Matches(input);
-        var orderedMatches = matches.Cast<Match>().Select(m => m.Groups["var"].Value).OrderByDescending(m => m.Length);
+        var orderedMatches = matches.Cast<Match>().Select(m => m.Groups["var"].Value).Distinct().OrderByDescending(m => m.Length);
 
         foreach (var varName in orderedMatches)
         {
-            result = result.Replace($"${varName}",
-                variables.TryGetValue(varName, out string? value) || this.State.TryGetValue(varName, out value) ? value :
-                string.Empty);
+            if (variables.TryGetValue(varName, out string? value) || this.State.TryGetValue(varName, out value))
+            {
+                result = result.Replace($"${varName}", value);
+            }
         }
 
         return result;
--- a/dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
+++ b/dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
@@ -880,4 +880,50 @@ Previously:Outline section #1 of 3: Here is a 3 chapter outline about NovelOutli
         // Assert
         Assert.Equal(expected, result.Result);
     }
+
+    [Fact]
+    public async Task CanExecutePlanWithExpandedAsync()
+    {
+        // Arrange
+        var kernel = new Mock<IKernel>();
+        var log = new Mock<ILogger>();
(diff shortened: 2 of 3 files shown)


dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -406,9 +406,10 @@ internal string ExpandFromVariables(ContextVariables variables, string input)
 
         foreach (var varName in orderedMatches)
         {
-            result = result.Replace($"${varName}",
+            var replaceString = $"${varName}";
+            result = result.Replace(replaceString,
                 variables.TryGetValue(varName, out string? value) || this.State.TryGetValue(varName, out value) ? value :
-                string.Empty);
+                replaceString); // leave as-is if not match found
```
shawncal, line 412: nit for readability: can you just do this?:
```
if (variables.TryGetValue(varName, out string? value) || this.State.TryGetValue(varName, out value))
{
   result = result.Replace($"${varName}, value)";
}
```

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -406,9 +406,10 @@ internal string ExpandFromVariables(ContextVariables variables, string input)
```
shawncal, line 406: Why `OrderByDescending(m => m.Length)`?

shawncal, line 406: Should we add a `.Distinct` here to remove duplicates? (The `.Replace(...)` below will replace all instances, so it would save an extra lookup, and the subsequent Replace calls for the same varName would not find the varName in the string anymore anyway).

lemillermicrosoft: This ensures that longer variable names are replaced first before shorter variable names.

shawncal (approved):

PR #1517, opened 2023-06-16
ISKFunction SimpleInvokeAsync 

RogerBarreto:
### Motivation and Context

We do have very simple scenarios where we just want to output the LLM result without having to manage or handle the SKContext result.

This is similar to what can be done with the `ITextCompletion.CompleteAsync` extension method with the additional benefit of using the Template Engine provided by the Kernel to execute your function (in case it's a Semantic Function)

### Description

Adding a `ISKFunction.SimpleInvokeAsync` extension methods to return the string result from a ISKFunction call.

### Contribution Checklist


dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
```diff
@@ -115,8 +115,99 @@ public static ISKFunction UseFrequencyPenalty(this ISKFunction skFunction, doubl
...
-        // Log a warning if the given input is overriding a different input in the context
+        var resultContext = await InternalInvokeAsync(
+             function,
+             input,
+             context,
+             mutableContext,
+             settings
+             ).ConfigureAwait(false);
+
+        return resultContext.Variables.Input;
```
lemillermicrosoft, line 149:
```suggestion
        return resultContext.Result;
```

lemillermicrosoft: Also below.

samples/dotnet/kernel-syntax-examples/Example06_TemplateLanguage.cs
```diff
@@ -48,7 +49,7 @@ public static async Task RunAsync()
 
         // Show the result
         Console.WriteLine("--- Semantic Function result");
-        var result = await kindOfDay.InvokeAsync();
+        var result = await kindOfDay.SimpleInvokeAsync();
```
lemillermicrosoft, line 52: These examples don't do a great job showing the benefit of this change. Can you add some new ones? I'm also concerned about the naming.

RogerBarreto: Not sure if I follow, this example is a really good one because it this scenario there was not need to manipulate the SKContext object or get specifics (its properties)

shawncal:
I don't fully understand when this would be used, and I don't think SimpleInvoke really explains how it's any different. Could these just be more "InvokeAsync" overloads?
Let's follow up on this change offline.

shawncal:
@RogerBarreto Flipping to draft for now. Let's look at this in context of the bigger changes we want to make with SKContext, etc.

PR #1538, opened 2023-06-17
Python: add Azure Credentials authentication and api_key renewal

ianchi:
### Motivation and Context
When using tokens obtained from Azure AD, they have an expiration, so in long running processes they need to be periodically renewed.
Currently there is no way of updating the OpenAI api_key other than accessing the private `_api_key` member.

### Description

Simplify authentication using Azure AD:
- enable auto login using `DefaultAzureCredential()`
- allow to pass a `TokenCredential`
- automatically renew of token before it expires (when using credentials)
- add method to manually update api_key, without having to access private members. Usefull if not using credentials

### Contribution Checklist


ianchi:
Hi, any comment on this?

shawncal:
Hi @ianchi, this looks like a good direction! Apologies for the delay. I'll assign to our Python crew, and we'll take it for a spin next week. 
Thanks for the awesome contribution!

python/semantic_kernel/connectors/ai/open_ai/services/azure_credentials_mixin.py
```diff
@@ -0,0 +1,65 @@
+import time
```
awharrison-28, line 1:
```suggestion
# Copyright (c) Microsoft. All rights reserved.

import time
```

awharrison-28 (changes requested):
Could you put together a quick syntax example under samples/python/kernel-syntax-examples?

Other than that and the missing copywrite header, this PR is great alignment with existing functionality in the .NET SK. Thank you!

awharrison-28:
@ianchi I added an example usage file under the kernel-syntax-examples folder and applied the pre-commit formatting.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
```diff
@@ -21,7 +30,7 @@ def __init__(
         api_key: Optional[str] = None,
         api_version: str = "2023-03-15-preview",
         logger: Optional[Logger] = None,
-        ad_auth=False,
+        ad_auth: Union[bool, str, TokenCredential] = False,
```
awharrison-28, line 33: Note sure what the purpose of this being a boolean is. If set to True, the code points to the error path 

> ValueError: The Azure API key cannot be `None` or empty` when not using Azure Credentials

I think it makes more sense to make it:

```suggestion
        ad_auth: Optional[Union[bool, str, TokenCredential]] = None,
```

ianchi: Hi, my reasoning is that if you are not using AD you MUST use an api_key to authenticate.
So even if the default changes to None, IF it is "None" and no key is provided, we should also change the code path to raise the same error.
There MUST be either a Key or the use of AD. 
I left it as "False" just to keep the same default that was already there. But with the appropriate check changes both are ok for me.

ianchi:
Hi @awharrison-28, I'll try to update it this week.
I had already started with the example, and I think a further change is needed.
Right now to get the config from the .env file you are forced to define an api_key variable even if you only intend to use AD auth.
Besides in the provided example ad_auth is set to True but no key is passed. This won't work.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
```diff
@@ -2,14 +2,23 @@
...
+from semantic_kernel.connectors.ai.chat_request_settings import ChatRequestSettings
+from semantic_kernel.connectors.ai.open_ai.services.azure_credentials_mixin import (
+    AzureCredentialMixin,
+)
 from semantic_kernel.connectors.ai.open_ai.services.open_ai_chat_completion import (
     OpenAIChatCompletion,
 )
 
+if TYPE_CHECKING:
+    pass
```
ianchi, line 18: I don't understand this change. 

What's the use of keeping a this IF with only a "pass"?

The idea of the conditional import was that azure credentials is an "optional" dependency, and so it is not necessarily installed in the target system, if Azure OAI with tokens is used, for instance.
Using the type annotation as string also allowed it to run without the library installed if you were not using Azure Credentials.

With this change I fear that it won't be backwards compatible, and it will be a breaking change as a new dependency is mandatory even if new feature is not used

ianchi:
I rebased this branch onto master to keep cleaner commits and included everything discussed, except this two pending conversations:

+ I kept conditionally importing azure credentials
+ I kept the default ad_auth=False

Once we finish those discussions I'll update accordingly.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
ianchi: If it is set to `True` it also needs an `api_key` as an authentication mechanism instead of az credentials, so it is correct to point to the error path.
That is the current behavior.

awharrison-28: > Hi, my reasoning is that if you are not using AD you MUST use an api_key to authenticate.

By this logic, then wouldn't it make the most sense for ad_auth have to either be false or None to use an api_key? Maybe I'm missing something here, but it looks like you need ad_auth to be a str or TokenCredential to use it for authentication, so for it to ever be 'True' doesn't make sense to me.

ianchi: I'm not exactly sure what is the use case. But that is actually the only case currently supported, and I didn't want to introduce a breaking change.
If you set ad_auth to True, you still need to send an api_key as authentication, but the api_type changes from azure to azure_ad.

Again, I don't know about the logic of that but it is the current behavior.

awharrison-28:
> I rebased this branch onto master to keep cleaner commits and included everything discussed, except this two pending conversations:
> 
> * I kept conditionally importing azure credentials
> * I kept the default ad_auth=False
> 
> Once we finish those discussions I'll update accordingly.

@ianchi after discussing with the team, we would like to combine the fields `ad_auth` and `api_key` - this would be a pretty large change, so we can leave this PR as is and create a new issue to link work under.

shawncal:
> @ianchi after discussing with the team, we would like to combine the fields `ad_auth` and `api_key` - this would be a pretty large change, so we can leave this PR as is and create a new issue to link work under.

@awharrison-28 would you recommend we complete and merge this PR, then apply the above (combine `ad_auth` and `api_key`), or we pursue that change instead of this one? If it's the latter, and you'd like to keep this change only as a reference, then please flip this to "draft" so that it's not on daily triage. And, of course, make sure to add @ianchi as a co-contributor on the new change ;)

awharrison-28:
@shawncal I've logged a task to combine ad_auth and api_key into a single authentication method - this PR will be merged with existing APIs

PR #1558, opened 2023-06-19
Copilot Chat: CI Tests

amsacha:
### Motivation and Context
Adding tests to our pipeline to catch bugs and issues earlier and have a level of sanity checking present when reasoning about issues.

### Description
Added testing for the following: 
- Server is running
- User login
- Page has correct title
- Send a message to the bot, and expect a response
- Chat History check
- Start New Chat
- SK Skill test: joke and fun fact
- Change chat title
- Document upload
- Plan generation and execution for the Klarna plugin
- Plan generation and execution for the Jira plugin 
- Plan generation and execution for the Github plugin 
- Multiuser chat: sharing and joining a chat session 

In response to: [#1775](https://github.com/microsoft/semantic-kernel/issues/1775)

### Contribution Checklist


samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,15 +1,42 @@
+/* eslint-disable testing-library/prefer-screen-queries */
```
TaoChenOSU, line 1: nit: breaking the tests into multiple files, i.e chat, plugin, multi-user, basic-ui-interaction, etc.

amsacha: I mentioned this in the description but when there are multiple simultaneous async tests running, they interfere with each other, by randomly sending messages to the chat server (all the tests share the same chat server), and this confuses the LLM. So a lot of the checks for chat history has n messages or last message was by copilot fail. Sometimes plan execution is also just abandoned cause the LLM thinks we've moved on.
The execution of all of these tests also requires knowing the exact state of the webpage, since without that locators wont work. and when tests start randomly after each other this cant be guaranteed.

TaoChenOSU: One thing we can do is to create a new chat session for each test.

samples/apps/copilot-chat-app/webapp/src/components/chat/chat-history/ChatHistoryItem.tsx
```diff
@@ -93,11 +98,17 @@ export const ChatHistoryItem: React.FC<ChatHistoryItemProps> = ({ message, getRe
         content = <ChatHistoryTextContent message={message} />;
     }
 
+    const contentAsString = (message.content as string).trim()
+                            .replace(/^sk:\/\/.*$/gm, (match: string) => createCommandLink(match))
+                            .replace(/^!sk:.*$/gm, (match: string) => createCommandLink(match));
```
dehoward, line 103: since `ChatHistoryTextContent` is already performing this logic as well, can we create a shared helper fn that is used in both?
imo it makes sense to put it and `createCommandLink` both in `utils/TextUtils.ts`.

dehoward: or better yet, could the `data-content` attribute just exist on the `div` in `ChatHistoryTextContent`?

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,239 @@
...
 import { expect, test } from '@playwright/test';
+import * as util from './testUtilCommon'
 
 /*
-This test is a simple end-to-end test that verifies the app can be
-loaded, and signed in. The user can send a message to the bot, and expect
-a response.
+This is a test suite to test the functionality of the copilot sample app.
+The Test suite goes through a bunch of tests, opening a new chat session
```
dehoward, line 7: nit: you probably don't need this piece and the sentence could just be:
```suggestion
A new chat session is opened for each test so that the chat history is not polluted and the LLM is not confused.
```

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,239 @@
...
+- Start New Chat
+- SK Skill test: joke and fun fact
+- Change chat title
+- Document upload
+- Plan generation and execution for the Klarna plugin
+- Plan generation and execution for the Jira plugin 
+- Plan generation and execution for the Github plugin 
 */
-test('get response from bot', async ({ page }) => {
+test.describe('Copilot Chat App Test Suite', () => {
```
dehoward, line 23: as the suite continues to grow larger I think we should consider having multiple test files instead a monolith one.
I could see us having a structure similar to this:
```suggestion
.
└── tests/
    ├── chat.test.ts
    ├── multiuser.test.ts
    └── plugins/
        ├── jira.test.ts
        ├── github.test.ts
        └── ...
```

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,239 @@
...
+    // Expect the last message to be the bot's response.
+    await expect(chatHistoryItems.last()).toHaveAttribute('data-username', 'Copilot');
+
+    await util.PostUnitTest(page);
+}
+
+async function ChatTitleChange(page) {
+    await util.LoginAndCreateNewChat(page);
+    
+    await page.getByRole('button', { name: 'Edit conversation name' }).click();
```
dehoward, line 93: using DOM text in our locators can make our tests brittle since the text could change at anytime.
imo we shouldn't be using hardcoded `name`s when calling `getByRole()` and should also stray away from using `getByPlaceholder()`.

using `getByTestId()` is more resilient to UX changes that may be made.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,239 @@
...
+import * as util from './testUtilCommon'
 
 /*
-This test is a simple end-to-end test that verifies the app can be
-loaded, and signed in. The user can send a message to the bot, and expect
-a response.
+This is a test suite to test the functionality of the copilot sample app.
+The Test suite goes through a bunch of tests, opening a new chat session
+for each test so that chat history is not polluted and the LLM isn't confused; 
+Here is a list of things we test right now, either standalone or as part of another test:
```
dehoward, line 9: as the suite continues to grow larger this list may become difficult to maintain and also grow long. 
instead of adding onto it each time, I think we could aim to have good readabilitity throughout the tests themselves so that this list isn't needed? what do you think?

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,239 @@
...
+- Page has correct title
+- Send a message to the bot, and expect a response
+- Chat History check
+- Start New Chat
+- SK Skill test: joke and fun fact
+- Change chat title
+- Document upload
+- Plan generation and execution for the Klarna plugin
+- Plan generation and execution for the Jira plugin 
+- Plan generation and execution for the Github plugin
```
dehoward, line 21: as the suite continues to grow larger this list may become difficult to maintain.
instead of adding onto it each time, I think we could aim to have good readability throughout the tests themselves so that this list isn't needed? what do you think?

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,239 @@
...
+            await GithubTest(page) });
+
+        // Multi-User Chat
+        test('Multi-user chat', async ({ page }) => { 
+            test.setTimeout(util.TestTimeout);
+            await MultiUserTest(page) });
+    });
+});
+
+async function ServerHealth( page ) {
```
dehoward, line 65: nit: `lowerCamelCase` is the convention for TS functions

samples/apps/copilot-chat-app/webapp/tests/testUtilCommon.ts
```diff
@@ -0,0 +1,155 @@
+import { expect } from '@playwright/test';
```
dehoward, line 1: nit: file naming, `uitls.ts` for brevity

dehoward (changes requested):
yay for more tests! 🥳

samples/apps/copilot-chat-app/webapp/src/components/chat/chat-history/ChatHistoryItem.tsx
amsacha: Moved the functions into the text util.
I will leave the data-content on the div its currently on, since we already find and query it for things and it is the container for the ChatHistoryTextContent.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
amsacha: this is a good idea, Im going to keep all the planner tests together though instead of in separate files.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
amsacha: Switched over to getByTestId() where possible.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,50 @@
-import { expect, test } from '@playwright/test';
+/* eslint-disable testing-library/prefer-screen-queries */
+import { test } from '@playwright/test';
+import * as util from './utils'
+import * as simpletests from './testsBasic'
+import * as plannertests from './testsPlanner'
+import { multiUserTest } from './testsMultiuser'
```
dehoward, line 6: nit: a little redundant
```suggestion
test suite for the functionality
```

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,50 @@
...
-/*
-This test is a simple end-to-end test that verifies the app can be
-loaded, and signed in. The user can send a message to the bot, and expect
-a response.
-*/
-test('get response from bot', async ({ page }) => {
-    // Make sure the server is running.
-    await page.goto('https://localhost:40443/healthz');
-    expect(page.getByText('Healthy')).toBeDefined();
+test.describe('Copilot Chat App Test Suite', () => {
```
dehoward, line 8: nit: I think we could shorten this to:
```suggestion
A new chat session is opened for each test so that the chat history is not polluted and the LLM is not confused.
```

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,50 @@
...
-    await page.getByPlaceholder('Password').fill(process.env.REACT_APP_TEST_USER_PASSWORD as string);
-    await page.getByRole('button', { name: 'Sign in' }).click();
+    test.describe('A, runs in parallel with B', () => {
+        test.describe.configure({ mode: 'parallel' });
+        // Server Tests
+        test('Server Health', async ({ page }) => { await simpletests.serverHealth(page) });
+        // Basic Operations
+        test('Basic Bot Responses', async ({ page }) => { 
+            test.setTimeout(util.TestTimeout);
+            await simpletests.basicBotResponses(page) });
```
dehoward, line 19: why not the MS Graph plugin too?

amsacha: MS Graph connects to more external Microsoft Apps, like Calendar and email and todo, and to properly validate the tests we would need to do more work to access the APIs better. It was a lot of work and out of scope.
Plus the Core Kernel does a lot of testing for the MS Graph Skill already;

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
amsacha: The change you're suggesting makes the comment one big line instead of 2 short ones; I think the multiple line approach is more readable (no scrolling or needing to have word wrap enabled in the IDE). We also follow this approach else where in the code.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
amsacha: renamed the test to be a bit more specific

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,53 @@
...
-    // Clicking the login button should redirect to the login page.
-    await expect(page).toHaveURL(new RegExp('^' + process.env.REACT_APP_AAD_AUTHORITY));
-    // Login with the test user.
-    await page.getByPlaceholder('Email, phone, or Skype').click();
-    await page.getByPlaceholder('Email, phone, or Skype').fill(process.env.REACT_APP_TEST_USER_ACCOUNT as string);
-    await page.getByRole('button', { name: 'Next' }).click();
-    await page.getByPlaceholder('Password').click();
-    await page.getByPlaceholder('Password').fill(process.env.REACT_APP_TEST_USER_PASSWORD as string);
-    await page.getByRole('button', { name: 'Sign in' }).click();
+    test.describe('A, runs in parallel with B', () => {
```
dehoward, line 12: in my comment on splitting up the files I assumed we'd break the `describe()` blocks too. for example:
```suggestion
    // basic.test.ts

    test.describe('Basic tests', () => {
        test('Basic Bot Responses, ...)
        // ... other basic tests
    });
```

and then the same structure for the other files.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
```diff
@@ -1,51 +1,53 @@
...
+        test('Planner Test: Klarna', async ({ page }) => { 
+            test.setTimeout(util.TestTimeout);
+            await plannertests.klarnaTest(page) });
+        test('Planner Test: Jira', async ({ page }) => { 
+            test.setTimeout(util.TestTimeout);
+            await plannertests.jiraTest(page) });
 
-    // After login, the page should redirect back to the app.
-    await expect(page).toHaveTitle('Copilot Chat');
+        // Todo: Action Planner intermittently returns a 400 error
```
dehoward, line 42: do we have already have a bug tracking this?

amsacha: We should. @teresaqhoang should be on it. I'll link it here when I find it.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
amsacha: Why would you not want to parallelize tests?
This is meant to run after each pull request is merged into main (cant do it prior because it relies on secrets). We want the iteration time on these things to be as fast as possible.

samples/apps/copilot-chat-app/webapp/tests/chat.test.ts
amsacha: https://github.com/microsoft/semantic-kernel/issues/1726

dehoward (approved):

PR #1612, opened 2023-06-20
Copilot Chat: Reduce calls to MSAL getIdToken

TaoChenOSU:
### Motivation and Context

Many Copilot Chat app React components call getActiveAccount() to get the logged in user id/name as one of the setup steps. This will trigger MSAL to retrieve ID tokens from the cache, resulting in unnecessary calls to getIdToken as shown below.

24cfecf0-4896-4561-98ba-3b65a9947c5a

### Description

Cache the logged in user id/name in the Redux store so that components don't have to repeatedly call getActiveAccount(). 

### Contribution Checklist


samples/apps/copilot-chat-app/webapp/src/App.tsx
```diff
@@ -69,26 +70,38 @@ const App: FC = () => {
...
-            dispatch(setLoggedInUserId(account.homeAccountId));
-
-            if (appState === AppState.LoadingChats) {
-                // Load all chats from the backend.
-                async function loadChats() {
-                    if (await chat.loadChats()) {
-                        setAppState(AppState.Chat);
-                    }
+        if (isAuthenticated) {
+            if (loggedInUserInfo === undefined) {
```
amsacha, line 80: Lets add a timestamp to the loggedInUserInfo object. If that timestamp is more than x mins old, say 30 mins, then we also try and fetch the activeAccount.

amsacha (changes requested):

samples/apps/copilot-chat-app/webapp/src/App.tsx
TaoChenOSU: Why do we need to fetch the account information if it hasn't changed? It will only change if the user signs out.

samples/apps/copilot-chat-app/webapp/src/redux/features/conversations/ConversationsState.ts
```diff
@@ -10,13 +10,19 @@ export type Conversations = {
 export interface ConversationsState {
     conversations: Conversations;
     selectedId: string;
-    loggedInUserId: string;
+    loggedInUserInfo?: LoggedInUserInfo;
```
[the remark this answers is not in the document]
TaoChenOSU: Moved

samples/apps/copilot-chat-app/webapp/src/App.tsx
```diff
@@ -69,26 +69,38 @@ const App: FC = () => {
...
+                    dispatch(addAlert({ type: AlertType.Error, message: 'Unable to get active logged in account.' }));
+                } else {
+                    dispatch(setLoggedInUserInfo({
+                        id: account.homeAccountId,
+                        email: account.username,
+                        fullName: account.name ?? account.username,
+                    }));
                 }
+            } else {
+                if (appState === AppState.LoadingChats) {
```
dehoward, line 91: nit: `else if`

dehoward (approved):

samples/apps/copilot-chat-app/webapp/src/redux/features/app/AppState.ts
```diff
@@ -3,6 +3,13 @@ import { AlertType } from '../../../libs/models/AlertType';
 // Copyright (c) Microsoft. All rights reserved.
 export interface AppState {
     alerts?: Alerts;
+    loggedInUserInfo?: LoggedInUserInfo;
+}
+
+export interface LoggedInUserInfo {
+    id: string;
+    email: string;
+    fullName: string;
```
teresaqhoang, line 12: nit: two rename suggestions:
- `LoggedInUserInfo` to `ActiveUserInfo` since MSAL supports multiple logged in users at one time
- `fullName` to `username` since it can fallback onto email when user's actual name is available

samples/apps/copilot-chat-app/webapp/src/App.tsx
teresaqhoang: the if statement should be embedded in an else {} block because we only want to load the chats if the LoggedInUserInfo has been defined since we have dependencies on the user id

teresaqhoang (approved):

samples/apps/copilot-chat-app/webapp/src/App.tsx
dehoward: since the `if` statement is the only logic in the `else` block, the `else if` will be equivalent

samples/apps/copilot-chat-app/webapp/src/App.tsx
```diff
@@ -69,26 +69,38 @@ const App: FC = () => {
...
-            dispatch(setLoggedInUserId(account.homeAccountId));
-
-            if (appState === AppState.LoadingChats) {
-                // Load all chats from the backend.
-                async function loadChats() {
-                    if (await chat.loadChats()) {
-                        setAppState(AppState.Chat);
-                    }
+        if (isAuthenticated) {
+            if (loggedInUserInfo === undefined) {
```
dehoward, line 79: nit: `!loggedInUserInfo`

samples/apps/copilot-chat-app/webapp/src/App.tsx
TaoChenOSU: Actually we should get rid of the else block because we want to carry on loading the chats after the user info is set.

amsacha (dismissed):

samples/apps/copilot-chat-app/webapp/src/App.tsx
amsacha: The thought was more from a security perspective and logging out stale accounts. In retrospect its probably something someone adapting the sample app should look at based on their use case.

teresaqhoang (dismissed):

dehoward (approved):

amsacha (approved):

samples/apps/copilot-chat-app/webapp/src/App.tsx
```diff
@@ -69,16 +69,29 @@ const App: FC = () => {
...
+    const { activeUserInfo } = useAppSelector((state: RootState) => state.app);
     const isAuthenticated = useIsAuthenticated();
 
     const chat = useChat();
 
     useEffect(() => {
-        if (isAuthenticated && account) {
-            dispatch(setLoggedInUserId(account.homeAccountId));
+        if (isAuthenticated) {
+            let isActiveUserInfoSet = activeUserInfo !== undefined;
```
dehoward, line 79: should we add `activeUserInfo` to the dependency array for the hook?

dehoward (approved):

samples/apps/copilot-chat-app/webapp/src/App.tsx
teresaqhoang: I don't think that's necessary because isActiveUserInfoSet is set to true, so the next if block should run / chats should be loaded as soon as the activeUser is set in app state.

teresaqhoang (dismissed):

samples/apps/copilot-chat-app/webapp/src/App.tsx
dehoward: it's good practice to add it unless we have a specific reason to exclude it. I can push this change here in a sec and will make sure it doesn't cause any unnecessary renders.
ideally, we can move away from `// eslint-disable-next-line react-hooks/exhaustive-deps` lines except for in rare cases where unavoidable.

teresaqhoang: @dehoward I tested it and it seems like it introduces more calls to the token cache? Can you test it on a production build and see if you're able to repro?
image

dehoward: I just noticed this too, looks like there are a couple of more calls. let's not block the PR for this. 

I will go through soon and look to remove all of the `eslint-disable-next-line` lines and make sure we don't add any new calls, as usually it will reduce them when done correctly.

dehoward:
resolving conflicts

dehoward (approved):

amsacha (approved):

Issue #1483, opened 2023-06-14
Sequential Planner: Plan parsing failed: Invalid plan: Failed to parse plan xml.


mtechvance:
**Describe the bug**
When using the sequential planner with an open api plugin, the payload attribute value in the plan xml is not properly escaped. It uses '/' escapes instead of the xml accepted &quot; escape. I overrode the sequential planner default prompt and added one additional line to it to have it properly escape the quotes.

'2. Quotes inside of xml attributes should be escaped with '&quot;'.'

**To Reproduce**
1. Import an OpenAPI plugin that has a post request with a payload
2. Create a SequentialPlanner
3. Create a plan with a goal that will call the open api plugin function that has a payload
4. Error should occur on Run or Invoke

**Expected behavior**
Plan parsing succeeds

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - IDE: Visual Studio 2022
 - 0.15.230609.2-preview

**Additional context**

- I started with the OpenAPI doc found [here](https://github.com/MicrosoftDocs/vsts-rest-api-specs/blob/master/specification/search/7.1/search.json) for Dev Ops search 
- An example of valid function XML would be this
`<plan>
    <function.CodeSearch.codesearchresults_post organization="MyOrg"
                                                project="Proj"
                                                api_version="7.1-preview.1"
                                                server_url="https://almsearch.dev.azure.com/"
                                                payload="{&quot;searchText&quot;:&quot;test&quot;,&quot;$top&quot;:3,&quot;filters&quot;:{&quot;Repository/Project&quot;:[&quot;Proj&quot;],&quot;Repository/Repository&quot;:[&quot;Repo&quot;]}}"
                                                content_type="application/json"
                                                appendToResult="RESULT__TOP_THREE_RESULTS" />
</plan>`
- Truncated call stack
Plan parsing failed: Invalid plan: Failed to parse plan xml.[2023-06-14T16:14:49.628Z] Executed 'Functions.PerformCodeSearch' (Failed, Id=c1362f61-e771-4075-971d-6c39adf551b4, Duration=12217ms)
[2023-06-14T16:15:38.997Z] System.Private.CoreLib: Exception while executing function: Functions.PerformCodeSearch. System.Private.CoreLib: Result: Failure
Exception: System.AggregateException: One or more errors occurred. (Invalid plan: Plan parsing error, invalid XML)
[...]

PR #1615, opened 2023-06-20
Improve sequential plan parser with support for missing functions.

### LINKED ISSUE #159: Improve Planner prompt to support invalid ASK

lemillermicrosoft:
### Motivation and Context
This pull request improves the ability to create sequential plans by parsing the XML plan text and matching the function names and parameters with the registered plugin functions. It also improves the error handling and logging in the sequential planner by adding more specific error codes, messages, and exceptions for different scenarios. The main motivation for these changes is to make the plan creation more robust and flexible, and to avoid throwing exceptions when the plan XML is not well-formed or contains unknown functions. The new option, `AllowMissingFunctions`, can be set to true or false in the SequentialPlannerConfig class, and it determines whether the plan creation will fail or succeed when a function is missing in the skill library. The plan parsing logic has been improved to use a regular expression to extract the <plan> element from the XML string, and to handle any XmlException that may occur when loading the XML document.

Related to #1483 
Fixes #159

### Description
- Added a new method `CreateKernelAndFunctionCreateMocks` in `SequentialPlanParserTests` to create a mock kernel and register mock plugin functions for testing
- Added three new test cases in `SequentialPlanParserTests` to test the parsing of plans with OpenApi plugin functions, with different input formats and parameters
- Modified `SequentialPlanParser` to handle the case where the plan text is not valid XML by using a regular expression to try to extract the plan XML from the input string and throwing a `PlanningException` if it fails
- Added a new error code `CreatePlanError` for cases where the plan creation fails due to invalid goal or plan XML
- Added a new option `AllowMissingFunctions` to the `ToPlanFromXml` method, which allows the parser to append missing function nodes as plain text steps instead of throwing an exception
- Updated the `skprompt.txt` file with a new instruction to return an empty plan if a plan cannot be created with the available functions
- Fixed a bug in the `SequentialPlanParser` where an exception was thrown twice when the plan XML was invalid
- Added more specific error messages and exceptions for different scenarios, such as invalid configuration, unknown error, or missing function
- Improved the logging of the plan parsing and processing by adding more trace and error messages

### Contribution Checklist

### DIFF
--- a/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
+++ b/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
@@ -58,7 +58,24 @@ internal static class SequentialPlanParser
             }
             catch (XmlException e)
             {
-                throw new PlanningException(PlanningException.ErrorCodes.InvalidPlan, "Failed to parse plan xml.", e);
+                // xmlString wasn't valid xml, let's try and parse it out.
+                // From my python code, for regex for getting json.
+                //                 # Filter out good JSON from the result in case additional text is present
+                // json_regex = r"\{(?:[^{}]|(?R))*\}"
+                // I want similar, but for xml that starts with <plan> and ends with </plan>
+                // Regex planRegex = new(@"\<plan\>(?:[^\<\>]|(?R))*\<\/plan\>", RegexOptions.Singleline);
+                // Match match = planRegex.Match(xmlString);
+                // if (match.Success)
+                // {
+                //     string planXml = match.Value;
+                //     xmlDoc.LoadXml("<xml>" + planXml + "</xml>");
+                // }
+                // else
+                // {
+                //     throw new PlanningException(PlanningException.ErrorCodes.InvalidPlan, $"Failed to parse plan xml: {xmlString}", e);
+                // }
+
+                throw new PlanningException(PlanningException.ErrorCodes.InvalidPlan, $"Failed to parse plan xml: {xmlString}", e);
             }
 
             // Get the Solution
--- a/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
+++ b/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
@@ -166,7 +166,7 @@ internal static class SequentialPlanParser
                             if (allowMissingFunctions)
                             {
                                 context.Log.LogTrace("{0}: allowing missing function node {1}", parentNodeName, skillFunctionName);
-                                plan.AddSteps(new Plan(childNode.InnerText));
+                                plan.AddSteps(new Plan(skillFunctionName));
                             }
                             else
                             {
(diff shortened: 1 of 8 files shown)


dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
```diff
@@ -135,23 +163,35 @@ internal static Plan ToPlanFromXml(this string xmlString, string goal, SKContext
...
-                            context.Log.LogTrace("{0}: appending function node {1}", parentNodeName, skillFunctionName);
-                            plan.AddSteps(new Plan(childNode.InnerText));
+                            if (allowMissingFunctions)
+                            {
+                                context.Log.LogTrace("{0}: allowing missing function node {1}", parentNodeName, skillFunctionName);
+                                plan.AddSteps(new Plan(childNode.InnerText));
+                            }
+                            else
+                            {
+                                throw new PlanningException(PlanningException.ErrorCodes.InvalidPlan, $"Failed to find function '{skillFunctionName}' in skill '{skillName}'.");
```
teresaqhoang, line 173: Other than the planner pulling functions from the [EXAMPLES] section of the prompt, what other cases can cause missing functions? 

I see the benefit in allowing missing functions to call out possible helpful capabilities that a developer might've missed, but if `allowMissingFunctions == false`, there should be some kind of safeguard against the above scenario, otherwise, it's not really an actionable error on the developer's part as much as it is a hallucination

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
```diff
@@ -135,23 +163,35 @@ internal static Plan ToPlanFromXml(this string xmlString, string goal, SKContext
                         }
                         else
                         {
-                            context.Log.LogTrace("{0}: appending function node {1}", parentNodeName, skillFunctionName);
-                            plan.AddSteps(new Plan(childNode.InnerText));
+                            if (allowMissingFunctions)
+                            {
+                                context.Log.LogTrace("{0}: allowing missing function node {1}", parentNodeName, skillFunctionName);
+                                plan.AddSteps(new Plan(childNode.InnerText));
```
teresaqhoang, line 169: from CC testing, the InnerText usually is empty so there's no telling what function this was unless you look at the XML. Can you use childNode.name or skillFunctionName here instead?

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
lemillermicrosoft: >  it's not really an actionable error on the developer's part as much as it is a hallucination

Depending on the model, it can hallucinate functions that are not present in the [EXAMPLES] and also not part of [AVAILABLE FUNCTIONS].

This will be caught and returned to the developer/user as a CreatePlanError indicating it's unable to create a plan.

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
lemillermicrosoft: Good call. Updating.

shawncal (approved):

PR #1616, opened 2023-06-20
Moving KernelSyntaxExamples into the dotnet/ folder

shawncal:
"KernelSyntaxExamples" is the only outstanding project under /samples that still uses project references to access the kernel libraries.

This change consolidates all the building, linked dotnet SDK code in SK-dotnet.sln into a single top-level folder. This will simplify build jobs, automatic code owner assignment, and other dev ops tasks.

Coming next: other projects in samples will get removed from the dotnet/SK-dotnet.sln solution, and a new solution with dedicated builds created for the /samples folder.

TBD:
- Language-specific code "samples" will move into a similar subfolder for each language.
- The current "/samples" folder (at the root) will likely become something like "apps", with the demo apps promoted to the top-level folder for better discoverability. 
- notebooks may also move to this folder in the future.


lemillermicrosoft (approved):

shawncal:
Updating docs now...

PR #1678, opened 2023-06-22
Update sequential planner syntax and documentation

lemillermicrosoft:
This pull request adds an END comment to the sequential planner XML format, to mark the end of the plan after the closing </plan> tag. This makes the format more consistent and easier to parse. It also simplifies the instructions for creating a plan, by removing some redundant or outdated information and clarifying some points. The examples and available functions are removed accordingly.

### Description
Details:
- Add an END comment to the sequential planner XML format, to mark the end of the plan after the closing </plan> tag. This is done by changing the StopSequence constant in SequentialPlanner.cs and updating the skprompt.txt file.
- Simplify the instructions for creating a plan, by removing some redundant or outdated information and clarifying some points. Specifically, the following changes are made:
  - Remove the requirement to check that a function is present in the most recent available functions list, as this is implied by the fact that the list is updated for each goal.
  - Remove the statement that a function has a single input and output, as some functions may have additional parameters as defined in the available functions list.
  - Remove the statement that input does not need to be specified if it consumes the output of the previous function, as this is not always the case and may cause confusion.
  - Change the syntax for saving an output to a context variable or a result, by using "=" instead of ":" and removing the "/" at the end of the function tag.
  - Clarify that the END comment should be appended after the final closing </plan> tag, not inside it.
- Update the examples and available functions in skprompt.txt to reflect the changes in the format and the instructions.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SequentialPlanParserTests.cs
+++ b/dotnet/src/Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SequentialPlanParserTests.cs
@@ -198,11 +198,9 @@ public class SequentialPlanParserTests
         // Assert
         Assert.NotNull(plan);
         Assert.Equal(goalText, plan.Description);
-        Assert.Equal(2, plan.Steps.Count);
+        Assert.Equal(1, plan.Steps.Count);
         Assert.Equal("MockSkill", plan.Steps[0].SkillName);
         Assert.Equal("Echo", plan.Steps[0].Name);
-        Assert.Equal("This is some text", plan.Steps[1].Description);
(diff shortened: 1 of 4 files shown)


lemillermicrosoft:
A follow up to #1615 that adds to the improvements.

teresaqhoang (approved):

lemillermicrosoft:
Tabling this for the moment. Really want to rid the examples in the prompt and focus on defining how to properly call a function. Struggling to consistently follow rules though to 1) not try and index a variable like it's an object `$SYPNOPSIS[2]` and 2) XML escape the parameters it generates.

lemillermicrosoft:
Condensed changes into #1727 1727

Issue #1606, opened 2023-06-20
Python: Integrate chat stream into kernel


joowon-dm-snu:
I checked 2 PRs(#1115, #1316) for streaming. Both PRs are defining `kernel = sk.Kernel()` but not using them at all.
Current chat streaming is complex to use with `Kernel`. 

hope streaming can be used via kernel like below
```python
kernel = sk.Kernel()
kernel.add_chat_service(
        "gpt-3.5-turbo",
        sk_oai.AzureChatCompletion(
            "gpt-35-turbo",
            os.environ.get("AZURE_OPENAI_ENDPOINT"),
            os.environ.get("AZURE_OPENAI_KEY"),
        ),
    )
skills = kernel.import_semantic_skill_from_directory(skill_dir, skill)
kernel.run_stream_async(skill["TranslationSkill"]["toEnglish"], input_vars=...)
```

alexchaomander:
Thanks for the feedback @joowon-dm-snu! It's good feedback that we can improve the interface. Putting this on the backlog for @awharrison-28 and @mkarle to review when they are back from vacation!

joowon-dm-snu:
Thank you. If it's not a high priority, I'll open a PR myself. I want to use it as soon as possible :).

Currently, I'm using a temporary method where I can simply append the code into `kernel.run_stream_async` with minimal additional code to implement.
```python
skills = kernel.import_semantic_skill_from_directory(skill_dir, skill)
func = skills["TranslationSkill"]["toEnglish"]

# Get the closure variables as function's config does not saved in `SKFunctionBase` itself.
closure_vars = func._function.__closure__  
for var in closure_vars:
    if isinstance(var.cell_contents, SemanticFunctionConfig):
        function_config = var.cell_contents
        break
request_settings = func._chat_request_settings
client = func._ai_service

prompt = await function_config.prompt_template.render_async(context)

...
```
If your team want this to be more accurate, it seems like `SKFunctionBase` should add new function, `invoke_stream_async`. But this might take a bit longer than former one, so if it's okay, I can quickly open the former PR and gradually add the necessary functions to SKFunctionBase. 

I'm curious what you think about that.

alexchaomander:
I personally think it makes sense and would be a nice interface to add. 

@dluc what do you think?

joowon-dm-snu:
I opened PR with my temporal method!

PR #1691, opened 2023-06-23
Python: temporal kernel run stream async

joowon-dm-snu:
### Motivation and Context

Temporal solution for #1606 

### Contribution Checklist


python/semantic_kernel/kernel.py
```diff
@@ -133,6 +134,97 @@ def register_semantic_function(
 
         return function
 
+    async def run_stream_async(
+        self,
+        func: SKFunction,
```
awharrison-28, line 139: Is this intentionally written to only run a single SK function at a time? It might be a bit unclear to users that you cannot run a collection of functions with this method as opposed to it's `run_async` non-streaming equivalent. 

I think this should have the same function signature as `run_async` below. 

There are of course some decisions to make here. If this method is changed to consume _n_ functions, should only the last one be streamed, or should all?

@shawncal

awharrison-28 (changes requested):

python/semantic_kernel/kernel.py
joowon-dm-snu: i understand and agree with this should have same function signature as `run_async`
I  changed function signature. 

and originally i wanted to add `SKFunction.invoke_stream_async`. 
If you think this good idea, i can try it too

awharrison-28: I like the idea of invoke_stream_async on SKFunction. I would just add it to sk_function.py for now, not sk_function_base.

awharrison-28 (dismissed):
Thank you for the contribution @joowon-dm-snu! 

I hope you don't mind, I moved the new integration tests under the azure endpoint tests for better test stability.

awharrison-28 (approved):

joowon-dm-snu:
> Thank you for the contribution @joowon-dm-snu!
> 
> I hope you don't mind, I moved the new integration tests under the azure endpoint tests for better test stability.
@awharrison-28 
no problem at all!
i just want to check adding `invoke_stream_async` on `SKFunction` is valuable and needs the other PR!

python/semantic_kernel/kernel.py
```diff
@@ -133,6 +134,110 @@ def register_semantic_function(
...
+                prompt = await function_config.prompt_template.render_async(context)
+                async for stream_message in client.complete_stream_async(
+                    prompt, stream_function._ai_request_settings
+                ):
+                    completion += stream_message
+                    yield stream_message
+                context.variables.update(completion)
+
+        except Exception as e:
+            # TODO: "critical exceptions"
```
VladTolm, line 238: It will be very helpful to rise exception here.
If you import skills from file and your config does not contain "number_of_responses": 1 you will not get anything.

The error occurs in OpenAIChatCompletion at line 141 "request_settings.number_of_responses > 1"

Error ""'>' not supported between instances of 'NoneType' and 'int'""

Issue #1633, opened 2023-06-21
Large documents cause stack overflow in TextChunker.SplitPlainTextParagraphs()


bdobabraham:
**Describe the bug**
When calling `TextChunker.SplitPlainTextLines()` followed by `TextChunker.SplitPlainTextParagraphs()` with a large document (7k+ lines output from `SplitPlainTextLines()` call.) The internal chunking logic causes a stack overflow exception in the final line of `TextChunker.BuildParagraph()`.

**To Reproduce**
1. Setup CoPilot demo
2. Find a large document with lots of text
3. Upload document to CoPilot app
4. Crash will happen during upload process

**Expected behavior**
Document should upload, chunk, and process without issue.

**Screenshots**
![image](https://github.com/microsoft/semantic-kernel/assets/135271843/66398e13-be9f-455c-ad9c-8462062f2259)

**Desktop (please complete the following information):**
 - OS: Windows
 - IDE: Visual Studio 17.6.3
 - NuGet Package Version: 0.16.230615.1-preview

**Additional context**
Document stats:
Characters: 650k+
Words: 115k+
Output from `TextChunker.SplitPlainTextLines()`: 7,214 lines

PR #1709, opened 2023-06-26
Avoid recursion in TextChunker.BuildParagraph

stephentoub:
### Motivation and Context

TextChunker.BuildParagraph is unnecessarily recursive and can lead to stack overflows on large inputs.

### Description

Fixes https://github.com/microsoft/semantic-kernel/issues/1633

Also happens to make it more efficient with a few tweaks.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/Text/TextChunker.cs
+++ b/dotnet/src/SemanticKernel/Text/TextChunker.cs
@@ -131,31 +131,34 @@ public static class TextChunker
 
     private static List<string> BuildParagraph(List<string> truncatedLines, StringBuilder paragraphBuilder, List<string> paragraphs, int maxTokensPerParagraph, Func<string, int, List<string>> longLinesSplitter)
     {
-        // Base case: no more elements in the list
-        if (truncatedLines.Count == 0)
+        while (true)
         {
-            // Adding any remaining paragraph
-            if (paragraphBuilder.Length > 0)
+            if (truncatedLines.Count == 0)
+            {
+                // Adding any remaining paragraph
+                if (paragraphBuilder.Length > 0)
+                {
+                    paragraphs.Add(paragraphBuilder.ToString().Trim());
+                }
+                return paragraphs;
+            }
+
+            string line = truncatedLines[0];
+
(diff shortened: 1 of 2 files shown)


dotnet/src/SemanticKernel/Text/TextChunker.cs
```diff
@@ -131,31 +131,34 @@ private static List<string> InternalSplitTextParagraphs(List<string> lines, int
 
     private static List<string> BuildParagraph(List<string> truncatedLines, StringBuilder paragraphBuilder, List<string> paragraphs, int maxTokensPerParagraph, Func<string, int, List<string>> longLinesSplitter)
     {
-        // Base case: no more elements in the list
-        if (truncatedLines.Count == 0)
+        while (true)
```
dmytrostruk, line 134: nit: Looks like the only condition to escape from infinite loop is `truncatedLines.Count == 0`, so probably we can replace `while (true)` with `while (truncatedLines.Count > 0)`?

Unit-tests are passed with this approach:
```csharp
while (truncatedLines.Count > 0)
{
    string line = truncatedLines[0];

    if (paragraphBuilder.Length > 0 && TokenCount(paragraphBuilder.Length) + TokenCount(line.Length) + 1 >= maxTokensPerParagraph)
    {
        paragraphs.Add(paragraphBuilder.ToString().Trim());

        // next paragraph
        paragraphBuilder.Clear();
    }
    else
    {
        paragraphBuilder.AppendLine(line);

        truncatedLines.RemoveAt(0);
    }
}

// Adding any remaining paragraph
if (paragraphBuilder.Length > 0)
{
    paragraphs.Add(paragraphBuilder.ToString().Trim());
}

return paragraphs;
```

dmytrostruk (approved):

dotnet/src/SemanticKernel/Text/TextChunker.cs
stephentoub: Yes, it's the same thing, just reorganized.  I was trying to minimize churn, but I can restructure it if that's desired.

dmytrostruk (approved):

lemillermicrosoft (approved):
Thank you!

PR #1712, opened 2023-06-26
Java: Http skill Added

yoshioterada:
Added HttpSkill for Java

### Motivation and Context

Based on the [http_skill.py](https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/core_skills/http_skill.py)

I implemented the Http Skill by using Mono of Project Reactor.
And I implemented the Test code with Mock.

### Contribution Checklist

### DIFF
--- a/java/semantickernel-core-skills/src/main/java/com/microsoft/semantickernel/coreskills/HttpSkill.java
+++ b/java/semantickernel-core-skills/src/main/java/com/microsoft/semantickernel/coreskills/HttpSkill.java
@@ -46,7 +46,7 @@ public class HttpSkill {
      */
     public Mono<String> getAsync(String url) {
         if (url == null || url.isEmpty()) {
-            throw new IllegalArgumentException("url cannot be `null` or empty");
+            return Mono.error(new IllegalArgumentException("url cannot be `null` or empty"));
         }
         HttpRequest request = new HttpRequest(HttpMethod.GET, url);
         return httpClient.send(request).flatMap(response -> response.getBodyAsString());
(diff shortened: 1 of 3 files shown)


java/semantickernel-core-skills/src/main/java/com/microsoft/semantickernel/coreskills/HttpSkill.java
```diff
@@ -0,0 +1,100 @@
...
+ * <p>{{http.putAsync $url}}
+ *
+ * <p>{{http.deleteAsync $url}}
+ */
+public class HttpSkill {
+
+    private final HttpClient httpClient;
+
+    public HttpSkill() {
+        this.httpClient = new NettyAsyncHttpClientBuilder().build();
```
brunoborges, line 34: It is perhaps better to reference HttpClient.createDefault() so that it uses the SPI mechanism of the Azure SDK to load an Http Client.

This way, we don't need to have a dependency on the Netty one, and users can load whatever they want.

https://github.com/Azure/azure-sdk-for-java/wiki/HTTP-clients/0b5cb17511eb25be4ab7b6a78c765d27748fa907#configuring-http-clients

brunoborges (changes requested):
Don't create a dependency on Netty Http Client.

java/semantickernel-core-skills/src/main/java/com/microsoft/semantickernel/coreskills/HttpSkill.java
brunoborges: Does the project still builds and the unit test for this skill passes without this dependency on test scope?

brunoborges: Excellent!

markwallace-microsoft (approved):

PR #1717, opened 2023-06-26
Update README.md

VinayAvilala:
### Contribution Checklist


shawncal:
In particular: 
- dotnet 7.0 is not a requirement (you can use dotnet 6 or 8)
- checking prerelease box isn't a requirement in command line or all IDEs, but it would be good to call out that it is a prerelease, in case people have trouble finding it. Note: we're discussing just removing the prerelease on our weekly packages.

Closing this PR as I don't think this is quite the right text, but you're right in pointing out that we can highlight some additional steps for clarity. @matthewbolanos can you please review this readme and make any changes needed?

PR #1718, opened 2023-06-26
Exception handling ADR

SergeyMenshykh:
### Motivation
The purpose of this ADR is to describe the existing challenges with SK error handling mechanism, propose solutions for improvement, and share points to consider along with questions to think about.


docs/decisions/0004-error-handling.md
```diff
@@ -0,0 +1,59 @@
...
+
+- **Improper exception usage**: Modify SK code to throw .NET standard exceptions, such as ArgumentOutOfRangeException or ArgumentNullException, when class argument values are not provided or are invalid, instead of throwing custom SK exceptions. Analyze SK exception usage to identify other potential areas where standard .NET exceptions can be used instead.
+
+- **Exception hierarchy inconsistency**: Refactor SK custom exceptions that are not derived from SKException to derive from SKException. This ensures consistency in the SK exception hierarchy.
+
+- **Missing original exception details**: Identify all cases where the original exception or failure is not preserved either as an inner exception or as a property/message of the rethrown exception, and address them. For example, AI service connectors that use HttClient should preserve the original status code by either setting it as an inner exception or including it as part of the rethrown exception's property.
```
stephentoub, line 38: > For example, AI service connectors that use HttClient should preserve the original status code by either setting it as an inner exception or including it as part of the rethrown exception's property.

Even if the original status code is copied out to whatever top-level exception gets thrown, there's still value in including the original exception as the inner exception.  In particular, any evolution of information exposed on that exception type will be available, for example if/when HttpRequestException is augmented with an HttpRequestError property as in https://github.com/dotnet/runtime/issues/76644#issue-1396942718.

docs/decisions/0004-error-handling.md
```diff
@@ -0,0 +1,59 @@
...
+
+- **Other**:
+  - Remove any code that wraps unhandled exceptions into AIException or any other SK exception solely for the purpose of wrapping. In most cases, this code does not provide useful information to help SK client code handle the exception, apart from a generic and uninformative "Something went wrong" message. Regardless of whether the unhandled exception is wrapped or not, SK client code cannot handle or recover from it without prior knowledge of the exception and proper exception handling logic.
+
+## Questions and consideration
+
+- Should the exception hierarchy follow the SK components hierarchy? For example, should there be a PlanningException for planners and a MemoryStorageException for memory storage connectors? What are the benefits for SK client code? Can the SK client code operate on the exception?
```
stephentoub, line 50: This comes down to what does the consuming code need to do.  What are the scenarios where consuming code needs to differentiate an exception related to planning from an exception related to memory storage? e.g. what are the scenarios where code would do:
```C#
try
{
   ...
}
catch (PlanningException pe)
{
     ... // do one thing
}
catch (MemoryStorageException mse)
{
    ... // do something else
}
```
If the consuming code would benefit from doing something different programmatically based on the exception type, differentiating can be valuable.  If not, differentiating adds little value and only cost.  One of my most costly mistakes for the .NET ecosystem was introducing a particular exception type when I shouldn't have.

docs/decisions/0004-error-handling.md
```diff
@@ -0,0 +1,59 @@
...
+  On the other hand, having base exception types per component or set of related operations can be convenient from a code reusability perspective because derived types can inherit common component exception properties relevant to that component.
+- Should all custom SK exceptions be derived from the SKException type?  Currently, SKException derives from the Exception type and does not add any additional information. What value can SK client code get from catching these exceptions compared to catching them as Exception types? Should SKException be used in all cases when SK needs to indicate that something went wrong without providing details through strongly typed properties, and neither custom nor standard exception types exist that could be used for that purpose?
+- Should we keep the IsCriticalException functionality that SK currently has to avoid handling several system/critical exceptions such as OutOfMemoryException and StackOverflowException? Some of them cannot be intercepted.
```
stephentoub, line 54: StackOverflowException isn't actually thrown; if you stack overflow, you crash, and the stack isn't unwound, e.g. your catch / finally blocks don't execute.  There's no benefit in trying to special-case it.

And OutOfMemoryExceptions don't necessarily mean you can't run recovery code.  The failure might be due, for example, to code trying to allocate an exorbitant amount of memory even though a smaller allocation might have succeeded.

Especially with moving to a model where exceptions aren't caught and stored, I don't think it's valuable to try to special-case certain exception types.

docs/decisions/0004-error-handling.md
```diff
@@ -0,0 +1,59 @@
...
+
+- **Exception hierarchy inconsistency**: Refactor SK custom exceptions that are not derived from SKException to derive from SKException. This ensures consistency in the SK exception hierarchy.
+
+- **Missing original exception details**: Identify all cases where the original exception or failure is not preserved either as an inner exception or as a property/message of the rethrown exception, and address them. For example, AI service connectors that use HttClient should preserve the original status code by either setting it as an inner exception or including it as part of the rethrown exception's property.
+
+- **No consolidated way for exception handling**:
+  - [Memory connectors]: Introduce a new exception called VectorDbStorageException, along with derived exceptions like UpsertRecordException and RemoveRecordException, if additional details need to be captured and propagated to SK client code. Refactor all SK memory component implementations to use these exceptions instead of implementation-specific memory storage exceptions.
```
dmytrostruk, line 41: I'm not sure about `VectorDbStorageException` naming. Since we have `IMemoryStore` interface and different implementations (e.g. `ChromaMemoryStore`, `RedisMemoryStore`, I would probably name the exception as `MemoryStoreException`.

Regarding `UpsertRecordException` and `RemoveRecordException`: currently it's possible to use only one memory store per Kernel instance, but in the future, it will be possible to register multiple memory stores. If I register both Chroma and Redis memory stores, when my Kernel instance throws `UpsertRecordException`, how should I know which memory provider throws it? There might be a case, when I want to handle the same operation failure (like upserting record) differently for different memory providers (for example, for MemoryStore1 I want to log error and re-throw, while for MemoryStore2 I want to trace it and continue, since it's expected behavior).

docs/decisions/0004-error-handling.md
```diff
@@ -0,0 +1,59 @@
...
+  - [AI service connectors]: Restrict the usage of AIException to AI service connectors only and replace its usage in other SK components with more relevant exceptions. For cases where AIException is thrown, use HttpOperationException describved below as an inner exception to allow SK client code to access the original status code and exception/error details, if necessary.
```
dmytrostruk, line 42: Regarding multiple memory stores issue, same applies for AI service connectors. Since it's possible today to register multiple AI services, as a result, we expect only `AIException` exception, but it's not clear how to handle each AI service exception separately. Also, I would probably rename `AIException` to `AIServiceException`.

dmytrostruk, line 42:
```suggestion
  - [AI service connectors]: Restrict the usage of AIException to AI service connectors only and replace its usage in other SK components with more relevant exceptions. For cases where AIException is thrown, use HttpOperationException described below as an inner exception to allow SK client code to access the original status code and exception/error details, if necessary.
```

docs/decisions/0004-error-handling.md
```diff
@@ -0,0 +1,59 @@
...
+- Should all custom SK exceptions be derived from the SKException type?  Currently, SKException derives from the Exception type and does not add any additional information. What value can SK client code get from catching these exceptions compared to catching them as Exception types? Should SKException be used in all cases when SK needs to indicate that something went wrong without providing details through strongly typed properties, and neither custom nor standard exception types exist that could be used for that purpose?
+- Should we keep the IsCriticalException functionality that SK currently has to avoid handling several system/critical exceptions such as OutOfMemoryException and StackOverflowException? Some of them cannot be intercepted.
+- An idea has been proposed to add SK wide error codes to exceptions in order to identify and locate the source of the exception. The question is, is it necessary? Aren't the exception type, message, values of its properties, and stack trace enough to precisely identify the exception and locate its source?
```
dmytrostruk, line 55: I think it depends how granular our exception hierarchy will be. Error codes are usually short, and they are useful when you try to search specific error in your logs (although you can still search by exception type), or if you try to find specific problem in documentation (again, still searchable by exception type).

The other point is - error codes will allow to align exception strategy across different programming languages (C#, Python, Java and more in the future). While exception hierarchy can be designed with inheritance in those languages, the naming convention for exceptions might be different (for example `MemoryStoreException` in C# and `MemoryStoreError` in Python). At the same time, error code like `SK001` can point to the same problem in all languages.

Do we actually need to align exceptions across different languages? I believe that is separate topic to discuss :)

docs/decisions/0004-error-handling.md
SergeyMenshykh: Be honest, at the moment, I don't see any scenario where consuming code needs to differentiate exceptions between different SK components.

docs/decisions/0004-error-handling.md
SergeyMenshykh: Yep, good point, updated the ADR.

docs/decisions/0004-error-handling.md
SergeyMenshykh: MemoryStoreException is definitely a better name for the exception than VectorDbStorageException one. I updated the document to reflect that.

Regarding your second point on handling exceptions from different memory storages differently - Kernel is not the appropriate place for such logic. If Kernel starts distinguishing between memory storages, it will inevitably become coupled to them by referencing the corresponding memory storage specific exception and memory storage projects implicitly.

One possible solution that would enable different handling of upserts elegantly is to use the decorator pattern. The decorator would wrap a specific memory storage and add the required behavior. Since the decorator would implement the same interface as the memory storage, it could seamlessly replace the wrapped memory storage. The Kernel would interact with it without realizing that it is a decorator and not a memory storage.

This approach would prevent the need to modify the memory storage class and avoid cluttering Kernel with memory storage specifics.

Considering that the decorator is designed for a specific memory storage, there is no need to have exceptions that are specific to a particular memory storage implementation. The decorator itself provides enough context for all MemoryStoreException exceptions thrown by the wrapped memory storage.

dmytrostruk: > If Kernel starts distinguishing between memory storages, it will inevitably become coupled to them by referencing the corresponding memory storage specific exception and memory storage projects implicitly.

This is not going to be the case. Kernel only works with `IMemoryStore` interface, Each memory store implementation will be responsible for throwing an exceptions, and Kernel will be responsible to pass it through to the caller. For example, if I have two (potentially in the future) registered memory providers - `ChromaMemoryStore` and `RedisMemoryStore`, then on `kernel.Memory.SaveInformationAsync` in case of exception it will throw `ChromaMemoryStoreException` or `RedisMemoryStoreException` instead of `UpsertRecordException`. I believe there is no need in decorator.

SergeyMenshykh: Do you see any practical scenarios for the caller of the kernel.Memory.SaveInformationAsync or any other public Kernel method to distinguish between ChromaMemoryStoreException and RedisMemoryStoreException?

docs/decisions/0004-error-handling.md
SergeyMenshykh: As usual, **well-thought-through** scenarios would help here.

docs/decisions/0004-error-handling.md
SergeyMenshykh: Cross-language alignment may seem good, but it can come at a cost of unnecessary coordination and maintenance, slowing down the process and teams involved.

docs/decisions/0004-error-handling.md
dmytrostruk: This can be useful only if we provide a support for multiple memory stores in one Kernel instance (I think there is a plan to add that possibility in the future). In this case I'm not sure how to react to `UpsertRecordException` and what memory provider should be checked.

But even for our current implementation with one memory provider, I'm not sure `UpsertRecordException` will provide some benefits. I think base `MemoryStoreException` with exception message should be enough.

docs/decisions/0004-error-handling.md
SergeyMenshykh: Yep, removed this question and added a step to remove those two exceptions.

shawncal (approved):

Issue #1730, opened 2023-06-27
Python: Core skills include incorrect syntax in usage documentation


HenriSchulte-MS:
Some core skills include example code showing how to import the skill. Some of these examples, use an invalid syntax where arguments are in the reversed order, e.g. `kernel.import_skill("wait", WaitSkill());` instead of the valid `kernel.import_skill(WaitSkill(), "wait")`.

The following core skills include the incorrect syntax:

- https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/core_skills/file_io_skill.py
- https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/core_skills/math_skill.py
- https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/core_skills/text_skill.py
- https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/core_skills/time_skill.py
- https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/core_skills/wait_skill.py

alexchaomander:
Thanks for pointing these out! We'll fix them!

HenriSchulte-MS:
Was fixed in #1731

PR #1731, opened 2023-06-27
Python: Correct syntax in core skill examples

HenriSchulte-MS:
Fix for #1730 
Some core skills include example code showing how to import the skill. Some of these examples, used an invalid syntax where arguments were in the reversed order, e.g. kernel.import_skill("wait", WaitSkill()); instead of the valid kernel.import_skill(WaitSkill(), "wait").

### DIFF
--- a/python/semantic_kernel/core_skills/file_io_skill.py
+++ b/python/semantic_kernel/core_skills/file_io_skill.py
@@ -11,7 +11,7 @@ class FileIOSkill:
     Description: Read and write from a file.
 
     Usage:
-        kernel.import_skill(FileIOSkill(), skill_name"file")
+        kernel.import_skill(FileIOSkill(), skill_name="file")
 
     Examples:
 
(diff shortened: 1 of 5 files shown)


python/semantic_kernel/core_skills/file_io_skill.py
```diff
@@ -11,7 +11,7 @@ class FileIOSkill:
     Description: Read and write from a file.
 
     Usage:
-        kernel.import_skill("file", FileIOSkill());
+        kernel.import_skill(FileIOSkill(), skill_name"file")
```
amsacha, line 14:
```suggestion
        kernel.import_skill(FileIOSkill(), skill_name="file")
```

amsacha (changes requested):
there's a small typo but besides that looks good to me!

amsacha (approved):

shawncal (approved):

shawncal:
Good catch. Thanks for the contribution @HenriSchulte-MS!

PR #1735, opened 2023-06-27
Re-implementing the Postgres memory store

JadynWong:
### Motivation and Context

Reimplement Postgres in-memory storage based on the discussion in #1338.

### Description

- Re-implementing the Postgres memory store, mapping between SK collection to Postgres table.
- PostgresMemoryStore no longer implements `IDisposable` pattern.
- No longer execute the enable pgvector extension statement, add it to README.md to be executed by the user. It only needs to be executed once in the database, and the extension may be enabled differently for different hosting methods.

### Contribution Checklist


JadynWong:
Hi @dmytrostruk,
Following the previous discussion on implementing the new design, I hope you can review it and give suggestions. 
Thank you.

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
```diff
@@ -345,4 +242,42 @@ private async Task<DatabaseEntry> ReadEntryAsync(NpgsqlDataReader dataReader, bo
...
+    /// <param name="conn">An opened <see cref="NpgsqlConnection"/> instance.</param>
+    /// <param name="collectionName">The name assigned to a collection of entries.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns></returns>
+    private async Task CreateIndexAsync(NpgsqlConnection conn, string collectionName, CancellationToken cancellationToken = default)
+    {
+        using NpgsqlCommand cmd = conn.CreateCommand();
+        cmd.CommandText = $@"
+            CREATE INDEX IF NOT EXISTS {collectionName}_ix
+            ON {this._schema}.""{collectionName}"" USING ivfflat (embedding vector_cosine_ops) WITH (lists = 1000)";
```
dmytrostruk, line 278: What is `1000` in this case? Would be nice to document it and/or put it in separate variable with naming.
Should we provide an option for users to provide their own value of this parameter when initializing `PostgresMemoryStore`?

dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresMemoryStore.cs
```diff
@@ -10,34 +10,29 @@
...
-        dataSourceBuilder.UseVector();
-
-        var memoryStore = new PostgresMemoryStore(dataSourceBuilder.Build());
-        using NpgsqlConnection dbConnection = await memoryStore._dataSource.OpenConnectionAsync(cancellationToken).ConfigureAwait(false);
-        await memoryStore._dbConnector.CreatePgVectorExtensionAsync(dbConnection, cancellationToken).ConfigureAwait(false);
-        await memoryStore._dbConnector.CreateTableAsync(dbConnection, vectorSize, cancellationToken).ConfigureAwait(false);
-        await memoryStore._dbConnector.CreateIndexAsync(dbConnection, cancellationToken).ConfigureAwait(false);
-        return memoryStore;
+        this._dataSource = dataSource;
+        this._dbConnector = new Database(schema, vectorSize);
```
dmytrostruk, line 35: We should also provide a way for users to inject their own implementation of `Database`. Please do the following:
1. As mentioned in previous comment, rename `Database` to `PostgresDbClient` (or `PostgresDatabaseClient`).
2. Create interface based on previous step (e.g. `IPostgresDbClient`) and define all methods which are needed for `PostgresMemoryStore` for interaction.
3. Create new constructor overload for `PostgresMemoryStore` and allow to inject new interface.

Advantages:
1. Users will be able to inject their own implementation or use the default one in case of already existing constructor.
2. It will be possible to unit test `PostgresMemoryStore` with mocked `IPostgresDbClient` interface.

Please check `ChromaMemoryStore` implementation for reference.

dotnet/src/Connectors/Connectors.Memory.Postgres/README.md
```diff
@@ -16,15 +20,84 @@ How to install the pgvector extension, please refer to its [documentation](https
...
+```sql
+-- Create new tables, each with the name of the collection field value
+DO $$
+DECLARE
+    r record;
+BEGIN
+    FOR r IN SELECT DISTINCT collection FROM sk_memory_table LOOP
+
+        -- Drop Table
+        EXECUTE format('DROP TABLE IF EXISTS %I;', r.collection);
```
dmytrostruk, line 71: Thank you for this migration script!
One note: we should not drop user existing tables. If previously user had `contacts` collection in `sk_memory_table` and separate `contacts` table in Postgres database - during migration script execution we should throw an error that table with the same name already exists, so user will need to decide what to do next, leave it as it is or drop table manually. 
Migration script should not brake user's current DB architecture.

dmytrostruk (changes requested):
@JadynWong thank you for contribution and quick action for Postgres memory store. While this implementation has better design and resolves a lot of issues of previous version, there are still a couple of problems to resolve. I think they are not critical but will improve the functionality even more. Thanks again!

dotnet/src/IntegrationTests/Connectors/Memory/Postgres/PostgresMemoryStoreTests.cs
```diff
@@ -637,85 +506,181 @@ public async Task GetNearestMatchesDifferentiatesIdenticalVectorsByKeyAsync()
...
+        }
+
+        await using (NpgsqlConnection conn = await this._dataSource.OpenConnectionAsync())
+        {
+            await using (NpgsqlCommand command = new("CREATE EXTENSION vector", conn))
+            {
+                await command.ExecuteNonQueryAsync();
+            }
+            await conn.ReloadTypesAsync();
+        }
```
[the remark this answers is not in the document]
JadynWong: The first DB connection is connected to the `postgres` database. It is used to create a new test database.

The second DB connection is connected to the new test database. It is used to enable pgvector extension on the new test database.

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
```diff
@@ -11,113 +10,98 @@
...
     /// <summary>
-    /// Create pgvector extensions.
+    /// Check if a collection exists.
     /// </summary>
     /// <param name="conn">An opened <see cref="NpgsqlConnection"/> instance.</param>
+    /// <param name="collectionName">The name assigned to a collection of entries.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns></returns>
-    public async Task CreatePgVectorExtensionAsync(NpgsqlConnection conn, CancellationToken cancellationToken = default)
+    public async Task<bool> DoesCollectionExistsAsync(NpgsqlConnection conn,
```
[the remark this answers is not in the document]
JadynWong: According to the npgsql documentation it seems that `NpgsqlDataSource` is better recommended.

Also when I first implemented it, I tried to pass only `NpgsqlConnection` to `PostgresMemoryStore`. It opens only once, but I encountered some errors, and it seems that it does not support parallel use.

Can we still use `NpgsqlDataSource`, but pass it to Database(PostgresDbClient) as a private variable?

https://www.npgsql.org/doc/release-notes/7.0.html#new-features

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
JadynWong: You are correct. According to the documentation this value should be set according to the number of records. But I'm not sure if there is a single setting that applies to each collection. Do you recommend using the initial setting for all collections?

dotnet/src/Connectors/Connectors.Memory.Postgres/README.md
JadynWong: I commented out this sql block, and added a comment, just like drop sk_memory_table, to let the user decide whether to enable it or not, so that the script can be executed multiple times. What do you think?

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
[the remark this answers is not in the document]
dmytrostruk: > Can we still use `NpgsqlDataSource`, but pass it to Database(PostgresDbClient) as a private variable?

Sounds good, thanks!

dotnet/src/Connectors/Connectors.Memory.Postgres/Database.cs
dmytrostruk: We can provide this setting as optional parameter of `PostgresMemoryStore` and specify in XML documentation that it will be applied to all collections. If user will need more granular setting - it will be possible to create index manually or provide IPostgresDbClient implementation with custom behavior.

dotnet/src/Connectors/Connectors.Memory.Postgres/README.md
dmytrostruk: > I commented out this sql block, and added a comment, just like drop sk_memory_table, to let the user decide whether to enable it or not, so that the script can be executed multiple times. What do you think?

Sounds awesome, thank you!

dotnet/src/IntegrationTests/Connectors/Memory/Postgres/PostgresMemoryStoreTests.cs
```diff
@@ -637,85 +504,181 @@ public async Task GetNearestMatchesDifferentiatesIdenticalVectorsByKeyAsync()
...
+
+        // Assert
+        Assert.NotNull(keys);
+        Assert.Equal(numRecords, keys.Count - 3);
+        Assert.Equal(numRecords, resultRecords.ToEnumerable().Count());
+    }
+
+    #region private ================================================================================
+
+    private const string ConnectionString = "Host=localhost;Database={0};User Id=postgres";
```
dmytrostruk, line 610: This connection string misses `Password` property, so it's not possible to run integration tests without modifying it. 

Please add section for Postgres in this configuration file with empty value:
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/IntegrationTests/testsettings.json

Then read Postgres section and password like we do this for OpenAI for example:
https://github.com/microsoft/semantic-kernel/blob/16d52c155a2c8193fef7207a8816ac8c4a8eed2c/dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs#L32-L38

Later we have a plan to start Docker containers during build pipeline, so it will be possible to run all integration tests which currently rely on Docker and are ignored.

This change will allow to store password for Postgres together with other keys and secrets for all our integration tests.

dmytrostruk (approved):
@JadynWong thank you for quick action on PR comments! I tested the functionality using integration tests together with `Example39_Postgres` - it works as expected! 🎉 
Thanks for adding unit tests, taking into account that integration tests are temporarily ignored, unit tests will increase our code coverage.

I left minor comments in documentation and tests. As soon as they are resolved, this PR will be ready to merge to main.

<hr>

Ideas for improvements in future (**this can be done in follow-up PRs**):
Since we have `PostgresDbClient` which is responsible for sending Postgres queries, it would be better if we operate in this class with Postgres entities. 

For example, `IPostgresDbClient` interface has method `GetCollectionsAsync`, but there is no entity `collection` in Postgres, there is `table`. So, what if methods in `IPostgresDbClient` will be like `GetTablesAsync`, `CreateTableAsync` etc? 

Another example, in `PostgresDbClient` we have method `CreateCollectionAsync` and in this method there are two calls `CreateTableAsync` and `CreateIndexAsync`, which means that this class defines a logic for creating specific `table` in Postgres which should act as SK `collection`, but this is responsibility of `PostgresMemoryStore` class.

If we follow single responsibility principle, there would be:
- PostgresDbClient - which operates only with Postgres entities and is responsible for sending simple queries for create/delete tables, records, search.
- PostgresMemoryStore - which will work as adapter between Semantic Kernel and Postgres, using `PostgresDbClient` to operate with Postgres entities.

Same applies to other connectors where we have the same problem.

JadynWong:
> Ideas for improvements in future (**this can be done in follow-up PRs**): Since we have `PostgresDbClient` which is responsible for sending Postgres queries, it would be better if we operate in this class with Postgres entities.
> 
> For example, `IPostgresDbClient` interface has method `GetCollectionsAsync`, but there is no entity `collection` in Postgres, there is `table`. So, what if methods in `IPostgresDbClient` will be like `GetTablesAsync`, `CreateTableAsync` etc?
> 
> Another example, in `PostgresDbClient` we have method `CreateCollectionAsync` and in this method there are two calls `CreateTableAsync` and `CreateIndexAsync`, which means that this class defines a logic for creating specific `table` in Postgres which should act as SK `collection`, but this is responsibility of `PostgresMemoryStore` class.
> 
> If we follow single responsibility principle, there would be:
> 
> * PostgresDbClient - which operates only with Postgres entities and is responsible for sending simple queries for create/delete tables, records, search.
> * PostgresMemoryStore - which will work as adapter between Semantic Kernel and Postgres, using `PostgresDbClient` to operate with Postgres entities.
> 
> Same applies to other connectors where we have the same problem.

Great advice, I'd like to finish it in a follow up PR.

dmytrostruk (approved):
LGTM!

Issue #1443, opened 2023-06-13
The sample is not working as intended. (Example07_BingAndGoogleSkills)


iyulab-caveman:
**Describe the bug**
  When running the `Example07_BingAndGoogleSkills` example in `KernelSyntaxExamples`, I get an incorrect result...
  
  **To Reproduce**
1. Run `Example07_BingAndGoogleSkills` Project
2. Step `Example2Async` Method

** Result Console **
```
======== Bing and Google Search Skill ========
What's the largest building in the world?
----
Historically, the world's tallest man-made structure was the Great Pyramid of Giza in Egypt, which ...
----
======== Use Search Skill to answer user questions ========
Who is the most followed person on TikTok right now? What's the exchange rate EUR:USD?
---- Fetching information from Bing...
Information found:

반의어: answer. 참조: 함께 사용되는 동사 : [동사 + question] ask (sb) | have | address, put | bombard sb with, fire | frame, phrase | face | answer, reply to, respond to | do (used only about written questions) | field | avoid, dodge, evade, ignore. If you have any questions, please let me know. 질문이 있으시면, 저에게 ....
warn: object[0]
      Function _GLOBAL_FUNCTIONS_.funccfde0e2bde40429face53e4eefb8dc26 has been invoked with an explicit input text that is different and overrides the input text defined in the context
```

PR #1757, opened 2023-06-28
Fix SKFunction.InvokeAsync to respect `input` when passed in explicitly.

NTaylorMullen:
- The defaults for `mutableContext` are `true` which make it super easy to fall into this trap. I slammed my head against the wall for a while until realizing that the SK InvokeAsync method wasn't properly utilizing the `input` parameter.

### DIFF
--- a/dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
+++ b/dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
@@ -125,19 +125,14 @@ public static class SKFunctionExtensions
                 function.SkillName, function.Name);
         }
 
-        if (mutableContext)
+        if (!mutableContext)
         {
-            // Mutate context to utilize the passed in input
-            context.Variables.Update(input);
-            return function.InvokeAsync(context, settings);
+            // Create a copy of the context, to avoid editing the original set of variables
+            context = context.Clone();
         }
 
-        // Create a copy of the context, to avoid editing the original set of variables
-        SKContext contextClone = context.Clone();
-
-        // Store the input in the context clone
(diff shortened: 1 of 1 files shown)


dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
```diff
@@ -127,6 +127,8 @@ public static ISKFunction UseFrequencyPenalty(this ISKFunction skFunction, doubl
 
         if (mutableContext)
```
danpere, line 128: If that's the intended behavior, this could be refactored to eliminate duplicate code to make it more clear what's going on:
```cs
        if (!mutableContext)
        {
            // Create a copy of the context, to avoid editing the original set of variables
            context = context.Clone();
        }

        // Store the input in the context
        context.Variables.Update(input);
        return function.InvokeAsync(context, settings);
```

NTaylorMullen: ooo I like

stephentoub:
@shawncal, @dluc, what are the scenarios for having this mutableContext parameter at all?  I agree with @NTaylorMullen this is a pit of failure.  In general these SKContexts are mutable... if for a given call someone wants to instead clone it and operate on a clone, why not just have them do exactly that, i.e. pass in `context.Clone()` rather than `context`?

stephentoub (dismissed):

shawncal:
@stephentoub, @NTaylorMullen Agreed, we need to drop this function.

Partial fix was checked in with PR:
https://github.com/microsoft/semantic-kernel/pull/1768
...to fix bug #1443

shawncal (approved):

PR #1772, opened 2023-06-29
Fix to use semantic function default values

markwallace-microsoft:
### Motivation and Context

  1. Why is this change required? Default values specified in a semantic function config.json are not being used when executing the semantic function

### Description

1. Add logic to use default values when executing a semantic function
2. Add a new extension method to allow execution of the text prompt

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
+++ b/dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
@@ -264,12 +264,10 @@ public sealed class SKFunction : ISKFunction, IDisposable
 
         if (this.IsSemantic)
         {
-            var addedVariables = this.AddDefaultValues(context.Variables);
+            this.AddDefaultValues(context.Variables);
 
             var resultContext = await this._function(this._aiService?.Value, settings ?? this._aiRequestSettings, context).ConfigureAwait(false);
             context.Variables.Update(resultContext.Variables);
-
-            this.RemoveDefaultValues(context.Variables, addedVariables);
         }
         else
         {
@@ -302,11 +300,8 @@ public sealed class SKFunction : ISKFunction, IDisposable
         ILogger? logger = null,
         CancellationToken cancellationToken = default)
     {
-        ContextVariables variables = new(input);
-        this.AddDefaultValues(variables);
-
         SKContext context = new(
-            variables: variables,
+            new ContextVariables(input),
             memory: memory,
             skills: this._skillCollection,
             logger: logger,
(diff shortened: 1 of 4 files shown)


dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -264,8 +264,12 @@ public async Task<SKContext> InvokeAsync(SKContext context, CompleteRequestSetti
 
         if (this.IsSemantic)
         {
+            var addedVariables = this.AddDefaultValues(context.Variables);
+
             var resultContext = await this._function(this._aiService?.Value, settings ?? this._aiRequestSettings, context).ConfigureAwait(false);
             context.Variables.Update(resultContext.Variables);
+
+            this.RemoveDefaultValues(context.Variables, addedVariables);
```
dmytrostruk, line 272: I think it's a little bit dangerous to remove something from the context after function execution based on function configuration.

Scenario:
_SKFunction1_ has `input` parameter with default value, value is not presented by user.
_SKFunction2_ has `input` parameter without default value, which should be taken from _SKFunction1_.

Action:
I call two functions in pipeline - `kernel.RunAsync(skFunction1, skFunction2)`

Expected behavior:
_SKFunction1_ returns result and put it in SKContext as `input` for next _SKFunction2_.
_SKFunction2_ function will use `input` parameter with response from _SKFunction1_.

Actual behavior:
_SKFunction1_ adds default value to the context and executes. In `addedVariables` we will have one parameter `input`. After execution it removes `input` from SKContext.
_SKFunction2_ has no default parameter value (as described in Scenario) and input from previous _SKFunction1_ was removed. So _SKFunction2_ will be without `input`.

I may have missed something, let me know what you think about this case. Maybe worth to check this behavior with unit-tests.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -1033,5 +1040,29 @@ private static void TrackUniqueParameterType(ref bool hasParameterType, MethodIn
     /// <summary>Formatter functions for converting parameter types to strings.</summary>
     private static readonly ConcurrentDictionary<Type, Func<object?, CultureInfo, string>?> s_formatters = new();
 
+    /// <summary>Add default values to the context variables if the variable is not defined</summary>
+    private List<string> AddDefaultValues(ContextVariables variables)
+    {
+        var addedVariables = new List<string>();
+        foreach (var parameter in this.Parameters)
+        {
+            if (!variables.ContainsKey(parameter.Name) && parameter.DefaultValue != null)
```
dmytrostruk, line 1049: Should we also handle the case when parameter exists in context variables, but it's null or empty?

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
markwallace-microsoft: Good suggestion.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
markwallace-microsoft: The current thinking is only use defaults if the parameter has not been set. Depending on the prompt null or empty string could be expected values.

@shawncal what do you think?

shawncal: Yeah, I'd prefer to use default value if `null` (not set). If user has set value to `string.Empty`, that's a valid, explicit choice that could be a way of saying "ignore this value -- add nothing to the prompt here" and a way of inserting optional clauses into a prompt.

shawncal (approved):

PR #1791, opened 2023-06-30
Added example 52 which returns Citations from the Grounding Context

sandeepvootoori:
### Motivation and Context
With this PR, I am adding an example which has prompt fine tuned to return citation at the end of the response in markdown format. This will help other developers of SK to return Bing or Microsoft 365 Copilot like citations or references in their responses.


sandeepvootoori:

shawncal:
Hi @sandeepvootoori Thanks for the contribution! This looks like a neat semantic function.

I don't think KernelSyntaxExamples is quite the right place for this -- that project is meant to show how to use specific kernel features, whereas this is a great semantic function that could be built for SK.

Can I recommend publishing this in a new standalone repo, as a new Skill, and linking back to SK? We'll happily highlight and showcase it as part of a demo or blog post.

Issue #1793, opened 2023-06-30
Smeantic-Kernal NuGets fail in Function Apps


wrharper-AASP:
**Describe the bug**
If you use Azure Function Apps with version 15.230531.5-preview it will work.
Every version past that gives the error in the screenshot. 0.15.320609.2-preview and onward.

**To Reproduce**
Just make an Azure Function App in Visual Studio 2022 as a httptrigger and import these files. After you run it, you will see the error.

**Expected behavior**
Needs to be able to load latest versions.

**Screenshots**
![image](https://github.com/microsoft/semantic-kernel/assets/117128832/fbbdc315-ef58-4812-b135-75bdbc7276c1)

**Desktop (please complete the following information):**
In descriptions already, on windows 11

wrharper-AASP:
I believe this is the problem in the patch notes:
Upgrade System.Text.Json and Microsoft.Extensions.Logging versions to 7.0 by @stephentoub in https://github.com/microsoft/semantic-kernel/pull/1297

wrharper-AASP:
I reverted changes (manually created a build) the best I could from: https://github.com/microsoft/semantic-kernel/pull/1297/commits/bc7627ae41dac088d0990a8e7cd0bff3d902041e until this is resolved.

ericstj:
cc @fabiocav
AzureFunctions has a managed host and loads functions into the same process -- as a result the host will *pin* a number of the assemblies it uses since it needs to share types from those assemblies with functions and agree on the type definitions.  As a result - those functions cannot upgrade the assemblies pinned by the host.  It looks like Azure functions doesn't do anything to notify the function project when it's in a state where it's referenced versions of these that cannot be loaded by the host.

Here's a repro project for the broken in-process azure function: 
[FunctionApp1.zip](https://github.com/microsoft/semantic-kernel/files/11919528/FunctionApp1.zip)

Azure functions did create a solution for this problem, and that's the "Isolated" function model.  When you create your function, make sure to select "Isolated" as shown below.
<img width="470" alt="image" src="https://github.com/microsoft/semantic-kernel/assets/8918108/3c9bb388-b2cf-4bbc-9d39-552bb9653482">

Here's a repro project for the working isolated process azure function:
[FunctionApp2.zip](https://github.com/microsoft/semantic-kernel/files/11919532/FunctionApp2.zip)

[...]

PR #1799, opened 2023-06-30
Downgrade several 7.0.* packages to 6.0.*

stephentoub:
This pains me, as it's a technical step backwards.  But there are two problems today:
1) The .NET support policy today unfortunately applies not only to the .NET SDK and runtime but also to nuget packages released in the same wave, which means that if someone is relying on eg System.Text.Json and is running on .NET 6, having SK depend on System.Text.Json with a minimum version of 7.0 forces the consumer to also use the 7.0 version of System.Text.Json from the nuget, which then means that one assembly falls under the STS support policy instead of LTS support policy, which means that one assembly will have support end for it a few months earlier than it otherwise would.  I've raised this issue for further evaluation on the .NET side of things, but in the meantime, this lowers the version number to remove the perceived problem and possible barrier to adoption.
2) Azure Functions has two deployment models: in-process and isolated.  Ideally functions use isolated, which gives them the freedom to reference whatever they need.  The in-process model is exactly what it sounds like: the function runs in the same process as the host, and the host pins several dependencies at 6.0 versions.  That means that if a function references a 7.0 version, it'll fail to load in the in-process model.  While we'd like for folks to be using the isolated model, we can't force it, and we don't want a need for the in-process model to block SK usage.

This commit downgrades back to the 6.0 versions, at least where possible.  Some of the connectors reference libraries (e.g. NRedisStack, pgvector, etc.) that themselves have a 7.0 dependency.

Closes https://github.com/microsoft/semantic-kernel/issues/1793

### DIFF
--- a/dotnet/src/Connectors/Connectors.UnitTests/Connectors.UnitTests.csproj
+++ b/dotnet/src/Connectors/Connectors.UnitTests/Connectors.UnitTests.csproj
@@ -20,6 +20,10 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+
+    <!-- Because some of the referenced projects have dependencies that themselves have System.Text.Json set with a minimum of 7.0. -->
+    <PackageReference Include="System.Text.Json" />
+    <PackageVersion Update="System.Text.Json" Version="7.0.3" />
   </ItemGroup>
 
   <Import Project="$(RepoRoot)/dotnet/src/InternalUtilities/src/InternalUtilities.props" />
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,12 @@ updates:
     directory: "dotnet/"
     schedule:
       interval: "weekly"
+    ignore:
+      # For all System.* and Microsoft.Extensions.* packages, ignore all major version updates
+      - dependency-name: "System.*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "Microsoft.Extensions.*"
+        update-types: ["version-update:semver-major"]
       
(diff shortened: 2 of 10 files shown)


stephentoub:
cc: @ericstj

dotnet/src/Connectors/Connectors.UnitTests/Connectors.UnitTests.csproj
```diff
@@ -20,6 +20,10 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+
+    <!-- Because some of the referenced projects have dependencies that themselves have System.Text.Json set with a minimum of 7.0. -->
```
ericstj, line 24: In those cases can't you just let those references win?  Why do you need an additional reference here?

stephentoub: Because otherwise it produces nuget conflicts warnings

shawncal:
Could you please also add the downgraded packages to [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) sections in https://github.com/microsoft/semantic-kernel/blob/main/.github/dependabot.yml?
Otherwise, we'll be right back where we are in a couple days ;p

stephentoub:
> Could you please also add the downgraded packages to [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) sections

I tried.  Let me know if this looks right / wrong.

shawncal (approved):

Woland2k:
Thank you for the update and listening for the community on this!

Issue #1726, opened 2023-06-27
ActionPlanner should be more forgiving when parsing out plan results.


lemillermicrosoft:
cc @amsacha 

Related to #1496 and #1615

PR #1802, opened 2023-06-30
Extracting plan json from ActionPlanner result

teresaqhoang:
### Motivation and Context

LLM can return additional texts (i.e., preambles, explanations) around the proposed plan, so this PR extracts a well-formed JSON from the result of the planner using a RegEx and balanced group concepts. 

In response to: https://github.com/microsoft/semantic-kernel/issues/1726

### Description

Also, updates any use of Plan State to Plan Parameters due to previous Planner refactor.

### Contribution Checklist


dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -83,20 +84,35 @@ public async Task<Plan> CreatePlanAsync(string goal)
...
         ActionPlanResponse? planData;
-        try
+
+        /* Filter out good JSON from the result in case additional text is present:
+        * Follows the balancing group regex defined here: https://learn.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#balancing-group-definitions 
+        */
+        Regex planRegex = new("^[^{}]*(((?'Open'{)[^{}]*)+((?'Close-Open'})[^{}]*)+)*(?(Open)(?!))", RegexOptions.Singleline);
+        Match match = planRegex.Match(result.ToString());
+
+        if (match.Success && match.Groups["Close"].Length > 0)
```
dmytrostruk, line 94: I would move this logic to separate private method. It will allow to: 
- Replace comments directly in the code with XML documentation above the method, which is responsible for Regex manipulations.
- Reduce code nesting for Regex `if` statement.
- Keep method `CreatePlanAsync` simple.

dmytrostruk (commented):
Thanks @teresaqhoang !
It would be great to add unit test under `Extensions.UnitTests\Planning` to verify and showcase how new regex resolves an issue.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
teresaqhoang: Moved to private `ParsePlannerResult` method

PR #1830, opened 2023-07-03
Removing Copilot Chat from the SK-dotnet.sln

shawncal:
### Motivation and Context
Copilot chat now builds, and generate packages, from its own series of Actions scripts that reference the csproj directly. This project, despite being included in the sln file, is not project-linked to any other (it uses nuget packages instead), nor does it share common dependencies. 

Having it in the sln...
- produces longer build times, 
- raises dependabot updates for copilot changes mislabeled as "...under /dotnet"
- does not get any copilot test coverage

### Description
Removing the Copilot Chat project file from the general SK-dotnet.sln file.
Updating GitHub Actions scripts.

### Contribution Checklist


gitri-ms:
Should we add a new .sln for Copilot Chat to the repo?

hathind-ms (approved):

shawncal:
> Should we add a new .sln for Copilot Chat to the repo?

Certainly can -- I don't think it's (currently) necessary, as the cp chat build scripts just use the csproj file as a target, instead of a sln. If there end up being more than one dotnet project, then yes, certainly a solution would be a good idea.

PR #1832, opened 2023-07-04
.Net: Update KernelSyntaxExamples env variables

shawncal:
### Description
Updating KernelSyntaxExamples to use the same environment variable keys as the Integration tests. This is a step toward making these tests simpler to run, and even running them as automation in CI tests.

This is a first pass -- more cleanup to be done.

### Contribution Checklist


dmytrostruk (commented):
@shawncal I just tried to set my key with command:
```
dotnet user-secrets set "OpenAI:ApiKey" "..."
```

If I try to fetch key with `Env.Var("OpenAI__ApiKey")` - it's not working, configuration doesn't exist, but when I use `Env.Var("OpenAI:ApiKey")` - it's working as expected.

I already saw this conversion pattern from `:` to `__`, but in my case the conversion is not needed. Is it something wrong with my setup?

dmytrostruk (approved):

dmytrostruk (approved):

dmytrostruk (approved):

PR #1841, opened 2023-07-04
[.Net CoreSkills]: Added ability to send custom Headers in HttpSkill

sus19agg:
### Motivation and Context

Currently, the HttpSkill present as part of the core skills in Semantic Kernel dotnet source does not have an option to add any custom headers to the http request. This makes it hard to make many of the API calls using this Core skill as generally some or the other authentication headers or other keys need to be added to the headers.

### Description

- This change adds the ability to apply custom headers to the Http request by adding a new optional parameter to the various method functions of Http skill. 
- This new parameter is optional headers for the request and is a serialized List of keyVaulePair types. HttpHeaders is also an enumerable of KeyValuePairs (https://learn.microsoft.com/en-us/dotnet/api/system.net.http.headers.httpheaders?view=net-7.0). 
- This List should be serialized using System.Text.Json which is then used to deserialize the headers before adding to the request. Serialization is important as Semantic Kernel operates on strings only.
- The "serializedHeaders" variable is completely optional and not necessary for the request.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Skills/Skills.Core/HttpSkill.cs
+++ b/dotnet/src/Skills/Skills.Core/HttpSkill.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Net.Http;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.SemanticKernel.SkillDefinition;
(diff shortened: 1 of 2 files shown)


sus19agg:
@shawncal - created this new PR with the custom headers option in the HTTP Skill which is not yet covered in the earlier pull request.

shawncal:
SKFunctions are primarily meant for the AI to call; it's very unlikely that the AI will know the specific serialized input format this headers arg should take.

Custom headers can be passed into the skill via the HttpClient (constructor arg). This is useful for setting up an HttpSkill where all calls should include a consistent set of headers. If there are situations where headers should be tuned for each specific call, then we'd recommend creating a new Skill class and SKFunctions, to ensure that its inputs and format are very clear to the AI and properly crafted for that specific API.

PR #1842, opened 2023-07-04
Copilot Chat: Url DataAnnotation is not compatible with Uri. Should be string

jpaarhuis:
Having Url annotation check with an Uri type always fails.

The reason is that the UrlAttribute IsValid method always checks if the value is of type string:
```cs
public override bool IsValid(object? value)
{
    if (value == null)
    {
        return true;
    }

    return value is string valueAsString &&
        (valueAsString.StartsWith("http://", StringComparison.OrdinalIgnoreCase)
        || valueAsString.StartsWith("https://", StringComparison.OrdinalIgnoreCase)
        || valueAsString.StartsWith("ftp://", StringComparison.OrdinalIgnoreCase));
}
```

Changing Uri to string here works fine (tested) while keeping the annotation check.

For reference see: https://github.com/dotnet/runtime/issues/71008

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/ConfigurationExtensions.cs
+++ b/samples/apps/copilot-chat-app/webapi/ConfigurationExtensions.cs
@@ -37,7 +37,7 @@ internal static class ConfigExtensions
                 reloadOnChange: true);
 
             // For settings from Key Vault, see https://learn.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-8.0
-            string? keyVaultUri = builderContext.Configuration["KeyVaultUri"];
+            string? keyVaultUri = builderContext.Configuration["Service:KeyVault"];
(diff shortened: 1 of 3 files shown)


jpaarhuis:

gitri-ms:
FYI the markdown check failing is not due to this PR, it's a separate issue with a fix incoming

jpaarhuis:
> FYI the markdown check failing is not due to this PR, it's a separate issue with a fix incoming

What about the merge gatekeeper? It's failing, but don't know why.

gitri-ms:
> > FYI the markdown check failing is not due to this PR, it's a separate issue with a fix incoming
> 
> What about the merge gatekeeper? It's failing, but don't know why.

Merge gatekeeper is failing because the markdown check failed. Nothing to worry about there either.

gitri-ms (approved):

PR #1869, opened 2023-07-06
GPT3 tokenizer

milderhc:
Add GPT3 tokenizer.

Based on .NET implementation.


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/tokenizers/GPT3Tokenizer.java
```diff
@@ -0,0 +1,290 @@
...
+            return new ArrayList<>();
+        }
+    }
+
+    private static int encodingUtf8GetByteCount(String str) {
+        return str.getBytes(StandardCharsets.UTF_8).length;
+    }
+
+    private static int encodingUtf8GetBytes(String str, byte[] bytes) {
+        byte[] strBytes = str.getBytes(StandardCharsets.UTF_8);
```
dsgrieve, line 198: My first reaction here was 'possible NPE'. But then I looked closer and see that the caller ensures str is not null and bytes is not null. I think this is a good place for an `assert str != null` and `assert bytes != null`. This is just my opinion but this tells the reader that you expect the caller to check for null. In other words, you are asserting that str is not null. The assert is only enabled with -ea.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/tokenizers/GPT3Tokenizer.java
```diff
@@ -0,0 +1,290 @@
...
+
+        while (word.size() >= 2) {
+            for (int pairIndex = 0; pairIndex < word.size() - 1; pairIndex++) {
+                Tuple2<String, String> pair =
+                        Tuples.of(word.get(pairIndex), word.get(pairIndex + 1));
+
+                long pairRank =
+                        GPT3Settings.bpeRanks.containsKey(pair)
+                                ? GPT3Settings.bpeRanks.get(pair)
+                                : 100_000_000_000L;
```
dsgrieve, line 232: Love it!

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/tokenizers/GPT3Tokenizer.java
```diff
@@ -0,0 +1,290 @@
...
+        for (char c : token.toCharArray()) {
+            word.add(String.valueOf(c));
+        }
+
+        long smallestRank = Long.MAX_VALUE;
+        Tuple2<String, String> smallestPair = Tuples.of("", "");
+        List<String> newWord;
+
+        while (word.size() >= 2) {
+            for (int pairIndex = 0; pairIndex < word.size() - 1; pairIndex++) {
```
dsgrieve, line 225: Just another one of my things. It doesn't have to be your thing. But something to consider. 
`for (int pairIndex = 0, maxIndex = word.size() - 1; pairIndex < maxIndex; pairIndex++)`
Sometimes calculating the size of something can be expensive. It shouldn't be. But it could be. This is just a habit I've developed over the years.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/tokenizers/GPT3Tokenizer.java
```diff
@@ -0,0 +1,290 @@
...
+
+            if (!GPT3Settings.bpeRanks.containsKey(smallestPair)) {
+                break;
+            }
+
+            String first = smallestPair.getT1();
+            String second = smallestPair.getT2();
+
+            newWord = new ArrayList<>(word.size());
+            for (int i = 0; i < word.size(); i++) {
```
dsgrieve, line 248: I know this is a port, but I'd consider making this for loop a subroutine just to improve readability and shorten the method up. Not a requirement, and it is probably better to stick with the reference impl, so ignore this. :^)

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/tokenizers/settings/EmbeddedResource.java
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft. All rights reserved.
+package com.microsoft.semantickernel.tokenizers.settings;
+
+import java.io.*;
```
dsgrieve, line 4: Please import fully qualified classes instead of *. I've got no idea what you are using from java.io at this point. If your IDE is doing this automagically, you can adjust the setting.

dsgrieve (approved):

java/semantickernel-api/pom.xml
```diff
@@ -39,4 +39,11 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <resources>
+            <resource>
+                <directory>src/main/java/com/microsoft/semantickernel/tokenizers/settings/resources</directory>
```
johnoliver, line 45: this should remain at default(src/main/resources), your later files should be at:

```
src/main/resources/com/microsoft/semantickernel/tokenizers/settings/encoder.json
src/main/resources/com/microsoft/semantickernel/tokenizers/settings/vocab.pbe
```

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/tokenizers/GPT3Tokenizer.java
```diff
@@ -0,0 +1,286 @@
...
+                (char) 0x00F8, (char) 0x00F9, (char) 0x00FA, (char) 0x00FB, (char) 0x00FC,
+                        (char) 0x00FD, (char) 0x00FE, (char) 0x00FF
+            };
+
+    private static final Pattern encodingRegex =
+            Pattern.compile(
+                    "'s|'t|'re|'ve|'m|'ll|'d| ?\\p{L}+| ?\\p{N}+|"
+                            + " ?[^\\s\\p{L}\\p{N}]+|\\s+(?!\\S)|\\s+");
+
+    private static final Map<String, List<String>> bpeCache = new HashMap<>();
```
johnoliver, line 94: since this is a static, this is a bit of an uncontrolled memory leak as this map will grow over time. 

- Define a constant that can be changed via properties MAX_TOKENIZER_CACHE_SIZE
- Use LinkedHashMap and limit the size

markwallace-microsoft: /spotless

java/semantickernel-gpt3-tokenizer/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+    <parent>
+        <groupId>com.microsoft.semantickernel</groupId>
+        <artifactId>semantickernel-parent</artifactId>
+        <version>0.2.2-beta-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>semantickernel-gpt3-tokenizer</artifactId>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
```
johnoliver, line 15: Should leave these properties at defaults, is there a need for Java 11?

java/semantickernel-gpt3-tokenizer/src/main/java/com/microsoft/semantickernel/GPT3Tokenizer.java
```diff
@@ -0,0 +1,304 @@
...
+    private static final String MAX_TOKENIZER_CACHE_SIZE_DEFAULT = "100000";
+
+    static {
+        // Limit the maximum cache size
+        int cacheSize =
+                Integer.parseInt(
+                        System.getProperty(
+                                MAX_TOKENIZER_CACHE_SIZE_KEY, MAX_TOKENIZER_CACHE_SIZE_DEFAULT));
+        bpeCache =
+                new LinkedHashMap<String, List<String>>() {
```
brunoborges, line 106: Shouldn't the `LinkedHashMap` be created with the `cacheSize` value in its constructor?

java/semantickernel-gpt3-tokenizer/src/main/java/com/microsoft/semantickernel/GPT3Tokenizer.java
```diff
@@ -0,0 +1,304 @@
...
+    /**
+     * The tokenizer uses a byte-pair encoding (BPE) algorithm to split words into sub-words based
+     * on frequency and merges rules. It can handle out-of-vocabulary words, punctuation, and
+     * special tokens.
+     *
+     * @param text Text to tokenize
+     * @return List of token IDs
+     */
+    public static List<Integer> encode(String text) {
+        List<Integer> bpeTokens = new ArrayList<>();
```
brunoborges, line 123: Is there a way to know how many tokens may be present in the `text`, so the `ArrayList` is created with a sizeable array?

brunoborges (commented):
Besides the comments I've added, another one:

What is the purpose of the `encoder.json`?

java/semantickernel-gpt3-tokenizer/src/main/java/com/microsoft/semantickernel/GPT3Tokenizer.java
milderhc: No, the constructor allocates the initial capacity of the table, but it does not limit it. We could add it, but it doesn't really matter.

java/semantickernel-gpt3-tokenizer/src/main/java/com/microsoft/semantickernel/GPT3Tokenizer.java
milderhc: Not really

milderhc:
It maps the tokens to their IDs, which is memory efficient. The original string can be restored afterwards with those IDs.

johnoliver (approved):

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/syntaxexamples/Example29_Tokenizer.java
```diff
@@ -0,0 +1,84 @@
...
+        ---
+
+
+        The following example uses Windows '\r\n' line separator.
+        ---
+        Some text on
+        two lines
+        Tokens: 7
+        ---
+         */
```
markwallace-microsoft, line 82: Should this comment be part of the method description?

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/syntaxexamples/Example29_Tokenizer.java
```diff
@@ -0,0 +1,84 @@
...
+
+        // Example 2
+        sentence = "⭐⭐";
+        tokenCount = GPT3Tokenizer.encode(sentence).size();
+
+        System.out.println("The following example contains emojis which require several tokens.");
+        System.out.println("---");
+        System.out.println(sentence);
+        System.out.println("Tokens: " + tokenCount);
+        System.out.println("---\n\n");
```
markwallace-microsoft, line 31: This could be extracted out to a utility method

markwallace-microsoft (approved):
Just a couple of nits

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/syntaxexamples/Example29_Tokenizer.java
dsgrieve: My $.02 is that examples are harder to follow when you have to take a detour to some utility method. 

I'll add, however, that this could be easier to read if it used printf.

milderhc: It could, but as printing the tokenCount is what matters the most in the example I would prefer to leave it explicitly here. Also, because it keeps parity with the .NET example: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example29_Tokenizer.cs

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/syntaxexamples/Example29_Tokenizer.java
milderhc: Not really, in all the .NET examples and in the Java ones we have so far, we have the output as a comment block at the end.

johnoliver (approved):

PR #1877, opened 2023-07-06
Java: Configure pom for releases

johnoliver:
Update Pom with settings needed for releasing

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.microsoft.semantickernel</groupId>
     <artifactId>semantickernel-parent</artifactId>
-    <version>0.2.1-beta</version>
+    <version>0.2.2-beta-SNAPSHOT</version>
     <packaging>pom</packaging>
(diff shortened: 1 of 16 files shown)


java/pom.xml
```diff
@@ -628,5 +627,95 @@
...
+                    <id>${releaseRepoId}</id>
+                    <url>${releaseRepoUrl}</url>
+                    <name>${releaseRepoName}</name>
+                </repository>
+            </distributionManagement>
+
+        </profile>
     </profiles>
+
+    <scm>
```
dsgrieve, line 715: In addition to <scm/> Microsoft recommends copyright, <developer/> and <license/> elements. See https://docs.opensource.microsoft.com/releasing/publish-binaries/maven/

Aren't these elements usually at the beginning of a pom? Not that order matters. I wonder if it would be worth applying something like tidy-maven-plugin to our poms.

dsgrieve (approved):

shawncal (approved):

PR #1881, opened 2023-07-06
Removing TrustService from Kernel.Core (to become extension)

shawncal:
### Motivation and Context
The `TrustService`, along with related types like `TrustAwareString` showcase a really interesting pattern, but we'd like to repackage them into a `SemanticKernal.Extensions.Trust` package with a lot more configurability, rather than baking these into the SK Core. With the upcoming support for pre- and post- completion hooks in a Plan, we believe we can build a better solution that leverages those hooks, but is optional, configurable, and fully replaceable if our solution doesn't meet your needs. 

Other reasons:
 - The `TrustAwareString` type is at odds with the transition we're making to allow any objects (not just strings) in SK's context variables, function args, and return types.
 - Security issues have been raised by architects that indicate that this pattern may give a false sense of security, as the strings in the `TrustAwareString` objects can be manipulated by static casts, and are therefore unreliable.
 
### Description
 - Stripping out all `TrustService`, `TrustAwareString`, `IsSensitive` and related references from the SK.Core and Abstractions packages. 

**These will be added back in the form of a new `SemanticKernel.Extensions.Trust` namespace with optional packages and more features, including an auditable trust chain.

### Contribution Checklist


dotnet/src/IntegrationTests/Planning/PlanTests.cs
```diff
@@ -146,7 +146,7 @@ public async Task CanExecutePanWithTreeStepsAsync()
     }
 
     [Theory]
-    [InlineData(null, "Write a poem or joke and send it in an e-mail to Kai.", null)]
+    //[InlineData(null, "Write a poem or joke and send it in an e-mail to Kai.", null)]
```
dmytrostruk, line 149: Is it accidental comment or on purpose? Same for test methods below.

shawncal:
> @shawncal Code changes look good to me, but it looks like something is wrong with Git, since it's grabbing 362 commits and making 58 participants of this PR :) Probably because of merge-rebase issues.

Closing this PR, will resend from a fresh branch.

Issue #1876, opened 2023-07-06
Sample apps depend on outdated Semantic Kernel


victor-v-rad:
**Describe the bug**
Sample apps, such as Copilot Chat, are out of sync with the latest Semantic Kernel dotnet-0.17.230629.1-preview. See [CopilotChatWebApi.csproj](https://github.com/microsoft/semantic-kernel/blob/69fb41131d8e5ba293e409c5ceb451ae753a2811/samples/apps/copilot-chat-app/webapi/CopilotChatWebApi.csproj) and [OpenApiSkillsExample.csproj](https://github.com/microsoft/semantic-kernel/blob/16d52c155a2c8193fef7207a8816ac8c4a8eed2c/samples/dotnet/openapi-skills/OpenApiSkillsExample.csproj#L34) for instance; they both depend on the 0.16.230615.1-preview version.

**Expected behavior**
For each new release, the sample apps (like Copilot Chat) are updated to reflect the latest SK packages.

**Screenshots**
Sample apps, such as Copilot Chat, currently depend on old SK packages.

Thank you for your consideration.

samleegithub:
https://github.com/microsoft/semantic-kernel/pull/1811

Copilot Chat is being updated with this pull request.

PR #1886, opened 2023-07-07
Updating dotnet samples to 17.230704.3

shawncal:
Also included:
 - Removing FileCompression project
   - no longer used
 - Merging GitHubSkill into KernelHttpServer
   - rename to GitHubPlugin (...the first of many to receive such treatment)
   - this is where it's used, and it will allow us to instantiate the github skill without passing the access token as a context variable (this change not yet implemented).
 - Removing LoadPromptsFromCloud
   - we are moving to a "plugin-first" approach - loading plugins, not raw prompts. This approach is no longer recommended.

Resolves https://github.com/microsoft/semantic-kernel/issues/1876


shawncal:
Resolves #1876

dmytrostruk (approved):
LGTM!

samples/dotnet/KernelHttpServer/Plugins/GitHubPlugin.cs
```diff
@@ -102,55 +72,48 @@ public GitHubSkill(IKernel kernel, ILogger<GitHubSkill>? logger = null)
...
-    [SKFunctionContextParameter(Name = SearchPatternParamName, Description = "The search string to match against the names of files in the repository")]
-    public async Task SummarizeRepositoryAsync(string source, SKContext context)
+    /// <param name="input">URI to download the repository content to be summarized</param>
+    /// <param name="repositoryBranch">Name of the repository repositoryBranch which will be downloaded and summarized</param>
+    /// <param name="searchPattern">The search string to match against the names of files in the repository</param>
+    /// <param name="patToken">Personal access token for private repositories</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Name of the memory collection used to store the code summaries.</returns>
+    [Description("Downloads a repository and summarizes the content")]
+    public async Task<string?> SummarizeRepositoryAsync(
```
gitri-ms, line 82: I am seeing this error:

> Microsoft.SemanticKernel.KernelException: Function not available: Function not available GitHubPlugin.SummarizeRepository

With the `SKFunctionName` attribute removed, how does this function get associated with the name `SummarizeRepository`?

shawncal: There's logic in there that's supposed to find the function with and without the "Aync" suffix, but it might not be working as I expected. I'll add the SKName attributes here and the other functions. Thanks for testing, and good find!

shawncal: @gitri-ms Updated -- can you try again please?

I'll also get my environment set up to run these too. Thanks again!

samples/dotnet/graph-api-skills/MsGraphSkillsExample.csproj
gitri-ms, line 1: The MsGraphSkillsExample also seems to be broken, getting this exception:

> System.Collections.Generic.KeyNotFoundException: 'The given key 'GetFileContentAsync' was not present in the dictionary.'

Need to investigate further...

gitri-ms (approved):

PR #1889, opened 2023-07-07
.Net: Update PromptTemplateEngine to render code blocks in parallel

dehoward:
### Motivation and Context

this PR looks to add basic parallelization when rendering code blocks in the template engine. 
today, functions are called sequentially but have the ability to be invoked in either way. 

### Description

- changed `PromptTemplateEngine` to use `Task.WhenAll` instead of `await` when calling `ICodeRendering.RenderCodeAsync`.
  to ensure the code blocks are appended to the prompt in the same order, a placeholder is inserted into the output before the function call has completed. once it's done, the placeholder, a `Guid` in `RenderAsync` and `null` in `RenderCodeAsync`, is replaced with the result of the call.
- added a test for `PromptTemplateEngine.RenderCodeAsync`.

### Contribution Checklist


dehoward:
fyi, we already mention the ability for parallelization in a comment at the top of the file:
https://github.com/microsoft/semantic-kernel/blob/8d705b30adf0ef94792dacb86424b172b613ffcd/dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs#L22C9-L22C136

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -105,20 +119,32 @@ public IList<Block> RenderVariables(IList<Block> blocks, ContextVariables? varia
...
             if (block.Type != BlockTypes.Code)
             {
-                updatedBlocks.Add(block);
+                result.Add(block);
             }
             else
             {
-                var codeResult = await ((ICodeRendering)block).RenderCodeAsync(executionContext).ConfigureAwait(false);
-                updatedBlocks.Add(new TextBlock(codeResult, this._log));
+                result.Add(null);
```
dmytrostruk, line 136: nit: Instead of adding `null` to the list we can define `result` as array `new Block[blocks.Count]`. In this case array items will be `null` by default and we will set result blocks directly by index.

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -65,7 +66,11 @@ public async Task<string> RenderAsync(string templateText, SKContext context)
     public async Task<string> RenderAsync(IList<Block> blocks, SKContext context)
```
dmytrostruk, line 66: Should we add separate unit-test method for `RenderAsync` function in the same way you added for `RenderCodeAsync` method?

dmytrostruk: Would be also interesting to see unit/integration tests to verify how `SKContext` and `ContextVariables` are handled when functions are invoked in parallel and what happens if `Func2` depends on input from `Func1` (if that's a valid case of course).

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -65,7 +66,11 @@ public async Task<string> RenderAsync(string templateText, SKContext context)
...
     {
         this._log.LogTrace("Rendering list of {0} blocks", blocks.Count);
+
         var result = new StringBuilder();
+        var tasks = new List<Task<string>>();
+        var guids = new List<Guid>();
```
shawncal, line 72: We don't need to track requests by guid... see suggestion below.

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -75,7 +80,10 @@ public async Task<string> RenderAsync(IList<Block> blocks, SKContext context)
                     break;
```
shawncal, line 80:
```suggestion
                    tasks.Add(Task.FromResult<string>(staticBlock.Render(context.Variables)));
                    break;
```
(...and remove the result.Append here)

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -75,7 +80,10 @@ public async Task<string> RenderAsync(IList<Block> blocks, SKContext context)
...
 
                 case ICodeRendering dynamicBlock:
-                    result.Append(await dynamicBlock.RenderCodeAsync(context).ConfigureAwait(false));
+                    var guid = Guid.NewGuid();
+                    result.Append(guid);
+                    guids.Add(guid);
+                    tasks.Add(dynamicBlock.RenderCodeAsync(context));
```
shawncal, line 86:
```suggestion
                    tasks.Add(dynamicBlock.RenderCodeAsync(context));
```

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -85,6 +93,12 @@ public async Task<string> RenderAsync(IList<Block> blocks, SKContext context)
             }
         }
 
+        var codeBlocks = await Task.WhenAll(tasks).ConfigureAwait(false);
+        for (var i = 0; i < guids.Count; i++)
+        {
+            result.Replace(guids[i].ToString(), codeBlocks[i]);
```
shawncal, line 99: foreach (Task<string> t in 
```suggestion
        foreach (Task<string> t in tasks)
        {
            result.Append(await t.ConfigureAwait(false));
        }
```
This allows the tasks to all run in parallel, but they'll be processed in order. Static blocks are instantly completed (awaiting a Task.FromResult is a noop). You don't need to WaitAll to get started with the string builder... just need to await each task in order to ensure it's completed by the time you get to it.

shawncal (changes requested):

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
dehoward: thanks for this suggestion, added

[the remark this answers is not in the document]
dehoward: this is a good idea, thanks for this.

see Shawn's suggestion below which helped further simplify the initial approach, I've added that into the changes.

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -105,20 +111,31 @@ public IList<Block> RenderVariables(IList<Block> blocks, ContextVariables? varia
         SKContext executionContext)
```
shawncal, line 111: As far as I can tell, there's nobody using this method. It also doesn't seem like a super useful method to be part of the public SDK surface. @dehoward Instead of updating this method, could you please just remove it?
(Also remove it from the interface, IPromptTemplateEngine, and any unit tests that hit this method directly).

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -65,17 +65,17 @@ public async Task<string> RenderAsync(string templateText, SKContext context)
...
         this._log.LogTrace("Rendering list of {0} blocks", blocks.Count);
-        var result = new StringBuilder();
+        var tasks = new List<Task<string>>();
         foreach (var block in blocks)
         {
             switch (block)
             {
                 case ITextRendering staticBlock:
-                    result.Append(staticBlock.Render(context.Variables));
+                    tasks.Add(Task.FromResult<string>(staticBlock.Render(context.Variables)));
```
shawncal, line 74: On second glace, I think you can drop the `<string>` part, as it'll be implicitly typed by the compiler.

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -65,17 +65,17 @@ public async Task<string> RenderAsync(string templateText, SKContext context)
     public async Task<string> RenderAsync(IList<Block> blocks, SKContext context)
     {
         this._log.LogTrace("Rendering list of {0} blocks", blocks.Count);
-        var result = new StringBuilder();
+        var tasks = new List<Task<string>>();
```
shawncal, line 68: May as well pass in the count to reserve the space, since we know exactly how many there will be!
```suggestion
        var tasks = new List<Task<string>>(blocks.Count);
```

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
dehoward: I've added a unit test for ensuring that variables remain immutable throughout the rendering process. lmk what you think. I with this one and the existing we are covering all scenarios

dotnet/src/SemanticKernel/TemplateEngine/PromptTemplateEngine.cs
```diff
@@ -85,11 +85,22 @@ public async Task<string> RenderAsync(IList<Block> blocks, SKContext context)
...
         // TODO: remove PII, allow tracing prompts differently
         this._log.LogDebug("Rendered prompt: {0}", result);
         return result.ToString();
     }
 
+    public Task<IList<Block>> RenderCodeAsync(IList<Block> blocks, SKContext executionContext)
+    {
+        throw new System.NotImplementedException();
+    }
+
```
shawncal, line 103: What's this for? Now that it's removed from the interface, we shouldn't need this anymore...

dehoward: VSCode autofix added this since I changed the `class` first🤦🏾 removed.

shawncal (approved):

PR #1890, opened 2023-07-07
Remove `dotnet format` check from PR template

dehoward:
proposing that we make this change since 
1) .NET formatting is verified by the PR build, and 
2) there are now several different languages used in the repo.


dmytrostruk:
I would probably keep check for following coding conventions. We can make it more general, so it will refer to C#/Python/Java.

dmytrostruk (commented):
I would probably keep check for following coding conventions. We can make it more general, so it will refer to C#/Python/Java.

shawncal:
I've been thinking we should update this to acknowledge x-language too...
Some notes: "dotnet format" only runs on release builds, which not everybody does.
Dotnet has a "build.sh"/"build.cmd" (Windows) script available that builds, checks format, runs UTs and integration tests...

Maybe we should have similar scripts for python, samples. Or, even better, the script could detect what files are changed and run the appropriate language builds/formatters/linters/tests before submitting a PR... (just an idea, beyond the scope of the PR).

dehoward:
> I would probably keep check for following coding conventions. We can make it more general, so it will refer to C#/Python/Java.

@dmytrostruk that's the thing, we're already checking for coding conventions by default in the PR build. 

to give some context, we recently enabled ESLint in the Copilot Chat build (#1710) and following that I want to integrate Prettier soon as well. it feels a bit cumbersome to continue to update the list of languages in the description as we add more in the repo.

dehoward:
> I've been thinking we should update this to acknowledge x-language too... Some notes: "dotnet format" only runs on release builds, which not everybody does. Dotnet has a "build.sh"/"build.cmd" (Windows) script available that builds, checks format, runs UTs and integration tests...
> 
> Maybe we should have similar scripts for python, samples. Or, even better, the script could detect what files are changed and run the appropriate language builds/formatters/linters/tests before submitting a PR... (just an idea, beyond the scope of the PR).

@shawncal python and java currently have Github workflows (`python-lint.yml` and `java-format.yml`) that verify the formatting for PRs and imo it seems redundant to mention these in the PR description since they will run regardless.
there is also already a workflow for running `dotnet format`:
image

maybe we can scrap this workflow too? since we are already running release builds for PRs?

dmytrostruk:
> there is also already a workflow for running `dotnet format`

As far as I know, it is running only a check, whether the code is formatted, but it's not formatting code automatically, so currently you need to do that manually. When developers submit PRs, they should check whether this build step is successful. If no, then they need to run that script locally to apply formatting changes and commit it. 

Of course, this can be automated, for example by adding pre-commit hooks. 

But it's not only about formatting, but naming convention and other things, so for these things we should use analyzers for automation. But it's not in place yet, that's why this step is present in PR description template :)

So quick resolution would be to mention that in PR description for all languages and for long term - add automation.

shawncal:
> > I've been thinking we should update this to acknowledge x-language too... Some notes: "dotnet format" only runs on release builds, which not everybody does. Dotnet has a "build.sh"/"build.cmd" (Windows) script available that builds, checks format, runs UTs and integration tests...
> > Maybe we should have similar scripts for python, samples. Or, even better, the script could detect what files are changed and run the appropriate language builds/formatters/linters/tests before submitting a PR... (just an idea, beyond the scope of the PR).
> 
> @shawncal python and java currently have Github workflows (`python-lint.yml` and `java-format.yml`) that verify the formatting for PRs and imo it seems redundant to mention these in the PR description since they will run regardless. there is also already a workflow for running `dotnet format`: image
> 
> maybe we can scrap this workflow too? since we are already running release builds for PRs?

The 'Release' builds that we run in GitHub Actions exclude the dotnet format check. It takes a long time to run, and since there are several build flavors, there's no need to run format check on all of them, and surface the same warning n times. The 'dotnet format' GH action is also much more selective about which projects it checks, based on the files changed in the PR. Most times, it only needs to check one or two, instead of the whole solution.

A couple other notes and ideas....
Dmytro's right: the format checks on the builds are just checks and, unlike running `dotnet format` on your machine, they do not automatically fix anything. Instead, they just fail the build. Precommit hooks are a good idea, but require additional setup on each clone -- we can check in the precommit hooks config files, but it would still require all developers to install the tool and initialize the clone on their machine, so it's not a whole lot more effective than "please run a Release build".

Our Java peers have implemented a neat toy for PRs... if java formatting check fails, you can type a certain command into the PR discussion, which kicks off a format job. ("slash spotless"... I won't spell it out here since that'll actually trigger the job). That format, if successful, will submit a PR to the PR branch itself with the fixes.
https://github.com/microsoft/semantic-kernel/blob/main/.github/workflows/java-format.yml

I had a similar setup for "forward slash" dotnet format, but removed it as it was using a 3P action that was no longer compatible with the latest dotnet format command. I do plan to follow the Java example and implement this manually for dotnet and python.

*RECOMMENDATIONS*
@dehoward Here's what I think we should do...

THIS PR:
- Update the [CONTRIBUTING.md](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) file with the instructions we'd like people to run before pushing a PR, including how to run the formatting and tests in each language. (You can start with just dotnet step of "run [build.cmd]([build.cmd](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/build.cmd))/[build.sh]([build.sh](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/build.sh))" in this PR, but leave room and ask others to add the java/python/ts steps in separate PRs.).
- Remove the dotnet-specific line in the template, as you've already done. In the line above, add a bit about formatting. Something like:
 ```
The PR follows [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the (pre-submission formatting script)[[link to CONTRIBUTING section with instructions]] raises no warnings.
```

FUTURE PRs:
- When dotnet format, python lint, node eslint or java's spotless checks fail in the PR, we have the style check Action script automatically post a PR comment instructing the developer how to fix it. That can include the "/format" commands and the instructions for doing so manually.

dehoward:
@shawncal, @dmytrostruk thanks for the great level of detail. I've updated the PR with Shawn's recommended changes and added a "Development scripts" section in `CONTRIBUTING.md`. lmk if we want to tweak the wording.

for python I just pointed to a section in the existing [`python/DEV_SETUP.md`](https://github.com/microsoft/semantic-kernel/blob/main/python/DEV_SETUP.md#pipeline-checks), maybe java/dotnet should have this file as well? just a thought.

also, @shawncal do we want to create item(s) to track the work for setting up the `/format` command?

shawncal (approved):

PR #1893, opened 2023-07-07
Add/Update Action Context variables to main Context for Stepwise Planner

akshaykokane:
Add ActionContext variable to main Context

### Motivation and Context

In stepwise planner, for every action we create new Action Context. Each action can result into adding new context variables. These context variables are potentially used for executing of next action. Currently, the context variables created from Action are ignored and only the "Result" is added back to context. So the action only adds Result to SKContext and doesn't update the context variables in Kernel context.

- Chaining of multiple skills using Stepwise Planner and Context Variables (inline with SequentialPlanner)
- Before implementing Action, reads from main context to check if variable is available if not try to get it from {actionVariables}

  Stepwise Planner uses Context Variables from previous actions to invoke next action

4.  If it fixes an open issue, please link to the issue here. https://github.com/microsoft/semantic-kernel/issues/1894

### Contribution Checklist


shawncal:
@akshaykokane Thanks for the contribution! Looking through this, even after a rebase, I see some use of classes that no longer exist (i.e TrustAwareString) which only means it look us too long to get to this PR. I'm going to close it, only because it's clear that the solution is no longer applicable. That said, the Issue #1894 remains open -- if you can confirm the issue is still... an issue, feel free to re-open the PR with an updated solution, and add me (@shawncal and Roger @RogerBarreto) as reviewers.

Thanks! And apologies for the delay in reviewing this. Your contributions are appreciated, and we'd love to see more!

PR #1899, opened 2023-07-07
Copilot Chat: Display usernames in Copilot Chat

teresaqhoang:
### Motivation and Context

This PR adds support for fetching user data for each user in a chat session, allowing us to display usernames instead of GUIDs in user lists. 

### Description

- User data is loaded per conversation for optimization but only fetched per user. 
- Fetched user data is stored and indexed globally, since users can be part of multiple chats. 
- User data is fetched only if the user exists in the same tenant as the active user due to AAD access restrictions, otherwise, user display name will be "External Guest".

> Note: right now, external user's names will show up in the chat history since we save user names as part of the chat message, but this will be removed in future work: https://github.com/orgs/microsoft/projects/852?pane=issue&itemId=29734028 

image

### Contribution Checklist


teresaqhoang:
Had to create own ParticipantsList component because the AvatarGroupPopover did not allow for custom styling, resulting in extra space when participants list <=2 
image

samples/apps/copilot-chat-app/webapp/src/libs/services/GraphService.ts
```diff
@@ -0,0 +1,58 @@
...
+        const result = await this.getResponseAsync<object>(
+            {
+                commandPath: this.getCommandPath(`users/${userId}`),
+                method: 'GET',
+            },
+            accessToken,
+        );
+
+        return result;
+    };
```
gitri-ms, line 57: Is this used anywhere?

samples/apps/copilot-chat-app/webapp/src/libs/useGraph.ts
```diff
@@ -0,0 +1,156 @@
...
+        const usersToRetry: string[] = [];
+
+        // Copy current state of user data
+        const userData = { ...users };
+
+        // Filter user Ids list to optimize fetch
+        userIds.forEach((userId) => {
+            const ids = userId.split('.');
+            const oid = ids[0];
+            const tid = ids[1];
```
gitri-ms, line 35: Can we make these variable names a little clearer? (I'm pretty sure tid is tenant id, but not totally sure what oid stands for.)

teresaqhoang: Changed to more clarifying variable names and added comment

samples/apps/copilot-chat-app/webapp/src/libs/services/GraphService.ts
teresaqhoang: No, I thought it might be helpful for future work when we load user info for each message, but I'll remove until we actually need it

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
```diff
@@ -98,14 +100,28 @@ const useClasses = makeStyles({
...
 
-    const chatService = new ChatService(process.env.REACT_APP_BACKEND_URI as string);
+    useEffect(() => {
+        if (!conversations[selectedId].userDataLoaded) {
+            void msGraph.loadUsers(chatUsers.map((user: IChatUser) => user.id)).then(() => {
+                dispatch(setUsersLoaded(selectedId));
+            });
+        }
+
+        // Limiting dependencies or else this effect is triggered on all user typing events
```
TaoChenOSU, line 122: nit: list the events that will trigger this effect that we want to exclude.

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
```diff
@@ -98,14 +100,28 @@ const useClasses = makeStyles({
...
+
     const [title = '', setTitle] = useState<string | undefined>(selectedId);
     const [isEditing, setIsEditing] = useState<boolean>(false);
-    const { instance, inProgress } = useMsal();
-    const { alerts } = useAppSelector((state: RootState) => state.app);
 
-    const chatService = new ChatService(process.env.REACT_APP_BACKEND_URI as string);
+    useEffect(() => {
+        if (!conversations[selectedId].userDataLoaded) {
+            void msGraph.loadUsers(chatUsers.map((user: IChatUser) => user.id)).then(() => {
```
TaoChenOSU, line 117: nit: `void` seems unnecessary.

gitri-ms (dismissed):

samples/apps/copilot-chat-app/webapp/src/components/chat/controls/ParticipantsList.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return (
+        <Popover positioning={'below-end'} size="small">
+            <PopoverTrigger>
+                <Tooltip content="Chat participants" relationship="label">
+                    <Avatar initials={`+${Math.min(participants.length, 100)}`} />
+                </Tooltip>
+            </PopoverTrigger>
+            <PopoverSurface className={classes.root}>
+                <div className={classes.list}>
+                    {participants.map((user) => (
```
TaoChenOSU, line 56: nit: `user` -> `participant`

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
```diff
@@ -98,14 +100,28 @@ const useClasses = makeStyles({
...
+                dispatch(setUsersLoaded(selectedId));
+            });
+        }
+
+        // Limiting dependencies or else this effect is triggered on all user typing events
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [selectedId, chatUsers.length]);
```
TaoChenOSU, line 124: What will happen if `chatUsers.length` has changed but `conversations[selectedId].userDataLoaded` is already true? Will it load the new user data?

teresaqhoang: This should never happen because the state for both is updated at the same time and Redux/React is single-threaded: 
image

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
teresaqhoang: This is necessary to adhere to linting rules @dehoward just enabled 
image

TaoChenOSU: Can you call catch on the promise at the end: https://typescript-eslint.io/rules/no-floating-promises/?

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
TaoChenOSU: Ahh, Ic

samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
teresaqhoang: A catch would be redundant here because all the business logic is in the try block inside the `loadUsers` function, so any errors are caught and handled appropriately.  It makes more sense to keep the error handling contained there rather than where the function is invoked

TaoChenOSU (dismissed):

samples/apps/copilot-chat-app/webapp/src/components/chat/controls/ParticipantsList.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const msGraph = useGraph();
+
+    const chatUsers = conversations[selectedId].users;
+    const [loadedParticipants, setLoadedParticipants] = useState(mapParticipantsNames(participants, users));
+
+    useEffect(() => {
+        if (!conversations[selectedId].userDataLoaded) {
+            void msGraph.loadUsers(chatUsers.map((user: IChatUser) => user.id)).then((loadedUsers?: Users) => {
+                if (loadedUsers) {
+                    dispatch(setUsersLoaded(selectedId));
```
dehoward, line 60: then this should likely be reflected in the typing of `UsersState`

samples/apps/copilot-chat-app/webapp/src/redux/features/users/UsersState.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface UsersState {
+    users: Users;
+}
+
+export const initialState: UsersState = {
+    users: {},
+};
+
+export type Users = Record<string, UserData>;
```
dehoward, line 11: building on my previous comment, this seems it should be like below to represent the fact that the data my not be there.
```suggestion
export type Users = Record<string, UserData | undefined>;
```

samples/apps/copilot-chat-app/webapp/src/libs/useGraph.ts
```diff
@@ -0,0 +1,160 @@
...
+
+            // Filter user Ids list to optimize fetch
+            userIds.forEach((userId) => {
+                const ids = userId.split('.');
+                const objectId = ids[0]; // Unique GUID assigned to each user in their home tenant
+                const tenantId = ids[1]; // Home tenant id
+
+                // Active user can only access user data within their own tenant
+                // Mark chat users outside of tenant as External
+                if (activeUserInfo && tenantId !== activeUserInfo.id.split('.')[1]) {
```
dehoward, line 40: nit: can we save this call to a variable so we don't have to do the `.split()` each time?

samples/apps/copilot-chat-app/webapp/src/libs/useGraph.ts
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                }
+            });
+
+            if (usersToLoad.length > 0) {
+                await makeBatchGetUsersRequest(usersToLoad, loadedUsers, usersToRetry);
+
+                // Retry any users that failed with transient (5xx) errors up to 3 times
+                while (usersToRetry.length > 0 && retries <= MAX_RETRIES) {
+                    console.log(`Retrying batch request  ${retries}/${MAX_RETRIES}`);
```
dehoward, line 58: nit: if you're in VSCode then Prettier should've fixed this by default based on our repo's settings. might be worth checking to see if you have the extension installed?

dehoward, line 58: nit: double space after "request"

dehoward (approved):

PR #1914, opened 2023-07-09
.Net: Cleanup from Resharper analysis

jsboige:
### Motivation and Context

 It helps clean the code up
 Better adhere to code standards
 A large range of code areas are affected, but all changes should be neutral
  7. If it fixes an open issue, please link to the issue here.

### Description

Those are a handful of suggestions coming out of a solution wide Resharper code inspection. out of the 5000+ warnings that Resharper oututs, not all are actionable (e.g when it thinks a public property should be made private because it sees no use of it), but there certainly at least another couple hundred low hanging fruits like the ones that this PR illustrates.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example19_Qdrant.cs
```diff
@@ -24,7 +24,7 @@ public static async Task RunAsync()
             .Build();
 
         Console.WriteLine("== Printing Collections in DB ==");
-        var collections = memoryStore.GetCollectionsAsync();
+        var collections = memoryStore.GetCollectionsAsync().ConfigureAwait(false);
```
shawncal, line 27: These ConfigureAwaits shouldn't be required as KernelSyntaxExamples is an app, not an SDK. I see they were being enforced because the .editorconfig that enabled the rule was in /dotnet. I moved it to /dotnet/src, so this project should now be free of its tyranny. Please sync run again and, if I'm not mistaken, these ConfigureAwaits can now be reverted!

dotnet/src/Connectors/Connectors.AI.OpenAI/Tokenizers/Settings/GPT3Settings.cs
```diff
@@ -17,12 +17,9 @@ internal static class GPT3Settings
...
     /// <summary>Lazy load the cached encoding table (encoder.json).</summary>
-    private static readonly Lazy<Dictionary<string, int>> s_encoder = new(() =>
-    {
-        return JsonSerializer.Deserialize<Dictionary<string, int>>(EmbeddedResource.ReadEncodingTable())
-               ?? throw new AIException(AIException.ErrorCodes.InvalidConfiguration,
-                   "Encoding table deserialization returned NULL");
-    });
+    private static readonly Lazy<Dictionary<string, int>> s_encoder = new(() => JsonSerializer.Deserialize<Dictionary<string, int>>(EmbeddedResource.ReadEncodingTable())
+                                                                                ?? throw new AIException(AIException.ErrorCodes.InvalidConfiguration,
+                                                                                    "Encoding table deserialization returned NULL"));
```
shawncal, line 22: The spacing seems to have gone wonky here. Far too wide.
Happy to get rid of the braces and return, but can we keep JsonSerialier... on a new line, and tab everything back?

dotnet/src/Connectors/Connectors.Memory.Qdrant/Diagnostics/ValidateExtensions.cs
```diff
@@ -8,12 +8,12 @@ namespace Microsoft.SemanticKernel.Connectors.Memory.Qdrant.Diagnostics;
 
 internal static class ValidateExtensions
 {
-    public static void Validate(IValidatable target)
+    public static void Validate(IValidatable? target)
     {
         target?.Validate();
     }
```
shawncal, line 14: We may want to look at this one more closely. Why would we want to pass in something null to validate, then simply ignore if it's null? If anything, I'd think that target being null would be a problem worthy of failing the validator.

Perhaps better to remove the nullable on both places, instead of adding to both places?

dotnet/src/Connectors/Connectors.Memory.Qdrant/Diagnostics/ValidateExtensions.cs
```diff
@@ -8,12 +8,12 @@ namespace Microsoft.SemanticKernel.Connectors.Memory.Qdrant.Diagnostics;
...
 
-    public static void Validate(params IValidatable[] targets)
+    public static void Validate(IEnumerable<IValidatable>? targets)
```
shawncal, line 16: This actually changes the calling semantics, no? What resharper rule triggered this? We may want to disable it.

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
```diff
@@ -48,7 +48,7 @@ public static class TextCompletionExtensions
     {
         var completionResults = textCompletion.GetStreamingCompletionsAsync(text, requestSettings, cancellationToken);
 
-        await foreach (var completionResult in completionResults)
+        await foreach (var completionResult in completionResults.WithCancellation(cancellationToken))
```
shawncal, line 51: Not needed -- the implementations use [EnumeratorCancellation] attibute, which does the same thing. Perhaps we should squash this resharper rule.

dotnet/src/Skills/Skills.Grpc/Extensions/GrpcOperationExtensions.cs
```diff
@@ -21,13 +21,13 @@ internal static class GrpcOperationExtensions
...
-
-        // Register the "payload" parameter to be used as gRPC operation request message.
-        parameters.Add(new ParameterView(GrpcOperation.PayloadArgumentName, "gRPC request message.", string.Empty));
+        var parameters = new List<ParameterView>
+        {
+            // Register the "address" parameter so that it's possible to override it if needed.
+            new ParameterView(GrpcOperation.AddressArgumentName, "Address for gRPC channel to use.", string.Empty),
+            // Register the "payload" parameter to be used as gRPC operation request message.
+            new ParameterView(GrpcOperation.PayloadArgumentName, "gRPC request message.", string.Empty)
+        };
```
shawncal, line 30: Not a huge fan of the shorthand here, with the comments iline, and the fact that the param values themselves are method calls... it's a little cluttered. Is this a r# rule? I think, at minimum, a newline between the first value and the second comment would help.

shawncal (changes requested):
Thanks for doing this! Lots of good, helpful cleanup.
Just a few minor comments and questions.

shawncal:
Note: samples/dotnet/FileCompression/ZipFileCompressor.cs is gone, and GitHub's requiring a command-line merge.

dotnet/samples/KernelSyntaxExamples/Example19_Qdrant.cs
jsboige: reverted with [that commit](https://github.com/microsoft/semantic-kernel/pull/1914/commits/bfeeef7ff9cf6c18e312fb833cbca5c7951f2138)

dotnet/src/Connectors/Connectors.AI.OpenAI/Tokenizers/Settings/GPT3Settings.cs
jsboige: reworked the indentation with [this commit](https://github.com/microsoft/semantic-kernel/commit/f6b0f079bc335f90677e5d003ed77e2669ff27a7).

dotnet/src/Connectors/Connectors.Memory.Qdrant/Diagnostics/ValidateExtensions.cs
jsboige: reverted in [that commit](https://github.com/microsoft/semantic-kernel/commit/8b0eb80f816715c5aecb81d9ec56457016c5c82c).

dotnet/src/Connectors/Connectors.Memory.Qdrant/Diagnostics/ValidateExtensions.cs
jsboige: reverted in [that commit](https://github.com/microsoft/semantic-kernel/commit/8b0eb80f816715c5aecb81d9ec56457016c5c82c).

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/TextCompletionExtensions.cs
jsboige: Reverted in [that commit](https://github.com/microsoft/semantic-kernel/pull/1914/commits/07d24c03677da3ebbc1df463f7612a95835c7e4d)

dotnet/src/Skills/Skills.Grpc/Extensions/GrpcOperationExtensions.cs
jsboige: Reworked in [that commit](https://github.com/microsoft/semantic-kernel/pull/1914/commits/831905fba119720d71021337e14091de99dd2d05)

shawncal (approved):

PR #1941, opened 2023-07-11
Python: Milvus memory connector

filip-halt:
This change brings in Milvus to the semantic kernel connectors.


filip-halt:
@fzliu

filip-halt:
@microsoft-github-policy-service agree company="Zilliz"

tawalke:
@filip-halt Following up on your offline note. Looks like it's already assigned for review. 

@alexchaomander for visibility.

python/semantic_kernel/connectors/memory/milvus/milvus_memory_store.py
```diff
@@ -0,0 +1,469 @@
...
+from pymilvus.milvus_client import milvus_client
+from semantic_kernel.memory.memory_record import MemoryRecord
+from semantic_kernel.memory.memory_store_base import MemoryStoreBase
+from semantic_kernel.utils.null_logger import NullLogger
+
+
+def memoryrecord_to_milvus_dict(mem: MemoryRecord) -> dict:
+    """Convert a memoryrecord into a dict.
+    Args:
+        mem (MemoryRecord): MemoryRecrod to convert.
```
awharrison-28, line 16:
```suggestion
        mem (MemoryRecord): MemoryRecord to convert.
```

python/tests/integration/connectors/memory/test_milvus.py
```diff
@@ -0,0 +1,216 @@
...
+from semantic_kernel.connectors.memory.milvus import MilvusMemoryStore
+from semantic_kernel.memory.memory_record import MemoryRecord
+
+try:
+    from milvus import default_server  # noqa: F401
+
+    default_server.start()
+    URI = "http://127.0.0.1:" + str(default_server.listen_port)
+    TOKEN = None
+    milvus_installed = True
```
awharrison-28, line 15: If the server is already running on the machine, milvus_installed will resolve as False. So running the tests twice on the same machine will be successful on the first run, but will skip on the second. I'd recommend some test teardown logic. Test-scoped is easier, but if the server setup takes too long, module scoped could work as well.

awharrison-28 (changes requested):

python/tests/integration/connectors/memory/test_milvus.py
filip-halt: Good point,resolved

awharrison-28:
Hi @filip-halt the code looks good, but I'm hitting errors with the integration tests where the default server is refusing to start due to communication timeouts. I'm seeing the issue standalone as well. I'm not sure if it's just a me problem, or if there's a bug with the default server - this wasn't an issue for me last week. Are you able to run these locally today?

python/semantic_kernel/connectors/memory/milvus/milvus_memory_store.py
```diff
@@ -0,0 +1,470 @@
...
+
+        Returns:
+            List[str]: List of collection names.
+        """
+        return self._client.list_collections()
+
+    async def delete_collection_async(
+        self, collection_name: str = "", all: bool = False
+    ) -> None:
+        """Delete the speficied collection.
```
awharrison-28, line 154:
```suggestion
        """Delete the specified collection.
```

filip-halt:
I would also like to know what kind of error is being seen? I just tested locally and everything seems to be running fine. Did the error happen on a first runthrough or after a few?

awharrison-28:
> I would also like to know what kind of error is being seen? I just tested locally and everything seems to be running fine. Did the error happen on a first runthrough or after a few?

Hmmm my machine might be in a bad state. I'll have someone else test locally. 

I don't believe I had the errors on the first runthrough, but definitely every run after that. 
image
image

all from the default_server.start() call.

filip-halt:
@awharrison-28  I put the call to preclean the service before running. This might help with your errors.

awharrison-28 (approved):
Looks like I had an orphan process hogging the default server. Restarting my machine did the trick :)

python/semantic_kernel/connectors/memory/milvus/milvus_memory_store.py
```diff
@@ -0,0 +1,470 @@
...
+    return MemoryRecord(
+        is_reference=milvus_dict.get("is_reference", None),
+        external_source_name=milvus_dict.get("external_source_name", None),
+        id=milvus_dict.get("id", None),
+        description=milvus_dict.get("description", None),
+        text=milvus_dict.get("text", None),
+        additional_metadata=milvus_dict.get("additional_metadata", None),
+        embedding=embedding,
+        key=milvus_dict.get("key", None),
+        timestamp=milvus_dict.get("timestamp", None),
```
roji, line 53: What representation actually gets saved to Milvus, given that it [doesn't seem to support](https://milvus.io/docs/create_collection.md) any sort of date/time type?

roji (commented):
@awharrison-28 hey, I'm working on a .NET version of the Milvus connector - see #2694. Below are some questions that came up while implementing this, can you give these a quick look? It would be great if we can more or less align across to connectors to ensure a uniform experience etc.

python/semantic_kernel/connectors/memory/milvus/milvus_memory_store.py
awharrison-28: my understanding is that timestamp resolves to a string

python/semantic_kernel/connectors/memory/milvus/milvus_memory_store.py
```diff
@@ -0,0 +1,470 @@
...
+        self._metric_cache[collection_name] = search_type
+
+        cleaned_results = []
+
+        if with_embeddings:
+            ids = []
+
+        # Clean up results, filter and get ids for fetch
+        for x in results:
+            if min_relevance_score is not None and x["distance"] < min_relevance_score:
```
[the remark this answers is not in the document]
awharrison-28: Thank you for raising the issue here, I was not aware.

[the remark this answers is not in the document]
roji: Update: the recently-released Milvus 2.3 ([release notes](https://github.com/milvus-io/milvus/releases/tag/v2.3.0)) has a new feature - [range search](https://milvus.io/docs/within_range.md#Conduct-a-Range-Search) - which allows doing this. This would be considerably better than the current client-side implementation.

PR #1947, opened 2023-07-11
Removing usernames from webapi stores

teresaqhoang:
### Motivation and Context

This PR removes all instances of saved usernames from backend memory stores to alleviate any PII concerns. We now only store user Id, which is used in the UI to query user data and display usernames.

Displaying usernames follow the same convention set in this PR: https://github.com/microsoft/semantic-kernel/pull/1899, where the active user only has access to user info within their own tenant, so any user outside of the tenant will be displayed as "External User".
image

### Contribution Checklist

### DIFF
--- a/samples/apps/copilot-chat-app/webapi/CopilotChat/Models/ChatMessage.cs
+++ b/samples/apps/copilot-chat-app/webapi/CopilotChat/Models/ChatMessage.cs
@@ -68,12 +68,6 @@ public class ChatMessage : IStorageEntity
     public string UserId { get; set; }
 
     /// <summary>
-    /// Name of the user who sent this message.
-    /// </summary>
-    [JsonPropertyName("userName")]
-    public string UserName { get; set; }
-
-    /// <summary>
     /// Id of the chat this message belongs to.
     /// </summary>
     [JsonPropertyName("chatId")]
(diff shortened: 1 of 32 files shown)


samples/apps/copilot-chat-app/webapp/src/redux/features/users/UsersState.ts
```diff
@@ -0,0 +1,22 @@
...
+    users: Users;
+}
+
+export const initialState: UsersState = {
+    users: {
+        bot: Constants.bot.profile,
+    },
+};
+
+export type Users = Record<string, UserData>;
```
dehoward, line 15: from previous PR: this should be `UserData | undefined` if some IDs won't have populated objects during loading

samples/apps/copilot-chat-app/webapp/src/libs/useGraph.ts
```diff
@@ -0,0 +1,158 @@
...
+        try {
+            // Filter user Ids list to optimize fetch
+            userIds.forEach((userId) => {
+                const ids = userId.split('.');
+                const objectId = ids[0]; // Unique GUID assigned to each user in their home tenant
+                const tenantId = ids[1]; // Home tenant id
+
+                // Active user can only access user data within their own tenant
+                // Mark chat users outside of tenant as External
+                if (activeUserInfo && tenantId !== activeUserInfo.id.split('.')[1]) {
```
dehoward, line 40: nit: can we save this call to a variable so we don't have to do the .split() each time?

samples/apps/copilot-chat-app/webapp/src/libs/useGraph.ts
```diff
@@ -0,0 +1,158 @@
...
+                    }
+                }
+            });
+
+            if (usersToLoad.length > 0) {
+                await makeBatchGetUsersRequest(usersToLoad, loadedUsers, usersToRetry);
+
+                // Retry any users that failed with transient (5xx) errors up to 3 times
+                while (usersToRetry.length > 0 && retries <= MAX_RETRIES) {
+                    console.log(`Retrying batch request  ${retries}/${MAX_RETRIES}`);
```
dehoward, line 58: nit: double space after "request"

samples/apps/copilot-chat-app/webapi/CopilotChat/Models/ChatMessage.cs
```diff
@@ -157,7 +149,7 @@ public string ToFormattedString()
             content = $"Sent a file named \"{documentDetails?.Name}\" with a size of {documentDetails?.Size}.";
         }
 
-        return $"[{this.Timestamp.ToString("G", CultureInfo.CurrentCulture)}] {this.UserName}: {content}";
+        return $"[{this.Timestamp.ToString("G", CultureInfo.CurrentCulture)}]: {content}";
```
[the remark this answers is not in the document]
TaoChenOSU: Or we can send a mapping of user id to username with the request to the backend.

shawncal:
@teresaqhoang Closing as no longer in this repo!

PR #1955, opened 2023-07-11
Copilot Chat: fix regression of missing bot name buttons

dehoward:
fixing regression introduced from a change (https://github.com/microsoft/semantic-kernel/pull/1307/commits/36385888a993e3503015b700d4fc1b95c7d2cd1e) in #1307.

before:
image

after:
gif


samples/apps/copilot-chat-app/webapp/src/components/chat/ChatWindow.tsx
```diff
@@ -138,14 +143,18 @@ export const ChatWindow: React.FC = () => {
         setTitle(data.value);
     };
 
+    const handleSave = () => {
+        onSave().catch((e: any) => {
```
TaoChenOSU, line 147: Should we reset the `chatName` and `isEditing` if `onSave` fails? This is probably an old issue.

dehoward: not necessarily, since the user should be able to retry the operation in case of a failure and setting `isEditing` to false would close the `Popover`.

gitri-ms (approved):

TaoChenOSU (approved):

PR #1957, opened 2023-07-11
Java: Fix invoking Action planner not copying the invocation state

johnoliver:
- Fix bug with ActionPlanner not copying the invocation state when the plan is invoked
- Improve the pretty printing of plans
- Cleanup ActionPlan returned by completion before parsing it
- Allow ActionPlanner to invoke native functions


java/semantickernel-extensions-parent/actionplanner-extension/src/main/java/com/microsoft/semantickernel/planner/actionplanner/ActionPlanner.java
```diff
@@ -36,6 +40,8 @@
 public class ActionPlanner {
     private static final Logger LOGGER = LoggerFactory.getLogger(ActionPlanner.class);
 
+    private static final Pattern CLEAN_PLAN =
```
dsgrieve, line 43: Would be helpful to have comments on patterns. It looks to me like this trying to get at the inner braces in something like {{$input}} (group 1 would be "{$input}"). But when I play with it in jshell, group one is "{{$input}}".

dsgrieve (approved):

markwallace-microsoft (approved):

markwallace-microsoft (approved):

PR #1959, opened 2023-07-11
.Net: Set default max tokens for SequentialPlanner

shawncal:
Setting SequentialPlanner's max tokens to 512 in cases where it's left null. Without this, plans are never completed and cannot be parsed.

Required to fix dotnet integration tests.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
+++ b/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
@@ -42,7 +42,7 @@ public sealed class SequentialPlanner
             skillName: RestrictedSkillName,
             description: "Given a request or command or goal generate a step by step plan to " +
                          "fulfill the request using functions. This ability is also known as decision making and function flow",
-            maxTokens: this.Config.MaxTokens,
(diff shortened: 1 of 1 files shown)


awharrison-28 (dismissed):
Non-blocking, but the smallest token window I'm aware of for a model capable of generating a sequential plan is 2k. 

I would suggest making the default either 1024 or 2048

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
```diff
@@ -42,7 +42,7 @@ public sealed class SequentialPlanner
             skillName: RestrictedSkillName,
             description: "Given a request or command or goal generate a step by step plan to " +
                          "fulfill the request using functions. This ability is also known as decision making and function flow",
-            maxTokens: this.Config.MaxTokens,
+            maxTokens: this.Config.MaxTokens ?? 512,
```
awharrison-28, line 45:
```suggestion
            maxTokens: this.Config.MaxTokens ?? 1024,
```

awharrison-28 (approved):

PR #1999, opened 2023-07-14
Copilot Chat: Java: add Example15_MemorySkill

dsgrieve:
### Motivation and Context

Adding Example15_MemorySkill
### Description

Added java/samples/sample-code/src/main/java/com/microsoft/semantickernel/syntaxexamples/Example15_MemorySkill.java
Made changes necessary to get the example to work. 

### Contribution Checklist


java/semantickernel-core-skills/src/main/java/com/microsoft/semantickernel/coreskills/TextMemorySkill.java
```diff
@@ -97,102 +108,115 @@ public Mono<String> retrieveAsync(
...
-    [SKFunctionContextParameter(Name = CollectionParam, Description = "Memories collection to search", DefaultValue = DefaultCollection)]
-    [SKFunctionContextParameter(Name = RelevanceParam, Description = "The relevance score, from 0.0 to 1.0, where 1.0 means perfect match",
-        DefaultValue = DefaultRelevance)]
-    [SKFunctionContextParameter(Name = LimitParam, Description = "The maximum number of relevant memories to recall", DefaultValue = DefaultLimit)]
-    public string Recall(string text, SKContext context)
+    /// <param name="text">The input text to find related memories for.</param>
+    /// <param name="collection">Memories collection to search.</param>
+    /// <param name="relevance">The relevance score, from 0.0 to 1.0, where 1.0 means perfect match.</param>
+    /// <param name="limit">The maximum number of relevant memories to recall.</param>
+    /// <param name="context">Contains the memory to search.</param>
```
johnoliver, line 156: can convert to javadoc

java/semantickernel-core/src/main/java/com/microsoft/semantickernel/orchestration/NativeSKFunction.java
```diff
@@ -229,6 +235,11 @@ private static SKNativeTask<SKContext> getFunction(Method method, Object instanc
 
                 return mono.map(
                         it -> {
+                            if (it instanceof Iterable) {
+                                // Handle return from things like Mono<List<?>>
+                                // from {{function 'input'}} as part of the prompt.
+                                it = ((Iterable<?>) it).iterator().next();
```
johnoliver, line 241: Think taking only the first element from an iterable may be non-obvious behaviour for users

dsgrieve: Thanks for reminding me. I meant to file an issue for this and will do straight away. 

The other workaround, at least to fit into the current context API, was to change `context.update((String) it)` to `context.update(String.valueOf(it))`, but that would be very strange indeed. Ultimately, I think the fix needs to come in SKContext itself.

dsgrieve: https://github.com/microsoft/semantic-kernel/issues/2190

java/semantickernel-core/src/main/java/com/microsoft/semantickernel/orchestration/NativeSKFunction.java
```diff
@@ -346,7 +357,54 @@ private static String getArgumentValue(
...
+                                + " expected "
+                                + type.getSimpleName()
+                                + " but got "
+                                + arg);
+            }
+        }
+        return value;
+    }
+
+    private static final Map<Class<?>, Function<String, ?>> converters = new HashMap<>();
```
dsgrieve, line 390: @johnoliver You had a question about these converters having Boolean and boolean (for example). The reason for this is that the user might put `@SKFunctionParameters(type=boolean.class)`.  The other option would be to box the return value from the annotation's `type()` method, but you'd still end up with some sort of `if (type==boolean.class) type = Boolean.class` logic.

Another question you had was whether we could add Iterable to the converters and get rid of this bit of code
```
                            if (it instanceof Iterable) {
                                // Handle return from things like Mono<List<?>>
                                // from {{function 'input'}} as part of the prompt.
                                it = ((Iterable<?>) it).iterator().next();
                            }
```
The converters are for converting a parameter value according to the `type` in the SKFunctionParameters annotation. I don't think we allow Iterable<?> as a parameter to a native function, do we?

johnoliver (approved):

PR #2000, opened 2023-07-14
.Net: Add custom PromptConfig to Stepwise Planner

akshaykokane:
### Motivation and Context

In case of custom prompt, allow sending custom prompt config. This is useful for custom prompt to Stepwise Planner having additional parameters.

### Description

This allow stepwise planner to be created with custom prompt and custom prompt config

### Contribution Checklist


dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -37,7 +37,8 @@ public class StepwisePlanner
     public StepwisePlanner(
         IKernel kernel,
         StepwisePlannerConfig? config = null,
-        string? prompt = null)
+        string? prompt = null.
+        string? config = null)
```
dmytrostruk, line 41: 1. It won't compile. There is dot instead of comma in line 40, and also there are two `config` parameters in this method.
2. We should think whether to accept `string` or `PromptTemplateConfig` for prompt configuration. If we use strong type instead of string, it will reduce the number of errors in string parsing logic.

akshaykokane: Yes, I agree. Passing PromptTemplateConfig, could be good idea, instead of string to reduce the error. Updated the PR

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -37,18 +37,19 @@ public class StepwisePlanner
...
         Verify.NotNull(kernel);
         this._kernel = kernel;
 
         this.Config = config ?? new();
         this.Config.ExcludedSkills.Add(RestrictedSkillName);
 
-        var promptConfig = new PromptTemplateConfig();
+        var promptConfig = promptUserConfig ?? new PromptTemplateConfig();
         var promptTemplate = prompt ?? EmbeddedResource.Read("Skills.StepwiseStep.skprompt.txt");
         string promptConfigString = EmbeddedResource.Read("Skills.StepwiseStep.config.json");
```
shawncal, line 51: This `EmbeddedResource` should only be read from file if the `promptUserConfig` arg was not passed in.

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -37,18 +37,19 @@ public class StepwisePlanner
...
-        if (!string.IsNullOrEmpty(promptConfigString))
+        if (!string.IsNullOrEmpty(promptConfigString) && promptUserConfig != null)
```
shawncal, line 52: ` && promptUserConfig != null`
I think this is inverse of what we'd want, no? If promptUserConfig *IS* null, then we set promptConfig to the value loaded from file.

shawncal (changes requested):

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
akshaykokane: Make sense. Updated

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
akshaykokane: My bad. Fixed

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -37,22 +37,27 @@ public class StepwisePlanner
...
         {
-            promptConfig = PromptTemplateConfig.FromJson(promptConfigString);
+            string promptConfigString = EmbeddedResource.Read("Skills.StepwiseStep.config.json");
+            if (!string.IsNullOrEmpty(promptConfigString))
+            {
+                promptConfig = PromptTemplateConfig.FromJson(promptConfigString);
+            }
         }
-
+
```
shawncal, line 60:
```suggestion

```
Whitespace formatting

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -37,22 +37,26 @@ public class StepwisePlanner
...
+        if (promptUserConfig == null)
         {
-            promptConfig = PromptTemplateConfig.FromJson(promptConfigString);
+            string promptConfigString = EmbeddedResource.Read("Skills.StepwiseStep.config.json");
+            if (!string.IsNullOrEmpty(promptConfigString))
+            {
+                promptConfig = PromptTemplateConfig.FromJson(promptConfigString);
+            }
         }
```
shawncal, line 59:
```suggestion
        }

```

shawncal (approved):

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -37,20 +37,25 @@ public class StepwisePlanner
     public StepwisePlanner(
         IKernel kernel,
         StepwisePlannerConfig? config = null,
-        string? prompt = null)
+        string? prompt = null,
+        PromptTemplateConfig? promptUserConfig = null)
```
shawncal, line 41: `promptUserConfig` requires a matching XML tag in the method comments above. (PR checks are blocking on this)

shawncal (changes requested):

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
akshaykokane: Thanks. Added

shawncal (approved):

PR #2007, opened 2023-07-15
.Net: Simplify chatgpt plugin import method names

craigomatic:
### Description
Several extension methods were previously supplied that allow loading plugins from remote URI, local files, streams and from a given directory. They also were specific to OpenAPI and ChatGPT.

This PR marks those extensions as Obsolete (KernelChatGptPluginExtensions.cs and KernelOpenApiExtensions.cs) and supplies a single `ImportAIPluginAsync` in a new extension file which expects a URI or a Stream to import from. For resources of any other form, the caller is expected to carry the burden of getting them to a stream or a URI.

A single integration test has been added that imports the Klarna AI Plugin using both it's ChatGPT representation (https://www.klarna.com/.well-known/ai-plugin.json) and the OpenAPI representation (https://www.klarna.com/us/shopping/public/openai/v0/api-docs/)

### Contribution Checklist


dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,439 @@
...
+                    cancellationToken: cancellationToken)
+                .ConfigureAwait(false);
+        }
+
+        //assume OpenAPI v3
+        var parser = new OpenApiDocumentParser(kernel.Log);
+
+        using (var documentStream = new MemoryStream(System.Text.Encoding.UTF8.GetBytes(pluginJson)))
+        {
+            var operations = await parser.ParseAsync(documentStream, executionParameters?.IgnoreNonCompliantErrors ?? true, cancellationToken).ConfigureAwait(false);
```
SergeyMenshykh, line 72: Allowing the ignoring of non-compliant errors by default will prevent SK consumer code developers from being aware of the non-compliant OpenAPI plugin they may want to work with. As a result, some issues may pop up later during the plugin's execution phase rather than the import phase, when it may be a little bit tricky to debug without knowing that the plugin does not follow the OpenAPI specification. 

Disabling it by default, on the other hand, will allow SK consuming code developers to make an explicit decision whether to use a particular plugin or not after a failure during the import phase, it will also provide an additional context that will make troubleshooting easier if the plugin misbehaves during the execution phase.

craigomatic: Ah good catch, I originally considered flipping this then forgot to unwind. Switched back to false.

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,439 @@
...
```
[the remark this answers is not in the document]
craigomatic: Modified - I went with a private method (which is a copy/paste from RegisterOpenApiSkillAsync). The prior extension methods have been marked as Obsolete

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,411 @@
...
+                .ConfigureAwait(false);
+        }
+
+        return await LoadSkill(
+            kernel,
+            skillName,
+            executionParameters,
+            internalHttpClient,
+            pluginJson,
+            cancellationToken).ConfigureAwait(false);
```
SergeyMenshykh, line 73: The code starting on line 56 and ending at line 73 is repeated three times in this file. Consider moving it to a separate private method to avoid duplication.

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,411 @@
...
+                    kernel.Log.LogWarning(ex, "Something went wrong while rendering the Rest function. Function: {0}.{1}. Error: {2}",
+                        skillName, operation.Id, ex.Message);
+                }
+            }
+
+            return skill;
+        }
+    }
+
+    private static async Task<string> LoadJsonFromUri(
```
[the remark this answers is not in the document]
craigomatic: Renamed as suggested

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
craigomatic: Cleaned up as suggested

SergeyMenshykh (dismissed):

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,406 @@
...
+    /// <param name="skillName">Skill name.</param>
+    /// <param name="filePath">The file path to the AI Plugin</param>
+    /// <param name="executionParameters">Skill execution parameters.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns></returns>
+    public static async Task<IDictionary<string, ISKFunction>> ImportAIPluginAsync(
+        this IKernel kernel,
+        string skillName,
+        string filePath,
+        OpenApiSkillExecutionParameters? executionParameters = null,
```
shawncal, line 39: I think we'll need an (optional) `HttpClient? httpClient = null` arg here.
In the public methods here, we always load the HttpClientProvider one, but there's a high likelihood that some of these AI plugin APIs will require custom bearer tokens and such. We should allow optional passing, and if null, fall back to what you have below.

@SergeyMenshykh to confirm

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,406 @@
...
+    /// <summary>
+    /// Imports an AI plugin that is exposed as an OpenAPI v3 endpoint or through OpenAI's ChatGPT format.
+    /// </summary>
+    /// <param name="kernel">Semantic Kernel instance.</param>
+    /// <param name="skillName">Skill name.</param>
+    /// <param name="filePath">The file path to the AI Plugin</param>
+    /// <param name="executionParameters">Skill execution parameters.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns></returns>
+    public static async Task<IDictionary<string, ISKFunction>> ImportAIPluginAsync(
```
shawncal, line 35: Love the unified call. This is fantastic!

One thing that comes to mind though is, in the planned renaming of all "Skills" to "Plugins", these Kernel overloads will appear alongside (possibly conflict with) any other "Kernel.ImportPlugin" calls load what are currently skills with native or semantic functions.

Let's discuss offline. This is certainly on the right track, and will make the development story a lot cleaner!

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,406 @@
...
+        Uri uri,
+        OpenApiSkillExecutionParameters? executionParameters = null,
+        CancellationToken cancellationToken = default)
+    {
+        Verify.NotNull(kernel);
+        Verify.ValidSkillName(skillName);
+
+#pragma warning disable CA2000 // Dispose objects before losing scope. No need to dispose the Http client here. It can either be an internal client using NonDisposableHttpClientHandler or an external client managed by the calling code, which should handle its disposal.
+        var internalHttpClient = HttpClientProvider.GetHttpClient(kernel.Config, executionParameters?.HttpClient, kernel.Log);
+#pragma warning restore CA2000 // Dispose objects before losing scope. No need to dispose the Http client here. It can either be an internal client using NonDisposableHttpClientHandler or an external client managed by the calling code, which should handle its disposal.
```
shawncal, line 86: Don't need the duplicate commentary on the `#... restore` calls

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,406 @@
...
+    /// </summary>
+    /// <param name="kernel">Semantic Kernel instance.</param>
+    /// <param name="skillName">Skill name.</param>
+    /// <param name="uri">A local or remote URI referencing the AI Plugin</param>
+    /// <param name="executionParameters">Skill execution parameters.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns></returns>
+    public static async Task<IDictionary<string, ISKFunction>> ImportAIPluginAsync(
+        this IKernel kernel,
+        string skillName,
```
shawncal, line 76: Should skillName be a required arg? I would expect this can be grabbed from the "name_for_model" property in the plugin schema, if not provided by a user.

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
craigomatic: I believe this is passed in via the OpenApiSkillExecutionParameters: https://github.com/microsoft/semantic-kernel/blob/adc59234aa6d1f91fe65376b3a9a5074b4f0dae9/dotnet/src/Skills/Skills.OpenAPI/Extensions/OpenApiSkillExecutionParameters.cs#L19

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
craigomatic: Fixed!

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
craigomatic: Seems reasonable - what happens if a dev imports a plugin without a name (which we supply via name_for_model or maybe the Title if it's an OpenAPI endpoint we are loading), then later wants to reference the plugin by name? 

They would have the object reference given the current method signature, but not the assigned name unless we modify the signature

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
craigomatic: @SergeyMenshykh anything to add here?

dotnet/src/IntegrationTests/IntegrationTests.csproj
```diff
@@ -11,6 +11,10 @@
     <UserSecretsId>b7762d10-e29b-4bb1-8b74-b6d69a667dd4</UserSecretsId>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Remove="Plugins\instacart-ai-plugin.json" />
```
glahaye, line 15: Why is that here?

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -0,0 +1,406 @@
...
+{
+    /// <summary>
+    /// Imports an AI plugin that is exposed as an OpenAPI v3 endpoint or through OpenAI's ChatGPT format.
+    /// </summary>
+    /// <param name="kernel">Semantic Kernel instance.</param>
+    /// <param name="skillName">Skill name.</param>
+    /// <param name="filePath">The file path to the AI Plugin</param>
+    /// <param name="executionParameters">Skill execution parameters.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns></returns>
```
glahaye, line 34: Nit: provide description of returned value

craigomatic: Done!

dotnet/src/IntegrationTests/IntegrationTests.csproj
craigomatic: It's used during the test that we can successfully load the plugin from a stream, I think a byproduct of how VS handles Content with copy-always:

image

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelAIPluginExtensions.cs
craigomatic: Spoke offline, will revisit this in a separate issue.

shawncal (approved):

shawncal (approved):

dotnet/src/IntegrationTests/Plugins/PluginTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            // Act
+            var result = await skill[functionName].InvokeAsync(new SKContext(contextVariables));
+
+            // Assert
+            Assert.False(result.ErrorOccurred);
+        }
+    }
+
+    [Theory]
+    [InlineData(".\\Plugins\\instacart-ai-plugin.json",
```
shawncal, line 116:
```suggestion
    [InlineData("Plugins/instacart-ai-plugin.json",
```

shawncal (approved):

PR #2016, opened 2023-07-17
.Net: Merge feature/oobabooga branch to main

dmytrostruk:
### Motivation and Context

Merge [feature/oobabooga](https://github.com/microsoft/semantic-kernel/tree/feature/oobabooga) branch to `main` with [Oobabooga](https://github.com/oobabooga/text-generation-webui) AI Connector functionality.
Functionality verified with unit and integration testing.

### Description

From original PR (https://github.com/microsoft/semantic-kernel/pull/1357):

> This PR adds to the solution a project similar to HuggingFace connectors project, and an additional integration test also similar to HuggingFace connector's
The code for the connector was based on the existing HuggingFace's, with a couple improvements (e.g. using web sockets for streaming API)

### Contribution Checklist

Co-authored-by: Jean-Sylvain Boige <jsboige@gmail.com>


dmytrostruk:
Hi @jsboige !
Just want to inform you, that I removed some long-running unit tests, since their duration is significantly higher comparing to existing unit tests (see screenshots below). This will lead to increased duration of build and release pipelines, which we want to avoid. Unit tests should be fast and lightweight, so we can consider moving long-running tests to integration tests in follow-up PRs. Thanks!

Before:
unit_tests_1

After:
unit_tests_2

jsboige:
>Just want to inform you, that I removed some long-running unit tests, since their duration is significantly higher comparing to existing unit tests (see screenshots below).

@dmytrostruk, @shawncal,  no pb at all. Some of those tests were more like stress tests that unit tests, and there's certainly no need to hammer the mock web socket server and connector with tens of thousands of web socket messages on each build as part of ci.

2 remarks :

- Maybe it would have been useful to keep and simply trim down the 2 tests that were testing the web socket pooling capabilities, with and without a semaphore, since that capability does not seem unit-tested anymore. The one without semaphore tested for 1000 multi-message stream calls, and the one with semaphore tested for 10000 such calls.
- When I was doing my tests of alternative implementation of concurrent consumption of streams, it seemed like BroadcastBlocks had better performance than Channels (by the way I'm not sure to understand how the later actually support that scenario, but that's pretty much what the last unit test that you did keep exhibits). Now I guess the most important is to properly support concurrent calls to the connector, rather that concurrent consumption of a single connector stream, and I can't think of a scenario for now where we would need massive concurrent observers of a single stream, but for what it's worth, I did notice that Channels performances degraded pretty badly when increasing the number of concurrent observers for each one of concurrent multi-streams, whereas BroadcastBlocks did not take that much of a hit. But then BroadcastBlocks had their own issues, which I wasn't able to fix, so I guess the current implementation seems good enough for the current use cases. 

Well this is to say, concurrent observers to concurrent web socket based stream is trickier that it looks, I would have gone with BroadcastBlocks if I had not pushed past 1000s of concurrent calls that were necessary to exhibit some weird race condition, and the current implementation with Chanels seems to work well with massive concurrent calls, but I wouldn't advice increasing the number of concurrent consumers of the connector without doing some further investigations.

Overall I'm very glad that this PR finally made it to the main, thanks for that, and I'll be happy to contribute the chat completion complement to the connector in the following weeks.

PR #2026, opened 2023-07-17
Java: Move samples to new package; Add jbang support

brunoborges:
- Fixes the template of `conf.properties` to not have quotes;
- Moves samples to a sub package to avoid conflict when running `jbang`
- Adds `jbang` support to execute the samples from command line


java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/Config.java
```diff
@@ -64,9 +65,11 @@ public static OpenAIAsyncClient getClient() throws IOException {
     }
 
     private static Supplier<ClientSettings<?>> getOpenAIClientFromDefaultPropertiesLocations(ClientType type) {
+        LOGGER.info("Create supplier for settings of "+type);
         return () -> {
             try {
                 for (String location : DEFAULT_PROPERTIES_LOCATIONS) {
+                    LOGGER.info("Attempting config file at "+location);
```
dsgrieve, line 72: First info should be inside the Supplier body - tell me when the Supplier is called, not when the getOAICFDPL method is called. I think this should be debug, not info. 
Second log should be debug, for sure. Don't annoy me with what the code is doing unless I decide that. 
The one log.info I'd have is one that says what file you are using. Maybe that is already in AIProviderSettings.get* methods.

brunoborges: This code is part of the sample, not the library. The reason I went with `info` is because right now there is no `logging.properties` configuration file (or something similar). 

But this did raise a challenge with the current AISettings API design: once an instance of `ClientSettings` is available and `isValid()`, it is not possible to know where it was loaded from.

@johnoliver and @milderhc, should we reevaluate this?

brunoborges: There's actually an xml already. I missed that. Changes made.

johnoliver:
Can bump to latest version, also some test failures

java/samples/example.conf.properties
```diff
@@ -1,6 +1,6 @@
-client.openai.key: "OPEN_AI_KEY"
```
maxandersen, line 1: any reason why using non-conventional :'s instead of = ?

java/samples/example.conf.properties
```diff
@@ -1,6 +1,6 @@
...
-client.openai.organizationid: "OPEN_AI_ORGANIZATION_ID"
```
maxandersen, line 2: when I copy this file to `conf.properties` I still get this exception:

```
jbang sample-code/src/main/java/com/microsoft/semantickernel/samples/Example00_GettingStarted.java
Exception in thread "main" java.io.IOException: Unable to find OpenAI settings. Please provide an endpoint and organisation Id.
        at com.microsoft.semantickernel.samples.Config$ClientType.buildClient(Config.java:187)
        at com.microsoft.semantickernel.samples.Config$ClientType$1.getClient(Config.java:93)
        at com.microsoft.semantickernel.samples.Example00_GettingStarted.run(Example00_GettingStarted.java:75)
        at com.microsoft.semantickernel.samples.Example00_GettingStarted.main(Example00_GettingStarted.java:81)
       ```
       
whats the expected location or way to call it?

File: java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/Config.java [hunk 4]
```diff
@@ -25,7 +26,7 @@ public class Config {
 
     private static final List<String> DEFAULT_PROPERTIES_LOCATIONS = Arrays.asList(
             "java/samples/conf.properties",
-            System.getProperty("user.home") + "/.sk/conf.properties"
+            System.getProperty("user.home") + File.separator + ".sk" + File.separator + "conf.properties"
```
maxandersen, line 29: I also tried putting it in ~/.sk/conf.properties - gives same exception.

java/samples/example.conf.properties
maxandersen: found the cause - I had not specified organization id (since no other openai API I used *require* it). when specifying it things got better.

any reason why require specifying org id?

brunoborges: We noticed this bug yesterday where the organization id is truly not required, but in some cases the API will not consider the setting valid.

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/Config.java
brunoborges: Hey @maxandersen , did you fix this?

java/samples/example.conf.properties
brunoborges: Check this PR for the upcoming Settings feature: https://github.com/microsoft/semantic-kernel/pull/2115

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/Config.java
maxandersen: Yeah. It was just the missing org key.

markwallace-microsoft (approved):

PR #2045, opened 2023-07-18
.Net: SKFunction / Plan - Pre/Post Handlers (ADR + Impl Example)

### LINKED ISSUE #1714: Be able to Handle Function Calls: Phase 1
### IKernel Handlers - Notification

An IKernel.RunAsync Caller needs:

- Know and be notified before and after each function is executed with information it can get or change prior and after invocation.

  - Pre-Invocation

    - Get: Function details
    - Get: Prompt generated by the current Kernel `TemplateEngine` before calling the LLM
    - Get: Current settings used
    - Get: SKContext
      - Get: State of variables (args)

    - Set: Cancel execution
    - Set: Modify parameters (Context Variables)

  - Post-Invocation

    - Get: Function details
    - Get: Prompt generated by the current Kernel `TemplateEngine` before calling the LLM
    - Get: SKContext
      - Get: LLM Model Result (Tokens Usage, Stop Sequence, ...)
      - 
    - Set: Modify parameters (Context Variables)

RogerBarreto:
### Motivation and Context

Be able to add handlers to function executions in a way the caller can be notified and get information before and after an ISKFunction is executed (Including Plan Steps)

Resolves #1714 

### Description

This is WIP and there are still Open Questions pending to decide in the ADR.

This implementation adds 2 new public interface methods to an ISKFunction allowing setting Pre/Post handlers to it:

- ISKFunction.SetPreExecutionHandler
- ISKFunction.SetPostExecutionHandler

With this initial change allow scenarios like:

- Pre-Execution

    - Get: Prompt generated by the current Kernel `TemplateEngine` before calling the LLM
    - Get: SKContext

- Post-Execution
    - Get: SKContext

Added Plan specific methods to set handlers to all children steps at once.

- Plan.SetStepsPreExecutionHandler
- Plan.SetStepsPostExecutionHandler

## Out of scope of this PR.

- Using handlers to Set/Change data
- Using handlers to abort/cancel a function call

### Contribution Checklist


dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -382,6 +382,58 @@ public ISKFunction SetAIConfiguration(CompleteRequestSettings settings)
             : this.Function.SetAIConfiguration(settings);
     }
 
+    public ISKFunction SetPreExecutionHook(Func<SKContext, string, SKContext>? preHook)
```
RogerBarreto, line 385: Currently considering bringing those as extension methods instead. Thoughts?

dmytrostruk: Since hooks are optional parameters in `SKFunction` and `Plan`, I think the methods to set them should be optional as well, when you implement `ISKFunction` interface. Extension methods should help to achieve that.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -168,14 +171,23 @@ public sealed class SKFunction : ISKFunction, IDisposable
...
 
             try
             {
                 string renderedPrompt = await functionConfig.PromptTemplate.RenderAsync(context).ConfigureAwait(false);
+
+                // If there is a pre execution hook, call it
+                if (preExecutionHook is not null)
+                {
+                    // Allow the pre execution hook to update the context if needed
+                    context = preExecutionHook(context, renderedPrompt);
```
dmytrostruk, line 188: I think just `context` and `renderedPrompt` might not be enough to pass in hook. For example, in ADR you also mentioned about passing request settings, like temperature and so on. We can:
- Create dedicated class which will hold pre and post hook arguments needed to be passed to consumer. In this way it will be easier to extend list of arguments in the future.
- Maybe we can re-use `SKContext` for those purposes, by placing rendered prompt and other required arguments in there. Not 100% sure about this approach though, it's worth to brainstorm it.

Same applies for post hook.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -321,15 +340,15 @@ public string ToString(bool writeIndented)
 
     private static readonly JsonSerializerOptions s_toStringStandardSerialization = new();
     private static readonly JsonSerializerOptions s_toStringIndentedSerialization = new() { WriteIndented = true };
-    private Func<ITextCompletion?, CompleteRequestSettings?, SKContext, Task<SKContext>> _function;
+    private Func<ITextCompletion?, CompleteRequestSettings?, SKContext, Func<SKContext, string, SKContext>?, Func<SKContext, SKContext>?, Task<SKContext>> _function;
```
dmytrostruk, line 343: This delegate definition becomes too complicated. Maybe for pre and post hooks we can define new delegate type (in similar way how `Action` and `Func` are defined) and use it where needed. 
Something like `PreExecutionHook` delegate or `PreExecutionCallback` delegate.

dotnet/src/SemanticKernel/Planning/Plan.cs
RogerBarreto: Those can't be extension methods because both are setting internal fields in the SKFunction implementation, any thoughts?

rysweet:
commentary/feedback. love it and definitely need some way to look at the completed prompt before execution, especially when building compound skills. 

I am wondering if having a function hook is the right pattern vs the more typical (in .net) pattern of having an extension of the Builder that lets you look into the kernel before the *Async invocation?

dotnet/samples/KernelSyntaxExamples/Example53_FunctionHooks.cs
```diff
@@ -0,0 +1,79 @@
...
+            return Task.CompletedTask;
+        }
+
+        Task MyPostHook(PostExecutionContext executionContext)
+        {
+            Console.WriteLine($"Post Hook - Total Tokens: {executionContext.SKContext.ModelResults.First().GetOpenAITextResult().Usage.TotalTokens}");
+            return Task.CompletedTask;
+        }
+
+        excuseFunction.SetPreExecutionHook(MyPreHook);
```
lemillermicrosoft, line 67: What if my semantic function calls other semantic functions? Will the token usage be cumulative across everything in the three?

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -293,6 +304,32 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
         return this;
     }
 
+    private async Task<SKContext> CallPostExecutionHook(SKContext context)
```
lemillermicrosoft, line 307: Curious if we can just depend on the fact that a Plan is an ISKFunction and that things will inherit appropriately.

dotnet/samples/KernelSyntaxExamples/Example53_FunctionHooks.cs
RogerBarreto: @lemillermicrosoft In the PreExecutionHook the generated prompt will be the version that the template engine generates at that stage which by definition will be cumulative of all dependencies the template engine has to generate the prompt.

dotnet/src/SemanticKernel/Planning/Plan.cs
RogerBarreto: Nope, I started that way but it breaks down in those scenarios:

1. When Plans are not functions/steps. (Steps within the plan will not trigger the the hook)
2. When new steps are added to the plan. (Those will not have the hooks, so the caller would need to manage that)
3. When hooks are cancelled at the plan level I would have to manage the multiple references in the steps level (for each function individually)
4. Some more :)

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
```diff
@@ -80,4 +80,20 @@ public interface ISKFunction
...
     /// <returns>Self instance</returns>
     ISKFunction SetAIConfiguration(CompleteRequestSettings settings);
+
+    /// <summary>
+    /// Used for setting a pre-execution hook to a function.
+    /// </summary>
+    /// <remarks>Using more than once in the same function will override the previous pre-execution hook, avoid overriding when possible.</remarks>
+    /// <param name="preHook">Pre-hook delegate</param>
+    /// <returns>Self instance</returns>
+    HookRequest<PreExecutionContext> SetPreExecutionHook(ExecutionHook<PreExecutionContext> preHook);
```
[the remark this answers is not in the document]
RogerBarreto: That makes sense, using the `Handler` naming instead of `Hook`

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/PreExecutionContext.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public SKContext SKContext { get; }
+
+    /// <summary>
+    /// Prompt that was generated prior to sending to the LLM.
+    /// </summary>
+    /// <remarks>
+    /// May be null for native functions.
+    /// </remarks>
+    public string? Prompt { get; }
```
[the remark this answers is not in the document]
RogerBarreto: Currently (as is), having the nullable prompt property might work, but we can think in a more generic way, removing the Property `Prompt` and changing it to a method `T? GetPrompt<T>()` and/or `bool TryGetPrompt<T?>(out T? generatedPrompt)` which the caller can specify what type he is expecting.

What you think?


Regarding your suggestion I might missed something but not sure how we would make this simple from the implementation and caller perspective. 

Made some trials in the code and wasn't able to get it in a simple way without having to resort into adding some extra methods, checks and complexity.

AFAIK using the specialization approach there would have 2 potential outcomes:

1. The caller would need to check for the `IsSemantic` flag in the ISKFunction interface and explicitly cast the ExecutionContext type to the specific SemanticFunctionExecutionContext to get access to the prompt property... OR rely in an extra extension method that does this check and return the prompt.

2. The caller would have to know prior what are the functions type and setup a hook using the correct specialized `ExecutionContext` implementation.

dotnet/src/SemanticKernel/Planning/Plan.cs
RogerBarreto: @shawncal Agree, I think this decomposition of the TemplateEngine in a TemplatePlanner is something we need to address shortly. Using the Template as Plan, is another think we would need to think if by design the Hooks defined in the plan should be triggered for `TemplateGeneration` functions. But this could be something we can address at the Hooking set definition for `IPlan` interfaces.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/HookRequest.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Diagnostics;
+
+namespace Microsoft.SemanticKernel.SkillDefinition;
+
+/// <summary>
+/// Request to the execution hook. Can be used to cancel the execution.
+/// </summary>
+/// <typeparam name="TExecutionContext"></typeparam>
+public sealed class HookRequest<TExecutionContext>
```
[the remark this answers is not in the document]
RogerBarreto: Renamed to `HandlerRegistration`. 

Not sure about using `Function` name as those registrations can be made into `Plan` and potentially in the `Kernel` scope in the future.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -195,6 +232,24 @@ public ISKFunction SetAIConfiguration(CompleteRequestSettings settings)
         return this;
     }
 
+    /// <inheritdoc/>
+    public HookRequest<PreExecutionContext> SetPreExecutionHook(ExecutionHook<PreExecutionContext> preHook)
```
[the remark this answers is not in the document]
RogerBarreto: Today we only allow the Kernel to create sealed SemanticFunctions and NativeFunctions, if we make `SKFunction` a public abstract we allow any new SKFunction specialization to be created by external projects.

I personally don't see a problem here, but the original design was not allowing it. If we are ok with it I may proceed in this direction, which makes a lot of sense from OOP prespective.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -173,6 +179,37 @@ public FunctionView Describe()
...
+    private async Task<SKContext> CallPreExecutionHook(SKContext context)
+    {
+        if (this._preExecutionHookRequest is not null)
+        {
+            var preExecutionContext = new PreExecutionContext(context);
+
+            await this._preExecutionHookRequest.InvokeAsync(preExecutionContext).ConfigureAwait(false);
+
+            // Allow the pre execution hook to update the context variables if needed
+            context.Variables.Update(preExecutionContext.SKContext.Variables, true);
```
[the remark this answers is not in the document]
RogerBarreto: You are right, I can remove that line or leave a TODO here so we can think on the immutable approach for the SKContext while exposing it in the `ExecutionContext`

dotnet/samples/KernelSyntaxExamples/Example55_PlanHandlers.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.SemanticKernel.Planning;
+using Microsoft.SemanticKernel.SkillDefinition;
+using RepoUtils;
+
+// ReSharper disable once InconsistentNaming
+public static class Example55_PlanHandlers
+{
+    public static async Task RunAsync()
+    {
+        Console.WriteLine("======== Sequential Planner - Using Step Execution Handlers ========");
```
lemillermicrosoft, line 16: Naming.

dotnet/samples/KernelSyntaxExamples/Example54_FunctionHandlers.cs
```diff
@@ -0,0 +1,79 @@
...
+            return Task.CompletedTask;
+        }
+
+        Task MyPostExecutionHandler(PostExecutionContext executionContext)
+        {
+            Console.WriteLine($"Post Execution Handler - Total Tokens: {executionContext.SKContext.ModelResults.First().GetOpenAITextResult().Usage.TotalTokens}");
+            return Task.CompletedTask;
+        }
+
+        excuseFunction.SetPreExecutionHandler(MyPreHandler);
```
lemillermicrosoft, line 67: Is there a way to set the pre execution handler on the kernel such that any function that gets called is routed through the same handler?

lemillermicrosoft: How will things look for ImportSkill for example -- do users need to iterate all the functions and add the handlers?

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -245,21 +245,32 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
...
             var step = this.Steps[this.NextStepIndex];
-
             // Merge the state with the current context variables for step execution
             var functionVariables = this.GetNextStepVariables(context.Variables, step);
 
             // Execute the step
             var functionContext = new SKContext(functionVariables, context.Skills, context.Logger);
+
+            string? renderedPrompt = null;
+            if (step.Function is not null && step.Function is SemanticFunction semanticFunction)
```
lemillermicrosoft, line 255: A few things feel odd:

- Calling RenderAsync twice for a single semantic function
- Every step in a plan will call the handler, but if I have a semantic or native function that's composed of native and semantic functions, I'll get only a single callback.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -360,6 +397,32 @@ public ISKFunction SetAIConfiguration(CompleteRequestSettings settings)
             : this.Function.SetAIConfiguration(settings);
     }
 
+    /// <summary>
+    /// Used for setting a pre-execution handler to a plan and its children.
```
lemillermicrosoft, line 401: nit: maybe steps instead of children to match the property name

RogerBarreto:
Closing as we are moving to IKernel handlers centric for the moment.

PR #2053, opened 2023-07-18
.Net: Added better formatting for responses from Bing Searches & Ability to use custom Bing Search endpoint.

SOE-YoungS:
### Motivation and Context
Returning just the page snippet from a Bing search results in a lack of context or ability to follow up on a response.

This PR addresses that by adding better formatting for Bing Search responses via the "Skills.Web" package. 
It also adds a method to use a custom endpoint for Bing Search.

#### Formatting Changes
Previous response format example:
image

New response format example:
image

As can be seen by the examples provided, the new response format contains much more context & provides the user with the URL of the search result so that they can click through and read further.

This can be obtained programmatically later on using regex matching, then could be fed into the "[Search Url Skill](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Skills/Skills.Web/SearchUrlSkill.cs)", to scrape the page directly, before finally summarizing and returning a complete summary of the page to the user.

#### Endpoint Changes
The ability to use a custom endpoint for Bing Search also enables a path to use an API Management instance as a front end for the Bing Search API. This is required by some end users for scenarios such as enterprise logging and usage counting for cross charge.

The default behaviour for the BingConnector is to target the Bing Search endpoint directly. It is entirely optional to use a custom endpoint. Doing so, does not introduce any other code change requirements to achieve use of Bing Search.

### Contribution Checklist


dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -23,14 +23,16 @@ public sealed class BingConnector : IWebSearchEngineConnector
     private readonly ILogger _logger;
     private readonly HttpClient _httpClient;
     private readonly string? _apiKey;
+    private readonly string? _uri;
+    private readonly string defaultUri = "https://api.bing.microsoft.com/v7.0/search?q";
```
dmytrostruk, line 27: 1. Naming convention is not followed here - name should start with underscore - `_defaultUri`.
2. This string should be just constant, there will be no need to update it in constructor. If we convert it to constant, the name should be `DefaultUri`.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -23,14 +23,16 @@ public sealed class BingConnector : IWebSearchEngineConnector
...
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BingConnector"/> class.
     /// </summary>
     /// <param name="apiKey">The API key to authenticate the connector.</param>
     /// <param name="logger">An optional logger to log connector-related information.</param>
-    public BingConnector(string apiKey, ILogger<BingConnector>? logger = null) :
-        this(apiKey, new HttpClient(NonDisposableHttpClientHandler.Instance, false), logger)
+    public BingConnector(string apiKey, string uri = "", ILogger<BingConnector>? logger = null) :
```
dmytrostruk, line 34: If we mark `defaultUri` as constant, we can then use it here as default value - `string uri = DefaultUri`. In this case it will be visible for class consumers which URI will be used by default:
image

Same applies for constructor below.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -75,9 +78,17 @@ public async Task<IEnumerable<string>> SearchAsync(string query, int count = 1,
...
+        {
+            output.Add(string.Format("Bing Search Results\n\n" +
+                "----\n\n" +
+                $"Page Name: {data?.WebPages?.Value.FirstOrDefault().Name}\n\n" +
+                $"Page Snippet: {data?.WebPages?.Value.FirstOrDefault().Snippet}...\n\n" +
+                $"Page Url: {data?.WebPages?.Value.FirstOrDefault().Url}"));
+        }
 
-        return results == null ? Enumerable.Empty<string>() : results.Select(x => x.Snippet);
+        return output.Count == 0 ? Enumerable.Empty<string>() : output.Take(1);
```
dmytrostruk, line 91: Previously we were returning `results.Select(x => x.Snippet)` - all snippets that come in HTTP response, based on `count` property in request query parameter. Now it looks like we will return only first result, which definitely changes the way how this functionality works.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -75,9 +78,17 @@ public async Task<IEnumerable<string>> SearchAsync(string query, int count = 1,
...
 
-        WebPage[]? results = data?.WebPages?.Value;
+        List<string> output = new();
+        if (data?.WebPages?.Value != null)
+        {
+            output.Add(string.Format("Bing Search Results\n\n" +
+                "----\n\n" +
+                $"Page Name: {data?.WebPages?.Value.FirstOrDefault().Name}\n\n" +
+                $"Page Snippet: {data?.WebPages?.Value.FirstOrDefault().Snippet}...\n\n" +
+                $"Page Url: {data?.WebPages?.Value.FirstOrDefault().Url}"));
```
dmytrostruk, line 88: I'm not sure about this format. Method `IWebSearchEngineConnector.SearchAsync` clearly describes that we are expecting snippet only. If we want to make this result more extended, we can add another method, which will return not a string, but specific model, with properties like `Name`, `Snippet` and `Url`. In this case there will be no need to parse string result using regular expressions and so on.

dmytrostruk (commented):
@SOE-YoungS thank you for contribution.

I left initial comments, but it looks like this PR generates 1 error and 5 warnings:
image

Please, make sure to resolve all errors and warnings, and add unit and/or integration tests for new functionality. Thanks again!

SOE-YoungS:
> @SOE-YoungS thank you for contribution.
> 
> I left initial comments, but it looks like this PR generates 1 error and 5 warnings: image
> 
> Please, make sure to resolve all errors and warnings, and add unit and/or integration tests for new functionality. Thanks again!

Thanks for the feedback, I'll look into the items mentioned and will update the PR in due course.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
SOE-YoungS: Updated to a constant as suggested.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
SOE-YoungS: Agreed, went with your suggestion.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
SOE-YoungS: Rewritten the method to use generics. Both IEnumerable<string> & IEnumerable<WebPage> types are now supported.

SOE-YoungS: > Rewritten the method to use generics. Both `IEnumerable<string>` & `IEnumerable<WebPage>` types are now supported. This allows usage such as,
> 
> ```
>             // Init the bingConnector and import the skill
>             var bingConnector = new BingConnector(_config["Bing_Search_API_Key"], 
>                 uri: "https://soe-azureoai.azure-api.net/bingsearch/v7.0/search?q=");
> 
>             var searchSkill = _sk.ImportSkill(new WebSearchEngineSkill(bingConnector), "bing");
> 
>             ContextVariables contextVars = new($"{text}");
>             contextVars.Set(WebSearchEngineSkill.Parameters.CountParam, $"3");
>             contextVars.Set(WebSearchEngineSkill.Parameters.OffsetParam, $"0");
> 
>             // Perform the search and return the full result object.
>             var search = await _sk.RunAsync(contextVars, searchSkill["GetSearchResults"]);
>             var webPages = JsonConvert.DeserializeObject<List<WebPage>>(search.Result);
> 
>             string results = $"Bing Search Results\n\n" +
>                     "----\n\n";
> 
>             switch(webPages.Count)
>             {
>                 case 0:
>                     results += "No results found.";
>                     break;
>                     case 1:
>                     results += $"Page Name: {webPages[0].Name}\n\n" +
>                         $"Page Snippet: {webPages[0].Snippet}...\n\n" +
>                         $"Page Url: {webPages[0].Url}";
>                     break;
>                 default:
>                     for (int i = 0; i < webPages.Count; i++)
>                     {
>                         results += $"--- Result {i+1} ---\n\n" +
>                             $"Page Name: {webPages[i].Name}\n\n" +
>                             $"Page Snippet: {webPages[i].Snippet}...\n\n" +
>                             $"Page Url: {webPages[i].Url}";
> 
>                         if (i+1 < webPages.Count)
>                         {
>                             results += $"\n\n";
>                         }
>                     }
>                     break;
>             }
>             
>             //// Perform the search and return snippets.
>             // var search = await _sk.RunAsync(contextVars, searchSkill["GetSearchResults"]);
>             // var results = JsonConvert.DeserializeObject<List<string>>(search.Result);
>             
>             // Process results here
> ```

See previous to related comment.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
SOE-YoungS: > Rewritten the method to use generics. Both `IEnumerable<string>` & `IEnumerable<WebPage>` types are now supported. This allows usage such as,
> 
> ```
>             // Init the bingConnector and import the skill
>             var bingConnector = new BingConnector(_config["Bing_Search_API_Key"], 
>                 uri: "https://soe-azureoai.azure-api.net/bingsearch/v7.0/search?q=");
> 
>             var searchSkill = _sk.ImportSkill(new WebSearchEngineSkill(bingConnector), "bing");
> 
>             ContextVariables contextVars = new($"{text}");
>             contextVars.Set(WebSearchEngineSkill.Parameters.CountParam, $"3");
>             contextVars.Set(WebSearchEngineSkill.Parameters.OffsetParam, $"0");
> 
>             // Perform the search and return the full result object.
>             var search = await _sk.RunAsync(contextVars, searchSkill["GetSearchResults"]);
>             var webPages = JsonConvert.DeserializeObject<List<WebPage>>(search.Result);
> 
>             string results = $"Bing Search Results\n\n" +
>                     "----\n\n";
> 
>             switch(webPages.Count)
>             {
>                 case 0:
>                     results += "No results found.";
>                     break;
>                     case 1:
>                     results += $"Page Name: {webPages[0].Name}\n\n" +
>                         $"Page Snippet: {webPages[0].Snippet}...\n\n" +
>                         $"Page Url: {webPages[0].Url}";
>                     break;
>                 default:
>                     for (int i = 0; i < webPages.Count; i++)
>                     {
>                         results += $"--- Result {i+1} ---\n\n" +
>                             $"Page Name: {webPages[i].Name}\n\n" +
>                             $"Page Snippet: {webPages[i].Snippet}...\n\n" +
>                             $"Page Url: {webPages[i].Url}";
> 
>                         if (i+1 < webPages.Count)
>                         {
>                             results += $"\n\n";
>                         }
>                     }
>                     break;
>             }
>             
>             //// Perform the search and return snippets.
>             // var search = await _sk.RunAsync(contextVars, searchSkill["GetSearchResults"]);
>             // var results = JsonConvert.DeserializeObject<List<string>>(search.Result);
>             
>             // Process results here
> ```

See previous to related comment.

SOE-YoungS:
> @SOE-YoungS thank you for contribution.
> 
> I left initial comments, but it looks like this PR generates 1 error and 5 warnings: image
> 
> Please, make sure to resolve all errors and warnings, and add unit and/or integration tests for new functionality. Thanks again!

I have resolved the previous unit test error, however, having written a new unit test for the added functionality, I am now stuck as the new unit test passes if I test the method manually, but fails for the unit test (I cannot pinpoint it to a code issue).

SOE-YoungS:
While addressing the issues in this PR, I updated the GoogleConnector to unify the return types and set user expectations.

However, I do not have facility or capacity to test the changes to the google connector.

I do not expect any errors as a result of these changes to the GoogleConnector, but will commit to resolving any issues that crop up as a result of the change (if someone can tell me what errors occur).

dmytrostruk (commented):
Hi @SOE-YoungS !
Thank you for your contribution.

It looks like changes in this PR still generate some warnings. Unfortunately, we won't be able to merge it until all warning are resolved:
image

Secondly, I think the solution for this problem may be more complex. Currently, Kernel works only with strings, which means that even if Bing/Google connectors will work with complex types, we still need to serialize/deserialize results to be used in Kernel. 

We are considering the possibility to make Kernel more generic to work not only with strings, but with complex types. But this part of work will require a lot of changes and testing. 

My suggestion would be to wait with this PR until we refactor Kernel and add possibility to process complex types, so in this case your idea for Bing/Google connectors can be applied more smoothly.

Please let me know if that makes sense. Thanks again!

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -75,9 +80,29 @@ public async Task<IEnumerable<string>> SearchAsync(string query, int count = 1,
...
+                foreach (var webPage in data?.WebPages?.Value)
+                {
+                    webPages.Add(webPage);
+                }
+                returnValues = webPages.Take(count).ToList() as List<T>;
+            }
+            else
+            {
+                throw new NotSupportedException($"Type {typeof(T)} is not supported.");
+            }
```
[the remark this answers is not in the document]
SOE-YoungS: Agreed, this should be considered, however, it will be some time before the suggested change to allow complex types is implemented.

I feel that there is a benefit for some to be able to use this PR for now, considering returning just a "snippet" can only really be used as context for chained input to the LLM. Whereas some would prefer to simply use the skill to return results of a search to the user. This PR caters for both scenarios.

I'll address the warnings :)

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -108,15 +133,15 @@ private sealed class BingSearchResponse
 
     [SuppressMessage("Performance", "CA1812:Internal class that is apparently never instantiated",
         Justification = "Class is instantiated through deserialization.")]
-    private sealed class WebPages
+    public sealed class WebPages
```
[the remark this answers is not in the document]
SOE-YoungS: Can you clarify why these should be private please?

I've opted to expose these classes, to enable the caller to be able to use the class objects in calls & returned output.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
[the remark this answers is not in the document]
dmytrostruk: > considering returning just a "snippet" can only really be used as context for chained input to the LLM. Whereas some would prefer to simply use the skill to return results of a search to the user.

Well, I believe this is the purpose of these connectors in Semantic Kernel SDK - to be used as context for chained input to the LLM :)
If there is a need to get more complex responses from Bing/Google, it's possible to use Bing/Google API/SDK directly. 

I agree that complex responses have benefits, and we will definitely add them, as soon as Kernel will know how to work with them properly.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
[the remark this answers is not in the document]
dmytrostruk: We should be careful what we expose to SDK users. If we make some class `public`, it means that SDK user will get access to it. In this case we should think about more specific naming (e.g. `BingWebPage`), add detailed XML documentation what this class is responsible for and provide documentation for each public member. And most importantly - we need to think whether we should provide this class to the end users at all. 

As I mentioned in previous comment, Kernel doesn't know how to work with this class (yet). Which means that it will be useful only when you call this skill directly without a Kernel. In this case it will look like we are developing Bing SDK, which is not the case. This skill should be useful in a context of Semantic Kernel, rather than separately.

dotnet/src/Skills/Skills.WebTests/Skills.WebTests.csproj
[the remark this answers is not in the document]
SOE-YoungS: Was accidentally added out of habit. I've removed it :)

dotnet/src/Skills/Skills.WebTests/Bing/BingConnectorTests.cs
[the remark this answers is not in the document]
SOE-YoungS: The class has been removed.

dotnet/Directory.Packages.props
```diff
@@ -12,6 +12,8 @@
     <PackageVersion Include="Microsoft.Bcl.HashCode" Version="[1.1.0, )" />
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Http" Version="6.0.0" />
+    <PackageVersion Include="MSTest.TestAdapter" Version="2.2.10" />
+    <PackageVersion Include="MSTest.TestFramework" Version="2.2.10" />
```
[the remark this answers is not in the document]
SOE-YoungS: These references have been removed in [f56f18ee9405ae30b96cdbd1a7c67f0b1451e9f9](commit)

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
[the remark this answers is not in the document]
dmytrostruk: > If a Skill exists in a library that is already in use, surely re-use of that library is more efficient

It means that we will need to implement Bing API completely, support and maintain it. But we don't want to re-invent the wheel, especially if such library already exists:
https://learn.microsoft.com/en-us/bing/search-apis/bing-web-search/quickstarts/sdk/web-search-client-library-csharp

Maybe it would be better to re-use already existing SDK in this connector, but on the other hand, it will bring more dependencies and functionality, that is not going to be used by Semantic Kernel (yet). Which is why this connector is lightweight at the moment, and we will extend it as soon as Kernel will know how to work with complex types.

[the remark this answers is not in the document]
SOE-YoungS: Forgive me, but I am a little confused here.

I have not changed the implementation of the API request in this Skill / connector at all (which equates to a simple Rest API call to the Bing Search API endpoint and is as lightweight as it gets). 

I have simply added an (optional) alternative output from the connector back to the caller, the original functionality should not be impacted at all.

I fail to see where the issue is here, beyond the response not being in a compatible format to use with the Kernel in a chaining scenario?

As can be seen from the example taken from our prod code (Teams bot)... The kernel works as expected.

```public async Task<BotResponse> Search(string text)
        {
            // Init the bingConnector and import the skill
            var bingConnector = new BingConnector(_config["BingSearchAPIKey"], uri: _config["BingSearchUrl"]);
            var searchSkill = _sk.ImportSkill(new WebSearchEngineSkill(bingConnector), "bing");

            ContextVariables contextVars = new($"{text}");
            contextVars.Set(WebSearchEngineSkill.Parameters.CountParam, $"3");
            contextVars.Set(WebSearchEngineSkill.Parameters.OffsetParam, $"0");

            // Perform the search and return the snippet(s) only.
            // var search = await _sk.RunAsync(contextVars, searchSkill["Search"]);
            // var snippets = JsonConvert.DeserializeObject<List<string>>(search.Result);

            // Perform the search and return the full result.
            var search = await _sk.RunAsync(contextVars, searchSkill["GetSearchResults"]);
            var WebPages = JsonConvert.DeserializeObject<List<WebPage>>(search.Result);

            string results = $"Bing Search Results\n\n" +
                    "----\n\n";

            switch(WebPages.Count)
            {
                case 0:
                    results += "No results found.";
                    break;
                    case 1:
                    results += $"Page Name: {WebPages[0].Name}\n\n" +
                        $"Page Snippet: {WebPages[0].Snippet}...\n\n" +
                        $"Page Url: {WebPages[0].Url}";
                    break;
                default:
                    for (int i = 0; i < WebPages.Count; i++)
                    {
                        results += $"--- Result {i+1} ---\n\n" +
                            $"Page Name: {WebPages[i].Name}\n\n" +
                            $"Page Snippet: {WebPages[i].Snippet}...\n\n" +
                            $"Page Url: {WebPages[i].Url}";

                        if (i+1 < WebPages.Count)
                        {
                            results += $"\n\n";
                        }
                    }
                    break;
            }
```
As can be seen, this example matches the format as defined in existing Skill examples found in the repo IE: [graph-api-skills](https://github.com/microsoft/semantic-kernel/blob/main/samples/dotnet/graph-api-skills/Program.cs) & is also compatible with the example provided for the Bing Search Skill itself: [10-BingSearch-using-kernel](https://github.com/microsoft/semantic-kernel/blob/main/samples/notebooks/dotnet/10-BingSearch-using-kernel.ipynb)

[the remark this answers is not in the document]
dmytrostruk: > I have simply added an (optional) alternative output from the connector back to the caller, the original functionality should not be impacted at all.
> 
> I fail to see where the issue is here, beyond the response not being in a compatible format to use with the Kernel in a chaining scenario?

@SOE-YoungS An (optional) alternative output it's always good, but if it's not going to be used in the context of Semantic Kernel, then in my opinion it should live in another place. Skills functionality presented in this repository should be pluggable to Semantic Kernel.

But that's only my opinion :) If you disagree with it - that's totally fine.

A couple of suggestions from me based on recent changes:
1. Instead of `Newtonsoft.Json` you can use `System.Text.Json` if needed, to avoid extra dependency.
2. If you want to add an alternative optional output to Bing/Google connectors, you can add new separate methods in these classes without modifying already existing methods and `IWebSearchEngineConnector`.
3. It looks like you suppressed all the warnings, but we should avoid suppression and design a code in a way that it won't generate any warnings.

Thank you for contribution and sorry if my comments confused you.

dotnet/src/IntegrationTests/WebSkill/WebSkillTests.cs
```diff
@@ -40,28 +43,64 @@ public WebSkillTests(ITestOutputHelper output)
...
             search["Search"]
         );
 
         // Assert
         Assert.Contains(expectedAnswerContains, result.Result, StringComparison.OrdinalIgnoreCase);
     }
 
+    [Theory(Skip = "Bing search results not consistent enough for testing.")]
+    [InlineData("What is generally recognized as the tallest building in Seattle, Washington, USA?", "Columbia Center")]
+    public async Task BingSkillFullResultTestAsync(string prompt)
```
lemillermicrosoft, line 75: Second parameter is missing.

dotnet/src/IntegrationTests/WebSkill/WebSkillTests.cs
```diff
@@ -40,28 +43,64 @@ public WebSkillTests(ITestOutputHelper output)
...
         IKernel kernel = Kernel.Builder.WithLogger(this._logger).Build();
 
         using XunitLogger<BingConnector> connectorLogger = new(this._output);
-        BingConnector connector = new(this._bingApiKey, connectorLogger);
+        BingConnector connector = new(this._bingApiKey, logger: connectorLogger);
         Assert.NotEmpty(this._bingApiKey);
 
         WebSearchEngineSkill skill = new(connector);
+
+        ContextVariables contextVariables = new(prompt);
```
lemillermicrosoft, line 57: How come these changes?

dotnet/src/IntegrationTests/WebSkill/WebSkillTests.cs
```diff
@@ -40,28 +43,64 @@ public WebSkillTests(ITestOutputHelper output)
...
+        // Act
+        var task = await kernel.RunAsync(
+            contextVariables,
+            search["GetSearchResults"]
+        );
+
+        var webPages = JsonConvert.DeserializeObject<List<WebPage>>(task.Result);
+
+        // Assert
+        Assert.True(webPages != null && webPages.Count > 0);
```
lemillermicrosoft, line 101: Count should equal 1 yes?

dotnet/src/IntegrationTests/WebSkill/WebSkillTests.cs
```diff
@@ -40,28 +43,64 @@ public WebSkillTests(ITestOutputHelper output)
...
-            prompt,
+            contextVariables,
             search["Search"]
         );
 
         // Assert
         Assert.Contains(expectedAnswerContains, result.Result, StringComparison.OrdinalIgnoreCase);
     }
 
+    [Theory(Skip = "Bing search results not consistent enough for testing.")]
```
lemillermicrosoft, line 73: I bet if we increase the count, we can allow these tests to start running.

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -23,14 +23,19 @@ public sealed class BingConnector : IWebSearchEngineConnector
...
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BingConnector"/> class.
     /// </summary>
     /// <param name="apiKey">The API key to authenticate the connector.</param>
+    /// <param name="uri"></param>
     /// <param name="logger">An optional logger to log connector-related information.</param>
-    public BingConnector(string apiKey, ILogger<BingConnector>? logger = null) :
-        this(apiKey, new HttpClient(NonDisposableHttpClientHandler.Instance, false), logger)
+    [SuppressMessage("Performance", "CA1054:Change the type of parameter 'uri'...",
```
lemillermicrosoft, line 35: You could make the parameters Uri and all null and in the constructor fallback to `private static readonly Uri DefaultUri = new("https://api.bing.microsoft.com/v7.0/search?q");`

dotnet/src/Skills/Skills.Web/Bing/BingConnector.cs
```diff
@@ -114,9 +145,16 @@ private sealed class WebPages
         public WebPage[]? Value { get; set; }
     }
 
+    /// <summary>
+    /// A sealed class containing the deserialized response from the Bing Search API.
+    /// </summary>
+    /// <returns>A WebPage object containing the Bing Search API response data.</returns>
+    /// <remarks><b>Note:</b> Complex types are not compatible with the kernel for chaining yet. See <see href="https://github.com/microsoft/semantic-kernel/pull/2053#discussion_r1274782835">PR#2053</see></remarks>
     [SuppressMessage("Performance", "CA1812:Internal class that is apparently never instantiated",
```
lemillermicrosoft, line 153: This suppression shouldn't be needed as-is -- but I think it's better to make this internal for tests.

dotnet/src/Skills/Skills.Web/Google/GoogleConnector.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.SemanticKernel.Diagnostics;
+using static Microsoft.SemanticKernel.Skills.Web.Bing.BingConnector;
```
lemillermicrosoft, line 13: Making the GoogleConnector depend on the BingConnector feels odd -- maybe it's best to refactor that class out into it's own. Curious what others may think re: it being public or internal.

lemillermicrosoft (commented):
Thanks so much for the PR. Leaving some comments for now.

lemillermicrosoft:
This has been open for a while, so for clarity and visibility, I am going to close this for now. If you'd like to continue to work on this though given the feedback, please feel free to re-open and we will make sure we get changes in 2-3 days after things are updated and passing the gates.

SOE-YoungS:
Can we re-open this please? 

I am in the process of completing the work required to bring this PR up to date with the current changes.

joslat:
This is a pity, this would be such a great addition to the web search plugin...

PR #2069, opened 2023-07-19
Python: added settings function to load direct into constructor

eavanvalkenburg:
### Motivation and Context

Noticed that another commonly edited environment setting for api_version was missing from the settings util. And want to pass the env to the constructor.

### Description

Added api_version to the settings util so that that can be controlled through .env as well.
Added convenience function that outputs a dict that can be passed directly into Completion constructors.
Updated sample to use that.

### Contribution Checklist


lemillermicrosoft:
@shawncal can you confirm whether we want to do this and if so link to an issue to ensure parity with dotnet, java, etc.

awharrison-28:
Functionally, this PR is fine, but this is a commonly used API that does break now that the return is 4 values instead of 3. If we do want this feature, we'll need to make sure that the kernel_syntax_examples and jupyter notebooks are consistent. @shawncal

eavanvalkenburg:
@awharrison-28 @lemillermicrosoft I've made a change that is slightly uglier but no longer break the API, when setting include_api_version to True (default is False) you get 4 strings back, otherwise still three.

lemillermicrosoft (approved):

awharrison-28 (approved):

PR #2075, opened 2023-07-19
Rename Azure Search to Azure Cognitive Search

dluc:
### Motivation and Context

The service was renamed to Azure Cognitive Search in 2019, so we should follow the new convention. The only thing still named "Azure Search" is the old Azure SDK, out of scope.

### Description

* Revert nuget package name to the previous name (we'll delete the new package, which has been out only for few days)
* Rename C# and Python classes and namespaces


shawncal (approved):

awharrison-28 (approved):
Python side looks good. There's an unclosed client warning that I think is windows specific. I'm logging an item to track.

dluc:
@awharrison-28 yeah those warnings are showing on Macos too...

PR #2080, opened 2023-07-19
Java: Allow functions to be loaded from the classpath

johnoliver:
Allow importing skills from the classpath using `getResourceAsStream` so that people can embed skills inside their jar


markwallace-microsoft (dismissed):
LGTM and this is a really nice addition.

Using this in some samples would be nice also.

brunoborges:
Should we consider a default directory inside META-INF?

johnoliver:
added example code

markwallace-microsoft (approved):

markwallace-microsoft (approved):

PR #2083, opened 2023-07-20
Python: Azure Cognitive Search bug fixes

dluc:
* close search clients
* fix get_collections_async()
* remove code duplication
* removed unused get_collection_async() method
* add more integration tests


awharrison-28 (changes requested):
```
    def __del__(self):
        try:
            loop = asyncio.get_event_loop()
            if loop.is_running():
                loop.create_task(self.close())
            else:
                loop.run_until_complete(self.close())
        except Exception:
            pass
```

I believe should be 


```
def __del__(self):
        try:
            loop = asyncio.get_event_loop()
            if not loop.is_running():
                loop.create_task(self.close())
            else:
                loop.run_until_complete(self.close())
        except Exception:
            pass
```

awharrison-28 (approved):

PR #2086, opened 2023-07-20
.Net: Remove CancellationToken from SKContext

shawncal:
### Motivation and Context

The CancellationToken is currently propagated to SKFunction calls and Plans via the SKContext. While this works as expected, it results in nonstandard call patterns that confuse the developer and the IDE alike.

Some issues:
- (for functions) Without CancellationToken specified in a function's signature, the analyzers fail to suggest that the token should be further passed through to async calls made within the function. These calls (i.e. Http calls, file IO...) are usually the long-running asynchronous operations in which CancellationToken is actually monitored and respected. They're no good if they're not passed in!
- (for SK callers) Setting SKContext.CancellationToken before calling a function is very unnatural, and is pretty much hidden from the developer.

### Description

In this change, the CancellationToken is passed through all async functions as an explicit arg with a default value, just as the tools and the .Net developer would expect.

Additionally, a new IKernel.RunAsync overload is provided with a single function, and CancellationToken as the final arg, aligning with the standard .net method signature expectation.

**BREAKING CHANGE:** Some functions developers may be loading the CancellationToken from an SKContext input arg, or callers setting it in the SKContext when calling Kernel operations, per our guidance and examples.

Following this change, all examples and notebooks will be cleaned of such use patterns.

### Contribution Checklist


dmytrostruk (approved):
Left small comment, but overall, I really like this change!

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -87,6 +87,16 @@ public interface IKernel
...
 
+    /// <summary>
+    /// Run a single synchronous or asynchronous <see cref="ISKFunction"/>.
+    /// </summary>
+    /// <param name="skFunction">A Semantic Kernel function to run</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Result of the function composition</returns>
+    Task<SKContext> RunAsync(
+        ISKFunction skFunction,
+        CancellationToken cancellationToken); // TODO: add `= default` value after the `params` overloads are removed, or it will be ambiguous.
```
stephentoub, line 98: What's an example where it's ambiguous?  I'd expect this overload to always win for a single ISKFunction argument, even with the default argument.

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -87,6 +87,16 @@ public interface IKernel
...
+
```
stephentoub, line 99: What's the plan for the below overloads, e.g.
```C#
    Task<SKContext> RunAsync(
        string input,
        CancellationToken cancellationToken,
        params ISKFunction[] pipeline);
```
?  Should they be obsoleted / deleted?

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -188,8 +191,8 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
 
             try
             {
-                context.CancellationToken.ThrowIfCancellationRequested();
-                context = await f.InvokeAsync(context).ConfigureAwait(false);
+                cancellationToken.ThrowIfCancellationRequested();
```
stephentoub, line 194: This line is fine, but also shouldn't be necessary.

stephentoub (approved):
Very nice.

RogerBarreto (dismissed):
Nice! LGTM

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -321,15 +324,15 @@ public string ToString(bool writeIndented)
 
     private static readonly JsonSerializerOptions s_toStringStandardSerialization = new();
     private static readonly JsonSerializerOptions s_toStringIndentedSerialization = new() { WriteIndented = true };
-    private Func<ITextCompletion?, CompleteRequestSettings?, SKContext, Task<SKContext>> _function;
+    private Func<ITextCompletion?, CompleteRequestSettings?, SKContext, CancellationToken, Task<SKContext>> _function;
```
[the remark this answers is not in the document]
shawncal: Yeah, had the same thought. I didn't want to break the glue that does all this magical mapping, so we'll follow up with that in an isolated PR. Good call.

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
shawncal: I believe one below with `**params** ISKFunction[]` will result in ambiguity if you only pass
   `k.RunAsync(oneFunction)`

No?

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
shawncal: Yeah, we'll obsolete the "params" ones. They don't allow for any arg passing or anything. 
Current thinking: Kernel either runs a `Function` or a `Plan`.
...where a `Plan` has support for chaining, args, other complexity.

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
stephentoub: I wouldn't expect it to. params should lose. Do you see otherwise?

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -143,6 +143,10 @@ public void RegisterMemory(ISemanticTextMemory memory)
         this._memory = memory;
     }
 
+    /// <inheritdoc/>
+    public Task<SKContext> RunAsync(ISKFunction skFunction, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
shawncal: Same comment - I think this would be (temporarily) ambiguous with the params... overload, when no token is passed.
I can try it and see if the compiler squawks.

dotnet/src/SemanticKernel/TemplateEngine/Blocks/ICodeRendering.cs
```diff
@@ -14,6 +15,7 @@ public interface ICodeRendering
     /// Render the block using the given context, potentially using external I/O.
     /// </summary>
     /// <param name="context">SK execution context</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>Rendered content</returns>
-    public Task<string> RenderCodeAsync(SKContext context);
+    public Task<string> RenderCodeAsync(SKContext context, CancellationToken cancellationToken);
```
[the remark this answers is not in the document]
shawncal: Yup, this one's just missing. Thanks!

dotnet/src/SemanticKernel/Kernel.cs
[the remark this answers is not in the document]
shawncal: Nevermind. Seems fine! Added

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
shawncal: Yup, you're right. Updated

dotnet/src/SemanticKernel/Kernel.cs
shawncal: It's just a way of bailing out of the function loop if cancellation is requested, instead of putting the onus of checking on the downstream function calls. Do you think we should let the functions themselves handle this instead? Or will the runtime bail out of the Task for us upon cancellation requested?

dotnet/src/Extensions/Extensions.UnitTests/Planning/ActionPlanner/ActionPlannerTests.cs
```diff
@@ -35,8 +35,8 @@ public async Task ExtractsAndDeserializesWellFormedJsonFromPlannerResult()
             functionsView.AddFunction(functionView);
 
             mockFunction.Setup(x =>
-                    x.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<CompleteRequestSettings>()))
-                .Returns<SKContext, CompleteRequestSettings>((context, settings) =>
+                    x.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<CompleteRequestSettings>(), default))
```
shawncal, line 38:
```suggestion
                    x.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<CompleteRequestSettings>(), It.IsAny<CancellationToken>()))
```

dotnet/src/Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SequentialPlanParserTests.cs
```diff
@@ -77,7 +75,7 @@ private static Mock<ISKFunction> CreateMockFunction(FunctionView functionView, s
 
             var result = this.CreateSKContext(kernel);
             result.Variables.Update(resultString);
-            mockFunction.Setup(x => x.InvokeAsync(It.IsAny<SKContext>(), null))
+            mockFunction.Setup(x => x.InvokeAsync(It.IsAny<SKContext>(), null, default))
```
shawncal, line 78:
```suggestion
            mockFunction.Setup(x => x.InvokeAsync(It.IsAny<SKContext>(), null, It.IsAny<CancellationToken>()))
```

dotnet/src/Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SequentialPlannerTests.cs
```diff
@@ -43,8 +43,8 @@ public async Task ItCanCreatePlanAsync(string goal)
             functionsView.AddFunction(functionView);
 
             mockFunction.Setup(x =>
-                    x.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<CompleteRequestSettings>()))
-                .Returns<SKContext, CompleteRequestSettings>((context, settings) =>
+                    x.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<CompleteRequestSettings>(), default))
```
shawncal, line 46:
```suggestion
                    x.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<CompleteRequestSettings>(), It.IsAny<CancellationToken>()))
```

dotnet/src/Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SequentialPlannerTests.cs
```diff
@@ -140,7 +141,7 @@ public async Task EmptyGoalThrowsAsync()
         var planner = new Microsoft.SemanticKernel.Planning.SequentialPlanner(kernel.Object);
 
         // Act
-        await Assert.ThrowsAsync<PlanningException>(async () => await planner.CreatePlanAsync(""));
+        await Assert.ThrowsAsync<PlanningException>(async () => await planner.CreatePlanAsync("", default));
```
shawncal, line 144:
```suggestion
        await Assert.ThrowsAsync<PlanningException>(async () => await planner.CreatePlanAsync(""));
```

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -161,9 +171,8 @@ public interface IKernel
     /// <summary>
     /// Create a new instance of a context, linked to the kernel internal state.
     /// </summary>
-    /// <param name="cancellationToken">Optional cancellation token for operations in context.</param>
     /// <returns>SK context</returns>
-    SKContext CreateNewContext(CancellationToken cancellationToken = default);
+    SKContext CreateNewContext();
```
shawncal, line 175: /// <summary>
    /// Create a new instance of a context, linked to the kernel internal state.
    /// </summary>
    /// <param name="cancellationToken">Optional cancellation token for operations in context.</param>
    /// <returns>SK context</returns>
    [Obsolete("SKContext no longer contains the CancellationToken. Use CreateNewContext().")]
    SKContext CreateNewContext(CancellationToken cancellationToken);

    /// <summary>
    /// Create a new instance of a context, linked to the kernel internal state.
    /// </summary>
    /// <returns>SK context</returns>
    SKContext CreateNewContext();

shawncal, line 175:
```suggestion
    SKContext CreateNewContext();
    
    /// <summary>
    /// Create a new instance of a context, linked to the kernel internal state.
    /// </summary>
    /// <param name="cancellationToken">Optional cancellation token for operations in context.</param>
    /// <returns>SK context</returns>
    [Obsolete("SKContext no longer contains the CancellationToken. Use CreateNewContext().")]
    SKContext CreateNewContext(CancellationToken cancellationToken);
```

awharrison-28 (approved):

dotnet/src/SemanticKernel/Kernel.cs
stephentoub: It's fine to do this if we think it's likely the functions themselves won't be checking the token.  You could also do it in SKFunction.InvokeAsync to avoid doing this check anywhere InvokeAsync is called.

PR #2101, opened 2023-07-20
Adding example notebook - Native Functions in C#

efficienthacks:
Example Notebook of how to use SequentialPlanner in C# with Native Functions

### Motivation and Context

@alexchaomander suggested that I add this example (of using the Planner with Native C# Functions) as a sample Notebook.  The existing notebook only uses Semantic Functions.

This will help C# developers that already have existing methods, but want to use a Planner to chain them together.

### Description

Added an example notebook - tried to follow the same structure as https://github.com/microsoft/semantic-kernel/blob/main/samples/notebooks/dotnet/05-using-the-planner.ipynb 

Feel free to rename this to whatever convention you feel is best.

### Contribution Checklist


lemillermicrosoft:
Thanks for the PR! A few thoughts (cc @alexchaomander  and @matthewbolanos)

- I'm now realizing we didn't have *any* notebooks go over the creation of native functions/plugins. Maybe we should have a top level notebook just for that, before planner. Curious what others think and whether you'd like to do that @efficienthacks 
- I really appreciate the serialization examples you've added, it would make sense in the general planner notebook I think.

I realize it's been a month, so sorry for the delay here.

lemillermicrosoft:
Again, sorry for the delay. For clarity and visibility, I am going to close this for now. If you'd like to continue to work on this though given the feedback, please feel free to re-open and we will make sure we get changes in 2-3 days after things are updated.

alexchaomander:
I do think having a notebook like or having this be a part of our documentation would be good. It's definitely good instructional material for people who aren't familiar with all the SK concepts. 

@matthewbolanos maybe this can be something to include on our docs site?

PR #2109, opened 2023-07-21
.Net: Sync HTTP User Agent across HTTP clients/Azure resources

dluc:
Sync HTTP User Agent value across clients.
Support Azure telemetry env var when using Azure Core SDK.


dotnet/src/SemanticKernel.Abstractions/Diagnostics/MicrosoftDiagnostics.cs
dmytrostruk, line 1: Overall, I have mixed feelings about this class.

1. Property `IsTelemetryEnabled` and related helper members (constant `TelemetryDisabledEnvVar`, method `EnvVarToBool`) - as per XML documentation and logic in general - it's clear that this logic applies only to Azure SDK. So, I'm not sure whether it can be located in `SemanticKernel.Abstractions` package, even if it's used both in ACS memory connector and OpenAI connector packages. I think method for boolean conversion can be located in `InternalUtilities` while constant `AZURE_TELEMETRY_DISABLED` can live in all places where needed separately. 

2. It looks like `AZURE_TELEMETRY_DISABLED` is already fetched on [Azure SDK side](https://github.com/Azure/azure-sdk-for-net/blob/b9a4d4b46abcabf77450ae74ae6db842163e0cc9/sdk/core/Azure.Core/src/DiagnosticsOptions.cs#L73C24-L73C24). As far as I understand, we need to reset it on SK side in case if `ClientOptions` was not passed by consumer (we have separate constructors for this option). But do we need to rely on `Environment.GetEnvironmentVariable` in this case? It looks like it's the first time when we use it to fetch settings directly in SK Core. Instead, we can use already existing approach - extend AzureOpenAI / OpenAI constructors and add separate property to enable/disable telemetry in Azure SDK. In this way, consumer will decide where to store this property (maybe another place than env var).

3. Constant `HttpUserAgent` is used in places:
    - As `ApplicationId` property of `DiagnosticsOptions` class from Azure SDK. From documentation about this property, it looks like it is related to `User-Agent` header.
    - As HTTP User Agent for all connectors - Hugging Face, Oobabooga, OpenAI.

Even if for both purposes we need to use the same value, I think it's better to define two separate constants:
- One for Azure SDK diagnostics and put it in a place where we use Azure SDK.
- Another one for HTTP Client and put it in place where common logic for HTTP operations lives (e.g. `InternalUtilities/Http`).

4. I'm not sure about naming `MicrosoftDiagnostics`. I think the better name will come when previous points in this list will be addressed. It may turn out that this class will no longer be necessary :)

dluc: The code is very much coupled to Azure SDK, e.g. the boolean logic, the env var name, the term "ApplicationId" are copied from Azure and I prefer having it in one class. The name "MicrosoftDiagnostics" captures the intention that this code is about "diagnosing our Microsoft code, within Azure and outside to 3P services". Given that the scenario is one ("identifying SK") I think it's best having a unique constant. The fact that Azure SDK calls it ApplicationId is something outside our control and scope.

dotnet/src/SemanticKernel.Abstractions/Diagnostics/MicrosoftDiagnostics.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+
+namespace Microsoft.SemanticKernel.Diagnostics;
+
+public static class MicrosoftDiagnostics
```
shawncal, line 7: I feel like this class name should have something more focused, since the namespace is already Diagnostics, and the only thing Microsoft-specific is that Azure env key. Can we use something with "Telemetry"?

dotnet/src/SemanticKernel.Abstractions/Diagnostics/MicrosoftDiagnostics.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        if (string.Equals(bool.FalseString, value, StringComparison.OrdinalIgnoreCase) ||
+            string.Equals("0", value, StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
+
+        return null;
+    }
```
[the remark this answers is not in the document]
dluc: sure though bear in mind that this logic is copied from the Azure SDK, so it's not something we'll be able to improve or change without the risk of breaking the integration with Azure

dotnet/src/Skills/Skills.OpenAPI/Extensions/KernelChatGptPluginExtensions.cs
```diff
@@ -49,10 +49,9 @@ public static class KernelChatGptPluginExtensions
 
         using var requestMessage = new HttpRequestMessage(HttpMethod.Get, url);
 
-        if (!string.IsNullOrEmpty(executionParameters?.UserAgent))
-        {
-            requestMessage.Headers.UserAgent.Add(ProductInfoHeaderValue.Parse(executionParameters!.UserAgent));
-        }
+        requestMessage.Headers.UserAgent.Add(!string.IsNullOrEmpty(executionParameters?.UserAgent)
```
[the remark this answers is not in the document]
dluc: this is one place where it shows the coupling with Azure Core SDK. The SDK allows only one header with a max length of 24 chars.

dotnet/src/SemanticKernel.Abstractions/Diagnostics/MicrosoftDiagnostics.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.SemanticKernel.Diagnostics;
+
+public static class MicrosoftDiagnostics
+{
+    /// <summary>
+    /// Env var used in Azure to enable/disable telemetry
+    /// See: https://learn.microsoft.com/en-us/dotnet/api/azure.core.diagnosticsoptions.istelemetryenabled?view=azure-dotnet
+    /// </summary>
+    private const string TelemetryDisabledEnvVar = "AZURE_TELEMETRY_DISABLED";
```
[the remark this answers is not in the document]
dluc: your call, I don't know if that's needed. The goal of the PR is supporting the existing user expectations, e.g. devs setting `AZURE_TELEMETRY_DISABLED` and our Azure SDK integration code honoring it

dotnet/src/SemanticKernel.Abstractions/Diagnostics/MicrosoftDiagnostics.cs
dluc: tbh I don't care about the name, I'll change it to Telemetry :-) 

My main concern is that someone will reuse this code for other things and we might end up breaking the integration with Azure Core SDK

dmytrostruk (approved):

PR #2118, opened 2023-07-21
Python: Qdrant Memory Store

awharrison-28:
### Motivation and Context
This PR adds Qdrant support for SK memory to the Python SK

### Description
- added `qdrant_memory_store.py`
- added `test_qdrant_memory_store.py`
- added dependency group `qdrant` to the pyproject.toml
- added dependency group `qdrant` to the `--without` list for python unit tests
- updated poetry

### Contribution Checklist

- [ ] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#dev-scripts) raises no violations


python/tests/integration/connectors/memory/test_qdrant_memory_store.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+@pytest.mark.asyncio
+async def test_does_collection_exist_async():
+    qdrant_mem_store = QdrantMemoryStore(vector_size=TEST_VECTOR_SIZE, local=True)
+
+    await qdrant_mem_store.create_collection_async("test_collection")
+    result = await qdrant_mem_store.does_collection_exist_async("test_collection")
+    assert result is True
+    result = await qdrant_mem_store.does_collection_exist_async("test_collection2")
```
dmytrostruk, line 112: In test `test_get_collections_async` we are creating collection with name `test_collection2`, but not removing it. In this test we expect `False` from `does_collection_exist_async` method for collection with the same name `test_collection2`. I'm wondering, what would be the behavior if we run tests in parallel, is this a case? There is a chance that `does_collection_exist_async` will actually return `True`, so the test will fail.

We can add logic that will clean all collections before each test or use random names for test collections (e.g. using GUIDs) so the tests won't overlap. 

Please let me know if I missed something.

dmytrostruk (approved):
Left one comment regarding tests, but in general LGTM!

python/tests/integration/connectors/memory/test_qdrant_memory_store.py
awharrison-28: We certainly could randomize collection names, but I wasn't too worried about it here because the Qdrant integration tests use a local, embedded client that does not persist between test runs, nor is it accessible outside of the test runner. 

If were to change this and point the tests to a standalone Qdrant service, I would completely agree with adding more cleanup logic.

dmytrostruk (approved):

dmytrostruk (approved):

PR #2147, opened 2023-07-23
.Net: Expose the token counter in the TextChunker

### LINKED ISSUE #2146: .Net: Modify the TextChunker to allow for the caller to pass token count logic
The TextChunker uses a rough estimate (length divided by 4) to determine the token length of the input. The token counter logic should be exposed to be caller configurable. This way, token counts can be calculated based on the model that will be used to consume the chunk.

anthonypuppo:
### Motivation and Context

The TextChunker currently uses a rough estimate of the total number of tokens in a string (input divided by 4). This PR aims to expose the counter logic to be caller configurable, enabling accurate token counts based on which model will be consuming the chunks.

Fixes #2146 

### Description

Adds an optional token counter function parameter to the TextChunker methods. If not supplied, it falls back to the base token counter logic.

### Contribution Checklist


dotnet/src/SemanticKernel/Text/TextChunker.cs
```diff
@@ -259,10 +273,8 @@ private static (List<string>, bool) Split(ReadOnlySpan<char> input, string? inpu
         return (result, inputWasSplit);
     }
 
-    private static int TokenCount(int inputLength)
+    private static int FallbackTokenCounter(string input)
```
shawncal, line 276: Can you please rename this to "DefaultTokenCounter"?

shawncal:
Cool change. I'll take a closer look in a bit. Also requesting review from @RogerBarreto

dotnet/src/SemanticKernel/Text/TextChunker.cs
anthonypuppo: Done! Thanks for the tip, that method name is much better than "Fallback".

dotnet/src/SemanticKernel/Text/TextChunker.cs
```diff
@@ -87,14 +101,14 @@ private static List<string> InternalSplitTextParagraphs(List<string> lines, int
...
-        var paragraphs = BuildParagraph(truncatedLines, adjustedMaxTokensPerParagraph, longLinesSplitter);
+        var paragraphs = BuildParagraph(truncatedLines, adjustedMaxTokensPerParagraph, longLinesSplitter, tokenCounter);
         // distribute text more evenly in the last paragraphs when the last paragraph is too short.
         if (paragraphs.Count > 1)
         {
             var lastParagraph = paragraphs[paragraphs.Count - 1];
             var secondLastParagraph = paragraphs[paragraphs.Count - 2];
 
-            if (TokenCount(lastParagraph.Length) < adjustedMaxTokensPerParagraph / 4)
+            if (tokenCounter(lastParagraph) < adjustedMaxTokensPerParagraph / 4)
```
lemillermicrosoft, line 111: I'm curious now if we were ever supposed to be doing the division by 4 here -- I can't recall the reasoning....

dotnet/src/SemanticKernel/Text/TextChunker.cs
```diff
@@ -59,13 +70,16 @@ public static List<string> SplitPlainTextParagraphs(List<string> lines, int maxT
     /// <param name="lines">Lines of text.</param>
     /// <param name="maxTokensPerParagraph">Maximum number of tokens per paragraph.</param>
     /// <param name="overlapTokens">Number of tokens to overlap between paragraphs.</param>
+    /// <param name="tokenCounter">Function to count tokens in a string. If not supplied, the default counter (input length divided by 4) will be used.</param>
```
lemillermicrosoft, line 73: nit: I don't love leaving the implmentation details in the comments as it could change and become out of sync.

dotnet/src/SemanticKernel/Text/TextChunker.cs
```diff
@@ -129,14 +143,14 @@ private static List<string> InternalSplitTextParagraphs(List<string> lines, int
...
-    private static List<string> BuildParagraph(IEnumerable<string> truncatedLines, int maxTokensPerParagraph, Func<string, int, List<string>> longLinesSplitter)
+    private static List<string> BuildParagraph(IEnumerable<string> truncatedLines, int maxTokensPerParagraph, Func<string, int, List<string>> longLinesSplitter, TokenCounter tokenCounter)
     {
         StringBuilder paragraphBuilder = new();
         List<string> paragraphs = new();
 
         foreach (string line in truncatedLines)
         {
-            if (paragraphBuilder.Length > 0 && TokenCount(paragraphBuilder.Length) + TokenCount(line.Length) + 1 >= maxTokensPerParagraph)
+            if (paragraphBuilder.Length > 0 && tokenCounter(paragraphBuilder.ToString()) + tokenCounter(line) + 1 >= maxTokensPerParagraph)
```
lemillermicrosoft, line 153: nit: You could wrap the delegate with a couple internal methods (see the Encode method support in GPT3Tokenizer).

lemillermicrosoft (commented):
A few comments. Also, it'd be nice to add or update the KernelSyntaxExamples to show how this could be used with the GPT3Tokenizer as an example. 

I'd love to add this to ConversationSummarySkill which uses the TextChunker-- but I don't yet have a great answer on how we can get the model being used from the SKContext to know which tokenizer to use. cc @shawncal

lemillermicrosoft:
One more thought: Please add some UTs to dotnet\src\SemanticKernel.UnitTests\Text\TextChunkerTests.cs

dotnet/src/SemanticKernel/Text/TextChunker.cs
anthonypuppo: It's a rough check to see if the last paragraph is too short (less than 1/4 of the maximum tokens per paragraph) and if so, validates and merges it with the second to last paragraph. Best to leave as is and potentially refactor once the TextChunker is enhanced with some more configuration options?

dotnet/src/SemanticKernel/Text/TextChunker.cs
anthonypuppo: Good point, removed.

anthonypuppo:
> Also, it'd be nice to add or update the KernelSyntaxExamples to show how this could be used with the GPT3Tokenizer as an example.

Added as example 55.

> One more thought: Please add some UTs to dotnet\src\SemanticKernel.UnitTests\Text\TextChunkerTests.cs

Added a test for each method overload. Assume the extras are redundant, but I can add if you think it's necessary.

dotnet/src/SemanticKernel/Text/TextChunker.cs
anthonypuppo: I think the GPT3Tokenizer Encode methods differ slightly in that they're publicly exposed and all call the same base method (not utilized again internally). These internal methods would need to be called quite a bit which would make even basic calls decently verbose?

Something like this...
```
private static int CountTokens(string input, TokenCounter tokenCounter) => tokenCounter(input);
private static int CountTokens(StringBuilder stringBuilder, TokenCounter tokenCounter) => tokenCounter(stringBuilder.ToString());
```

Seems overly complex for the gain.

dotnet/samples/KernelSyntaxExamples/Example55_TextChunker.cs
```diff
@@ -0,0 +1,77 @@
...
+            if (i < paragraphs.Count - 1)
+            {
+                Console.WriteLine("------------------------");
+            }
+        }
+    }
+
+    private static int TokenCounter(string input)
+    {
+        var tokens = GPT3Tokenizer.Encode(input);
```
lemillermicrosoft, line 73: <3

lemillermicrosoft (approved):
Awesome, thanks for the contribution!

hario90 (approved):

PR #2154, opened 2023-07-24
.Net: Make MaxTokens nullable for consistency

### LINKED ISSUE #2153: .Net: MaxTokens nullable consistency
Make the maxTokens argument default to null at `InlineFunctionsDefinitionExtension.InvokeSemanticFunctionAsync` for consistency.

anthonypuppo:
### Motivation and Context

MaxTokens defaults to null throughout SK. The `InlineFunctionsDefinitionExtension.InvokeSemanticFunctionAsync` method should be updated to remain consistent.

Fixes #2153 

### Description

Make the `maxTokens` argument nullable at `InlineFunctionsDefinitionExtension.InvokeSemanticFunctionAsync`.

### Contribution Checklist


dotnet/src/SemanticKernel/SkillDefinition/InlineFunctionsDefinitionExtension.cs
```diff
@@ -126,7 +126,7 @@ public static class InlineFunctionsDefinitionExtension
         string? functionName = null,
         string? skillName = null,
         string? description = null,
-        int maxTokens = 256,
+        int? maxTokens = null,
```
dmytrostruk, line 129: I think we should be careful with this change as it can break our samples, where this parameter is not specified and also break our customers who uses default value. As soon as users will update package version, they may receive this exception in runtime:
https://github.com/microsoft/semantic-kernel/blob/f8b6e2af3f862f244ec9c380951a0e1ad5671eb3/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs#L360-L368

I think we can convert type to nullable but keep `256` as default value, since this method was added as helper to quickly create semantic functions inline without need to specify `maxTokens` every time.

shawncal: Yeah, we've been running into test breaks, planner breaks, and user complaints since we made other max tokens nullable and removed the defaults. Unfortunately, the service's defaults are inadequate for most interesting prompts.

Echoing @dmytrostruk, I'm happy to make it nullable, but we should keep a reasonable default -- otherwise, this is an issue that'll be almost immediately hit by all new users.

anthonypuppo: This was discussed in my original PR to make max tokens nullable (discussion here [here](https://github.com/microsoft/semantic-kernel/pull/1367#discussion_r1223582142)).

My stance is still the same. In my opinion, the entire purpose of an orchestration library like SK is to provide a generic interface for leveraging AI models in modern apps without breaking separation of concerns. If responses are being cut off that is the responsibility of the caller to understand the API they're consuming. If a default is going to be set, to me it makes more sense set it at the connector level so it can be more platform specific and not as high up the chain as the initial point of invocation.

With that being said, this is all just my opinion, if keeping the default at 256 makes more sense I can refactor 😄.

shawncal: I can confidently that, since we removed the default values, we've had lots of new issues and complaints. The problem is, as a nullable and *optional* arg, it appears as though it's... optional. That is, it should work fine without it. But for most cases, including our own planners, integration test, etc. it doesn't work at all. The defaults for some models are so low, they're effectively useless. When you hit this (right away), you really gotta start surfing your intellisense to find that sneaky optional arg to get yourself unblocked -- not the idea first-run experience.

I agree that we shouldn't be the ones trying to pick a default that works for everyone, but we do seek to make LLM usage turnkey and easy.

This specific extension method being the one where we choose a default value -- probably not quite the right place. I'd expect it to be downstream somewhere. i.e. you can pass null all the way down to the AI Connector, but if it encounters it there, that's where we might set a sensible default value. That's especially true if we know that the connector's API/models have a very low default.

anthonypuppo: > I can confidently that, since we removed the default values, we've had lots of new issues and complaints. The problem is, as a nullable and _optional_ arg, it appears as though it's... optional. That is, it should work fine without it. But for most cases, including our own planners, integration test, etc. it doesn't work at all. The defaults for some models are so low, they're effectively useless. When you hit this (right away), you really gotta start surfing your intellisense to find that sneaky optional arg to get yourself unblocked -- not the idea first-run experience.

This is really all a moot point with OpenAI announcing their completion based models as legacy. Once the examples are moved off and reference the chat models the API defaults to whatever's left within the models max token window after the input is processed. If anything, the hardcoded token values will do more harm than good in not allowing the model to max out it's context window.

> I agree that we shouldn't be the ones trying to pick a default that works for everyone, but we do seek to make LLM usage turnkey and easy.

Agreed, but this should be on the vendor, not the abstraction library.

> This specific extension method being the one where we choose a default value -- probably not quite the right place. I'd expect it to be downstream somewhere. i.e. you can pass null all the way down to the AI Connector, but if it encounters it there, that's where we might set a sensible default value. That's especially true if we know that the connector's API/models have a very low default.

I think this makes the most sense. To your point of nullable vs optional here I'm starting to agree more and more that a mandatory _nullable_ is the way to go. Not having it optional at least forces the developer acknowledge the value exists, even if null gets passed in every time. Very much future proof and vendor agnostic.

Going to close this as I think this is part of a larger discussion/plan/refactor. I'll bring it up in the next community office hours meeting.

PR #2167, opened 2023-07-25
.Net: [DRAFT] SKContext abstract base class

shawncal:
### Description
Creating an abstract base class for SKContext. The implementation is internal. This hides the constructor, discouraging creation of `new SKContext()`, as this is meant to be created and managed by the Kernel.

Deprecating additional members of SKContext, that appear to not be used, even in the repo samples.

Updating tests and samples accordingly.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -82,118 +54,78 @@ public CultureInfo Culture
...
     /// </summary>
-    /// <param name="skillName">Skill name</param>
-    /// <param name="functionName">Function name</param>
-    /// <returns>Delegate to execute the function</returns>
-    public ISKFunction Func(string skillName, string functionName)
-    {
-        return this.Skills.GetFunction(skillName, functionName);
-    }
+    /// <returns>Processed input, aka result, or last exception message if any</returns>
+    public abstract override string ToString();
```
stephentoub, line 79: This was already virtual, and making this abstract forces the derived class to override it... is that the intent?

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -82,118 +54,78 @@ public CultureInfo Culture
...
 
     /// <summary>
-    /// App logger
+    /// Create a clone of the current context, using the same kernel references (memory, skills, logger)
```
stephentoub, line 82: Is this comment still accurate? Wasn't memory removed?

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
```diff
@@ -0,0 +1,126 @@
...
+internal sealed class DefaultSKContext : SKContext
+{
+    private CultureInfo _culture;
+    private bool _errorOccurred = false;
+    private Exception? _lastException;
+    private string _lastErrorDescription = string.Empty;
+    private IReadOnlyCollection<ModelResult> _modelResults;
+    private ContextVariables _variables;
+    private IReadOnlySkillCollection _skills;
+    private ILogger _logger;
```
stephentoub, line 26: Some these can be readonly now

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
```diff
@@ -0,0 +1,126 @@
...
+    internal static SKContext FromException(Exception exception)
+    {
+        DefaultSKContext context = new();
+        context.Fail(exception.Message, exception);
+        return context;
+    }
+
+    public override string Result => this.Variables.ToString();
+
+    public override bool ErrorOccurred => this._lastException != null;
```
stephentoub, line 37: Is this property even necessary if all it's doing is seeing whether LastException is null?

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    public override string LastErrorDescription => this._lastErrorDescription;
```
stephentoub, line 39: Even if we keep the error-related stuff on SKContext, do we need both LastErrorDescription and LastException?  If someone wants to provide a description, can they just provide an exception instead with the description as the message?

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    public override Exception? LastException => this._lastException;
+
+    public override IReadOnlyCollection<ModelResult> ModelResults => this._modelResults;
+
+    public override CultureInfo Culture
+    {
+        get => this._culture;
+    }
```
stephentoub, line 48: Nit:
```suggestion
    public override CultureInfo Culture => this._culture;
```

dotnet/src/InternalUtilities/test/MockSKContext.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.SemanticKernel.Orchestration;
+using Microsoft.SemanticKernel.SkillDefinition;
+
+namespace SemanticKernel.UnitTests;
+
+internal sealed class MockSKContext : SKContext
+{
+    private CultureInfo _culture;
+    private bool _errorOccurred = false;
```
stephentoub, line 16: Where is this read? Seems like it can be deleted?

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Semantic Kernel context.
+/// </summary>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+internal sealed class DefaultSKContext : SKContext
+{
+    private CultureInfo _culture;
+    private bool _errorOccurred = false;
```
stephentoub, line 20: Where is this read? Seems like it can be deleted?

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -182,27 +182,25 @@ public sealed class SKFunction : ISKFunction, IDisposable
...
+                contextResult.Variables.Update(completion);
+                return contextResult;
             }
             catch (AIException ex)
             {
                 const string Message = "Something went wrong while rendering the semantic function" +
                                        " or while executing the text completion. Function: {0}.{1}. Error: {2}. Details: {3}";
                 logger?.LogError(ex, Message, skillName, functionName, ex.Message, ex.Detail);
-                context.Fail(ex.Message, ex);
+                return DefaultSKContext.FromException(ex);
```
stephentoub, line 195: The returned context doesn't need to preserve things like input variables?

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
shawncal: That was the intent, only because there's this very specific description in the comments of what "ToString" should to. That said, I don't like relying on ToString for anything specific, so I'll just remove the whole method from the base.

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
shawncal: No, we do absolutely do not. I had the same thought -- we should just remove this with the Exceptions work.

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
shawncal: Oh, I removed this, then restored it because in the current "Fail()" function, the Exception parameter is optional. My shortcut of ErrorOccurred => _errorOccurred not null; wont work if you call Fail(string).
- Fail() should go away too.

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
shawncal: Unclear. This is a stepping stone change to one where we replace the return type of all kernel operations, SKContext, with something specifically resulty, like SKResult.

Good point though, I'll look at keeping as much in the context as possible (for now), even in a Fail condition.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -15,65 +13,39 @@ namespace Microsoft.SemanticKernel.Orchestration;
...
     /// Error details.
     /// </summary>
-    public string LastErrorDescription { get; private set; } = string.Empty;
+    public abstract string LastErrorDescription { get; }
 
     /// <summary>
     /// When an error occurs, this is the most recent exception.
     /// </summary>
-    public Exception? LastException { get; private set; }
+    public abstract Exception? LastException { get; }
```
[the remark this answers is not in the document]
shawncal: No, this is just to minimize churn until new mechanism is in place. If we agree that we still need an "error info" section at all, I except it will not be in SKContext, but in a new SKResult (name tbd) class that strictly captures execution output. We'll follow up offline.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
shawncal: Wow, no, is it ever not accurate. I've reduced to "Creates a clone of the current context".
I believe the Clone() method can also be obsoleted (at least on the public base class) once Kernel itself is responsible for instantiating and managing all Context objects. TBD whether Planner extensions will still require this.

dotnet/src/SemanticKernel/Orchestration/DefaultSKContext.cs
shawncal: No, this was just intended to keep the current SDK consistent until we overhaul the exceptions. Then it should all go away.

PR #2188, opened 2023-07-26
.Net: Obsolete SKContext[...] indexer

shawncal:
### Motivation and Context
The SKContext[...] indexer is simply shorthand for SKContext.Variables[...]. With additional routes to doing the same thing, there's more confusion about which to use when. In this case, there's a great deal of confusion about the mutability of these variables and whether they can/should be changed within functions or between functions.

### Description
This PR starts to address the above issues by simplifying the SDK surface. The first step is to remove the "extra" route to accessing the same data, the indexer on SKContext. 

**This is a breaking change.** The mitigation is simply to update all uses of `context["MY_VALUE"]` to `context.Variables["MY_VALUE"]`

In the PRs that follow, the mutability issue will be addressed by making it clear which variables are input, how and where to set output, and which can/cannot be transformed by an SKFunction.

### Contribution Checklist
   => This is a breaking change.

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
@@ -30,31 +30,10 @@ public sealed class SKContext
     public string Result => this.Variables.ToString();
 
     /// <summary>
-    /// Whether an error occurred while executing functions in the pipeline.
-    /// </summary>
-    public bool ErrorOccurred { get; private set; }
-
-    /// <summary>
-    /// Error details.
-    /// </summary>
-    public string LastErrorDescription { get; private set; } = string.Empty;
-
-    /// <summary>
-    /// When an error occurs, this is the most recent exception.
-    /// </summary>
-    public Exception? LastException { get; private set; }
-
-    /// <summary>
     /// When a prompt is processed, aka the current data after any model results processing occurred.
     /// (One prompt can have multiple results).
     /// </summary>
     public IReadOnlyCollection<ModelResult> ModelResults { get; set; } = Array.Empty<ModelResult>();
-
-    /// <summary>
-    /// The token to monitor for cancellation requests.
-    /// </summary>
-    [Obsolete("Add a CancellationToken param to SKFunction method signatures instead of retrieving it from SKContext.")]
-    public CancellationToken CancellationToken { get; } = default;
 
     /// <summary>
     /// The culture currently associated with this context.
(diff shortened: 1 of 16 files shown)


dotnet/src/Skills/Skills.Core/HttpSkill.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.SemanticKernel.Skills.Core;
 /// <example>
 /// Usage: kernel.ImportSkill("http", new HttpSkill());
 /// Examples:
-/// SKContext["url"] = "https://www.bing.com"
+/// SKcontext.Variables["url"] = "https://www.bing.com"
```
stephentoub, line 17: I don't understand the casing change here.  Is there something that's actually cased as "SKcontext"?

stephentoub (dismissed):

dotnet/src/Skills/Skills.Core/HttpSkill.cs
shawncal: Nope, just a typo in the comments. Will fix

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -196,4 +140,65 @@ private string DebuggerDisplay
...
+    /// <summary>
+    /// The token to monitor for cancellation requests.
+    /// </summary>
+    [Obsolete("Add a CancellationToken param to SKFunction method signatures instead of retrieving it from SKContext.")]
+    public CancellationToken CancellationToken { get; } = default;
+
+    /// <summary>
+    /// Semantic memory
+    /// </summary>
+    [Obsolete("Memory no longer passed through SKContext. Instead, initialize your skill class with the memory provider it needs.")]
```
markwallace-microsoft, line 197: It would be nice to reference the kernel syntax sample that shows the new pattern so that there is no confusion about what change needs to be made.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -196,4 +140,65 @@ private string DebuggerDisplay
...
+    #region Error handling
+    /// <summary>
+    /// Whether an error occurred while executing functions in the pipeline.
+    /// </summary>
+    public bool ErrorOccurred { get; private set; }
+
+    /// <summary>
+    /// Error details.
+    /// </summary>
+    public string LastErrorDescription { get; private set; } = string.Empty;
```
markwallace-microsoft, line 153: Should this be nullable?

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -196,4 +140,65 @@ private string DebuggerDisplay
             return display;
         }
     }
+
+    #region Error handling
+    /// <summary>
+    /// Whether an error occurred while executing functions in the pipeline.
+    /// </summary>
+    public bool ErrorOccurred { get; private set; }
```
RogerBarreto, line 148: Would remove completely the ErrorOccurred, LastErrorDescription and Leave only the `LastException`, if `LastException` is not null (Alone this means an ErrorOccurred) + LastErrorDescription = (`LastException.Message`)

shawncal: Yup, agreed. (We were also discussing in another related PR). For this change, I'm trying to keep it focused to just the indexer[] and leave the rest of the API as intact as possible. Going forward, I expect ALL of these exception fields will be removed from the SKContext. Instead, they'll either throw directly (critical exceptions) or, if we do want to capture error conditions in the functions, the LastException (plus Result/ModelResults) should be moved to a new `SKReturn` class.

RogerBarreto (dismissed):

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
shawncal: Good call. Keeping this PR focused on the indexer[] (this part of the diff is just moving the exception-related stuff around within the file). Making it nullable does trigger a few warnings scattered in files throughout the repo, so I'll make that one separately.

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
shawncal: Created issue #2193 to track. I'll try to get that one out today.

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
shawncal: This Memory obsoleted thing is already in and released (this PR just moves the method within the file). But your point stands... should point at new use pattern where applicable.
In this case, I don't think we have any examples other than the Planners, of functions using Memory. That's one of the reasons this deprecation was a fairly easy choice to make; I don't believe this access pattern has much usage.

PR #2198, opened 2023-07-27
.Net: Obsolete SKContext.Fail, .Func, .LastErrorDescription

### LINKED ISSUE #2191: .Net: Add [EditorBrowsable(EditorBrowsableState.Never)] on Obsolete properties.

### LINKED ISSUE #2193: .Net: Update SKContext.LastErrorDescription to string?

shawncal:
### Description
In the effort to clean up SDK and access patterns, we are deprecating, and adding the [Obsolete] attribute to, the following `SKContext` members:
- `LastErrorDescription`
  - use SKContext.LastError.Message instead
- `Fail( ... )`
  - encourage SKFunction developers to throw in the function body instead 
- `Func( ... )`
  - unused

Also included:
- adding `[EditorBrowsable(EditorBrowsableState.Never)]` to all obsoleted members.
- updating callers (tests, samples)
- moving obsoleted members to a `#region Obsolete` at the bottom of each file.

Fixes #2193
Fixes #2191

### Contribution Checklist
   => This is a BREAKING CHANGE


dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -151,60 +132,90 @@ private string DebuggerDisplay
     /// <summary>
     /// Whether an error occurred while executing functions in the pipeline.
     /// </summary>
-    public bool ErrorOccurred { get; private set; }
-
-    /// <summary>
-    /// Error details.
-    /// </summary>
-    public string LastErrorDescription { get; private set; } = string.Empty;
+    public bool ErrorOccurred => this.LastException != null;
```
dmytrostruk, line 135: nit: Maybe we can rename this property to `ExceptionOccurred` for better consistency.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -200,15 +197,15 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
...
+                        pipelineStepCount, f.SkillName, f.Name, context.LastException?.Message);
                     return context;
                 }
             }
             catch (Exception e) when (!e.IsCriticalException())
             {
                 this.Logger.LogError(e, "Something went wrong in pipeline step {0}: {1}.{2}. Error: {3}",
                     pipelineStepCount, f.SkillName, f.Name, e.Message);
-                context.Fail(e.Message, e);
+                context.LastException = e;
```
dmytrostruk, line 208: Do we want to re-throw it here?

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -231,24 +231,20 @@ public async Task<SKContext> InvokeAsync(SKContext context, CompleteRequestSetti
...
-                this._logger.LogError(e, Message, this._function.Method.Name, e.Message);
-                context.Fail(e.Message, e);
-            }
+            return await this._function(null, settings, context, cancellationToken).ConfigureAwait(false);
+        }
+        catch (Exception e) when (!e.IsCriticalException())
+        {
+            const string Message = "Something went wrong while executing the native function. Function: {0}. Error: {1}";
+            this._logger.LogError(e, Message, this._function.Method.Name, e.Message);
+            context.LastException = e;
```
dmytrostruk, line 245: Same here, it looks like in some places we re-throw and in some places we just set `LastException` property. Not sure it's the scope of this PR though :)

dmytrostruk (dismissed):

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
shawncal: Yeah, I ran into this too. My first instinct was to just throw here, but that broke some of the UTs because they're expecting to successfully complete the "InvokeAsync" call and then check for failure in the context. I decided to keep the same functionality for now, and to overhaul it when we introduce the SKResult type. I also want to coordinate with @SergeyMenshykh about how such errors should propagate, and not overlap with his PR.

dotnet/src/SemanticKernel/Kernel.cs
shawncal: Same as below... that changes the flow from the caller perspective: currently, a failure in the SK function will allow the RunAsync() call to complete, then the caller can check the ErrorOccurred property. If we throw here, the RunAsync call itself will fail, which means that they won't get a chance to get to the ErrorOccurred property.

We may need to overhaul this with a breaking change. To coordinate with @SergeyMenshykh, and will address this if needed once his exception handling PR is complete.

dmytrostruk (approved):

PR #2203, opened 2023-07-27
Java: Add more setting options to context variable builder

johnoliver:
- Allows users to set values in the ContextVariable builder
- Allows constructing a mutable ContextVariable

### DIFF
--- a/java/semantickernel-core/src/main/java/com/microsoft/semantickernel/orchestration/DefaultContextVariables.java
+++ b/java/semantickernel-core/src/main/java/com/microsoft/semantickernel/orchestration/DefaultContextVariables.java
@@ -148,19 +148,39 @@ class DefaultContextVariables implements ContextVariables, WritableContextVariab
 
     public static class Builder implements ContextVariables.Builder {
 
+        private final Map<String, String> variables;
+
+        public Builder() {
+            variables = new CaseInsensitiveMap<>();
+            this.variables.put(MAIN_KEY, "");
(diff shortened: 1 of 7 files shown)


shawncal:
@johnoliver Let's sync offline about current thinking for ContextVariables, mutability. There's an opportunity to align plans for .net, java here before we implement changes.

java/semantickernel-core/src/main/java/com/microsoft/semantickernel/orchestration/DefaultContextVariables.java
```diff
@@ -148,19 +148,39 @@ public WritableContextVariables build(Map<String, String> map) {
...
 
+        private final Map<String, String> variables;
+
+        public Builder() {
+            variables = new CaseInsensitiveMap<>();
+            this.variables.put(MAIN_KEY, "");
+        }
+
+        @Override
+        public WritableContextVariables buildWritable() {
```
dsgrieve, line 159: I don't think this is needed. One can do `new Builder().build().writableClone()` 
In my opinion, this builder would be better if the API had only this:
```
        ContextVariables build();
        ContextVariables withInput(String input);
        ContextVariables withVariables(Map<String, String> map);
        ContextVariables withVariable(String key, String value);
```

dsgrieve (changes requested):

dsgrieve (approved):

PR #2205, opened 2023-07-27
.Net: Introduce new pattern for creating a semantic function

markwallace-microsoft:
### Motivation and Context

  1. Why is this change required? Enhancements to the current patterns used to create a Semantic Functions.

The PR contains an ADR which answer the above questions.

### Description

1. Introduces the following new classes:
    1. `PromptConfig` - Defines a prompt (aka semantic function) and the capabilities we want a prompt to support.
    2. `PromptRequestSettings` - Defines the request settings used to execute the prompt with a specific model in an AI agnostic fashion.
2. Mark `PromptTemplateConfig` and `SemanticFunctionConfig` as obsolete.
5. Using the new code path the function does not have a default text completion service associated with it. Instead this is determined when the function is invoked.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptConfig.cs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// <summary>
+    /// Settings for a prompt request. This supports text completions, image generation, etc.
+    /// </summary>
+    public class PromptRequestSettings
+    {
+        /// <summary>
+        /// Order for this model.
+        /// </summary>
+        public int? order { get; set; }
```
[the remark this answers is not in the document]
markwallace-microsoft: Hi @SergeyMenshykh 
I'll fix the capitalisation. Order won't be the only option used to determine which model is selected but supporting different strategies to select the model is not part of this PR it will come later.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptConfig.cs
```diff
@@ -0,0 +1,278 @@
...
+
+/// <summary>
+/// Prompt configuration.
+/// </summary>
+public class PromptConfig
+{
+    /// <summary>
+    /// Input parameter for semantic functions.
+    /// </summary>
+    public class InputParameter
```
dmytrostruk, line 26: As an example, regarding naming, here we have `prompt` configuration, but input parameter for `semantic functions`. For users it may be confusing and not clear whether it's the same thing or not. I would probably try to align the naming in scope of this refactoring/redesign.

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptConfig.cs
```diff
@@ -0,0 +1,278 @@
...
+    /// </summary>
+    public class PromptRequestSettings
+    {
+        /// <summary>
+        /// Identifier for an AI model.
+        /// </summary>
+        public string? ModelId { get; set; }
+
+        /// <summary>
+        /// Identifier for a AI service provider. This is deployment agnostic.
```
dmytrostruk, line 62:
```suggestion
        /// Identifier for an AI service provider. This is deployment agnostic.
```

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptConfig.cs
```diff
@@ -0,0 +1,278 @@
...
+        /// </summary>
+        public JsonObject Properties { get; set; } = new();
+    }
+
+    /// <summary>
+    /// Prompt version
+    /// </summary>
+    [JsonPropertyName("promptVersion")]
+    [JsonPropertyOrder(1)]
+    public int PromptVersion { get; set; } = 1;
```
dmytrostruk, line 77: Is it `prompt` version or `config` schema version? If it's schema version, I would probably rename this property to `ConfigVersion`, `SchemaVersion` or `ConfigSchemaVersion`. If it's prompt version, I'm not sure what does it mean :)

docs/decisions/0005-create-semantic-function.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+#### Changes:
+
+The code looks similar to the current pattern but has the following key differences:
+
+1. Introduces the following new classes:
+    1. `PromptConfig` - Defines a prompt (aka semantic function) and the capabilities we want a prompt to support.
+    2. `PromptRequestSettings` - Defines the request settings used to execute the prompt with a specific model in an AI agnostic fashion.
+2. Mark `PromptTemplateConfig` and `SemanticFunctionConfig` as obsolete.
```
dmytrostruk, line 105: Looks like in actual code these classes are not marked as obsolete, is there a plan to do it in follow-up PR?

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptConfig.cs
```diff
@@ -0,0 +1,278 @@
...
+            // when the semantic function is invoked.
+            var defaultSettings = this.RequestSettings[0];
+            return defaultSettings.Properties.ToJson();
+        }
+
+        return null;
+    }
+
+    /// <summary>
+    /// Convert SemanticFunctionConfig to the new abstraction PromptConfig
```
dmytrostruk, line 234: nit: If you use `see` attribute for class names, it will be easier to maintain XML documentation in the future, when you rename some class or delete it for example. Same applies to other places in the code where class names are used. Can be done in follow-up PRs.
```suggestion
    /// Convert <see cref="SemanticFunctionConfig"/> to the new abstraction <see cref="PromptConfig"/>
```

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -194,7 +194,15 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
...
-                context = await f.InvokeAsync(context, cancellationToken: cancellationToken).ConfigureAwait(false);
+
+                if (f.IsSemantic && f is SKFunction func && !func.HasAIService())
+                {
+                    context = await f.InvokeAsync(context, textCompletion: this.GetService<ITextCompletion>(), cancellationToken: cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    context = await f.InvokeAsync(context, cancellationToken: cancellationToken).ConfigureAwait(false);
+                }
```
dmytrostruk, line 205: I think we should investigate in more detail how to inject AI service based on future functionality like:
1. Allowing different types of AI services (not only text completion).
2. Fallback logic when first AI service is unresponsive etc.

Also, I'm wondering what was the reason of adding `ITextCompletion` parameter to `InvokeAsync` function? Because in this case we need to call `InvokeAsync` with different set of parameters, which makes it a little bit complicated, especially if the list of parameters will grow.

For example, we can inject AI service with `f.SetAIService(() => this.GetService<ITextCompletion>());` and use the same `InvokeAsync` signature which will simplify the code and remove the need to update all unit tests to add new `ITextCompletion` parameter.

docs/decisions/0005-create-semantic-function.md
```diff
@@ -0,0 +1,141 @@
...
+{
+    Template: "Say hello in German",
+    Description = "Say hello in German",
+    InputParameters = new(),
+    skillName: "GermanSkill",
+    functionName: "SayHello",
+    PromptRequestSettings = new List<PromptRequestSettings>() { … }
+};
+
+var function = kernel.CreateSemanticFunction(promptConfig);
```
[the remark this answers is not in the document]
markwallace-microsoft: The current implementation uses the `SemanticFunctionConfig` class name. I can change this implementation to reuse that class and modify it to support the new functionality.
So instead of introducing `PromptConfig` I would make some changes to `SemanticFunctionConfig` and deprecate some of the properties it supports. That's how the first version of the changes worked but there was feedback that a new implementation might be cleaner. I'm happy to revert the implementation, @dmytrostruk @shawncal ?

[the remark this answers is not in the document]
dmytrostruk: Got it, thanks. I agree that it would be better and easier to add improvements from scratch rather than modifying already existing classes, but we should also think about naming and whether we want to keep previous naming or introduce a new one.

PR #2210, opened 2023-07-27
Java: Update jbang configuration in samples 

brunoborges:


java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/Example00_GettingStarted.java
```diff
@@ -1,7 +1,8 @@
 ///usr/bin/env jbang "$0" "$@" ; exit $?
-//DEPS com.microsoft.semantic-kernel:semantickernel-core:0.2.6-alpha
-//DEPS com.microsoft.semantic-kernel:semantickernel-core-skills:0.2.6-alpha
-//DEPS com.microsoft.semantic-kernel.connectors:semantickernel-connectors:0.2.6-alpha
+// Note: to run this with jbang, you first must run `./mvnw install` in the root of the java project.
+//DEPS com.microsoft.semantic-kernel:semantickernel-core:0.2.8-alpha-SNAPSHOT
```
brunoborges, line 3: I wonder if there is a way for the maven-release process to update this value automatically for us.

Meaning: every time we release, it will remove the SNAPSHOT so we get the code with the actual released version in the tag (and on Maven Source JAR), and then updated to the next SNAPSHOT version for the post-release update.

@joe-braley @Luigi96  any ideas?

dsgrieve: I've done this in gradle.

I have a version.properties which contains `version=${version ?: 'unknown'}`
processResources will expand ${version} when the resource is getting copied from the src tree to the build tree. In this particular case, if there is no "version" property in the project, version is set to 'unknown'. 

I imagine there is something similar in maven.
```
processResources {
    include '**/version.properties'
    expand(version: project.property("version"))
}
```

```
compileJava {
    dependsOn processResources
}
```

dsgrieve (approved):

PR #2211, opened 2023-07-27
.Net: Redis memory store constructor additions

### LINKED ISSUE #2208: .Net: improve Redis memory store constructor
As discussed in #2132 , it would be helpful to allow the user to customize the vector attributes used in similarity searching, and also create a `RedisMemoryStore` instance from a connection string than having to make the database instance first.

I will implement these features by adding parameters to the constructor and overloading the existing constructor, thank you!

sneha-afk:
### Motivation and Context

RediSearch allows for vector similarity queries with a variety of settings. Previously, the C# Redis connector defaulted all the vector settings other than the vector size. Also, users only had the option of passing their own `IDatabase` instance into the constructor.

Closes #2208 
### Description

1. Added vector attributes to the `RedisMemoryStore` constructor: type, indexing algorithm, distance metric, and query dialect. These maintain the previous defaults but are now parameters that can be changed.
2. Added a new constructor that takes in a connection string to a server instead of a database instance

### Contribution Checklist

Note: I get `Warnings were encountered while loading the workspace. Set the verbosity option to the 'diagnostic' level to log warnings.` when I try to run `dotnet format`. This occurs on a clean clone of the repo as well. The verbose output mentions how each project in the repo has its own `.editorconfig`, so no files are formatted.


dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
```diff
@@ -200,31 +232,9 @@ public async Task RemoveBatchAsync(string collectionName, IEnumerable<string> ke
...
-    /// <summary>
-    /// Supported distance metric, one of {L2, IP, COSINE}. The default value is "COSINE".
-    /// </summary>
-    internal const string VECTOR_DISTANCE_METRIC = "COSINE";
-
-    /// <summary>
-    /// Query dialect. To use a vector similarity query, specify DIALECT 2 or higher. The default value is "2".
-    /// <see href="https://redis.io/docs/stack/search/reference/vectors/#querying-vector-fields"/>
-    /// </summary>
-    internal const int QUERY_DIALECT = 2;
```
dmytrostruk, line 223: I think we should keep these constants. It would be great to do the following:
1. Convert access modifiers from `internal` to `private`.
2. Change naming to follow .NET naming convention (e.g. `QUERY_DIALECT` => `QueryDialect`)
3. Use it in constructors as default parameters (e.g. `RedisMemoryStore(..., int queryDialect = QueryDialect`)

Same applies to all other constants in this class.

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
```diff
@@ -27,14 +27,46 @@ public sealed class RedisMemoryStore : IMemoryStore
...
+    /// <summary>
+    /// Create a new instance of semantic memory using Redis.
+    /// <see href="https://redis.io/docs/interact/search-and-query/search/vectors/#creation-attributes-per-algorithm"/>
+    /// <see href="https://redis.io/docs/interact/search-and-query/search/vectors/#querying-vector-fields"/>
+    /// </summary>
+    /// <param name="connectionString">Provide connection URL to a Redis instance</param>
+    /// <param name="vectorSize">Embedding vector size, defaults to 1536</param>
+    /// <param name="vectorIndexAlgorithm">Indexing algorithm for vectors, defaults to "HNSW"</param>
+    /// <param name="vectorType">Vector type, defaults to "FLOAT32"</param>
+    /// <param name="vectorDistanceMetric">Metric for measuring vector distances, defaults to "COSINE"</param>
```
dmytrostruk, line 59: I'm wondering how we can minimize error probability for these parameters. Since they are strings, user can pass any value. 
1. We can list all possible values in XML documentation, so user will know which values can be accepted.
2. We can convert their type from `string` to `Enum`, in this case it won't be possible to pass incorrect value. 

I would probably go with `Enum` option in C#. It has disadvantages as well, for example as soon as Redis will accept new value, we will have to add new enum value. But looking at nature of these parameters, I don't think they will be updated a lot. While this will make our class more stable.

sneha-afk: Good idea, we can use `Enum.TryParse` with the `Schema.VectorField.VectorAlgo`  for interoperability with Redis, since they define VectorAlgo with an enum as well.
I'll add enums to a new file called `RedisVectorAttributes.cs` so its easier to find valid values.

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
```diff
@@ -28,13 +28,58 @@ public sealed class RedisMemoryStore : IMemoryStore
...
+        VectorIndexAlgorithms vectorIndexAlgorithm = VectorIndexAlgorithm,
+        VectorTypes vectorType = VectorType,
+        VectorDistanceMetrics vectorDistanceMetric = VectorDistanceMetric,
+        int queryDialect = QueryDialect)
     {
         this._database = database;
         this._vectorSize = vectorSize;
         this._ft = database.FT();
+        // Default to HNSW if parsing error occurs
+        if (!Enum.TryParse<Schema.VectorField.VectorAlgo>(vectorIndexAlgorithm.ToString(), out this._vectorIndexAlgorithm))
```
glahaye, line 47: Do we want to go to a default know value when we can't parse, or do we want to throw an exception so that the user knows right away that something is wrong?

sneha-afk: Good point, an exception would be more beneficial. Another missing exception handling is the vector size.

dotnet/src/Connectors/Connectors.Memory.Redis/RedisVectorAttributes.cs
dmytrostruk, line 1: I will have a couple of requests for enums, if possible:
1. Keep each enum in separate file.
2. Keep each enum values in separate line.
3. Add XML documentation for enum and enum values. It would be great if there will be direct link to Redis documentation.

Thanks a lot!

sneha-afk:
I have implemented the suggested changes, enums are in separate files with XML, and there is error handling for vector size and parsing into a different enum in the constructor.

To note, when I was testing the preexisting example file `Example47_Redis.cs` with different vector settings, the only one that does not properly index in the Redis database is when the vector type is FLOAT64 (double) since the current setup of memory records has the embedding as type float. When upserting a record, it is currently only supporting the float type. My C# knowledge is limited, but I assume there needs to be another intermediate conversion between a float to double if the user wants to store FLOAT64 vectors?
https://github.com/microsoft/semantic-kernel/blob/e00c0f3360d92c4c8e3abb7f90d4c336b0973f27/dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs#L167

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
```diff
@@ -28,13 +28,65 @@ public sealed class RedisMemoryStore : IMemoryStore
...
+        if (vectorSize <= 0)
+        {
+            throw new ArgumentException("Vector size must be a positive integer.");
+        }
+        if (!Enum.TryParse<Schema.VectorField.VectorAlgo>(vectorIndexAlgorithm.ToString(), out this._vectorIndexAlgorithm))
+        {
+            throw new ArgumentException("Unsupported vector indexing algorithm.");
+        }
+
+        this._database = ConnectionMultiplexer.Connect(connectionString).GetDatabase();
```
JadynWong, line 84: I have a question. Is it necessary to consider managing and dispose `ConnectionMultiplexer` instances?

JadynWong:
> I have implemented the suggested changes, enums are in separate files with XML, and there is error handling for vector size and parsing into a different enum in the constructor.
> 
> To note, when I was testing the preexisting example file `Example47_Redis.cs` with different vector settings, the only one that does not properly index in the Redis database is when the vector type is FLOAT64 (double) since the current setup of memory records has the embedding as type float. When upserting a record, it is currently only supporting the float type. My C# knowledge is limited, but I assume there needs to be another intermediate conversion between a float to double if the user wants to store FLOAT64 vectors?
> 
> https://github.com/microsoft/semantic-kernel/blob/e00c0f3360d92c4c8e3abb7f90d4c336b0973f27/dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs#L167


`MemoryRecord.Embedding` is fixed set to `Embedding<float>`. I think Embedding needs to be adjusted first if double types are to be supported.

https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecord.cs#L19

Also direct conversion of float to double results in loss of precision.

This may also require some conversion.
https://stackoverflow.com/questions/29492878/convert-float-to-double

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
sneha-afk: Do you mean keeping a reference to the `ConnectionMultiplexer` within the memory store instance? I would think if someone has more complex use for the `ConnectionMultiplexer` that they should use the constructor that takes in an `IDatabase` instance so they can do their own management of it.

sneha-afk:
> I think Embedding needs to be adjusted first if double types are to be supported.

Here's an approach I tried within `UpsertAsync`:
```c#
        byte[]? bytesEmbed = null;
        if (this._vectorType == "FLOAT32")
        {
            bytesEmbed = MemoryMarshal.Cast<float, byte>(record.Embedding.AsReadOnlySpan()).ToArray();
        }
        else if (this._vectorType == "FLOAT64")
        {
            double[] toDouble = MemoryMarshal.Cast<float, double>(record.Embedding.AsReadOnlySpan()).ToArray();
            bytesEmbed = MemoryMarshal.Cast<double, byte>(toDouble).ToArray();
        }

        await this._database.HashSetAsync(GetRedisKey(collectionName, record.Key), new[] {
            new HashEntry("key", record.Key),
            new HashEntry("metadata", record.GetSerializedMetadata()),
            new HashEntry("embedding", bytesEmbed),
            new HashEntry("timestamp", ToTimestampLong(record.Timestamp))
        }, flags: CommandFlags.None).ConfigureAwait(false);
```
However, the embedding bytes from the `double` conversion were not properly indexed within Redis, indicating this approach is not abiding by a FLOAT64 schema. Is there there anything I'm missing here? Thanks :)

JadynWong:
> > I think Embedding needs to be adjusted first if double types are to be supported.
> 
> Here's an approach I tried within `UpsertAsync`:
> 
> ```cs
>         byte[]? bytesEmbed = null;
>         if (this._vectorType == "FLOAT32")
>         {
>             bytesEmbed = MemoryMarshal.Cast<float, byte>(record.Embedding.AsReadOnlySpan()).ToArray();
>         }
>         else if (this._vectorType == "FLOAT64")
>         {
>             double[] toDouble = MemoryMarshal.Cast<float, double>(record.Embedding.AsReadOnlySpan()).ToArray();
>             bytesEmbed = MemoryMarshal.Cast<double, byte>(toDouble).ToArray();
>         }
> 
>         await this._database.HashSetAsync(GetRedisKey(collectionName, record.Key), new[] {
>             new HashEntry("key", record.Key),
>             new HashEntry("metadata", record.GetSerializedMetadata()),
>             new HashEntry("embedding", bytesEmbed),
>             new HashEntry("timestamp", ToTimestampLong(record.Timestamp))
>         }, flags: CommandFlags.None).ConfigureAwait(false);
> ```
> 
> However, the embedding bytes from the `double` conversion were not properly indexed within Redis, indicating this approach is not abiding by a FLOAT64 schema. Is there there anything I'm missing here? Thanks :)

```csharp
// see https://stackoverflow.com/questions/29492878/convert-float-to-double
// var toDouble = record.Embedding.Vector.Select(x=>(double)new decimal(x)).ToArray();
var toDouble = new[] { 0.1f, 0.2f, 0.3f }.Select(x => (double)new decimal(x)).ToArray();

var bytes = MemoryMarshal.Cast<double, byte>(toDouble).ToArray();

var doubles = MemoryMarshal.Cast<byte, double>(bytes).ToArray();

Debug.Assert(toDouble.SequenceEqual(doubles));
```

If you need to support `double` type, remember that everything involving `Embedding` type requires a similar conversion.

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
dmytrostruk: @sneha-afk You are right. For complex scenarios when developers want to control disposable `ConnectionMultiplexer`, they can use constructor that accepts `IDatabase` interface. We use similar approach in AI Connectors for `HttpClient`. It's possible to pass custom `HttpClient` to control object lifetime, or we will provide our own instance of `HttpClient` as default option.

dmytrostruk (approved):
@sneha-afk Thank you for testing `FLOAT64` vector type. @JadynWong has a good point - in `MemoryRecord` type `Embedding` accepts only `float` but not `double`. Which means that for Redis only `FLOAT32` option will work at the moment. I removed vector type enum and added `FLOAT32` as the only one default option in this commit: https://github.com/microsoft/semantic-kernel/pull/2211/commits/791f7aab70f54e013167fc2dc407614823c1118a

PR itself looks good to me!

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
JadynWong: In fact, I deliberately did not implement `IDisposable' at all, and did not allow the user to pass a connection string. The reason for this is mentioned in PR #1338, and I'm not sure which side of the issue this falls on. `IKernel` is registered as scoped lifetime with DI, `MemoryStore` is registered as singleton lifetime, and whenever `IKernel` is freed, it tries to dispose all internal instances that implement `IDisposable`. This results in the instances inside the MemoryStore being disposed of after the first request, unless the application is restarted.

https://github.com/microsoft/chat-copilot/blob/07f2fb09689aa007f22310db4769fe650d33af40/webapi/SemanticKernelExtensions.cs#L41

https://github.com/microsoft/chat-copilot/blob/main/webapi/SemanticKernelExtensions.cs#L99

https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel/Kernel.cs#L250

https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel/Memory/SemanticTextMemory.cs#L132

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
```diff
@@ -23,18 +23,71 @@ namespace Microsoft.SemanticKernel.Connectors.Memory.Redis;
...
+
+        if (!Enum.TryParse<Schema.VectorField.VectorAlgo>(vectorIndexAlgorithm.ToString(), out this._vectorIndexAlgorithm))
+        {
+            throw new ArgumentException("Unsupported vector indexing algorithm.");
+        }
+
         this._database = database;
         this._vectorSize = vectorSize;
         this._ft = database.FT();
+        this._vectorDistanceMetric = vectorDistanceMetric.ToString();
```
markwallace-microsoft, line 56: What is the likelihood of a new vector distance metric being defined?

When that happens our code needs to change whereas if we just take a string we are future proof but lose some strictness. Which is the lesser evil?

dotnet/src/Connectors/Connectors.Memory.Redis/RedisVectorDistanceMetrics.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.Connectors.Memory.Redis;
+
+/// <summary>
+/// Supported distance metrics are {L2, IP, COSINE}. The default value is "COSINE".
+/// <see href="https://redis.io/docs/interact/search-and-query/search/vectors/"/>
+/// </summary>
+public enum VectorDistanceMetrics
```
markwallace-microsoft, line 9: DO use a singular type name for an enumeration unless its values are bit fields.
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations

dotnet/src/Connectors/Connectors.Memory.Redis/RedisVectorIndexAlgorithms.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.Connectors.Memory.Redis;
+
+/// <summary>
+/// Vector similarity index algorithm. Supported algorithms are {FLAT, HNSW}. The default value is "HNSW".
+/// <see href="https://redis.io/docs/interact/search-and-query/search/vectors/#create-a-vector-field"/>
+/// </summary>
+public enum VectorIndexAlgorithms
```
markwallace-microsoft, line 9: DO use a singular type name for an enumeration unless its values are bit fields.
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations

dotnet/src/Connectors/Connectors.Memory.Redis/RedisVectorDistanceMetrics.cs
dmytrostruk: Good catch, resolved.

dotnet/src/Connectors/Connectors.Memory.Redis/RedisMemoryStore.cs
dmytrostruk: Ideal scenario is when Redis SDK provides enum for this parameter (like for `vectorIndexAlgorithm` for example), and we just re-use it. As soon as new distance metric is added to enum, we just update Redis SDK package version. 

As for now, I would choose more secure scenario, when we provide enum on our side, as we know that these are supported values and we decrease the probability of error. I don't think supported values will change a lot, so the cost of maintenance is low.

markwallace-microsoft (approved):

PR #2212, opened 2023-07-28
.Net: Simplifying ISKFunction.InvokeAsync

shawncal:
### Motivation and Context
Simplifying ISKFunction.InvokeAsync, and making most samples call through the Kernel.

With these changes, callers should rarely have to create or set properties on an `SKContext` or `ContextVariables` directly. Instead, they pass only what they need, and the Kernel takes care of the rest.

### Description
**ISKFunction Updates**
- Simplifying the interface by reducing to ONE version of `InvokeAsync` -- the one with the SKContext param.
- This makes it simpler to implement the ISKFunction interface
- Cleaned up `SKFunctionExtensions` to compensate. Now, a caller can run `InvokeAsync(...)` with either:
  - A single `SKContext` param (used by the Kernel), or
  - With the individual components of an `SKContext`, as an extension method (bypasses the kernel -- mostly for testing)

### Contribution Checklist

  => This is a BREAKING CHANGE


dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -26,10 +26,26 @@ public sealed class ContextVariables : IDictionary<string, string>
     /// <summary>
     /// Constructor for context variables.
     /// </summary>
-    /// <param name="value">Optional value for the main variable of the context including trust information.</param>
-    public ContextVariables(string? value = null)
+    /// <param name="input">Optional value for the main variable of the context including trust information.</param>
```
markwallace-microsoft, line 29: Is this "trust information" part of this comment still accurate?

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -26,10 +26,26 @@ public sealed class ContextVariables : IDictionary<string, string>
...
+    /// <param name="args">Optional list of key-value pairs to initialize the context with.</param>
+    public ContextVariables(
+        string? input = null,
+        IEnumerable<KeyValuePair<string, string>>? args = null)
     {
-        this._variables[MainKey] = value ?? string.Empty;
+        this.Set(MainKey, input ?? string.Empty);
```
markwallace-microsoft, line 35: Why change a null value to an empty string?

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
```diff
@@ -59,20 +58,6 @@ public interface ISKFunction
...
-    /// <param name="input">String input</param>
-    /// <param name="settings">LLM completion settings (for semantic functions only)</param>
-    /// <param name="logger">Application logger</param>
-    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <returns>The updated context, potentially a new one if context switching is implemented.</returns>
-    Task<SKContext> InvokeAsync(
-        string? input = null,
-        CompleteRequestSettings? settings = null,
-        ILogger? logger = null,
-        CancellationToken cancellationToken = default);
```
markwallace-microsoft, line 74: Is it worth having a CHANGELOG.md file in the repo where we list the breaking changes for each release?

Format could be something like this:

## Next Release

### New features:

-   ...

### Breaking changes:

- Removed `ISKFunction.InvokeAsync(string, CompletionRequestSettings, ILogger, CancellationToken)`, use `ISKFunction.InvokeAsync(string, CompletionRequestSettings, CancellationToken)` instead

## Release 0.18.230725.3-preview

### New features:

-   ...
-   ...

### Breaking changes:

- ...
- ...

dotnet/samples/KernelSyntaxExamples/Example16_CustomLLM.cs
```diff
@@ -106,7 +106,7 @@ private static async Task CustomTextCompletionWithSKFunctionAsync()
 
         var textValidationFunction = kernel.CreateSemanticFunction(FunctionDefinition);
 
-        var result = await textValidationFunction.InvokeAsync("I mised the training session this morning");
+        var result = await kernel.RunAsync(textValidationFunction, input: "I mised the training session this morning");
```
markwallace-microsoft, line 109: Typo should be "missed"

dotnet/src/SemanticKernel.Abstractions/KernelExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.SemanticKernel;
+
+public static class KernelExtensions
+{
+    public static Task<SKContext> RunAsync(
+        this IKernel kernel,
+        string skillName,
+        string functionName,
+        string? input = null,
```
RogerBarreto, line 17: Should we remove this input all the way? 
Or just make a call to the args, passing the input as another argument in the dictionary.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
```diff
@@ -487,10 +487,9 @@ public async Task CanExecutePlanWithOneStepAndStateAsync()
             .Returns(() => Task.FromResult(returnContext));
 
         var plan = new Plan(mockFunction.Object);
-        plan.State.Set("input", "Cleopatra");
 
         // Act
-        var result = await plan.InvokeAsync();
+        var result = await plan.InvokeAsync(input: "Cleopatra");
```
RogerBarreto, line 492: Shouldn't we also use `Kernel.RunAsync(plan, "Cleopatra");` ?

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
```diff
@@ -85,55 +86,42 @@ public static ISKFunction UseFrequencyPenalty(this ISKFunction skFunction, doubl
...
+        IReadOnlySkillCollection? skills = null,
+        CultureInfo? culture = null,
+        CompleteRequestSettings? settings = null,
+        ILogger? logger = null,
+        CancellationToken cancellationToken = default)
     {
-        // Log a warning if the given input is overriding a different input in the context
-        var inputInContext = context.Variables.Input;
-        if (!string.IsNullOrEmpty(inputInContext) && !string.Equals(input, inputInContext, StringComparison.Ordinal))
+        ContextVariables contextVariables = new(input);
```
RogerBarreto, line 110: Suggest: 
Get rid of `input` as a field (input is variables["input"])

```suggestion
        ContextVariables contextVariables = new(variables);
```

RogerBarreto (changes requested):

dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
```diff
@@ -139,26 +142,32 @@ private static async Task Example2Async(IKernel kernel)
 
         var oracle = kernel.CreateSemanticFunction(SemanticFunction, maxTokens: 200, temperature: 0, topP: 1);
 
-        var context = kernel.CreateNewContext();
-        context.Variables["externalInformation"] = "";
-        var answer = await oracle.InvokeAsync(questions, context);
+        var answer = await kernel.RunAsync(oracle,
+            input: questions,
```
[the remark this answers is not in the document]
dmytrostruk: I think separate optional `input` parameter is not a problem, it will simplify the usage in case there is only one input parameter. We can support both scenarios at the same time if needed.

Just as idea: we can also provide a signature where args are passed as `params`, in similar way how `string.Format` works. In this case the usage will be easier comparing to `Dictionary` initialization syntax. In case of `params` it won't be possible to map parameter key to parameter value, but values can be applied to keys by following the same order how they are defined in prompt and how they are passed to the method (like in `logger.LogInformation(str, params)`).

dotnet/samples/KernelSyntaxExamples/Example16_CustomLLM.cs
dmytrostruk: Quick fix 😄 
```suggestion
        var result = await kernel.RunAsync(textValidationFunction, input: "I missed the training session this morning");
```

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
dmytrostruk: It looks like if the value is null, the variable will be deleted. But I believe there should be a better way how to handle null-empty scenarios. Probably out of scope of this PR.

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
shawncal: Nope. Will remove

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
shawncal: Yeah, I did simplify it just to 'null', and it ended up breaking a bunch of UTs. I figured it'd be best to fix that case and all of its UTs in another change. Also, we should expect the "input" to move out of this class soon. In principle, I agree that null strings should remain null, if not provided.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
shawncal: I'm planning on putting the breaking changes section in the Releases page. That doesn't have the "next section" though. It's a good idea -- let's talk offline and sync across kernel languages

dotnet/src/SemanticKernel.Abstractions/KernelExtensions.cs
shawncal: it's a good call, and one we'll need to decide soon. I've seen, heard, and made compelling arguments for both directions. I'd like to write down the pros and cons of each approach and get the architects together to hash it out. For now, I'm just trying to make it more clear what's what, so there's not a reliance on the "magic string" entry in the dictionary -- a secret special "INPUT" key.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
shawncal: yup, can do.

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -26,10 +26,26 @@ public sealed class ContextVariables : IDictionary<string, string>
...
+
+        if (args != null)
+        {
+            this.SetAll(args);
+        }
+    }
+
+    internal void SetAll(IEnumerable<KeyValuePair<string, string>> values)
```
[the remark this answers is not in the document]
shawncal: Trying to avoid adding too many new public methods onto SKContext and ContextVariables until we have a final design for what to do with them. 

For instance, I (and some others) believe that ContextVariables can go away entirely, and be replaced with a standard Dictionary. If "input" sticks around, it might become a `Stream` to support more complex types (i.e. json, base64 images, etc), and would likely be moved out of Variables and stored as a separate property, either on the `SKContext` directly, or potentially on a new, immutable object representing the request information (i.e. input + args + any other caller information).

To be clear, none of the above is committed, and it requires an ADR, which is why I'm trying to change/expose/introduce as little as possible right now, until we have a chance to review. Just cleaning up overloads, and the reduction of ISKFunction interface in particular, is the main thing addressed in this PR.

Note @dmytrostruk, @RogerBarreto, if there are enough questions/concerns about the ContextVariables and/or IKernel parts of this PR, I can reduce it to just the ISKFunction reduction change, and then reintroduce in the others later.

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
```diff
@@ -85,55 +86,42 @@ public static ISKFunction UseFrequencyPenalty(this ISKFunction skFunction, doubl
...
+    /// <param name="logger">Logger to use for the function execution</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The result of the function execution</returns>
     public static Task<SKContext> InvokeAsync(this ISKFunction function,
-        string input,
-        SKContext context,
-        bool mutableContext = true,
-        CompleteRequestSettings? settings = null)
+        string? input = null,
+        IReadOnlyDictionary<string, string>? variables = null,
```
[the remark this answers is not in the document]
shawncal: Yup, good catch. I started with IReadOnlyDictionary, then moved to IEnum to avoid forcing the additional using statement, and to keep it more general. Will update to IEnum.

dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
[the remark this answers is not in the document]
shawncal: I agree, and would really rather not initialize the new Dictionary.

params is a neat trick, but unfortunately gets in the way of passing the CancellationToken as a final arg (see IKernel.RunAsync overloads).
It looks really weird when you use params, and have CancellationToken in the middle of the method signature.

I have an idea to try out here... stay tuned.

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
shawncal: Created Issue to track:
https://github.com/microsoft/semantic-kernel/issues/2216

dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
```diff
@@ -78,7 +78,10 @@ private static async Task Example1Async(IKernel kernel, string searchSkillId)
 
         // Run
         var question = "What's the largest building in the world?";
-        var result = await kernel.Func(searchSkillId, "search").InvokeAsync(question);
+        var result = await kernel.RunAsync(
+            skillName: searchSkillId,
```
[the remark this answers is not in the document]
shawncal: Generally, I'd agree. In the code examples though, I think it helps to really show what we're passing in and why. It helps as a teaching tool.

From a development standpoint, it also forces the IDE to resolve to the right overload/extension.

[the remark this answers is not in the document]
shawncal: Again, I added this extension to clean up the sample and give a "nicer" call pattern, but on second thought, this exposes new public APIs that, as you mention, are likely to change (or at least having naming review) soon. I'll revert these calls and remove the related new IKernelExtention call for now -- that can come later.

dotnet/src/SemanticKernel.Abstractions/KernelExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Orchestration;
+
+namespace Microsoft.SemanticKernel;
+
+public static class KernelExtensions
```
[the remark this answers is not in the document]
shawncal: Classic example of "just one more thing"... expanding the scope of a PR beyond its original intent. I've deleted this class to avoid introducing yet more public methods until we can finalize.

dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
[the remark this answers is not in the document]
shawncal: Reverted

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
shawncal: oh, this is a UT that doesn't maintain a kernel instance (it has only a mock). In this case, since we're testing one specific piece of code, it's ok to use the lower-level .InvokeAsync() call rather than going through the kernel. That ensures that we're only testing this "unit" and not any potential problems the kernel is introducing as a wrapper.

dotnet/src/SemanticKernel.Abstractions/KernelExtensions.cs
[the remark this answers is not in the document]
shawncal: Deleted this file.

dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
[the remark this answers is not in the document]
shawncal: Cleaned up.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
shawncal: reverted (reduced the scope of the PR)

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
shawncal: Reverted changes to ContextVariables. We'll tackle that another day (soon)

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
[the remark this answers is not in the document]
shawncal: Reverted changes to ContextVariables.

dotnet/src/SemanticKernel.Abstractions/KernelExtensions.cs
shawncal: Reverted this file and calling format (reduced scope of PR).

RogerBarreto (dismissed):
LGTM

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ISKFunction.cs
lemillermicrosoft: This breaking change broke one of our samples. https://github.com/microsoft/semantic-kernel/issues/2361

dotnet/src/SemanticKernel/SkillDefinition/SKFunction.cs
```diff
@@ -247,21 +250,6 @@ public async Task<SKContext> InvokeAsync(SKContext context, CompleteRequestSetti
         }
     }
 
-    /// <inheritdoc/>
-    public Task<SKContext> InvokeAsync(
```
lemillermicrosoft, line 251: We should have done the `obsolete` pattern we've been following.

PR #2215, opened 2023-07-28
.Net: Align Chroma memory store behavior with other stores

jsboige:
### Motivation and Context

This is an attempt to account for the [following PR](https://github.com/microsoft/semantic-kernel/pull/1989#issuecomment-1637218332).
Chroma logs an error when calling DoesCollectionExistAsync on a non existing collection, it should log at debug level instead
Also Volatile Store and QDrant return an empty collection when calling GetNearestMatchesAsync whereas Chroma throws an exception. This PR aligns behavior with other stores.

### Description

Align chroma store behavior with other common memory stores

### Contribution Checklist


dmytrostruk (commented):
@jsboige thank you for contribution.

Based on our conversation in this PR https://github.com/microsoft/semantic-kernel/pull/1989 and my comment:
> Regarding alignment between different connectors - we should definitely do that at some point. And I think that for `Search` operation:
> 
> * If collection exists but it's empty - we should return empty collection.
> * If collection doesn't exist - we should throw an exception to the caller, because it will clearly describe that something is wrong with memory connector setup.
> 
> In this case it will be easier to differentiate two scenarios and react on them appropriately.

For example, in Qdrant, memory record operations like `Get`, `Upsert`, `Remove` will throw exception in case you try to perform operation in non-existent collection. I don't see any reasons for `Search` operation to behave in different way and return empty result instead.

In case we make an agreement on this, then we should leave Chroma implementation as it is and align other connectors which don't follow this pattern.

Please let me know what you think about it.

glahaye (approved):

dmytrostruk (changes requested):
@jsboige please see my comment above.

jsboige:
@dmytrostruk OK I'm happy either :

- to revert the change in GetNearestMatchesAsync,, only keep the log event type change in this PR, leaving the distinct memory providers non aligned in their behavior concerning NearestMatch for missing collections. 
- to revert this one to throwing exceptions and also align the Volatile, Qdrant and potentially other providers to throw exceptions systematically in GetNearestMatchesAsync if the collection does not exist, but that will break the Copilot App until my other PR is merged, and potentially other apps that use the same pattern of create-on-insert / Search regardless of existence, so this is a breaking change that should be advertised as such, whether it is considered in this PR or postponed.
- to keep the PR where the behavior is aligned to be more lenient allowing GetNearestMatchesAsync to return empty when collection does not exist yet. That was my choice because I think that this is the currently expected behavior, and it would make sense for all the providers to change in sync with the appropriate warning for better coherence.  

Can you make a firm decision?

dmytrostruk:
@jsboige Please use this option for now:
> * to revert the change in GetNearestMatchesAsync,, only keep the log event type change in this PR, leaving the distinct memory providers non aligned in their behavior concerning NearestMatch for missing collections.

We will focus on aligning every connector to follow the same flow, I created issue for that - https://github.com/microsoft/semantic-kernel/issues/2269

The preferred way would be to throw an exception when collection doesn't exist.

Thank you!

jsboige:
@dmytrostruk I reverted the over reach to only keep the logger update

dmytrostruk (approved):
@jsboige Thanks a lot!

Issue #2221, opened 2023-07-29
.Net: Python: implement stepwise planner


joowon-dm-snu:
@alexchaomander @awharrison-28 
does your team has plan for stepwise planner?
if nothing decided, i think i can take it ;)

++ I'm curious about the future plans on python now that the C# Telemetry feature has been officially released since the main product of the company I work for is logging and analyzing interactions between users and LLM-based products. If telemetry or callback feature is also added to Python, I'd like to know if it would be possible to commit our product's integration.

alexchaomander:
@joowon-dm-snu Right now no one has taken this up yet! Of course, if you have the time, you're welcome to do it! 

From my perspective, after these planners, the telemetry and pre/post hooks integrations (i.e. callbacks) are going to be the next things to tackle so that Python SK is truly at parity with C#. Those things are still being architected in C# but the plan is once that happens, we'll want to bring the same to Python. 

@dmytrostruk @RogerBarreto for visibility as they are working on the telemetry and pre/post hooks.

matthewbolanos:
Closing since this is now complete

PR #1727, opened 2023-06-27
SequentialPlanner - Allow custom manual discovery and loading.

lemillermicrosoft:
### Motivation and Context
This change is part of a need to enable planners that can reference a large list of potential plugins that aren't loaded in a kernel instance. Additionally, consumers of planners need to be able to load plugins on demand when needed by a plan. The change includes some additional improvements to the sequential plan parser and prompt.

Part of #1728 

### Description
This pull request refactors and improves the sequential plan parser and config, simplifying the code, adding new features, and fixing some issues. The main changes are:

- Use the Azure Chat Completion Service instead of the Azure Text Completion Service for the sequential planner, and lower the relevancy threshold for creating a plan. This allows for more flexibility and creativity in generating plans, and improves the performance and accuracy of the planner.        
- Refactor the plan parsing logic to simplify the code and improve readability. Remove unnecessary try-catch blocks, logging statements, and string splitting logic. Reorder the parameters of the ToPlanFromXml method to match the order of the XML attributes. Ignore text and comment nodes in the plan XML, as they are not valid steps and could cause confusion or errors. Remove the logic that adds empty nodes as steps, as they are not valid functions and could cause errors or confusion.
- Add two optional functions to the SequentialPlannerConfig class that allow the planner to use custom logic for getting available functions and finding functions by name. These functions can be useful for scenarios where the planner needs to access functions from different sources or with different criteria, such as filtering by skill category or function type. The default behavior of the planner remains unchanged if these functions are not specified.
- Simplify and standardize the syntax for calling functions in the plan XML, using fully qualified function names and consistent attribute names for inputs, outputs, and parameters. Update the skprompt.txt file to reflect the new syntax and add more instructions and examples for creating a plan. Update the sequential plan parser tests to use the new syntax and function names, and add a helper method to get the function from the kernel context.      
- Fix some minor typos and errors in the code and the skprompt.txt file.

Details:

- dotnet/src/IntegrationTests/Planning/SequentialPlanner/SequentialPlannerTests.cs
  - Changed the KernelSyntaxExamples and SequentialPlannerTests to use the Azure Chat Completion Service instead of the Azure Text Completion Service, by replacing the deployment name, endpoint, and key environment variables.
  - Removed the EmailSkillFake import from the InitializeKernel method, as it is not used in any test case and is already imported in the TestHelpers class.
[...]

PR #2223, opened 2023-07-30
Python: add stepwise planner

joowon-dm-snu:
### Motivation and Context

implement stepwise planner #2221

### Contribution Checklist


joowon-dm-snu:
i think [bc61d2e](https://github.com/microsoft/semantic-kernel/pull/2223/commits/bc61d2e6183c4d47669aa7edf4a2a6f9ac4dd4ca) this commit should be handled more globally in other PR.  #2301

python/semantic_kernel/utils/settings.py
```diff
@@ -79,3 +79,19 @@ def pinecone_settings_from_dot_env() -> Tuple[str, str]:
     assert environment, "Pinecone environment not found in .env file"
 
     return api_key, environment
+
+
+def bing_search_settings_from_dot_env() -> str:
+    """Reads the Bing Search API key and endpoint from the .env file.
+
+    Returns:
+        Tuple[str, str]: The Bing Search API key, the Bing Search endpoint
```
TaoChenOSU, line 88: Comments are inaccurate. This function only returns the api key.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,461 @@
...
+
+        if prompt_user_config is None:
+            prompt_config = PromptTemplateConfig.from_json(
+                read_file(PROMPT_CONFIG_FILE_PATH)
+            )
+
+        prompt_config.completion.max_tokens = self.config.max_tokens
+
+        self._system_step_function = self.import_semantic_function(
+            kernel, "StepwsieStep", prompt_template, prompt_config
```
awharrison-28, line 95: Looks like a typo
```suggestion
            kernel, "StepwiseStep", prompt_template, prompt_config
```

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,461 @@
...
+        )
+        self._native_functions = self._kernel.import_skill(self, RESTRICTED_SKILL_NAME)
+
+        self._context = kernel.create_new_context()
+        self._logger = self._kernel.logger
+
+    def create_plan(self, goal: str) -> Plan:
+        if len(goal) == 0:
```
awharrison-28, line 103:
```suggestion
        if goal is None or goal == "":
```

awharrison-28: is_null_or_empty could work too

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,461 @@
...
+THOUGHT = "[THOUGHT]"
+OBSERVATION = "[OBSERVATION]"
+SCRATCH_PAD_PREFIX = (
+    "This was my previous work (but they haven't seen any of it!"
+    " They only see what I return as final answer):"
+)
+
+
+def is_null_or_empty(value: str) -> bool:
+    return value is None or len(value) == 0
```
awharrison-28, line 63:
```suggestion
    return value is None or value == ""
```

awharrison-28 (changes requested):
Hi @joowon-dm-snu,

Thanks for porting stepwise planner to python SK. And thank for you identifying the bug in plan.py. I am going to cherry pick the plan.py change out to a separate PR and add a new test for it as it's needed downstream for something else.

joowon-dm-snu:
> Hi @joowon-dm-snu,
> 
> Thanks for porting stepwise planner to python SK. And thank for you identifying the bug in plan.py. I am going to cherry pick the plan.py change out to a separate PR and add a new test for it as it's needed downstream for something else.

@awharrison-28 
Thanks :)

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,461 @@
...
+        self._system_step_function = self.import_semantic_function(
+            kernel, "StepwiseStep", prompt_template, prompt_config
+        )
+        self._native_functions = self._kernel.import_skill(self, RESTRICTED_SKILL_NAME)
+
+        self._context = kernel.create_new_context()
+        self._logger = self._kernel.logger
+
+    def create_plan(self, goal: str) -> Plan:
+        if is_null_or_empty(goal):
```
mkarle, line 103: Why this instead of `if not goal`?

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,461 @@
...
+)
+
+
+def is_null_or_empty(value: str) -> bool:
+    return value is None or value == ""
+
+
+class StepwisePlanner:
+    config: StepwisePlannerConfig
+    _context: "SKContext"
```
mkarle, line 68: Why do we save _context instead of pass the given context through?

python/tests/integration/planning/stepwise_planner/test_stepwise_planner.py
```diff
@@ -0,0 +1,180 @@
...
+from semantic_kernel.planning import StepwisePlanner
+from semantic_kernel.planning.stepwise_planner.stepwise_planner_config import (
+    StepwisePlannerConfig,
+)
+from semantic_kernel.skill_definition import sk_function, sk_function_context_parameter
+
+
+class TempWebSearchEngineSkill:
+    """
+    TODO: replace this class with semantic_kernel.core_skills.web_search_engine_skill.WebSearchEngineSkill
```
mkarle, line 24: Why use the temp instead of the WebSearchEngineSkill?

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
joowon-dm-snu: suggestion by harrison!

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
joowon-dm-snu: @mkarle we need context to execute `get_available_functions` function. actually it can be handled by `execute_plan_async`'s context but this is to match the C# version as closely as possible.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,461 @@
...
+        )
+
+    def create_scratch_pad(self, question: str, steps_taken: List[SystemStep]) -> str:
+        if len(steps_taken) == 0:
+            return ""
+
+        scratch_pad_lines: List[str] = []
+
+        # Add the original first thought
+        # joowon-dm-snu: does prompt really need SCRATCH_PAD_PREFIX? it looks like it breaks thought process pattern
```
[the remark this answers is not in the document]
joowon-dm-snu: I don't think there's a right answer, it's an engineering thing and it's probably going to feel different to different people. I'm going to delete it because I'm just making a note to check it out in another branch.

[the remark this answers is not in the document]
joowon-dm-snu: Personally, I would prefer to remove the prefix, as I think maintaining the pattern helps the behavior of the model.

python/tests/integration/planning/stepwise_planner/test_stepwise_planner.py
joowon-dm-snu:
```python
# python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
def create_plan(self, goal: str) -> Plan:
    ...
    function_descriptions = self.get_function_descriptions()

def get_function_descriptions(self) -> str:
    available_functions = self.get_available_functions()

def get_available_functions(self) -> List[FunctionView]:
    functions_view = self._context.skills.get_functions_view() <- problem here
```

`get_functions_view` function only recognize parameters that decorated by `sk_function_context_parameter` not function arguments. Currently, `WebSearchSkill`'s `query` is not decorated by `sk_function_context_parameter`, so the planner does not know that the skill has a parameter called query, and eventually it cannot enter the variable properly, causing an error. 

image

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,459 @@
...
+        plan_step._outputs.append("skill_count")
+        plan_step._outputs.append("steps_taken")
+
+        plan = Plan(goal)
+
+        plan.add_steps([plan_step])
+
+        return plan
+
+    # TODO: sync C# with https://github.com/microsoft/semantic-kernel/pull/1195
```
lemillermicrosoft, line 125: Will update this comment thread when future dotnet changes are available for review.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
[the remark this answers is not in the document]
lemillermicrosoft: We should make sure we have a set of tests that we can use to validate performance and reliability of the planners against changes.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner_config.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+from typing import List, Optional
+
+
+class StepwisePlannerConfig:
```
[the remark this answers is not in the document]
joowon-dm-snu: The lists look nice. I especially like the introduction of various callback functions. I have a few questions and additional points. (There might be some misunderstandings on my part due to the lack of examples.)

- I'm not quite sure what `GetSkillFunction` is.
- `GetPromptTemplate`, `PreferTextCompletion`, and `PromptUserConfig` seem like they could already be included in the constructor. Are these moving to `StepwisePlannerConfig` or new concept?
- Personally, since I use `kernel.import_semantic_skill_from_directory` a lot, it would be convenient to expose `SKFunction` for using custom prompt.

Since we're introducing callbacks, it would be nice to include them in other parts a bit more. (I'm curious about what parameters would be included to be used as callbacks and I've provided ideas and opinions with the possibility of manipulating internal flows.). It might increase the complexity of using the planner, but advanced users would probably appreciate it.

1) Providing user callbacks in the `var nextStep = this.ParseResult(actionText);` part could potentially allow for controlling things in the desired way even if errors occur in between. There could be approaches like forcibly creating actions.
image  

2) If user callbacks were provided before `return context;`, accessibility for the final result could improve. It seems like it could also solve the issue mentioned in my proposal #2394.  
image

3) Although a clear use case doesn't come to mind immediately, having callback functionality before and after the execution of `InvokeActionAsync` would be beneficial.
image

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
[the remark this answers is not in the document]
joowon-dm-snu: @lemillermicrosoft 
I completely agree. Does you team have any plans to provide such functionalities within the semantic-kernel in the future? Or plan to separate them into a different repository, similar to `semantic-memory`?

python/semantic_kernel/planning/stepwise_planner/stepwise_planner_config.py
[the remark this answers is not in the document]
lemillermicrosoft: Here's some more context: https://github.com/microsoft/semantic-kernel/pull/1727

>  having callback functionality before and after the execution of InvokeActionAsync would be beneficial.

Agreed! This sounds like part of other efforts right now: https://github.com/microsoft/semantic-kernel/issues/1714

I thought putting them on all the config would be convenient -- what are your thoughts?

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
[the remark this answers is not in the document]
lemillermicrosoft: In short, the kernel syntax examples project is something we'd like to transform into integration tests.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner_config.py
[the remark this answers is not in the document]
joowon-dm-snu: I agree. It's also good to have everything in the Config.

One additional suggestion is that even though the detailed workings of each planner may be slightly different, the underlying principle is similar (asking questions to the LLM to solve goals and converting them into actions). Because of this, I believe a certain level of abstraction is possible.

However, currently, the implementation of the sequential and step-wise planners is somewhat different (sequential uses `PlanObject`, while step-wise directly calls `InvokeActionAsync`), which might lead to some ambiguity. Nevertheless, in the end, having a unified execution flow would be advantageous in terms of maintenance and user understanding.

I suggest taking this opportunity to abstract appropriately and integrating it into the step-wise planner first, then later refactoring other planners.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
lemillermicrosoft: Good call though @mkarle -- probably could be worth changing at some point.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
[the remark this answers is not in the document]
lemillermicrosoft: I understand the concern about breaking the pattern, the scope however is only when MaxTokens limitations are hit and we need to truncate the thought process. I've updated to `... I've removed some of my previous work to make room for the new stuff ...` at the moment and have had satisfactory results with that (and other changes, see draft: #2504 and it's cousin #2444). I wouldn't worry about it too much right now though, we can iterate and improve.

python/tests/integration/planning/stepwise_planner/test_stepwise_planner.py
lemillermicrosoft: I'd definitely recommend updating the actual implementation. Additionally, I'd suggest raising the default value for num_results. Currently, both Google and Bing APIs have a default value of 10 for this parameter. Increasing it to at least 5 could significantly enhance semantic comprehension, as it would fetch a broader array of results. Interestingly, I've noticed that the LLM doesn't consistently make use of these parameters. Sometimes, when the default is set higher than 1, the model oddly forces it back to 1. It can be a bit frustrating at times! 😄

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
lemillermicrosoft: #2504

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+def read_file(file_path: str) -> str:
+    with open(file_path, "r") as file:
+        return file.read()
+
+
+# TODO: Original C# uses "StepwisePlanner_Excluded" for RESTRICTED_SKILL_NAME
+RESTRICTED_SKILL_NAME = "StepwisePlanner"
+S_FINAL_ANSWER_REGEX = re.compile(r"\[FINAL ANSWER\](?P<final_answer>.+)", re.DOTALL)
```
lemillermicrosoft, line 47: I've noted some models having trouble here at times with this being multi word label. Sometimes it produces [FINAL_ANSWER] or [FINAL-ANSWER] and [FINALANSWER]. Probably could make it even more robust to parsing issues.

```suggestion
S_FINAL_ANSWER_REGEX = re.compile(r"\[FINAL[_\s\-]ANSWER\](?P<final_answer>.+)", re.DOTALL)
```

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
```diff
@@ -0,0 +1,459 @@
...
+            f"{skill}({action_counts[skill]})" for skill in action_counts
+        ]
+        skill_call_list_with_counts = ", ".join(skill_call_list_with_counts)
+        skill_call_count_str = str(sum(action_counts.values()))
+
+        context.variables.set(
+            "skill_count", f"{skill_call_count_str} ({skill_call_list_with_counts})"
+        )
+
+    def create_scratch_pad(self, question: str, steps_taken: List[SystemStep]) -> str:
```
lemillermicrosoft, line 292: Just a note: in dotnet I'm looking to move away from this logic and manage the message stack in a ChatHistory object which can then be used with both text completion and chat completion models. This is fine for now though. THanks!

lemillermicrosoft:
FYI @joowon-dm-snu I pushed a commit to fix the UTs.

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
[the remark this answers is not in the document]
joowon-dm-snu: @lemillermicrosoft Agree with MaxTokens limitations! I'll use the planner as much as possible for my projects. and will try to give you some good feedback on to improve it :)

python/tests/integration/planning/stepwise_planner/test_stepwise_planner.py
joowon-dm-snu: @lemillermicrosoft 
Thanks for sharing good tips. But the code was written in a different PR, so I'm going to open an issue about that!

python/tests/integration/planning/stepwise_planner/test_stepwise_planner.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+    return kernel
+
+
+@pytest.mark.parametrize(
+    "use_chat_model, prompt, expected_function, expected_skill",
+    [
+        (
+            False,
+            "Who is the current president of the United States? What is his current age divided by 2",
```
awharrison-28, line 90: I believe it makes more sense to run an integration test against a fact that is permanent - this test will have to be changed at some point otherwise. I also find that this test fails for me regularly on my machine.

awharrison-28: Maybe "How long was Henry VIII king of England? How many times was he married divided by 2?"

python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
joowon-dm-snu: @lemillermicrosoft 
I think it's a good change, and it would be nice to see synchronization from other PRs when that part is updated.

python/tests/integration/planning/stepwise_planner/test_stepwise_planner.py
awharrison-28: Update: the above prompt works well with the tests for me

joowon-dm-snu: @awharrison-28  I think your case is much better ! I'll update it :)

awharrison-28 (approved):

python/tests/integration/planning/stepwise_planner/test_stepwise_planner.py
lemillermicrosoft: Doing similar in #2550 2550

PR #2232, opened 2023-07-31
.Net: ConversationSummarySkill assign explicit function names

### LINKED ISSUE #2230: .Net: Give ConversationSummarySkill functions explicit names
Give `ConversationSummarySkill` semantic functions explicit names so they are not assigned random ones at runtime. This allows them to be more easily recognized when described (such as `GetFunctionsView`).

anthonypuppo:
### Motivation and Context

Give ConversationSummarySkill semantic functions explicit names so they are not assigned random ones at runtime. This allows them to be more easily recognized when described (such as GetFunctionsView).

Fixes #2230 

### Description

Give ConversationSummarySkill semantic functions explicit names so they are not assigned random ones at runtime.

### Contribution Checklist


glahaye (approved):

shawncal (dismissed):

dotnet/src/Skills/Skills.Core/ConversationSummarySkill.cs
```diff
@@ -37,6 +37,7 @@ public ConversationSummarySkill(IKernel kernel)
     {
         this._summarizeConversationFunction = kernel.CreateSemanticFunction(
             SemanticFunctionConstants.SummarizeConversationDefinition,
+            functionName: nameof(SummarizeConversationAsync),
```
shawncal, line 40: In our native functions, we have code that grabs the method name, but also strips off the "Async" suffix, as the AI shouldn't know or care about that little detail... I'm thinking, instead of trying to do that automatically here, could you add string constants for each of these semantic skill definitions in the same file as the prompts themselves:
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Skills/Skills.Core/SemanticFunctionConstants.cs

...where SummarizeConversationAsync is stored as "SummarizeConversation", etc?

anthonypuppo:
@shawncal Stripping async and assigning the semantic function names would directly conflict with the existing native functions. I've updated the naming on the native functions to make it more clear to the caller what each function is designed for.

- Create constants in [SemanticFunctionConstants.cs](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Skills/Skills.Core/SemanticFunctionConstants.cs) for semantic function names
- Insert "Long" into the native function definitions to give an indication they are designed for long inputs
- Keep [Example13_ConversationSummarySkill.cs](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example13_ConversationSummarySkill.cs) consistent in that it runs the semantic functions (the example conversation transcript is only ~1,300 tokens so it will fit in model context windows fine)
- Update [AISummary.tsx](https://github.com/microsoft/semantic-kernel/blob/main/samples/apps/chat-summary-webapp-react/src/components/AISummary.tsx) to use the "long" functions as the input is variable

shawncal:
> @shawncal Stripping async and assigning the semantic function names would directly conflict with the existing native functions. I've updated the naming on the native functions to make it more clear to the caller what each function is designed for.
> 
> * Create constants in [SemanticFunctionConstants.cs](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Skills/Skills.Core/SemanticFunctionConstants.cs) for semantic function names
> * Insert "Long" into the native function definitions to give an indication they are designed for long inputs
> * Keep [Example13_ConversationSummarySkill.cs](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example13_ConversationSummarySkill.cs) consistent in that it runs the semantic functions (the example conversation transcript is only ~1,300 tokens so it will fit in model context windows fine)
> * Update [AISummary.tsx](https://github.com/microsoft/semantic-kernel/blob/main/samples/apps/chat-summary-webapp-react/src/components/AISummary.tsx) to use the "long" functions as the input is variable

Ah, I see the problem. Because the semantic functions are being "registered" with the kernel, we effectively end up with two "SummarizeConversation" functions, two "GetConversationActionItems" functions, etc. when looking at the FunctionsView. Problem is, that'll all be given to the Planners too... either with the random names (without this change) or the "Long" names (with this change). Neither is ideal.

We've got changes coming soon to how semantic functions are handled by the kernel, and this highlights a good case for why they shouldn't always be registered in the kernel, but rather sometimes they should merely be transient and invokable by a kernel (without registration).

Let me take a look at the work there and see if this case would be handled.

lemillermicrosoft:
To follow up on this, we do plan on making changes with regards to how semantic functions can be created in a more lightweight fashion, reducing the need for a change like this. For now, the obfuscation of the semantic functions by name is intentional. Thanks for your time in filing the issue and the pr itself.

PR #2235, opened 2023-07-31
Python: fix for lack of persist on persistentclient with chroma > 0.4

eavanvalkenburg:
### Motivation and Context

I noticed that the new version of ChromaDB (after 0.4.0) no longer needs an explicit call to persist, so got rid of that.
 
### Description

Deleted all persist calls, as well as the `self._persist_directory` since that was no longer used.

### Contribution Checklist


dmytrostruk (approved):

awharrison-28:
@eavanvalkenburg There was another PR that removed the persist calls within the chroma memory code. I would like to keep persist_directory however - users may want to place the database files somewhere other than the default location. and we actually do this for the integration tests. Thank you for your contribution though :)

eavanvalkenburg:
I agree that putting it in a different place is needed, but chromadb handles that in it's settings so the extra variable in the wrapper isn't needed AFAIK? @awharrison-28

PR #2236, opened 2023-07-31
Python: added num_records parameter to text memory skill

eavanvalkenburg:
### Motivation and Context

The `text_memory_skill` in the core skills was setup to only return 1 record, while the underlying memories all have support for multiple, so I added a parameter to the recall function to use that.

### Description

Added parameter, passed that on to the memory call.
### Contribution Checklist


python/semantic_kernel/core_skills/text_memory_skill.py
```diff
@@ -39,10 +46,11 @@ async def recall_async(self, ask: str, context: SKContext) -> str:
         Args:
             ask -- The question to ask the memory
             context -- Contains the 'collection' to search for information
-                and the 'relevance' score to use when searching
+                , the 'relevance' score to use when searching
+                and the 'num_records' to retrieve.
 
         Returns:
-            The nearest item from the memory store
+            The nearest item from the memory store as a comma-separated string or empty string if not found.
```
shawncal, line 53: Not sure comma-separated is useful here, as the memories themselves can contain punctuation. In the dotnet flavor, we accept a "limit" param and return the results as a json collection:
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Skills/Skills.Core/TextMemorySkill.cs

There's a good opportunity to align the languages here (parameter naming and output expectation)

eavanvalkenburg: This also adds the limit parameter, but still returns a single string, that is comma-separated when multiple results are returned (aka when limit > 1 and there are more then 1 results that match the threshold)

eavanvalkenburg: ah sorry, I see that the naming is different I will update to get it inline with the dotnet version

python/semantic_kernel/core_skills/text_memory_skill.py
```diff
@@ -39,10 +47,11 @@ async def recall_async(self, ask: str, context: SKContext) -> str:
         Args:
             ask -- The question to ask the memory
             context -- Contains the 'collection' to search for information
-                and the 'relevance' score to use when searching
+                , the 'relevance' score to use when searching
+                and the 'limit' to retrieve.
```
TaoChenOSU, line 51: nit: 'limit' of relevant memories to retrieve.

eavanvalkenburg:
I already solved that!

alexchaomander (approved):

PR #2242, opened 2023-07-31
Python: casting device int to str

jetjodh:
fixes an issue when gpu is enabled to run HF models.

### Motivation and Context

      running hf models on GPU causes an issue with device string
      fixed string concatenation issue
      Helps when running HF models on GPU

### Contribution Checklist


shawncal:
@mkarle this is a super simple fix (thanks @jetjodh!) but it's failing linting validation. Can you please take a look, and help merge this, if it's the right fix?

PR #2257, opened 2023-08-01
.Net: Adding Kusto as an external memory

danield137:
### Description

Adding Kusto as an external memory.

### Contribution Checklist


dotnet/src/Connectors/Connectors.Memory.Kusto/README.md
```diff
@@ -0,0 +1,35 @@
+﻿# Microsoft.SemanticKernel.Connectors.Memory.Kuato
+
+This connector uses Kusto (Azure Data Explorer)[https://learn.microsoft.com/en-us/azure/data-explorer/] to implement Semantic Memory.
```
zivcaspi, line 3: >Kusto (Azure Data Explorer) [](http://example.com/codeflow?start=20&length=27)

Azure Data Explorer (Kusto)

dotnet/src/Connectors/Connectors.Memory.Kusto/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Quick start
+
+1. Create a cluster and database in Kusto (Azure Data Explorer) - see https://learn.microsoft.com/en-us/azure/data-explorer/create-cluster-and-database?tabs=free
+
+2. To use Kusto as a semantic memory store:
+
+```csharp
+var connectionString = new KustoConnectionStringBuilder("https://mycluster.kusto.windows.net").WithAadUserPromptAuthentication();
```
zivcaspi, line 12: >KustoConnectionStringBuilder [](http://example.com/codeflow?start=27&length=28)

Either show the "using" statement, or provide full name here. You might also want to mention which Nuget the user needs.

dotnet/src/Connectors/Connectors.Memory.Kusto/README.md
```diff
@@ -0,0 +1,35 @@
...
+    .WithLogger(ConsoleLogger.Log)
+    .WithOpenAITextCompletionService(modelId: TestConfiguration.OpenAI.ModelId, apiKey: TestConfiguration.OpenAI.ApiKey)
+    .WithOpenAITextEmbeddingGenerationService(modelId: TestConfiguration.OpenAI.EmbeddingModelId,apiKey: TestConfiguration.OpenAI.ApiKey)
+    .WithMemoryStorage(memoryStore)
+    .Build();
+```
+
+## Important notes
+
+### Coisine similarity
```
zivcaspi, line 25: >Coisine [](http://example.com/codeflow?start=4&length=7)

Cosine

dotnet/src/Connectors/Connectors.Memory.Kusto/README.md
```diff
@@ -0,0 +1,35 @@
...
+At this time, coisine similiarity is not baked in into Kusto. It is add to the database hosting the collections when the first collection is created.  
+This is done by creating a function in the database. 
+This function is not deleted when the last collection is deleted. This is done to avoid the cost of creating the function each time a collection is created. 
+If you want to delete the function, you can do it manually using the Kusto explorer. The function is called `series_cosine_similarity_fl` and is located in the `Functions` folder of the database.
```
zivcaspi, line 29: That's too much info :)
Just say something like:

A function to calculate cosine similarity is automatically added to the Kusto data when the first collection is created. This function (`series_cosine_similarity_fl`) is not removed. You might want to delete it manually if you stop using the Kusto database as a semantic memory store.

dotnet/src/Connectors/Connectors.Memory.Kusto/Connectors.Memory.Kusto.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Kusto.Data" />
+    <PackageReference Include="Microsoft.Azure.Kusto.Ingest" />
```
zivcaspi, line 11: ><PackageReference Include="Microsoft.Azure.Kusto.Ingest" /> [](http://example.com/codeflow?start=4&length=59)

Is this actually used?

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,393 @@
...
+    public async IAsyncEnumerable<string> UpsertBatchAsync(string collectionName, IEnumerable<MemoryRecord> records, CancellationToken cancellationToken = default)
+    {
+        // Upserts don't exist in Kusto, as it is an append only store.
+        // However, since we have an easy PK, we can just insert a new record, as our query always picks the latest row of that PK.
+        // An intresting scenario is deletion after a lot of "Upserts". This could be a heavy operation, as in theory there could be many old version that we now need to remove.
+        // Plus, deleting those records is a "soft delete" operation.
+        // For simplicty purposes (and under the assumption that in most systems, upserts are rare), we will ignore the potential build-up of "garbage" records,
+        // as Kusto is generally good with large amounts of data.
+
+        var inlineIngestionCommand = $".ingest inline into table {GetTableName(collectionName)} <|\n";
```
zivcaspi, line 339: >inlineIngestionCommand [](http://example.com/codeflow?start=12&length=22)

Doesn't CslCommandGenerator already have this template?

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,393 @@
...
+        this.Embedding = KustoSerializer.DeserializeEmbedding(embedding);
+        this.Timestamp = timestamp;
+    }
+
+    public MemoryRecord ToMemoryRecord()
+    {
+        return new MemoryRecord((MemoryRecordMetadata)this.Metadata, (Embedding<float>)this.Embedding, this.Key, this.Timestamp);
+    }
+
+    public string ToCsvString()
```
zivcaspi, line 93: >ToCsvString [](http://example.com/codeflow?start=18&length=11)

We already have a CsvWriter functionality, don't recreate it (is this handling escaping characters for example?)

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,393 @@
...
+/// <summary>
+/// An implementation of <see cref="IMemoryStore"/> backed by a Kusto database.
+/// </summary>
+/// <remarks>The embedded data is saved to the Kusto database specified in the constructor.
+/// Similarity search capability is provided through a cosine similarity function (added on first collection creation). Use Kusto's "Table" to implement "Collection".
+/// </remarks>
+public class KustoMemoryStore : IMemoryStore
+{
+    private string _database;
+    private static ClientRequestProperties GetClientRequestProperties() => new()
```
zivcaspi, line 112: >GetClientRequestProperties [](http://example.com/codeflow?start=43&length=26)

You really want to create a new ClientRequestId here.

zivcaspi (changes requested):
:clock1:

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
danield137: I am, just a shortened syntax for function expression.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
danield137: changed per your recommendation

dotnet/src/Connectors/Connectors.Memory.Kusto/Connectors.Memory.Kusto.csproj
danield137: removed

dotnet/src/Connectors/Connectors.Memory.Kusto/README.md
danield137: done

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
danield137: yes. fixed. thank you!

dotnet/src/Connectors/Connectors.Memory.Kusto/README.md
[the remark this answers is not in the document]
danield137: changed

dotnet/src/Connectors/Connectors.Memory.Kusto/README.md
danield137: thanks :) ran a spell check

danield137:
You guys have an impressive array of auto-checks. Kudos.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="KustoMemoryStore"/> class.
+    /// </summary>
+    /// <param name="kcsb">Kusto Connection String Builder.</param>
+    /// <param name="database">The database used for the tables.</param>
+    public KustoMemoryStore(KustoConnectionStringBuilder kcsb, string database)
+    {
+        this._database = database;
+        this._queryClient = KustoClientFactory.CreateCslQueryProvider(kcsb);
+        this._adminClient = KustoClientFactory.CreateCslAdminProvider(kcsb);
```
dmytrostruk, line 158: I think we will need another constructor which will accept `ICslQueryProvider` and `ICslAdminProvider` as parameters. This will allow to add unit tests for this class.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+            $"| where Key in ({inClauseValue}) " +
+            "| project " +
+            // Key
+            $"{s_collectionColumns[0].Name}, " +
+            // Metadata
+            $"{s_collectionColumns[1].Name}=tostring({s_collectionColumns[1].Name}), " +
+            // Timestamp
+            $"{s_collectionColumns[3].Name}, " +
+            // Embedding
+            $"{s_collectionColumns[2].Name}=tostring({s_collectionColumns[2].Name})";
```
dmytrostruk, line 219: Accessing column names by using array with index numbers is not very readable, so you had to add comments. 

We can initialize each column separately, so in this case we will be able to access specific column by name and avoid index numbers and comments.

Definition:
```csharp
private static readonly ColumnSchema KeyColumn = new("Key", typeof(string).FullName);
private static readonly ColumnSchema MetadataColumn = new("Metadata", typeof(object).FullName);
private static readonly ColumnSchema EmbeddingColumn = new("Embedding", typeof(object).FullName);
private static readonly ColumnSchema TimestampColumn = new("Timestamp", typeof(DateTime).FullName);

private static readonly ColumnSchema[] s_collectionColumns = new ColumnSchema[]
{
    KeyColumn,
    MetadataColumn,
    EmbeddingColumn,
    TimestampColumn
};
```

Usage:
```csharp
var query = $"{this.GetBaseQuery(collectionName)} " +
    $"| where Key in ({inClauseValue}) | project " +
    $"{KeyColumn.Name}, " +
    $"{MetadataColumn.Name}=tostring({MetadataColumn.Name}), " +
    $"{TimestampColumn.Name}, " +
    $"{EmbeddingColumn.Name}=tostring({EmbeddingColumn.Name})";
```

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+            ).ConfigureAwait(false);
+
+        while (reader.Read())
+        {
+            var key = reader.GetString(0);
+            var metadata = reader.GetString(1);
+            DateTime? timestamp = reader.IsDBNull(2) ? null : reader.GetDateTime(2);
+            var _embedding = withEmbeddings ? reader.GetString(3) : default;
+
+            var kustoRecord = new KustoMemoryRecord(key, metadata, _embedding, timestamp);
```
dmytrostruk, line 242:
```suggestion
            var embedding = withEmbeddings ? reader.GetString(3) : default;

            var kustoRecord = new KustoMemoryRecord(key, metadata, embedding, timestamp);
```

dmytrostruk (commented):
@danield137 Thank you for contribution!
I left some comments, but in general implementation looks great. As soon as comments are resolved, we will test this integration manually and if everything works as expected - merge it to main! 

One note: since examples are only for manual usage (at the moment), it would be nice to add unit and/or integration tests to enable them in build pipeline to keep code coverage high. 

Thanks again!

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+/// </summary>
+/// <remarks>The embedded data is saved to the Kusto database specified in the constructor.
+/// Similarity search capability is provided through a cosine similarity function (added on first collection creation). Use Kusto's "Table" to implement "Collection".
+/// </remarks>
+public class KustoMemoryStore : IMemoryStore
+{
+    private string _database;
+    private static ClientRequestProperties GetClientRequestProperties() => new()
+    {
+        Application = "SemanticKernel",
```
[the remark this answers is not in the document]
danield137: I'm not sure I understand why it matters what is the user agent we report for our internal telemetry, but sure, I don't mind using a const. However, if we will build dashboards based on that, and it will change, that might break us.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+
+    private static readonly ColumnSchema[] s_collectionColumns = new ColumnSchema[]
+    {
+        new ColumnSchema("Key", typeof(string).FullName),
+        new ColumnSchema("Metadata", typeof(object).FullName),
+        new ColumnSchema("Embedding", typeof(object).FullName),
+        new ColumnSchema("Timestamp", typeof(DateTime).FullName),
+    };
+
+    private const string c_collectionPrefix = "sk_memory_";
```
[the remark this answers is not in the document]
danield137: I've seen other connectors prefix tables the same way. It is helpful to distinguish between tables managed by an external library vs internally in Kusto.

Is this a hard no for you or can we keep the prefix? I'll change according to your input.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+
+namespace Microsoft.SemanticKernel.Connectors.Memory.Kusto;
+
+internal static class KustoSerializer
+{
+    public static string SerializeEmbedding(Embedding<float> embedding)
+    {
+        return JsonConvert.SerializeObject(embedding.Vector);
+    }
+    public static Embedding<float> DeserializeEmbedding(string? embedding)
```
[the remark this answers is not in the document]
danield137: Will do. Maybe you can add that to your style checker?

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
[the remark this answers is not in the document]
danield137: I should note that without a prefix, I won't have an easy way to list the collections in the database. I'd have to resort to matching the schema, which I feel is less intuitive.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+        similarityQuery += "| project " +
+            // Key
+            $"{s_collectionColumns[0].Name}, " +
+            // Metadata
+            $"{s_collectionColumns[1].Name}=tostring({s_collectionColumns[1].Name}), " +
+            // Timestamp
+            $"{s_collectionColumns[3].Name}, " +
+            "similarity, " +
+            // Embedding
+            $"{s_collectionColumns[2].Name}=tostring({s_collectionColumns[2].Name})";
```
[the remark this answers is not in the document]
danield137: The queries are not identical. Carving out common code for potential future re-use doesn't seem like a good investment IMHO. If further changes will be needed in the future, I can definitely do that, but I feel like that's somewhat unlikely.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
```diff
@@ -0,0 +1,411 @@
...
+                GetClientRequestProperties(),
+                cancellationToken
+            ).ConfigureAwait(false);
+        while (reader.Read())
+        {
+            var key = reader.GetString(0);
+            var metadata = reader.GetString(1);
+            DateTime? timestamp = reader.IsDBNull(2) ? null : reader.GetDateTime(2);
+            var similarity = reader.GetDouble(3);
+            var _embedding = withEmbeddings ? reader.GetString(4) : default;
```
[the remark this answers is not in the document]
danield137: Same comment as above. Although it is possible with more effort to create a more "standard" way of doing things, I'm not sure the investment right now is worth it (nor am I sure it would make the code more "future proof")

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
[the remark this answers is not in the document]
dmytrostruk: > I'm not sure I understand why it matters what is the user agent we report for our internal telemetry

The value itself doesn't matter, you are right. It's just the approach we are using to organize our code. It's also good to keep Application identifier consistent across different integrations.

> if we will build dashboards based on that, and it will change, that might break us

Value `Semantic-Kernel` most probably won't be changed.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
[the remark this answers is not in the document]
dmytrostruk: > I've seen other connectors prefix tables the same way.

It was previously, currently we don't have table prefixing in other connectors. If it's the case, please do let me know, since we will need to remove prefixing in these connectors as well.

> It is helpful to distinguish between tables managed by an external library vs internally in Kusto.

The end users should be responsible for naming tables/collections in a way that works for them. It's similar scenario as in Entity Framework. If you use SQL Server for example, Entity Framework (as an external library) doesn't label tables during creation process.


> Is this a hard no for you or can we keep the prefix?

This question was discussed already with internal team and also with open-source community. We decided to remove prefixing from all the connectors.



> I won't have an easy way to list the collections in the database. I'd have to resort to matching the schema, which I feel is less intuitive.

I think you can list all of them. The end user is responsible for DB architecture, so if user inserts record in unmatched schema, that is the same type of problem when using some ORM with DB - it will throw an exception.



> it would force me to check for similarity function existence

That's why you probably need to move `InitializeVectorFunctionsAsync` from `CreateCollectionAsync` to some other place. I think constructor should be a good one. It will be called only once during initialization and used for passed `database` parameter.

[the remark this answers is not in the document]
danield137: It is usually considered a bad practice to put initialization code in constructors. 
But, if that is the convention, I'll follow along 😃

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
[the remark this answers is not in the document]
dmytrostruk: Got it, I see `similarity` property there. Keeping them in both places will also work at the moment. As soon as we add similar query a third time, most probably it will go to separate method, following the "Rule of Three" :)

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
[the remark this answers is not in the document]
dmytrostruk: Sounds good, feel free to skip it.

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs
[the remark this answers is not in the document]
danield137: > I think you can list all of them. The end user is responsible for DB architecture, so if user inserts record in unmatched schema, that is the same type of problem when using some ORM with DB - it will throw an exception.

I somewhat disagree with that statement, because it makes the memory connector non-standalone. Kusto is often used in a way where databases are shared, and contain large number of tables. Creating a database requires elevated permissions, so if a user was to try out the connector on an existing database, it would require them to be _very_ careful in their usage, which is more error prone. 

But again, I'm mostly stating the Kusto perspective. I will align with Semantic Kernel preferences.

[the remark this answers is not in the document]
dmytrostruk: Agreed in other communication channel to put initialization logic in `GetNearestMatchesAsync` method with additional flag to avoid calling initialization logic every time before Search operation. @danield137 thanks again! :)

dmytrostruk (approved):
@danield137 thank you for contribution!

I tested functionality manually using `Example53_Kusto` - it works as expected.

awharrison-28 (approved):

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Connectors.Memory.Kusto;
+
+internal static class KustoExtensions
+{
+    public static async ValueTask<T?> FirstOrDefaultAsync<T>(this IAsyncEnumerable<T> source, CancellationToken cancellationToken = default)
```
shawncal, line 13: Since you're importing the "InternalUtilities" in the csproj:
`  <Import Project="$(RepoRoot)/dotnet/src/InternalUtilities/src/InternalUtilities.props" />`
...you should already have access to this extension from [AsyncEnumerable](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/InternalUtilities/src/Linq/AsyncEnumerable.cs) by simply `using System.Linq`. 

Can you try removing this file?

shawncal (changes requested):

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoExtensions.cs
dmytrostruk: @shawncal Good catch! I removed this class and used already existing `AsyncEnumerable` extension method in this commit https://github.com/microsoft/semantic-kernel/pull/2257/commits/3bd036a047feec55bd492ad2b626bf48d1c79551 - it's working as expected.

shawncal (approved):

danield137:
@shawncal , is there anything else required to in order to merge this?

PR #2076, opened 2023-07-19
Python: Add Google PaLM connector with text completion and example file

am831:
### Motivation and Context

Implementation of Google PaLM connector with text completion and an example file to demonstrate its functionality. 
Closes #1979

### Description

1. Implemented Google Palm connector with text completion 
2. Added example file to ```python/samples/kernel-syntax-examples```
3. Added integration tests with different inputs to kernel.run_async
4. Added unit tests to ensure successful initialization of the class and successful API calls
5. 3 optional arguments (top_k, safety_settings, client) for google.generativeai.generate_text were not included. See more information about the function and its arguments: https://developers.generativeai.google/api/python/google/generativeai/generate_text

I also opened a PR for text embedding and chat completion #2258

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#dev-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

Currently no warnings, there was 1 warning when first installing genai with `poetry add google.generativeai==v0.1.0rc2` from within poetry shell: "The locked version 0.1.0rc2 for google-generativeai is a yanked version. Reason for being yanked: Release is marked as supporting Py3.8, but in practice it requires 3.9". We would need to require later versions of python to fix it.

PR #2258, opened 2023-08-01
Python: Add Google PaLM connector with chat completion and text embedding

### LINKED ISSUE #2098: Python: Add text embedding and chat completion to Google PaLM connector
I want to add text embedding and chat completion to the Google PaLM connector. I submitted a PR for the GP connector with text completion: #2076 
See the PR for this issue: #2258

am831:
### Motivation and Context

Implementation of Google PaLM connector with chat completion, text embedding and three example files to demonstrate their functionality.
Closes #2098
I also opened a PR for text completion #2076

### Description
**What's new:**
1. Implemented Google Palm connector with chat completion and text embedding
2. Added 3 example files to `python/samples/kernel-syntax-examples`
3. The example files show PaLM chat in action with skills, memory, and just with user messages. The memory example uses PaLM embedding as well  
5. Added integration tests to test chat with skills and embedding with kernel.memory functions
  6. Added unit tests to verify successful class initialization and correct behavior from class functions
  
**There are some important differences between google palm chat and open ai chat:**

  1. PaLM has two functions for chatting, `chat` and `reply`. The chat function in google's genai library starts a new conversation, and reply continues the conversation. Reply is an attribute of the response object returned by chat. So an instance of the `GooglePalmChatCompletion` class needs a way to determine which function to use, which is why I introduced a private attribute to store the response object. See https://developers.generativeai.google/api/python/google/generativeai/types/ChatResponse
  2. PaLM does not use system messages. Instead, the chat function takes a parameter called context. It serves the same purpose, to prime the assistant with certain behaviors and information. So when the user passes a system message to `complete_chat_async`, it is passed to `chat` as the context parameter.
  3. Semantic memory works with the chat service as long as the user creates a chat prompt template. The prompt containing the memory needs to be added to the chat prompt template as a system message. See `python\samples\kernel-syntax-examples\google_palm_chat_with_memory.py` for more details. If the only purpose of `complete_async` in `GooglePalmChatCompletion` is to send memories + user messages to the chat service as a text prompt, then `complete_async` is not fulfilling its intended purpose. A possible solution would be to send the text prompt as a request to the text service within `complete_async`.

### Contribution Checklist

Currently no warnings, there was 1 warning when first installing genai with `poetry add google.generativeai==v0.1.0rc2` from within poetry shell: "The locked version 0.1.0rc2 for google-generativeai is a yanked version. Reason for being yanked: Release is marked as supporting Py3.8, but in practice it requires 3.9". We would need to require later versions of python to fix it.


awharrison-28:
@am831 merged with main and resolved conflicts caused by the merge of PR #2076. Added some additional logic to make sure that test and module imports only occur with Python 3.9 or greater.

awharrison-28 (approved):

mkarle (approved):

PR #2270, opened 2023-08-01
Java: Apply proposed builder refactor to make them more consistent

johnoliver:
Apply proposed Builder updates to make them more consistent


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/builders/BuildersSingleton.java
```diff
@@ -51,28 +50,33 @@ public enum BuildersSingleton {
...
 
-    private final Map<Class, Supplier> builders = new HashMap<>();
+    private final Map<Class<? extends SemanticKernelBuilder>, Supplier<? extends Buildable>>
+            builders = new HashMap<>();
 
     BuildersSingleton() {
         try {
-            registerBuilder(FunctionBuilders.class, FALLBACK_FUNCTION_BUILDER_CLASS);
-            registerBuilder(Kernel.InternalBuilder.class, FALLBACK_KERNEL_BUILDER_CLASS);
+            registerBuilder(
```
dsgrieve, line 58: If the getInstance method took a String fallbackBuilderClass, then 1) registration would happen lazily and 2) when you add a new builder, you don't have to touch this class at all. The implementation of getInstance would be something like 
```
builders.computeIfAbsent(
                clazz, (k,v) ->
                (Supplier<? extends Buildable>)
                        ServiceLoadUtil.findServiceLoader(clazz, fallbackClassName));
```
and you'd get rid of registerBuilder altogether

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/builders/SemanticKernelBuilder.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft. All rights reserved.
+package com.microsoft.semantickernel.builders;
+
+public interface SemanticKernelBuilder<T extends Buildable> {
```
dsgrieve, line 4: Needs javadoc.

dsgrieve (changes requested):

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/builders/BuildersSingleton.java
dsgrieve: After discussion, this change would make it so SKBuilders could not stand on its own. I withdraw this comment.

dsgrieve (approved):

PR #2281, opened 2023-08-02
.Net: Removing internal HttpRetry Abstraction and Implementation

### LINKED ISSUE #2271: Obsolete the public "WithRetry" and internal Retry logic implementation

RogerBarreto:
### Motivation and Context

⚠️ This is a breaking change.

Reduce the responsibility surface of Semantic Kernel regarding Resiliency implementations, keeping only the abstractions that can be delegated and handling by well know resiliency libraries and custom implementations.

Resolves #2271
Closes #2271 

### Description

This change removes and deprecates the usage of Retry implementations in our Resiliency namespace with `NullHttpHandlerFactory` implementations.

Changed our examples using Polly Handlers to handle most common retrial scenarios.

### Contribution Checklist


shawncal:
Pausing this PR due to its impact, until after next nuget release. Please continue to review and apply changes -- just don't merge!

shawncal:
@SergeyMenshykh any thoughts or concerns here? Possible conflicts with error handling plans?

lemillermicrosoft:
So by default now there will be no retry handling by default? That seems like a big change that will have negative impact in developer and user experience.

dotnet/src/SemanticKernel.Abstractions/Reliability/DefaultHttpRetryHandlerFactory.cs
```diff
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+using System;
 using System.Net.Http;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.SemanticKernel.Reliability;
 
+[Obsolete("Usage of Semantic Kernel internal retry abstractions is deprecated")]
```
lemillermicrosoft, line 9: Can we move this to an extensions package? Same for Polly implementation? `Microsoft.SemanticKernel.Reliability` and/or `Microsoft.SemanticKernel.Reliability.Polly`

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -41,11 +41,6 @@ public static IKernel Create()
     /// <returns>Kernel instance</returns>
     public IKernel Build()
     {
-        if (this._httpHandlerFactory != null)
-        {
-            this._config.SetHttpRetryHandlerFactory(this._httpHandlerFactory);
```
lemillermicrosoft, line 46: Shouldn't this be replaced with call to `SetHttpHandlerFactory`

dotnet/src/SemanticKernel.Abstractions/Reliability/NullHttpHandler.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Net.Http;
+
+namespace Microsoft.SemanticKernel.Reliability;
+
+/// <summary>
+/// A http retry handler that does not retry.
```
lemillermicrosoft, line 8: Is it a retry handler or just an http handler? I think we are expecting callers to do other http handling actions here, is that right? Makes me wonder if `Reliability` is the right namespace any longer.

dotnet/samples/KernelSyntaxExamples/Example51_StepwisePlanner.cs
```diff
@@ -111,12 +111,7 @@ private static IKernel GetKernel(bool useChat = false)
 
         var kernel = builder
             .WithLogger(ConsoleLogger.Logger)
-            .Configure(c => c.SetDefaultHttpRetryConfig(new HttpRetryConfig
-            {
-                MaxRetryCount = 3,
-                UseExponentialBackoff = true,
-                MinRetryDelay = TimeSpan.FromSeconds(3),
-            }))
+            .Configure(c => c.SetHttpHandlerFactory(new PollyHttpHandlerFactory<PollyRetryThreeTimesWithBackoff>()))
```
lemillermicrosoft, line 114: Want to highlight how we are kind of admitting that you need to have Polly to use things like StepwisePlanner, now. If it remains that way, I think we should make it easier somehow.

dotnet/samples/KernelSyntaxExamples/Example42_KernelBuilder.cs
```diff
@@ -80,8 +80,7 @@ public static Task RunAsync()
         var templateEngine = new PromptTemplateEngine(logger);
         var kernelConfig = new KernelConfig();
 
-        using var httpHandler = new DefaultHttpRetryHandler(new HttpRetryConfig(), logger);
-        using var httpClient = new HttpClient(httpHandler);
+        using var httpClient = new HttpClient();
```
lemillermicrosoft, line 83: Do we even need to create this client, now?

dotnet/samples/KernelSyntaxExamples/Example22_OpenApiSkill_AzureKeyVault.cs
```diff
@@ -32,11 +32,9 @@ public static async Task RunAsync()
 
     public static async Task GetSecretFromAzureKeyVaultWithRetryAsync(InteractiveMsalAuthenticationProvider authenticationProvider)
     {
-        var retryConfig = new HttpRetryConfig() { MaxRetryCount = 3, UseExponentialBackoff = true };
-
         var kernel = new KernelBuilder()
             .WithLogger(ConsoleLogger.Logger)
-            .Configure(c => c.SetDefaultHttpRetryConfig(retryConfig))
+            .Configure(c => c.SetHttpHandlerFactory(new PollyHttpHandlerFactory<PollyRetryThreeTimesWithBackoff>()))
```
lemillermicrosoft, line 37: I'm curious if retry is needed for this example more often than not.

dotnet/src/SemanticKernel.Abstractions/KernelConfig.cs
```diff
@@ -13,36 +14,30 @@ public sealed class KernelConfig
...
-    public IDelegatingHandlerFactory HttpHandlerFactory { get; private set; } = new DefaultHttpRetryHandlerFactory(new HttpRetryConfig());
+    public IDelegatingHandlerFactory HttpHandlerFactory { get; private set; } = new NullHttpHandlerFactory();
 
     /// <summary>
     /// Default HTTP retry configuration for built-in HTTP handler factory.
     /// </summary>
-    public HttpRetryConfig DefaultHttpRetryConfig { get; private set; } = new();
+    [Obsolete("Usage of Semantic Kernel internal retry abstractions is deprecated")]
+    public HttpRetryConfig DefaultHttpRetryConfig =>
+        throw new NotSupportedException("Usage of Semantic Kernel internal retry abstractions is deprecated");
```
lemillermicrosoft, line 24: What should developers do who were previously tweaking config settings? We should point them in the right direction.

dotnet/src/SemanticKernel.UnitTests/KernelConfigTests.cs
```diff
@@ -27,11 +27,11 @@ public KernelConfigTests()
     public void HttpRetryHandlerFactoryIsSet()
```
lemillermicrosoft, line 27: If renaming others, rename this.

lemillermicrosoft (changes requested):
Sorry for the random comments.

dotnet/src/SemanticKernel.Abstractions/KernelConfig.cs
RogerBarreto: Agree, but currently we don't have a readme file with suggested packages supporting our HttpHandler Resiliency abstractions. 

I will keep track and as soon we have a readme file to relate will update the message with the reference.

dotnet/samples/KernelSyntaxExamples/Example22_OpenApiSkill_AzureKeyVault.cs
RogerBarreto: I'm not sure, I just substituted what we had with the Polly impl.

dotnet/samples/KernelSyntaxExamples/Example42_KernelBuilder.cs
RogerBarreto: This is for example purposes only when the caller has his own a httpclient implementation that he wants to use in the Kernel instance..

Took the opportunity to add another example using the `OpenAIClient` as well.

dotnet/samples/KernelSyntaxExamples/Example51_StepwisePlanner.cs
RogerBarreto: I would not use the `Need` because it all depends on the resiliency of the current infrastructure used by customer. 

Stepwise planner would work assuming high 9 figures of LLM service resiliency. That said, regarding which resiliency implementation used, yes, `Polly` seems a quite good example to work with but this could be still just one of the potential recommendations we have and could point to a future `readme` files with good resiliency extensions built by the community.

dotnet/src/SemanticKernel.Abstractions/Reliability/NullHttpHandler.cs
RogerBarreto: That's a good point, might need to move this out of `Resiliency` namespace as this is a more general purpose abstraction.

dotnet/src/SemanticKernel.Abstractions/Reliability/DefaultHttpRetryHandlerFactory.cs
RogerBarreto: We can do it in a dedicated PR. Some unit tests using the Polly will need to migrate to there as well.

dotnet/src/SemanticKernel/KernelBuilder.cs
RogerBarreto: That's a good observation, actually we don't use the Kernel private field _httpHandlerFactory in our code, I just removed it.

When we create our Text/Chat completion services we use the `HttpClientProvider` which uses the `KernelConfig` instance to grab the HttpHandler reference.

I just updated the Kernel method `WithHttpHandlerFactory` to call the KernelConfig method: `this._config.SetHttpHandlerFactory(httpHandlerFactory);`, and added Obsolete flag to the existing `WithRetryHandlerFactory`

lemillermicrosoft (approved):
Very concerned about this removal without a replacement that doesn't require the user to copy code from our KernelSyntaxExamples project.  This feels like it should be a core concept of SK and helping users leverage LLMs.

At minimum, there should be an ADR to make sure the various SDK flavors are aligned on expectations going forward.

 Approving though to remove my blocking review.

https://platform.openai.com/docs/guides/rate-limits/retrying-with-exponential-backoff

https://api.python.langchain.com/en/latest/_modules/langchain/llms/openai.html#completion_with_retry

RogerBarreto:
Closing this in favor of PR #2656

PR #2283, opened 2023-08-02
.Net: Add support for custom authentication provider for OpenAPI

mojanas:
### Motivation and Context

  1. **Why is this change required?** This change adds support for other authentication methods defined in OpenAPI specs. 
  2. **What problem does it solve?** Some APIs use other authentication methods that define a value in a custom header. As an example, [many PlayFab APIs](https://learn.microsoft.com/en-us/gaming/playfab/api-references/#security) are authenticated using secret key or user token in a custom header. The current authentication providers are not sufficient for onboarding these APIs as skills.
  3. **What scenario does it contribute to?** This change will enable APIs using custom header authentication to onboard their OpenAPI specs as skills in Semantic Kernel.
  4. **If it fixes an open issue, please link to the issue here.**

### Description

This change adds a new OpenAPI authentication provider that enables specifying a custom header to supply the authentication key or value.

### Contribution Checklist


glahaye:
@gitri-ms This is related to work you did in the past. Adding you as a reviewer.

shawncal:
I believe one can already do this by passing in an HttpClient with the DefaultHeaders populated, which is our recommended approach for similar situations across SK. (We're trying to reduce the number of routes we provide to do things that are already doable via common dotnet patterns, or can be added as extensions, to reduce the surface and complexity of the SDK)

@SergeyMenshykh can you please confirm, and take a look at the PR and the needs it addresses?

SergeyMenshykh (approved):

dotnet/src/Skills/Skills.OpenAPI/Authentication/CustomAuthenticationProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Skills.OpenAPI.Authentication;
+
+/// <summary>
+/// Retrieves authentication content (scheme and value) via the provided delegate and applies it to HTTP requests.
+/// </summary>
+public class CustomAuthenticationProvider
```
RogerBarreto, line 12: nit. 
`sealed` ? :)

mojanas: Updated

shawncal (approved):

PR #2289, opened 2023-08-02
Python: add Wikipedia search connector

### LINKED ISSUE #2272: Python: Wikipedia search connector
Wikipedia has its [MediaWiki API](https://www.mediawiki.org/wiki/API:Opensearch#API_documentation) that allows for queries returning information and article links. Having this feature would complement the new Bing and Google connectors. 

 I would like to work on making a connector for this, thank you!

sneha-afk:
### Motivation and Context

Adding onto the current search engine connectors provided, Google and Bing, that work with the web search engine skill. The upsides of Wikipedia's API is not needing an API key, simple uses of the connector (see example file) does not incur any charges as long as a reasonable rate limit is followed😄 

Closes #2272 

### Description

1. Added new file called `wikipedia_connector.py` whose `search_async` function sends a search request to the MediaWiki API (defaulted to route to the English Wiki, can be changed via constructor) and returns results in the form of [(article_title1, snippet1), (article_title2, snippet2), ...]
2. Added an example file called `wikipedia_search_skill.py` to show some use cases for the connector. This includes simple search cases without using the AI, and other cases of combining its functionality with semantic functions.

### Contribution Checklist


sneha-afk:
I had some suggestions for improving the structure of the search engine connectors, if there is agreement I can open another PR.

- `python\semantic_kernel\connectors\search_engine\connector.py` should be renamed to something more specific like `search_engine_connector_base.py` and the class should follow a similar convention, `SearchEngineConnectorBase`.
https://github.com/microsoft/semantic-kernel/blob/419e5969f717b52ed12ec8ff7cc22d0a412d8a20/python/semantic_kernel/connectors/search_engine/connector.py#L6

- The base should be an abstract class, and its sole method `search_async` should be decorated as an abstract method.

- The `search_async` method within `web_search_engine_skill.py` gets the `num_results` and `offset` variables as strings from an `SKContext`. I would think its more intuitive for these two variables to be converted into `int` before passing it onto the connector's `search_async`. This would mean changing the parameters in the connector class to be type int, which makes sense with the purposes of those variables.
https://github.com/microsoft/semantic-kernel/blob/419e5969f717b52ed12ec8ff7cc22d0a412d8a20/python/semantic_kernel/core_skills/web_search_engine_skill.py#L49-L52
https://github.com/microsoft/semantic-kernel/blob/419e5969f717b52ed12ec8ff7cc22d0a412d8a20/python/semantic_kernel/connectors/search_engine/connector.py#L11

- As seen above, the return values of two `search_async`'s are not consistent. My Wiki connector returned a `List[Tuple[str, str]` because I didn't think it was useful to only return the Wiki article title or its snippet. I believe the skill should just return the connector's return list plainly without string-ifying it.
- Perhaps the connector's `search_async` can have a return type of `List[Any]` so different connectors can have different implementations?  I can see how using `List[Any]` introduces uncertainty with the skill's return type, this can be discussed further.

awharrison-28:
Hi @sneha-afk 

This is a great connector, but we're in discussions about whether lumping its source code together with the core kernel makes sense. We will be releasing guidance on how tie external connectors to SK soon. I'll keep this PR open, but marked as draft until we can provide you with a concrete path forward.

lemillermicrosoft:
Thanks for your PR and contribution. For now, we are closing old drafts. Please re-open as appropriate.

PR #2299, opened 2023-08-03
.Net: [Error handling] [Part3] Replacing a few connector exceptions by SKException

SergeyMenshykh:
### Motivation and Context
This PR is the first in a chain of following PRs aimed at providing a simple, consistent, and extensible exception model. You can find more details in the [ADR](https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0004-error-handling.md) discussing error handling.

### Description
This PR changes PineconeMemoryStore, QdrantMemoryStore, WeaviateMemoryStore, etc.. connectors to throw SKException instead of their dedicated exceptions and removes those exceptions.

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaInvalidResponseException.cs
RogerBarreto, line 1: This is a scenario where having the Custom Exception justifies as it has a specific ResponseData field.

In this case I suggest extending from SKException instead of deleting it

dotnet/src/Connectors/Connectors.Memory.Chroma/ChromaClientException.cs
```diff
@@ -1,45 +0,0 @@
...
-    /// <param name="innerException">Instance of inner exception.</param>
-    public ChromaClientException(string message, Exception innerException) : base(message, innerException)
-    {
-    }
-
-    /// <summary>
-    /// Checks if Chroma API error means that collection does not exist.
-    /// </summary>
-    /// <param name="collectionName">Collection name.</param>
-    public bool CollectionDoesNotExistException(string collectionName) =>
```
RogerBarreto, line 43: Not sure if this method should be the responsability of this exception, maybe a property?

RogerBarreto (changes requested):

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaInvalidResponseException.cs
dmytrostruk: I'm not sure it is necessary to include response body in exception, what do you think?

dotnet/src/Connectors/Connectors.Memory.Chroma/ChromaMemoryStore.cs
```diff
@@ -70,10 +71,10 @@ public async Task DeleteCollectionAsync(string collectionName, CancellationToken
         {
             await this._chromaClient.DeleteCollectionAsync(collectionName, cancellationToken).ConfigureAwait(false);
         }
-        catch (ChromaClientException e) when (e.CollectionDoesNotExistException(collectionName))
+        catch (SKException e) when (e.Message.Contains(string.Format(CultureInfo.InvariantCulture, CollectionDoesNotExistErrorFormat, collectionName)))
```
dmytrostruk, line 74: nit: I would place this long condition in separate method since it is used in two places in this class. It will also improve the readability of `catch` block.

SergeyMenshykh: Yep, good idea, done.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/OobaboogaInvalidResponseException.cs
SergeyMenshykh: The exception was used in two places where the response body deserialization operation has succeeded but returned null. I was not able to find any official documentation describing cases when JsonSerializer.Deserialize would return null. However, according to unofficial (web search) documentation, this is possible when a valid JSON 'null' string is being deserialized. In all other cases, when the JSON string is invalid, the method should throw exceptions.

Taking that into account, I don't think this case deserves a dedicated exception type with a strongly typed property. I think one of the strong cases for having a strongly-typed property on top of derived ones from the Exception base class would be when either SK itself or SK consumer code needs to build actionable logic (fallback, resilience, etc.) based on the property value. So, I would propose postponing the decision with a dedicated exception until there's a valid case requested internally or externally. Meanwhile, we can put the body content into the Data collection so that it's not lost and can be analyzed if necessary.

shawncal (approved):

RogerBarreto (approved):
LGTM! Nice changes

shawncal (approved):

PR #2300, opened 2023-08-03
.Net: Change `VolatileMemoryStore.CreateCollectionAsync` to not throw when collection already exists

dehoward:
### Motivation and Context

today the `VolatileMemoryStore` will throw when creating a collection if it already exists. other stores (e.g. Postgres, DuckDB) follow a different pattern and return gracefully. throwing an error can create a problem for parallel writes to the memory store, which is being added in ChatCopilot [here](https://github.com/microsoft/chat-copilot/pull/83).

### Contribution Checklist


dehoward:
> I think this is a good change but it is contrary to the contract of CreateCollectionAsync in IMemoryStore. If you're going to do this, then I think you need to change the contract of CreateCollectionAsync and make this change to all implementations of IMemoryStore.

contrary in what way? the contract of `IMemoryStore.CreateCollectionAsync` doesn't specify what the behavior should be here. also, most stores already implement this behavior:

_DuckDB:_
image

_Postgres:_
image

_Sqlite:_
image

dotnet/src/SemanticKernel/Memory/VolatileMemoryStore.cs
```diff
@@ -22,7 +23,11 @@ public class VolatileMemoryStore : IMemoryStore
     /// <inheritdoc/>
     public Task CreateCollectionAsync(string collectionName, CancellationToken cancellationToken = default)
     {
-        if (!this._store.TryAdd(collectionName, new ConcurrentDictionary<string, MemoryRecord>()))
+        try
+        {
+            this._store.TryAdd(collectionName, new ConcurrentDictionary<string, MemoryRecord>());
```
glahaye, line 28: The name of the method called here (TryAdd) implies that the call WON'T throw.

Maybe it's in THAT method that the change should be made???

dmytrostruk (commented):
> DuckDB, Postgres, and Redis already implement this while Pinecone, Redis, and Chroma don't, so there doesn't seem to be a clear pattern here

@dehoward @dsgrieve  I think it is expected behavior because different DBs have different API. It depends how consumers will use `IMemoryStore` interface. For example, `SemanticTextMemory` has a check if collection exists before its creation:
https://github.com/microsoft/semantic-kernel/blob/8552943c5afc866b70e3dc771089b0342342e467/dotnet/src/SemanticKernel/Memory/SemanticTextMemory.cs#L42-L45

The only difference would be when you use concrete DB directly instead of using it through Kernel.

In case of `VolatileMemoryStore`, since it's `ConcurrentDictionary`, I wouldn't throw an exception, just to make `CreateCollection` operation idempotent.

dmytrostruk (approved):

shawncal:
> > > I think this is a good change but it is contrary to the contract of CreateCollectionAsync in IMemoryStore. If you're going to do this, then I think you need to change the contract of CreateCollectionAsync and make this change to all implementations of IMemoryStore.
> > 
> > 
> > @dsgrieve contrary in what way? the contract of `IMemoryStore.CreateCollectionAsync` doesn't specify what the behavior should be. DuckDB, Postgres, and Redis already implement this while Pinecone, Redis, and Chroma don't, so there doesn't seem to be a clear pattern here.
> > with that said, I think I could update this change to check to see if the collection already exists before throwing the error, which is more similar to what we're doing in other places.
> 
> At the time I ported IMemoryStore.cs to Java, I recall the code throwing an exception if the collection already existed. Now, it seems that the .NET implementation has drifted quite far from the Java implementation and I have some catching up to do.
> 
> Sorry for the noise.

Not noise. This is a good consideration, and was also my instinct too. I'm not sure this is what that function should to, but whatever it does, we should document it and standardize it everywhere (all connectors, all languages).

We may want to change method to (or add a new one) that's named more clearly, like `CreateCollectionIfNotExists` or
`GetCollection(bool createIfNotExists)`

shawncal:
> Member

My 2c: these functions are not really meant for the C# developer; they're for the AI to call. Throwing exceptions will just inevitably break AI-driven scenarios, so we should make SKFunctions as resilient as possible. When in doubt, defer to the intent of the function, and make it do the logical thing, not relying on exceptions as control logic. In this case, the intent is clearly to create the collection -- throwing because it already exists it a technicality that makes sense from a strictly-code Collections library, but cannot possibly be productive for the AI that just wants to make something happen.

shawncal (approved):

PR #2305, opened 2023-08-03
.Net: The Type property is added to the ParameterView class

SergeyMenshykh:
### Motivation and Context

This change was requested by an internal team to enable access to OpenAPI operation parameter types without parsing the OpenAPI documents themselves.

### Description

This PR extends the ParameterView class with a new Type property, which is assigned from the OpenAPI operation parameter type.

### Contribution Checklist


dotnet/src/Skills/Skills.UnitTests/OpenAPI/Extensions/KernelOpenApiExtensionsTests.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Skills.OpenAPI.OpenApi;
+using SemanticKernel.Skills.UnitTests.OpenAPI.TestSkills;
+using Xunit;
+
+namespace SemanticKernel.Skills.UnitTests.OpenAPI.Extensions;
+public sealed class KernelOpenApiExtensionsTests : IDisposable
```
dmytrostruk, line 13: 😃 
```suggestion

public sealed class KernelOpenApiExtensionsTests : IDisposable
```

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ParameterView.cs
```diff
@@ -37,6 +37,11 @@ public string Name
     /// </summary>
     public string? DefaultValue { get; set; }
 
+    /// <summary>
+    /// Parameter type.
+    /// </summary>
+    public string? Type { get; set; }
```
shawncal, line 43: I assume we expect this type will be limited to a few options. Can we use the fake enum trick here (define a type that looks like an enum but returns strings, and is extensible in case we miss one)?
Examples: 
 - https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/AuthorRole.cs
 - https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpmethod?view=net-7.0
 - https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpMethod.cs

SergeyMenshykh: @shawncal please have a look at the implementation.

shawncal (approved):

PR #2328, opened 2023-08-04
Java: Integrate EmbeddingVector into Embedding

dsgrieve:
### Motivation and Context

EmbeddingVector and the vectoroperations package are down in semantickernel-core, which makes it impossible to use the cosine similarity function needed to calculate the similarity between embeddings. 
### Description

EmbeddingVector was integrated into Embedding and the vectoroperations package was refactored into a VectorOperations class which is in the same package as Embedding. Now, one can do:
`embedding.cosineSimilarity(otherEmbedding)` instead of having to new up an EmbeddingVector from embedding.getVector(). 

### Contribution Checklist


brunoborges (dismissed):
LGTM

johnoliver (dismissed):

johnoliver (approved):

PR #2335, opened 2023-08-06
.Net: Update Kernel.cs

j4t-git:
minor fix

### DIFF
--- a/dotnet/src/SemanticKernel/Kernel.cs
+++ b/dotnet/src/SemanticKernel/Kernel.cs
@@ -301,8 +301,8 @@ public sealed class Kernel : IKernel, IDisposable
     private static Dictionary<string, ISKFunction> ImportSkill(object skillInstance, string skillName, ILogger logger)
     {
         MethodInfo[] methods = skillInstance.GetType().GetMethods(BindingFlags.Static | BindingFlags.Instance | BindingFlags.Public);
-        logger.LogTrace("Importing skill name: {0}. Potential methods found: {1}", skillName, methods.Length);
-
+        logger.LogTrace($"Importing skill name: {skillName}. Potential methods found: {methods.Length}");
+    
         // Filter out non-SKFunctions and fail if two functions have the same name
         Dictionary<string, ISKFunction> result = new(StringComparer.OrdinalIgnoreCase);
         foreach (MethodInfo method in methods)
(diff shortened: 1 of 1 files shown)


dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -301,28 +301,30 @@ public void Dispose()
     private static Dictionary<string, ISKFunction> ImportSkill(object skillInstance, string skillName, ILogger logger)
     {
         MethodInfo[] methods = skillInstance.GetType().GetMethods(BindingFlags.Static | BindingFlags.Instance | BindingFlags.Public);
-        logger.LogTrace("Importing skill name: {0}. Potential methods found: {1}", skillName, methods.Length);
-
+        logger.LogTrace($"Importing skill name: {skillName}. Potential methods found: {methods.Length}");
```
glahaye, line 304: Revert this change.

See the following for the reason why:
https://blog.objektkultur.de/structured-logging-with-.net-ilogger/

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -301,28 +301,30 @@ public void Dispose()
...
                 }
-
-                result.Add(function.Name, function);
             }
         }
-
-        logger.LogTrace("Methods imported {0}", result.Count);
-
+    
+        logger.LogTrace($"Methods imported {result.Count}");
```
glahaye, line 326: Revert this too.  Same reason as above. The old code enabled structured logging.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -301,28 +301,30 @@ public void Dispose()
...
         Dictionary<string, ISKFunction> result = new(StringComparer.OrdinalIgnoreCase);
         foreach (MethodInfo method in methods)
         {
             if (method.GetCustomAttribute<SKFunctionAttribute>() is not null)
             {
                 ISKFunction function = SKFunction.FromNativeMethod(method, skillInstance, skillName, logger);
-                if (result.ContainsKey(function.Name))
+                try
+                {
+                    result.Add(function.Name, function);
```
glahaye, line 315: What does this change fix? The PR's description doesn't describe any motivation for this.

The new code just seems less readable. What's the upside? Is there anything I am missing?

glahaye (changes requested):
I am not sure this change is warranted at all.

At the very least, the changes that break structured logging need to be reverted.

shawncal:
This does not appear to fix anything -- there's no attached Issue number. It contains changes we cannot accept for reasons described above. Closing, but if there is a bug you've found, please let us know.

PR #2339, opened 2023-08-07
.Net: Add Sql Server Support for memory in .NET SDK

### LINKED ISSUE #2312: .Net: Implement SQLServer MemoryStore
Following [this article](https://devblogs.microsoft.com/azure-sql/vector-similarity-search-with-azure-sql-database-and-openai/), it seems that it could be possible to have a memory store using vectors on SQL server.

kbeaugrand:
### Motivation and Context

Fix #2312

As discussed in [https://github.com/microsoft/semantic-kernel/discussions/2265](https://github.com/microsoft/semantic-kernel/discussions/2265), the objective is to add support of MSSQL Server (and Azure SQL) for vector database of the project.

### Description

I created the SQL server support based on what is presented in the following article : [https://learn.microsoft.com/en-us/samples/azure-samples/azure-sql-db-openai/azure-sql-db-openai/](https://learn.microsoft.com/en-us/samples/azure-samples/azure-sql-db-openai/azure-sql-db-openai/)

The code structure corresponds to that already in place with PostgreSQL or SQLite. A DB client performs actions towards the database, while the MemoryStore uses this client to make the necessary calls.

The DB Client creates 2 basic tables: 
- SKMemoryCollections (list of collections)
- SKMemories (contains all memory entries, by collection)  

In addition, when requesting the creation of a new collection, the client creates a new table ``SKEmbedding_{collectionName}`` for indexing all vector data.
_The clustered column store index for vector search is therefore present in the last table._

When deleting collections, the client deletes the corresponding embedding table and deletes the entry in the collections table.

> Note that the embedding is store twice (1 in the Memories tables + in the SKEmbedding_* table). This is necessary since, we have to provide also get feature withEmbeddings. It's faster to read the embedding directly in the column of Memory instead of constructing it from the vectors table. 

As far as unit and integration tests are concerned, I've reproduced the tests available for PostgreSQL and adapted them to SqlServer.

### Contribution Checklist


nacharya1:
@kbeaugrand  FYI: We recently updated our contributing guidelines [here](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#adding-plugins-and-memory-connectors , stating we are not directly adding plugin and connectors directly into the SDK repository. We are encouraging the community to host their own plugins and connectors separately in their own repositories. At the time we are exploring how we can provide a way to discover plugins and connectors and should have an update soon. I am going to close this issue given our change in contribution guidelines.

PR #2349, opened 2023-08-08
.Net: Splitting semantic and native function internal implementations

shawncal:
### Description

Splitting semantic and native functions into `internal` implementations of ISKFunction. This helps to isolate the (largely independent) code paths and bring clarity to the internal implementation.

This change does NOT introduce breaking changes; all public properties and methods are preserved. 

Note: In a future change, the interface and properties will be further edited, and that will be a breaking change. This PR is a first pass, and aims to clean things up as much as possible with zero impact to the public API surface.

Steps ahead:
 - Removing methods from the public ISKFunction interface if they only apply to one type
 - Adding helpers to identify and cast to the correct type
 - Splitting semantic functions into its own package -- the SemanticKernel.Core/Abstractions packages will have no direct knowledge of prompt templates or semantic functions as it will be exposed as a plugin package (that functionality will remain available by default for anyone using the `Microsoft.SemanticKernel` metapackage).

### Contribution Checklist


dotnet/src/SemanticKernel/SemanticFunctions/PromptFunction.cs
```diff
@@ -0,0 +1,238 @@
...
+using Microsoft.SemanticKernel.Diagnostics;
+using Microsoft.SemanticKernel.Orchestration;
+using Microsoft.SemanticKernel.SkillDefinition;
+
+namespace Microsoft.SemanticKernel.SemanticFunctions;
+
+#pragma warning disable format
+
+/// <summary>
+/// A Semantic Kernel "Semantic" prompt function.
```
lemillermicrosoft, line 23: What about these prompt functions are semantic?

dotnet/src/SemanticKernel/SkillDefinition/DotNetNativeFunction.cs
```diff
@@ -0,0 +1,861 @@
...
+
+#pragma warning disable format
+
+/// <summary>
+/// Standard Semantic Kernel callable function.
+/// SKFunction is used to extend one C# <see cref="Delegate"/>, <see cref="Func{T, TResult}"/>, <see cref="Action"/>,
+/// with additional methods required by the kernel.
+/// </summary>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+internal sealed class DotNetNativeFunction : ISKFunction, IDisposable
```
lemillermicrosoft, line 33: ChatGPT plugins would now be considered a DotNetNativeFunction even if they are pointing to code that is not running DotNet -- might be confusing.

shawncal: The ChatGPT functions are (currently) calling dotnet code - they are dotnet delegate functions.

A future refactor we can plan for may have them further split into their own custom implementation of ISKFunction, but for now they're using this one.

Remember, this is an internal implementation class, not part of the public SDK. I'm not worried about confusing ourselves.

dotnet/src/SemanticKernel/SemanticFunctions/PromptFunction.cs
shawncal: Real question?

Internal class & implementation - if the comment is confusing, we can just remove it.

lemillermicrosoft: I just wanted to make sure we weren't limiting what the results|output of a prompt function could be. 

not a blocker for pr.

dotnet/src/SemanticKernel/SkillDefinition/DotNetNativeFunction.cs
lemillermicrosoft: I'd vote for just NativeFunction and SemanticFunction since that's what we've always used internally. But not a blocking vote.

lemillermicrosoft (approved):
Stampy.

shawncal:
> Stampy.

Went with your advice (NativeFunction and SemanticFunction). Keeps it clear. Another glace pls.

PR #2353, opened 2023-08-08
.Net: Removing unused files

shawncal:
### Description

Cleaning up files no longer in use.

### Contribution Checklist


dmytrostruk (commented):
@shawncal Since these interfaces and static classes are `public`, should we apply `Obsolete` strategy here? Especially for `ContextVariablesExtensions.ToContextVariables` and `EmbeddingIndexExtensions.GetNearestMatchAsync` extension methods, which may be used by developers at the moment, so it may be breaking change.

shawncal:
> @shawncal Since these interfaces and static classes are `public`, should we apply `Obsolete` strategy here? Especially for `ContextVariablesExtensions.ToContextVariables` and `EmbeddingIndexExtensions.GetNearestMatchAsync` extension methods, which may be used by developers at the moment. We can also proceed with this PR but mark it as breaking change.

I'd rather just break, than keep dead classes around. My assumption: If nobody's using IEmbeddingIndex, then most likely nobody's using IEmbeddingIndex.GetNearestMatchAsync either. 

string.ToContextVariables -- also very unlikely to be used, I imagine, since most places we can accept a contextvariables, we can alternatively accept a string.

Marked as "breaking change"

dmytrostruk (approved):

PR #2354, opened 2023-08-08
.Net: Changing StringExtensions to "System" namespace

shawncal:
Changing the (internal)  StringExtensions to "System" namespace, like… the String class itself, so they're more broadly available across the project.

### Contribution Checklist


dotnet/src/InternalUtilities/src/System/StringExtensions.cs
```diff
@@ -2,7 +2,10 @@
 
 using System.Diagnostics.CodeAnalysis;
 
-namespace Microsoft.SemanticKernel.Text;
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+// ReSharper disable once CheckNamespace
+namespace System;
+#pragma warning restore IDE0130
```
stephentoub, line 8: The .NET 6+ API for `NormalizeLineEndings` below is called `ReplaceLineEndings`. Consider renaming it, such that this effectively just becomes a polyfill.

Also, do we need the IsNullOr{Empty/WhiteSpace} extensions? I find the syntax quite odd, encouraging you to call what looks like an instance method on something that's obviously potentially null.

shawncal: Rename to ReplaceLineEndings -- the behavior's a little different: we need to normalize to "\n" regardless of OS, as that produces more consistent results from the OpenAI models. If we just leveraged the .net6+ RepaceLineEndings(), it would normalize to "\r\n" on Windows and "\n" on Linux. The alternative would be add the extension RepaceLineEndings(str replacementText) to match .net, and get all callers to specify "\n" arg.

The string.IsNull... extensions haha age-old debate. I agree that it's weird to have nullable strings leveraging extension methods to confirm if it's null. These extensions are also included in pretty much every .net project I've ever been involved with, for the convenience. Given we're using nullability operators in these projects, I'll just remove these to avoid confusion.

Issue #2315, opened 2023-08-03
Azure Open AI Functions calling integration into Kernel


nacharya1:
With the announcement of Azure Function Calling (https://techcommunity.microsoft.com/t5/azure-ai-services-blog/function-calling-is-now-available-in-azure-openai-service/ba-p/3879241) evaluate how SK planner and kernel can incorporate support for our customers.

lemillermicrosoft:
Related to #1450

Issue #2175, opened 2023-07-25
Add support for Open AI Functions


nacharya1:
https://openai.com/blog/function-calling-and-other-api-updates?ref=upstract.com

craigomatic:
Dupe/related: https://github.com/microsoft/semantic-kernel/issues/1450

nacharya1:
closing as dupe of #2315  and #1450

Issue #1450, opened 2023-06-13
OpenAI Chat Service classes should support function calling and other API updates


awharrison-28:
OpenAI just released new features to the chat completion API. 

https://openai.com/blog/function-calling-and-other-api-updates

Our chat services objects should accommodate these updates. 
```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/semantic-kernel/issues/2315
- [ ] https://github.com/microsoft/semantic-kernel/issues/2932
- [ ] https://github.com/microsoft/semantic-kernel/issues/3017
- [ ] Explore kernel integration with function calling execution
- [ ] Add telemetry for function calling feature
```

gmilano:
Do you have some plan on how to implement this ? 

The sk has a dependency with azure sdk, the current implementation rely on Azure SDK event for abstract chat concepts,  for example on the Azure.AI.OpenAI.ChatRole class, are you thinking to wait for this  https://github.com/Azure/azure-sdk-for-net/issues/37013 ?

I think that should be better to have the model for chatting without dependencies with any other external library, so that different services provider (OpenAI, Azure Open AI or any other provider) can evolve at their own rythm.

daxian-dbw:
The latest release of the [Azure.AI.OpenAI](https://www.nuget.org/packages/Azure.AI.OpenAI/1.0.0-beta.6) already supports [Chat Functions](https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.OpenAI_1.0.0-beta.6/sdk/openai/Azure.AI.OpenAI/README.md#use-chat-functions). Can someone please share the plans on how this feature will be accommodated to Semantic Kernel (especially, the planner)?

samleegithub:
Just saw that too. Here is the blog post announcing this yesterday: https://techcommunity.microsoft.com/t5/azure-ai-services-blog/function-calling-is-now-available-in-azure-openai-service/ba-p/3879241

This would greatly improve the reliability of planners in SK in generating JSON.

hooyao:
What's the difference between SK's planner + custom skills and function calling?

eavanvalkenburg:
Just created a first implementation for this #2356

Kevdome3000:
What's the status on the dotnet implementation? I've done my own and if there isn't any plan on for this on the teams end I can submit a PR

daxian-dbw:
[...]

PR #2356, opened 2023-08-08
Python: Implement Function calling for Chat

eavanvalkenburg:
### Motivation and Context

I built support for function calling into SK!

Related to/fixes: 
- #2315 
- #2175 
- #1450 

### Description

This implementation builds on top of all the existing pieces, but did require some major work, so feel free to comment on where that is or is not appropriate.

- Added a `ChatMessage` class to capture the relevant pieces of function calling (name and content)
- Added a `complete_chat_with_functions_async` into OpenAIChatCompletions class
- Added a `function_call` field to ChatRequestSettings class
- Added several helper functions and smaller changes
- Added a sample with updated core_skill that uses function calling to demonstrate
- Added a second sample that shows how to use function_calling with non-sk functions.

### Contribution Checklist


python/samples/kernel-syntax-examples/chat_gpt_api_function_calling.py
```diff
@@ -0,0 +1,79 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+import asyncio
+
+from core_skills_function_call.math_skill import MathSkill
```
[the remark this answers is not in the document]
eavanvalkenburg: I didn't want to make changes to the built-in core skills quite yet, but you are right that can be done, I'll have a look. Will also require some work in the setup to disable function calling even though some skills have it (which is needed anyway, since not all ai services will have that ability (both from model and api perspective)

python/semantic_kernel/orchestration/sk_function.py
```diff
@@ -422,10 +483,15 @@ async def _invoke_semantic_async(self, context, settings):
                     "Semantic functions must have either an AI service or Chat service",
                 )
 
+        functions = self.describe_function_call()
+
         service = (
             self._ai_service if self._ai_service is not None else self._chat_service
         )
-        new_context = await self._function(service, settings, context)
+        if service.function_calling_enabled and functions:
```
awharrison-28, line 491: Lots of confusing use of the word functions 

Some suggestions:
- function_completion_enabled
- has_function_completion_api

@gitri-ms since you'll be working on this for .NET, I'd like your thoughts here as well

python/semantic_kernel/orchestration/sk_function_base.py
```diff
@@ -85,6 +85,14 @@ def is_native(self) -> bool:
         """
         pass
 
+    @property
+    @abstractmethod
+    def is_function_call(self) -> bool:
+        """
+        Whether the function is a function call that is passed to the API.
```
awharrison-28, line 92: Semi-called it out earlier, but the functions API is really specific to OpenAI. I think it would be better to have a method in the base class called "serialize" or similar, and then pass a serialization function to that method. It's useful to be able to serialize the calling pattern of an SK function, but the structure of the data may need to differ depending on how it's used (ex: OpenAI functions API, Planner, etc.)

python/semantic_kernel/semantic_functions/chat_prompt_template.py
```diff
@@ -35,29 +43,65 @@ async def render_async(self, context: "SKContext") -> str:
...
+        self.add_message(
+            ChatMessage(
+                role="user",
+                message=message,
+            )
+        )
 
     def add_assistant_message(self, message: str) -> None:
-        self.add_message("assistant", message)
+        self.add_message(
```
awharrison-28, line 57: I think this would fine combined with add_assistent_message_with_function_call and have the function_call be an optional parameter. Was there specific rationale for separating the two?

python/semantic_kernel/semantic_functions/chat_prompt_template.py
```diff
@@ -15,8 +16,15 @@
     from semantic_kernel.orchestration.sk_context import SKContext
 
 
+class ChatMessage(SKBaseModel):
```
awharrison-28, line 19: I'd like to see unit tests for adding, rendering, and manipulating chat message/message histories.

python/semantic_kernel/skill_definition/parameter_view.py
```diff
@@ -8,9 +9,27 @@ class ParameterView(SKBaseModel):
...
+    required: bool = False
 
-    def __init__(self, name: str, description: str, default_value: str) -> None:
+    def __init__(
+        self,
+        name: str,
+        description: str,
+        default_value: str,
+        type_: str = "string",
+        required: bool = False,
```
awharrison-28, line 21: I'm thinking it makes sense to make required default to true since the standard for most (all?) languages is that params are required unless the language allows for optional or a default.

python/semantic_kernel/skill_definition/sk_function_context_parameter_decorator.py
```diff
@@ -2,7 +2,12 @@
 
 
 def sk_function_context_parameter(
-    *, name: str, description: str, default_value: str = ""
+    *,
+    name: str,
+    description: str,
+    default_value: str = "",
+    type_: str = "string",
+    required: bool = False
```
awharrison-28, line 10: same comment on default = true vs false above

python/semantic_kernel/skill_definition/skill_collection.py
```diff
@@ -106,6 +106,13 @@ def has_native_function(
             skill_name, function_name
         )
 
+    def has_function_call_function(
```
awharrison-28, line 109: We need some clearer naming conventions here

@gitri-ms

awharrison-28 (changes requested):

python/semantic_kernel/semantic_functions/chat_prompt_template.py
eavanvalkenburg: agreed, this is def missing

python/semantic_kernel/orchestration/sk_function_base.py
eavanvalkenburg: yeah, serialize or something similar makes sense, I'll have a look

python/semantic_kernel/orchestration/sk_function.py
eavanvalkenburg: If there are any naming conventions for this, please share, the openai chat completion api does use the term function_call so I went mostly from that, but getting a more generic terminology would be good, especially as other LLM's start adding similar functionality

python/semantic_kernel/skill_definition/sk_function_decorator.py
```diff
@@ -24,6 +26,9 @@ def decorator(func):
         func.__sk_function_name__ = name or func.__name__
         func.__sk_function_input_description__ = input_description or ""
         func.__sk_function_input_default_value__ = input_default_value or ""
+        func.__sk_function_function_completion_enabled__ = (
```
awharrison-28, line 29: Was this intended to be __sk_function_function_.... or just __sk_function_?

lemillermicrosoft:
Handful of checks are not yet passing, so will turn this one to draft for now. Please feel free to re-open when things are resolved.

python/semantic_kernel/skill_definition/sk_function_decorator.py
eavanvalkenburg: Indeed, intentional, but not pretty.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
```diff
@@ -49,6 +53,7 @@ def __init__(
             org_id {Optional[str]} -- OpenAI organization ID.
                 This is usually optional unless your
                 account belongs to multiple organizations.
+            has_function_completion {bool} -- Whether or not the model supports function calling.
```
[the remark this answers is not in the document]
eavanvalkenburg: Was removed, no longer relevant.

python/semantic_kernel/semantic_functions/chat_prompt_template.py
eavanvalkenburg: Mostly that usually only message or function_call is returned, but it can be both, which would mean setting both message and function_call as optional, and then checking if at least one is present.

eavanvalkenburg:
@gitri-ms I got a hanging PR earlier, naming is now updated.
@awharrison-28 added in the implementation of the chat_messages incl tests from #2442

eavanvalkenburg:
@lemillermicrosoft fixed all tests

samples/skills/FunSkill/Joke/config.json
```diff
@@ -2,6 +2,7 @@
   "schema": 1,
   "description": "Generate a funny joke",
   "type": "completion",
+  "function_completion_enabled": true,
```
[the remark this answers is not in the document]
dmytrostruk: As far as I know, it should not break .NET version. During deserialization process this field should be just ignored.

awharrison-28 (approved):

samples/skills/FunSkill/Joke/config.json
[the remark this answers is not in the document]
eavanvalkenburg: Indeed, also good for @gitri-ms to look at for the.net side

gitri-ms, line 5: I think you missed one in the string changes:
```suggestion
  "function_calling_enabled": true,
```

gitri-ms (approved):
Just one tiny thing to fix, then LGTM!

python/semantic_kernel/orchestration/sk_function.py
```diff
@@ -115,40 +122,88 @@ def from_semantic_config(
...
+                    return context
+            except Exception as e:
+                # TODO: "critical exceptions"
+                context.fail(str(e), e)
+                return context
 
-                    # Similar to non-chat, render prompt (which renders to a
-                    # list of <role, content> messages)
-                    messages = await as_chat_prompt.render_messages_async(context)
+            functions = kwargs.get("functions")
```
lemillermicrosoft, line 140: This could be context.skill_collection after removing the `functions_calling_enabled`

lemillermicrosoft (changes requested):
Left some comments, I think we should remove changes to the template config where they are not required and lean on the skill collection for supplying available functions.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
```diff
@@ -184,57 +217,50 @@ async def _send_chat_request(
...
+            "logit_bias": (
+                request_settings.token_selection_biases
+                if request_settings.token_selection_biases is not None
+                and len(request_settings.token_selection_biases) > 0
+                else {}
+            ),
+        }
+
+        if request_settings.function_call is not None:
+            model_args["functions"] = functions
```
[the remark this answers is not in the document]
eavanvalkenburg: Good call

python/semantic_kernel/orchestration/sk_function.py
```diff
@@ -87,9 +91,11 @@ def from_native_method(method, skill_name="", log=None) -> "SKFunction":
...
         ):
             input_param = ParameterView(
-                "input",
-                method.__sk_function_input_description__,
-                method.__sk_function_input_default_value__,
+                name="input",
+                description=method.__sk_function_input_description__,
+                default_value=method.__sk_function_input_default_value__,
+                type="string",
+                required=True,
```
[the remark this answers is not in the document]
eavanvalkenburg: @awharrison-28 also commented on this, and we figured it makes sense to set all parameters as required for function calling, instead of setting it to default False and thereby inviting more cases where GPT doesn't respond with a value for that parameter. We might want to consider renaming this field to something like function_call_required to clarify that is is used only for function calling.

lemillermicrosoft:
Also, right now, having trouble getting it to work.

image

python/semantic_kernel/orchestration/sk_function_base.py
```diff
@@ -82,6 +82,14 @@ def is_native(self) -> bool:
         """
         pass
 
+    @property
+    @abstractmethod
+    def function_calling_enabled(self) -> bool:
```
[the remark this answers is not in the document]
eavanvalkenburg: not sure I understand what you mean with this one.

python/semantic_kernel/skill_definition/functions_view.py
```diff
@@ -23,6 +24,8 @@ def add_function(self, view: FunctionView) -> "FunctionsView":
                 self.native_functions[view.skill_name] = []
             self.native_functions[view.skill_name].append(view)
 
+        if view.function_calling_enabled:
```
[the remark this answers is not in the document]
eavanvalkenburg: that would make it a much bigger change, and when you are not carefull you will pass dozens of functions to the API, running out of tokens fast, while this makes sure that you have to specificly call out that you want this to be used with a function call. I guess it also depends on how often app and skill developers work independently, I do see a need for controlling which functions you want to include at runtime (app developer)

python/semantic_kernel/skill_definition/functions_view.py
```diff
@@ -60,3 +63,6 @@ def is_native(self, skill_name: str, function_name: str) -> bool:
             )
 
         return as_nf
+
+    def is_callable(self, skill_name: str, function_name: str) -> bool:
```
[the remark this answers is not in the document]
eavanvalkenburg: not now I think, but I figured keep parity across the three collections.

python/semantic_kernel/skill_definition/parameter_view.py
```diff
@@ -8,9 +11,15 @@ class ParameterView(SKBaseModel):
     name: str
     description: str
     default_value: str
+    type_: str = Field(default="string", alias="type")
+    required: bool = True
```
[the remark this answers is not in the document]
eavanvalkenburg: I had that initially, but after discussing with @awharrison-28 changed it, there is a good case for either way, but also see the other comment

python/semantic_kernel/orchestration/sk_function.py
eavanvalkenburg: correct

python/semantic_kernel/orchestration/sk_function.py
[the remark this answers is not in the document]
lemillermicrosoft: There's overlap here with needs of planner, as well. So far, implementations have relied on dictating whether a value is required in the description of the parameter -- hacky but sufficient in a semantic world. Whatever changes we make should be aligned across these scenarios.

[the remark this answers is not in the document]
lemillermicrosoft: For semantic functions, all parameters are technically optional even if they don't have a default value. For native functions, we can be smarted to determine requiredness.

python/samples/kernel-syntax-examples/chat_gpt_api_function_calling.py
```diff
@@ -42,6 +42,8 @@
 kernel.import_skill(MathSkill(), skill_name="math")
 
 # enabling or disabling function calling is done by setting the function_call parameter for the completion.
+# when the function_call parameter is set to "auto" the model will decide which function to use, if any.
```
lemillermicrosoft, line 45: So this introduces a requirement that a function ought never be named 'auto' -- yeah? Maybe this should be a fully qualified `SkillName.FunctionName`?

eavanvalkenburg: under the covers the name being used towards openai is already skillname-functionname, since only a single word is allowed, even when using global skill that will be used in the same way as well. Better docs for using the function_call param is then def needed, good call.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
```diff
@@ -60,26 +65,54 @@ def __init__(
         self._messages = []
 
     async def complete_chat_async(
+        self, messages: List[Dict[str, str]], request_settings: ChatRequestSettings
+    ) -> Union[str, List[str]]:
+        # TODO: tracking on token counts/etc.
+        response = await self._send_chat_request(
+            messages, request_settings, False, functions
```
gitri-ms, line 72: `functions` is undefined here... should it be `None` instead?

eavanvalkenburg:
Did refactoring on this one to bring it more inline with the dotnet implementation by @gitri-ms Also, some major cleanup.

eavanvalkenburg:
There is now one big architectural question coming in, and that is where to put the function_call response, in dotnet there is a notion of a specific response object per service, but we don't have that in python (yet?), so I've now added the function_call under that name as a variable in the context, which is not ideal and the context variables are also quite focused on strings, rather then other objects types. So have a look @awharrison-28 @lemillermicrosoft @shawncal @gitri-ms @nacharya1

nacharya1:
@gitri-ms - can you review and ensure this aligns with the .net approach.

python/semantic_kernel/connectors/ai/complete_request_settings.py
```diff
@@ -21,6 +21,7 @@ class CompleteRequestSettings:
     logprobs: int = 0
     token_selection_biases: Dict[int, int] = field(default_factory=dict)
     chat_system_prompt: str = "Assistant is a large language model."
+    function_call: Optional[str] = None
```
gitri-ms, line 24: Is this used for text completion? I believe it's only supported for chat.

gitri-ms (changes requested):

python/semantic_kernel/orchestration/sk_function.py
```diff
@@ -260,6 +288,17 @@ def __init__(
...
+            "name": self.full_name,
+            "description": description,
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    param.name: param.callable_function_object for param in parameters
+                },
+                "required": [p.name for p in parameters if p.required],
+            },
+        }
```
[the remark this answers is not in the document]
eavanvalkenburg: moved to utils in connectors.ai.openai

python/semantic_kernel/connectors/ai/complete_request_settings.py
eavanvalkenburg: good one

python/semantic_kernel/skill_definition/read_only_skill_collection.py
```diff
@@ -125,3 +125,65 @@ def _normalize_names(
 
         s_name, f_name = s_name.lower(), f_name.lower()
         return s_name, f_name
+
+    def get_function_calling_object(
```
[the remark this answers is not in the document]
eavanvalkenburg: moved

python/semantic_kernel/orchestration/sk_context.py
```diff
@@ -5,6 +5,9 @@
 
 import pydantic as pdt
 
+from semantic_kernel.connectors.ai.open_ai.models.chat.function_call import (
+    FunctionCall,
+)
```
gitri-ms, line 10: Would really like to avoid having `SKContext` take a dependency on the OpenAI connector. One option would be to move the `FunctionCall` model elsewhere --  although we've previously spoken about trying to keep function calling code in the connector as much as possible. Alternatively, is there another way to track the function call without using `SKContext`?

python/semantic_kernel/semantic_functions/chat_prompt_template.py
```diff
@@ -1,8 +1,11 @@
 # Copyright (c) Microsoft. All rights reserved.
 
+import asyncio
 from logging import Logger
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
+from semantic_kernel.connectors.ai.open_ai.models.chat.chat_message import ChatMessage
+from semantic_kernel.connectors.ai.open_ai.models.chat.function_call import FunctionCall
```
gitri-ms, line 8: Same comment about dependencies here -- I think it might make the most sense to move both these models out of the connector so that core classes don't take dependency on OpenAI connector.

python/semantic_kernel/orchestration/sk_context.py
eavanvalkenburg: now created a objects dict in context that can be used to hold anything other then a string (since we use variables for those), and adapted the other code to that model.

python/semantic_kernel/semantic_functions/chat_prompt_template.py
eavanvalkenburg: Now referencing a generic chat message (that is not related to OpenAI, just role and content), this is inherited by a OpenAIChatMessage. and that is used in a OpenAIChatPromptTemplate class.

gitri-ms (approved):

PR #2358, opened 2023-08-08
Python: Adding USearch memory connector

AleksandrKent:
### Motivation and Context

The integration of [USearch](https://github.com/unum-cloud/usearch) as a memory connector to Semantic Kernel (SK). 

### Description

     
The USearch `Index` does not natively have the ability to store different collections, and it only stores embeddings without other attributes like `MemoryRecord`.

The `USearchMemoryStore` class encapsulates these capabilities. It uses the USearch `Index` to store a collection of embeddings under unique IDs, with original collection names mapped to those IDs. Other `MemoryRecord ` attributes are stored in a `pyarrow.Table`, which is mapped to each collection.

It's important to note the current behavior when a user removes a record or upserts a new one with an existing ID: the old row is not removed from the `pyarrow.Table`. This is done for performance reasons but could lead to the table growing in size.

By default, `USearchMemoryStore` operates as an in-memory store. To enable persistence, you must set the persist mode with calling appropriate `__init__ `, supplying a path to the directory for the persist files. For each collection, two files will be created: `{collection_name}.usearch` and `{collection_name}.parquet`. Changes will only be dumped to the disk when `close_async` is called. Due to the interface provided by the base class `MemoryStoreBase`, this happens implicitly when using a context manager, or it may be called explicitly.

Since collection names are used to store files on disk, all names are converted to lowercase.

To ensure efficient use of memory, you should call `close_async`.

### Contribution Checklist


ashvardanian:
This is exciting! [We are also working on C# bindings for USearch to allow broader integration with SK](https://github.com/unum-cloud/usearch/issues/160) 🤗 cc @dluc

AleksandrKent:

dluc:
awesome, thank you @ashvardanian - I'll take a look asap (FYI, there's a quick git conflict to fix when you have a chance)

ashvardanian:
Hey, @dluc! @AleksandrKent has updated the poetry file. It seems to be the only collision. But it will re-appear as soon as you have any other dependency updates, so we should try merging this sooner. Please let us know if anything has to be polished.

awharrison-28 (approved):
Thank you for this contribution :)

python/tests/integration/connectors/memory/test_usearch.py
```diff
@@ -0,0 +1,387 @@
+from datetime import datetime
```
dluc, line 1:
```suggestion
# Copyright (c) Microsoft. All rights reserved.

from datetime import datetime
```

python/semantic_kernel/connectors/memory/usearch/__init__.py
```diff
@@ -0,0 +1,5 @@
+from semantic_kernel.connectors.memory.usearch.usearch_memory_store import (
```
dluc, line 1:
```suggestion
# Copyright (c) Microsoft. All rights reserved.

from semantic_kernel.connectors.memory.usearch.usearch_memory_store import (
```

python/semantic_kernel/connectors/memory/usearch/usearch_memory_store.py
```diff
@@ -0,0 +1,636 @@
+import itertools
```
dluc, line 1:
```suggestion
# Copyright (c) Microsoft. All rights reserved.

import itertools
```

dluc (commented):
missing file headers

dluc (approved):

PR #2397, opened 2023-08-10
.Net: Fix for ScoredValue structure

dmytrostruk:
### Motivation and Context

Should fix: https://github.com/microsoft/semantic-kernel/issues/616

Structure `ScoredValue` contains generic property `T Value` and structure property `Score Score`. Implemented operators `==` and `!=` include both properties `Value` and `Score` while operators `<`, `>`, `<=`, `>=` include property `Score` only. In this case operators `==`, `!=` and `<=`, `>=` may produce different results based on `Score` and `Value` property values.

This PR contains changes to:
- Remove `Score` structure, since it contains just one `double` property without additional functionality. `Score` structure can be easily replaced by `double`.
- Fix inconsistent results for `==`, `!=` and `<=`, `>=` operators.

New logic works in following way:
- Operators `==` and `!=` rely on `Score` equality AND `Value` equality based on `T` implementation of `IEquatable<T>` interface or `Object.Equals`.
- Operators `<` and `>` rely on `Score` comparison first, then on `Value` comparison in case if `Score` is equal and `T` implements `IComparable<T>` interface.
- Operators `<=` and `>=` rely on `<` and `>` operators first (backed by `IComparable<T>` implementation) OR `==` operator (backed by `IEquatable<T>` or `Object.Equals` implementation).

### Description
1. Updated `ScoredValue` structure to follow the logic described above.
2. Replaced `Score` structure by `double`.
3. Added unit tests to verify behavior of all `ScoredValue` comparison operators.

### Contribution Checklist


dotnet/src/SemanticKernel/Memory/Collections/ScoredValue.cs
```diff
@@ -18,11 +18,18 @@ public ScoredValue(T item, double score)
...
     public int CompareTo(ScoredValue<T> other)
     {
-        return this.Score.CompareTo(other.Score);
+        var result = this.Score.CompareTo(other.Score);
+
+        if (result == 0 && this.Value is IComparable<T> comparableValue)
+        {
+            result = comparableValue.CompareTo(other.Value);
+        }
+
```
shawncal, line 31: We're making some statements here about how to "break a tie" by comparing score and value... a clever idea. That said, if I was checking for whether two entries are equal, I think I'd want to know that they are equal Score -- not sure I'd expect them to be differently ordered even when their Scores are identical. It's a design choice, and one that should be well-documented.

dotnet/src/SemanticKernel/Memory/Collections/TopNCollection.cs
```diff
@@ -62,7 +62,7 @@ public void Add(ScoredValue<T> value)
         this._heap.Add(value);
```
shawncal, line 62: General comment: I know we looked at this once before, but this might be a good opportunity to bring in the PriorityQueue to replace TopNCollection. It's only available on .net6+ so we'd have to bring it in as a code copy for builds that are below .net6, with a compiler directive to ignore the code if net6+ (we have other examples of this).

Then we could entirely remove the whole "Collections" folder.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs

shawncal (approved):

dotnet/src/SemanticKernel/Memory/Collections/TopNCollection.cs
dmytrostruk: I tried to copy `PriorityQueue` implementation, but it appeared as non-trivial task, because this data structure depends on .NET internal helpers (e.g. `EnumerableHelpers`), which should be copied as well. Also, there is a big list of warnings generated by our convention rules, and it will take some time to resolve them all (`PriorityQueue` has near 1000 lines of code).
I would probably leave `MinHeap` as it is right now.

stephentoub: > e.g. EnumerableHelpers

FWIW, that's just the implementation of Enumerable.ToArray, which PQ can't rely on in the actual implementation because of layering (System.Linq.dll depends on System.Collections.dll, so the latter can't reference the former).

dotnet/src/SemanticKernel/Memory/Collections/ScoredValue.cs
```diff
@@ -18,11 +18,18 @@ public ScoredValue(T item, double score)
     }
 
     public T Value { get; }
-    public Score Score { get; }
+    public double Score { get; }
 
     public int CompareTo(ScoredValue<T> other)
```
stephentoub, line 23: What is the intended usage of ScoredValue? I'm not sure what to make of the `Value` playing a role in its `CompareTo`.

dotnet/src/SemanticKernel/Memory/Collections/ScoredValue.cs
```diff
@@ -78,7 +85,7 @@ public override int GetHashCode()
 
     public static bool operator <=(ScoredValue<T> left, ScoredValue<T> right)
     {
-        return left.CompareTo(right) <= 0;
+        return left < right || left.Equals(right);
```
stephentoub, line 88: So if left is `new ScoredValue(4, "abc")` and right is `new ScoredValue(5, "def")`, left is `<=` right, but if left is `new ScoredValue(4, "abc")` and right is `new ScoredValue(4, "def")`, left isn't `<=` right?

Given this definition of `<=` and `>=`, doesn't this mean that you can have two ScoredValues where both `<=` and `>=` are false, if they contain the same score but different values?

dotnet/src/SemanticKernel/Memory/Collections/ScoredValue.cs
dmytrostruk: @stephentoub The main purpose of `ScoredValue` structure is to be used in `TopNCollection` (it's `MinHeap`, which keeps only top `N` entries by score). We use this collection to find top `N` similar records from DB based on `Score`. So, the actual value doesn't matter, and we need to sort entries only by score. Primary scenario in SDK is when `T` is `MemoryRecord` object.

In `ScoredValue` structure we have following methods and operators to implement - `CompareTo`, `Equals`, `GetHashCode`, and `<`, `>`, `<=`, `>=`, `==`, `!=`. If we want to keep all method and operator results consistent - we will have to do the comparison in all places using the same set of properties:
- In case we use both `Value` and `Score` in all places - all results will be consistent when `T Value` implements `IComparable` interface.
- In case we use just `Score` in all places - comparison operators will be consistent, while I'm not sure how `Equals` and `GetHashCode` will serve their purpose (for example in `Dictionary`, even though it's not the main purpose of `ScoredValue`, but it's still `public` and accessible by users). 

Currently in `main` branch we use just `Score` for comparison, while both `Value` and `Score` for equality, which creates inconsistency between `<=`, `>=` and `==` operators, that's why GH issue and this PR were created.

I'm wondering, what would be the right solution here, because it looks like we have to sacrifice some functionality to make it work consistent for SDK use-case only.

dotnet/src/SemanticKernel/Memory/Collections/ScoredValue.cs
dmytrostruk: @stephentoub 

> So if left is `new ScoredValue(4, "abc")` and right is `new ScoredValue(5, "def")`, left is `<=` right, but if left is `new ScoredValue(4, "abc")` and right is `new ScoredValue(4, "def")`, left isn't `<=` right?

Left is `<=` right in both cases with this implementation:
image

> Given this definition of `<=` and `>=`, doesn't this mean that you can have two ScoredValues where both `<=` and `>=` are false, if they contain the same score but different values?

This is true, when `T Value` doesn't implement `IComparable` interface. Otherwise `<=` and `>=` results will be opposite.

dotnet/src/SemanticKernel/Memory/Collections/ScoredValue.cs
stephentoub: > The main purpose of ScoredValue structure is to be used in TopNCollection

Do we need TopNCollection and ScoredValue to public?  Could we just rely on OrderBy(...).Take(...)?

dmytrostruk: > Do we need TopNCollection and ScoredValue to public? Could external consumers just rely on OrderBy(...).Take(...)?

Good questions. As far as I can see they are used in a couple of memory connectors. Also, we still call `TopNCollection.SortByScore` method manually, which performs in-place sorting. I'm not sure about performance advantages here, taking into account that we shouldn't operate with big amount of records in runtime, ideally the sorting should be performed on memory DB side. 

@shawncal Do you have any thoughts regarding `OrderBy(...).Take(...)` instead of `TopNCollection`?

shawncal: The only real advantage to TopNCollection is that, as values are added, it keeps only the top (scoring) values, discarding others. This makes for very fast additions, as there are only N items to compare to at any given time. Most other collections would hold onto every item, then sort all of those items, then remove most of those items, putting potentially a lot more memory pressure on these searches.

That said, I would very much like to remove TopNCollection and ScoredValue from the public API surface. We could just copy these classes into the few connectors that use them, or we could use those linq methods in the connectors and use these classes in our own internal implementation only.

dmytrostruk:
`TopNCollection` and `ScoredValue` structures will be removed from public API surface in scope of separate PR.

PR #2419, opened 2023-08-10
.Net: Represent all embeddings as `ReadOnlyMemory<float>`

stephentoub:
### Motivation and Context

Across AI-related libraries, we'd like a consistent representation of an embedding.  It needs to support efficient handling of the data, with the ability to get a span over the data so that the data can be manipulated efficiently, which rules out interfaces like `IEnumerable<float>` and `IList<float>`. It should ideally encourage immutability but not enforce it, such that developers have escape hatches that don't require them to continually allocate and copy around large objects. It should be usable in both synchronous and asynchronous contexts. And it should be a ubiquitous core type that's available everywhere.  Based on that, `ReadOnlyMemory<float>` is the winning candidate.

### Description

This PR overhauls SK to use `ReadOnlyMemory<float>` for embeddings. In addition to the advantages listed above, I think it also cleans things up nicely.

Note that in the System.Text.Json v8.0.0 nuget package, ReadOnlyMemory is implicitly supported. Once SK moves from having a v6 to a v8 dependency, we can delete the ReadOnlyMemoryConverter and all mentions of it.

### Contribution Checklist


dmytrostruk (approved):
Nice changes! I also double-checked memory connectors to make sure we don't break the schema for already existing records in DBs and tested manually Redis and Chroma connectors - everything works as expected.

shawncal (approved):

umeshma:
While I think its makes sense low level math routines to work with ReadOnlyMemory,  the big problem I see with this change is that there is _meaning_ and behavior associated with Embedding: it was made a strong type deliberately and has been so in the SK codebase from the very beginning. Just like there is meaning associated with System.Numerics.ComplexNumber.  

The original intent for Embedding was to be lightweight wrapper around the underlying vector... so people didn't just pass around random arrays. We wanted them to pass around _Embeddings_. However, the current implementation and its read-only nature unfortunately makes some operations harder. 

ReadOnlyMemory means nothing here. It is just an array.

stephentoub:
> The original intent for Embedding was to be lightweight wrapper around the underlying vector... so people didn't just pass around random arrays or spans or readonly memory

I hear what you're saying.  And if it was a closed ecosystem that consisted only of SK, I could understand it. But it's not, and the moment you cross the boundary into another component, you have to go from that Embedding to something else, or from that something else to the Embedding. That means frequently ToArray'ing back and forth, which is not only inefficient, it loses the meaning of having Embedding.  The strong typing isn't really buying anything; if you have an array, ok, well, just wrap it in an Embedding, and if you need an array, ok, well, just ToArray it.  In my opinion, it ends up adding cost for little-to-no benefit.

> However, the current implementation and its read-only nature unfortunately makes some operations harder.

Embedding was trying to be readonly as well. Its ctor would make a copy of the input (I previously special-cased some internal constructions to avoid that), and it only exposed the internal array via read-only means.  With `ReadOnlyMemory<float>`, there are actually multiple ways to extract the data to be non-readonly, you just have to make an explicit "I know what I'm doing" gesture to do so, e.g. via the MemoryMarshal type.

umeshma:
Imagine if as a programmer, I was just given a (void *pVoid) or pByte everywhere.  And yes, we've all been there :-) and there is a place for it. Instead of String or ComplexNumber or something with meaning and associated code readability. There is a place for Pvoid of course. I don't think we should let just our desired to get maximum performance override the opportunity to introduce the right vocabulary for how to program this stuff into .NET. 

IMHO: ReadOnlyMemory is not the right vocabulary. 

You are working with an Embedding. People retrieve embeddings from models. Ada 002 is an embedding model. It returns embeddings, not ReadOnlyMemory. They do similarity searches by finding the closest embedding. Vector databases store embeddings. We should therefore use vocabulary that is natural here.. let developers work with Embedding instead of a PByte.    

The ToArray has always been the issue in the current impl because it tries to do too much. The original embedding was a _lightweight_ struct that avoided all that because it was designed to merely _hold the vector_ and return it as is - no copying every - that was its contract. Just a couple of implicit operator overloads. Too bad C# does not have a **typedef**, but that is effectively what it was. The syntactic sugar introduced by the typedef did matter because it let the code and the interfaces and the associated verbiage use the right vocabulary without losing perf.

PR #2421, opened 2023-08-11
.Net: Combining memory connector examples into TextMemorySkill 

shawncal:
Combining all the (nearly identical) memory connector syntax examples into the single Example15_TextMemorySkill.cs example.

### Motivation and Context
General cleanup and code reduction. 

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example15_TextMemorySkill.cs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    private static async Task<IMemoryStore> CreateSamplePostgresMemoryStoreAsync()
+    {
+        NpgsqlDataSourceBuilder dataSourceBuilder = new(TestConfiguration.Postgres.ConnectionString);
+        dataSourceBuilder.UseVector();
+        await using NpgsqlDataSource dataSource = dataSourceBuilder.Build();
+        IMemoryStore store = new PostgresMemoryStore(dataSource, vectorSize: 1536, schema: "public");
+        return store;
+    }
```
dmytrostruk, line 132: Redis and Postgres connectors will throw an exception on a first request, because here we are disposing `ConnectionMultiplexer` in case of Redis and `NpgsqlDataSource` in case of Postgres at the end of these initialization methods. We should initialize them on higher level (maybe at the beginning of `RunAsync` method) and use them here or skip disposal since these are just samples.

dotnet/samples/KernelSyntaxExamples/Example15_TextMemorySkill.cs
```diff
@@ -0,0 +1,280 @@
...
+    private static IMemoryStore CreateSampleKustoMemoryStore()
+    {
+        var connectionString = new Kusto.Data.KustoConnectionStringBuilder(TestConfiguration.Kusto.ConnectionString).WithAadUserPromptAuthentication();
+        IMemoryStore store = new KustoMemoryStore(connectionString, "MyDatabase");
+        return store;
+    }
+
+    private static async Task RunWithStoreAsync(IMemoryStore memoryStore, CancellationToken cancellationToken)
+    {
+        var logger = ConsoleLogger.Logger;
```
dmytrostruk, line 143: I just tried to run example with Chroma and this logger outputs too many messages with transient errors and warnings, so console window is overloaded and it's hard to find actual function results. What if we set `NullLogger.Instance` by default and provide a comment that for detailed logs it's possible to use `ConsoleLogger.Logger`? 

It applies for loggers on Kernel and each memory store level.

dmytrostruk (changes requested):
I did a manual testing with Chroma and it works only partially :) 
For some reason during last requests, it tries to search memory record in collection with name `generics` which doesn't exist. 

Previous examples were simpler - save records, get records, search, output collections.
This example includes semantic function invocation and responses are not that reliable. Most probably we will have to re-test each memory connector for this scenario and check how each works. Otherwise, we may receive new issues that samples don't work as expected.

shawncal:
> 

I'll try pulling in the examples that the other memory providers were following. Those should be simpler.

Thanks for running!

Flipping to Draft for now

dotnet/samples/KernelSyntaxExamples/Example15_TextMemorySkill.cs
lemillermicrosoft: Aside: I've been updating ConsoleLogger locally to expose the LogFactory so I can create loggers more easily and tune filters per category/name/type. I'll try and get that pulled out of a draft and merged with the recent ILogger changes.

samples/notebooks/dotnet/config/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "openai",
```
lemillermicrosoft, line 2: Is this related? or included on accident?

lemillermicrosoft (approved):

dmytrostruk (approved):
Tested manually with Redis, Chroma and Postgres.

hario90 (approved):

RogerBarreto (approved):
Not the scope of this PR but I'm not a very fan of leaving commented code in the examples for the other alternative examples. 

Since this is a console app, we could introduce some command interactions allowing the user to chose which memory it want to run.

PR #2425, opened 2023-08-11
Python: Update 03-semantic-function-inline.ipynb

demongodYY:
### Motivation and Context

Example notebook has some params error when I learn from it, so I fix this simple params error.

### Description

Fix AzureChatCompletion params error, Azure OpenAI Service should pass DEPLOY_NAME, but not MODEL_NAME

### Contribution Checklist


shawncal (approved):
Good find! Thanks for the fix.

awharrison-28 (approved):

PR #2428, opened 2023-08-11
SKException consistency with docs/decisions/0004-error-handling.md

dsgrieve:
### Motivation and Context

Where it makes sense, align Java with https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0004-error-handling.md

### Description

For the most part, Java is in compliance with the decision record. This PR tries to make SKException and its inheritors consistent with each other in that each exception has a ErrorCodes enum and an ErrorCodes member with a getter, that the enums use upper case with underscore, and that the formatting of the error message is consistent. This also tries to address "Unnecessary and verbose exceptions", but there was not much to be gained there. 

### Contribution Checklist


milderhc (approved):

java/connectors/semantickernel-connectors-memory-azurecognitivesearch/src/main/java/com/microsoft/semantickernel/connectors/memory/azurecognitivesearch/AzureCognitiveSearchMemory.java
```diff
@@ -323,8 +326,7 @@ private static String normalizeNullToEmptyString(@Nullable String value) {
     // <returns>Normalized name</returns>
     private static String normalizeIndexName(String indexName) {
         if (indexName.length() > 128) {
-            throw new AzureCognitiveSearchMemoryException(
-                    "The collection name is too long, it cannot exceed 128 chars");
+            throw new IllegalArgumentException("The collection name cannot exceed 128 chars");
```
johnoliver, line 329: Seems to have gone less specific

dsgrieve: One of the points in the error handling ADR is about unnecessary exceptions, the concern being that clients may become dependent on them making it harder to change in the future. I felt that the AzureCognitiveSearchMemoryException was such a case as it was only used in this case, and two others which could be up-leveled. Further, I felt that this was not truly a "memory exception" as this is not something unexpected that happened during processing, such as not being able to parse a JSON value. This is a case where the programmer has done the wrong thing and, as such, this should be caught during development. Still, the javadoc does not reflect this limitation so I will fix the javadoc for this.

johnoliver (approved):

PR #2442, opened 2023-08-14
Python: add parsing of (Azure) OpenAI into classes, including ChatMessages

eavanvalkenburg:
### Motivation and Context

1. There were a lot of opinions in what is and isn't returned by the different OpenAI classes, that is now much more complete (things like Usage are returned as well)
2. Tricks had to be done to deal with 1 or more results in the openai class, that now moves to downstream classes, or user code.
3. This will also make Function Calling much easier
4. Adds Usage on both service and function level
5. This will make it much easier to store and restore chat messages

One note, most of the completion results and other classes are based on the openai api definitions, which are probably the most prevalent, but a mapping of these and HF and other ai services is needed.

### Description

1. Added models folder to store `completion results`, `usage_result`, `finish_reason`, `completion_content` and appropriate base classes as well as the `chat` versions of these, and a Role class.
2. Added models into `connectors\ai\openai\models` to store subclasses of above, with specific classmethods that create these items from openai_objects.
3. Adapted chat_completion_client_base and text_completion_client_base to have ai_service_client_base as the base which houses a logger and usage, adapted all downstream methods for these objects (`complete_chat_async` and the like)
4. adapted tests for all this and added unit tests for all new classes (except enums)

### Contribution Checklist


lemillermicrosoft:
Marking as draft while this is broken down into smaller pieces. We should also make there is an accompanying issue to make sure we maintain parity across .Net and python.

eavanvalkenburg:
Makes sense @lemillermicrosoft it did get a bit big. I did hear there were already pieces of this (using a object, rather then a tuple for the response) in the .net version, but did one did go a lot further, should I create the issue for this or will you?

PR #2444, opened 2023-08-14
.Net: Improvements to StepwisePlanner for higher accuracy and fewer LLM calls

andhesky:
### Motivation and Context

The current stepwise planner generally takes 5-10 iterations to get to an answer many of which involve no detected action or an action that does not exist. Even after all those iterations, it is frequently wrong when answering the integration test question. Also, it generally just does not work or takes even longer when using the chat interface which is required for GPT-4.

**NOTE** This is a draft PR for feedback on a few areas. It generally works but there are outstanding issues and open questions listed below

### Description
* Use the ChatHistory object instead of ScratchPad
* When using the TextCompletion, convert it to a ChatCompletion using a wrapper
* Give configurable few-shot examples
* Automatically exclude the "time" plugin as callable as there are lots of functions and they confuse the planner

With these changes, it generally (>80%) answers the integration test question accurately in 2-3 steps.

### Open Questions
* Do we want to have the wrapping of the TextCompleting be explicit by just throwing if there is no ChatCompletion?
* Are the classes in the right assemblies?

### Outstanding issues
* Currently there is no culling of the ChatHistory.
* PromptConfig is not honored
* Stop sequences is hacky
* Filtering time plugin is hacky, but it is currently required with the base system prompt which is confusing
* Text streaming is not supported
* Comments and formatting

### Contribution Checklist


dotnet/src/Skills/Skills.Web/WebSearchEngineSkill.cs
```diff
@@ -28,7 +28,7 @@ public WebSearchEngineSkill(IWebSearchEngineConnector connector)
     [SKFunction, Description("Perform a web search.")]
     public async Task<string> SearchAsync(
         [Description("Text to search for")] string query,
-        [Description("Number of results")] int count = 1,
+        [Description("Number of results")] int count = 5,
```
lemillermicrosoft, line 31: Yes, strong agree. I even updated mine to 10 as this is the bing default. This really helps.

lemillermicrosoft: Thought: Still, I'm surprised the LLM never seems to try and override this default. Been considering ways to encourage this.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/TextCompletionChatWrapper.cs
```diff
@@ -0,0 +1,60 @@
...
+        this._wrappedCompletion = wrappedCompletion;
+        this._converter = converter ?? new ChatToTextConverter();
+    }
+
+    public ChatHistory CreateNewChat(string? instructions = null)
+    {
+        var chatHistory = new ChatHistory();
+        if (!string.IsNullOrWhiteSpace(instructions))
+        {
+            // TODO reconcile ChatHistory with OpenAiChatHistory, why is there a different class?
```
lemillermicrosoft, line 25: cc @RogerBarreto

dotnet/src/Extensions/Planning.StepwisePlanner/Skills/StepwiseStep/SystemPrompt.txt
```diff
@@ -0,0 +1,47 @@
+[INSTRUCTIONS]
+* Answer the user as accurately as possible using the provided functions.
+* Break down the problem step by step and think about the best approach
+* Generate a single [THOUGHT] and an optional single [ACTION]. The resulting output of an [ACTION] will be provided as an [OBSERVATION]
+* Always use the functions provided in the [AVAILABLE FUNCTIONS] section when possible regardless of your internal knowledge or information.
+* Once the necessary observations have been gathered and required actions performed, provide the [FINAL ANSWER] in a clear and human-readable format.
+
+
+[AVAILABLE FUNCTIONS]
```
lemillermicrosoft, line 9: I anecdotally was finding chat model would hallucinate functions quickly until I pulled the available functions out of here and into the first user message before the question. Did you experience anything similar?

andhesky: It did pretty well if I created few shot examples, but I didn't do a lot of experimentation. It certainly seemed to pay more attention to the examples than this definition.

dotnet/src/Extensions/Planning.StepwisePlanner/Skills/StepwiseStep/SystemPrompt.txt
```diff
@@ -0,0 +1,47 @@
...
+[ACTION]
+$JSON_BLOB
+[OBSERVATION]
+The result of the action will be provided here.
+... (These Thought/Action/Observation can repeat until the final answer is reached.)
+[FINAL ANSWER]
+Once I have gathered all the necessary observations and performed any required actions, I can provide the final answer in a clear and human-readable format.
+[END THOUGHT PROCESS]
+
+At the beginning of this chat exchange, the time is {{time.UtcNow}}
```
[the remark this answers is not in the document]
andhesky: Yea, it might be useful to have an {{$additionalInstructions}} on line 7 for example. Though honestly it might be better just to provide a few example txt files and let them copy/paste/modify. With a few examples for dynamic content like function definitions or current time, I found a txt file is a lot easier to understand than the complexity introduced with a bunch of extension points.

Also, as a user, once I start customizing the system prompt I think that I expect to be able to control the whole thing when lends itself better to examples rather than substitution.

dotnet/src/Skills/Skills.Web/WebSearchEngineSkill.cs
andhesky: I was conservative with 5, but it was good enough to make it work reliably. Yea, I suspect we could put some guidance about when to override in the system prompt, but based on my knowledge, that guidance would be something like "always use 5".

This is mostly unrelated so I can actually break into a separate PR so we can pick the right number. I think 1 is almost certainly a bad default.

dotnet/src/Extensions/Planning.StepwisePlanner/Skills/StepwiseStep/SystemPrompt.txt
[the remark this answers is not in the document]
andhesky: Regarding the time plugin requirement that stinks too especially because we require it and then exclude it from the callable functions with no user feedback about either the requirement or the exclusion. Need to think of a better way to at least make these obvious and intuitive or maybe use a different mechanism that requires less of the caller.

andhesky:
Replaced by https://github.com/microsoft/semantic-kernel/pull/2504

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -98,92 +127,109 @@ public Plan CreatePlan(string goal)
...
+        for (int i = 0; i < this.Config.MaxIterations; i++)
+        {
+            // TODO request settings and cancellation token
+            IReadOnlyList<IChatResult> llmResponse = await this._chatCompletion.GetChatCompletionsAsync(chatHistory).ConfigureAwait(false);
+            IChatResult llmResult = llmResponse.Single(); // TODO, is there a case where we get multiple messages?
+            ChatMessageBase responseMessage = await llmResult.GetChatMessageAsync().ConfigureAwait(false);
 
-                string actionText = llmResponse.Result.Trim();
-                this._logger?.LogTrace("Response: {ActionText}", actionText);
+            string actionText = responseMessage.Content.Trim();
```
lemillermicrosoft, line 168: Another note for comparison -- I found it beneficial to not give the raw response as the message, but instead the parsed version (if valid) to help define the expectations better.

PR #2464, opened 2023-08-15
.Net: IKernel - Pre/Post RunAsync Handlers (ADR + Impl Example)

### LINKED ISSUE #1714: Be able to Handle Function Calls: Phase 1

RogerBarreto:
### Motivation and Context

Be able to add handlers to function executions using `Kernel.RunAsync` in a way the caller can be notified and get or change information before and after an function is executed by the given Kernel.

Resolves #1714 

### Description

This implementation adds 2 new public interface events to a IKernel interface allowing setting Pre(Running)/Post(Ran) handlers to it:

- IKernel.FunctionInvoking
- IKernel.FunctionInvoked

With this initial change allow scenarios like:

- Pre-Execution (Invoking)

    - Get: Prompt generated by the current Kernel `TemplateEngine` before calling the ISKFunction.InvokeAsync
    - Get: SKContext
    - Set: Modify the SKContext variables
    - Set: Cancel further RunAsync function executions

- Post-Execution (Invoked)
    - Get: SKContext
    - Set: Modify the SKContext variables
    - Set: Cancel further RunAsync function executions

## Out of scope of this PR.

- Plan/Step Handlers execution support
- Individual ISKFunction.InvokeAsync execution support

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -186,8 +187,17 @@ public interface IKernel
     /// <returns>Instance of T</returns>
     T GetService<T>(string? name = null) where T : IAIService;
 
-    #region Obsolete
+    /// <summary>
+    /// Used for registering a pre-run event handler.
+    /// </summary>
+    event EventHandler<KernelRunningEventArgs> Running;
```
lemillermicrosoft, line 193: nit: Maybe something like OnExecuting and OnExectued?

RogerBarreto: I did this thinking in the RunAsync, should we use ExecuteAsync then ?

shawncal: The On* naming, IMO would typically be for the handler (not the event itself)
`     kernel.Running += this.OnRunning`

We probably do need event names that better indicate that this will be run before and after each step (not just once for the whole pipeline). This is a tricky one... thinking of recommendations.

- FunctionStarting // FunctionCompleted
- StepRunning // StepCompleted
...

RogerBarreto: Right, as discussed, I'm going to use `FunctionInvoking` and `FunctionInvoked` as those events, which better fits the current behavior and Function.InvokeAsync method.

docs/decisions/0005-function-execution-handlers.md
```diff
@@ -0,0 +1,133 @@
...
+  - Get: Generated Prompt
+  - Get: SKContext
+  - Get: Output parameters
+
+  - Set: Modify a prompt content after getting it from LLM
+  - Set: Modify output parameters content (before returning the output)
+
+## Decision Drivers
+
+- Architecture changes and the associated decision making process should be transparent to the community.
```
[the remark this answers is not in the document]
RogerBarreto: The Decision is not final yet, but as soon I get the final agreement (Approval) that we are going with the EventHandler approach will update the Decision

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -324,6 +347,31 @@ public void Dispose()
...
+    /// <param name="functionView">Function view details</param>
+    /// <param name="context">SKContext before function invocation</param>
+    /// <param name="prompt">Generated prompt when using semantic functions</param>
+    /// <returns>Returns true if no cancellation was attempted.</returns>
+    private bool OnFunctionInvoking(FunctionView functionView, SKContext context, string? prompt)
+    {
+        var args = new FunctionInvokingEventArgs(functionView, context, prompt);
+        this.FunctionInvoking?.Invoke(this, args);
+
+        return !args.Cancel;
```
stephentoub, line 362: Better to avoid allocating the arguments if there's no handler registered.
```suggestion
        var func = this.FunctionInvoking;
        if (func is not null)
        {
            var args = new FunctionInvokingEventArgs(functionView, context, prompt);
            func(this, args);
            return !args.Cancel;
        }
        
        return true;
```

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -194,6 +201,20 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
             try
             {
                 cancellationToken.ThrowIfCancellationRequested();
+                var functionDetails = f.Describe();
+                string? renderedPrompt = null;
+
+                if (f is SemanticFunction semanticFunction)
+                {
+                    renderedPrompt = await semanticFunction._promptTemplate.RenderAsync(context, cancellationToken).ConfigureAwait(false);
+                }
```
stephentoub, line 210: renderedPrompt is only used as part of the FunctionInvoking handler?  Does that means we'll end up rendering it twice, once here for the event and another time inside of the actual invocation?  Three issues:
1. Duplicated effort.
2. Couldn't that mean we actually get a different prompt here than what's actually submitted, if there's any non-determinism involved? Or is it always 100% deterministic?
3. We're doing the work here even if there's no handler registered for FunctionInvoking.

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -186,8 +187,19 @@ public interface IKernel
...
+    /// Used for registering a function invoking event handler.
+    /// Triggers before each function invocation.
+    /// </summary>
+    event EventHandler<FunctionInvokingEventArgs> FunctionInvoking;
 
+    /// <summary>
+    /// Used for registering a function invoked event handler.
+    /// Triggers after each function invocation.
+    /// </summary>
+    event EventHandler<FunctionInvokedEventArgs> FunctionInvoked;
```
stephentoub, line 200: These should be nullable as well, just like on Kernel, right?

dotnet/src/SemanticKernel/Kernel.cs
RogerBarreto: @stephentoub, that's a good point.
1. We are planning (very soon) to move PromptTemplateEngine logic away from the function execution and actually pass the generated prompt as an input to the function itself in the meanwhile, yes, this is a duplicated effort.

2. It can be potentially non deterministic if the template actually calls other semantic functions (non deterministic by nature) or Native skills like random number generators. That said, those scenarios aren't common.

3. That's new for me, I thought the `FunctionInvoking?.Invoke` would do the null check before trying to call it and not be expensive.

RogerBarreto: As a way to overcome numbers 1 and 2, I added an `internal GetAndForgetRenderedPrompt` method to SKContext which will preserve the GeneratedPrompt in the InvokeAsync, so this way, when using both `IKernel.RunAsync` and `ISKFunction.InvokeAsync` no duplicated effort will happen as well as any problems that could arise from it.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -194,14 +201,41 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
...
+
+                if (!this.OnFunctionInvoking(functionDetails, context, renderedPrompt))
+                {
+                    this.Logger.LogInformation("Function invocation was cancelled {0}: {1}.{2}.", pipelineStepCount, f.SkillName, f.Name);
+                    break;
+                }
+
+                if (semanticFunction is not null)
+                {
+                    context = await semanticFunction.InternalInvokeAsync(context, addDefaultValues: false, cancellationToken: cancellationToken).ConfigureAwait(false);
```
lemillermicrosoft, line 224: Have you considered updating the signature of InvokeAsync to take a callback?

There's a lot of code paths here, can we add some tests for them?

lemillermicrosoft: The SetRenderedPrompt on the context feels kind of hacky. What's going to happen if a pre-execution handler ends up calling a semantic function using that context, etc.

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokedEventArgs.cs
```diff
@@ -0,0 +1,31 @@
...
+
+/// <summary>
+/// Event arguments available to the Kernel.FunctionInvoked event.
+/// </summary>
+public sealed class FunctionInvokedEventArgs : EventArgs
+{
+    internal FunctionInvokedEventArgs(FunctionView functionView, SKContext context)
+    {
+        Verify.NotNull(context);
+        this.FunctionView = functionView;
```
lemillermicrosoft, line 18: <3

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
```diff
@@ -0,0 +1,41 @@
...
+/// </summary>
+public sealed class FunctionInvokingEventArgs : CancelEventArgs
+{
+    internal FunctionInvokingEventArgs(FunctionView functionView, SKContext context, string? prompt)
+    {
+        Verify.NotNull(context);
+
+        this.FunctionView = functionView;
+        this.SKContext = context;
+        this.Prompt = prompt;
```
lemillermicrosoft, line 21: Might be worth including on the FunctionInvokedEventArgs as well? That way folks don't have to register both if they don't intend on modifying a prompt or context before execution.

dotnet/src/SemanticKernel/Kernel.cs
RogerBarreto: Thanks for the feedback.

I managed to make it more clean and friendly, moving `renderedPrompt` as a parameter to the `InternalInvokeAsync`. This way I don't need to hacky using the SKContext or using the `bool addDefaultValues` parameter.

Quick note: As the responsibility for rendering prompt is not of the function moving this logic outside of the Invoke implementation is a step forward in that direction as a plan to move the Rendering logic to be a `Plan`.

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
RogerBarreto: Done, good point!

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -194,14 +201,40 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
...
-                context = await f.InvokeAsync(context, cancellationToken: cancellationToken).ConfigureAwait(false);
+                var functionDetails = f.Describe();
+                string? renderedPrompt = null;
+
+                SemanticFunction? semanticFunction = f as SemanticFunction;
+
+                if (semanticFunction is not null)
+                {
+                    semanticFunction.AddDefaultValues(context.Variables);
+                    renderedPrompt = await semanticFunction._promptTemplate.RenderAsync(context, cancellationToken).ConfigureAwait(false);
```
markwallace-microsoft, line 212: Will `RenderAsync` get called again when the function is invoked?

If yes, this could be a problem because the prompt could contain functions and the values those functions return could change.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -53,6 +54,12 @@ public sealed class Kernel : IKernel, IDisposable
     /// </summary>
     public static KernelBuilder Builder => new();
 
+    /// <inheritdoc/>
+    public event EventHandler<FunctionInvokingEventArgs>? FunctionInvoking;
```
shawncal, line 58: Can we use `AsyncEventHander`?... 
```
public delegate Task AsyncEventHandler<TEventArgs>(
    object? sender,
    TEventArgs args,
    CancellationToken cancellationToken = default) where TEventArgs : EventArgs;
```

That way, handlers can naturally be written as awaitable Task methods. Then, when invoked, each callback is awaited. (it's likely that many handlers won't need to be awaitable, but it's far easier/cleaner to implement an async-over-sync handler than a sync-over-async method).

`public event AsyncEventHandler<FunctionInvokingEventArgs>? FunctionInvoking;`

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -324,6 +357,37 @@ public void Dispose()
...
+    /// <param name="functionView">Function view details</param>
+    /// <param name="context">SKContext before function invocation</param>
+    /// <param name="renderedPrompt">Rendered prompt prior to semantic function invocation</param>
+    /// <returns>Returns true if no cancellation was attempted.</returns>
+    private bool OnFunctionInvoking(FunctionView functionView, SKContext context, string? renderedPrompt)
+    {
+        if (this.FunctionInvoking is not null)
+        {
+            var args = new FunctionInvokingEventArgs(functionView, context, renderedPrompt);
+            this.FunctionInvoking.Invoke(this, args);
```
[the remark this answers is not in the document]
RogerBarreto: Initially I considered this approach where the Cancel would abort further callers.

As an alternative I'm adding two new properties to `FunctionInvokingEventArgs`:

Added `Skip` will skip current Function, moving to the next skill in the pipeline/plan. 
Renamed `Cancel` to `Abort` will stop all execution.

For both properties, once set to true by one handler the values cannot be unset by any further  handler.
Any further attempt to unset it will throw.

dotnet/src/SemanticKernel/Kernel.cs
RogerBarreto: There is no AsyncEventHandler supported by dotnet.

One hacky way would be something like Example below:
 
```
public event Func<object, EventArgs, Task>? FunctionInvoking;
``` 
https://codereview.stackexchange.com/questions/261396/asynchronous-event-handler

Keep note that trying this approach got [CA1003 Warning](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1003)

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    public SKContext SKContext { get; }
+
+    /// <summary>
+    /// Prompt that was rendered prior to the function execution.
+    /// </summary>
+    /// <remarks>
+    /// May be null for native functions.
+    /// </remarks>
+    public string? Prompt { get; }
```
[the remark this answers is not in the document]
RogerBarreto: Agree, we can start this way for now and once we have a definition how Prompt will be passed down to functions as arguments I can revisit this implementation (redirecting to a SKContext Variable + Obsolete it)

dotnet/src/SemanticKernel/Kernel.cs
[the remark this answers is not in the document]
RogerBarreto: Resolved that, this will introduce a Breaking Change label to this PR.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -197,14 +204,41 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
...
+
+                SemanticFunction? semanticFunction = f as SemanticFunction;
+
+                if (semanticFunction is not null)
+                {
+                    semanticFunction.AddDefaultValues(context.Variables);
+                    renderedPrompt = await semanticFunction.RenderPromptTemplateAsync(context, cancellationToken).ConfigureAwait(false);
+                }
+
+                var cancelled = this.OnFunctionInvoking(functionDetails, context, renderedPrompt);
```
shawncal, line 218:
```suggestion
                SKContext promptContext = this.CreateNewContext();
                promptContext.Update(renderedPrompt);
                var cancelled = this.OnFunctionInvoking(functionDetails, promptContext);
```
...then we don't need `string? Prompt` to be a hard-coded property on the EventArgs classes.

dotnet/src/SemanticKernel/SkillDefinition/FunctionBase.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.SemanticKernel.AI.TextCompletion;
+using Microsoft.SemanticKernel.Diagnostics;
+using Microsoft.SemanticKernel.Events;
+using Microsoft.SemanticKernel.Orchestration;
+
+namespace Microsoft.SemanticKernel.SkillDefinition;
+internal abstract class FunctionBase : ISKFunction,
```
lemillermicrosoft, line 15: Can you add some documentation on this class and its purpose? Why does a Plan not implement FunctionBase?

dotnet/src/SemanticKernel/SemanticFunctions/PromptTemplate.cs
```diff
@@ -22,6 +22,8 @@ public sealed class PromptTemplate : IPromptTemplate
     // ReSharper disable once NotAccessedField.Local
     private readonly PromptTemplateConfig _promptConfig;
 
+    internal string Template => this._template;
```
lemillermicrosoft, line 25: Trying to find where this is used...

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
```diff
@@ -0,0 +1,36 @@
...
+
+/// <summary>
+/// Event arguments available to the Kernel.FunctionInvoking event.
+/// </summary>
+public class FunctionInvokingEventArgs : SKCancelEventArgs
+{
+    /// <summary>
+    /// Indicates if the function execution should be skipped.
+    /// </summary>
+    public bool IsSkipRequested => this._skipRequested;
```
lemillermicrosoft, line 16: Can the rendered prompt be part of the event args?

This would:
- Remove need for `InternalVariables` on `SKContext`
- Enable callers to modify the rendered prompt before invocation (not clear how that is possible now if at all)

lemillermicrosoft (commented):
Leaving some comments for now while continuing to test async.

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
RogerBarreto: This was the previous approach and @shawncal did a good comment on it, saying that we don't have to add dedicated `Prompt`

>  As far as I can see, the words "prompt" or "completion" or anything like that... they don't need to be anywhere in the Kernel.Core/Abstractions, because the Kernel shouldn't care what that string is going to be used for. It just says "Run next function"

RogerBarreto: > Enable callers to modify the rendered prompt before invocation (not clear how that is possible now if at all)

Now that's not possible but can be easily achieved adding a `Set` to the `SemanticKernelEventArgs.RenderedPrompt` property

lemillermicrosoft: Isn't that one of the goals of this PR?
> - Set: Modify a prompt content before sending it to LLM

SergeyMenshykh (approved):

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -292,6 +320,30 @@ public void Dispose()
...
+    {
+        if (eventHandler is null)
+        {
+            return null;
+        }
+
+        if (function is ISKFunctionEventSupport<TEventArgs> supportedFunction)
+        {
+            var eventArgs = await supportedFunction.PrepareEventArgsAsync(context).ConfigureAwait(false);
+            eventHandler.Invoke(this, eventArgs);
```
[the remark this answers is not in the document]
RogerBarreto: Yes, by default, if for any reason this becomes undesired, we can add a behavior option to the kernel an implement that. 

For now, I'm going to add this info to the ADR as the expected behavior.

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
RogerBarreto: It's part of the Phase 2 implementation.

dotnet/src/SemanticKernel/SkillDefinition/FunctionBase.cs
RogerBarreto: Using the `FunctionBase` was not initially in the scope of this PR but I did it due to the simplicity + benefits except for Plans:

1. `FunctionBase` is intended to be internal while `Plan` is public.
2. `FunctionBase` implements IList<ParameterView> in Parameters property while `Plan` has `ContextVariables` as Parameters property
3. Doing that for plans in this PR brings more complexity to the actual Plan implementation (many different contructors that would need to the current FunctionBase abstraction.

dotnet/src/SemanticKernel/SemanticFunctions/PromptTemplate.cs
RogerBarreto: It was, removed.  😁

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <param name="functionView">Function view details</param>
+    /// <param name="context">Context related to the event</param>
+    public FunctionInvokingEventArgs(FunctionView functionView, SKContext context) : base(functionView, context)
+    {
+    }
+
+    /// <summary>
+    /// Skip the current function invoking attempt.
+    /// </summary>
+    public void Skip()
```
lemillermicrosoft, line 30: thought: semantic functions call other functions, and we do the rendering (and therefore calling of other functions) before we give them the chance to skip.

dotnet/src/SemanticKernel.Abstractions/Events/SKEventArgs.cs
```diff
@@ -0,0 +1,34 @@
...
+        this.SKContext = context;
+    }
+
+    /// <summary>
+    /// Function view details.
+    /// </summary>
+    public FunctionView FunctionView { get; }
+
+    /// <summary>
+    /// SKContext prior function execution.
```
lemillermicrosoft, line 31: nit: prior only in the one case, for invoked that's not correct.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -25,6 +25,11 @@ public sealed class SKContext
     /// </summary>
     private CultureInfo _culture;
 
+    /// <summary>
+    /// Internal variables for Kernel use only.
+    /// </summary>
+    internal ContextVariables InternalVariables { get; }
```
lemillermicrosoft, line 31: Is it expected this will still be around after all the phases are complete?

lemillermicrosoft (approved):
Still need to understand some things, approving as this is phase 1 of a few though.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
RogerBarreto: This might go away if we have a way of passing parameters to ISKFunction that will not be part of the context (using attributes or something of like)

dotnet/src/SemanticKernel.Abstractions/Events/FunctionInvokingEventArgs.cs
RogerBarreto: Yep, you are right, currently event's don't trigger when Rendering (Even if the template makes reference to other Semantic/Native functions.

There are ongoing discussions how to solve that:

1. Add more rendering related events to Kernel
2. Add rendering events to ITemplateEngine interface
3. Add the rendering callbacks to the current Invoked/Invoking

RogerBarreto:
Closing this one, as the latest one will be the path moving forward.

PR #2468, opened 2023-08-15
.Net: [Error handling] [Part5] Replacing TemplateException exception by SKException

SergeyMenshykh:
### Motivation and Context

This PR is one of the follow-up PRs aimed at gradually replacing SK custom exceptions with SKException to provide a simple, consistent, and extensible exception model. You can find more details in the [ADR](https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0004-error-handling.md) discussing error handling.

### Description

This PR updates the template engine functionality to throw an SKException instead of a TemplateException and removes the TemplateException one.

### Contribution Checklist


dmytrostruk (approved):

dotnet/src/SemanticKernel.UnitTests/TemplateEngine/Blocks/CodeBlockTests.cs
```diff
@@ -56,11 +53,8 @@ public async Task ItThrowsIfAFunctionCallThrowsAsync()
         this._skills.Setup(x => x.GetFunction("functionName")).Returns(function.Object);
         var target = new CodeBlock("functionName", this._logger.Object);
 
-        // Act
-        var exception = await Assert.ThrowsAsync<TemplateException>(async () => await target.RenderCodeAsync(context));
-
-        // Assert
-        Assert.Equal(TemplateException.ErrorCodes.RuntimeError, exception.ErrorCode);
+        // Act & Assert
+        await Assert.ThrowsAsync<SKException>(async () => await target.RenderCodeAsync(context));
```
RogerBarreto, line 57: Not sure if this test method should be a `Task`, the others aren't. I would move this one to be sync or the others to be async

dotnet/src/SemanticKernel.UnitTests/TemplateEngine/CodeTokenizerTests.cs
```diff
@@ -137,7 +138,7 @@ public void ItThrowsWhenSeparatorsAreMissing()
         var template2 = @"call 'f\\'x";
 
         // Act
-        Assert.Throws<TemplateException>(() => this._target.Tokenize(template1));
-        Assert.Throws<TemplateException>(() => this._target.Tokenize(template2));
+        Assert.Throws<SKException>(() => this._target.Tokenize(template1));
```
RogerBarreto, line 141: Suggest making this a `[Theory]` with the multiple templates as InlineParameters parameters.

dotnet/src/SemanticKernel/TemplateEngine/Blocks/FunctionIdBlock.cs
```diff
@@ -22,8 +23,7 @@ public FunctionIdBlock(string? text, ILogger? logger = null)
         if (functionNameParts.Length > 2)
         {
             this.Logger.LogError("Invalid function name `{0}`", this.Content);
```
RogerBarreto, line 25: I would make the LogError consistent with the exception.

Instantiate the exceptions, send it to the LogError and throw it.

```
var ex = new SKExceptions(...)
this.Logger.LogError(ex);
throw ex;
```

dotnet/src/SemanticKernel/TemplateEngine/Blocks/VarBlock.cs
```diff
@@ -70,7 +71,7 @@ public string Render(ContextVariables? variables)
         {
             const string ErrMsg = "Variable rendering failed, the variable name is empty";
             this.Logger.LogError(ErrMsg);
-            throw new TemplateException(TemplateException.ErrorCodes.SyntaxError, ErrMsg);
+            throw new SKException(ErrMsg);
```
RogerBarreto, line 74: Same here, create an exceptions and pass it to the logger default behavior and throw it after.

RogerBarreto (approved):
nit. Changes!

All in all, looks good! 👍🏻

dotnet/src/SemanticKernel.UnitTests/TemplateEngine/Blocks/CodeBlockTests.cs
SergeyMenshykh: This test is for the `target.RenderCodeAsync` method, which works asynchronously and should be awaited. The test method should return a `Task` and be marked as `async` for XUnit to run and await it. While we could use the `Wait()` method to make the test method synchronous and wait for `RenderCodeAsync` to complete, why would we do that?

dotnet/src/SemanticKernel/TemplateEngine/Blocks/FunctionIdBlock.cs
SergeyMenshykh: Done, although slightly differently from the proposed solution. The reason is that there isn't a `LogError` method that accepts only an instance of an exception; an argument for a message parameter needs to be provided, whether it's a valid string or null. When the message argument is provided, the compiler starts complaining about needing a static string to support structural/semantic logging. So, I addressed the comment in a classic way by declaring a string and reusing it for both logging and the exception. The compiler warning is suppressed until we decide to implement structural/semantic logging in SK.
image

dotnet/src/SemanticKernel/TemplateEngine/Blocks/VarBlock.cs
SergeyMenshykh: Will keep it like this in order not to pass null as an argument for the message parameter of the LogError method and not to suppress CA2254 warning as a result.

dmytrostruk (approved):

dotnet/src/SemanticKernel/TemplateEngine/Blocks/VarBlock.cs
RogerBarreto: This file already suppress CA2254, move it further down into the Render method.

```
var exception = new SKException("Variable rendering failed, the variable name is empty")
this.Logger.LogError(exception, exception.Message);
throw exception;
```

PR #2472, opened 2023-08-15
.Net: Change public APIs to accept ILoggerFactory

stephentoub:
### Motivation and Context

Following best practices for libraries outlined at https://learn.microsoft.com/en-us/dotnet/core/extensions/logging-library-authors, SK should generally be taking an ILoggerFactory rather than an ILogger. This allows individual components to each get a logger with its own category applied.

### Contribution Checklist


dotnet/src/Skills/Skills.MsGraph/TaskListSkill.cs
```diff
@@ -37,19 +37,19 @@ public static class Parameters
...
-    /// <param name="logger">Logger.</param>
-    public TaskListSkill(ITaskManagementConnector connector, ILogger<TaskListSkill>? logger = null)
+    /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
+    public TaskListSkill(ITaskManagementConnector connector, ILoggerFactory? loggerFactory = null)
     {
         Ensure.NotNull(connector, nameof(connector));
 
         this._connector = connector;
-        this._logger = logger ?? new NullLogger<TaskListSkill>();
+        this._logger = loggerFactory is not null ? loggerFactory.CreateLogger(nameof(TaskListSkill)) : NullLogger.Instance;
```
shawncal, line 52: What are your thoughts about the generic versions of ILogger<T> and ILoggerFactory.CreateLogger<T>()?

i.e. syntax sugaring this line:
`var logger = kernel.LoggerFactory?.CreateLogger<TaskListSkill>() ?? NullLogger<TaskListSkill>.Instance;`

stephentoub: I think there are actually three questions here:
1. Should this use `ILogger<T>` instead of `ILogger`?
2. Should this use `NullLogger<T>.Instance` instead of `NullLogger.Instance`?
3. Should this use `CreateLogger<T>()` instead of `CreateLogger(nameof(T))`?

For (1), `ILogger<T>` is a strange API. Its sole purpose to my knowledge is in support of DI, so that you can have an `ILogger<Foo>` in your ctor and DI will end up supplying it with the result of effectively `GetService<ILoggerFactory>().CreateLogger<Foo>())`, such that it's an ILogger created with the "Foo" category; the actual `ILogger<T>` interface isn't any more useful to a consumer than just `ILogger`. So, in this use (and every other similar use in SK), there's no need for `_logger` to be an `ILogger<TaskListSkill>`, just `ILogger`.

For (2), a NullLogger doesn't do any logging, so category doesn't matter. And since with (1) we just decided we don't need `ILogger<T>`, there's no benefit to using `NullLogger<T>.Instance` instead of `NullLogger.Instance`.

(3) is the interesting one. The purpose of all of `CreateLogger<T>()`, `CreateLogger(typeof(T))`, and `CreateLogger(string)` is to supply a category to the logger; that's it, nothing more. The latter is the cheapest: it's calling to the actual interface method on ILoggerFactory, just providing the category name.  The built in LoggerFactory will use that string to look up an `ILogger` instance in its cache, and just return the existing one it has: it tries to only ever create one `ILogger` instance per category and return the same instance for all requests for that category. `CreateLogger(typeof(T))` is an extension method that creates a string from the specified Type; in .NET 8's nuget package for this, that's been made much cheaper, at least, although there's still a bit of logic executed to determine that it can just use Type.FullName (when it can... it can't in some situations)... on earlier nuget packages, this always ends up creating a new string.  And then there's `CreateLogger<T>()`. This one is frustrating, as it _always_ ends up allocating a new `Logger<T>` instance: that's because it's statically typed to return an `ILogger<T>` rather than an `ILogger`, but ILoggerFactory.CreateLogger returns an `ILogger`, and so it can't just delegate to the instance method as does `CreateLogger(Type)`.  All of that is to say, the cheapest answer here is to use `CreateLogger(nameof(Whatever))`, and switching to `CreateLogger<Whatever>()` adds unnecessary and unavoidable cost.  In a case like this, where it's done once per skill, that overhead doesn't really matter; it's more impactful in situations where you might be getting a logger per log entry, in which case you really want to use a cached instance over and over again.  Hence why I went with always using `CreateLogger(string)`.  I could switch the ones that aren't hot path to be `CreateLogger<T>` if you'd prefer.

dluc: fwiw I've rewritten the logger approach in a separate repo, using ILogger<T> and replacing NullLogger<T> with DefaultLogger<T>. The final result has improved a lot:
* DefaultLogger follows settings in appsettings.json and appsettings.development.json
* ILogger<T> makes logs more readable thanks to the inclusion of the class name, and more configurable with filters in appsettings.json

dluc:
```csharp
public static class DefaultLogger<T>
{
    public static readonly ILogger<T> Instance = GetLogFactory()
        .CreateLogger<T>();

    private static ILoggerFactory GetLogFactory()
    {
        return WebApplication.CreateBuilder().Build().Services.GetService<ILoggerFactory>()
               ?? throw new ConfigurationException("Unable to provide logger factory");
    }
}
```

example usage in a ctor:

```csharp
public AzureCognitiveSearchMemory(
    AzureCognitiveSearchConfig config,
    ILogger<AzureCognitiveSearchMemory>? log = null)
{
    this._log = log ?? DefaultLogger<AzureCognitiveSearchMemory>.Instance;
```

shawncal (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -26,9 +26,9 @@ public abstract class ClientBase
     private const int MaxResultsPerPrompt = 128;
 
     // Prevent external inheritors
-    private protected ClientBase(ILogger? logger = null)
+    private protected ClientBase(ILoggerFactory? loggerFactory = null)
     {
-        this.Logger = logger ?? NullLogger.Instance;
+        this.Logger = loggerFactory is not null ? loggerFactory.CreateLogger(this.GetType().Name) : NullLogger.Instance;
```
lemillermicrosoft, line 31: Curious your thoughts on the following options @stephentoub :

```suggestion
        this.Logger = loggerFactory is not null ? loggerFactory.CreateLogger(this.GetType()) : NullLogger.Instance;
```


```suggestion
        this.Logger = loggerFactory is not null ? loggerFactory.CreateLogger(this.GetType().FullName) : NullLogger.Instance;
```

stephentoub: Using `this.GetType()` is a bit more expensive, as CreateLogger needs to format the type into the name. In .NET 8, that's at least reasonably cheap in the most common case; for less common cases, e.g. where the type is generic, it'll involve allocations.

Using `this.GetType().FullName` is a fine choice, and more closely matches what you'd get in the common case with just `this.GetType()`.  I think it ends up coming down to how verbose we want those categories to be. If we want the full name, that's fine, we should just consistently use it in all such places.

dotnet/src/Connectors/Connectors.AI.OpenAI/CustomClient/OpenAIClientBase.cs
```diff
@@ -26,11 +26,11 @@ public abstract class OpenAIClientBase
...
     /// </summary>
     /// <param name="httpClient">The HttpClient used for making HTTP requests.</param>
-    /// <param name="logger">The ILogger used for logging. If null, a NullLogger instance will be used.</param>
-    private protected OpenAIClientBase(HttpClient? httpClient, ILogger? logger = null)
+    /// <param name="loggerFactory">The ILoggerFactory used to create a logger for logging. If null, no logging will be performed.</param>
+    private protected OpenAIClientBase(HttpClient? httpClient, ILoggerFactory? loggerFactory = null)
     {
         this._httpClient = httpClient ?? new HttpClient(NonDisposableHttpClientHandler.Instance, disposeHandler: false);
-        this._logger = logger ?? NullLogger.Instance;
+        this._logger = loggerFactory is not null ? loggerFactory.CreateLogger(this.GetType().Name) : NullLogger.Instance;
```
lemillermicrosoft, line 33: Similar, curious about this instead:

```suggestion
        this._logger = loggerFactory is not null ? loggerFactory.CreateLogger(this.GetType()) : NullLogger.Instance;
```

stephentoub: Same as https://github.com/microsoft/semantic-kernel/pull/2472#discussion_r1307615364 :)

Issue #2326, opened 2023-08-04
Call Azure OpenAI Studio as model endpoint


nacharya1:
Call Azure OpenAI Studio completion endpoint to allow customers to upload docs and talk to them through the chat completition end points of Azure OpenAI Studio.

PR #2498, opened 2023-08-18
.Net: Azure OpenAI Chat Completion with data

dmytrostruk:
### Motivation and Context

Resolves https://github.com/microsoft/semantic-kernel/issues/2326

Implementation of Azure OpenAI Chat Completion with data feature:
https://learn.microsoft.com/en-us/azure/ai-services/openai/use-your-data-quickstart

### Description

1. Implemented `AzureChatCompletionWithData` class, that uses `HttpClient` for API requests.
2. Added necessary internal classes for request/response serialization/deserialization.
3. Added unit tests.
4. Added usage example.

### Contribution Checklist


dmytrostruk:
Received great feedback from @RogerBarreto , will convert it to draft to add missing functionality.

lemillermicrosoft (commented):
Initial review, LGTM overall. Will do more when published.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
```diff
@@ -10,11 +10,21 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
 public class SKChatMessage : ChatMessageBase
```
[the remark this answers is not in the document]
dmytrostruk: Yeah, I'm not sure about that. Maybe some ambiguity between SK ChatMessage entity and ChatMessage from Azure SDK.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletionWithData/AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            yield return result;
+        }
+    }
+
+    private async Task<HttpResponseMessage> GetResponse(
+        ChatWithDataRequest request,
+        CancellationToken cancellationToken = default)
+    {
+        using var httpRequestMessage = HttpRequest.CreatePostRequest(this.GetRequestUri(), request);
```
SergeyMenshykh, line 187: nit: The code for creating an HTTP request message can be moved to the GetRequest method, and this method can be renamed to SendRequestAsync to better align method names with what they actually do.

dmytrostruk: Good suggestion, thanks! Code updated.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletionWithData/AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            yield return (ITextStreamingResult)result;
+        }
+    }
+
+    #region private ================================================================================
+
+    private const string ServerEventPayloadPrefix = "data:";
+    private const string DefaultApiVersion = "2023-06-01-preview";
+    private const string EndpointUriFormat = "{0}/openai/deployments/{1}/extensions/chat/completions?api-version={2}";
```
SergeyMenshykh, line 121: nit point to think about, and not related to the PR: In most cases, we create constants at the class level by declaring and initializing them as **private** const class fields. One of the reasons for doing this could be to avoid "magic numbers" - _literals used in the code without any explanation of their meaning_ - by giving them the right name and making them constants. However, if that is the reason, why do we declare the constants at the class level and not at the method level? Considering that almost always the constants are used by only one method of the class, as long as they are named properly, declaring them inside the method that needs them will improve the readability of the method. Developers can have their values in front of their eyes within the method that uses them, and they won't need to go back and forth between the method code and constant declarations to access the const values. There could be a performance aspect to consider, but I think it can be negligible as .NET runtimes are very sophisticated and can optimize method-scoped constants.

SergeyMenshykh (approved):

dotnet/samples/KernelSyntaxExamples/Example54_AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,108 @@
...
+        var completionWithDataConfig = GetCompletionWithDataConfig();
+
+        IKernel kernel = new KernelBuilder()
+            .WithAzureChatCompletionService(config: completionWithDataConfig)
+            .Build();
+
+        var semanticFunction = kernel.CreateSemanticFunction("Question: {{$input}}");
+
+        // First question without previous context based on uploaded content.
+        var result = await semanticFunction.InvokeAsync("How did Emily and David meet?");
```
RogerBarreto, line 75: With the kernel I would expect `kernel.RunAsync()`

dotnet/samples/KernelSyntaxExamples/Example54_AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        // Output: Emily and David, both passionate scientists, met during a research expedition to Antarctica.
+        Console.WriteLine(result);
+        Console.Write(Environment.NewLine);
+
+        // Second question based on uploaded content.
+        result = await semanticFunction.InvokeAsync("What are Emily and David studying?");
```
RogerBarreto, line 82: Same here.

`await kernel.RunAsync("What...", semanticFunction);`

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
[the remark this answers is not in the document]
RogerBarreto: @lemillermicrosoft This was a concrete implementation of our `ChatMessageBase` abstraction. So this is not a SDK class, but an extension specialization from our Abstraction, `SK` in the name as @dmytrostruk pointed out is to avoid the conflict with the `ChatMessage` class specific to the SDK.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatCompletionExtensions.cs
```diff
@@ -30,10 +30,37 @@ await foreach (var chatMessageStream in chatCompletionResult.GetStreamingChatMes
...
 
+    /// <summary>
+    /// Generate a new chat message
+    /// </summary>
+    /// <param name="chatCompletion">Target interface to extend</param>
+    /// <param name="chat">Chat history</param>
+    /// <param name="requestSettings">AI request settings</param>
+    /// <param name="cancellationToken">Async cancellation token</param>
+    /// <returns>Stream the generated chat message in string format</returns>
+    public static async IAsyncEnumerable<string> GenerateMessagesStreamAsync(
```
RogerBarreto, line 46: Remove this extension, the first will work for your scenarios as the `WithData` doesn't support multiple Results / Choices.

This method will also Bug/Freeze out if used with the current Azure SDK Completions implementation.

RogerBarreto (changes requested):

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatCompletionExtensions.cs
dmytrostruk: Thanks! First extension doesn't work in my scenario, so I decided to use `IChatStreamingResult` directly.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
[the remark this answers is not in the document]
lemillermicrosoft: I'd argue the private `ChatMessage` should be `SKChatMessage` and the extension be `ChatMessage` as there is nothing really SK about it, is there? It's in the AzureSdk namespace, which makes me wonder if it should be named AzureChatMessage?

@RogerBarreto  -- is there a reason why `ChatMessage` is private and not public?

dotnet/samples/KernelSyntaxExamples/Example54_AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,111 @@
...
+        // Second question based on uploaded content.
+        result = await kernel.RunAsync("What are Emily and David studying?", semanticFunction);
+
+        // Output:
+        // They are passionate scientists who study glaciology,
+        // a branch of geology that deals with the study of ice and its effects.
+        Console.WriteLine(result);
+
+        Console.Write(Environment.NewLine);
+        Console.Write(Environment.NewLine);
```
RogerBarreto, line 93:
```suggestion
        Console.WriteLine(Environment.NewLine);
```

dotnet/samples/KernelSyntaxExamples/Example54_AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                // Output:
+                // They are passionate scientists who study glaciology,
+                // a branch of geology that deals with the study of ice and its effects.
+                Console.Write(message.Content);
+            }
+        }
+
+        Console.Write(Environment.NewLine);
+        Console.Write(Environment.NewLine);
```
RogerBarreto, line 62: Is there a special reason for double lines breaks?

nit. One line code with same effect.

```suggestion
        Console.WriteLine(Environment.NewLine);
```

RogerBarreto (approved):
LGTM. 

nit comments.

dotnet/samples/KernelSyntaxExamples/Example54_AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,109 @@
...
+            .Build();
+
+        var semanticFunction = kernel.CreateSemanticFunction("Question: {{$input}}");
+
+        // First question without previous context based on uploaded content.
+        var result = await kernel.RunAsync("How did Emily and David meet?", semanticFunction);
+
+        // Output: Emily and David, both passionate scientists, met during a research expedition to Antarctica.
+        Console.WriteLine(result);
+        Console.Write(Environment.NewLine);
```
RogerBarreto, line 81:
```suggestion
        Console.WriteLine();
```

dotnet/samples/KernelSyntaxExamples/Example54_AzureChatCompletionWithData.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        // First question without previous context based on uploaded content.
+        chatHistory.AddUserMessage("How did Emily and David meet?");
+
+        // Chat Completion example
+        string reply = await chatCompletion.GenerateMessageAsync(chatHistory);
+
+        // Output: Emily and David, both passionate scientists, met during a research expedition to Antarctica.
+        Console.WriteLine(reply);
+        Console.Write(Environment.NewLine);
```
RogerBarreto, line 44:
```suggestion
        Console.WriteLine();
```

Issue #2431, opened 2023-08-12
Python: Echo_text_completion service for Prompt construction testing


a-little-srdjan:
What:
There does not appear to be a simple way to retrieve the actual prompt that is sent to *_text_completion services from an SKFunction. In particular the _function param in SKFunction cannot be queried for the prompt it will send.

Why:
When testing complex SKFunction and Context chains, it would be useful to be able to inspect the final prompt that is being sent. This can help both with unit testing complex chains as well as help with better understanding of how SKFunctions construct the prompts.

How:
I will implement a simple ECHO service that simply returns the prompt it was passed as the output in Context. This ought to accomplish the above. Please assign the issue to me.

nacharya1:
@a-little-srdjan thanks for letting us know and we will look into improve support for this in the future.

a-little-srdjan:
Hey @nacharya1 I am almost done with a patch that adds Echo like classes that can be used when instantiating a kernel object. I should have the first version wrapped up this week. If you want please assign this to me.

a-little-srdjan:
Simple example:
```
from semantic_kernel.connectors.ai.echo import EchoTextCompletion
...
kernel.add_text_completion_service("e", EchoTextCompletion())
summarize = kernel.create_semantic_function("{{$input}}\n\nOne line TLDR with the fewest words."
print(summarize("prompt"))
```

output:

```
prompt


One line TLDR with the fewest words.
```

lemillermicrosoft:
Assigning to @shawncal to help advise what set of services we'd like to have in the kernel.

a-little-srdjan:
Core python functionality: https://github.com/microsoft/semantic-kernel/pull/2500

matthewbolanos:
Pre/Post hooks should be leveraged to support this functionality.

PR #2500, opened 2023-08-18
Python: Issue 2431

a-little-srdjan:
### Motivation and Context

This is an implementation for #2431  

The echo completion service allows for easier unit testing of prompt constructions (e.g. through skill chaining), and a way to easier debug prompt construction issues.

The key motivation for writing up this echo is because the final prompt is constructed by _function delegate within SKFunction that does not record constructed prompts. Thus they are not available for introspection.

### Description

The changes are minimal, the introduced classes satisfy the Base interface and simply echo back the prompt.

### Contribution Checklist


a-little-srdjan:

lemillermicrosoft:
Thanks again for the PR @a-little-srdjan -- I'm going to mark it as Draft for now until we get confirmation in the issue #2431 that this is something we want to make available by default. I definitely understand the desire, I've built this myself in my projects. But we may not choose to ship this in the core kernel. cc @shawncal

a-little-srdjan:
hey @lemillermicrosoft of course.  To your point, I've had to do it a few times as well, so that internally we can post the prompts to OpenAI wrapper APIs, hence the idea. Nevertheless, pausing on this makes perfect sense. I was also thinking if it makes more sense to move the code under test module that way it is clearer that it is not some sort of a core functionality.

Thanks again for the write up

evchaki:
@moonbox3 can you see if this is something that is still needed.

moonbox3:
Closing this PR as we are going to use pre/post hooks to support this. Thank you for your effort, @a-little-srdjan!

PR #2263, opened 2023-08-01
.Net: [Error handling] [Part1] SKException is instantiable and public

SergeyMenshykh:
### Motivation and Context

This PR is the first in a chain of following PRs aimed at providing a simple, consistent, and extensible exception model. You can find more details in the [ADR](https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0004-error-handling.md) discussing error handling.

### Description

This PR changes SKException class so that it's instantiable and public, throws SKException instead of  AzureSearchMemoryException, and removes AzureSearchMemoryException.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #2516, opened 2023-08-21
.Net: [Error handling] [Part7] Removing IsCriticalException extension method

SergeyMenshykh:
### Motivation and Context
This PR is one of the follow-up PRs aimed at gradually replacing SK custom exceptions with SKException to provide a simple, consistent, and extensible exception model. You can find more details in the [ADR](https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0004-error-handling.md) discussing error handling.

### Description
This PR removes the IsCriticalException extension method and updates all the code that used it. In a few cases, the change triggers the `CA1031 - Do not catch general exception types` warning, which is temporarily suppressed to keep this PR focused on removing the IsCriticalException extension method. The proper fix for these cases will be applied by one of the following PRs dedicated to the fix.

### Contribution Checklist


dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -160,11 +161,12 @@ public Plan CreatePlan(string goal)
                             nextStep.Observation = result;
                         }
                     }
-                    catch (Exception ex) when (!ex.IsCriticalException())
+                    catch (Exception ex)
                     {
                         nextStep.Observation = $"Error invoking action {nextStep.Action} : {ex.Message}";
                         this._logger?.LogWarning(ex, "Error invoking action {Action}", nextStep.Action);
                     }
+#pragma warning restore CA1031 // Do not catch general exception types. The code will change in one of the next PRs to allow SK to throw exceptions instead of swallowing them. As a result, this suppression will be removed as well.
```
lemillermicrosoft, line 169: Catching them is by design here.

lemillermicrosoft (commented):
Why are we removing `IsCriticalException`? The ADR only mentions it in that it will be updated to include `StackOverflowException and OutOfMemoryException`.

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
SergeyMenshykh: Fixed.

dmytrostruk (approved):
As mentioned by @lemillermicrosoft , would be nice to see updates in ADR with details why `IsCriticalException` is removed/replaced.

SergeyMenshykh:
> Why are we removing `IsCriticalException`? The ADR only mentions it in that it will be updated to include `StackOverflowException and OutOfMemoryException`.

Yes, right, it's not part of ADR. It popped up later as a comment in one of the 'Erro handling' PRs where we (@RogerBarreto , @stephentoub and me) agreed that there's no value in having it.
image
Link - https://github.com/microsoft/semantic-kernel/pull/2263/files/02abc1c868a259e081bb144f869cf09fcecd8416#diff-160896cef4bb4f4f3256d65cadf09f8d0e223b06a1b1c1184647780e8d319a37

Let's discuss this on parking lot. CC: @shawncal

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -120,7 +120,7 @@ private async Task<IReadOnlyList<ITextStreamingResult>> ExecuteGetCompletionsAsy
 
             return completionResponse.ConvertAll(c => new TextCompletionStreamingResult(c));
         }
-        catch (Exception e) when (e is not AIException && !e.IsCriticalException())
+        catch (Exception e) when (e is not AIException)
```
RogerBarreto, line 123: MSIL for catch + when is way more costly than using the below. Apply the same for the others.

```suggestion
        catch (AIException aiex) 
        { 
            throw; 
        }
        catch (Exception ex)
```

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -160,7 +162,7 @@ public Plan CreatePlan(string goal)
                             nextStep.Observation = result;
                         }
                     }
-                    catch (Exception ex) when (!ex.IsCriticalException())
+                    catch (Exception ex)
                     {
                         nextStep.Observation = $"Error invoking action {nextStep.Action} : {ex.Message}";
                         this._logger?.LogWarning(ex, "Error invoking action {Action}", nextStep.Action);
```
RogerBarreto, line 168: The exception is being swallowed and no log data is being stored about it.

LogWarning should be LogError or at least include the exception details.

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -357,11 +360,12 @@ private async Task<string> InvokeActionAsync(string actionName, Dictionary<strin
 
             return result.Result;
         }
-        catch (Exception e) when (!e.IsCriticalException())
+        catch (Exception e)
         {
             this._logger?.LogError(e, "Something went wrong in system step: {0}.{1}. Error: {2}", targetFunction.SkillName, targetFunction.Name, e.Message);
             return $"Something went wrong in system step: {targetFunction.SkillName}.{targetFunction.Name}. Error: {e.Message} {e.InnerException.Message}";
```
RogerBarreto, line 366:
```suggestion
            return $"Something went wrong in system step: {targetFunction.SkillName}.{targetFunction.Name}. Error: {e.Message} {e.InnerException?.Message}";
```

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -160,17 +160,19 @@ public FunctionView Describe()
         CompleteRequestSettings? settings = null,
         CancellationToken cancellationToken = default)
     {
+#pragma warning disable CA1031 // Do not catch general exception types. The code will change in one of the next PRs to allow SK to throw exceptions instead of swallowing them. As a result, this suppression will be removed as well.
         try
         {
             return await this._function(null, settings, context, cancellationToken).ConfigureAwait(false);
         }
-        catch (Exception e) when (!e.IsCriticalException())
+        catch (Exception e)
```
RogerBarreto, line 168: By removing the function `IsCriticalException` we are pratically considering any exceptions to be "critical" (should throw).

Two options: 
1. Remove all the catch block
2. Keep the catch just to log and `throw` after

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -203,13 +204,14 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
                     return context;
                 }
             }
-            catch (Exception e) when (!e.IsCriticalException())
+            catch (Exception e)
```
RogerBarreto, line 207: By removing the function `IsCriticalException` we are pratically considering any exceptions to be "critical" (should throw).

Two options: 
1. Remove all the catch block
2. Keep the catch just to log and `throw` after

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -203,13 +204,14 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
...
             {
                 this.Logger.LogError(e, "Something went wrong in pipeline step {0}: {1}.{2}. Error: {3}",
```
RogerBarreto, line 209:
```suggestion
                this._logger.LogError(e, "Something went wrong in pipeline step {0}: {1}.{2}. Error: {3}",
```

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -506,7 +508,7 @@ private static (Func<ITextCompletion?, CompleteRequestSettings?, SKContext, Canc
                     {
                         return parser(value, context.Culture);
                     }
-                    catch (Exception e) when (!e.IsCriticalException())
+                    catch (Exception e)
```
RogerBarreto, line 511: By removing the function `IsCriticalException` we are pratically considering any exceptions to be "critical" (should throw).

Two options: 
1. Remove all the catch block
2. Keep the catch just to log and `throw` after

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -738,14 +740,16 @@ private static void TrackUniqueParameterType(ref bool hasParameterType, MethodIn
 
                     // First try to parse using the supplied culture (or current if none was supplied).
                     // If that fails, try with the invariant culture and allow any exception to propagate.
+#pragma warning disable CA1031 // Do not catch general exception types. The code will change in one of the next PRs to allow SK to throw exceptions instead of swallowing them. As a result, this suppression will be removed as well.
                     try
                     {
                         return converter.ConvertFromString(context: null, cultureInfo, input);
                     }
-                    catch (Exception e) when (!e.IsCriticalException() && cultureInfo != CultureInfo.InvariantCulture)
+                    catch (Exception) when (cultureInfo != CultureInfo.InvariantCulture)
```
RogerBarreto, line 748: By removing the function `IsCriticalException` we are pratically considering any exceptions to be "critical" (should throw).

Two options: 
1. Remove all the catch block
2. Keep the catch just to log and `throw` after

RogerBarreto (changes requested):
By removing the function `IsCriticalException` we are pratically considering any exceptions to be "critical" (should throw), I would expect the change to be either: 

1. Remove all the catch block
2. Keep the catch just to log and `throw` after

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
dmytrostruk: I will tag @lemillermicrosoft for input here.

> The exception is being swallowed

As far as I know, this is expected behavior for StepwisePlanner, when it receives an exception, it will put details in `Observation` field and proceed to the next step.

> LogWarning should be LogError

In case this is correct behavior, we shouldn't log it as error, because it doesn't break the application flow, instead it's just transient error. Previously we used Debug level here, and I updated it to Warning, and it's still not clear whether it was right choice.

> or at least include the exception details.

As far as I can see, we include `ex` instance as first parameter to `LogWarning` method.

lemillermicrosoft: Yes thank you @dmytrostruk . I think LogDebug is probably best in this scenario (I was gonna change in another PR).

SergeyMenshykh:
Closing the PR due to a lack of consensus regarding the removal of the method. Additionally, it seems that the method can be useful in distinguishing between actionable and non-actionable exceptions.

PR #2528, opened 2023-08-22
.Net: Support named args

hario90:
### Motivation and Context

  1. Why is this change required? Native functions invoked within semantic functions are currently limited to being called with at most one argument for the input variable.
  2. What problem does it solve? Calling native functions with mutliple arguments from semantic functions.
  3. What scenario does it contribute to? Template Engine

### Description

Please see the ADR for more context on decisions.

To understand the implementation, I would start with the changes in CodeTokenizer.cs and CodeBlock.cs.

### Contribution Checklist


RogerBarreto (commented):
IMHO I would prefer a pattern like: 
```
{{Skill.MyFunction(street="123 Main St", age=20)}}
{{Skill.MyFunction(address.street="123 Main St", address.latitude=-20.0)}}
``` 

Pros:
- Legibility / Clarity about it being a function
- Parameters can be string, numbers, booleans between `,` 
- Easy to distinguish and parse
- Quotes are not needed for numbers and booleans

Const:
- Little more verbose (extra comas between parameters)

hario90:
@RogerBarreto While I think the syntax that includes parenthesis and commas is more similar to how other languages support function calls, I think the downsides are:

- that this syntax isn't similar to [Guidance's syntax](https://github.com/guidance-ai/guidance). Aligning with Guidance's syntax was a driving concern for Matthew.
- this could introduce a major breaking change since we currently don't support parenthesis for function calls. 

I think as long as we're supporting only function arguments inside of curly brackets when the code starts with the function name, it should still be readable. If the plan is to eventually support more than function arguments, then that would require a way to separate function calls from everything else, e.g:

```
Normal text
{{MySkill.FunctionName(arg1="some text", arg2="more text") $randomVariable}} More text
``` 

But currently, I don't think there's a need for that kind of flexibility.

docs/decisions/005-support-multiple-named-args-in-template-function-calls.md
```diff
@@ -0,0 +1,111 @@
...
+
+Example
+
+```handlebars
+
+{{MyFunction "inputVal" street="123 Main St" zip="98123" city="Seattle"}}
+
+```
+
+* Don't allow whitespace before and after equals sign for named args because without commas, differentiating between arguments becomes more difficult to read.
```
SergeyMenshykh, line 103: Isn't this restriction too strict? It may frustrate plugin developers and be a very frequent source of issues.

dotnet/src/SemanticKernel/TemplateEngine/Blocks/CodeBlock.cs
```diff
@@ -54,27 +54,15 @@ public override bool IsValid(out string errorMsg)
...
-                this.Logger.LogError(errorMsg);
-                return false;
-            }
+            errorMsg = "Unexpected named argument found. Expected function name first.";
+            this.Logger.LogError(errorMsg);
+            return false;
         }
 
-        if (this._tokens.Count > 2)
+        if (this._tokens.Count > 1 && !this.IsValidFunctionCall(out errorMsg))
```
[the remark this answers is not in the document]
hario90: The helper function `IsValidFunctionCall` will handle logging an error message in this case.

dotnet/src/SemanticKernel/TemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,105 @@
...
+internal sealed class NamedArgBlock : Block, ITextRendering
+{
+    internal override BlockTypes Type => BlockTypes.NamedArg;
+
+    internal string Name { get; } = string.Empty;
+
+    public string GetValue(ContextVariables? variables)
+    {
+        var valueIsValidValBlock = this._valBlock != null && this._valBlock.IsValid(out var errorMessage);
+        var valueIsValidVarBlock = this._argValueAsVarBlock != null && this._argValueAsVarBlock.IsValid(out var errorMessage2);
```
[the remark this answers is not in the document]
hario90: Fixed

dotnet/src/SemanticKernel/TemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,105 @@
...
+        if (valueIsValidValBlock && this._valBlock != null)
```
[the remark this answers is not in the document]
hario90: I included the null checks because I get this warning when I don't include them: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings?f1url=%3FappId%3Droslyn%26k%3Dk(CS8602).

dotnet/src/SemanticKernel/TemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        if (valueIsValidVarBlock && this._argValueAsVarBlock != null)
+        {
+            return this._argValueAsVarBlock.Render(variables);
+        }
+
+        return string.Empty;
+    }
+
+    private readonly VarBlock _argNameAsVarBlock;
```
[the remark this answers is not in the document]
hario90: Fixed

dotnet/src/SemanticKernel/TemplateEngine/Blocks/NamedArgBlock.cs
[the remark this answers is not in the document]
SergeyMenshykh: You can use the `!` null-forgiving operator to suppress the warnings. It's safe because the null check has already been performed, and it's a little bit easier to read - there's no need to think about why the null check has been done two times.
image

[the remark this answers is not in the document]
hario90: Forgot about that, good idea.

docs/decisions/005-support-multiple-named-args-in-template-function-calls.md
hario90: So I updated the decision outcome to allow whitespace because it turns out to not be that much more complicated to support. I have pushed the updated code in the latest commits as well.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,156 @@
...
+
+namespace Microsoft.SemanticKernel.TemplateEngine.Prompt.Blocks;
+
+internal sealed class NamedArgBlock : Block, ITextRendering
+{
+    internal override BlockTypes Type => BlockTypes.NamedArg;
+
+    internal string Name { get; } = string.Empty;
+
+    public string GetValue(ContextVariables? variables)
```
SergeyMenshykh, line 15: Add XML comments to all classes and their members in the PR.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,156 @@
...
+            this.Logger.LogError("Invalid named argument `{0}`.", this.Content);
+            throw new SKException($"Invalid named argument `{this.Content}`. A named argument can contain at most one equals sign separating the arg name from the arg value.");
+        }
+
+        if (argParts.Length == 2)
+        {
+            var trimmedArgParts = NamedArgBlock.GetTrimmedParts(text);
+            this.Name = trimmedArgParts[0];
+            this._argNameAsVarBlock = new VarBlock($"{Symbols.VarPrefix}{trimmedArgParts[0]}");
+            var secondPart = trimmedArgParts[1];
```
SergeyMenshykh, line 47: nit: Consider giving a meaningful name to the expression on the right side of the equal sign. Instead of naming it `secondPart`, use a proper name that reflects its value, such as `argumentValue`.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,156 @@
...
+                this.Logger.LogError("Invalid named argument `{0}`", this.Content);
+                throw new SKException($"A function named argument must contain a quoted value or variable after the '{Symbols.NamedArgBlockSeparator}' character.");
+            }
+            else if (secondPart[0] == Symbols.VarPrefix)
+            {
+                this._argValueAsVarBlock = new VarBlock(secondPart);
+            }
+            else
+            {
+                this._valBlock = new ValBlock(trimmedArgParts[1]);
```
SergeyMenshykh, line 59: nit: Use `secondPart` variable instead of 'trimmedArgParts[1]'.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,156 @@
...
+        return string.Empty;
+    }
+
+    public NamedArgBlock(string? text, ILoggerFactory? logger = null)
+        : base(NamedArgBlock.TrimWhitespace(text), logger)
+    {
+        var argParts = this.Content.Split(Symbols.NamedArgBlockSeparator);
+        if (argParts.Length > 2)
+        {
+            this.Logger.LogError("Invalid named argument `{0}`.", this.Content);
```
SergeyMenshykh, line 38: You can use either `this.Content` or `text`, but not both, to reduce the mental load while reading the code.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,156 @@
...
+            var trimmedArgParts = NamedArgBlock.GetTrimmedParts(text);
+            this.Name = trimmedArgParts[0];
+            this._argNameAsVarBlock = new VarBlock($"{Symbols.VarPrefix}{trimmedArgParts[0]}");
+            var secondPart = trimmedArgParts[1];
+            if (secondPart.Length == 0)
+            {
+                this.Logger.LogError("Invalid named argument `{0}`", this.Content);
+                throw new SKException($"A function named argument must contain a quoted value or variable after the '{Symbols.NamedArgBlockSeparator}' character.");
+            }
+            else if (secondPart[0] == Symbols.VarPrefix)
```
SergeyMenshykh, line 53: nit: the `else` keyword is not necessary here.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,156 @@
...
+        var argParts = this.Content.Split(Symbols.NamedArgBlockSeparator);
+        if (argParts.Length > 2)
+        {
+            this.Logger.LogError("Invalid named argument `{0}`.", this.Content);
+            throw new SKException($"Invalid named argument `{this.Content}`. A named argument can contain at most one equals sign separating the arg name from the arg value.");
+        }
+
+        if (argParts.Length == 2)
+        {
+            var trimmedArgParts = NamedArgBlock.GetTrimmedParts(text);
```
SergeyMenshykh, line 44: There's no need for this line since the `argParts` array already contains trimmed argument name and value. The array was created by splitting the `this.Content` variable value, which was trimmed during the initialization of the class using `base(NamedArgBlock.TrimWhitespace(text))`."

hario90: Good catch!

SergeyMenshykh (approved):

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,182 @@
...
+        return string.Empty;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="NamedArgBlock"/> class.
+    /// </summary>
+    /// <param name="text">Raw text parsed from the prompt template.</param>
+    /// <param name="logger">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
+    /// <exception cref="SKException"></exception>
+    public NamedArgBlock(string? text, ILoggerFactory? logger = null)
```
dmytrostruk, line 54: I would place constructor above `GetValue` method and remove `return` statement, since it's not required in constructor. Ideally, it would be great to move the logic with exception throwing and logging to separate methods to keep constructor clean and simple.

dotnet/samples/KernelSyntaxExamples/Example56_TemplateNativeFunctionsWithMultipleArguments.cs
```diff
@@ -0,0 +1,84 @@
...
+        var promptRenderer = new PromptTemplateEngine();
+        var renderedPrompt = await promptRenderer.RenderAsync(FunctionDefinition, context);
+        Console.WriteLine(renderedPrompt);
+
+        // Run the prompt / semantic function
+        var haiku = kernel.CreateSemanticFunction(FunctionDefinition, maxTokens: 150);
+
+        // Show the result
+        Console.WriteLine("--- Semantic Function result");
+        var result = await kernel.RunAsync(variables, haiku);
```
lemillermicrosoft, line 69:
```suggestion
        var result = await kernel.RunAsync(context.Variables, haiku);
```

Is this possible and then you can remove the creation of `variables`.

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/PromptTemplateEngineTests.cs
```diff
@@ -75,18 +78,24 @@ public void ItRendersVariables()
...
+        Assert.Equal("food a='b' c = $d", blocks[9].Content);
+        Assert.Equal("food a='b' c = $d", updatedBlocks[9].Content);
+        Assert.Equal(BlockTypes.Code, blocks[9].Type);
+        Assert.Equal(BlockTypes.Code, updatedBlocks[9].Type);
+
         // Arrange
         this._variables.Set("x11", "x11 value");
         this._variables.Set("a", "a value");
         this._variables.Set("_a", "_a value");
+        this._variables.Set("c", "c value");
```
lemillermicrosoft, line 90: What is this doing? Should this be `d` instead?

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -0,0 +1,182 @@
...
+        return this.Content;
+    }
+
+    /// <summary>
+    /// Returns whether the named arg block has valid syntax.
+    /// </summary>
+    /// <param name="errorMsg">An error message that gets set when the named arg block is not valid.</param>
+    /// <returns></returns>
+#pragma warning disable CA2254 // error strings are used also internally, not just for logging
+    public override bool IsValid(out string errorMsg)
```
dmytrostruk, line 101: I found an issue in this method. I tested it using your example from unit test, but I extended it a little bit:
`{{my.sayAge name=\"Mario\" birthdate=$birthdate exclamation='Wow, that\\'s surprising' test=5}}`, so I added `test=5` parameter. 

Probably out of scope of this PR, but it looks like passing plain numbers is not supported at the moment, which is a little bit confusing. I'm wondering if it makes sense to add this functionality in the future. Maybe it worth to create separate issue for that and investigate it further.

The problem in this method - if I use `test=5` syntax, which is invalid at the moment, the value check will fail in line 114 and there will be some information in `errorMsg` out variable. Then, in line 128 we are performing name check, and in my case the name is valid, but it uses the same `errorMsg` out variable, which means that this variable will be overwritten, and as a result - empty. In this case, method `RenderAsync` will throw `SKException` without any information:
image

Overall, I think we should avoid passing `errorMsg` as out variable through the stack trace, because it's hard to maintain its state. I'm wondering, if we can just immediately throw an exception in case of invalid syntax with error message. This will resolve the problem which I described above and potential similar problems in the future.

As part of this PR, it would be great to add more unit tests, specifically with invalid scenarios and verify the expected behavior.

hario90: Discussed offline - here's the summary:

* Keep or remove the out variable for `IsValid` a method on the _base_ class for NamedArgBlock?
  * Outcome: keep. Re-evaluate as a separate issue.
  * Reason: This would change the interface for many other classes, and I don't fully understand the original intent behind it
* How to avoid confusing blank exceptions?
  * Outcome: throw immediately after encountering an issue.
  * Reason: Simplest solution, less error prone, doesn't require API change.
*  Add more unit tests for valid and invalid scenarios
* To support numbers/unquoted arg values?
  * Currently working as designed (see ADR). Only quoted and variable values are supported.
  * I'm open to re-evaluating this as a separate issue but wanted to keep things simple.
  
The reason I came to this decision had to do with this line from the ADR for supporting multiple args in native functions:

> Added support for non-string arguments. All C# primitive types and many core .NET types are supported, with their corresponding TypeConverters used to parse the string context variable into the appropriate type. Custom types attributed with TypeConverterAttribute may also be used, and the associated TypeConverter will be used as is appropriate.

So the args are expected to be saved as strings in context anyways even if the value looks like an int in the template. However, it is still possible for the engine to be flexible enough to save all non-variable values as strings to context.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
hario90: Makes sense. This started out as a property and never got moved. And good catch on the `return`.

dotnet/samples/KernelSyntaxExamples/Example56_TemplateNativeFunctionsWithMultipleArguments.cs
hario90: Yep, tested locally and this works. Thanks!

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/PromptTemplateEngineTests.cs
hario90: It's setting a default value for c which should get overwritten by the value for `$d` which currently isn't set. I can update the test to include setting `$d` as well. It's not super important since the prompt template engine doesn't render variables for function calls.

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/NamedArgBlock.cs
hario90: Added some more tests for the error message in my latest commit. I already had a test for unquoted/non-variable values, so I updated that to check the error message.

RogerBarreto (approved):
LGTM

PR #2539, opened 2023-08-23
.Net: Update Plan Result in State with matching outputs

### LINKED ISSUE #2525: .Net: Sequential Planner results are always prepended with a newline character
**Describe the bug**
I tried notebook 05-using-the-planner, planner results are always prepended with a newline character.
I also tried a simple my process, but the result was the same.

**To Reproduce**
Try notebook 05-using-the-planner, and see results below.
You can see the blank line.
image

**Expected behavior**
Output the original return value of the function verbatim.

**Screenshots**

**Platform**
 - OS: Mac
 - IDE: VS Code
 - Language:  C#
 - Source: Microsoft.SemanticKernel 0.20.230821.4-preview

**Additional context**
I tried a simple my process. Function return "Long sleeve shirt", but results are prepended with a newline character.
image
image
image

lemillermicrosoft:
Update Plan Result in State with matching outputs to include the new result value. If there is already a current plan result, append the new result value to it. If there is no current plan result, set the result value as the new plan result.

Resolves #2525

### Contribution Checklist


SergeyMenshykh (approved):

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -268,8 +268,14 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
...
-                this.State.TryGetValue(DefaultResultKey, out string? currentPlanResult);
-                this.State.Set(DefaultResultKey, string.Join("\n", currentPlanResult?.Trim(), resultValue));
+                if (this.State.TryGetValue(DefaultResultKey, out string? currentPlanResult))
+                {
+                    this.State.Set(DefaultResultKey, $"{currentPlanResult}\n{resultValue}");
+                }
+                else
+                {
+                    this.State.Set(DefaultResultKey, resultValue);
+                }
```
RogerBarreto, line 278: nit.

```suggestion
                this.State.TryGetValue(DefaultResultKey, out string? currentPlanResult)
                this.State.Set(DefaultResultKey, (currentPlanResult is null) ? resultValue : $"{currentPlanResult}\n{resultValue}");
```

RogerBarreto (approved):

dotnet/src/SemanticKernel/Planning/Plan.cs
lemillermicrosoft: Heh yes I actually had this originally and then felt the urge to leverage the result of TryGetValue. Will leave as-is for now.

PR #2542, opened 2023-08-23
.Net: [OpenAPI] Fallback to operation summary if no description provided

SergeyMenshykh:
### Motivation and Context

The code provides a fallback mechanism to allow the use of operation summary if no operation description is provided.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Skills/Skills.OpenAPI/OpenApi/OpenApiDocumentParser.cs
+++ b/dotnet/src/Skills/Skills.OpenAPI/OpenApi/OpenApiDocumentParser.cs
@@ -186,7 +186,7 @@ internal sealed class OpenApiDocumentParser : IOpenApiDocumentParser
                 string.IsNullOrEmpty(serverUrl) ? null : new Uri(serverUrl),
                 path,
                 new HttpMethod(method),
-                operationItem.Description,
+                string.IsNullOrEmpty(operationItem.Description) ? operationItem.Summary : operationItem.Description,
                 CreateRestApiOperationParameters(operationItem.OperationId, operationItem.Parameters),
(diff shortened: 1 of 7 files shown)


lemillermicrosoft (approved):

dotnet/src/Skills/Skills.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -186,7 +186,7 @@ private static List<RestApiOperation> CreateRestApiOperations(string? serverUrl,
                 string.IsNullOrEmpty(serverUrl) ? null : new Uri(serverUrl),
                 path,
                 new HttpMethod(method),
-                operationItem.Description,
+                string.IsNullOrEmpty(operationItem.Description) ? operationItem.Summary : operationItem.Description,
```
markwallace-microsoft, line 189: Use the null-coalescing operator ??

markwallace-microsoft (approved):
LGTM just one minor comment

dotnet/src/Skills/Skills.OpenAPI/OpenApi/OpenApiDocumentParser.cs
dmytrostruk: @markwallace-microsoft I think it's not going to work if `operationItem.Description` is not null, but empty string.

SergeyMenshykh: It won't compile because the `string.IsNullOrEmpty(operationItem.Description)` method returns Boolean while the `operationItem.Summary` property returns string:
image

PR #2533, opened 2023-08-22
Update python-integration-tests.yml

shawncal:
Adding back the paths filter on the Python integration tests GitHub actions.

This ensures that the tests only run when changes under the python/ folder are included. Otherwise, it skips to the final validation step.

PR #2557, opened 2023-08-24
Python: Integration tests 'needs' path filter job

awharrison-28:
### Motivation and Context

PRs #2533 and #2555 updated the integration tests to trigger off of python changes only, but did not actually have the resulting jobs require the path filtering. This means that the tests never run. 

### Description
Add path-filter job as a requirement for subsequent jobs.

### Contribution Checklist

### DIFF
--- a/python/samples/kernel-syntax-examples/action_planner.py
+++ b/python/samples/kernel-syntax-examples/action_planner.py
@@ -13,7 +13,7 @@ async def main():
     api_key, org_id = sk.openai_settings_from_dot_env()
 
     kernel.add_chat_service(
-        "gpt-3.5", OpenAIChatCompletion("gpt-3.5-turbo", api_key, org_id)
+        "chat-gpt", OpenAIChatCompletion("gpt-3.5-turbo", api_key, org_id)
     )
     kernel.import_skill(MathSkill(), "math")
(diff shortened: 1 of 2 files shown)


python/samples/kernel-syntax-examples/action_planner.py
```diff
@@ -13,7 +13,7 @@ async def main():
     api_key, org_id = sk.openai_settings_from_dot_env()
 
     kernel.add_chat_service(
-        "gpt-3.5", OpenAIChatCompletion("gpt-3.5-turbo", api_key, org_id)
+        "chat-gpt", OpenAIChatCompletion("gpt-3.5-turbo", api_key, org_id)
```
alexchaomander, line 16: Did this need to change for this PR?

lemillermicrosoft: I'm guessing it's to help test the behavior of the template.

awharrison-28: Yes, it's to help test the template behavior. It's harmless - just the lookup name for the service. The service itself is not changing and it's a syntax example, so no tests are affected.

dmytrostruk (approved):

Issue #2527, opened 2023-08-22
Extract template engine from Kernel.Core


markwallace-microsoft:
Template engine will be moved to it's own project i.e. `Extensions\TemplateEngine`

```[tasklist]
### Tasks
- [x] Create new project and move code
- [ ] Create an ADR to describe the change, project naming and impact for consumers
- [ ] Make required build/pipeline changes
```

PR #2562, opened 2023-08-24
.Net: Extract template engine from Kernel.Core

markwallace-microsoft:
### Motivation and Context

  1. Why is this change required? This is part of the work to extract functionality from Semantic Kernel core prior to 1.0 release.
  2. What problem does it solve? Provides a pattern to follow for additional template engine implementations.
  3. What scenario does it contribute to? Contributes to the goal of reducing the Semantic Kernel core to focus on AI orchestration.
  4. If it fixes an open issue, please link to the issue here. https://github.com/microsoft/semantic-kernel/issues/2527

### Description

- Create new template engine project in the extensions folder
- Move template engine code from semantic kernel abstractions and core
- Refactor `IPromptTemplateEngine` so that it doesn't depend on details of the default implementation
- Refactor `PromptTemplateEngine` to match the new interface definition
- Move template engine unit tests to the extensions unit tests project

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/TemplateEngine/IPromptTemplateEngine.cs
```diff
@@ -14,14 +13,12 @@ namespace Microsoft.SemanticKernel.TemplateEngine;
...
-    /// Given a prompt template string, extract all the blocks (text, variables, function calls)
+    /// Given a prompt template string, extract all the variable names
     /// </summary>
     /// <param name="templateText">Prompt template (see skprompt.txt files)</param>
-    /// <param name="validate">Whether to validate the blocks syntax, or just return the blocks found, which could contain invalid code</param>
-    /// <returns>A list of all the blocks, ie the template tokenized in text, variables and function calls</returns>
-    IList<Block> ExtractBlocks(
-        string? templateText,
-        bool validate = true);
+    /// <returns>A list of all the variable names</returns>
```
lemillermicrosoft, line 19: nit: If we move towards something like `Guidance` -- we'll have both input and output variable names. Might want to consider that with renaming now (ExtractInputVariableNames or something)

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/PromptTemplateEngineTests.cs
```diff
@@ -9,11 +9,11 @@
 using Microsoft.SemanticKernel.TemplateEngine;
 using Microsoft.SemanticKernel.TemplateEngine.Blocks;
 using Moq;
-using SemanticKernel.UnitTests.XunitHelpers;
+using SemanticKernel.Extensions.UnitTests.XunitHelpers;
 using Xunit;
 using Xunit.Abstractions;
 
-namespace SemanticKernel.UnitTests.TemplateEngine;
+namespace SemanticKernel.Extensions.UnitTests.TemplateEngine;
```
lemillermicrosoft, line 16: Should we add some tests for the new interface method, too?

dotnet/src/SemanticKernel.Abstractions/TemplateEngine/IPromptTemplateEngine.cs
markwallace-microsoft: I'm removing this method completely

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/PromptTemplateEngineTests.cs
markwallace-microsoft: I have removed the method

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -221,4 +230,71 @@ public KernelBuilder Configure(Action<KernelConfig> configure)
...
+        }
+
+        return null;
+    }
+}
+
+/// <summary>
+/// No-operation IPromptTemplateEngine which performs no rendering of the template.
+/// </summary>
+internal class NoopPromptTemplateEngine : IPromptTemplateEngine
```
SergeyMenshykh, line 298: It's usually a good practice to have a file per class or interface, so consider to move this class declaration to a separete file.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel/KernelBuilder.cs
RogerBarreto: Should we consider calling it `NullPromptTemplateEngine` ?

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -221,4 +230,75 @@ public KernelBuilder Configure(Action<KernelConfig> configure)
...
+            // Unable to load the Microsoft.SemanticKernel.TemplateEngine assembly
+        }
+        catch (InvalidOperationException)
+        {
+            // Assembly does not contain typed named PromptTemplateEngine
+        }
+        catch (Exception ex) when (!ex.IsCriticalException())
+        {
+            // Something unexpected by not critical
+        }
```
RogerBarreto, line 289: If no catching logic is being done, suggest to use `catch` without Exception type. 

```suggestion
        catch 
        {
            return null;
        }
```

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -221,4 +230,75 @@ public KernelBuilder Configure(Action<KernelConfig> configure)
...
+    /// <summary>
+    /// Create a default prompt template engine.
+    ///
+    /// This is a temporary solution to avoid breaking existing clients.
+    /// There will be a separate task to add support for registering instances of IPromptTemplateEngine and obsoleting the current approach.
+    ///
+    /// </summary>
+    /// <param name="loggerFactory">Logger factory to be used by the template engine</param>
+    /// <returns></returns>
+    private IPromptTemplateEngine CreateDefaultPromptTemplateEngine(ILoggerFactory? loggerFactory = null)
```
RogerBarreto, line 243: Consider moving this logic to a `NullTemplateEngine` static `Instance`, and call it from the Builder `this._promptTemplateEngine ?? NullTemplateEngine.Instance`

dotnet/src/SemanticKernel/SemanticFunctions/PromptTemplate.cs
```diff
@@ -64,16 +63,6 @@ public IList<ParameterView> GetParameters()
...
-        // Parameters from the template
-        foreach (var block in this._templateEngine.ExtractBlocks(this._template))
-        {
-            string? blockName = (block as VarBlock)?.Name;
-            if (!string.IsNullOrEmpty(blockName) && !result.ContainsKey(blockName!))
-            {
-                result.Add(blockName!, new ParameterView(blockName!));
-            }
-        }
-
```
RogerBarreto, line 76: Removing this logic can bring some strong consequences.

Shouldn't we have a `GetParameters()` in the `_templateEngine` `IPromptTemplateEngine` interface? 

So this call would be something like:

```
foreach (var p in this._templateEngine.GetParameters())
{
     result[p.Name] = new ParameterView(p.Name, p.Description, p.DefaultValue);
}
```

dotnet/src/SemanticKernel/KernelBuilder.cs
dmytrostruk: +1 for `NullPromptTemplateEngine` :)

dotnet/SK-dotnet.sln
```diff
@@ -152,6 +152,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ApplicationInsightsExample"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Connectors.Memory.Kusto", "src\Connectors\Connectors.Memory.Kusto\Connectors.Memory.Kusto.csproj", "{E07608CC-D710-4655-BB9E-D22CF3CDD193}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TemplateEngine", "src\Extensions\TemplateEngine\TemplateEngine.csproj", "{10E4B697-D4E8-468D-872D-49670FD150FB}"
```
dmytrostruk, line 155: Have you considered naming like `SemanticKernel.TemplateEngine` or `Extensions.TemplateEngine`? 
Looking at solution structure now, and from the name it feels like this project should belong to some logical group:
image

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -25,6 +31,9 @@ public sealed class KernelBuilder
     private IPromptTemplateEngine? _promptTemplateEngine;
     private readonly AIServiceCollection _aiServices = new();
 
+    private static bool _promptTemplateEngineInit = false;
```
dmytrostruk, line 34: I suggest using more detailed naming for better understanding, for example `_templateEngineInitialized`.

dotnet/src/SemanticKernel/KernelBuilder.cs
dmytrostruk: I agree, we should not keep `PromptTemplateEngine`-related logic in `KernelBuilder`. Regarding `this._promptTemplateEngine ?? NullTemplateEngine.Instance`, we can put this logic in place where template engine is actually used, in similar way how we do this with logging.

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -221,4 +230,75 @@ public KernelBuilder Configure(Action<KernelConfig> configure)
...
+        }
+
+        if (_promptTemplateEngineType is not null)
+        {
+            var constructor = _promptTemplateEngineType.GetConstructor(new Type[] { typeof(ILoggerFactory) });
+            if (constructor is not null)
+            {
+#pragma warning disable CS8601 // Null logger factory is OK
+                return (IPromptTemplateEngine)constructor.Invoke(new object[] { loggerFactory });
+            }
```
dmytrostruk, line 258: If this warning suppression is only for one line, then we should restore warning after that line. Otherwise warning suppression will be applied to the rest of the file.
```suggestion
#pragma warning disable CS8601 // Null logger factory is OK
                return (IPromptTemplateEngine)constructor.Invoke(new object[] { loggerFactory });
#pragma warning restore CS8601
            }
```

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -221,4 +230,75 @@ public KernelBuilder Configure(Action<KernelConfig> configure)
...
+
+    /// <summary>
+    /// Create a default prompt template engine.
+    ///
+    /// This is a temporary solution to avoid breaking existing clients.
+    /// There will be a separate task to add support for registering instances of IPromptTemplateEngine and obsoleting the current approach.
+    ///
+    /// </summary>
+    /// <param name="loggerFactory">Logger factory to be used by the template engine</param>
+    /// <returns></returns>
```
dmytrostruk, line 242:
```suggestion
    /// <returns>Instance of <see cref="IPromptTemplateEngine"/>.</returns>
```

or

```suggestion
```

dotnet/src/SemanticKernel/KernelBuilder.cs
markwallace-microsoft: This is deliberate. The goal with this PR is just to move the code and not cause a breaking change. We need to decide on a pattern for how to remove the logic from the `KernelBuilder` for the template engine (and also memory) but that is out of scope for this PR.

dotnet/src/SemanticKernel/SemanticFunctions/PromptTemplate.cs
markwallace-microsoft: I made that change in an earlier commit and when I reviewed with Shawn the decision was made that we would remove this code if it doesn't cause any regressions. This code is a bit hacky and likely cannot be easily support by other template engines. So in the interest of keeping things simple we are removing it.

dotnet/SK-dotnet.sln
markwallace-microsoft: This naming follows our current convention and changing it will be a breaking change because the namespace would also need to change. Shawn has asked me to check with @lemillermicrosoft on naming so I'm waiting to see his response.

dotnet/src/SemanticKernel/KernelBuilder.cs
dmytrostruk: You can create separate class `DefaultPromptTemplateEngine` temporarily, place it in separate file and move all the logic there. You will have to do the same for `NoopPromptTemplateEngine` in order not to keep two classes in same file.

dotnet/SK-dotnet.sln
dmytrostruk: In this case, I think the name of the project should be `SemanticKernel.TemplateEngine` in order to follow naming convention and keep the namespace. But it's not critical.

markwallace-microsoft: Yeah, that introduces an inconsistency with the other extension projects e.g.,

AssemblyName: `Microsoft.SemanticKernel.Planning.ActionPlanner`
RootNamespace: `Microsoft.SemanticKernel.Planning.Action`
Project name: `Planning.ActionPlanner`

For template engine we have RootNamespace: `Microsoft.SemanticKernel.TemplateEngine` which we don't want to change. So following the above pattern the Project name should be `TemplateEngine`.

But I'll happily defer to whatever @lemillermicrosoft suggests.

lemillermicrosoft: Maybe this:

ProjectName: `TemplateEngine.PromptTemplateEngine`

RootNamespace: `Microsoft.SemanticKernel.TemplateEngine.Prompt`

AssemblyName: `Microsoft.SemanticKernel.TemplateEngine.PromptTemplateEngine`

lemillermicrosoft: And even with that namespace, the `PromptTemplateEngine` class could still be declared in `Microsoft.SemanticKernel.TemplateEngine` -- the other classes could be in `...Prompt` namespace though.

RogerBarreto (approved):
CLG

dotnet/SK-dotnet.sln
lemillermicrosoft: Also would be fine leaving as-is and then doing that breakdown for future.

lemillermicrosoft (approved):

dotnet/SK-dotnet.sln
markwallace-microsoft: @lemillermicrosoft I've pushed the naming changes also. I created two commits, with the second one just changing the namespace for the `PromptTemplateEngine` class. This change impacts samples and could impact clients but I would expect most people rely on the fact that the `PromptTemplateEngine` class is loaded by default.

dotnet/src/SemanticKernel.Abstractions/TemplateEngine/IPromptTemplateEngine.cs
```diff
@@ -13,16 +11,6 @@ namespace Microsoft.SemanticKernel.TemplateEngine;
 /// </summary>
 public interface IPromptTemplateEngine
 {
-    /// <summary>
-    /// Given a prompt template string, extract all the blocks (text, variables, function calls)
-    /// </summary>
-    /// <param name="templateText">Prompt template (see skprompt.txt files)</param>
-    /// <param name="validate">Whether to validate the blocks syntax, or just return the blocks found, which could contain invalid code</param>
-    /// <returns>A list of all the blocks, ie the template tokenized in text, variables and function calls</returns>
-    IList<Block> ExtractBlocks(
```
hario90, line 22: Can we keep this? It could be useful on the extension side for highlighting user and SDK/AI provided values in the prompt preview. Determining these values using a diff didn't always provide an accurate result and I think we shouldn't need to re-calculate this again on the extension side.

hario90 (changes requested):
Everything looks good but I'd like to see if we can keep the ExtractBlocks method because this would be helpful on the extension side for prompt highlighting.

markwallace-microsoft:
> Everything looks good but I'd like to see if we can keep the ExtractBlocks method because this would be helpful on the extension side for prompt highlighting.

Hi @hario90 
we are taking this out of the API signature because it's very tied to our template engine implementation and other template engines e.g. Handlebars may not be able to support this.

dotnet/src/SemanticKernel/SemanticFunctions/PromptTemplate.cs
lemillermicrosoft: #3115

Issue #2592, opened 2023-08-27
Python: Implement skill for python REPL


joowon-dm-snu:
As shown by the code interpreter of GPT-4, defining a problem in code through the concept of a Planner or Agent and utilizing the execution results in connection with LLM can help to handle more issues easily and accurately, especially in tasks related to numbers.

It would be great if a python REPL was added to the CoreSkill so that the Planner can utilize it.
The addition of this skill can be better utilized when linked with #2385.

joowon-dm-snu:
Although some code overlaps, the reason I opened it again as a different PR is mentioned in #2593.

nacharya1:
@joowon-dm-snu thanks for creating the issue to align to your PR.
cc: @awharrison-28

PR #2385, opened 2023-08-09
Python: Query CSV and pandas dataframes

am831:
### Motivation and Context

Implementation of two new skills called `data_skill` and `code_skill` that allow the user to query structured data using natural language. This provides an approach to querying data that is conversational, as opposed to writing code or using SQL queries to work with data.
I am working with @sneha-afk on this.
Closes #1968 

### Description

1. Implemented `code_skill` class which sends the user's prompt to the LLM along with instructions to generate python code to answer the question, it then executes the code
2. Implemented `data_skill` class that first converts all CSV data to pandas dataframes. It has a function called `get_df_data` that creates a prompt with the first 3 rows of every dataframe. And another function called `query_async` that receives the user's ask and returns the answer. It sends a prompt to the `code_skill` class that contains the first 3 rows of each dataframe along with instructions to write python code and the user's ask, `code_skill` then executes LLM generated code and `query_async` sends the answer back in a natural language format.
3. `transform_async` will allow the user to ask to transform data and receive the new dataframe
4.  One other function is prototyped for `data_skill`. `plot_async` will allow the user to ask for bar charts and line graphs from matplotlib. We would like high level feedback on this idea, do we want the user to be able to generate plots?
5. There are two new example files demonstrating the usage of both skills

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #2593, opened 2023-08-27
Python: add python repl skill

joowon-dm-snu:
### Motivation and Context

implement #2592 

As mentioned in #2385, I believe that the `custom_execute_async` in the `CodeSkill` is a necessary feature. 
However, the PR has been stagnant for a long time, and it seems a bit complicated mixed with the Chat feature.

### Description

The python REPL execution part is entirely separated as a Native function, and the stdout generated by the print command is optionally returned. 

For security, considering a separate thread and strict management of global and local variables can be taken into account. However, it's currently left as a TODO since it's complex to manage and time-consuming. 
Also, it seems much better to input the include_stdout variable directly rather than using the context.

### Contribution Checklist


awharrison-28:
@nacharya1 the comment from #2385 applies to this as well

PR #2638, opened 2023-08-29
.Net: Implementing cosine similarity in duckdb engine

colombod:
### Motivation and Context

This pr moves cosine similarity calculation into duckDb engine to improve performance and memory usage.

### Description

Moving cosine similarity calculation on the server there is no need to marshal out more data than the one scoring enough and meeting the result limit.

### Contribution Checklist


dotnet/src/Connectors/Connectors.Memory.DuckDB/DuckDBMemoryStore.cs
```diff
@@ -152,22 +150,16 @@ public async Task RemoveBatchAsync(string collectionName, IEnumerable<string> ke
             yield break;
         }
 
-        var collectionMemories = new List<MemoryRecord>();
         TopNCollection<MemoryRecord> embeddings = new(limit);
 
-        await foreach (var record in this.GetAllAsync(collectionName, cancellationToken))
+        await foreach (var dbEntry in this._dbConnector.GetNearestMatchesAsync(this._dbConnection, collectionName, embedding.ToArray(), limit, minRelevanceScore, cancellationToken))
```
lemillermicrosoft, line 155: I think you can delete `GetAllAsync` now too, right?

dotnet/src/Connectors/Connectors.Memory.DuckDB/Database.cs
lemillermicrosoft, line 1: Same here, I think `ReadAllAsync` can be removed.

lemillermicrosoft: And `DeleteEmptyAsync`

lemillermicrosoft (approved):
LGTM -- a few methods I think can be removed though. Thanks @colombod!

dotnet/src/Connectors/Connectors.Memory.DuckDB/DuckDBMemoryStore.cs
colombod: let me look!

dotnet/src/Connectors/Connectors.Memory.DuckDB/Database.cs
colombod: done! thanks for pointing that out

dotnet/src/Connectors/Connectors.Memory.DuckDB/Database.cs
```diff
@@ -98,14 +117,23 @@ public async Task CreateCollectionAsync(DuckDBConnection conn, string collection
         }
     }
 
-    public async IAsyncEnumerable<DatabaseEntry> ReadAllAsync(DuckDBConnection conn,
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Security", "CA2100:Review SQL queries for security vulnerabilities", Justification = "Internal method sserializing array of float and numbers")]
```
dmytrostruk, line 120:
```suggestion
    [System.Diagnostics.CodeAnalysis.SuppressMessage("Security", "CA2100:Review SQL queries for security vulnerabilities", Justification = "Internal method serializing array of float and numbers")]
```

dmytrostruk (approved):
@colombod Thank you for contribution, having similarity calculation on server side is a huge improvement. 

Idea for future improvements - currently we save all collections and records in one table. I'm wondering if it will be possible to create new table per collection. From DB architecture and data isolation perspective, I think this is the right approach, but would be great to know your opinion on this.

colombod:
> @colombod Thank you for contribution, having similarity calculation on server side is a huge improvement.
> 
> Idea for future improvements - currently we save all collections and records in one table. I'm wondering if it will be possible to create new table per collection. From DB architecture and data isolation perspective, I think this is the right approach, but would be great to know your opinion on this.

Very good point. I kept it this way because, with polyglot notebook, it is possible to send commands to the duckDB and manage persistency and loading from parquet files. From my understanding they deal with a single table.

lemillermicrosoft (approved):

PR #2656, opened 2023-08-30
.Net: Extract Retry Implementation

### LINKED ISSUE #2271: Obsolete the public "WithRetry" and internal Retry logic implementation

RogerBarreto:
### Motivation and Context

Reduce the responsibility surface of Semantic Kernel regarding Resiliency implementations, keeping only the abstractions that can be delegated and handling by well know resiliency libraries and custom implementations.

Extract existing logic from Kernel Core to an extension project that can handle retry logic similar to the prior using Polly Policies.

Resolves #2271
Closes #2271 

⚠️ Small breaking changes.

### Description

This change removes and deprecates the usage of Retry implementations in our Resiliency namespace with `NullHttpHandlerFactory` implementations.

### Contribution Checklist


dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandlerFactory.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Net.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.SemanticKernel.Config;
+using Microsoft.SemanticKernel.Diagnostics;
+
+namespace Microsoft.SemanticKernel.Reliability.Polly.Config;
```
lemillermicrosoft, line 8: Do we need all three factories in a single package? I was imagining a Reliability.Default and a Reliability.Polly where Reliability.Default would be part of the Meta kernel package with the existing implementation. It's also odd that this is in the Config namespace and while this leverages HttpRetryConfig, none of the others do.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    internal DefaultHttpRetryHandler(HttpRetryConfig? config, ITimeProvider? timeProvider, ILoggerFactory? loggerFactory)
+    {
+        this._logger = loggerFactory?.CreateLogger<DefaultHttpRetryHandler>();
+        this._config = config ?? new();
+        this._policy = this.GetPolicy();
+        this._timeProvider = timeProvider ?? new DefaultTimeProvider();
+    }
+
+    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
```
SergeyMenshykh, line 34: Several members of this class, as well as others, don't have XML comments.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+
+        return await this._policy.ExecuteAsync(async () =>
```
SergeyMenshykh, line 38: Consider passing cancellation token as an argument to the `ExecuteAsync` method.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
```diff
@@ -0,0 +1,135 @@
...
+        var timeToWait = retryAfter > this._config.MaxRetryDelay
+            ? this._config.MaxRetryDelay
+            : retryAfter < this._config.MinRetryDelay
+                ? this._config.MinRetryDelay
+                : retryAfter ?? default;
+
+        // If exponential backoff is enabled, double the delay for each retry
+        if (this._config.UseExponentialBackoff)
+        {
+            timeToWait = TimeSpan.FromTicks(timeToWait.Ticks * 2);
```
SergeyMenshykh, line 114: 1. This line only doubles the wait time instead of the intended exponential increase, which can be achieved using the `retryCount` parameter.

2. I'm not sure if we should exponentially increase the wait time for responses with `RetryAfter.Date` or `RetryAfter.Delta` headers. It makes sense for responses without a specified wait time, but if the server already provides a retry instruction, why increase it exponentially?

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
```diff
@@ -0,0 +1,135 @@
...
+        DateTimeOffset GetCurrentTime();
+    }
+
+    internal sealed class DefaultTimeProvider : ITimeProvider
+    {
+        public DateTimeOffset GetCurrentTime()
+        {
+            return DateTimeOffset.UtcNow;
+        }
+    }
```
SergeyMenshykh, line 134: Consider moving the TimeProvider interface and class to the abstractions project so it can be used by other components.

RogerBarreto: I think this deserves a dedicated PR. As TimeProvider as an abstraction would only make sense if used by more than one component. 

I can be wrong but currently I don't see any other Kernel component that uses the Clock.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandlerFactory.cs
RogerBarreto: Good point, we don't I changed all the others to be Internal and only be usable thru the Extension method.

Added a public handler that takes a Polly Policy for more customized scenarios.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
RogerBarreto: 1. Agree, this was the current default implementation in the Core Kernel. Changed.

2. You are right, good catch, I left this part with the original implementation.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
RogerBarreto: Thanks, added missing comments.

dotnet/src/Extensions/Extensions.UnitTests/Reliability/Polly/DefaultHttpRetryHandlerTests.cs
lemillermicrosoft, line 1: I see a `Reliability/Polly` and `Reliability.Polly` -- both intended?

dotnet/src/Extensions/Reliability.Polly/HttpRetryConfig.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Net;
+using System.Net.Http;
+using Microsoft.SemanticKernel.Diagnostics;
+
+namespace Microsoft.SemanticKernel.Reliability.Polly;
+
+/// <summary>
+/// Retry configuration for DefaultKernelRetryHandler that uses RetryAfter header when present.
+/// </summary>
+public sealed record HttpRetryConfig
```
lemillermicrosoft, line 14: It feels very odd to use this in conjunction with Polly when it already has its own configuration mechanisms. What is the goal here with retry support?

RogerBarreto: To not break current usage of Kernel Configuration file, (intermediate translation to Polly policies).

We might Obsolete this in the long run.

dotnet/samples/KernelSyntaxExamples/Example42_KernelBuilder.cs
```diff
@@ -153,15 +154,44 @@ public static Task RunAsync()
...
 
         var kernel9 = Kernel.Builder
-            .Configure(c => c.SetHttpRetryHandlerFactory(new NullHttpRetryHandlerFactory()))
+            .Configure(c => c.SetHttpHandlerFactory(new NullHttpHandlerFactory()))
             .Build();
 
-        var kernel10 = Kernel.Builder.WithRetryHandlerFactory(new RetryThreeTimesFactory()).Build();
+        var kernel10 = Kernel.Builder.WithHttpHandlerFactory(new RetryThreeTimesFactory()).Build();
+
+        var kernel11 = Kernel.Builder.WithHttpHandlerFactory(new PollyRetryThreeTimesFactory()).Build();
```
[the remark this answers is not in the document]
RogerBarreto: Good suggestion, added that factory + changed the example to use it.

dotnet/src/Extensions/Reliability.Polly/HttpRetryConfig.cs
lemillermicrosoft: But today kernel configuration is used only for the DefaultHttpRetryPolicy -- never related to Polly implementations of things. I really was hoping to keep the implementation, just in its own package, and not be related to Polly.

dotnet/samples/KernelSyntaxExamples/Example08_RetryHandler.cs
lemillermicrosoft, line 1: Depending on the final result, we should simplify this example. I think we should focus on:

- Default behavior, no retry
- Using Reliability.* extensions
- Using a custom handler that retries once for a given error.

dotnet/src/Extensions/Reliability.Polly/KernelConfigExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+/// </summary>
+public static class KernelConfigExtensions
+{
+    /// <summary>
+    /// Sets the default retry configuration for any kernel http request.
+    /// </summary>
+    /// <param name="kernelConfig">Target instance</param>
+    /// <param name="retryConfig">Retry configuration</param>
+    /// <returns>Returns target instance for fluent compatibility</returns>
+    public static KernelConfig SetHttpRetryConfig(this KernelConfig kernelConfig, HttpRetryConfig? retryConfig = null)
```
lemillermicrosoft, line 18: I think we should only support HttpRetryConfig in our own non-polly extension library. Reliability.Polly should be helper utilities for taking a Polly policy and supplying it as part of an IDelegatingHandlerFactory

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
lemillermicrosoft: > This line only doubles the wait time instead of the intended exponential increase

How would it be different? My understanding is that exponential increase is when you take the original delay and multiply by a factor of x>=2. When initial wait is 1 second, you get subsequent waits of 2, 4, 8, 16, 32. This is exponential growth.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
lemillermicrosoft: These were added to improve testability.

dotnet/src/Extensions/Reliability.Polly/Config/DefaultHttpRetryHandler.cs
SergeyMenshykh: > > This line only doubles the wait time instead of the intended exponential increase
> 
> How would it be different? My understanding is that exponential increase is when you take the original delay and multiply by a factor of x>=2. When initial wait is 1 second, you get subsequent waits of 2, 4, 8, 16, 32. This is exponential growth.

Yep, your understanding is correct; the subsequent waits should be 2, 4, 8, 16, 32 for exponential retry. However, with the current implementation where the `retryCount` parameter is not used, the waits will remain at 2, 2, 2, 2, 2, ... I think the line should use the `retryCount` to achieve that - `timeToWait = TimeSpan.FromTicks(timeToWait.Ticks * Math.Pow(2, retryCount));`

dotnet/src/Extensions/Reliability.Polly/KernelHttpRetryHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        // If the retry delay is longer than the max retry delay, use the max retry delay
+        var timeToWait = retryAfter > this._config.MaxRetryDelay
+            ? this._config.MaxRetryDelay
+            : retryAfter < this._config.MinRetryDelay
+                ? this._config.MinRetryDelay
+                : retryAfter ?? default;
+
+        // If exponential backoff is enabled, and the server didn't provide a RetryAfter header, double the delay for each retry
+        if (this._config.UseExponentialBackoff && response?.Headers.RetryAfter?.Date.HasValue == false)
```
SergeyMenshykh, line 128: There could be cases when server returns delata instead of time.

SergeyMenshykh (approved):

dotnet/src/Extensions/Reliability.Basic/KernelConfigExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </summary>
+public static class KernelConfigExtensions
+{
+    /// <summary>
+    /// Sets the default retry configuration for any kernel http request.
+    /// </summary>
+    /// <param name="kernelConfig">Target instance</param>
+    /// <param name="retryConfig">Retry configuration</param>
+    /// <returns>Returns target instance for fluent compatibility</returns>
+    public static KernelConfig SetRetryBasic(this KernelConfig kernelConfig, BasicRetryConfig? retryConfig = null)
```
lemillermicrosoft, line 20: Since this is the last thing in KernelConfig, I think we should kill it completely and add extensions for KernelBuilder.WithBasicRetryHandler(BasicRetryConfig config)

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -51,7 +51,7 @@ public IKernel Build()
     {
         if (this._httpHandlerFactory != null)
         {
-            this._config.SetHttpRetryHandlerFactory(this._httpHandlerFactory);
+            this._config.SetHttpHandlerFactory(this._httpHandlerFactory);
```
lemillermicrosoft, line 54: related to other comments, let's just expose the httpHandlerFactory

dotnet/src/SemanticKernel.Abstractions/Config/HttpHandlerFactory{THandler}.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Net.Http;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.SemanticKernel.Config;
+
+public abstract class HttpHandlerFactory<THandler> : IDelegatingHandlerFactory where THandler : DelegatingHandler
```
lemillermicrosoft, line 9: Maybe this is all better in Microsoft.SemanticKernel.Http or similar?

lemillermicrosoft (approved):
High level LGTM. I think we might want to re-think `Config` and those namespaces, but could also be fine doing that later as a follow-up (might be some breaking change noise to handle but also not too worried about it). Thanks a bunch Roger for the time and effort and patience on this one.

dotnet/src/Extensions/Reliability.Basic/Reliability.Basic.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
```
lemillermicrosoft, line 2: I'd also vote to add this to the meta package.

dotnet/src/SemanticKernel.Abstractions/Config/HttpHandlerFactory{THandler}.cs
RogerBarreto: Agree, moved to Http namespace

SergeyMenshykh (approved):

PR #2657, opened 2023-08-30
.Net: Add missing ModelResult abstraction in IChatResult interface

### LINKED ISSUE #1565: Hacky workaround to get token usage from chat completions api response
I am trying to get the token usage data from the response of a call to the OpenAI's chat completion api but it doesn't seem possible. Here are the details:

`IChatCompletion.GetChatCompletionsAsync` returns a list of `IChatResult` objects. It represents the response from OpenAI's chat completion API. The issue is that this interface does not have the `ModelResult` property which contains token usage data. Hence, there's no way to get the token usage data when using the chat completion api. In contrast, the `ITextCompletion.GetCompletionsAsync` returns a list of `ITextResult` objects which do contain the `ModelResult` property. 

Looking deeper into the code, it looks like the concrete object type of `IChatResult` which is the internal class `ChatResult` has the `ModelResult` property. So then I was able to cast the `IChatResult` to a `ITextResult` and access that property:

Screenshot 2023-06-19 174511

There should be a simple fix for this.

RogerBarreto:
### Motivation and Context

`IChatResult` interface misses the ModelResult abstraction

Resolves #1565
Closes #1565

### Description

Added missing abstraction.

### Contribution Checklist

### DIFF
--- a/dotnet/samples/KernelSyntaxExamples/Example34_CustomChatModel.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example34_CustomChatModel.cs
@@ -7,6 +7,7 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.SemanticKernel.AI.ChatCompletion;
+using Microsoft.SemanticKernel.Orchestration;
 
 /**
  * The following example shows how to plug use a custom chat model.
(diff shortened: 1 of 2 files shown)


markwallace-microsoft (approved):
LGTM

dmytrostruk (approved):

PR #2671, opened 2023-08-31
.Net: Adding chat system prompt to config file schema

### LINKED ISSUE #644: Is there any way to define the system/user/assistant messages in the prompt file with the `OpenAIChatCompletion`, but not call the `add_system_message`/`add_user_message`/`add_assistant_message` in code?
Is there any way to define the system/user/assistant messages in the prompt file with the `OpenAIChatCompletion`, but not call the `add_system_message`/`add_user_message`/`add_assistant_message` in code?

### LINKED ISSUE #2068: How to change system message when use skprompt.txt with Chat Completion API
When use semantic function which from skprompt.txt file, how to change system message?

I noticed that the default system message always is: You are an AI assistant that helps people find information.

RogerBarreto:
### Motivation and Context

Adding support to set the chat system message using `config.json` with Semantic Functions.

```
{
  ...
  "completion": {
    ...
    "chat_system_prompt": "My custom system chat prompt"
  }
}
```

Resolves #644 
Closes #644 
Resolves #2068 
Closes #2068 

### Description

Allows configuration files to specify what will be the system chat message used while requesting from LLMs.

### Contribution Checklist


lemillermicrosoft:
We should make sure this is done in Python as well (either open a different issue or don't close the linked upon completion.

lemillermicrosoft:
Should we also add support to `CreateSemanticFunction` extensions?

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/CompleteRequestSettings.cs
```diff
@@ -81,5 +81,12 @@ public static CompleteRequestSettings FromCompletionConfig(PromptTemplateConfig.
             MaxTokens = config.MaxTokens,
             StopSequences = config.StopSequences,
         };
+
+        if (config.ChatSystemPrompt is not null)
```
SergeyMenshykh, line 85: What if the `ChatSystemPrompt` property is empty? should it be added?

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/CompleteRequestSettings.cs
RogerBarreto: Good point, going to change it to `string.IsNullOrWhitespace()`

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -125,6 +126,34 @@ public void ItImportsSkillsNotCaseSensitive()
         Assert.True(skill.ContainsKey("GETANYVALUE"));
     }
 
+    [Theory]
+    [InlineData(null, "Assistant is a large language model.")]
```
dmytrostruk, line 130: Where did this `Assistant is a large language model.` message come from?

RogerBarreto: https://github.com/microsoft/semantic-kernel/blob/c8a7c6d134fff68e59dd8cc191e9a2fdc1d0937a/dotnet/src/SemanticKernel.Abstractions/AI/TextCompletion/CompleteRequestSettings.cs#L61

dmytrostruk (approved):

pablontiv:
Is this already available in Semantic Kernel Tools?

sandeepvootoori:
Is there a reason why this was implemented in the config.json? putting it in config.json makes it difficult to hydrate the system message with grounding data while using planners. Any plans in the future to make this part of skprompt.txt?

@thegovind

PR #2674, opened 2023-08-31
.Net: Use Cosine Distance for Azure Cognitive Search

dluc:
When using vector search with `VectorMetricType.Cosine`, Azure Cognitive Search search result score is expressed as Cosine Distance. The value `minRelevanceScore` passed to `IMemoryStore` is expressed as Cosine Similarity, so it's important to transform the values during the comparison and when returning results.

Formulas:

```csharp
double ScoreToCosineSimilarity(double score)
{
    return 2 - 1 / score;
}

double CosineSimilarityToScore(double similarity)
{
    return 1 / (2 - similarity);
}
```

### DIFF
--- a/dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
@@ -65,8 +65,8 @@ public class AzureCognitiveSearchMemoryStore : IMemoryStore
 
         return await indexes
             .AnyAsync(index =>
-                string.Equals(index, collectionName, StringComparison.OrdinalIgnoreCase) ||
-                string.Equals(index, normalizedIndexName, StringComparison.OrdinalIgnoreCase),
+                    string.Equals(index, collectionName, StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(index, normalizedIndexName, StringComparison.OrdinalIgnoreCase),
                 cancellationToken: cancellationToken
             )
             .ConfigureAwait(false);
@@ -169,6 +169,9 @@ public class AzureCognitiveSearchMemoryStore : IMemoryStore
         bool withEmbeddings = false,
         [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
+        // Cosine similarity range: -1 .. +1
+        minRelevanceScore = Math.Max(-1, Math.Min(1, minRelevanceScore));
+
         var normalizedIndexName = this.NormalizeIndexName(collectionName);
 
         var client = this.GetSearchClient(normalizedIndexName);
@@ -431,7 +434,9 @@ public class AzureCognitiveSearchMemoryStore : IMemoryStore
 
     private static double ScoreToCosineSimilarity(double score)
     {
-        return 1 - (1 - score) / score;
+        // Azure Cognitive Search score formula. The min value is 0.333 for cosine similarity -1.
+        score = Math.Max(score, 1.0 / 3);
+        return 2 - 1 / score;
     }
 
     private static double CosineSimilarityToScore(double similarity)
--- a/dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
@@ -198,13 +198,14 @@ public class AzureCognitiveSearchMemoryStore : IMemoryStore
 
         if (searchResult == null) { yield break; }
 
+        var minAzureSearchScore = CosineSimilarityToScore(minRelevanceScore);
         await foreach (SearchResult<AzureCognitiveSearchMemoryRecord>? doc in searchResult.Value.GetResultsAsync())
         {
-            if (doc == null || doc.Score < minRelevanceScore) { continue; }
+            if (doc == null || doc.Score < minAzureSearchScore) { continue; }
 
             MemoryRecord memoryRecord = doc.Document.ToMemoryRecord(withEmbeddings);
 
-            yield return (memoryRecord, doc.Score ?? 0);
+            yield return (memoryRecord, ScoreToCosineSimilarity(doc.Score ?? 0));
         }
     }
 
@@ -426,6 +427,16 @@ public class AzureCognitiveSearchMemoryStore : IMemoryStore
         {
             throw e.ToHttpOperationException();
         }
+    }
+
+    private static double ScoreToCosineSimilarity(double score)
+    {
+        return 1 - (1 - score) / score;
+    }
+
+    private static double CosineSimilarityToScore(double similarity)
+    {
+        return 1 / (2 - similarity);
     }
 
     #endregion
--- a/dotnet/src/SemanticKernel.Abstractions/Memory/IMemoryStore.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Memory/IMemoryStore.cs
@@ -106,7 +106,7 @@ public interface IMemoryStore
     /// <param name="collectionName">The name associated with a collection of embeddings.</param>
     /// <param name="embedding">The <see cref="Embedding{Single}"/> to compare the collection's embeddings with.</param>
     /// <param name="limit">The maximum number of similarity results to return.</param>
-    /// <param name="minRelevanceScore">The minimum relevance threshold for returned results.</param>
+    /// <param name="minRelevanceScore">The minimum cosine similarity threshold for returned results.</param>
     /// <param name="withEmbeddings">If true, the embeddings will be returned in the memory records.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>A group of tuples where item1 is a <see cref="MemoryRecord"/> and item2 is its similarity score as a <see cref="double"/>.</returns>
@@ -123,7 +123,7 @@ public interface IMemoryStore
     /// </summary>
     /// <param name="collectionName">The name associated with a collection of embeddings.</param>
     /// <param name="embedding">The <see cref="Embedding{Single}"/> to compare the collection's embeddings with.</param>
-    /// <param name="minRelevanceScore">The minimum relevance threshold for returned results.</param>
+    /// <param name="minRelevanceScore">The minimum cosine similarity threshold for returned results.</param>
     /// <param name="withEmbedding">If true, the embedding will be returned in the memory record.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>A tuple consisting of the <see cref="MemoryRecord"/> and the similarity score as a <see cref="double"/>. Null if no nearest match found.</returns>


dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -428,5 +429,15 @@ private static T RunMemoryStoreOperation<T>(Func<T> operation)
         }
     }
 
+    private static double DistanceToSimilarity(double distance)
+    {
+        return 1 - (1 - distance) / distance;
```
glahaye, line 434: I'm not sure if this can be an issue, but if distance is 0, the division will explode...

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -428,5 +429,15 @@ private static T RunMemoryStoreOperation<T>(Func<T> operation)
...
+    }
+
+    private static double SimilarityToDistance(double similarity)
+    {
+        return 1 / (2 - similarity);
+    }
```
glahaye, line 440: I'm not familiar with the formulas but, assuming possible similarity values in the range from 0 to 1, the possible distances given by SimilarityToDistance() range between 0.5 to 1.

Also, if distance ranges from 0 to 1 (does it?), then the similarities given by DistanceToSimilarity() range from 1 to infinity.  However, we know similarities range from 0 to 1, so this methid is not correct!

Isn't the relationship between distance and similarity simply CosineDistance = 1 - CosineSimilarity???

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: I've seen the same problem in all implementations of cosine similarity, when magnitude is 0 you get a division by zero.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: the `DistanceToSimilarity` to formula was shared by the Search team. This one is just the inverse of the other formula. I've seen other definitions of Distance, e.g. in StackOverflow, but these are specific to Azure Cognitive Search.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
momuno: assuming correctness, why not return
`return 2 - 1 / distance` ?  Fewer operations.  Not sure how the C# compiler works, but possibly this could result in fewer operations and reads of distance variable?

glahaye (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
momuno: Curious to know more.  We can chat in person.  The domain/range for these equations I'm getting is (-infinity, 2) for similarity, and (0, infinity) for distance.

glahaye (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: I renamed the function to "ScoreToSimilarity". When two vectors match, the score is 1, thus similarity 1

`1 - (1 - distance) / distance` => `1 - (1 - 1) / 1` => 1

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: happy to chat offline, I think here we're good to merge. 

This formula is just the inverse of the other one. 
**Similarity goes from 0 to 1**, hence:

min: `1 / (2 - 0)` => `0.5`
max: `1 / (2 - 1)` => `1.0`

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -198,13 +198,14 @@ public async IAsyncEnumerable<string> UpsertBatchAsync(string collectionName, IE
...
+        var minAzureSearchScore = CosineSimilarityToScore(minRelevanceScore);
         await foreach (SearchResult<AzureCognitiveSearchMemoryRecord>? doc in searchResult.Value.GetResultsAsync())
         {
-            if (doc == null || doc.Score < minRelevanceScore) { continue; }
+            if (doc == null || doc.Score < minAzureSearchScore) { continue; }
 
             MemoryRecord memoryRecord = doc.Document.ToMemoryRecord(withEmbeddings);
 
-            yield return (memoryRecord, doc.Score ?? 0);
+            yield return (memoryRecord, ScoreToCosineSimilarity(doc.Score ?? 0));
```
lemillermicrosoft, line 208:
```suggestion
            yield return (memoryRecord, ScoreToCosineSimilarity(doc.Score ?? 0.5));
```
Sounds like `score: [0.5,1.0]` so probably should use that here.

lemillermicrosoft (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: @momuno the operators precedence is different: `1 - (1 - distance) / distance` is `1 - ((1 - distance) / distance)` which could also be written as `(2 * distance -1) / distance`

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: actually the minimum is 1/3, I'll add the check to the function and leave it to 0 here

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -198,13 +198,14 @@ public async IAsyncEnumerable<string> UpsertBatchAsync(string collectionName, IE
...
+        var minAzureSearchScore = CosineSimilarityToScore(minRelevanceScore);
         await foreach (SearchResult<AzureCognitiveSearchMemoryRecord>? doc in searchResult.Value.GetResultsAsync())
         {
-            if (doc == null || doc.Score < minRelevanceScore) { continue; }
+            if (doc == null || doc.Score < minAzureSearchScore) { continue; }
 
             MemoryRecord memoryRecord = doc.Document.ToMemoryRecord(withEmbeddings);
 
-            yield return (memoryRecord, doc.Score ?? 0);
+            yield return (memoryRecord, ScoreToCosineSimilarity(doc.Score ?? 0.333333));
```
glahaye, line 208: Maybe have a constant like MaxAcsDistanceScore = 1 / 3; Since this is re-used below too.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -428,5 +429,16 @@ private static T RunMemoryStoreOperation<T>(Func<T> operation)
         }
     }
 
+    private static double ScoreToCosineSimilarity(double score)
```
glahaye, line 432: Again, I just hope this this right because the "similarities" returned by this method are always negative for input scores above 0!

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -428,5 +429,16 @@ private static T RunMemoryStoreOperation<T>(Func<T> operation)
...
+    {
+        score = Math.Max(score, 1.0 / 3);
+        return 1 - (1 - score) / score;
+    }
```
momuno, line 436: Not sure where to find prior conversation if this already was anwered...
Why use `return 1 - (1 - score) / score;` instead of `return 2 - 1 / score;`  ?

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
momuno: If there are restrictions on the input domain, why does the code not enforce that?

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
momuno: I see response in email that:
`1 - (1 - distance) / distance` is `1 - ((1 - distance) / distance) ` is `(2 * distance -1) / distance`  Agreed.  And this is `2 - (1 / distance)`  which is  `2 - 1 / distance`.

Why do we need to evaluate distance twice in the expression?  It can be evaluated once.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
momuno: Not sure why I couldn't see this thread earlier.  Please see my note below about furthering the simplification of the expression.

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: the value turns to positive after 0.5

for score = 0.3333: `1 - (1 - 0.3333) / 0.3333` => `-1`
for score = 0.4: `1 - (1 - 0.4) / 0.4` => `-0.5`
for score = 0.5: `1 - (1 - 0.5) / 0.5` => `0`
for score = 0.6: `1 - (1 - 0.6) / 0.6` => `0.3333`
for score = 0.7: `1 - (1 - 0.7) / 0.7` => `0.57`

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: pls see latest code, I logic is in the function

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
momuno: For score (D) to similarity (S) aka f(D) = S ....
The domain for D is [1/3, infinity).
This makes the range for S to be [-1, 2)

S is negative for D in [1/3, 1/2) , S is 0 for D at 1/2, and S is positive for D in (1/2, infinity)

glahaye: Oh, right!  Well, if you're happy with this, I'm happy too.

glahaye (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -428,5 +432,17 @@ private static T RunMemoryStoreOperation<T>(Func<T> operation)
         }
     }
 
+    private static double ScoreToCosineSimilarity(double score)
```
SergeyMenshykh, line 435: 1. Please add XML comment to the method and the one below.
2. Rename the method and the one below so that the start from verb - https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
```diff
@@ -169,6 +169,9 @@ public async IAsyncEnumerable<string> UpsertBatchAsync(string collectionName, IE
         bool withEmbeddings = false,
         [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
+        // Cosine similarity range: -1 .. +1
+        minRelevanceScore = Math.Max(-1, Math.Min(1, minRelevanceScore));
```
SergeyMenshykh, line 173: Move the line down closer to the place it's used to improve readability and decrease cognitive load.

SergeyMenshykh (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: this is validation logic, it's common practice checking/cleaning when entering a method before any other logic takes place

dotnet/src/Connectors/Connectors.Memory.AzureCognitiveSearch/AzureCognitiveSearchMemoryStore.cs
dluc: private methods are rarely commented and the name is self explanatory. About the verb, this is a transformation like `ToInt` `ToString`

PR #2688, opened 2023-09-01
.Net: Postgres memory store add simple constructor

JadynWong:
### Motivation and Context

The Postgres memory store adds a simple constructor using a connection string. The `NpgsqlDataSource` is managed internally by the postgres memory store.

### Description

- Similar to the recent changes to the redis memory store. which allows the user to pass the connection string directly. the `NpgsqlDataSource` is managed internally.
- Fixed postgres integration test (caused by #2419)

### Contribution Checklist


lemillermicrosoft (approved):

dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresMemoryStore.cs
```diff
@@ -1,4 +1,5 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+﻿
```
dmytrostruk, line 1: Empty line:
```suggestion
```

dotnet/src/IntegrationTests/Connectors/Memory/Postgres/PostgresMemoryStoreTests.cs
```diff
@@ -231,7 +231,7 @@ public async Task UpsertReplacesExistingRecordWithSameIdAsync()
         Assert.NotNull(actual);
         Assert.Equal(testRecord.Metadata.Id, key);
         Assert.Equal(testRecord2.Metadata.Id, actual.Key);
-        Assert.True(testRecord.Embedding.Span.SequenceEqual(actual.Embedding.Span));
+        Assert.False(testRecord.Embedding.Span.SequenceEqual(actual.Embedding.Span));
```
dmytrostruk, line 234: It's interesting that previously we checked that `testRecord` embeddings should be equal to `actual`. But it looks like they should be different based on test scenario, so this change is valid.

dmytrostruk (approved):
LGTM! Functionality verified with integration tests.

dotnet/src/IntegrationTests/Connectors/Memory/Postgres/PostgresMemoryStoreTests.cs
JadynWong: Yes, it originally verified that the embeddings value had been updated after the records with the same key were updated. However, when the embedded types are updated, the is updated to verity equal.

lemillermicrosoft (approved):

Issue #2640, opened 2023-08-29
Feature Request: ITextCompletion wrappers to support IChatCompletion constructs


lemillermicrosoft:
Opened in context of #2504 -- so the changes could be specific for StepwisePlanner and its implementation of Text and Chat completion support or more generic in kernel or abstractions.

andhesky:
I need this wrapper for another purpose, so I went ahead and finished the implementation. PR linked above.

matthewbolanos:
Closing since text completion will be deprecated

PR #2690, opened 2023-09-01
.Net: Text to chat wrapper

andhesky:
This is a draft. I can add comments and cleanup formatting if we want to go forward.

### Motivation and Context
Add a chat wrapper for text completions API to allow code to be written against the chat API and executed against either a Text or Chat API.

Resolves https://github.com/microsoft/semantic-kernel/issues/2640
Also fixes a bug in ChatStreamingResult.GetChatMessageAsync which would only ever return the last message in the steam instead of all messages. I also opened https://github.com/microsoft/semantic-kernel/issues/2730 to remove that method from the interface.

Issues with implementation:
* The ChatModelResult and TextModelResult are still different. Since these are taken directly from the Azure SDK I did not convert them. It seems there is an opportunity for an abstraction layer in SK so these objects are not exposed directly to callers. That would also give flexibility in the future to not use the Azure SDK. For now, just changed the ModelResult classes to share a single base class so at least as long as you are using common fields you can switch between the two.

### Description
Add wrapper classes for text completion objects and associated tests

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/TextWrapper/ChatToTextConverter.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/TextWrapper/ChatToTextConverter.cs
@@ -2,7 +2,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Data;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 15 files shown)


dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/TextWrapper/ChatToTextConverter.cs
```diff
@@ -0,0 +1,157 @@
...
+    private sealed class ChatResultFromTextResult : IChatResult
+    {
+        private readonly ITextResult _textResults;
+
+        public ChatResultFromTextResult(ITextResult textResults)
+        {
+            this._textResults = textResults;
+        }
+
+        public ModelResult ModelResult => this._textResults.ModelResult; // TODO return ChatModelResult
```
andhesky, line 43: >// TODO return ChatModelResult [](http://example.com/codeflow?start=73&length=30)

Not sure how/if to handle this. It is potentially a breaking difference between the wrapped interface and the native chat interface.

markwallace-microsoft: @RogerBarreto could you take a look at this comment?

RogerBarreto:
> If callers want their own implementation they can always write their own wrapper and register it as a chat completion and it will be picked up by this planner.

I agree that we can go ahead with this change except for the Conversion responsibility part in `Abstractions` namespace which could be a good candidate for another `Utility` package.

RogerBarreto:
Closing this due to inactivity.

PR #2717, opened 2023-09-05
.Net Removing KernelConfig, moving last property to KernelBuilder

RogerBarreto:
### Motivation and Context

⚠️ Breaking change.

As all the configuration is being centralized in the builder and `DelegatingHandlerFactory` was the last property living in this class, this PR completes the move and extinguish `KernelConfig`.

### Description

Move out the KernelConfig dependency to the builder

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
```diff
@@ -49,9 +50,9 @@ public static class OpenAIKernelBuilderExtensions
         bool setAsDefault = false,
         HttpClient? httpClient = null)
     {
-        builder.WithAIService<ITextCompletion>(serviceId, (loggerFactory, config) =>
+        builder.WithAIService<ITextCompletion>(serviceId, (loggerFactory) =>
         {
-            var client = CreateAzureOpenAIClient(loggerFactory, config, deploymentName, endpoint, new AzureKeyCredential(apiKey), httpClient);
+            var client = CreateAzureOpenAIClient(loggerFactory, builder.HttpHandlerFactory, deploymentName, endpoint, new AzureKeyCredential(apiKey), httpClient);
             return new AzureTextCompletion(deploymentName, client, loggerFactory);
         }, setAsDefault);
```
dmytrostruk, line 57: I'm not sure about this pattern: `builder.WithAIService(() => ...builder.HttpHandlerFactory...)`. 
Shouldn't it be like this?
```csharp
builder.WithAIService<ITextCompletion>(serviceId, (loggerFactory, httpHandlerFactory) =>
{
    var client = CreateAzureOpenAIClient(loggerFactory, httpHandlerFactory, deploymentName, endpoint, new AzureKeyCredential(apiKey), httpClient);
    return new AzureTextCompletion(deploymentName, client, loggerFactory);
}, setAsDefault);
```

Same applies to all other extension methods. In this case you don't need to expose `HttpHandlerFactory` from `KernelBuilder`.

dotnet/src/SemanticKernel.Abstractions/Http/NullHttpHandlerFactory.cs
```diff
@@ -5,8 +5,10 @@
 
 namespace Microsoft.SemanticKernel.Http;
 
-internal sealed class NullHttpHandlerFactory : IDelegatingHandlerFactory
+public sealed class NullHttpHandlerFactory : IDelegatingHandlerFactory
```
dmytrostruk, line 8: Public class and all class members need XML documentation :)

dotnet/src/SemanticKernel/KernelBuilder.cs
```diff
@@ -22,17 +22,18 @@ namespace Microsoft.SemanticKernel;
...
     private Func<IMemoryStore>? _memoryStorageFactory = null;
-    private IDelegatingHandlerFactory? _httpHandlerFactory = null;
+    private IDelegatingHandlerFactory _httpHandlerFactory = NullHttpHandlerFactory.Instance;
     private IPromptTemplateEngine? _promptTemplateEngine;
     private readonly AIServiceCollection _aiServices = new();
 
     private static bool _promptTemplateEngineInitialized = false;
     private static Type? _promptTemplateEngineType = null;
 
+    public IDelegatingHandlerFactory HttpHandlerFactory => this._httpHandlerFactory;
```
dmytrostruk, line 35: As mentioned above, I'm not sure it's required. Otherwise, it will also need XML documentation.

dotnet/samples/KernelSyntaxExamples/Example42_KernelBuilder.cs
```diff
@@ -80,8 +80,8 @@ public static Task RunAsync()
         using var memory = new SemanticTextMemory(memoryStorage, textEmbeddingGenerator);
         var skills = new SkillCollection();
         var templateEngine = new PromptTemplateEngine(loggerFactory);
-        var kernelConfig = new KernelConfig();
-        using var httpHandler = kernelConfig.HttpHandlerFactory.Create(loggerFactory);
+        var httpHandlerFactory = NullHttpHandlerFactory.Instance;
```
lemillermicrosoft, line 83: This doesn't need to be done, does it? Is this why it was made public? Maybe we keep it internal and use the Basic or Polly factory here instead.

dotnet/src/SemanticKernel/KernelBuilder.cs
lemillermicrosoft: I expected a WithDHFactory or something on Builder such that the factory could be passed similar to config to other object initialization.

RogerBarreto: It's not, thanks for the good sight!

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
RogerBarreto: Completely agree, missed this totally. Thanks for the review!

dmytrostruk (approved):

PR #2718, opened 2023-09-05
.Net: Hide memory collections and structures for internal usage only

dmytrostruk:
### Motivation and Context

Predecessor: https://github.com/microsoft/semantic-kernel/pull/2397
Resolves: https://github.com/microsoft/semantic-kernel/issues/616

This PR contains changes to hide memory-related collections and structures from public API surface in order to use them only in internal functionality. 

Reasons: 
- Structure `Score` contains only one `double Value` property, which means that `ScoredValue` can use `double` directly.
- Structure `ScoredValue` produces inconsistent results in equality and comparison operators. While it serves its purpose for internal functionality, it can behave incorrectly in user scenarios.
- Class `TopNCollection` can be replaced with already existing `PriorityQueue` in projects starting from .NET 6 version.

### Description

1. Marked `ScoredValue` and `TopNCollection` types as `internal`.
2. Removed `Score` structure and replaced it with `double`.
3. Updated memory connectors which used `Score`, `ScoredValue` and/or `TopNCollection` types directly.
4. Added small fix to `ScoredValue` structure in `Equals` method.
5. Fixed `Example25_ReadOnlyMemoryStore` based on updates from previous PRs.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example25_ReadOnlyMemoryStore.cs
```diff
@@ -148,11 +147,9 @@ public IAsyncEnumerable<string> GetCollectionsAsync(CancellationToken cancellati
                 }
             }
 
-            embeddings.SortByScore();
-
-            foreach (var item in embeddings)
+            foreach (var item in embeddings.OrderByDescending(l => l.Item2).Take(limit))
             {
-                yield return (item.Value, item.Score.Value);
+                yield return (item.Item1, item.Item2);
```
RogerBarreto, line 152: Use the tuple for clarity where possible. Not sure if you are able to use it in the OrderDescendingBy.

```suggestion
            foreach (var (memoryRecord, similarity) in embeddings.OrderByDescending((mr, similarity) => similarity).Take(limit))
            {
                yield return (memoryRecord, similarity);
```

dotnet/src/Connectors/Connectors.Memory.DuckDB/DuckDBMemoryStore.cs
```diff
@@ -153,7 +152,7 @@ public async Task RemoveBatchAsync(string collectionName, IEnumerable<string> ke
         }
 
         var collectionMemories = new List<MemoryRecord>();
-        TopNCollection<MemoryRecord> embeddings = new(limit);
+        List<(MemoryRecord, double)> embeddings = new();
```
RogerBarreto, line 155: Same as above.

RogerBarreto (commented):
Suggest changing the tuple impl for better clarity.

RogerBarreto (approved):

lemillermicrosoft (approved):

Issue #2228, opened 2023-07-31
.NET - Registering IKernel add a feature to register Azure Open AI, LLAMA or multiple models


praneetnadkar:
When we have prompts, we can dynamically inject the version of the IKernel we need, whether it is using OpenAI or LLAMA or anything other. There can be prompts or asks which dont require me to go to GPT-4 or 3.5 and I may use LLAMA for this. Or there can be prompts where I would specifically need GPT-4. 
Can SK Support multiple models based on the prompts?

Is this already supported by SK planners?

evchaki:
@praneetnadkar - in the prompt template you will be able to define which model you want to use.  We found a bug in doing this and we are working to fix it.

PR #2739, opened 2023-09-06
Add a feature to specify model for a semantic function

markwallace-microsoft:
### Motivation and Context

  1. Why is this change required? Allows a semantic function config to include an LLM service id
  2. If it fixes an open issue, please link to the issue here: https://github.com/microsoft/semantic-kernel/issues/2228

### Description

- Add ServiceId to PromptTemplateConfig to allow AI service to be configured

### Contribution Checklist

### DIFF
--- a/dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
+++ b/dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
@@ -10,6 +10,7 @@ using Microsoft.SemanticKernel;
 using Microsoft.SemanticKernel.Diagnostics;
 using Microsoft.SemanticKernel.Orchestration;
 using Microsoft.SemanticKernel.Reliability.Basic;
+using Microsoft.SemanticKernel.SemanticFunctions;
 using Microsoft.SemanticKernel.SkillDefinition;
 using SemanticKernel.IntegrationTests.TestSettings;
 using Xunit;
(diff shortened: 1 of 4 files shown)


dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
```diff
@@ -351,6 +352,103 @@ public async Task AzureOpenAIDefaultValueTestAsync()
         Assert.Contains("Bob", actual.Result, StringComparison.OrdinalIgnoreCase);
     }
 
+    [Fact]
+    public async Task MultipleServiceInlineServiceIdTestAsync()
```
dmytrostruk, line 356: While it's good to have more integration tests, I'm wondering if scenario in this PR can be verified with unit tests. Because these tests will make a call to Azure OpenAI endpoint every time, but they verify the logic with service registration only. With proper mocking it should be possible to verify which service is used during function invocation without actual call to Azure OpenAI. This should save some resources.

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
```diff
@@ -351,6 +352,103 @@ public async Task AzureOpenAIDefaultValueTestAsync()
...
+    {
+        // Arrange
+        var azureOpenAIConfiguration = this._configuration.GetSection("AzureOpenAI").Get<AzureOpenAIConfiguration>();
+        Assert.NotNull(azureOpenAIConfiguration);
```
dmytrostruk, line 360: It looks like this check is not needed here, because we already fetch the configuration and check if it's null in two methods below: `ConfigureAzureOpenAI` and `ConfigureInvalidAzureOpenAI`.

Same applies to other test methods where this check is performed.

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
```diff
@@ -427,6 +525,22 @@ private void ConfigureAzureOpenAI(KernelBuilder kernelBuilder)
...
             setAsDefault: true);
     }
+    private void ConfigureInvalidAzureOpenAI(KernelBuilder kernelBuilder)
+    {
+        var azureOpenAIConfiguration = this._configuration.GetSection("AzureOpenAI").Get<AzureOpenAIConfiguration>();
+
+        Assert.NotNull(azureOpenAIConfiguration);
+        Assert.NotNull(azureOpenAIConfiguration.DeploymentName);
+        Assert.NotNull(azureOpenAIConfiguration.Endpoint);
+        Assert.NotNull($"invalid-{azureOpenAIConfiguration.ServiceId}");
```
dmytrostruk, line 535: Value in this line will never be null.

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
markwallace-microsoft: Good catch, looks like the pattern was changed in some tests and this code was never removed and now I'm copying the wrong pattern

SergeyMenshykh (approved):

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
markwallace-microsoft: I'll replace this test with unit tests

dmytrostruk (approved):

PR #2775, opened 2023-09-11
.Net: Enable timestamp for Kusto records

freistli:
### Motivation and Context

Kusto query can use timestamp field to query records. However ever [SemanticTextMemory](https://github.com/microsoft/semantic-kernel/blob/cfebcaec6c1656f2c129ca3e40442c0ace9495e5/dotnet/src/SemanticKernel/Memory/SemanticTextMemory.cs#L45) 
 doesn't set this field in MemoryRecord, and Kusto MemoryStore doesn't set it as well if the timestamp is null in the input memory record.

### Description

This fix enables TimeStamp for Kusto records, therefore Kusto query can run with this pattern:

union *
| where Timestamp > ago(1h)

### Contribution Checklist


dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryRecord.cs
```diff
@@ -38,7 +38,7 @@ public sealed class KustoMemoryRecord
     /// Initializes a new instance of the <see cref="KustoMemoryRecord"/> class.
     /// </summary>
     /// <param name="record">Instance of <see cref="MemoryRecord"/>.</param>
-    public KustoMemoryRecord(MemoryRecord record) : this(record.Key, record.Metadata, record.Embedding, record.Timestamp) { }
+    public KustoMemoryRecord(MemoryRecord record) : this(record.Key, record.Metadata, record.Embedding, record.Timestamp ?? DateTimeOffset.UtcNow) { }
```
dmytrostruk, line 41: I'm not sure if this is the right approach to set `Timestamp` field. As Kusto connector, we should provide an interface to set `Timestamp` field, but we can't decide on behalf of user, what value we should use as default, for some users it can be `UtcNow`, other users may prefer local time based on their scenario. 

Regarding queries like `| where Timestamp > ago(1h)`, in this case I think field `Timestamp` should be as part of Kusto system fields, which should be set automatically on Kusto side.

But I do agree, that `SemanticTextMemory` should accept `timestamp` field as part of schema, because otherwise there is no option to provide this field, unless using `KustoMemoryStore` directly, which is not preferred way in context of Semantic Kernel.

dmytrostruk (commented):
@freistli Thank you for contribution. I think the absence of the `Timestamp` field as part of schema in `SemanticTextMemory` may be a problem, which we will need to resolve in different way (e.g. extending current interface). But before doing that, I will need to discuss it with internal team first and will let you know the results as soon as possible. Thanks again!

dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryRecord.cs
freistli: Thanks @dmytrostruk , if SemanticTextMemory provides parameter to set this value, can be another possible way, but not all memory store from connectors can handle timestamp correctly. 

For example, in original Kusto MemoryStore code, [here](https://github.com/microsoft/semantic-kernel/blob/cfebcaec6c1656f2c129ca3e40442c0ace9495e5/dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs#L134) and [here](https://github.com/microsoft/semantic-kernel/blob/cfebcaec6c1656f2c129ca3e40442c0ace9495e5/dotnet/src/Connectors/Connectors.Memory.Kusto/KustoMemoryStore.cs#L217) , it tries to read timestamp as string, but indeed this field is DateTimeOffset. 

Another thing is Kusto doesn't set this field automatically, that's why I have to set it in Kusto Connector in this way. 

From my point of view, before all connectors can well accept timestamp setting, and SemanticTextMemory starts well support this option, for Kusto Connector scenario, this method is an achievable solution. 

We define the default behavior is UTC DateTimeOffset. If SemanticTextMemory provides this value. Kusto can still use it because the code accept explicitly setting the value from record (which is from SemanticTextMemory(: record.Timestamp ?? DateTimeOffset.UtcNow

dmytrostruk: > but not all memory store from connectors can handle timestamp correctly

`Timestamp` parameter can be optional, as well as other parameters that currently exist. 

> it tries to read timestamp as string, but indeed this field is DateTimeOffset

I think there was some issue with `reader.GetDateTime(2)`, we will need to test it manually and double-check.

> We define the default behavior is UTC DateTimeOffset

As I said in previous comment, I'm not sure that we can/should define this default behavior. In my opinion, it should be `null` by default (as optional field), and users can define their default values as soon as we extend `SemanticTextMemory`. The change in this PR can fix some scenarios for now, but we need to think about long-term solution, which will bring benefits not only for Kusto connector, but other connectors as well.

dmytrostruk:
@freistli Thanks again for your contribution. As per discussion with internal team, it looks like adding `Timestamp` field to `ISemanticTextMemory` methods should be a right way how to resolve this problem. In this case, it will be possible to set default value on user side, so users can decide which default value works the best.

I created separate issue for this task: https://github.com/microsoft/semantic-kernel/issues/2792.
Feel free to work on it if you have an opportunity - your contributions are very welcome! 

This task is a good candidate for separate PR, so I will close this one for now. In case you want to proceed within this PR, feel free to re-open it. Thank you!

PR #2776, opened 2023-09-11
.Net [WIP] Kernel Streaming

RogerBarreto:
### Contribution Checklist


anthonypuppo:
Is there a specific reason to use `System.IO.Stream`? `System.Threading.Channels` could be a good alternative as it already provides support for the producer/consumer pattern, orchestrating streaming events from async operations (such as simultaneous LLM calls?), etc. See [docs](https://learn.microsoft.com/en-us/dotnet/core/extensions/channels) and [devblog](https://devblogs.microsoft.com/dotnet/an-introduction-to-system-threading-channels/). Easy to utilize without sacrificing performance over raw streams.

Would probably be beneficial to wrap the underlying channel instance in a custom class (`SKChannel`?), maybe with some helper method support like so:
```
public IAsyncEnumerable<T> ToAsyncEnumerable()
{
    return channel.Reader.ReadAllAsync();
}
```

If compatibility is a concern since streams are so widely supported the channel reader can create and/or write to a stream pretty easily:
```
public static async Task WriteChannelToStreamAsync(this ChannelReader<byte[]> reader, Stream stream)
{
    await foreach (var item in reader.ReadAllAsync())
    {
        await stream.WriteAsync(item, cancellationToken);
    }
}
```

Just a thought?

RogerBarreto:
Hi @anthonypuppo Thanks for your suggestion. 

Currently this is just a Draft idea on how we could be handling Streaming scenarios, not final thought and potentially will change completely, we are considering having a InvokeAsync supporting `Stream` or IAsyncEnumerable<byte> or IAsyncEnumerable<string> extensions.

Regarding the question why Stream.
The most important reason for using streaming approach is due to the nature of the operation and the multi modals related to streaming data (Video, Audio, Image, Files).

Normally Channels are recommended for scenarios like:

- When you need to establish a communication mechanism between different parts of your application that doesn't require real-time data streaming.
- Send discrete messages or events, such as notifications, updates, or commands.
- Used for implementing messaging systems within a microservices architecture, event-driven systems, or for handling asynchronous communication between components.

PR #2781, opened 2023-09-11
.Net Potential concurrency bugs in FunctionsView Fix

### LINKED ISSUE #604: Potential concurrency bugs in FunctionsView
FunctionsView appears to be trying to be thread-safe, as its SemanticFunctions and NativeFunctions are both ConcurrentDictionary.  However, in that case there are a couple of concurrency bugs lurking.

First, while the dictionaries are ConcurrentDictionary, the values stored for each key are `List<T>`.  The code for AddFunction looks like this:
```C#
            if (!this.SemanticFunctions.ContainsKey(view.SkillName))
            {
                this.SemanticFunctions[view.SkillName] = new();
            }

            this.SemanticFunctions[view.SkillName].Add(view);
```
If multiple threads concurrently try to add a FunctionView for the same skill, they could end up both mutating a `List<T>` concurrently, which is not safe.

Second, with that same code, if multiple threads both try to add a FunctionView for the same skill, and if both see that the SemanticFunctions doesn't yet contain that skill name, they will race to create the list to store for that skill.  While the dictionary itself won't be corrupted from a data structure perspective, one thread might add its list and then proceed to add to the list, and the other thread might then (having already checked ContainsKey and found it to return false) overwrite the original list with a new one, there by losing the data the other thread wrote.  Such initialization should be done using either TryAdd or GetOrAdd so that the dictionary provides the synchronization to ensure nothing is overwritten.

These issues can easily be fixed by using GetOrAdd and locking on the `List<>`... except the collections are exposed publicly from the FunctionsView, which means anyone else accessing these would also need to employ the same synchronization scheme.  That suggests the API should be designed differently.

(And related to these being exposed publicly, the properties also have public setters... if someone were to set these concurrently with other operations being performed, that could lead to other failure modes, e.g. because lookups are being done first by calling ContainsKey and then separately by using the indexer, the collection could actually be changed to a completely different instance between those calls.)

cc: @dluc, @shawncal

RogerBarreto:
### Motivation and Context

FunctionsView appears to be trying to be thread-safe, as its SemanticFunctions and NativeFunctions are both ConcurrentDictionary. However, in that case there are a couple of concurrency bugs lurking.

If multiple threads concurrently try to add a FunctionView for the same skill, they could end up both mutating a List<T> concurrently, which is not safe.

Resolves #604 

### Description

Applied suggestions.

- Using `GetOrAdd` instead of `Contains`
- Added lock when Adding new Function details to the list (prevent concurrency)
- Removed the FunctionView setters
- Improved `IsSemantic` and `IsNative` logic
- Added Concurrency UnitTest

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -38,23 +41,16 @@ public sealed class FunctionsView
     /// <returns>Current instance</returns>
     public FunctionsView AddFunction(FunctionView view)
     {
-        if (view.IsSemantic)
+        lock (this._lock)
```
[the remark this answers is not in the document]
RogerBarreto: Had a look, but BlockingCollection seems to work more for Stacking (FIFO / FILO) Producer / Consumer scenarios.

https://learn.microsoft.com/en-us/dotnet/standard/collections/thread-safe/blockingcollection-overview

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -17,19 +18,21 @@ namespace Microsoft.SemanticKernel.SkillDefinition;
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 public sealed class FunctionsView
 {
+    private object _lock = new();
```
dmytrostruk, line 21: Small formatting improvement:
```suggestion
    private object _lock = new();
    
```

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -17,19 +18,21 @@ namespace Microsoft.SemanticKernel.SkillDefinition;
...
     /// <summary>
     /// Collection of semantic skill names and function names, including function parameters.
     /// Functions are grouped by skill name.
     /// </summary>
-    public ConcurrentDictionary<string, List<FunctionView>> SemanticFunctions { get; set; }
-        = new(StringComparer.OrdinalIgnoreCase);
+    public IReadOnlyDictionary<string, IReadOnlyCollection<FunctionView>> SemanticFunctions => this.GetFunctions(nativeFunctionCheck: false);
```
dmytrostruk, line 26: Is `GetFunctions` method necessary here? 
Can we do just:
```csharp
public IReadOnlyDictionary<string, IReadOnlyCollection<FunctionView>> SemanticFunctions => new ReadOnlyFunctionsView(this._semanticFunctions);
```
Same applies for `NativeFunctions`.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -94,22 +77,120 @@ public bool IsSemantic(string skillName, string functionName)
...
+#pragma warning restore CA1859 // Use concrete types when possible for improved performance
+    {
+        ConcurrentDictionary<string, List<FunctionView>> targetFunctionDictionary = (nativeFunctionCheck)
+            ? this._nativeFunctions
+            : this._semanticFunctions;
+
+        return new ReadOnlyFunctionsView(targetFunctionDictionary);
+    }
+
+    private class ReadOnlyFunctionsView : IReadOnlyDictionary<string, IReadOnlyCollection<FunctionView>>
```
dmytrostruk, line 94: Classes `ReadOnlyFunctionsView` and `ReadOnlyFunctionsViewEnumerator` are nested, but they don't use any members from parent class. Also, they are located between `FunctionsView` class methods, which makes them harder to read. Can we move these classes to separate files and make them `internal`?

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -94,22 +77,120 @@ public bool IsSemantic(string skillName, string functionName)
...
+            return this._decorated.MoveNext();
+        }
+
+        public void Reset()
+        {
+            this._decorated.Reset();
+        }
+    }
+
+    private bool IsFunctionCheck(string skillName, string functionName, bool nativeFunctionCheck)
```
dmytrostruk, line 174: nit: I'm not sure about name `IsFunctionCheck` based on its implementation. Because we definitely know that it's a function, we just need to know whether it's native or semantic.

The usage looks like:
```csharp
this.IsFunctionCheck(skillName, functionName, nativeFunctionCheck: false);
```

Maybe we can rename it to something like:
```csharp
this.CheckFunction(skillName, functionName, nativeFunctionCheck: false);
// or
this.CheckFunctionType(skillName, functionName, isNative: false);
```

We can experiment with naming combinations here.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -94,22 +77,120 @@ public bool IsSemantic(string skillName, string functionName)
...
-        if (sf && nf)
+        if (foundSemanticFunction && foundNativeFunction)
         {
             throw new AmbiguousMatchException("There are 2 functions with the same name, one native and one semantic");
         }
 
-        return nf;
+        return (nativeFunctionCheck)
+            ? foundNativeFunction
+            : foundSemanticFunction;
```
dmytrostruk, line 191: nit: I think this can be one-liner without parentheses:
```suggestion
        return nativeFunctionCheck ? foundNativeFunction : foundSemanticFunction;
```

dotnet/src/SemanticKernel.UnitTests/SkillDefinition/FunctionsViewTests.cs
```diff
@@ -112,4 +115,86 @@ public void ItReturnsFunctionParams()
...
+        Assert.ThrowsAny<Exception>(() => (List<FunctionView>)result.Values);
+        Assert.ThrowsAny<Exception>(() => (IList<FunctionView>)result.Values);
+        Assert.ThrowsAny<Exception>(() => (ICollection<FunctionView>)result.Values);
+        Assert.ThrowsAny<Exception>(() => (List<FunctionView>)result.GetEnumerator().Current.Value);
+        Assert.ThrowsAny<Exception>(() => (IList<FunctionView>)result.GetEnumerator().Current.Value);
+        Assert.ThrowsAny<Exception>(() => (ICollection<FunctionView>)result.GetEnumerator().Current.Value);
+    }
+
+    [Fact]
+    public async Task ItAddSemanticAndNativeFunctionsShouldBeThreadSafe()
```
dmytrostruk, line 162:
```suggestion
    public async Task AddSemanticAndNativeFunctionsShouldBeThreadSafe()
```

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -17,19 +18,21 @@ namespace Microsoft.SemanticKernel.SkillDefinition;
...
 
     /// <summary>
     /// Collection of native skill names and function views, including function parameters.
     /// Functions are grouped by skill name.
     /// </summary>
-    public ConcurrentDictionary<string, List<FunctionView>> NativeFunctions { get; set; }
-        = new(StringComparer.OrdinalIgnoreCase);
+    public IReadOnlyDictionary<string, IReadOnlyCollection<FunctionView>> NativeFunctions => this.GetFunctions(nativeFunctionCheck: true);
+
+    private readonly ConcurrentDictionary<string, List<FunctionView>> _semanticFunctions = new(StringComparer.OrdinalIgnoreCase);
```
shawncal, line 34: We have no requirement for these to be Concurrent, and the rest of the Kernel is not built for that either. Can we just drop the overhead of concurrency, and just make it a Dictionary? This is just a Functions*View* - a snapshot of the functions - and shouldn't be modified by multiple (or even one) thread after creation anyway.

stephentoub: > Can we just drop the overhead of concurrency, and just make it a Dictionary?

+1. We should move in that direction everywhere we don't absolutely need the concurrency guarantees (especially since they often require more than just a different dictionary type).

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
RogerBarreto: They are not used outside of the FunctionsView class, might keep it inside a region and move the private property before them.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionsView.cs
```diff
@@ -17,19 +18,21 @@ namespace Microsoft.SemanticKernel.SkillDefinition;
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 public sealed class FunctionsView
```
shawncal, line 19: I actually think this whole class can be deleted, with references replaced by Dictionary<FunctionView> (for creation) and IReadOnlyDictionary<FunctionView> for accessing.

We'll be removing any places where there are differentiators between "NativeFunction" and "SemanticFunction", as they'll all just be functions. Once that's gone, this is just a wrapper around a Dictionary, and may as well be a Dictionary IMO.

RogerBarreto: That makes sense, lets do this when the Native/Semantic functions distinction is gone.

RogerBarreto:
Closing this PR as the `FunctionsView` class is being removed #2878

PR #2797, opened 2023-09-12
.Net: Fix null ref in HttpClientExtensions.SendWithSuccessCheckAsync

stephentoub:
### Motivation and Context

If the SendAsync fails, response is null, the the access to response.StatusCode null refs.

### Contribution Checklist


dotnet/src/InternalUtilities/src/Http/HttpClientExtensions.cs
```diff
@@ -41,7 +42,7 @@ internal static async Task<HttpResponseMessage> SendWithSuccessCheckAsync(this H
             }
             catch { } // We want to suppress any exceptions that occur while reading the content, ensuring that an HttpOperationException is thrown instead.
 
-            throw new HttpOperationException(response!.StatusCode, responseContent, e.Message, e);
+            throw new HttpOperationException(response?.StatusCode ?? HttpStatusCode.BadRequest, responseContent, e.Message, e);
```
RogerBarreto, line 45: Should we consider using (HttpStatusCode)0 ? NoResponse?

https://github.com/microsoft/semantic-kernel/blob/4ede383dcd77054066813b855b10009d5e4209c1/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/RequestFailedExceptionExtensions.cs#L22

shawncal: In [RequestFailedExceptionExtensions.cs](https://github.com/microsoft/semantic-kernel/blob/4ede383dcd77054066813b855b10009d5e4209c1/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/RequestFailedExceptionExtensions.cs#L22) it looks like, if the exception's statusCode is 0, it just replaces it with `null`. 

This PR seeks to avoid null status codes (and crashes), by setting them to a default value.

@RogerBarreto are you suggesting that we aligh the [RequestFailedExceptionExtensions.cs](https://github.com/microsoft/semantic-kernel/blob/4ede383dcd77054066813b855b10009d5e4209c1/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/RequestFailedExceptionExtensions.cs#L22) version to also avoid null status codes (instead of deliberately creating them)?

SergeyMenshykh: The reason for making the `StatusCode` property nullable was to communicate another "NoResponse" state via a `null` value. The property comment reflects that: "Gets or sets the HTTP status code. If the property is null, it indicates that no response was received."

An alternative solution, which is implemented in this PR, is to use the `BadRequest` status code instead of `null` to indicate a no-response situation. According to the `BadRequest` constant documentation, it can be used for that purpose: "... BadRequest is sent when no other error is applicable, or if **the exact error is unknown or does not have its own error code**."

Regardless of the chosen solution, the behavior should be consistent.

shawncal (approved):

dotnet/src/InternalUtilities/src/Http/HttpClientExtensions.cs
shawncal: @stephentoub I'll merge the PR now, but we should align these approaches as mentioned above.

PR #2800, opened 2023-09-12
Python: USearch dependency update and fix

### LINKED ISSUE #2676: Python: Upgrade USearch Memory Store to use v2.x.x
---
name: Upgrade USearch Memory Store to use the latest package version

This task cannot be done automatically with dependabot since the newest version (2.x.x) has breaking changes from the current version (1.x.x)

---

alexbarev:
### Motivation and Context

1) To keep the USearch dependency up-to-date by upgrading it to version 2.4.0

2) To fix the implementation of `min_relevance_score` in the `get_nearest_matches_async` method, aligning it more closely with USearch distance-based similarity measurement.

Resolves #2676 
Closes #2676 

### Description

Fix the implementation of `min_relevance_score` in the `get_nearest_matches_async` function for USearch.

**Background**

In USearch, similarity between vectors is determined by distance, not by relevance score. A lower distance indicates higher similarity between vectors.

**Before the Change**

Before, the same formula for all metrics was used to change distances into relevance scores that fell within the range $[0; 1]$ . Which works for distances in $[0; +\infty)$.

https://github.com/microsoft/semantic-kernel/blob/a28728898eb98ca6fb6a4d9db0968296de1bdfae/python/semantic_kernel/connectors/memory/usearch/usearch_memory_store.py#L560

This allowed user to filter out vectors from `get_nearest_matches_async` that had relevance scores higher than `min_relevance_score`. But this method was not accurate. Different metrics implementation have different distance ranges even with negative values.

**After the Change**

Now, the function will only return vectors that have a distance **less** than or equal to the specified `min_relevance_score`. This makes the function more accurate and aligns it with the way USearch actually works.

### Contribution Checklist


moonbox3:
Hi @alexbarev, thank you for your contribution. If these changes are still needed, can you please either re-create the PR against main, or work on fixing the conflicts? I will close this PR as of now.

PR #2801, opened 2023-09-12
.Net: Optimize 'Publish' Configuration DLLs

shawncal:
### Description
Optimize the 'Publish' builds by adding the Optimize=true compiler flag. Note: these are the NuGet-packed, shipping binaries.
Remove debugger overrides for Release and Debug builds (use defaults instead).

### Contribution Checklist

### DIFF
--- a/dotnet/Directory.Build.props
+++ b/dotnet/Directory.Build.props
@@ -19,7 +19,10 @@
   <PropertyGroup Condition="'$(Configuration)'=='Publish'">
     <Optimize>True</Optimize>
   </PropertyGroup>
+<<<<<<< HEAD
 
+=======
+>>>>>>> 04c3dee20aa7182d916a235d9cc16eb3458d4b89
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


dotnet/Directory.Build.props
```diff
@@ -16,13 +16,8 @@
...
 
-  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-  </PropertyGroup>
-
-  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
-    <DebugType>portable</DebugType>
+  <PropertyGroup Condition="'$(Configuration)'=='Publish'">
+    <Optimize>True</Optimize>
```
lemillermicrosoft, line 20: Do we need debugsymbols here? Folks just today were asking to use them and very happy they were available on nuget automatically.

lemillermicrosoft: Maybe we can revive later for `nightly` publishes only.

lemillermicrosoft (approved):

dotnet/Directory.Build.props
shawncal: Yeah, we can easily override these msbuild props in the `dotnet build...` command args, so customizing for certain pipelines is no problem. Good idea!

PR #2802, opened 2023-09-13
.Net: Allow Action Planner to Exclude Plugins and Functions

### LINKED ISSUE #1851: Action Planner - Allow to exclude Plugins/Functions
**Describe the bug**
Please create a way to exclude plugins/functions when using Action Planner. I am able to do this on Sequential and Stepwise Planners but not from Action Planner.

hario90:
### Motivation and Context
The Action Planner was missing the ability to exclude plugins and functions the way that Sequential and Stepwise Planners are able to.

Fixes #1851

### Description

Changes
* Added config to Action Planner constructor for parity with the other planners
* Created a base class for all planner configs to share common values 
* Filtered out excluded plugins and functions within the `ListOfFunctions` method of ActionPlanner

### Contribution Checklist


dotnet/src/SemanticKernel/Planning/PlannerConfigBase.cs
lemillermicrosoft, line 1: Awesome, thanks for taking the first step here. I think we can expand on this iteratively.

lemillermicrosoft (approved):
Overall LGTM. A few format issues atm.

dotnet/src/Extensions/Extensions.UnitTests/Planning/ActionPlanner/ActionPlannerTests.cs
```diff
@@ -1,12 +1,15 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+using System;
 using System.Collections.Generic;
+using System.Text;
```
dmytrostruk, line 5: `System.Text` is not needed here, that's why build pipeline is failing:
image

Before commit or PR creation you can run `dotnet/build.sh` or `dotnet/build.cmd`, which should resolve such issues automatically.

dotnet/src/Extensions/Extensions.UnitTests/Planning/ActionPlanner/ActionPlannerTests.cs
```diff
@@ -149,4 +179,41 @@ private static Mock<ISKFunction> CreateMockFunction(FunctionView functionView)
...
+                .Returns(mockFunction.Object);
+            ISKFunction? outFunc = mockFunction.Object;
+            skills.Setup(x => x.TryGetFunction(It.Is<string>(s => s == skillName), It.Is<string>(s => s == name), out outFunc)).Returns(true);
+        }
+
+        skills.Setup(x => x.GetFunctionsView(It.IsAny<bool>(), It.IsAny<bool>())).Returns(functionsView);
+        return skills;
+    }
+
+    private readonly string _validPlanString = "Here is a possible plan to accomplish the user intent:\n\n{\"plan\":{\n\"rationale\": \"the list contains a function that allows to list pull requests\",\n\"function\": \"GitHubSkill.PullsList\",\n\"parameters\": {\n\"owner\": \"microsoft\",\n\"repo\": \"semantic-kernel\",\n\"state\": \"open\"\n}}}\n\nThis plan uses the `GitHubSkill.PullsList` function to list the open pull requests for the `semantic-kernel` repository owned by `microsoft`. The `state` parameter is set to `\"open\"` to filter the results to only show open pull requests.";
```
dmytrostruk, line 218: This can be constant with name `ValidPlanString`. Since we declared separate field for this value, we can also reuse it in `ExtractsAndDeserializesWellFormedJsonFromPlannerResult` test, it looks like the values are the same. 

Improvement ideas: in IDE it's hard to read this value, you have to use horizontal scroll :) We can move this string to separate .txt file and read from it, or break string in multiple lines for better readability. For other long strings in this file (located in tests) we can use line break.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -55,10 +56,12 @@ public sealed class ActionPlanner : IActionPlanner
     /// <param name="kernel">The semantic kernel instance.</param>
     /// <param name="prompt">Optional prompt override</param>
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
+    /// <param name="config">The planner configuration.</param>
     public ActionPlanner(
         IKernel kernel,
         string? prompt = null,
-        ILoggerFactory? loggerFactory = null)
+        ILoggerFactory? loggerFactory = null,
+        ActionPlannerConfig? config = null)
```
dmytrostruk, line 64: Maybe we can add `ActionPlannerConfig? config` parameter after `IKernel kernel` parameter to follow the same pattern as in `SequentialPlanner` and `StepwisePlanner` constructors.
I would definitely put `config` parameter before `ILoggerFactory`. I think having `ILoggerFactory` as last optional parameter is a good pattern and is used in most cases across codebase.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -234,6 +238,10 @@ No parameters.
     }
 
     #region private ================================================================================
+    /// <summary>
```
dmytrostruk, line 241: Small formatting improvement:
```suggestion

    /// <summary>
```

dotnet/src/SemanticKernel/Planning/PlannerConfigBase.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System.Collections.Generic;
+namespace Microsoft.SemanticKernel.Planning;
+
+/// <summary>
+/// Base class for planner configs
+/// </summary>
+public abstract class PlannerConfigBase
+{
+    #region Use these to configure which functions to exclude
```
dmytrostruk, line 11: Do we need `region` here? The name of the region looks like the hint for user, but I'm not sure it's going to be visible for user, maybe only if user will open SK source code and check the implementation of this class. In general, I would avoid regions in model/configuration classes such as `PlannerConfigBase` and `StepwisePlannerConfig`.

dotnet/src/SemanticKernel/Planning/PlannerConfigBase.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+namespace Microsoft.SemanticKernel.Planning;
```
dmytrostruk, line 4: Small formatting improvement:
```suggestion

namespace Microsoft.SemanticKernel.Planning;
```

lemillermicrosoft: I thought `dotnet format` took care of this before... :/

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
dmytrostruk: I think we can move `prompt` to `config` as well. 
Regarding logging: having different logger factories for planner and kernel should provide flexibility, but I'm not sure it will be a common case, most probably users will want to configure logger once at kernel level and reuse it in all modules. We can also use provided in constructor `loggerFactory`, and in case it's `null` - use the one from Kernel. 

It can be done in separate PR, but in this one I would put `config` parameter after `kernel` for consistency.

dotnet/src/Extensions/Extensions.UnitTests/Planning/ActionPlanner/ActionPlannerTests.cs
hario90: Thanks!

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
hario90: I changed the order of the args. 

Regarding putting the prompt in the config and using the logger factory of the kernel - it would make sense to make ActionPlanner more consistent with the others in general. For the prompt - the sequential planner uses a delegate for this like:
```
this._promptTemplate = this.Config.GetPromptTemplate?.Invoke() ?? EmbeddedResource.Read("Skills.StepwiseStep.skprompt.txt");
```
I think it'd make sense to match this pattern for the ActionPlanner. For the loggerFactory, I don't see a reason to be inconsistent with the other planners and provide a constructor arg when the others don't.

Either way, since there's questions about which one to go forward with - it would make sense to make a separate PR.

dotnet/src/Extensions/Extensions.UnitTests/Planning/ActionPlanner/ActionPlannerTests.cs
hario90: Made this a `const` and added actual newlines to the invalid plan string in one of the tests.

dotnet/src/SemanticKernel/Planning/PlannerConfigBase.cs
hario90: Removed

dotnet/src/SemanticKernel/Planning/PlannerConfigBase.cs
```diff
@@ -0,0 +1,24 @@
...
+
+/// <summary>
+/// Base class for planner configs
+/// </summary>
+public abstract class PlannerConfigBase
```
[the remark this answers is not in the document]
hario90: I can see the benefit of this - for example, the planners usually exclude the functions used to create a plan and with direct access to the HashSet, developers may clear the HashSet and run into unintended problems like the excluded function getting included again. 

If we restrict the interactions with the underlying HashSet, does it make sense to only allow additions to the Set? Only allowing additions would make the planner configs single use so I'm thinking we should also allow removals. 
However, given that this is a breaking change, I propose addressing this in a separate PR once we have the interface we want to go with. I'd also like to hear what other folks think.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -55,10 +56,12 @@ public sealed class ActionPlanner : IActionPlanner
     /// <param name="kernel">The semantic kernel instance.</param>
     /// <param name="prompt">Optional prompt override</param>
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
+    /// <param name="config">The planner configuration.</param>
     public ActionPlanner(
         IKernel kernel,
         string? prompt = null,
-        ILoggerFactory? loggerFactory = null)
+        ILoggerFactory? loggerFactory = null,
```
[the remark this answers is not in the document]
hario90: fixed

dotnet/src/Extensions/Extensions.UnitTests/Planning/ActionPlanner/ActionPlannerTests.cs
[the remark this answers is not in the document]
hario90: Fixed

dotnet/src/Extensions/Extensions.UnitTests/Planning/ActionPlanner/ActionPlannerTests.cs
```diff
@@ -149,4 +179,41 @@ private static Mock<ISKFunction> CreateMockFunction(FunctionView functionView)
...
+
+            mockFunction.Setup(x =>
+                    x.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<CompleteRequestSettings>(), It.IsAny<CancellationToken>()))
+                .Returns<SKContext, CompleteRequestSettings, CancellationToken>((context, settings, CancellationToken) =>
+                {
+                    context.Variables.Update("MOCK FUNCTION CALLED");
+                    return Task.FromResult(context);
+                });
+
+            skills.Setup(x => x.GetFunction(It.Is<string>(s => s == skillName), It.Is<string>(s => s == name)))
```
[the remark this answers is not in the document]
hario90: Simplified this to `x.GetFunction(skillName, name)` - The `It`'s were not necessary.

dotnet/src/Extensions/Planning.ActionPlanner/ActionPlanner.cs
```diff
@@ -53,12 +54,15 @@ public sealed class ActionPlanner : IActionPlanner
...
+    /// <param name="config">The planner configuration.</param>
     /// <param name="prompt">Optional prompt override</param>
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
     public ActionPlanner(
         IKernel kernel,
+        ActionPlannerConfig? config = null,
         string? prompt = null,
-        ILoggerFactory? loggerFactory = null)
+        ILoggerFactory? loggerFactory = null
+        )
```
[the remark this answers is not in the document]
hario90: Fixed. Looks like the build was passing when the parenthesis looked like this. But the formatters also didn't complain when I moved the parenthesis.

dmytrostruk (approved):
LGTM!

RogerBarreto (approved):
LGTM

PR #2526, opened 2023-08-22
.Net: Tokens information in telemetry

dmytrostruk:
### Motivation and Context

This PR contains changes in telemetry to add logging and metering for prompt, completion and total tokens used for each request to Azure OpenAI/OpenAI APIs.

Example of logging in Application Insights:
image

Example of metering in Application Insights:
image

### Description

1. Added logging and metering for tokens information in `ClientBase` class of OpenAI connector.
2. Fixed examples.
3. Updated telemetry documentation.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

Issue #131, opened 2023-03-21
Python: Use Pydantic for configuration


AdityaGudimella:
We have a few configuration classes in the python version like `BackendConfig`, `OpenAIConfig`, `AzureOpenAIConfig` etc. We perform operations like loading these configs from .env files using helper methods like `utils.settings::openai_settings_from_dot_env` etc. We also have a class named `Verify` that verifies config values like checking that a value is not None etc. [Pydantic](https://docs.pydantic.dev) is a popular open-source library that is particularly used in such cases. All the above mentioned functionality is built into Pydantic. I propose that we use pydantic for our configuration and replace the `Verify` class and the helper functions like `utils.settings::openai_settings_from_dot_env` with pydantic functions.

What are your thoughts on this?

jjhenkel:
I want to chime in here and say that, based on other open source projects in the LLM space I've seen, using something like `pydantic` would align us more closely with community standards. When I wrote the initial Python port I tried to make minimal changes from the C# version, but the config classes and the `Verify` class (and the helpers I've written) are very un-pythonic. 

In my opinion, it'd be great to move toward `pydantic`! (And, related, it may be nice to simplify configuration at some point... I feel like there are so many classes that hold so many of the same values, something like making a `PromptTemplate` has a lot of "ceremony" around it.)

dluc:
thanks for the great feedback. Sharing a few notes for the sake of optimizing the port to python: we're redesigning the support for backend and we expect to be a substantial change. The goal is allowing any backend, removing the current first implementation that is very OpenAI centric. The work is still in progress and most probably it will involve how configuration is managed. So it might be worth waiting a couple of weeks until that's sorted out, in case there are reusable ideas, conf files, etc.

dluc:
@awharrison-28 @mkarle FYI

dluc:
quick update: work in progress

nacharya1:
@dluc can you please update this issue on the latest status?

nacharya1:
@awharrison-28 - can you please update target dates/close this issue as needed.

evchaki:
Completed.

PR #2804, opened 2023-09-13
Python: added token usage with client base

eavanvalkenburg:
### Motivation and Context

Added token usage tracking to the OpenAI classes, similar to how that is done in dotnet.
Similar approach as #2526 but for Python
With to_dict and from_dict methods on the (Azure)OpenAI service classes.
TODO: add a similar construct to python for the SKContext.LastResult.

### Description

Adds a completion client base class with trackers for the prompt, completion and total tokens.
This has a add_tokens method that can be used to increment the totals.
This is called from the OpenAI completion classes to add the tokens.
Also added serialization to the service classes, using pydantic, building on #131 

Added deprication for "logger" parameter on Azure OpenAI service classes, this was inconsistent with OpenAI, which used "log", added both to AzureOpenAIChatCompletion class with a warning when using "logger" to not break existing implementations.

Added tests for from_dict and to_dict on the Azure classes.

### Contribution Checklist


eavanvalkenburg:
Split this PR into two

Issue #2679, opened 2023-09-01
encoder.json/vocab.bpe show up in every project that uses SK


stephentoub:
Bringing in the Microsoft.SemanticKernel nuget package causes these files to show up in the consuming application:
![image](https://github.com/microsoft/semantic-kernel/assets/2642209/8889bec5-bd34-495e-9b57-a4d50d315f80)
and they end up in the output directory for the application, regardless of whether the app is using the tokenizer or not.
![image](https://github.com/microsoft/semantic-kernel/assets/2642209/c90f7e44-caaf-451e-80f7-09a41a614dbe)

I'd opened https://github.com/microsoft/semantic-kernel/pull/1800 to turn them into assembly resources instead, so that they'd simply be part of the assembly and not separate files, but it was closed due to a lack of a decision about what to do with it.

anthonypuppo:
Echoing my comment on #1800 

Bringing up [SharpToken](https://github.com/dmitry-brazhenko/SharpToken) as this would solve 1) embedding of the tokenization resource files 2) extracting the tokenization logic to a separate package (maybe wrapped as an official SK package in the future) and 3) bug https://github.com/microsoft/semantic-kernel/issues/2334.

lemillermicrosoft:
One other option could be to move the tokenizer into an extension package for OpenAI connector.

Issue #2508, opened 2023-08-20
.Net: Remove GPT tokenizer


dluc:
Scope
1. Remove GPT Tokenizer from SDK
2. Remove tokenizers from notebook 

This code works in a typical C# project, but fails in a notebook, because GPT resources are in a different path.

Console app:
```csharp
using Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers;

Console.WriteLine(GPT3Tokenizer.Encode("hello world").Count + " tokens");
```
Output:
> 2 tokens

C# notebook:

```csharp
#r "nuget: Microsoft.SemanticKernel.Connectors.AI.OpenAI, 0.19.230804.2-preview"

using Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers;

Console.WriteLine(GPT3Tokenizer.Encode("hello world").Count + " tokens");
```

Output:
> Error: System.IO.FileNotFoundException: vocab.bpe not found, path: '~/.nuget/packages/microsoft.semantickernel.connectors.ai.openai/0.19.230804.2-preview/lib/netstandard2.0/Tokenizers/Settings/vocab.bpe'
> at Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers.Settings.EmbeddedResource.ReadFile(String fileName)
> at Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers.Settings.EmbeddedResource.ReadBytePairEncodingTable()
> at Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers.Settings.GPT3Settings.<>c.<.cctor>b__6_1()
> at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
> at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
> at System.Lazy`1.CreateValue()
> at Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers.Settings.GPT3Settings.get_BpeRanks()
> at Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers.GPT3Tokenizer.BytePairEncoding(String token)
> at Microsoft.SemanticKernel.Connectors.AI.OpenAI.Tokenizers.GPT3Tokenizer.Encode(String text)
> at Submission#2.<<Initialize>>d__0.MoveNext()
> --- End of stack trace from previous location ---
[...]

Issue #2334, opened 2023-08-05
.Net: cl100k_base tokenizer


anthonypuppo:
The existing [GPT3Tokenizer](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.AI.OpenAI/Tokenizers/GPT3Tokenizer.cs) class uses p50k_base encoding. GPT-3.5, GPT-4, and text-embedding-ada-002 all rely on cl100k_base encoding.

The below table is extracted from the [OpenAI notebook](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb) on token counting.

| Encoding name | OpenAI models |
| ------------- | ------------- |
| cl100k_base | gpt-4, gpt-3.5-turbo, text-embedding-ada-002 |
| p50k_base | Codex models, text-davinci-002, text-davinci-003 |
| r50k_base (or gpt2) | GPT-3 models like davinci |

anthonypuppo:
[SharpToken](https://github.com/dmitry-brazhenko/SharpToken)

nacharya1:
@dluc are you able to look into this?

PR #2809, opened 2023-09-13
.Net: Removed GPT3Tokenizer and related files

dmytrostruk:
### Motivation and Context

Closes https://github.com/microsoft/semantic-kernel/issues/2679
Closes https://github.com/microsoft/semantic-kernel/issues/2508
Closes https://github.com/microsoft/semantic-kernel/issues/2334

This PR contains changes to remove `GPT3Tokenizer` from Semantic Kernel repository.

Alternatives:
https://github.com/microsoft/Tokenizer
https://github.com/dmitry-brazhenko/sharptoken

### Description

1. Removed tokenizer logic, tests and related files.
2. Updated documentation.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example55_TextChunker.cs
```diff
@@ -68,10 +67,8 @@ private static void WriteParagraphsToConsole(List<string> paragraphs)
         }
     }
 
-    private static int TokenCounter(string input)
+    private static int CustomTokenCounter(string input)
     {
-        var tokens = GPT3Tokenizer.Encode(input);
-
-        return tokens.Count;
+        return input.Length / 4;
```
stephentoub, line 72: Isn't it a meaningful gap now that our samples for chunking can't do it based on actual token count?

dmytrostruk: I think it's just a syntax showcase how to use custom token counter. Since `tokenCounter` parameter is optional and can be `null`, in this case the default one will be used, and it has the same logic:
https://github.com/microsoft/semantic-kernel/blob/34456aa1b49486bba0569697116119121bcdfcfb/dotnet/src/SemanticKernel/Text/TextChunker.cs#L281-L284

stephentoub: My point is, developers writing these solutions need to be able to compute token counts, whether for chunking or limiting the size of a prompt. What are SK samples for this going to show now?

dmytrostruk: As far as I understand, this PR shouldn't impact already existing SK samples. 
For chunking, users can replace `input.Length / 4` logic with their own logic or use already existing library. We can add more comments if needed. We can showcase how to use some external tokenizer library, but these examples should be already provided by library itself. 

In terms of limiting the size of a prompt, we don't have such examples at the moment. But again, tokenizer libraries should contain such examples. 

Do you have any recommendations here?

RogerBarreto: @dmytrostruk May be worth using the SharpToken in our examples, Like we did for the Polly reliability scenarios.

dmytrostruk: My idea was to avoid using specific tokenizer library in examples, since there are multiple implementations, but if this is preferred approach, it works for me as well. I've added SharpToken in `TextChunker` example in this commit: https://github.com/microsoft/semantic-kernel/pull/2809/commits/7b1cc1901ebe36ac40e3be76bd6dc966318de5ce
Thanks for feedback!

RogerBarreto (approved):

shawncal (approved):

Issue #2814, opened 2023-09-13
Consolidate model support across all readmes and repositories


momuno:
Supported model tables are found in semantic-kernel and Microsoft docs, along with in each sample readme.  This makes maintainability difficult.  We need one source of truth that all samples, notebooks, and documentation can point to.

Issue #2015, opened 2023-07-17
SK and Model Compatibility


Escoto:
Hi!

Working with `Azure OpenAI` I had some troubles getting the samples to work (`..\samples\notebooks\dotnet\00-getting-started.ipynb`). I tried multiple Models and I could only make the samples work with the `Davinci Models`. Is it that the `SK` only works with Davinci? or is there a list of compatible Models I didn't find? 

If it is the case that `SK` only works with `Davinci Models`, can we add that as a requirement?

nacharya1:
@Escoto  thanks for letting us know, we are looking into updating the samples to reflect text completion models.

matthewbolanos:
@momuno, would you be able to update the requirements in the notebooks to make it clear which models are supported?

momuno:
Yes.  It looks right now that text-completion models are no longer available via Azure OpenAI.  Only in OpenAI.  The notebooks may need a bigger revamp than just a supported-model update.  

To address this specific issue, I can put a not in the samples about this and see if I can suggest a temporary workaround... like only use OpenAI?

nacharya1:
that works @momuno , @matthewbolanos any concerns?

momuno:
Addressing specific dotnet issue and azure openai models here: https://github.com/microsoft/semantic-kernel/pull/2815
More expansive model compatibility consolidation task created here: https://github.com/microsoft/semantic-kernel/issues/2814

markwallace-microsoft:
We have to move away from text completion models by the end of the year.

matthewbolanos:
Samples have been moved to chat completion models

PR #2815, opened 2023-09-13
.Net: New models readme and update notebook to reference

momuno:
### Motivation and Context

  1. Why is this change required? clarity needed on what Azure OpenAI models are supported by SK
  2. What problem does it solve? missing info on what models are supported in dotnet notebooks
  3. What scenario does it contribute to? running dotnet notebooks
  4. If it fixes an open issue, please link to the issue here. #2015 Since issue reported, models supported has changed (azure openai no longer has text-completion models), and notebook location has changed.

### Description

This addresses the specific issue encountered using dotnet notebooks and missing model support table.  A permanent home for supported models still needs to be created - a separate task for that is #2814 (Currently there are multiple repos and readmes that store this information, making maintainability difficult.)

Chat completion models were tested for both openai and azure openai for the dotnet notebook 00-getting-started.

### Contribution Checklist

### DIFF
--- a/dotnet/notebooks/config/MODELS.md
+++ b/dotnet/notebooks/config/MODELS.md
@@ -1,0 +1,27 @@
+# Dotnet Notebooks - Supported Models
+The tables below show which models are available and supported with the dotnet notebooks. Please select from these tables when configuring your service variables to run the notebooks.
+
+## Azure OpenAI
+
+| Service      | Model type      | Model name             | Model version | Supported |
+| ------------ | --------------- | ---------------------- | ------------- | --------- |
+| Azure OpenAI | Chat Completion | gpt-35-turbo           | 0301          | ✅         |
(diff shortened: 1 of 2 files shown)


dotnet/notebooks/config/MODELS.md
```diff
@@ -0,0 +1,27 @@
+# Dotnet Notebooks - Supported Models
+The tables below show which models are available and supported with the dotnet notebooks. Please select from these tables when configuring your service variables to run the notebooks.
```
glahaye, line 2: I'm wondering whether we want to maintain this table ourselves our just point to the documentation from OpenAI and Azure OpenAI?

https://platform.openai.com/docs/models/overview
https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models

gitri-ms:
This PR is no longer being worked on, can reopen at a later date if we want to revisit it.

Issue #2735, opened 2023-09-06
CompleteRequestSettings is insufficiently abstract


douglasware:
Note: I could also file this as a bug as many models do not work with SK and the Oobabooga and HuggingFace connectors because of this problem.

When I started using other completion APIs, e.g. Oobabooga, HuggingFace, the design around Microsoft.SemanticKernel.AI.TextCompletion.CompleteRequestSettings gets inconvenient quickly because behind those APIs the parameters vary, and the SK implementation is specifically shaped around OpenAI API. 

Seems like this one design element is a big blocker to the vision of a pluggable system.

A simple (but not terribly robust) solution would be to get rid of CompleteRequestSettings altogether and allow the use of a name/value collection instead.

sammiie:
To address the issue of CompleteRequestSettings being insufficiently abstract and not accommodating different completion APIs (such as Oobabooga and HuggingFace), you can implement a more flexible and extensible design using a name/value collection. Below is a structured explanation, along with a diagram, to illustrate how this can be achieved:

Issue Explanation:

The CompleteRequestSettings class in Microsoft.SemanticKernel.AI.TextCompletion is tightly coupled with the OpenAI API and does not provide a generic way to accommodate various completion APIs, causing inconvenience and a lack of flexibility.

Solution: Flexible Request Settings with Name/Value Collection

1. Introducing RequestSettings Collection:

Create a new RequestSettings collection that is based on a name/value pair structure. This collection will serve as a generic container for request parameters.
2. Decoupling Request Settings:

Modify the code to decouple request settings from a specific API. Instead of having a dedicated class for each API (e.g., OpenAIRequestSettings, OobaboogaRequestSettings, HuggingFaceRequestSettings), use the generic RequestSettings collection.
3. Dynamic Configuration:

Allow developers to dynamically configure request parameters by adding or modifying name/value pairs within the RequestSettings collection.
4. Adapters for Specific APIs:

Implement separate adapter classes for each API (e.g., OpenAIAdapter, OobaboogaAdapter, HuggingFaceAdapter). These adapters will interpret the generic RequestSettings and convert them into API-specific parameters.
5. Diagram Structure:

Create a diagram to illustrate the revised architecture with the RequestSettings collection and API-specific adapters.
Here's a simplified diagram structure to explain the concept:

sql

+---------------------------------------------------+
|                   Application Code                |
|                                                 |
|    +---------------+         +-----------------+  |
|    | Request       |         | API-Specific   |  |
[...]

PR #2819, opened 2023-09-14
.Net: Option #2 - Refactor to support generic LLM request settings (use object) 

markwallace-microsoft:
### Motivation and Context

Currently the SK core uses a model for LLM request settings which is OpenAI specific. This is a refactor to remove this assumption and to allow SK to be LLM agnostic.

Resolves: https://github.com/microsoft/semantic-kernel/issues/2735

### Description

1. Remove `ChatRequestSettings` and `CompleteRequestSettings` and use dynamic instead
1. Create default OpenAI request settings classes and code to create instance from an object. The object can be an instance of the expected request settings, a `JsonElement` or an anonymous type.
1. Fix up planner, unit tests and integration tests
1. Add a new Kernel syntax example showing all options to configure LLM request settings

### Contribution Checklist


markwallace-microsoft:
See: https://github.com/microsoft/semantic-kernel/pull/2829

PR #2824, opened 2023-09-14
.Net: Add support for partial XML and global functions in plan parser

### LINKED ISSUE #2502: .Net: Function created with CreateSemanticFunction() not used in planner
**Describe the bug**
I recently updated a console project I use for testing from SK preview 14 to SK preview 19. It broke a bunch of stuff, which I fixed, but I'm having a new problem now. I load a few skills from disk, then call kernel.CreateSemanticFunction() using the stock "write in the style of Shakespeare" prompt. Then I call the planner. It used to rewrite the output in the style of Shakespeare, but it no longer includes this step. The documentation I found here https://learn.microsoft.com/en-us/semantic-kernel/get-started/quick-start-guide/using-the-planner?tabs=Csharp shows essentially the same steps.  Is there a new step I'm missing?

**To Reproduce**
```
    //Load up the skills the planner will use
    var skillsDirectory = Path.Combine(Directory.GetCurrentDirectory(), "skills");
    kernel.ImportSemanticSkillFromDirectory(skillsDirectory, "SummarizeSkill");
    kernel.ImportSemanticSkillFromDirectory(skillsDirectory, "WriterSkill");

    //Define a skill on the fly
    string skPrompt = """
        {{$input}}

        Rewrite the above in the style of Shakespeare.
    """;

    kernel.CreateSemanticFunction(skPrompt, "shakespeare", "ShakespeareSkill", maxTokens: 2000, temperature: 0.2, topP: 0.5);

    var goal = "Tomorrow is Valentine's day. I need to come up with a few date ideas.\r\nShe likes Shakespeare so write using his style. E-mail these ideas to my significant other.";
    var plannerConfig = new SequentialPlannerConfig { RelevancyThreshold = 0.5 };
    var planner = new SequentialPlanner(kernel, plannerConfig);
    var plan = await planner.CreatePlanAsync(goal);     //This line connects to Azure

    var i = 1;
    foreach (var step in plan.Steps)
        Console.WriteLine($"Step {i++}: {step.SkillName} {step.Description}");
    var results = await kernel.RunAsync(plan); //This line connects to Azure
    Console.WriteLine(results);
```

The output is:
Step 1: WriterSkill Given a goal or topic description generate a list of ideas
Step 2: WriterSkill Generate an acronym for the given concept or phrase
Step 3: WriterSkill Given a request to generate an acronym from a string, generate an acronym and provide the acronym explanation.
Step 4: WriterSkill Turn a scenario into a short and entertaining poem.
Step 5: WriterSkill Summarize given text in two sentences or less
Step 6: WriterSkill Turn bullet points into an email to someone, using a polite tone

Dear My Significant Other,

I hope you're doing well. I wanted to suggest some activities that we can do together to make our relationship even more special.

We could take a romantic walk in the park, have a picnic in the backyard, cook a romantic dinner together, enjoy a romantic movie night, take a cooking class together, visit a local art gallery, go for a bike ride, enjoy a romantic spa day, or take a dance class together. We could even go stargazing for a truly magical experience.

I'm looking forward to spending quality time with you and creating memories that will last a lifetime.

Thanks,
Me

lemillermicrosoft:
- Added support for parsing partial XML by appending a closing tag if not present.
- Enabled plan creation with global functions by handling cases where skill name is not provided.
- Updated unit tests to cover these new scenarios.

Fixes #2502

### Contribution Checklist

### DIFF
--- a/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
+++ b/dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
@@ -89,6 +89,12 @@ internal static class SequentialPlanParser
             // '</plan>': Matches the literal string "</plan>", indicating the closing tag of the <plan> element.
             Regex planRegex = new(@"<plan\b[^>]*>(.*?)</plan>", RegexOptions.Singleline);
             Match match = planRegex.Match(xmlString);
+
+            if (!match.Success)
+            {
+                match = planRegex.Match($"{xmlString}</plan>"); // try again with a closing tag
+            }
+
             if (match.Success)
             {
                 string planXml = match.Value;
@@ -205,7 +211,7 @@ internal static class SequentialPlanParser
     private static void GetSkillFunctionNames(string skillFunctionName, out string skillName, out string functionName)
     {
         var skillFunctionNameParts = skillFunctionName.Split('.');
-        skillName = skillFunctionNameParts?.Length > 0 ? skillFunctionNameParts[0] : string.Empty;
+        skillName = skillFunctionNameParts?.Length > 1 ? skillFunctionNameParts[0] : string.Empty;
         functionName = skillFunctionNameParts?.Length > 1 ? skillFunctionNameParts[1] : skillFunctionName;
     }
 
--- a/dotnet/src/Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SequentialPlanParserTests.cs
+++ b/dotnet/src/Extensions/Extensions.UnitTests/Planning/SequentialPlanner/SequentialPlanParserTests.cs
@@ -66,7 +66,7 @@ public class SequentialPlanParserTests
         var functionsView = new FunctionsView();
         foreach (var (name, skillName, description, isSemantic, resultString) in functions)
         {
-            var functionView = new FunctionView(name, skillName, description, new List<ParameterView>(), isSemantic, true);
+            var functionView = new FunctionView(name, skillName, description, new List<ParameterView>() { new(name: "param", description: "description") }, isSemantic, true);
             var mockFunction = CreateMockFunction(functionView);
             functionsView.AddFunction(functionView);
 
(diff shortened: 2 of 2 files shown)


SergeyMenshykh (approved):

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
```diff
@@ -89,6 +89,12 @@ internal static Plan ToPlanFromXml(this string xmlString, string goal, Func<stri
             // '</plan>': Matches the literal string "</plan>", indicating the closing tag of the <plan> element.
             Regex planRegex = new(@"<plan\b[^>]*>(.*?)</plan>", RegexOptions.Singleline);
             Match match = planRegex.Match(xmlString);
+
+            if (!match.Success)
+            {
+                match = planRegex.Match($"{xmlString}</plan>"); // try again with a closing tag
```
dmytrostruk, line 95: Idea for future improvements: 
It's strange that closing tag was not included in user scenario, even with `text-davinci-003` model. I'm wondering, if token limit could be a reason. If yes, what if returned plan will contain closing tag, but truncated, like `</plan` or `</pla` (is it possible at all?). 
We can investigate in scope of separate PR if it's possible to handle parsing without `plan` tag at all. We can ask model to return just a list of functions and for parsing we can wrap it on our side, in the same way as we do with `<xml>` tag. In this case we will use less tokens and should handle more scenarios. If `function` tag will be truncated or not completed, then we definitely should throw an exception, because the plan is not completed.

dmytrostruk (approved):

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanParser.cs
lemillermicrosoft: >I'm wondering, if token limit could be a reason.

Could be but I doubt it. I think it's more likely it's trying to end with this where `<!-- End -->` is a stop sequence:
```
<!-- End -->
</plan>
```

> We can investigate in scope of separate PR if it's possible to handle parsing without plan tag at all. 
Agreed.

PR #2829, opened 2023-09-14
.Net: Option #3 - Refactor to support generic LLM request settings (use AIRequestSettings)

markwallace-microsoft:
### Motivation and Context

Currently the SK core uses a model for LLM request settings which is OpenAI specific. This is a refactor to remove this assumption and to allow SK to be LLM agnostic.

Resolves: https://github.com/microsoft/semantic-kernel/issues/2735

### Description

1. Remove `ChatRequestSettings` and `CompleteRequestSettings` and use new class `AIRequestSettings` instead
1. Create default OpenAI request settings class `OpenAIRequestSetings` which extends  `AIRequestSettings`.
1. Fix up planner, unit tests and integration tests
1. Add a new Kernel syntax example showing all options to configure LLM request settings

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettingsConverter.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI;
+
+/// <summary>
+/// JSON converter for <see cref="OpenAIRequestSettingsConverter"/>
+/// </summary>
+public class OpenAIRequestSettingsConverter : JsonConverter<OpenAIRequestSettings>
```
lemillermicrosoft, line 13: This helps -- we could do similar with ooga booga too, right?

markwallace-microsoft: Strictly we don't even need to include this for this option (it's here because I created options 2,3 from the option 1 branch). This is needed in option 1&2 to convert any anonymous class using pascal case e.g. If I set `requestSettings: new { MaxTokens = 128 }` in code. We want to use the pascal case because it's C# code. But for option 3 we cannot support that option so this converter can be removed.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
```diff
@@ -0,0 +1,160 @@
...
+            MaxDepth = 20,
+            AllowTrailingCommas = true,
+            PropertyNameCaseInsensitive = true,
+            ReadCommentHandling = JsonCommentHandling.Skip,
+        };
+
+        options.Converters.Add(new OpenAIRequestSettingsConverter());
+
+        return options;
+    }
```
RogerBarreto, line 157: Can be simplified.

```suggestion
    private static readonly JsonSerializerOptions s_options => new()
       {
          WriteIndented = true,
          MaxDepth = 20,
          AllowTrailingCommas = true,
          PropertyNameCaseInsensitive = true,
          ReadCommentHandling = JsonCommentHandling.Skip,
          Converters = { new OpenAIRequestSettingsConverter() }
       };
```

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        if (openAIRequestSettings is null)
+        {
+            openAIRequestSettings = new OpenAIRequestSettings()
+            {
+                MaxTokens = defaultMaxTokens
+            };
+        }
+
+        return openAIRequestSettings;
```
RogerBarreto, line 136: Using return first approach this can reduce complexity and size.

```suggestion
        if (requestSettings is OpenAIRequestSettings openAIRequestSettings)
        {
            return openAIRequestSettings;
        }
        
        var json = JsonSerializer.Serialize(requestSettings);
        var openAIRequestSettings = JsonSerializer.Deserialize<OpenAIRequestSettings>(json, s_options);
        
        if (openAIRequestSettings is not null) 
        {
            return openAIRequestSettings;
        }
        
        return new OpenAIRequestSettings()
        {
            MaxTokens = defaultMaxTokens
        };
```

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettingsConverter.cs
```diff
@@ -0,0 +1,107 @@
...
+                    case "PresencePenalty":
+                        requestSettings.PresencePenalty = reader.GetDouble();
+                        break;
+                    case "max_tokens":
+                    case "MaxTokens":
+                        requestSettings.MaxTokens = reader.GetInt32();
+                        break;
+                    case "stop_sequences":
+                    case "StopSequences":
+                        requestSettings.StopSequences = JsonSerializer.Deserialize<IList<string>>(ref reader, options) ?? Array.Empty<string>();
```
RogerBarreto, line 52: `ref` is already implicit

```suggestion
                        requestSettings.StopSequences = JsonSerializer.Deserialize<IList<string>>(reader, options) ?? Array.Empty<string>();
```

dotnet/src/InternalUtilities/src/System/DynamicUtils.cs
RogerBarreto, line 1: Is this file being used?

dotnet/src/InternalUtilities/src/System/DynamicUtils.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel;
+
+internal static class DynamicUtils
+{
+    internal static T? TryGetPropertyValue<T>(object? obj, string propertyName, T? defaultValue = default)
```
RogerBarreto, line 7: Suggest adding `OrDefault` in the end.

```suggestion
    internal static T? TryGetPropertyValueOrDefault<T>(object? obj, string propertyName, T? defaultValue = default)
```

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptTemplateConfig.cs
```diff
@@ -151,7 +83,7 @@ public class InputConfig
     [JsonPropertyName("completion")]
     [JsonPropertyOrder(4)]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public CompletionConfig Completion { get; set; } = new();
+    public AIRequestSettings? Completion { get; set; }
```
RogerBarreto, line 86: Suggest changing the property to `RequestSettings` or `Settings` as well (Abstract away the Completion concept)

RogerBarreto (commented):
Some suggestions.

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/PromptTemplateConfig.cs
markwallace-microsoft: That change is outside of the scope of this PR. The `PromptTemplateConfig` class needs a number of changes so they'll be made separately.

dotnet/src/InternalUtilities/src/System/DynamicUtils.cs
markwallace-microsoft: Class will be deleted

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettingsConverter.cs
markwallace-microsoft: Changing results in this error:

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1503	Argument 1: cannot convert from 'System.Text.Json.Utf8JsonReader' to 'System.Text.Json.JsonDocument'	Connectors.AI.OpenAI	C:\SemanticKernel\skrepo\semantic-kernel\dotnet\src\Connectors\Connectors.AI.OpenAI\OpenAIRequestSettingsConverter.cs	52	Active

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettingsConverter.cs
lemillermicrosoft: > For Oobabooga it has defined it's own AI request settings class and it just needs to set the appropriate JSON property name attributes and things will work.

Okay cool, then I'd propose that connector does the conversion from `StopSequences` to `stopping_sequences` .

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -58,9 +58,6 @@ public class StepwisePlanner : IStepwisePlanner
         // Load or use default PromptConfig
         this._promptConfig = this.Config.PromptUserConfig ?? LoadPromptConfigFromResource();
 
-        // Set MaxTokens for the prompt config
```
lemillermicrosoft, line 61: I still think we need to be doing this somehow. And also maybe an opportunity to help define something missing in the kernel. The ability to defin 1) how big a prompt is allowed to get before submission and 2) the amount of tokens you want generated. As this has been lacking, the implementation now just uses the same value for each with the assumption that the user configures it to be at most half the allowable token limit of the model being used. This could be done separately, too. Just getting the thoughts out.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/TextCompletionRequest.cs
```diff
@@ -173,5 +174,5 @@ public sealed class TextCompletionRequest
     /// In addition to the defaults. Written between "" and separated by commas. For instance: "\nYour Assistant:", "\nThe assistant:"
     /// </summary>
     [JsonPropertyName("stopping_strings")]
-    public List<string> StoppingStrings { get; set; } = new List<string>();
+    public IList<string> StoppingStrings { get; set; } = new List<string>();
```
lemillermicrosoft, line 177: Can we make it so this will also read from `stop_sequences` or is that the responsibility of the plugin developer | planner developer to do:

```
ExtensionData = new Dictionary<string, object>()
{
  { "Temperature", 0.0 },
  { "StopSequences", new[] { StopSequence } }, // to support openai
  { "StoppingStrings", new[] { StopSequence } }, // to support oogabooga
  { "MaxTokens", this.Config.MaxTokens ?? 1024 },
}
```

docs/decisions/0010-support-generic-llm-request-settings.md
RogerBarreto, line 1: Not sure if you need a whole new example for this. Might worth adding this to the `Example_06`

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
markwallace-microsoft: I'm going to add back in the code that will provide the equivalent functionality so we don't have a regression. 

I agree we do need to have well defined patterns to allow the appropriate default request settings to be provided. I am working on a plan for this and will have a PR + ADR in the next week or so.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/TextCompletionRequest.cs
markwallace-microsoft: One problem with making this the responsibility of the connector is that every connector needs to be aware of every other connector so that it can perform this conversion. And if a new AI and connector appears existing connectors will need an update.

The pattern you have proposed above is the short term solution and will work with the changes in this PR.

The medium term goal is to support the ability to configure requesting settings on a per model/provider basis. So I should be able to configure the Oobabooga and OpenAI settings separately when I create a Semantic Function.

docs/decisions/0010-support-generic-llm-request-settings.md
markwallace-microsoft: I'm going to keep it as a separate example as I want to extend it to demonstrate configuring multiple models soon.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettingsConverter.cs
markwallace-microsoft: We want to avoid one connector needing to know the settings another connector supports or all connectors needing to know how to adapt OpenAI specific request settings to the pattern they support. If OpenAI bring out a new request setting then only it's connector should need to change.

dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj
```diff
@@ -18,6 +18,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
     <PackageReference Include="Microsoft.Bcl.HashCode" />
+    <PackageReference Include="Microsoft.CSharp" />
```
lemillermicrosoft, line 21: Why is this needed?

lemillermicrosoft (approved):

dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj
markwallace-microsoft: Good catch, that was for a different option. Will remove this.

dotnet/src/Connectors/Connectors.AI.Oobabooga/TextCompletion/TextCompletionRequest.cs
RogerBarreto: I would not dictate this to be a `IList` we can use `ICollection`

RogerBarreto (commented):
Change suggestions

RogerBarreto (approved):
Thanks for the changes!

LGTM

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -59,7 +59,11 @@ public class StepwisePlanner : IStepwisePlanner
         this._promptConfig = this.Config.PromptUserConfig ?? LoadPromptConfigFromResource();
 
         // Set MaxTokens for the prompt config
-        this._promptConfig.Completion.MaxTokens = this.Config.MaxTokens;
+        if (this._promptConfig.Completion is null)
+        {
+            this._promptConfig.Completion = new AI.AIRequestSettings();
```
dmytrostruk, line 64: I think using `AI` namespace explicitly is not required here, we can add `using Microsoft.SemanticKernel.AI;` at the top and use just `AIRequestSettings`.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
```diff
@@ -92,7 +92,7 @@ public async Task CanExecutePlanWithPlanStepAsync()
 
         var mockFunction = new Mock<ISKFunction>();
         mockFunction.Setup(x => x.InvokeAsync(It.IsAny<SKContext>(), null, It.IsAny<CancellationToken>()))
-            .Callback<SKContext, CompleteRequestSettings, CancellationToken>((c, s, ct) =>
+            .Callback<SKContext, dynamic, CancellationToken>((c, s, ct) =>
```
dmytrostruk, line 95: In these similar mocks I see the usage of `dynamic`, `object` and `AIRequestSettings`. I'm wondering what the reason is, and would it be possible to use just `AIRequestSettings` everywhere?

dmytrostruk (commented):
Small comments, in general LGTM. 
We should mark this PR with `breaking change` label since a couple of public interface signatures were changed.

dotnet/samples/KernelSyntaxExamples/Example58_ConfigureRequestSettings.cs
```diff
@@ -0,0 +1,92 @@
...
+        }";
+        var templateConfig = JsonSerializer.Deserialize<PromptTemplateConfig>(configPayload);
+        var func = kernel.CreateSemanticFunction(prompt, config: templateConfig!, "HelloAI");
+
+        result = await kernel.RunAsync(func);
+        Console.WriteLine(result.Result);
+
+        /* OUTPUT (using gpt4):
+Hello! As an AI language model, I can help you with a variety of
+Hello! As an AI language model, I can help you with a variety of tasks, such as:
```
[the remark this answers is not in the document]
markwallace-microsoft: Yes, the AI is called multiple times. First with a smaller max tokens and this is why the response is truncated.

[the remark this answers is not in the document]
dmytrostruk: Got it. I was confused because it looks like there are two result outputs (line 46, line 68), but output in comments contains three different results (line 71, 72 and 84).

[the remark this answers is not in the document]
dmytrostruk: nit: Also, looks like first call uses 256 tokens while second call uses 60, so the second output should be smaller in theory.

docs/decisions/0010-support-generic-llm-request-settings.md
```diff
@@ -0,0 +1,237 @@
...
+* Semantic Kernel implementation and sample code should avoid specifying OpenAI specific request settings in code that is intended to be used with multiple AI services.
+* Semantic Kernel implementation and sample code must be clear if an implementation is intended to be OpenAI specific.
+
+## Considered Options
+
+* Use `dynamic` to pass request settings
+* Use `object` to pass request settings
+* Define a base class for AI request settings which all implementations must extend
+
+Note: Using generics was discounted during an earlier investigation which Dmytro conducted, for more details go [here](https://microsofteur-my.sharepoint.com/:w:/g/personal/dmytrostruk_microsoft_com/EYThaoKwtvJLkQ53hv7bg7sBsil-aAHMbIU5jCsiwxofhg?e=hUUGay).
```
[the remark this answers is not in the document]
markwallace-microsoft: Done

dotnet/samples/KernelSyntaxExamples/Example58_ConfigureRequestSettings.cs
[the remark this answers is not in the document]
markwallace-microsoft: Have fixed this

dmytrostruk (approved):

dotnet/samples/NCalcPlugins/LanguageCalculatorPlugin.cs
```diff
@@ -71,9 +73,15 @@ public LanguageCalculatorPlugin(IKernel kernel)
...
-            topP: 1);
+            requestSettings: new AIRequestSettings()
+            {
+                ExtensionData = new Dictionary<string, object>()
+                {
+                    { "MaxTokens", 256 },
+                    { "Temperature", 0.0 },
+                    { "TopP", 1 },
+                }
+            });
```
SergeyMenshykh, line 84: We don't know in advance which LLM will be used to run this plugin, which means that some of the hardcoded parameters might not be relevant to that particular LLM, or additional ones specific to ther LLM might be required.

Can we extend the `LanguageCalculatorPlugin` constructor with the `AIRequestSettings` parameter, allowing the plugin consumers to provide settings specific to LLMs the plugin runs against?

markwallace-microsoft: We currently have hardcoded OpenAI specific request settings in multiple places in the code and this issue still remains to be resolved. This PR does not provide a solution to this problem. I have a task in the back log to address this issue.

SergeyMenshykh (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
```diff
@@ -0,0 +1,165 @@
...
+/// Request settings for an OpenAI completion request.
+/// </summary>
+public class OpenAIRequestSettings : AIRequestSettings
+{
+    /// <summary>
+    /// Temperature controls the randomness of the completion.
+    /// The higher the temperature, the more random the completion.
+    /// </summary>
+    [JsonPropertyName("temperature")]
+    [JsonPropertyOrder(1)]
```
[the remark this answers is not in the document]
markwallace-microsoft: No, this was in the original code. I'll remove this in a follow up PR because it makes extending this class problematic. I've created a task so I don't forget to do this.

PR #2834, opened 2023-09-14
.Net: REST API components builders

SergeyMenshykh:
### Motivation and Context
This change is necessary to improve maintainability by shifting the responsibility for constructing various components of REST API operations, such as paths, query strings, and payloads, to dedicated classes. This prevents the accumulation of code in the `RestApiOperation` and `RestApiOperationRunner` classes, making them easier to extend and mantain.

### Description
1. A dedicated `QueryStringBuilder` class has been introduced for constructing query strings for REST API operations.
2. The `OperationComponentBuilderFactory` class has been added to create builders for various REST API operation components. Currently, it only creates the `QueryStringBuilder`, but in the future, it will also create other builders for paths, payloads, and more.
3. The `RestApiOperationRunner` is updated to create the `QueryStringBuilder` using the `OperationComponentBuilderFactory` and delegates the task of query string construction to this builder.
4. The functionality for constructing query strings has been relocated from the `RestApiOperation` class to the `QueryStringBuilder` class without any changes.

Note: This change is refactoring - restructuring code, without changing its original functionality. This change will be used for the following PR to support various serialization mechanisms for query string parameters.

### Contribution Checklist


dotnet/src/Skills/Skills.OpenAPI/Builders/Query/QueryStringBuilder.cs
```diff
@@ -0,0 +1,47 @@
...
+
+/// <summary>
+/// Represents a query string builder for REST API operations.
+/// </summary>
+internal class QueryStringBuilder : IQueryStringBuilder
+{
+    ///<inheritdoc/>
+    public string Build(RestApiOperation operation, IDictionary<string, string> arguments)
+    {
+        var queryStringSegments = new List<string>();
```
shawncal, line 19:
```
NameValueCollection queryString = System.Web.HttpUtility.ParseQueryString(string.Empty);
queryString.Add("key1", "value1");
queryString.Add("key2", "value2");

return queryString.ToString(); // Returns "key1=value1&key2=value2", all URL-encoded
```

dotnet/src/Skills/Skills.OpenAPI/Builders/Query/QueryStringBuilder.cs
```diff
@@ -0,0 +1,47 @@
...
+        var queryStringParameters = operation.Parameters.Where(p => p.Location == RestApiOperationParameterLocation.Query);
+
+        foreach (var parameter in queryStringParameters)
+        {
+            //Resolve argument for the parameter.
+            if (!arguments.TryGetValue(parameter.Name, out var argument))
+            {
+                argument = parameter.DefaultValue;
+            }
+
```
shawncal, line 30: Not sure if we need this, but per url laws, `argument` could be a collection, i.e.
`&colors=blue,red,green`
...which is why [Microsoft.AspNetCore.Http.IQueryCollection](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.iquerycollection) uses [StringValues](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringvalues) in `KeyValuePair<string, StringValues>`

dotnet/src/Skills/Skills.OpenAPI/Builders/OperationComponentBuilderFactory.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using Microsoft.SemanticKernel.Skills.OpenAPI.Builders.Query;
+
+namespace Microsoft.SemanticKernel.Skills.OpenAPI.Builders;
+
+/// <summary>
+/// Defines factory for creating builders used to construct various components(path, query string, payload, etc...) of REST API operations.
+/// </summary>
+internal class OperationComponentBuilderFactory : IOperationComponentBuilderFactory
```
shawncal, line 11: I feel like this could be overkill... URLs are pretty standard and .net already has good tools. Do we expect anyone will want to override?

dotnet/src/Skills/Skills.OpenAPI/Builders/Query/QueryStringBuilder.cs
SergeyMenshykh: It's refactoring; no functionality has changed; the code has just moved as it was:
* To simplify the code review of this PR.
* To establish the "builder" structure as preparation for the next PR, which will completely replace the code to support various query string serialization styles.
image

The HttpUtility.ParseQueryString will be used in the next PR.

dotnet/src/Skills/Skills.OpenAPI/Builders/Query/QueryStringBuilder.cs
SergeyMenshykh: Right, support for various query string serialization styles will be introduced in the next PR:
image

dotnet/src/Skills/Skills.OpenAPI/Builders/OperationComponentBuilderFactory.cs
SergeyMenshykh: It's a little bit broader than URL building. The plan is to have a builder for each REST API operation component:
- Server URL: It can be parameterized, and the arguments for the parameters need to be resolved.
- Resource path: It can have various serialization styles, and the arguments for the parameters need to be resolved.
- Query string: It can have various serialization styles, and the arguments for the parameters need to be resolved.
- Headers: They can be serialized differently, and the arguments for the parameters need to be resolved.
- Body: It can have various media types such as application/json, text/plain, form, etc. The body needs to be built dynamically in a way specific to the media type, and the arguments for the parameters need to be resolved.
- Cookie: It can be serialized differently, and the arguments for the parameters need to be resolved."

On top of that, it's likely that the serialization style/format may differ between OpenAPI document versions. Therefore, it is better have a good abstraction/structure that enables us to handle serialization-version permutations with ease, rather than accumulating all the code within the `RestApiOperation` or `RestApiOperationRunner` classes, which would clutter them with `if(param.Type == "Query" && param.Style == "Form" && doc.Version == "2.0"...)` code and make them unmaintainable over time.

More details about various OpenAPI pa
* https://swagger.io/docs/specification/describing-parameters/
* https://swagger.io/docs/specification/serialization/
* https://swagger.io/docs/specification/describing-request-body/

shawncal (approved):

dotnet/src/Skills/Skills.OpenAPI/Skills.OpenAPI.csproj
```diff
@@ -23,6 +23,7 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="SemanticKernel.Skills.UnitTests" />
+    <InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
```
lemillermicrosoft, line 26: nit
```suggestion
    <InternalsVisibleTo Include="DynamicProxyGenAssembly2" /> <!-- Moq -->
```

lemillermicrosoft (approved):

PR #2835, opened 2023-09-14
.Net: SequentialPlanner skills lookup optimization

shawncal:
### Description
Where SKContext is used only to look for functions, replacing with IReadOnlySkillCollection instead.
Passing CancellationToken through async functions.

### Contribution Checklist


dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -359,28 +358,24 @@ private ChatHistory CreateChatHistory(IKernel kernel, out IAIService aiService)
...
-    }
+    private Task<string> GetSystemMessageAsync(SKContext context, CancellationToken cancellationToken)
+        => this._promptRenderer.RenderAsync(this._promptTemplate, context, cancellationToken);
 
     #endregion setup helpers
 
     #region execution helpers
 
-    private Task<string> GetNextStepCompletion(List<SystemStep> stepsTaken, ChatHistory chatHistory, IAIService aiService, int startingMessageCount, CancellationToken token)
+    private Task<string> GetNextStepCompletionAsync(List<SystemStep> stepsTaken, ChatHistory chatHistory, IAIService aiService, int startingMessageCount, CancellationToken token)
```
lemillermicrosoft, line 378: Why aren't these kind of rules getting flagged any longer??? Concerned there's a big pile of stuff we're missing now.

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
```diff
@@ -57,16 +57,15 @@ public async Task<Plan> CreatePlanAsync(string goal, CancellationToken cancellat
             throw new SKException("The goal specified is empty");
         }
 
-        string relevantFunctionsManual = await this._context.GetFunctionsManualAsync(goal, this.Config, cancellationToken).ConfigureAwait(false);
-        this._context.Variables.Set("available_functions", relevantFunctionsManual);
+        string relevantFunctionsManual = await this._kernel.CreateNewContext().GetFunctionsManualAsync(goal, this.Config, cancellationToken).ConfigureAwait(false);
 
-        this._context.Variables.Update(goal);
+        ContextVariables vars = new(goal) { ["available_functions"] = relevantFunctionsManual };
```
lemillermicrosoft, line 62: 👀

lemillermicrosoft (approved):
Thanks!

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
```diff
@@ -57,16 +57,15 @@ public async Task<Plan> CreatePlanAsync(string goal, CancellationToken cancellat
...
 
-        var planResult = await this._functionFlowFunction.InvokeAsync(this._context, cancellationToken: cancellationToken).ConfigureAwait(false);
+        var planResult = await this._kernel.RunAsync(this._functionFlowFunction, vars, cancellationToken).ConfigureAwait(false);
```
lemillermicrosoft, line 64: Ahh, this is breaking some of the UTs that only mocked the InvokeAsync. You'll need to mock out `RunAsync` for this.

lemillermicrosoft (changes requested):

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
shawncal: Not sure. I'll check the rules.

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
shawncal: Not sure. I'll check the rules.

SergeyMenshykh: Most likely because there's no 'async' keyword in the method signature.

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -518,7 +513,7 @@ protected internal virtual SystemStep ParseResult(string input)
         return result;
     }
 
-    private async Task<string> InvokeActionAsync(string actionName, Dictionary<string, string> actionVariables)
+    private async Task<string> InvokeActionAsync(string actionName, Dictionary<string, string> actionVariables, CancellationToken cancellationToken)
     {
         var availableFunctions = await this.GetAvailableFunctionsAsync().ConfigureAwait(false);
```
SergeyMenshykh, line 518: The cancellation token can be passed to the `GetAvailableFunctionsAsync()` method.

SergeyMenshykh (approved):

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
```diff
@@ -57,16 +57,15 @@ public async Task<Plan> CreatePlanAsync(string goal, CancellationToken cancellat
             throw new SKException("The goal specified is empty");
         }
 
-        string relevantFunctionsManual = await this._context.GetFunctionsManualAsync(goal, this.Config, cancellationToken).ConfigureAwait(false);
-        this._context.Variables.Set("available_functions", relevantFunctionsManual);
+        string relevantFunctionsManual = await this._kernel.CreateNewContext().GetFunctionsManualAsync(goal, this.Config, cancellationToken).ConfigureAwait(false);
```
hario90, line 60: Not related to this PR but I notice that the `SequentialPlanner` is the only planner that uses an extension method off of `SKContext` to get the available functions whereas the others use private methods on the planner classes. Moving the shared logic for this would be another opportunity to improve consistency of behavior between planners (not to mention make it easier to maintain).

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
hario90: We should probably use a `const` for this key to make this less error-prone.

hario90 (approved):

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
shawncal: This is a temporary workaround... with Roger's change, we should no longer need CreateNewContext() for anything.

lemillermicrosoft (approved):

hario90:
* Are there plans to update Action Planner to read skills from the kernel?
* Why does SKContext have a Skill collection when the Kernel does as well?

lemillermicrosoft (approved):

PR #2842, opened 2023-09-15
.Net: Support function calling

### LINKED ISSUE #2315: Azure Open AI Functions calling integration into Kernel

gitri-ms:
### Motivation and Context

Enable the use of [OpenAI's function calling capability](https://platform.openai.com/docs/guides/gpt/function-calling) with SK. However, since this feature is limited to OpenAI chat completion models, we want to do so without widespread changes to the kernel.
 Fixes #2315 
### Description

- Add ability to pass functions to chat completion model via request settings. If the model does not support function calling, this will be ignored.
- Add OpenAIFunction class to store function information when passing to the model. This reduces coupling with OpenAI's FunctionDefinition, but adds an extra layer of translation.
- Add translation layer between FunctionView and OpenAIFunction to easily utilize functions already registered with the kernel
- Add example demonstrating how to include functions, parse response, and invoke chosen function if registered with the kernel
- Add required field to ParameterView (default false). This improves the chances of getting correct function arguments from OpenAI.

### Contribution Checklist

- [x] Design decisions captured in MADR


dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -353,9 +353,14 @@ private static ChatCompletionsOptions CreateChatCompletionsOptions(ChatRequestSe
             NucleusSamplingFactor = (float?)requestSettings.TopP,
             FrequencyPenalty = (float?)requestSettings.FrequencyPenalty,
             PresencePenalty = (float?)requestSettings.PresencePenalty,
-            ChoiceCount = requestSettings.ResultsPerPrompt
+            ChoiceCount = requestSettings.ResultsPerPrompt,
         };
 
+        if (requestSettings.Functions is not null)
+        {
+            options.Functions = requestSettings.Functions.ToFunctionDefinitions();
```
eavanvalkenburg, line 361: I don't see the FunctionCall option ([here](https://github.com/Azure/azure-sdk-for-net/blob/d8a846377185b5f4d2f30c10746f309ae879e127/sdk/openai/Azure.AI.OpenAI/src/Custom/ChatCompletionsOptions.cs#L68C61-L68C61))exposed, while I do think that that is important to steer the behaviour, with None, when you want to turn it off, or with a function name when you just want it to use that function.

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
```diff
@@ -0,0 +1,104 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Threading.Tasks;
+using Kusto.Cloud.Platform.Utils;
```
dmytrostruk, line 5: This namespace is used for `string.IsNotNullOrEmpty()` extension method in line 68, but I would avoid using extension methods from other libraries in this example and just use `!string.IsNullOrEmpty(chatMessage.Content)`. In this case it will be easier for users to copy this example with minimal changes required.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatRequestSettings.cs
```diff
@@ -60,6 +61,11 @@ public class ChatRequestSettings
     /// </summary>
     public IDictionary<int, int> TokenSelectionBiases { get; set; } = new Dictionary<int, int>();
 
+    /// <summary>
+    /// The set of functions to choose from if function calling is enabled by the model.
+    /// </summary>
+    public FunctionsView? Functions { get; set; } = null;
```
dmytrostruk, line 67: So, taking into account this approach, we have `ChatRequestSettings` from `Abstractions` package, which now contains `FunctionsView` from `Abstractions` as well.

There is refactoring work in parallel to remove `ChatRequestSettings` from `Abstractions`, because it's not really abstract settings, they are related to OpenAI and should be placed in OpenAI Connector. There is PR where it's already implemented and most probably that PR or something similar will be applied: https://github.com/microsoft/semantic-kernel/pull/2829.

Based on that, having `FunctionsView` model from `Abstractions` in OpenAI request settings probably is not the best option. Another approach would be to expose `FunctionDefinition` class from Azure SDK. But this also has drawbacks. For example, as soon as Azure SDK introduces some breaking changes in this class, this will result in breaking changes in SK. Also, if we decide to remove Azure SDK at some point (just as assumption), it will break SK users, who already rely on `FunctionDefinition` class. In short - it's high coupling to Azure SDK.

In my opinion, one of the solutions would be to introduce new class model, something like `OpenAIFunction` in OpenAI Connector, then map `FunctionView` to `OpenAIFunction` and pass it to `IChatCompletion.GetChatCompletionsAsync` as part of request settings. Then, inside OpenAI Connector we can map `OpenAIFunction` to Azure SDK `FunctionDefinition`. This approach introduces more mapping, but at the same time - less coupling.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/FunctionsViewExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
+
+internal static class FunctionsViewExtensions
+{
+    public static IList<FunctionDefinition> ToFunctionDefinitions(this FunctionsView functionsView)
+    {
+        var functionDefinitions = new List<FunctionDefinition>();
+
+        ConcurrentDictionary<string, List<FunctionView>> nativeFunctions = functionsView.NativeFunctions;
```
SergeyMenshykh, line 16: nit: no need to declare the variables if you can use `functionsView.NativeFunctions` and `functionsView.SemanticFunctions` directly in the foreach loops below.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/FunctionViewExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        var paramProperties = new Dictionary<string, object>();
+        foreach (var param in functionView.Parameters)
+        {
+            paramProperties.Add(
+                param.Name,
+                new
+                {
+                    type = param.Type?.Name ?? "string",
+                    description = param.Description ?? "",
```
SergeyMenshykh, line 25: nit: `string.Empty` instead of "" ?

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        // Check for message response
+        var chatMessage = await chatResult.GetChatMessageAsync();
+        if (chatMessage.Content.IsNotNullOrEmpty())
+        {
+            Console.WriteLine(chatMessage.Content);
+        }
+
+        // Check for function response
+        var functionCall = chatResult.ModelResult.GetResult<ChatModelResult>().Choice.Message.FunctionCall;
```
SergeyMenshykh, line 74: I would propose to start adding additional behavior to our "data"/DTO/model classes to avoid the [train wrecks](https://wiki.c2.com/?TrainWreck) - `var functionCall = chatResult.ModelResult.GetResult<ChatModelResult>().Choice.Message.FunctionCall`

An alternative/example could be `var functionCall = chatResult.GetFunctionCall()`.

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
```diff
@@ -0,0 +1,104 @@
...
+        if (functionCall is not null)
+        {
+            FunctionCallResponse functionResponse = FunctionCallResponse.FromFunctionCall(functionCall);
+
+            Console.WriteLine("Function name: " + functionResponse.FunctionName);
+            Console.WriteLine("Skill name: " + functionResponse.SkillName);
+            Console.WriteLine("Arguments: ");
+
+            var context = kernel.CreateNewContext();
+            if (context.Skills!.TryGetFunction(functionResponse.SkillName, functionResponse.FunctionName, out ISKFunction? func))
```
SergeyMenshykh, line 84: I believe that the logic for the AI function call should reside within SK (`SemanticFunction`?), so SK consumers do not need to implement it.

From the perspective of SK consumer code, it should be as simple as this:

1. Register/import a semantic function.
2. Register/import available native functions that might be called.
3. Execute the semantic function by calling `kernel.RunAsync`.
4. Retrieve the result returned by the `RunAsync` method.

There's no need to write any additional code. Everything (receiving the AI function response, finding/calling the function, calling AI again, receiving/parsing/returning the result) should be handled behind the scenes by SK.

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
gitri-ms: Oh, that is a mistake. I think that was Visual Studio trying (but failing) to be helpful. Will change!

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatRequestSettings.cs
gitri-ms: I was using FunctionsView because I didn't want to take a dependency on OpenAI in the Abstractions package. But if ChatRequestSettings is being moved into the connector, then I like your suggestion. I was worried that FunctionView would be too restrictive because it's really only useful to use with functions that have been imported to the kernel -- using something like OpenAIFunction or FunctionDefinition directly would give the developer more options for where their functions are sourced from.

The other thing to keep in mind is whether or not other model providers would eventually add support for function calling... but maybe that's something to handle once it actually happens 😄

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
gitri-ms: It felt a little redundant -- if a list of functions is provided, FunctionCall is auto by default. If no functions provided, FunctionCall is none by default. However, my solution does not (yet) handle the single function option. I was concerned about taking a direct dependency on the Azure Sdk (via FunctionDefinition). But with Dmytro's suggestion in another comment, if we add a new class model to contain function information, that would make it easier to support that option.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatRequestSettings.cs
dmytrostruk: > whether or not other model providers would eventually add support for function calling

Yes, so in this case we are considering base `RequestSettings` which will have common properties shared by model providers. If we see that pattern with function calling is common, we will move it to the level above in settings. But for now, we should probably keep it in `OpenAIChatRequestSettings` using something like `OpenAIFunction` property instead of `FunctionView`. But it looks like you will have to integrate it with PR shared above with settings refactoring.

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
gitri-ms: Yeah, I can add it to ChatResult since that lives in the OpenAI connector. I'm hesitate to add it to IChatResult since that is in the Abstractions package -- but then can cast the result similar to GetResult<T>

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
lemillermicrosoft: I suggested Gina go down this path as a first step and then we could build on as appropriate. SK treats this as ITextCompletion with regards to semantic functions today so a lot more changes would be needed.

lemillermicrosoft: I think the logic you described makes sense in the scope of something like ActionPlanner which could leverage this functionality when supported by the available AIService

gitri-ms: In the discussions I've had with Shawn and others, there has been hesitance to automatically have SK call a function returned by the model without some sort of approval. (Mostly security and cost concerns) By leaving that step to the app, it gives the app more control over validating the function call. This is also how we currently handle planning - SK returns the plan, it's up to the app to execute it. I think it would be best to maintain that consistency in behavior.

However, we probably could consider moving the

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIChatRequestSettings.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using Microsoft.SemanticKernel.AI.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
+
+// TODO: remove this temporary class
```
gitri-ms, line 8: This class will be removed once #2829 comes in. I just needed a request settings object in the connector to not mess up the whole dependency tree while testing.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -353,9 +353,19 @@ private static ChatCompletionsOptions CreateChatCompletionsOptions(ChatRequestSe
...
             FrequencyPenalty = (float?)requestSettings.FrequencyPenalty,
             PresencePenalty = (float?)requestSettings.PresencePenalty,
-            ChoiceCount = requestSettings.ResultsPerPrompt
+            ChoiceCount = requestSettings.ResultsPerPrompt,
         };
 
+        // TODO: temporary cast until request settings refactor comes in
+        OpenAIChatRequestSettings? oaiSettings = requestSettings as OpenAIChatRequestSettings;
+        if (oaiSettings is not null)
+        {
```
gitri-ms, line 362: Remove this temp code once #2829 comes in

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
```diff
@@ -0,0 +1,103 @@
...
+        // Retrieve available functions from the kernel and add to request settings
+        OpenAIChatRequestSettings requestSettings = new();
+        requestSettings.Functions = kernel.Skills.GetFunctionsView().ToOpenAIFunctions();
+
+        // Send request
+        var chatResult = (await chatCompletion.GetChatCompletionsAsync(chatHistory, requestSettings))[0];
+
+        // Check for message response
+        var chatMessage = await chatResult.GetChatMessageAsync();
+        if (string.IsNullOrEmpty(chatMessage.Content))
```
gitri-ms, line 67: This condition is backward

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
```diff
@@ -98,6 +99,18 @@ public string ChatSystemPrompt
     [JsonPropertyOrder(9)]
     public IDictionary<int, int> TokenSelectionBiases { get; set; } = new Dictionary<int, int>();
 
+    /// <summary>
+    /// Indicates a specific function that OpenAI should use to respond to the chat request.
+    /// If you want the model to choose from multiple functions, leave this null and
+    /// instead set <see cref="OpenAIRequestSettings.Functions"/>.
+    /// </summary>
+    public OpenAIFunction? FunctionCall { get; set; } = null;
```
gitri-ms, line 107: Not sure if these need the Json decorations and to be included in the converter. I don't imagine these being defined in prompts/text files but maybe good to keep the option open.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -0,0 +1,105 @@
...
+        OpenAIRequestSettings requestSettings = new()
+        {
+            Functions = kernel.Skills.GetFunctionViews().Select(f => f.ToOpenAIFunction()).ToList()
+        };
+
+        // Send request
+        var chatResult = (await chatCompletion.GetChatCompletionsAsync(chatHistory, requestSettings))[0];
+
+        // Check for message response
+        var chatMessage = await chatResult.GetChatMessageAsync();
```
lemillermicrosoft, line 70: re: feedback on helper out developers here. I've started a PR to add some extension methods, but I think we can do that as a follow up this week.

https://github.com/gitri-ms/semantic-kernel/pull/1

cc @shawncal

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
lemillermicrosoft: Here is an example of how this could be done as an extension: https://github.com/gitri-ms/semantic-kernel/pull/1

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
lemillermicrosoft: cc @markwallace-microsoft

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIFunctionResult.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Collections.Generic;
+using System.Text.Json;
+using Azure.AI.OpenAI;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
+
+/// <summary>
+/// Object containing function information and parameter values for a function call generated by the OpenAI model.
+/// </summary>
+public class OpenAIFunctionResult
```
gitri-ms, line 13: Change to OpenAIFunctionResponse to avoid confusion with FunctionResult (#2864)

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ParameterView.cs
```diff
@@ -56,16 +61,19 @@ public ParameterView()
...
     /// <param name="defaultValue">Default parameter value, if not provided</param>
     /// <param name="type">Parameter type.</param>
+    /// <param name="isRequired">Whether the parameter is required.</param>
     public ParameterView(
         string name,
         string? description = null,
         string? defaultValue = null,
-        ParameterViewType? type = null)
+        ParameterViewType? type = null,
+        bool isRequired = false)
```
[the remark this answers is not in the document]
lemillermicrosoft: Also should/could we apply this change in `NativeFunction.cs@507` to include `parameter.IsOptional`? Maybe this should be separate? cc @shawncal

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -386,6 +386,7 @@ async Task<SKContext> ExecuteAsync(SKContext context)
                 Description = $"{p.Description ?? p.Name}{(p.IsRequired ? " (required)" : string.Empty)}",
```
lemillermicrosoft, line 386: We might want to reconsider this now that we have `IsRequired` -- will also need to update planner manual formatting though to accommodate that. I'm fine not doing that in this PR though. Can you just make sure an issue is filed to track the follow up @gitri-ms when this merges and assign to me?

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ParameterView.cs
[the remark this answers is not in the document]
gitri-ms: Nullable bool seems weird to me... I figured if `isRequired` is not provided, it would be sufficient to assume it is false / treat it as not required. I don't think there would be any semantic difference between `isRequired = null` and `isRequired = false`.

I would recommend making the change to `NativeFunction.cs` as a separate PR, as I imagine it could change some functionality around native functions. I can open an issue.

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
gitri-ms: Yes will open an issue

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
gitri-ms: I've added the `FunctionCall` parameter to the `OpenAIRequestSettings`. So the behavior would be:
- **to use a specific function:** provide that function via `FunctionCall`, the model would choose that function and populate the parameters
- **for the `auto` option:** provide a list of functions via `Functions`, the model can optionally choose a function from this list
- **for the `none` option:** leave both `Functions` and `FunctionCall` empty/null

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/FunctionViewExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <returns>An <see cref="OpenAIFunction"/> object.</returns>
+    public static OpenAIFunction ToOpenAIFunction(this FunctionView functionView)
+    {
+        var openAIParams = new List<OpenAIFunctionParameter>();
+        foreach (ParameterView param in functionView.Parameters)
+        {
+            openAIParams.Add(new OpenAIFunctionParameter
+            {
+                Name = param.Name,
+                Description = param.Description ?? string.Empty,
```
lemillermicrosoft, line 26: Consider putting the default value here.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
gitri-ms: @eavanvalkenburg I'm looking at your Python PR to see how you handled this. It looks like you set `function_call` to either "auto", "none", or the name of a function. If the latter, you look for that function name in the list of `functions`. Is that right? I can mirror that behavior here by changing `FunctionCall` to be a string instead.

lemillermicrosoft (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -349,9 +349,34 @@ private static ChatCompletionsOptions CreateChatCompletionsOptions(OpenAIRequest
...
+            options.FunctionCall = FunctionDefinition.None;
+        }
+        else if (requestSettings.FunctionCall == OpenAIRequestSettings.FunctionCallAuto)
+        {
+            options.FunctionCall = FunctionDefinition.Auto;
+        }
+        else if (!requestSettings.FunctionCall.IsNullOrEmpty())
+        {
+            OpenAIFunction? function = requestSettings.Functions.Where(
+                f => f.FunctionName.Equals(requestSettings.FunctionCall, StringComparison.OrdinalIgnoreCase))
```
gitri-ms, line 366: Should search on full name (including plugin) not just function name

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
eavanvalkenburg: that is correct, the name should refer to the name of the function that should be called (where the model mostly supplies the parameters), and then I filter the list of functions so that we save tokens.

dotnet/samples/KernelSyntaxExamples/Example58_FunctionCalling.cs
gitri-ms: I spoke with @shawncal and we have decided not to bring in that extension at this time. For now, with this PR, it will be up to the app to invoke the function. We have not written off the possibility of adding auto-execution in a future update though. I am making a note in the ADR and will open an issue to track follow-up on this.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
markwallace-microsoft: I think it's OK to leave this out for the initial PR and then add it later as needed

dotnet/src/SemanticKernel.Core/Functions/NativeFunction.cs
```diff
@@ -517,7 +517,8 @@ Task<FunctionResult> Function(ITextCompletion? text, AIRequestSettings? requestS
             var parameterView = new ParameterView(
                 name,
                 parameter.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description ?? string.Empty,
-                defaultValue?.ToString() ?? string.Empty);
+                defaultValue?.ToString() ?? string.Empty,
```
lemillermicrosoft, line 520: Do we need to do anything on 377 re SKFunctionAttribute?

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -0,0 +1,102 @@
...
+        chatHistory.AddUserMessage(ask);
+
+        // Send request
+        var chatResult = (await chatCompletion.GetChatCompletionsAsync(chatHistory, requestSettings))[0];
+
+        // Check for message response
+        var chatMessage = await chatResult.GetChatMessageAsync();
+        if (!string.IsNullOrEmpty(chatMessage.Content))
+        {
+            Console.WriteLine(chatMessage.Content);
```
gitri-ms, line 73: Add message response to chat history

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        // Check for function response
+        OpenAIFunctionResponse? functionResponse = chatResult.GetFunctionResponse();
```
gitri-ms, line 77: Add function response to chat history (may need to add serializer?)

dotnet/src/SemanticKernel.Core/Functions/NativeFunction.cs
gitri-ms: SKParameterAttribute is being removed (#2991)

crickman (approved):

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -0,0 +1,111 @@
...
+            .Build();
+
+        // Load functions to kernel
+        kernel.ImportFunctions(new TimePlugin(), "TimePlugin");
+        await kernel.ImportPluginFunctionsAsync("KlarnaShoppingPlugin", new Uri("https://www.klarna.com/.well-known/ai-plugin.json"), new OpenApiFunctionExecutionParameters());
+
+        return kernel;
+    }
+
+    private static async Task CompleteChatWithFunctionsAsync(string ask, ChatHistory chatHistory, IChatCompletion chatCompletion, IKernel kernel, OpenAIRequestSettings requestSettings)
```
stephentoub, line 61: What's the plan for including a variant of this "in the box"?  It's boilerplate that lots of folks are going to need to write, otherwise.

lemillermicrosoft: Tracked by #2932. Here is one potential way we could do this, would love your feedback there @stephentoub https://github.com/gitri-ms/semantic-kernel/pull/1

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -0,0 +1,111 @@
...
+            // you can invoke it using the following code.
+            if (kernel.Functions.TryGetFunctionAndContext(functionResponse, out ISKFunction? func, out ContextVariables? context))
+            {
+                var result = (await kernel.RunAsync(func, context)).GetValue<string>();
+                if (!string.IsNullOrEmpty(result))
+                {
+                    Console.WriteLine(result);
+
+                    // Add the function result to chat history
+                    chatHistory.AddAssistantMessage(result);
```
lemillermicrosoft, line 102: This should be a 'function' message and then a *new* chat message for the user should be generated.

image

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
lemillermicrosoft: There's some issues with the SKChatMessage and MessageBase models that's actually not allowing the full scenario here. Need to figure out how to resolve that.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
gitri-ms: Supporting this requires modifications to our chat message abstractions, here's what it would look like: https://github.com/gitri-ms/semantic-kernel/pull/2. Tracked by #3017 / #1450

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
stephentoub: Sorry, I missed these responses somehow.

Yes, something along the lines of https://github.com/gitri-ms/semantic-kernel/pull/1.  I'm not seeing a need for it to return the ChatHistory, since it looks like it's always just returning the one that was passed in.

From my perspective, this support is critical and is _the_ value add SK can have here.  It also just "makes sense": SK already has the functions, and has the metadata for them, and that metadata is being used to populate the functions description, and it knows how to invoke its functions, and it knows how to convert the arguments supplied by the LLM to the right types for the function invocation, and it knows how to marshal the results back... it has everything it needs to make what is otherwise complicated and cumbersome boilerplate into a one-liner.

PR #2846, opened 2023-09-15
Adding IKernel property the SKContext  - Phase 1

RogerBarreto:
### Motivation and Context

⚠️ Breaking Change

Adding Kernel instance as part of the SKContext enable functions to interact with other services available in the kernel from within the functions, including the ability to call other Functions using the same Kernel "context".

This will also allow Kernel.RunAsync to execute and trigger events seamlessly when running steps and preparing the prompts (template engine) without resorting to implement those events for Plans and Template Engine separately.

This change is also a step forward in making IKernel.RunAsync the main way to Invoke functions.

### Description

This first step is only adding Kernel as a required SKContext property, 

PS: Other changes will start to harnessing the SKContext.Kernel reference to call Steps and Prompt rendering (thru prompt template engine) moving forward.

### Contribution Checklist

### DIFF
--- a/dotnet/samples/KernelSyntaxExamples/Example05_InlineFunctionDefinition.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example05_InlineFunctionDefinition.cs
@@ -50,15 +50,15 @@ Event: {{$input}}
 
         var excuseFunction = kernel.CreateSemanticFunction(FunctionDefinition, maxTokens: 100, temperature: 0.4, topP: 1);
 
-        var result = await excuseFunction.InvokeAsync("I missed the F1 final race");
+        var result = await excuseFunction.InvokeAsync("I missed the F1 final race", kernel);
         Console.WriteLine(result);
 
-        result = await excuseFunction.InvokeAsync("sorry I forgot your birthday");
+        result = await excuseFunction.InvokeAsync("sorry I forgot your birthday", kernel);
         Console.WriteLine(result);
 
         var fixedFunction = kernel.CreateSemanticFunction($"Translate this date {DateTimeOffset.Now:f} to French format", maxTokens: 100);
 
-        result = await fixedFunction.InvokeAsync();
+        result = await fixedFunction.InvokeAsync(kernel);
         Console.WriteLine(result);
     }
 }
--- a/dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
@@ -77,7 +77,7 @@ public static class Example07_BingAndGoogleSkills
 
         // Run
         var question = "What's the largest building in the world?";
-        var result = await kernel.Func(searchSkillId, "search").InvokeAsync(question);
+        var result = await kernel.Func(searchSkillId, "search").InvokeAsync(question, kernel);
 
         Console.WriteLine(question);
         Console.WriteLine($"----{searchSkillId}----");
--- a/dotnet/samples/NCalcSkills/LanguageCalculatorSkill.cs
+++ b/dotnet/samples/NCalcSkills/LanguageCalculatorSkill.cs
@@ -26,7 +26,7 @@ namespace NCalcSkills;
 public class LanguageCalculatorSkill
 {
     private readonly ISKFunction _mathTranslator;
-
+    private readonly IKernel _kernel;
     private const string MathTranslatorPrompt =
         @"Translate a math problem into a expression that can be executed using .net NCalc library. Use the output of running this code to answer the question.
 Available functions: Abs, Acos, Asin, Atan, Ceiling, Cos, Exp, Floor, IEEERemainder, Log, Log10, Max, Min, Pow, Round, Sign, Sin, Sqrt, Tan, and Truncate. in and if are also supported.
@@ -75,6 +75,8 @@ Question: {{ $input }}
             maxTokens: 256,
             temperature: 0.0,
             topP: 1);
+
+        this._kernel = kernel;
     }
 
     /// <summary>
@@ -93,7 +95,7 @@ Question: {{ $input }}
 
         try
         {
-            answer = await this._mathTranslator.InvokeAsync(input).ConfigureAwait(false);
+            answer = await this._mathTranslator.InvokeAsync(input, this._kernel).ConfigureAwait(false);
         }
         catch (Exception ex)
         {
--- a/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
@@ -18,16 +18,6 @@ namespace Microsoft.SemanticKernel.Orchestration;
 public sealed class SKContext
 {
     /// <summary>
-    /// The culture currently associated with this context.
-    /// </summary>
-    private CultureInfo _culture;
-
-    /// <summary>
-    /// Kernel instance reference for this context.
-    /// </summary>
-    private readonly IKernel _originalKernel;
-
-    /// <summary>
     /// Print the processed input, aka the current data after any processing occurred.
     /// </summary>
     /// <returns>Processed input, aka result</returns>
@@ -56,7 +46,7 @@ public sealed class SKContext
     /// <summary>
     /// Read only skills collection
     /// </summary>
-    public IReadOnlySkillCollection skills { get; }
+    public IReadOnlySkillCollection Skills { get; }
 
     /// <summary>
     /// App logger
@@ -96,7 +86,7 @@ public sealed class SKContext
         // 
         this._originalKernel = kernel;
         this.Variables = variables ?? new();
-        this.skills = skills ?? NullReadOnlySkillCollection.Instance;
+        this.Skills = skills ?? NullReadOnlySkillCollection.Instance;
         this.LoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
         this._culture = CultureInfo.CurrentCulture;
     }
@@ -116,7 +106,7 @@ public sealed class SKContext
     /// Constructor for the context.
     /// </summary>
     /// <param name="kernel">Kernel instance parameter</param>
-    /// <param name="skills">Context variables to include in context.</param>
+    /// <param name="skills">Skills to include in context.</param>
     public SKContext(
         IKernel kernel,
         IReadOnlySkillCollection? skills = null) : this(kernel, null, skills, kernel.LoggerFactory)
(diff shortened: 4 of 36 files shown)


SergeyMenshykh:
@RogerBarreto I would like to understand the rationale behind the change a little better. Did we have a technical discussion? If not, let's have one.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -57,23 +53,88 @@ public CultureInfo Culture
...
+
+        // 
+        this._originalKernel = kernel;
         this.Variables = variables ?? new();
         this.Skills = skills ?? NullReadOnlySkillCollection.Instance;
         this.LoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
         this._culture = CultureInfo.CurrentCulture;
     }
 
+    /// <summary>
```
shawncal, line 94: I plan to hide the public constructor (future change) so that only a Kernel creates an instance of SKContext -- a couple practical considerations to sort out first. 

These extra constructors shouldn't be needed (the first one covers all cases).

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
```diff
@@ -86,8 +85,7 @@ public async Task CanExecutePlanWithPlanStepAsync()
         // Arrange
         var kernel = new Mock<IKernel>();
 
-        var returnContext = new SKContext(
-            new ContextVariables(stepOutput)
+        var returnContext = new SKContext(kernel.Object, new ContextVariables(stepOutput)
         );
```
shawncal, line 89: `);` : no newline needed

dotnet/samples/KernelSyntaxExamples/Example05_InlineFunctionDefinition.cs
```diff
@@ -50,15 +50,15 @@ public static async Task RunAsync()
 
         var excuseFunction = kernel.CreateSemanticFunction(FunctionDefinition, maxTokens: 100, temperature: 0.4, topP: 1);
 
-        var result = await excuseFunction.InvokeAsync("I missed the F1 final race");
+        var result = await excuseFunction.InvokeAsync("I missed the F1 final race", kernel);
```
shawncal, line 53: Please convert these to Kernel.RunAsync calls

dotnet/samples/KernelSyntaxExamples/Example07_BingAndGoogleSkills.cs
```diff
@@ -77,7 +77,7 @@ private static async Task Example1Async(IKernel kernel, string searchSkillId)
 
         // Run
         var question = "What's the largest building in the world?";
-        var result = await kernel.Skills.GetFunction(searchSkillId, "search").InvokeAsync(question);
+        var result = await kernel.Skills.GetFunction(searchSkillId, "search").InvokeAsync(question, kernel);
```
shawncal, line 80:
```suggestion
        var func = kernel.Skills.GetFunction(searchSkillId, "search");
        var result = await kernel.RunAsync(func, new() { ["input"] = question });
```

RogerBarreto: Used `kernel.RunAsync(question, func);` signature, which will be an extension in the future.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -57,23 +53,88 @@ public CultureInfo Culture
...
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
     public SKContext(
+        IKernel kernel,
         ContextVariables? variables = null,
         IReadOnlySkillCollection? skills = null,
         ILoggerFactory? loggerFactory = null)
     {
+        Verify.NotNull(kernel, nameof(kernel));
+
+        //
```
markwallace-microsoft, line 86: Remove empty comment

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -57,23 +53,88 @@ public CultureInfo Culture
...
     /// Constructor for the context.
     /// </summary>
+    /// <param name="kernel">Kernel reference</param>
     /// <param name="variables">Context variables to include in context.</param>
     /// <param name="skills">Skills to include in context.</param>
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
     public SKContext(
+        IKernel kernel,
         ContextVariables? variables = null,
         IReadOnlySkillCollection? skills = null,
```
markwallace-microsoft, line 81: Should we eliminate the ability to do this and also use the skill collection from the kernel?

RogerBarreto: @markwallace-microsoft At first I considered it, but I didn't want to dictate the skills to be of the Kernel only as we had this interface expecting the parameter already there.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -57,23 +53,88 @@ public CultureInfo Culture
...
         ILoggerFactory? loggerFactory = null)
     {
+        Verify.NotNull(kernel, nameof(kernel));
+
+        // 
+        this._originalKernel = kernel;
         this.Variables = variables ?? new();
         this.Skills = skills ?? NullReadOnlySkillCollection.Instance;
         this.LoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
```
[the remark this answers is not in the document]
RogerBarreto: This makes sense, I will remove the ILoggerFactory as an argument and always use the Kernel.

dotnet/samples/NCalcSkills/LanguageCalculatorSkill.cs
```diff
@@ -75,6 +75,8 @@ public LanguageCalculatorSkill(IKernel kernel)
             maxTokens: 256,
             temperature: 0.0,
             topP: 1);
+
+        this._kernel = kernel;
```
lemillermicrosoft, line 79: In this scenario, we really don't want to expose these 'inner' semantic functions to the outer kernel nor use them on invocation below. Do we have work tracked for making it easier to create a kernel with the available AIServices?

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -252,7 +252,7 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
             var functionVariables = this.GetNextStepVariables(context.Variables, step);
 
             // Execute the step
-            var functionContext = new SKContext(functionVariables, context.Skills, context.LoggerFactory);
+            var functionContext = new SKContext(context.Kernel, functionVariables, context.Skills);
```
lemillermicrosoft, line 255: We don't have coverage for it right now, but curious about how this will integrate with the ability to do this
```csharp
// SequentialPlannerConfig.cs
public Func<string, string, ISKFunction?>? GetSkillFunction { get; set; 
```

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -183,10 +183,7 @@ public Task<SKContext> RunAsync(string input, CancellationToken cancellationToke
     /// <inheritdoc/>
     public async Task<SKContext> RunAsync(ContextVariables variables, CancellationToken cancellationToken, params ISKFunction[] pipeline)
     {
-        var context = new SKContext(
-            variables,
-            this._skillCollection,
-            this.LoggerFactory);
+        var context = new SKContext(this, variables);
```
lemillermicrosoft, line 186: If the SKContext will contain a kernel reference -- couldn't the implementations of InvokeAsync do the handling of the events instead of RunAsync?

lemillermicrosoft (approved):

dotnet/src/SemanticKernel/Kernel.cs
RogerBarreto: @lemillermicrosoft Originally that was my idea implemented in this PR #2045 but it was thrown away in favor of the Kernel because `ISKFunction` shouldn't have additional responsibilities on how to handle events.

dotnet/samples/NCalcSkills/LanguageCalculatorSkill.cs
RogerBarreto: you are right, I did it wrong here.

dotnet/src/SemanticKernel/Planning/Plan.cs
RogerBarreto: The path moving forward would be something like:

```
context.Kernel.RunAsync(GetSkillFunction(plugin, name))
```

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
```diff
@@ -92,48 +92,48 @@ public static ISKFunction UseFrequencyPenalty(this ISKFunction skFunction, doubl
...
         CancellationToken cancellationToken = default)
     {
-        var context = new SKContext(variables, skills, loggerFactory)
+        var context = new SKContext(kernel, variables, skills)
         {
             Culture = culture!
         };
 
-        return function.InvokeAsync(context, settings, cancellationToken);
+        return function.InvokeAsync(context, settings ?? function.RequestSettings, cancellationToken);
```
markwallace-microsoft, line 115: Do we want to mark `InvokeAsync` as obsolete and direct people toward using `RunAsync` now?

dotnet/samples/NCalcPlugins/LanguageCalculatorPlugin.cs
```diff
@@ -93,7 +92,7 @@ public LanguageCalculatorPlugin(IKernel kernel)
 
         try
         {
-            answer = await this._mathTranslator.InvokeAsync(input).ConfigureAwait(false);
+            answer = await this._mathTranslator.InvokeAsync(input, context.Kernel).ConfigureAwait(false);
```
markwallace-microsoft, line 95: Should we use `RunAsync` here?

dotnet/src/SemanticKernel/SkillDefinition/SKFunctionExtensions.cs
RogerBarreto: @markwallace-microsoft Not the scope of this PR yet. But we will as soon we can fully substitute all InvokeAsync cases.

dotnet/samples/NCalcPlugins/LanguageCalculatorPlugin.cs
RogerBarreto: Agree, made the change.

markwallace-microsoft (approved):
LGTM

dotnet/samples/NCalcPlugins/LanguageCalculatorPlugin.cs
```diff
@@ -93,7 +92,7 @@ public LanguageCalculatorPlugin(IKernel kernel)
 
         try
         {
-            answer = await this._mathTranslator.InvokeAsync(input).ConfigureAwait(false);
+            answer = await context.Kernel.RunAsync(input).ConfigureAwait(false);
```
lemillermicrosoft, line 95: Whoops, I think this was wrong @RogerBarreto

Issue #2519, opened 2023-08-21
Feature Suggestion: Consistent Interface for Planners and Configs


jorisdg:
I realize the Stepwise Planner is brand new and perhaps this is in the works already, but it would be nice to have the planners all implement the same interface... similarly, the configs can implement the same interface or inherit an (abstract?) base class or something.
```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/semantic-kernel/issues/2847
- [ ] https://github.com/microsoft/semantic-kernel/issues/2911
- [x] [ActionPlanner] Add support for memory filtering config properties
- [ ] https://github.com/microsoft/semantic-kernel/issues/2973
- [ ] https://github.com/microsoft/semantic-kernel/issues/2848
- [ ] https://github.com/microsoft/semantic-kernel/issues/1851
- [ ] https://github.com/microsoft/semantic-kernel/issues/2975
- [ ] https://github.com/microsoft/semantic-kernel/issues/2074
```

lemillermicrosoft:
We had considered this before, would love to hear more thoughts and reasoning. I particularly agree with the configuration being shared but can see some configurations just won't apply to other planners so doesn't make sense.

lemillermicrosoft:
#2802 has started to tackle a piece of this.

PR #2849, opened 2023-09-15
.Net: make planner constructors more similar

### LINKED ISSUE #2847: Planner Constructor normalization

hario90:
### Motivation and Context

Fixes #2847,  #2519

Contains breaking changes.

This addresses inconsistencies in planner construction. The planners sometimes took a prompt override through a delegate in the config and sometimes as a constructor argument. Additionally, the ActionPlanner used to take a loggerFactory as a constructor argument while the other planners did not. 

The goal of these changes is to make working with planners more predictable.

### Description

* Added `GetPromptTemplate` delegate function to the planner base config. All planners will use this to get the prompt override.
* Removed logger factory argument from ActionPlanner constructor. Used the Kernel's logger factory to create a new logger within each planner constructor.
* Renamed private `Config` private instance field to `_config` 

### Contribution Checklist


dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
```diff
@@ -24,25 +24,25 @@ public sealed class SequentialPlanner : ISequentialPlanner
...
+        // Set up prompt template
+        string promptTemplate = this._config.GetPromptTemplate?.Invoke() ?? EmbeddedResource.Read("skprompt.txt");
 
         this._functionFlowFunction = kernel.CreateSemanticFunction(
             promptTemplate: promptTemplate,
             skillName: RestrictedSkillName,
             description: "Given a request or command or goal generate a step by step plan to " +
                          "fulfill the request using functions. This ability is also known as decision making and function flow",
-            maxTokens: this.Config.MaxTokens ?? 1024,
+            maxTokens: this._config.MaxTokens ?? 1024,
```
lemillermicrosoft, line 45: This stuff is going to be changing with the `AIRequestSettings` work. Just a heads up.

lemillermicrosoft (approved):

shawncal:
@hario90 We've got some conflicts. Can you please resolve, then ping me for final review and merge?

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
```diff
@@ -94,7 +94,7 @@ public async Task<Plan> CreatePlanAsync(string goal, CancellationToken cancellat
         return plan;
     }
 
-    private SequentialPlannerConfig Config { get; }
+    private SequentialPlannerConfig _config { get; }
```
SergeyMenshykh, line 97: nit: The property was renamed according to the field's rules. Its name follows camel case, and it is prefixed with `_`. So, if the intention of this change is to convert the property into a class field, please remove the remaining getter.

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -682,7 +682,7 @@ private static string ToFullyQualifiedName(FunctionView function)
     /// <summary>
     /// The configuration for the StepwisePlanner
     /// </summary>
-    private StepwisePlannerConfig Config { get; }
+    private StepwisePlannerConfig _config { get; }
```
SergeyMenshykh, line 685: nit: The property was renamed according to the field's rules. Its name follows camel case, and it is prefixed with _. So, if the intention of this change is to convert the property into a class field, please remove the remaining getter.

hario90: I confused this with an instance field - I'll change this back.

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
hario90: fixed - thanks!

SergeyMenshykh (approved):

banduki:
Is `GetPromptTemplate()` a delegate to allow override of the create plan prompt? -- or if not, what is it used for?

PR #2852, opened 2023-09-16
.Net: Making ContextVariables into a proper Dictionary.

shawncal:
### Description
Simplifying ContextVariables by making it into an actual Dictionary<string, string> instead of a reimplemenation.

### Contribution Checklist

    - BREAKING CHANGE - use normal Dictionary methods instead. Behavior change: retrieving missing values via indexer [key] will throw if argument is missing, instead of returning string.Empty.


dotnet/src/IntegrationTests/TemplateLanguage/tests.txt
```diff
@@ -18,9 +18,6 @@
 {{}}
 {{}}
 
-.{{asis}}.
-..
-
```
markwallace-microsoft, line 23: Why is this change required?

dotnet/src/IntegrationTests/Fakes/EmailSkillFake.cs
```diff
@@ -11,7 +11,7 @@ internal sealed class EmailSkillFake
 {
     [SKFunction, Description("Given an email address and message body, send an email")]
     public Task<string> SendEmailAsync(
-        [Description("The body of the email message to send.")] string input,
+        [Description("The body of the email message to send.")] string input = "",
```
markwallace-microsoft, line 14: nit: Use `string.Empty`

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -21,15 +14,19 @@ namespace Microsoft.SemanticKernel.Orchestration;
...
     /// </summary>
     /// <param name="value">Optional value for the main variable of the context including trust information.</param>
     public ContextVariables(string? value = null)
+        : base(StringComparer.OrdinalIgnoreCase)
     {
-        this._variables[MainKey] = value ?? string.Empty;
+        if (value != null)
+        {
+            this[MainKey] = value;
+        }
```
markwallace-microsoft, line 29: `MainKey` no longer defaults to an empty string, is that intentional?

If yes we should tag this as a breaking change as some clients may depend on this behaviour.

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -59,30 +56,7 @@ public ContextVariables Clone()
...
-            // If requested, discard old data and keep only the new one.
-            if (!merge) { this._variables.Clear(); }
-
-            foreach (KeyValuePair<string, string> varData in newData._variables)
-            {
-                this._variables[varData.Key] = varData.Value;
-            }
-        }
-
+        this.Set(MainKey, value);
```
markwallace-microsoft, line 59: Allowing entry for `MainKey` to be set to null here but not in the constructor. Behaviour should be consistent.

dotnet/src/IntegrationTests/Fakes/EmailSkillFake.cs
RogerBarreto: As this is a parameter we can't use `string.Empty` only const values.

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
RogerBarreto: @markwallace-microsoft. According with the `Set` implementation, setting `null` will remove the variable.

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
RogerBarreto, line 29: Call `Set(MainKey, value)` to ensure consistency.

```suggestion
        this.Set(MainKey, value);
```

dotnet/src/SemanticKernel.UnitTests/Orchestration/ContextVariablesTests.cs
```diff
@@ -239,36 +239,6 @@ public void GetNameThatDoesNotExistReturnsFalse()
         Assert.Null(value);
     }
 
-    [Fact]
```
RogerBarreto, line 242: Keep the test. This should work for cloned Contexts as well.

RogerBarreto (changes requested):

hario90 (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -21,15 +14,19 @@ namespace Microsoft.SemanticKernel.Orchestration;
 /// </summary>
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 [DebuggerTypeProxy(typeof(ContextVariables.TypeProxy))]
-public sealed class ContextVariables : IDictionary<string, string>
+public sealed class ContextVariables : Dictionary<string, string>
```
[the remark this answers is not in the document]
shawncal: Possibly... let's work with object separately. We'll likely roll this in with ITypeConverter support

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
```diff
@@ -49,7 +46,7 @@ public ContextVariables Clone()
 
     /// <summary>Gets the main input string.</summary>
     /// <remarks>If the main input string was removed from the collection, an empty string will be returned.</remarks>
-    public string Input => this._variables.TryGetValue(MainKey, out string? value) ? value : string.Empty;
+    public string Input => this.TryGetValue(MainKey, out string? value) ? value : string.Empty;
```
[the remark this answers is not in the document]
shawncal: I agree. This just maintains current behavior, keeping this PR simple. I did try making `Input` nullable, and it results in a lot of changes to samples, tests.

This PR just seeks to just swap out the data structure with minimal breaking or functional changes.
We should decide between removing `Input` altogether (my preference), or keeping it and updating to nullable (and update callers). Either way, it's more readable if all of that is changed in another PR.

dotnet/src/IntegrationTests/Fakes/EmailSkillFake.cs
```diff
@@ -20,8 +20,8 @@ internal sealed class EmailSkillFake
 
     [SKFunction, Description("Lookup an email address for a person given a name")]
     public Task<string> GetEmailAddressAsync(
-        [Description("The name of the person to email.")] string input,
-        ILogger logger)
+        ILogger logger,
+        [Description("The name of the person to email.")] string input = "")
```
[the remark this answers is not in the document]
shawncal: To maintain the logic in the method:
```
        if (string.IsNullOrEmpty(input))
        {
            logger.LogTrace("Returning hard coded email for {0}", input);
            return Task.FromResult("johndoe1234@example.com");
        }
```
If we want to allow no value to be passed, then we have to give it a default, as the property mapping uses defaults indicate optional parameters. `null` would fail to resolve from the dictionary and throw.
I'll update it to `string? input = null;` so that it's still optional, but not automatically empty.

dotnet/src/IntegrationTests/TemplateLanguage/tests.txt
shawncal: The "asis" function has a required string input:
```
        [SKFunction, Description("This is a test"), SKName("asis")]
        public string MyFunction2(string input)
        {
            return input;
        }
```

Calling `asis` on its own is will crash because there's no input value provided. I can update asis to accept a nullable `string? input = null`

dotnet/src/SemanticKernel.UnitTests/Orchestration/ContextVariablesTests.cs
shawncal: The `Update(ContextVariables)` method that this is testing has been removed.

I'll update the test so that it checks a similar scenario, but calls cv.Clone() instead.

dotnet/src/Skills/Skills.UnitTests/MsGraph/CalendarSkillTests.cs
```diff
@@ -225,16 +225,13 @@ public async Task AddEventAsyncWithoutSubjectFailsAsync()
 
         CalendarSkill target = new(connectorMock.Object);
 
-        // Act
-        var ex = await Assert.ThrowsAsync<ArgumentException>(() => FunctionHelpers.CallViaKernel(target, "AddEvent",
+        // Act & Assert
+        var ex = await Assert.ThrowsAsync<SKException>(() => FunctionHelpers.CallViaKernel(target, "AddEvent",
```
lemillermicrosoft, line 229: nit: What exception is expected?

lemillermicrosoft (approved):

dotnet/src/IntegrationTests/TemplateLanguage/tests.txt
shawncal: undone

dotnet/src/SemanticKernel.Abstractions/Orchestration/ContextVariables.cs
shawncal: tagged as breaking

dotnet/src/Skills/Skills.UnitTests/MsGraph/CalendarSkillTests.cs
shawncal: Missing arg throws:
SKException($"Missing value for parameter '{name}'");

(NativeFunction.cs:483)

shawncal: Added an additional check.

lemillermicrosoft (approved):

RogerBarreto (approved):

lemillermicrosoft (approved):

Issue #2851, opened 2023-09-15
Python: PromptTemplateConfig.from_dict clobbers default CompletionConfig values


ascander:
**Describe the bug**

Initializing `PromptTemplateConfig` via the `from_dict` method clobbers default `CompletionConfig` values.

**To Reproduce**

Run the following snippet in a notebook or test:

```python
from semantic_kernel.semantic_functions.prompt_template_config import PromptTemplateConfig

base_completion_config = PromptTemplateConfig.CompletionConfig()
updated_completion_config = PromptTemplateConfig.from_dict({ "completion": { "max_tokens": 666 }}).completion

base_completion_config.temperature == updated_completion_config.temperature
```

**Expected behavior**

Default values (eg. `number_of_responses = 1`) for the completion config will be unchanged in the updated config.

**Actual behavior**

Default values are overwritten with `None`, which causes downstream failures.

**Platform**
 - OS: Mac
 - IDE: Neovim
 - Language: Python
 - Source: `main` branch of repository

**Additional context**

Found this while attempting to use `SequentialPlanner` which did not inherit the default value of `1` for the `completion.number_of_responses` setting.

I'm happy take on the implementation for a fix. I have a local fork that fixes this behavior by delegating to `dataclasses.replace` instead of explicitly overwriting completion setting keys. If that's a suitable fix, I'd like to add tests, etc. before submitting the PR.

alexchaomander:
Thank you for bringing this up! @mkarle can you take a look at this and @ascander's PR here https://github.com/microsoft/semantic-kernel/pull/2857

PR #2853, opened 2023-09-16
Python: Fixed prompt_template_config.from_dict method overwriting default CompletionConfig param values

AdarshAcharya5:
Fixes #2851 

### Description
dict.get('key') returns None if the key isn't present in the dict, Hence the config params were getting assigned None.

### Contribution Checklist

- [X] The code builds clean without any errors or warnings
- [X] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [X] All unit tests pass, and I have added new tests where possible
- [X] I didn't break anyone :smile:

PR #2857, opened 2023-09-18
Python: Fixes PromptTemplateConfig from_dict method clobbering default completion settings

ascander:
Fixes https://github.com/microsoft/semantic-kernel/issues/2851

### Motivation and Context

As reported in https://github.com/microsoft/semantic-kernel/issues/2851, the `from_dict` static method in `PromptTemplateConfig` clobbers default `CompletionConfig` values that aren't specified in the input parameter. For example, initializing the `SequentialPlanner` results in an invalid config, as `number_of_responses` is not specified in the planner config, and the default value of 1 is clobbered by loading it.

### Description

The approach here is to delegate to `dataclasses.replace` for updating the fields of the dataclass. The code here will properly default to the right values for `PromptTemplateConfig`, and raise a `TypeError` if `dict` contains keys that are not fields for the dataclass. I've added tests for this scenario, as well as tests for default completion settings.

### Contribution Checklist


mkarle (approved):

ascander:

ascander:
I don't think I'm blocking anymore; I needed to accept the CLA. What's needed to get this review finished?

ascander:
Obviated by https://github.com/microsoft/semantic-kernel/pull/2853

PR #2864, opened 2023-09-18
.Net: New result types - FunctionResult and KernelResult

dmytrostruk:
### Motivation and Context

This PR contains changes to replace `SKContext` as result type with new types `FunctionResult` and `KernelResult`.

1. `FunctionResult` - new return type from single function invocation. It contains `object? Value` which can be any primitive, complex type or `IAsyncEnumerable<T>`.  It also contains `SKContext`, but internally. This is required to pass `SKContext` to the next function in pipeline and keep the implementation simple at first iteration. Context could be removed from `FunctionResult` if needed in the future. It's not publicly available as part of `FunctionResult`.
2. `KernelResult` - new return type in `Kernel.RunAsync` method. It doesn't contain `SKContext`, just `object? Value` to get execution result.

`FunctionResult` also contains `Metadata` - property, that contains additional data, including AI model response (e.g. token usage).
`KernelResult` contains collection of `FunctionResult` - all function results from pipeline, so it's possible to check result of each function in pipeline, observe result value, AI-related information, like amount of tokens etc.

Syntax in examples and tests were changed from `result.Result` to `result.GetValue<string>`, but now it should be possible to get any type as a result and not only string. 

### Description

1. Updated `ISKFunction` interface to return `FunctionResult` instead of `SKContext`.
2. Updated `IKernel` interface to return `KernelResult` instead of `SKContext`. 
3. Updated `NativeFunction` to support `IAsyncEnumerable<T>` return type.
4. Added tests to `SKFunctionTests2` to verify the behavior for different result types.
5. Removed test `KernelTests.ItProvidesAccessToFunctionsViaSKContextAsync`. Reason: Invalid scenario. Kernel should not provide public access to functions via `SKContext`.
6. Updated tests and examples across the codebase to align with new signature.

### Contribution Checklist

    - BREAKING CHANGE: This PR will modify syntax of getting result using `ISKFunction.InvokeAsync` and `IKernel.RunAsync` methods. Syntax changed from `SKContext.Result` to `FunctionResult/KernelResult.GetValue<T>()`.


dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -0,0 +1,67 @@
...
+
+/// <summary>
+/// Function result after execution.
+/// </summary>
+public sealed class FunctionResult
+{
+    /// <summary>
+    /// Additional data from model.
+    /// </summary>
+    public IReadOnlyCollection<ModelResult> ModelResults { get; internal set; } = Array.Empty<ModelResult>();
```
lemillermicrosoft, line 16: Do we expect NativeFunctions to populate this?

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -256,7 +256,7 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
 
             var result = await step.InvokeAsync(functionContext, cancellationToken: cancellationToken).ConfigureAwait(false);
 
-            var resultValue = result.Result.Trim();
+            var resultValue = result.Context.Result.Trim();
```
lemillermicrosoft, line 259: I'm wondering if this should be result.Value instead.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunctionDelegate.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.AI.TextCompletion;
+using Microsoft.SemanticKernel.Orchestration;
+
+namespace Microsoft.SemanticKernel.SkillDefinition;
+
+internal delegate Task<FunctionResult> NativeFunctionDelegate(
```
lemillermicrosoft, line 10: I know this is existing -- but why do we need textCompletion and requestSettings here?

lemillermicrosoft (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Function result after execution.
+/// </summary>
+public sealed class FunctionResult
+{
+    /// <summary>
+    /// Additional data from model.
```
markwallace-microsoft, line 14: Can you expand on this description to explain when this would be populated. The original comment was a bit vague also but a description of how multiple model results get populated would be useful e.g., if there are nested semantic functions which one is first and last in the list.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -0,0 +1,67 @@
...
+        this.Value = value;
+        this.Context = context;
+    }
+
+    /// <summary>
+    /// Returns function result value.
+    /// </summary>
+    /// <typeparam name="T">Target type for result value casting.</typeparam>
+    /// <exception cref="InvalidCastException">Thrown when it's not possible to cast result value to <typeparamref name="T"/>.</exception>
+    public T? GetValue<T>()
```
markwallace-microsoft, line 53: Should this ne named `GetValueOrDefault` or should it return null if the `Value` is null?

dotnet/src/SemanticKernel.Abstractions/Orchestration/KernelResult.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Kernel result after execution.
+/// </summary>
+public sealed class KernelResult
```
markwallace-microsoft, line 11: Should `FunctionResult` and `KernelResult` shared a common base class?

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
dmytrostruk: Regarding naming, I would like to avoid long name for this method, as it should be widely used, so keeping it as short as possible will simplify the usage. 
In terms of returning null - it will return null in case `Value` is reference type, since `default` value for reference types is `null`. Also, we cannot return null in all cases, since `T` can be non-nullable value type, like `int` for example.

dotnet/src/SemanticKernel.Abstractions/Orchestration/KernelResult.cs
dmytrostruk: Initially, I tried this approach, but then I decided to remove base class, since it introduces some coupling between FunctionResult and KernelResult, which I would like to avoid. I think these types shouldn't be related to each other, even if they have some common properties or methods. 

For common methods like `GetValue` we can extract it to some static class and re-use in both classes if needed, without introducing base class. 
In terms of fields like `Value` or `ModelResults`, again, not sure that moving them to the base class will introduce significant benefits, but on the other hand, coupling may introduce some issues later.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
dmytrostruk: No, it is specifically for AI model additional data during semantic function execution, like token usage for example. However, since `ModelResult` also contains `object` property, we may consider making it more generic (e.g. `Dictionary<string, object> ExtensionData`, which will also allow native functions to put some additional metadata during function execution if required.

I would probably make this change as part of separate PR, to keep this one as lightweight as possible.

dotnet/src/SemanticKernel/Planning/Plan.cs
dmytrostruk: Good catch! I made it `result.GetValue<string>() ?? result.Context.Result` just for backward compatibility, but I tested manually, and `result.GetValue<string>()` works as expected. We can remove `result.Context.Result` in the future.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunctionDelegate.cs
dmytrostruk: I had the same question here :) It looks like we pass `null` as `ITextCompletion` parameter:
image

It looks like there was/is desire to keep similar signatures for Semantic and Native functions, but it can be definitely simplified in case of native function.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
dmytrostruk: Thanks, updated. There is not a lot of information to put there since this field is populated immediately after function execution. The reason why it's a collection is because one AI response can have multiple results. I described it in the documentation.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
RogerBarreto: @dmytrostruk, suggest to remove this property in thie PR and add it with the ExtensionData approach later to avoid introducing something that we will break right after.

dmytrostruk: I decided to add this in scope of this PR, the implementation will be ready soon :)

markwallace-microsoft (commented):
@dmytrostruk can you create an ADR for this change. This will be helpful for the community and the Java team.

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -208,9 +212,13 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
                     continue;
                 }
 
-                context = await skFunction.InvokeAsync(context, cancellationToken: cancellationToken).ConfigureAwait(false);
+                functionResult = await skFunction.InvokeAsync(context, cancellationToken: cancellationToken).ConfigureAwait(false);
+
+                context = functionResult.Context;
+
+                allFunctionResults.Add(functionResult);
```
SergeyMenshykh, line 219: I'm just wondering why we need the results of all function executions. Is this something that was requested, or is it just nice to have? Can we postpone introducing it until it's requested? It's always easier to add later than to remove.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -543,25 +543,33 @@ private static (Func<ITextCompletion?, CompleteRequestSettings?, SKContext, Canc
...
             {
                 await ((ValueTask)ThrowIfNullResult(result)).ConfigureAwait(false);
-                return context;
+                return new FunctionResult(context);
             };
         }
 
         // SKContext, either synchronous (SKContext) or asynchronous (Task<SKContext> / ValueTask<SKContext>).
 
         if (returnType == typeof(SKContext))
```
SergeyMenshykh, line 552: Should we have a similar `if` operator for 'FunctionResult'? or the case is already handled?

dotnet/src/SemanticKernel/SkillDefinition/NativeFunctionDelegate.cs
shawncal: Yeah, leave it for now (to avoid a lot more churn) and we'll fix up the remainder of ISKFunction in another PR. There are several semantic-only methods still in there.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
SergeyMenshykh: I would propose naming the property 'Metadata' if we decide to go with it so that it's clear that it contains data describing the result, while the result itself is represented by the 'value' property.

dotnet/src/SemanticKernel/Kernel.cs
dmytrostruk: I'm trying to resolve the case of passing information like token usage to the end user. Previously it was part of SKContext. In this PR, initially I've added `ModelResults` to KernelResult. But then it was not clear, either to use ModelResults from last function in pipeline or combine all ModelResults from all functions together. Having ModelResults from last function won't work, because if user will receive information that 50 tokens were used for function execution, it will be only for last function, not for all of them. (I think it works in this way at the moment, not sure it's correct though.) So, returning ModelResults from all functions sounds more reasonable, but then, why we should return just one `Value` as last function result but all ModelResults? So, I decided to return all FunctionResults, which will contain value for each function result separately, as well as ModelResults separately. It should provide more visibility to the users how each function performed.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
dmytrostruk: Good idea. In this case it will be possible for plugin developers to reuse `FunctionResult` as return type for their functions. Will add that, thanks. Most probably we will have to remove `if` for `SKContext`, as we shouldn't encourage developers to use SKContext as return type.

dotnet/samples/NCalcPlugins/LanguageCalculatorPlugin.cs
```diff
@@ -88,11 +88,12 @@ public LanguageCalculatorPlugin(IKernel kernel)
...
         SKContext context)
     {
-        SKContext answer;
+        string answer;
 
         try
         {
-            answer = await context.Kernel.RunAsync(input).ConfigureAwait(false);
+            var result = await context.Kernel.RunAsync(input).ConfigureAwait(false);
+            answer = result.GetValue<string>() ?? throw new InvalidOperationException("No result from kernel.");
```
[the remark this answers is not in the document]
lemillermicrosoft: I'd probably just default to `string.Empty` in this sample -- the rest of the code will handle this properly.

dotnet/src/Extensions/Planning.SequentialPlanner/SequentialPlanner.cs
```diff
@@ -62,9 +62,11 @@ public async Task<Plan> CreatePlanAsync(string goal, CancellationToken cancellat
 
         this._context.Variables.Update(goal);
 
-        var planResult = await this._functionFlowFunction.InvokeAsync(this._context, cancellationToken: cancellationToken).ConfigureAwait(false);
+        FunctionResult planResult = await this._functionFlowFunction.InvokeAsync(this._context, cancellationToken: cancellationToken).ConfigureAwait(false);
 
-        string planResultString = planResult.Result.Trim();
+        string planResultString =
+            planResult.GetValue<string>()?.Trim() ??
+            throw new SKException($"Plan was not generated successfully.\nGoal:{goal}\nFunctions:\n{relevantFunctionsManual}");
```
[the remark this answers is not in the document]
dmytrostruk: Thanks, added.

dotnet/src/Extensions/Planning.StepwisePlanner/StepwisePlanner.cs
```diff
@@ -534,11 +534,12 @@ private async Task<string> InvokeActionAsync(string actionName, Dictionary<strin
 
             var actionContext = this.CreateActionContext(actionVariables);
 
-            var result = await function.InvokeAsync(actionContext).ConfigureAwait(false);
+            var functionResult = await function.InvokeAsync(actionContext).ConfigureAwait(false);
+            var result = functionResult.GetValue<string>() ?? throw new InvalidOperationException($"No result from function: {targetFunction.SkillName}.{targetFunction.Name}");
```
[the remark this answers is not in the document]
dmytrostruk: Good catch! I've fixed that to be the same as previous implementation.

dotnet/samples/NCalcPlugins/LanguageCalculatorPlugin.cs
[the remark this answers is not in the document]
dmytrostruk: Thanks for feedback, done!

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -0,0 +1,67 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Function result after execution.
+/// </summary>
+public sealed class FunctionResult
```
gitri-ms, line 11: I am using `OpenAIFunctionResult` to indicate the function information returned by OpneAI when using the fuction calling feature. Do we need to better differentiate these with respect to naming?

dmytrostruk: @gitri-ms How about `OpenAIFunctionCallResult`?

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -22,12 +20,6 @@ public sealed class SKContext
     /// <returns>Processed input, aka result</returns>
     public string Result => this.Variables.ToString();
 
-    /// <summary>
-    /// When a prompt is processed, aka the current data after any model results processing occurred.
-    /// (One prompt can have multiple results).
-    /// </summary>
-    public IReadOnlyCollection<ModelResult> ModelResults { get; set; } = Array.Empty<ModelResult>();
```
shawncal, line 29: Can you please throw an [Obsolete] on this one, with instructions guiding users where to find the new version?

(We don't need to add obsoletes everywhere, but this is one I know some people will hit)

You can just throw NotImplementedException instead of get/set, as long as there are instructions telling callers the new way of accessing this.

shawncal (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
dmytrostruk: Yeah, I thought about `FunctionExecutionResult` and `KernelExecutionResult` as well. I decided to make the name shorter.
I also thought about `FunctionResponse` and `KernelResponse`. But `Response` looks more closer to the networking communication, while `Result` should work better for code execution. By the way, what if you call your class `OpenAIFunctionCallResponse`?  

> as opposed to a result containing info about a function that has not been executed

It shouldn't be the case. We should have a result only when function is executed. In case of error, we should throw an exception.

gitri-ms: That's a fair point about Result vs. Response. I was actually using Response on my end but then changed to Result to better match ChatResult. But I think I'll go change it back. Thanks for talking this through with me!

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -256,7 +256,10 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
 
             var result = await step.InvokeAsync(functionContext, cancellationToken: cancellationToken).ConfigureAwait(false);
```
lemillermicrosoft, line 257: It feels like this should also maintain a list of all function results similar to kernel.runasync. Thoughts?

dmytrostruk: This is one of the problems I see with `Plan` as implementation of `ISKFunction`. Because from the first sight it looks like it's a function that just contains sub-functions, but in some cases, we can see that it's a little bit different. In my opinion - `Function`, as entity, should return single result model (inside result we can have everything), but `Plan`, same as Kernel, should return final result (from last function in plan), but also it can return all function results for better observability and maintenance.

RogerBarreto:
```
functionContext.Kernel.RunAsync(step) ...

Comment (dmytrostruk) [reply dotnet/src/SemanticKernel/Planning/Plan.cs:255]: @RogerBarreto In this case there will be still only one result, because you can get multiple results from kernel only if you call multiple functions in pipeline - `kernel.RunAsync("input", func1, func2, func3)`.

Comment (RogerBarreto) [reply dotnet/src/SemanticKernel/Planning/Plan.cs:255]: Actually a Plan Execution will return multiple results.

Check latest @shawncal PR #2914 and see if this resonates.

File: dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs (hunk 13 above)
Comment (dmytrostruk) [reply dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs:29]: Thanks, done!

File: dotnet/src/SemanticKernel/Kernel.cs (hunk 7 above)
Comment (dmytrostruk) [reply dotnet/src/SemanticKernel/Kernel.cs:219]: Resolving based on private discussions.

File: dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs (hunk 1 above)
Comment (dmytrostruk) [reply dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs]: Thanks for all the feedback! I replaced `ModelResults` with `Metadata`.

File: dotnet/src/SemanticKernel/Planning/Plan.cs (hunk 14 above)
Comment (lemillermicrosoft) [reply dotnet/src/SemanticKernel/Planning/Plan.cs:255]: Should a semantic function not also return all the various results from the function the template engine might call?

Comment (dmytrostruk) [reply dotnet/src/SemanticKernel/Planning/Plan.cs:260]: As result `Value`, it will always contain result from itself. While `FunctionResults` may return results from each child function, and I think it should be useful (information like amount of tokens used, model response etc.)

**Review by lemillermicrosoft — APPROVED** [review]

**Review by SergeyMenshykh — APPROVED** [review]

**Review by SergeyMenshykh — APPROVED** [review]

**Comment by stephentoub** [comment]
Was just updating to use the new SK beta that has this.  Did we consider overriding ToString to make experimentation / prototyping with this a bit easier? e.g.
```C#
public override string ToString() => Value?.ToString() ?? base.ToString();
```
or something like that?

Related, what do we expect a consumer to do if they don't know the `T` that the function might return? The only API available to get back any information about it is `GetValue<T>`.

dmytrostruk:
> Did we consider overriding ToString to make experimentation / prototyping with this a bit easier?

@stephentoub I thought about that, but this is something that can be provided separately as extension method to `FunctionResult` (e.g. `ReadAsString()`) or implemented on users' side, if user mainly works with strings. I decided to avoid `ToString` override, because type of `Value` property will be forced to have meaningful `ToString` override as well. Otherwise, if it's custom type for example, we will return `typeof(T).FullName`, which won't provide additional benefits. In my opinion, `FunctionResult` should remain type-agnostic.


> what do we expect a consumer to do if they don't know the `T` that the function might return?

It depends on return type of last function in pipeline.

- If it's Semantic Function - return type will be always `string`.
- If it's Native Function - return type will be always the same as return type of invoked function.

Also, it's possible to get plain `.GetValue<object>` and then try to cast it to any potential return types and see if it's successful.
Another way would be to try to search for potential TypeConverter, and convert object to another known type, as we did in in StepwisePlanner, since in planning functionality we don't know about last function return type in generated plan:
https://github.com/microsoft/semantic-kernel/blob/7b46188f6f32eb8f29ff5f586ca54cd3f09bc191/dotnet/src/Planners/Planners.Core/Stepwise/StepwisePlanner.cs#L534-L542

If function pipeline is static (i.e., last function in pipeline will be always the same), and user doesn't know its return type, when user will try to cast it to invalid type, the exception will mention type of `Value` property which should be used for casting.

stephentoub:
> because type of Value property will be forced to have meaningful ToString override as well.

It won't be forced to. If it doesn't, you'll get its type name, which is no worse than getting the type name of "FunctionResult". And without either a ToString or a DebuggerDisplay attribute, it's harder to use in a debugger, eg you need to expand it in a watch to see anything useful.

Everywhere I previously just output the result of a function invocation, I now need to know about and use `.GetValue<object>()`. It's less condusive to quick tests and experimentation and prototyping.

As a point of reference, there are multiple container types in .NET that serve to just wrap some T with additional functionality, like Lazy, Nullable, and ThreadLocal, and that override ToString to delegate to the wrapped T's ToString. It's not intended to be something someone parses and consumes programmatically, but an aid for logging, debugging, etc.

dmytrostruk:
> Everywhere I previously just output the result of a function invocation, I now need to know about and use `.GetValue<object>()`. It's less condusive to quick tests and experimentation and prototyping.

I agree, but taking into account that we are on the way to `v1.0` and usage of SK in production environments, is it something that we want to deliver out of the box? We can add `DebuggerDisplay` attribute to simplify debugging process, but `ToString` override, as you said, it's something that is useful for tests and prototyping, but should not be used in production environments.

If you have some business scenario in production, you should know about the result type you expect to be returned from Kernel, so you can take it and handle it as needed.

stephentoub:
Are `Lazy<T>`, `Nullable<T>`, `ThreadLocal<T>`, every arity of `Tuple<...>` and `ValueTuple<...>`, etc. not meant for production environments?

dmytrostruk:
> Are `Lazy<T>`, `Nullable<T>`, `ThreadLocal<T>`, every arity of `Tuple<...>` and `ValueTuple<...>`, etc. not meant for production environments?

Sorry, I missed that part about these types and their `ToString` overrides. If this is common implementation, then I will add `ToString` override as well as `DebuggerDisplay` attribute and we will deliver it as part of next release. Thanks for the information!

PR #2878, opened 2023-09-19
.Net: Replacing `FunctionsView` class with List

### LINKED ISSUE #604: Potential concurrency bugs in FunctionsView

shawncal:
### Motivation and Context
The `FunctionsView` class wraps two inner Dictionaries of `FunctionView` objects: Native and Semantic. Since we no longer differentiate between these two for any practical purpose, we could merge those into one list, and we no longer need a special class for that.

### Description

- Removing the different collections to store Semantic vs Native functions
- Removing `FunctionsView` collection
- From `SkillCollection.GetFunctionsView()`, returning an `IReadOnlyList<FunctionView>`
- Callers can then easily use LINQ to filter the resulting flat list, or group by skill name.
- Updating callers, tests

Resolves #604

### Contribution Checklist

    - BREAKING CHANGE as it modifies the public API. Low impact.


lemillermicrosoft (approved):

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/IReadOnlySkillCollection.cs
```diff
@@ -48,8 +49,6 @@ public interface IReadOnlySkillCollection
     /// <summary>
     /// Get all registered functions details, minus the delegates
     /// </summary>
-    /// <param name="includeSemantic">Whether to include semantic functions in the list</param>
-    /// <param name="includeNative">Whether to include native functions in the list</param>
     /// <returns>An object containing all the functions details</returns>
-    FunctionsView GetFunctionsView(bool includeSemantic = true, bool includeNative = true);
+    IReadOnlyList<FunctionView> GetFunctionsView();
```
markwallace-microsoft, line 53: This should be named `GetFunctionViews` as it returns a list of `FunctionView` instances. Current name reflects the old implementation.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/NullReadOnlySkillCollection.cs
```diff
@@ -38,9 +39,9 @@ public bool TryGetFunction(string skillName, string functionName, [NotNullWhen(t
     }
 
     /// <inheritdoc/>
-    public FunctionsView GetFunctionsView(bool includeSemantic = true, bool includeNative = true)
+    public IReadOnlyList<FunctionView> GetFunctionsView()
```
markwallace-microsoft, line 42: As above this should be named `GetFunctionViews`

dotnet/src/SemanticKernel/SkillDefinition/SkillCollection.cs
```diff
@@ -85,21 +85,15 @@ public bool TryGetFunction(string skillName, string functionName, [NotNullWhen(t
     }
 
     /// <inheritdoc/>
-    public FunctionsView GetFunctionsView(bool includeSemantic = true, bool includeNative = true)
+    public IReadOnlyList<FunctionView> GetFunctionsView()
```
markwallace-microsoft, line 88: Name change required here also

SergeyMenshykh (approved):

RogerBarreto:
Resolves #604

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/IReadOnlySkillCollection.cs
shawncal: Done

markwallace-microsoft (approved):

PR #2881, opened 2023-09-19
.Net: Removing IsSemantic and IsAsync properties from ISKFunction and FunctionView

shawncal:
### Motivation and Context
The Kernel no longer differentiates between Native and Semantic functions, nor between async and synchronous functions. These are abstractions that are resolved internally.

### Description
ISKFunction
 - Removing IsSemantic
 - Updating tests and callers

FunctionView
 - Removing IsSemantic
 - Removing IsAsync
 - Updating tests and callers

### Contribution Checklist
   BREAKING CHANGE - due to change in public APIs. Low impact.


dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/IPromptTemplate.cs
```diff
@@ -17,7 +17,7 @@ public interface IPromptTemplate
     /// Get the list of parameters required by the template, using configuration and template info.
     /// </summary>
     /// <returns>List of parameters</returns>
-    IList<ParameterView> GetParameters();
+    IReadOnlyList<ParameterView> GetParameters();
```
markwallace-microsoft, line 20: In another PR we will have `GetFunctionViews`, we should have consistent naming so either use `GetParameterViews` here or `GetFunctions` in the other PR. I would prefer the shorter function names.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ParameterView.cs
```diff
@@ -1,75 +1,16 @@
...
-        ? this.Name
-        : $"{this.Name} ({this.Description})";
-}
+/// <param name="Name">Parameter name. The name must be alphanumeric (underscore is the only special char allowed).</param>
+/// <param name="Description">Parameter description</param>
+/// <param name="DefaultValue">Default parameter value, if not provided</param>
+/// <param name="Type">Parameter type.</param>
+public sealed record ParameterView(
+    string Name,
+    string? Description = null,
```
markwallace-microsoft, line 14: Do we want to allow this? If no description is provided can the function be used with a planner?

SergeyMenshykh (approved):

RogerBarreto (commented):
Suggestions.

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/IPromptTemplate.cs
shawncal: Actually, in this case, I think we can just replace with a property accessor. This list should not change, ever.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -131,14 +128,7 @@ internal sealed class NativeFunction : ISKFunction, IDisposable
...
     {
-        return new FunctionView
-        {
-            IsSemantic = this.IsSemantic,
-            Name = this.Name,
-            SkillName = this.SkillName,
-            Description = this.Description,
-            Parameters = this.Parameters,
-        };
+        return new FunctionView(this.Name, this.SkillName, this.Description) { Parameters = this.Parameters };
```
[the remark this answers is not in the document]
shawncal: Good question. I'll take a look. Might be easier to leave Describe() for this PR, then update that (and all callers) immediately after.

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/ParameterView.cs
shawncal: It's already "allowed", in that you can create a semantic or native function without adding a param definition. If the function and arg names are obvious enough, the model's fine with it. i.e. SavePhoneNumber(string phoneNumber)... it'll figure it out. I think it has to be optional, given the current architecture, and the fact that the [Description] attribute is optional. Alternative could be an empty string...

dotnet/src/SemanticKernel.Abstractions/SkillDefinition/FunctionView.cs
```diff
@@ -1,79 +1,25 @@
...
-public sealed class FunctionView
+/// <param name="Name">Name of the function. The name is used by the skill collection and in prompt templates e.g. {{skillName.functionName}}</param>
+/// <param name="SkillName">Name of the skill containing the function. The name is used by the skill collection and in prompt templates e.g. {{skillName.functionName}}</param>
+/// <param name="Description">Function description. The description is used in combination with embeddings when searching relevant functions.</param>
+/// <param name="Parameters">Optional list of function parameters</param>
+public sealed record FunctionView(
+    string Name,
+    string SkillName,
+    string Description = "",
+    IReadOnlyList<ParameterView>? Parameters = null)
```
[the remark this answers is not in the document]
shawncal: I'd rather keep *List because params are ordered and IReadOnlyCollection doesn't have an indexers. I've got an update here that can help.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
```diff
@@ -40,16 +40,13 @@ internal sealed class NativeFunction : ISKFunction, IDisposable
...
-    public bool IsSemantic { get; } = false;
-
     /// <inheritdoc/>
     public CompleteRequestSettings RequestSettings { get; } = new();
 
     /// <summary>
     /// List of function parameters
     /// </summary>
-    public IList<ParameterView> Parameters { get; }
+    public IReadOnlyList<ParameterView> Parameters { get; }
```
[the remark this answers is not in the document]
shawncal: See above, let me know if you agree.

dotnet/src/InternalUtilities/src/Diagnostics/Verify.cs
```diff
@@ -84,7 +84,7 @@ internal static void DirectoryExists(string path)
     /// Make sure every function parameter name is unique
     /// </summary>
     /// <param name="parameters">List of parameters</param>
-    internal static void ParametersUniqueness(IList<ParameterView> parameters)
+    internal static void ParametersUniqueness(IReadOnlyList<ParameterView> parameters)
```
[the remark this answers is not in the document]
shawncal: Just tried applying this suggestion, thanks!. Using a HashSet will enforce the distinctness of the args, but it will not enforce arg order (collection vs list), and it will silently drop duplicates - did it drop arg 2 or arg 4?

For the "View" structures, I think these should be as simple as possible, and we need not enforce any laws in these classes. Instead, they should be enforced in the classes that manage these registrations (upon registration) and throw on invalid data.

dotnet/src/SemanticKernel/SkillDefinition/NativeFunction.cs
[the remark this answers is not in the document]
shawncal: Updated to use cached/lazy init. Will update to a property (+ callers) in next PR.

dotnet/src/SemanticKernel.Abstractions/SemanticFunctions/IPromptTemplate.cs
shawncal: Done!

dotnet/src/InternalUtilities/src/Diagnostics/Verify.cs
```diff
@@ -84,7 +84,7 @@ internal static void DirectoryExists(string path)
...
     {
         int count = parameters.Count;
```
RogerBarreto, line 89: I don't see a need to use index in here + This method is specific to ParameterView, might worth moving it as an internal impl of it used in the `ctor`

RogerBarreto (approved):
LGTM

dotnet/src/InternalUtilities/src/Diagnostics/Verify.cs
shawncal: Gotcha. Room for improvement in this function. I've filed #2908 to clean up this function and where it's implemented/called.

markwallace-microsoft (approved):
LGTM

PR #2884, opened 2023-09-19
Python: updates sample prompts

videlalvaro:
Updates the prompts for bing and google search examples. 

In the Bing example, the result has `M eryl Streep` with an actual space between the M and the rest of the name.


moonbox3:
Hi @videlalvaro, thank you for your contribution. As this PR has been stale for many months, I am closing. If you feel that the kernel examples still need to be updated, please re-create the PR against the latest in main. Thank you.

Issue #2825, opened 2023-09-14
.Net: Function to Function calling


nacharya1:
```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/semantic-kernel/pull/2903
- [x] Update template engine to use RunAsync
- [ ] Remaining InvokeAsync clean
```

nacharya1:
Cross-function discoverability, dependency, access restrictions

RogerBarreto:
Partially resolved by #2903

Issue #2324, opened 2023-08-04
.Net: Control logic in pre/post- hooks: Phase 2


nacharya1:
> [!IMPORTANT]  
> Labeled **Urgent** because it will require a breaking change.

Support events for Plans and Prompt Rendering

Pre-Invocation event:
  - See trace of what functions came before (TBD) 
    i.e. for TrustService 
 
  - Semantic functions:
    - See model/endpoint to be used

  - Actions:
    - Modify prompt (TBD)
 
In-Invocation event: (TBD)
  - Get output StreamReader

Post-Invocation event:
  - Get state succeded / faulted + details

  - Actions
    - Stop execution
    - Repeat step
    - Set next step/branch in the plan

```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/semantic-kernel/pull/2903
- [x] Change Prompt Template Engine impl to use Executor.ExecuteAsync()
- [ ] Change Plans to use Executor.ExecuteAsync()
- [ ] Update PrePost hooks examples and implementations to suffice this Issue completely
```

RogerBarreto:
This one will be partially addressed by the #1714 and the existing PR #2045.

RogerBarreto:
Update the description only with Features related to Phase 2.

RogerBarreto:
This one will be partially addressed by the #2846

markwallace-microsoft:
Removing from V1 RC1 as this cannot be contained

PR #2903, opened 2023-09-20
.Net Updates to SKContext.KernelContext Phase 2

RogerBarreto:
### Motivation and Context

⚠️ Breaking Change

Adding an Executor in the SKContext enable functions the ability to call other Functions using the same Kernel "context" as described in #2825 

This will also allow Kernel.RunAsync to execute and trigger events (Resolves Partially #2324) seamlessly when running plans, plans steps and rendering prompts from templates (template engine) without resorting to implement those events for Plans and Template Engine separately.

This change is also a step forward in making IKernel.RunAsync the main way to Invoke functions.

### Description

This second phase, adds more changes to SKContext as follows:

- SKContext.Executor is now behind a new interface `IFunctionExecutor` exposed only to functions listed as part of the execution context.
- Calls to `IFunctionExecutor.ExecuteAsync` will run in a volatile Kernel where its usage won't affect the invoking kernel.
- Changed IKernel.CreateNewContext to allow optionally passing `ContextVariables` and `IReadOnlyFunctionCollection` (supporting most of the `new SKContext` prior scenarios)
- Added possibility of Cloning `SKContext` specifying both `ContextVariables` and `IReadOnlyFunctionCollection` as overrids.
- `SKContext` constructor was changed to internal and the only way now to instantiate `SKContext` is using `IKernel.CreateNewContext(...)` or cloning it.

### Contribution Checklist


dotnet/src/SemanticKernel/Orchestration/KernelContext.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        return this._kernel.RunAsync(variables, cancellationToken, pipeline);
+    }
+
+    public void Dispose()
+    {
+        this._kernel.Dispose();
+    }
+
+    private SkillCollection GetEditableSkillCollection(IReadOnlySkillCollection readOnlyCollection)
```
shawncal, line 49:
```suggestion
    private static SkillCollection GetEditableSkillCollection(IReadOnlySkillCollection readOnlyCollection)
```

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelContext.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// <returns>SK context</returns>
+    SKContext CreateNewContext(
+        ContextVariables? variables = null,
+        IReadOnlySkillCollection? skills = null);
+}
+
+/// <summary>
+/// Extension methods for <see cref="IKernelContext"/>
+/// </summary>
+public static class KernelContextExtensions
```
shawncal, line 51: Since a IKernel : IKernelContext, these extensions should also work for (and could replace) the similar methods inside of IKernel, right?

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelContext.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.SemanticKernel.SkillDefinition;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Kernel context reference
+/// </summary>
+public interface IKernelContext
```
markwallace-microsoft, line 13: This name is pretty generic. I'd recommend using `IKernelExecutionContext` to make it clear the intended use. 

I'd also recommend adding a better summary which explains the lifecycle.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -45,17 +45,17 @@ public CultureInfo Culture
...
     /// App logger
     /// </summary>
-    public ILoggerFactory LoggerFactory { get; }
+    public ILoggerFactory LoggerFactory => this.KernelContext.LoggerFactory;
 
     /// <summary>
     /// Kernel context reference
     /// </summary>
-    public IKernel Kernel => this.GetKernelContext();
+    public IKernelContext KernelContext => this.GetKernelContext();
```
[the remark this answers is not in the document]
markwallace-microsoft: I'd suggest:

`public IKernelExecutionContext ExecutionContext => this.GetKernelExecutionContext();`

`SKContext` i.e. "SemanticKernel Context" and `IKernelContext` is confusing. Also `SKContext` is vague.

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelContext.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// Run a pipeline composed of synchronous and asynchronous functions.
+    /// </summary>
+    /// <param name="variables">Input to process</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <param name="pipeline">Target functions to run in sequence</param>
+    /// <returns>Result of the function composition</returns>
+    Task<SKContext> RunAsync(
+        ContextVariables variables,
+        CancellationToken cancellationToken = default,
+        params ISKFunction[] pipeline);
```
markwallace-microsoft, line 35: Should we make the `cancellationToken` the last parameter in this new interface?

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelContext.cs
shawncal: Discussed offline. Resolving.

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelContext.cs
shawncal: The `params` has to be last... one of the reasons we're not a fan of this pipeline syntax. We'd like to replace this with a Plan-based approach and remove these overloads, but we have a few pre-reqs to get there. Roger and I discussed.

This is included for consistency with the (current) Kernel, but they'll all be replaced.

@RogerBarreto If you need this for something that exists in the codebase today, it's fine to keep it here to unblock. If you're adding this just for consistency, with Kernel and we have no callers in the repo, we should avoid adding it now (...and replacing it later).

dotnet/src/SemanticKernel/Orchestration/KernelContext.cs
RogerBarreto: I liked the alternative approach. Thanks

dotnet/src/SemanticKernel/SemanticKernel.csproj
```diff
@@ -23,6 +23,7 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="SemanticKernel.UnitTests" />
+    <InternalsVisibleTo Include="SemanticKernel.Extensions.UnitTests" />
```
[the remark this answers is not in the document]
RogerBarreto: Needed in the Plan UnitTests, many of those use SKContext mocking

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -252,7 +249,7 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
             var functionVariables = this.GetNextStepVariables(context.Variables, step);
 
             // Execute the step
-            var functionContext = new SKContext(context.Kernel, functionVariables, context.Skills);
+            var functionContext = context.KernelContext.CreateNewContext(functionVariables, context.Skills);
```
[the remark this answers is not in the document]
RogerBarreto: I think this make sense also because I don't change a property again after introducing it.

Additionaly (Thoughts):
SKContext in my opinion could just be `ExecutionContext` (mixing some of Mark's idea), so anything in this class is understood as being in the `Context`.

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelContext.cs
RogerBarreto: @markwallace-microsoft @shawncal 
I started this way TBH but I got lost trying to correct the IKernel to follow the same pattern and undid it.

Will give it another go.

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelContext.cs
RogerBarreto: Agree, changed the both implementation and interface names to `IKernelExecutionContext`

dotnet/src/SemanticKernel/Planning/Plan.cs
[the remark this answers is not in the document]
dmytrostruk: I would like to understand, what is general and long-term plan of this feature. Syntax `context.KernelContext.CreateNewContext` looks like this is moving to the wrong direction :)
I still think that we should use Context only for function input/output parameter binding, but not for executing functions from other functions.

It would be nice if:
- Users will invoke functions only through `kernel.RunAsync` - as original goal of this PR.
- Functions can execute other functions directly or there will be some orchestrator which will scan function dependencies, build execution graph and execute all functions in required order.
- SKContext is created only once at the beginning of Kernel execution and passed through all functions as same instance. Functions won't be allowed to create new context by themselves.

This is how dependencies looked previously to call function from other function: Kernel -> Plugin -> Kernel.
This is how it looks now: Kernel -> Plugin -> SKContext -> Kernel.

In my opinion, the dependencies should look like: Kernel -> Plugin1 -> Plugin2 -> Plugin3.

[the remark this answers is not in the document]
RogerBarreto: > SKContext is created only once at the beginning of Kernel execution and passed through all functions as same instance. Functions won't be allowed to create new context by themselves.

Agree here, removed this method and updated with how this should work.

> Functions can execute other functions directly or there will be some orchestrator which will scan function dependencies, build execution graph and execute all functions in required order.

We have the Planner which is our Function -> Function orchestrator.

> Functions can execute other functions directly

This change is a way of allowing it. KernelExecutionContext is not a fullfledged kernel, it's intent is to live only during the execution of the function as a security boudary to what a function also has access.

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -230,10 +230,7 @@ public void AddSteps(params ISKFunction[] steps)
     /// </remarks>
     public Task<Plan> RunNextStepAsync(IKernel kernel, ContextVariables variables, CancellationToken cancellationToken = default)
     {
-        var context = new SKContext(
-            kernel,
-            variables,
-            kernel.Functions);
+        var context = kernel.CreateNewContext(variables);
```
markwallace-microsoft, line 233: Looks like we create this context just so we can pass the `kernel` and `variables` to `InvokeNextStepAsync`

Inside `InvokeNextStepAsync` it looks like we only use this new `context` to access the values, so why not pass `kernel` and `variables` directly?

```

File: dotnet/src/SemanticKernel/Planning/Plan.cs [hunk 9]
```diff
@@ -255,9 +252,7 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
             var functionVariables = this.GetNextStepVariables(context.Variables, step);
```
markwallace-microsoft, line 252: If we pass in `variables` this will become

`var functionVariables = this.GetNextStepVariables(variables, step);`

dotnet/src/SemanticKernel/Planning/Plan.cs
```diff
@@ -255,9 +252,7 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
...
 
             // Execute the step
-            var functionContext = new SKContext(context.Kernel, functionVariables, context.Functions);
-
-            var result = await step.InvokeAsync(functionContext, cancellationToken: cancellationToken).ConfigureAwait(false);
+            var result = await context.Kernel.RunAsync(functionVariables, step, cancellationToken).ConfigureAwait(false);
```
markwallace-microsoft, line 255: If we pass in `kernel` this will become

`var result = await kernel.RunAsync(functionVariables, step, cancellationToken).ConfigureAwait(false);`

dotnet/src/SemanticKernel/Kernel.cs
```diff
@@ -250,11 +254,19 @@ public async Task<SKContext> RunAsync(ContextVariables variables, CancellationTo
...
-            this,
-            functions: this._functionCollection);
+#pragma warning disable CA2000 // Dispose objects before losing scope
+        var kernelContext = new KernelExecutionContext(
+            functions ?? new FunctionCollection(this._functionCollection),
+            this._aiServiceProvider,
+            this.PromptTemplateEngine,
+            this.Memory,
+            this.HttpHandlerFactory,
+            this.LoggerFactory);
```
markwallace-microsoft, line 266: Can this be simplified to `var executionContext = new KernelExecutionContext(this, functions);`

dotnet/src/SemanticKernel/Planning/Plan.cs
RogerBarreto: This will be changed in a future PR update. (Phase 3 and 4 will be changing Plans and Template engines).

RogerBarreto: @lemillermicrosoft FYI

dotnet/src/SemanticKernel/Planning/Plan.cs
RogerBarreto: Good suggestion. 
All plan specific structural/logic changes will be in a next phase.

lemillermicrosoft (approved):
Thanks for chatting on this, understand the next phase. LGTM.

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContextExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Extension methods for <see cref="IKernelExecutionContext"/>
+/// </summary>
+public static class KernelExecutionContextExtensions
```
shawncal, line 11: These RunAsync overloads -- if each one is not being used right now by something in the repo, please don't add them. I'd like to work toward removing "pipeline" and instead have kernel (and kernelContextExtensions) operate with `IPlan`.

dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -194,6 +194,10 @@ public Task<SKContext> RunAsync(CancellationToken cancellationToken, params ISKF
     public Task<SKContext> RunAsync(string input, CancellationToken cancellationToken, params ISKFunction[] pipeline)
         => this.RunAsync(new ContextVariables(input), cancellationToken, pipeline);
 
+    /// <inheritdoc/>
```
shawncal, line 197: Trying to reduce these overloads. Why is this one needed? I'd remove it from both places

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContext.cs
```diff
@@ -0,0 +1,49 @@
...
+            new FunctionCollection(functionCollection),
+            aiServiceProvider,
+            kernel.PromptTemplateEngine,
+            kernel.Memory,
+            kernel.HttpHandlerFactory,
+            kernel.LoggerFactory);
+    }
+
+    /// <inheritdoc/>
+    public Task<SKContext> RunAsync(ContextVariables variables, ISKFunction[] pipeline, CancellationToken cancellationToken = default)
```
shawncal, line 39: I would much rather have only this version of RunAsync in this interface (no pipelines):
    
public Task<SKContext> RunAsync(ISKFunction skFunction,

Hopefully that alone is enough (I don't think we need to add the other extension overloads.) If we do, please only add strictly the ones we need today to unblock what's in the repo.

dotnet/src/SemanticKernel.Core/Functions/FunctionCollection.cs
```diff
@@ -29,13 +29,34 @@ public class FunctionCollection : IFunctionCollection
     /// <summary>
     /// Initializes a new instance of the <see cref="FunctionCollection"/> class.
     /// </summary>
+    /// <param name="readOnlyFunctionCollection">Optional skill collection to copy from</param>
     /// <param name="loggerFactory">The logger factory.</param>
-    public FunctionCollection(ILoggerFactory? loggerFactory = null)
+    public FunctionCollection(IReadOnlyFunctionCollection? readOnlyFunctionCollection = null, ILoggerFactory? loggerFactory = null)
     {
-        this._logger = loggerFactory?.CreateLogger(typeof(FunctionCollection)) ?? NullLogger.Instance;
+        this._logger = loggerFactory is not null ? loggerFactory.CreateLogger(typeof(FunctionCollection)) : NullLogger.Instance;
```
WeihanLi, line 36: Any reason for this change?

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContext.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Kernel execution context implementation
+/// This will live during the execution of functions.
+/// </summary>
+internal sealed class KernelExecutionContext : IKernelExecutionContext, IDisposable
+{
+    private Kernel _kernel;
```
WeihanLi, line 17: Seemed this can be `readonly`

dotnet/src/SemanticKernel.UnitTests/Planning/PlanSerializationTests.cs
```diff
@@ -15,7 +15,7 @@ namespace SemanticKernel.UnitTests.Planning;
 
 public sealed class PlanSerializationTests
 {
-    private readonly Mock<IKernel> _kernel = new();
+    private Mock<IKernelExecutionContext> _kernelContext = new();
```
WeihanLi, line 18: Why the `readonly` removed, seemed nowhere updates it

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContextExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /// <summary>
+    /// Run a pipeline composed of synchronous and asynchronous functions.
+    /// </summary>
+    /// <param name="kernelContext">Kernel context</param>
+    /// <param name="input">Input</param>
+    /// <param name="pipeline">Target functions to run in sequence</param>
+    /// <returns>Result of the function composition</returns>
+    public static Task<SKContext> RunAsync(this IKernelExecutionContext kernelContext, string input, ISKFunction pipeline)
```
WeihanLi, line 33: Maybe better with the optional cancellationToken `CancellationToken cancellationToken = default`

dotnet/src/SemanticKernel.Core/Kernel.cs
RogerBarreto: You are right, using just one impl. No more overloads.

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContext.cs
RogerBarreto: Agreed, overloads removed, using the suggested one.

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContext.cs
RogerBarreto: Agree, `kernel` will be lazy loaded at some point, but for now readonly works.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanSerializationTests.cs
RogerBarreto: Added.

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContextExtensions.cs
RogerBarreto: All overloads were removed.

dotnet/src/SemanticKernel.Core/Functions/FunctionCollection.cs
RogerBarreto: Auto merge did this. will revert.

lemillermicrosoft (approved):

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -132,7 +132,7 @@ public string Type05(SKContext context)
     public async Task<string> Type06Async(SKContext context)
     {
         var summarizer = context.Functions.GetFunction("SummarizePlugin", "Summarize");
-        var summary = await context.Kernel.RunAsync("blah blah blah", summarizer);
+        var summary = await context.Kernel.RunAsync(summarizer, new ContextVariables("blah blah blah"));
```
markwallace-microsoft, line 135: Do we have an obsolete method for the old method signature?

dotnet/samples/NCalcPlugins/NCalcPlugins.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\src\SemanticKernel.Abstractions\SemanticKernel.Abstractions.csproj" />
```
markwallace-microsoft, line 8: Why is this needed?

markwallace-microsoft (commented):
Feedback on the core code changes. I still need to review unit tests.

dotnet/src/SemanticKernel.Core/Extensions/SKFunctionExtensions.cs
```diff
@@ -46,10 +46,8 @@ public static ISKFunction UseCompletionSettings(this ISKFunction skFunction, AIR
         AIRequestSettings? requestSettings = null,
         CancellationToken cancellationToken = default)
     {
-        var context = new SKContext(kernel, variables, functions ?? kernel.Functions)
-        {
-            Culture = culture!
-        };
+        var context = kernel.CreateNewContext(variables, functions);
+        context.Culture = culture!;
```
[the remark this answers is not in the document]
RogerBarreto: Culture seems to not be inteded to be unique per SKContext, is bound to change.

Check UT: `ItUsesContextCultureForParsingFormattingAsync`

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
RogerBarreto: There wasn't an old method, the previous `RunAsync` was from IKernel interface. As this is a recent change (that don't change the method name) 

I don't see a strong reason to create a new ExtensionClass implementing all the IKernel overloads Obsoleted into KernelExecutionContext.

dotnet/src/SemanticKernel.Core/Orchestration/KernelExecutionContext.cs
```diff
@@ -0,0 +1,49 @@
...
+        IReadOnlyFunctionCollection functionCollection,
+        IAIServiceProvider aiServiceProvider)
+    {
+        this._kernel = new Kernel(
+            new FunctionCollection(functionCollection),
+            aiServiceProvider,
+            kernel.PromptTemplateEngine,
+            kernel.Memory,
+            kernel.HttpHandlerFactory,
+            kernel.LoggerFactory);
```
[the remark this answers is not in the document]
RogerBarreto: KernelExecutionContext needs all stuff needed to instantiate a standalone Kernel to run in the context.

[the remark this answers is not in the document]
markwallace-microsoft: So when I look at the implementation of ` public async Task<KernelResult> RunAsync(ContextVariables variables, CancellationToken cancellationToken, params ISKFunction[] pipeline)` I don't see references to template engine and memory. My understanding is that we want the execution context to be restricted to just what is needed so eliminating stuff would be good. Am I missing something?

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelExecutionContext.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Kernel execution context
+/// </summary>
+public interface IKernelExecutionContext
```
SergeyMenshykh, line 12: Can we name it so that it's clear that it's not a context and that it's purpose to execute functions? E.g. ISKFunctionRunner/Executor/Invoker?

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelExecutionContext.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>
+    /// App logger
+    /// </summary>
+    ILoggerFactory LoggerFactory { get; }
```
SergeyMenshykh, line 17: Can we inject the logger factory directly into the components that need it?

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelExecutionContext.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Read only skills collection
+    /// </summary>
+    IReadOnlyFunctionCollection Functions { get; }
```
SergeyMenshykh, line 22: SKContext already has Functions property. I would propose to use it instead of adding another one.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -64,57 +64,25 @@ public CultureInfo Culture
...
     /// </summary>
-    /// <param name="kernel">Kernel reference</param>
+    /// <param name="kernelContext">Kernel reference</param>
     /// <param name="variables">Context variables to include in context.</param>
-    /// <param name="functions">Functions to include in context.</param>
-    public SKContext(
-        IKernel kernel,
+    /// <param name="culture">Culture related to the context</param>
+    internal SKContext(
+        IKernelExecutionContext kernelContext,
```
SergeyMenshykh, line 77: We can keep the `IKernel kernel` parameter but return the kernel as an instance of the IKernelExecutionContext from the `Kernel` property getter. It'll allow to avoid putting the LoggerFactory and Functions into the IKernelExecutionContext interface.

RogerBarreto: Made the change following the discussion regarding `IFunctionExecutor`

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelExecutionContext.cs
RogerBarreto: Removed from the interface

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelExecutionContext.cs
RogerBarreto: Not needed, removed

dotnet/src/SemanticKernel.Abstractions/Orchestration/IKernelExecutionContext.cs
RogerBarreto: REnamed to `IFunctionExecutor`

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -45,17 +45,17 @@ public CultureInfo Culture
...
     /// App logger
     /// </summary>
-    public ILoggerFactory LoggerFactory { get; }
+    public ILoggerFactory LoggerFactory => this.Kernel.LoggerFactory;
 
     /// <summary>
     /// Kernel context reference
     /// </summary>
-    public IKernel Kernel => this.GetKernelContext();
+    public IKernelExecutionContext Kernel => this.GetKernelContext();
```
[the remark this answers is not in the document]
RogerBarreto: Changed to `IFunctionExecutor`

dotnet/src/SemanticKernel.Core/Orchestration/FunctionExecutor.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    internal FunctionExecutor(Kernel originKernel)
+    {
+        this._originKernel = originKernel;
+    }
+
+    /// <inheritdoc/>
+    public Task<KernelResult> ExecuteAsync(ISKFunction skFunction, ContextVariables variables, IReadOnlyFunctionCollection? functions = null, CancellationToken cancellationToken = default)
+    {
+        using var kernel = this._originKernel.Clone(functions);
```
SergeyMenshykh, line 23: Do we have a case when we need to have a kernel with a different set of functions than the original one?

markwallace-microsoft, line 23: Is this achieving anything?

E.g., when I load the semantic function it currently is bound to the kernel instance used to load it i.e. it has a reference to the kernel functions and gets the AI service from the kernel.

markwallace-microsoft (commented):
A few questions

dotnet/src/SemanticKernel.Abstractions/Orchestration/IFunctionExecutor.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Kernel execution context
+/// </summary>
+public interface IFunctionExecutor
```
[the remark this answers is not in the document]
SergeyMenshykh: @markwallace-microsoft I don't have strong preferences here, so I'll accept the other folk's decision.

dmytrostruk (approved):

dmytrostruk (commented):
Approved accidentally :) There are still some unresolved comments.

dotnet/src/SemanticKernel.Core/Orchestration/FunctionExecutor.cs
RogerBarreto: The original idea was to expose a kernel that could be changed only during the execution time and those changes wouldn't affect the original kernel.

dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -268,6 +276,23 @@ public void Dispose()
         if (this._functionCollection is IDisposable reg) { reg.Dispose(); }
     }
 
+    internal Kernel Clone(IReadOnlyFunctionCollection? functions = null)
```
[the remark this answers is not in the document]
RogerBarreto: With the latest discussions, I'm removing this method and using the same kernel.

dotnet/src/SemanticKernel.Core/Orchestration/FunctionExecutor.cs
RogerBarreto: Normally this can happen from the plan side, while you want to run from a specific set of functions, not all.

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -156,8 +157,16 @@ public interface IKernel
...
+    /// <param name="functions">Provide specific scoped functions. Defaults to all existing in the kernel</param>
+    /// <param name="loggerFactory">Logged factory used within the context</param>
+    /// <param name="culture">Optional culture info related to the context</param>
     /// <returns>SK context</returns>
-    SKContext CreateNewContext();
+    SKContext CreateNewContext(
+        ContextVariables? variables = null,
+        IReadOnlyFunctionCollection? functions = null,
+        ILoggerFactory? loggerFactory = null,
+        CultureInfo? culture = null);
```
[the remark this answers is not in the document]
RogerBarreto: Yep, this is very common to aggregate information using multiple functions.

dotnet/src/SemanticKernel.Abstractions/Orchestration/IFunctionRunner.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+/// <summary>
+/// Kernel execution context
```
SergeyMenshykh, line 9: nit: XML comment needs to be improved.

dotnet/src/SemanticKernel.Core/Orchestration/FunctionRunner.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Orchestration;
+
+internal class FunctionRunner : IFunctionRunner
```
SergeyMenshykh, line 9: Please add XML comments to the class and its members.

dotnet/src/SemanticKernel.Core/Orchestration/FunctionRunner.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    private readonly IKernel _kernel;
```
SergeyMenshykh, line 11:
```suggestion
    private readonly IKernel _kernel;
    
```

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -156,8 +157,16 @@ public interface IKernel
     /// <summary>
     /// Create a new instance of a context, linked to the kernel internal state.
     /// </summary>
+    /// <param name="variables">Initializes the context with the provided variables</param>
+    /// <param name="functions">Provide specific scoped functions. Defaults to all existing in the kernel</param>
+    /// <param name="loggerFactory">Logged factory used within the context</param>
+    /// <param name="culture">Optional culture info related to the context</param>
     /// <returns>SK context</returns>
-    SKContext CreateNewContext();
+    SKContext CreateNewContext(
```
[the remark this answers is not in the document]
RogerBarreto: We will be able to remove it once the decision on `Kernel.RunAsync` as the only option is made and the new implementation is done.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #2914, opened 2023-09-21
.Net: Extract Kernel.RunAsync overloads [DRAFT]

shawncal:
### Motivation and Context

The Kernel.RunAsync overloads that take params ISKFunction do not need to be part of the primary IKernel interface. Kernel itself needs only run the function and coordinate the pre/post execution hooks. Complex executions, with more than one function, can be represented as a Plan, then run via RunAsync like any other ISKFunction.  

### Description

Extract the "pipeline" overloads of Kernel.RunAsync into a static extension methods class.
Pipeline overloads create then run a Plan.
Change the Kernel so that the single remaining RunAsync executes a single ISKFunction, instead of a loop.
Handle CancellationToken within Plan.
Update tests.

Note: Callers can, and likely should, be updated to create a new Plan and run it on the Kernel, instead of the pipeline functions. These will soon be obsoleted and removed - for now, this removes them from IKernel without breaking callers.

### Contribution Checklist

### DIFF
--- a/dotnet/samples/KernelSyntaxExamples/Example12_SequentialPlanner.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example12_SequentialPlanner.cs
@@ -166,7 +166,7 @@ internal static class Example12_SequentialPlanner
             Console.WriteLine($"Restored plan (relevance={memory.Relevance}):");
 
             // Deseriliaze the plan from the description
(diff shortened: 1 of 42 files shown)


RogerBarreto:
THis moves Plan to a whole new level, I liked it. I think this changes also deserves a 

SemanticKernel.Abstractions.Planning.IPlan

RogerBarreto:
Closing to be revisited in the future.

PR #2943, opened 2023-09-22
.Net: Rename ImportXXXPlugins methods to ImportXXXFunctions for GRPC

markwallace-microsoft:
### Contribution Checklist


dotnet/src/Functions/Functions.Grpc/Extensions/KernelGrpcExtensions.cs
```diff
@@ -195,4 +196,47 @@ async Task<SKContext> ExecuteAsync(SKContext context)
...
+        string skillDirectoryName,
+        HttpClient? httpClient = null)
+    {
+        return kernel.ImportGrpcFunctionsFromDirectory(parentDirectory, skillDirectoryName, httpClient);
+    }
+#pragma warning restore CS1591
+
+    [Obsolete("Methods and classes which includes Skill in the name have been renamed to use Plugin. Use Kernel.ImportGrpcFunctionsFromFile instead. This will be removed in a future release.")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+#pragma warning disable CS1591
```
hario90, line 217: nit: use `SuppressMessage` attribute on these methods

hario90 (approved):

dotnet/src/Functions/Functions.Grpc/Extensions/KernelGrpcExtensions.cs
lemillermicrosoft: We've actually split usage quite a bit in the repo. I kind of prefer the pragma warnings just because it doesn't require a namespace using and can apply to files or specific lines more easily. Are there other considerations here on why we might want to use SuppressMessage instead of the pragma markers?

lemillermicrosoft (approved):

Issue #2619, opened 2023-08-29
Java: batch operations from SQL memory connectors aren't really a batch


milderhc:
Support batch operations in the SQL-related connectors.

PR #2960, opened 2023-09-25
Java: Add batch operations to SQL memory stores

milderhc:
### Motivation and Context

Fix https://github.com/microsoft/semantic-kernel/issues/2619

### Contribution Checklist


java/connectors/semantickernel-connectors-memory-jdbc/src/main/java/com/microsoft/semantickernel/connectors/memory/jdbc/JDBCConnector.java
```diff
@@ -81,8 +83,8 @@ public Mono<Void> createTableAsync() {
                                     "CREATE TABLE IF NOT EXISTS "
                                             + TABLE_NAME
                                             + " ("
-                                            + "collection TEXT, "
-                                            + "key TEXT, "
+                                            + "collection TEXT NOT NULL, "
+                                            + "key TEXT NOT NULL, "
                                             + "metadata TEXT, "
                                             + "embedding TEXT, "
                                             + "timestamp TEXT, "
```
johnoliver, line 90: Is this a TEXT or some other Data type?

java/connectors/semantickernel-connectors-memory-jdbc/src/main/java/com/microsoft/semantickernel/connectors/memory/jdbc/JDBCConnector.java
```diff
@@ -135,70 +137,84 @@ public Mono<Void> createCollectionAsync(String collectionName) {
...
+    }
+
+    public Mono<String> upsertAsync(
             String collection,
             String key,
             String metadata,
             String embedding,
             ZonedDateTime timestamp) {
-
+        final String upsertKey = resolveKey(key);
```
johnoliver, line 150: Is it consistent with .net that key can be null and will be generated for you? If so add Nullable annotation to the `key` parameter so people know it can be null

java/connectors/semantickernel-connectors-memory-jdbc/src/main/java/com/microsoft/semantickernel/connectors/memory/jdbc/JDBCConnector.java
milderhc: Solving `timestamp` type in a further PR.

johnoliver (approved):

PR #2976, opened 2023-09-25
925 stepwise functionresult

lemillermicrosoft:
### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/AIFunctionResultExtensions.cs
```diff
@@ -21,15 +21,30 @@ public static class AIFunctionResultExtensions
...
             return modelResults;
         }
 
         return null;
     }
 
+    /// <summary>
+    /// Get typed value from <see cref="FunctionResult"/> metadata.
+    /// </summary>
+    public static bool TryGetValue<T>(this FunctionResult result, string key, out T value)
```
dmytrostruk, line 35: Since `FunctionResult` already contains method `GetValue<T>`, it looks like this method `TryGetValue<T>` works in similar way, but with boolean result and `out` parameter. While it's working differently, since it returns value not from `object Value`, but from `Metadata`. So, the naming may be confusing here. I would name it `TryGetMetadata` or `TryGetMetadataValue` to clearly define that we work with `Metadata` here.

dotnet/src/SemanticKernel.Abstractions/AI/AIFunctionResultExtensions.cs
```diff
@@ -21,15 +21,30 @@ public static class AIFunctionResultExtensions
...
+    {
+        if (result.Metadata.TryGetValue(key, out object? valueObject) &&
+            valueObject is T valueT)
```
dmytrostruk, line 38: nit: I would rename `valueT` to `typedValue` for better readability.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -80,4 +81,15 @@ public FunctionResult(string functionName, string pluginName, SKContext context,
...
     }
+
+    /// <summary>
+    /// Populates metadata with variables from <see cref="SKContext"/>.
+    /// </summary>
+    internal void PopulateMetadata()
+    {
+        foreach (var variable in this.Context.Variables)
+        {
+            this.Metadata[variable.Key] = variable.Value;
```
dmytrostruk, line 92: I'm wondering, if it will be possible to use `Metadata` dictionary only for fields that are required to expose, but not all the variables from `Context`. Otherwise, we can just make `Context` or `Context.Variables` field public. But main idea here was to expose only what is needed for end-user and hide unnecessary context information. Because if context information is opened, the context is used in the wrong way for workarounds and hacks :)

In this scenario, we can add to metadata key `StepwisePlanner` with value, which could be complex object, including properties `StepCount`, `FunctionCount` and `Iterations`. These properties could be set somewhere in `StepwisePlanner` logic before returning `FunctionResult`, if that is possible.

lemillermicrosoft: Yeah this is fair. Will move to `Plan` object where `Outputs` can be directly referenced.

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -80,4 +80,20 @@ public FunctionResult(string functionName, string pluginName, SKContext context,
 
         throw new InvalidCastException($"Cannot cast {this.Value.GetType()} to {typeof(T)}");
     }
+
+    /// <summary>
+    /// Get typed value from metadata.
+    /// </summary>
+    public bool TryGetMetadataValue<T>(string key, out T value)
```
hario90, line 87: Given that this is a public method, some tests for this could help prevent regressions.

dotnet/src/SemanticKernel.Core/Planning/Plan.cs
```diff
@@ -359,6 +359,7 @@ public FunctionView Describe()
                 this.UpdateContextWithOutputs(context);
 
                 result = new FunctionResult(this.Name, this.PluginName, context, context.Result);
+                this.UpdateFunctionResultWithOutputs(result);
```
hario90, line 362: This added functionality could use some unit tests.

lemillermicrosoft: Thanks for the callout. This now has me reconsidering a few things. Marking as draft for now.

dmytrostruk (approved):

hario90 (approved):

PR #2986, opened 2023-09-26
.Net: Kernel Streaming Support

### LINKED ISSUE #1298: .NET Allow ISKFunction to return IAsyncEnumerable

### LINKED ISSUE #1649: .Net: Streaming support for semantic functions on IKernel
> [!IMPORTANT]  
> Labeled **Urgent** because it will require a breaking change.

Add streaming APIs to IKernel, in addition to the work that's been done for ITextCompletion and IChatCompletion. 
Add streaming from SK Functions

```[tasklist]
### Child issues
- [x] [Update PrePost hooks examples and implementations to suffice this Issue completely](https://github.com/microsoft/semantic-kernel/issues/2730)
```

RogerBarreto:
### Motivation and Context

- As a long-awaited feature request, this new allows streaming when calling Kernel.RunAsync or SKFunctions.InvokeAsync. 
Resolves #1649 
Resolves #1298 

- Adjusted `Completion` abstract interface methods/name following a common pattern 
- Added streaming component interface specializations. 

### Description

⚠️ Breaking Change

Streaming behavior by default is false since streaming responses from providers like OpenAI/Azure OpenAI won't give additional data like (PromptFilterResults, Usage(Completion Tokens, Prompt Tokens, Total Tokens, ...).

To enable streaming behavior set `Streaming = true` in a `AIRequestSettings` specialized class like `OpenAIRequestSettings`, once this is set for text/chat based models like OpenAi's GPT you can get streaming data using `GetValue<IAsyncEnumerable<string>>()` in the result of `Kernel.RunAsync` or `Function.InvokeAsync`.

How to:
```
myFunction = kernel.CreateSemanticFunction($"your prompt",
        requestSettings: new OpenAIRequestSettings
        {
            Streaming = true,
        });

var result = await kernel.RunAsync(myFunction);
await foreach (string token in result.GetValue<IAsyncEnumerable<string>>())
{
    Console.Write(token);
}
```

ℹ️ Any attempt on using `SKContext` from `FunctionResults` will cancel the streaming behavior as those require the full content to be available.

New Streaming flag added to `bool AIRequestSettings.Streaming` abstraction, this will signal that the caller has intention on use streaming when available.

New streaming interfaces added: 
- `ITextStreamingCompletion`
- `IChatStreamingCompletion`

Changes to the method names following the pattern (which service is it and what it is returning)

- IText(Streaming)Completion
    - Old: `List<ITextResult> GetCompletionsAsync` 
    New: `GetTextResultsAsync`
    - Old: `IAsyncEnumerable<ITextStreamingResult> GetStreamingCompletionsAsync` 
    New: `GetTextStreamingResultsAsync`

- IText(Streaming)Result
   - Old: `Task<string> GetCompletionAsync`
   New: `GetTextAsync`
   - Old: `IAsyncEnumerable<ITextStreamingResult> GetStreamingCompletionsAsync`
   New: `GetTextStreamingResultsAsync`

- IChat(Streaming)Completion
    - Old: `Task<IReadOnlyList<IChatResult>> GetChatCompletionsAsync`
    New: `GetChatResultsAsync`
    - Old: `IAsyncEnumerable<IChatStreamingResult> GetStreamingChatCompletionsAsync`
    New: `GetChatStreamingResultsAsync`
 

Services following this abstraction signales to Kernel it can use streaming when requested.

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/StreamingTextModelResult.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using Azure.AI.OpenAI;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
+
+/// <summary> Represents a singular result of a chat completion.</summary>
```
hario90, line 8: Is this description accurate?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/StreamingTextModelResult.cs
```diff
@@ -0,0 +1,34 @@
...
+public class StreamingTextModelResult
+{
+    /// <summary> A unique identifier associated with this chat completion response. </summary>
```
hario90, line 11:
```suggestion
    /// <summary> A unique identifier associated with this text completion response. </summary>
```

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIModelResultExtensions.cs
```diff
@@ -23,6 +23,16 @@ public static TextModelResult GetOpenAITextResult(this ModelResult resultBase)
         return resultBase.GetResult<TextModelResult>();
     }
 
+    /// <summary>
+    /// Retrieves a typed <see cref="Completions"/> OpenAI / AzureOpenAI result from text completion prompt.
```
hario90, line 27: It seems confusing that this returns the AzureOpenAI result type. Is this what you meant?
```
/// Retrieves a <see cref="StreamingTextModelResult "/> from the model result.
```

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
```diff
@@ -63,7 +63,22 @@ public class OpenAIRequestSettings : AIRequestSettings
     /// Use carefully and ensure that you have reasonable settings for max_tokens and stop.
     /// </summary>
     [JsonPropertyName("results_per_prompt")]
-    public int ResultsPerPrompt { get; set; } = 1;
+    public int ResultsPerPrompt
+    {
+        get
+        {
+            if (this.ExtensionData.TryGetValue("results_per_prompt", out var value))
```
hario90, line 70: suggestion: use a `const` for the string literal

dotnet/src/SemanticKernel.Core/Functions/SemanticFunction.cs
```diff
@@ -217,15 +211,36 @@ private void AddDefaultValues(ContextVariables variables)
...
-            string completion = await GetCompletionsResultContentAsync(completionResults, cancellationToken).ConfigureAwait(false);
+            var modelResults = new List<ModelResult>();
 
-            // Update the result with the completion
-            context.Variables.Update(completion);
-
-            result = new FunctionResult(this.Name, this.PluginName, context, completion);
-
-            var modelResults = completionResults.Select(c => c.ModelResult).ToArray();
+            if (requestSettings?.Streaming == true &&
```
dmytrostruk, line 216: I would suggest to move streaming and non-streaming logic to separate private methods to keep method `RunPromptAsync` as simple as possible.

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIRequestSettings.cs
dmytrostruk, line 70: If this property is applicable to general SemanticFunction logic (Method `GetResultsCountOrDefault`), we can:
- Move `ResultsPerPrompt` parameter to base `AIRequestSettings` with default value `1` - although it will mean that this parameter is applicable to every AI connector, which is not the case.
- Place logic for multiple results only where it's applicable - move it from SemanticFunction to OpenAI Connector.
- Remove this property from `OpenAIRequestSettings` completely and place it as parameter of function. (I had an idea to replace SKContext as input parameter for function/kernel with `FunctionRequest`/`KernelRequest`, and property `ResultsPerPrompt` can be good first candidate for request models, but this is definitely separate scope of work and PR.)

Main point here, that we should not use `ExtensionData` as a way of storing general settings for functions.

dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -228,8 +228,6 @@ public async Task<KernelResult> RunAsync(ContextVariables variables, Cancellatio
 
                 functionResult = await skFunction.InvokeAsync(context, cancellationToken: cancellationToken).ConfigureAwait(false);
 
-                context = functionResult.Context;
```
dmytrostruk, line 231: This is breaking change with unknown outcomes. That's the reason why property `Context` still exists in `FunctionResult` model internally. Because, some functions may update context inside (since context is mutable at the moment), but some of them can return new cloned context instance with updated parameters (that's probably the way how it should be), which means that next function in pipeline won't receive updated variables if we remove this line.

We should investigate and test all available scenarios to understand if we can safely do this change and if we need to do it at all.

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -152,45 +162,78 @@ public void ItAllowsToImportTheSamePluginMultipleTimes()
...
 
         var kernel = Kernel.Builder
             .WithAIService<ITextCompletion>("service1", mockTextCompletion1.Object, false)
             .WithAIService<ITextCompletion>("service2", mockTextCompletion2.Object, true)
             .Build();
 
         var templateConfig = new PromptTemplateConfig();
+        if (useStream is not null)
+        {
+            templateConfig.Completion = new() { Streaming = useStream };
```
dmytrostruk, line 193: As mentioned above, name `Streaming` here doesn't show that it's boolean flag, that's why this test contains parameter `useStream`. I would prefer to use the same naming in settings and tests, like `UseStreaming` or `IsStreamingEnabled`.

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -495,6 +566,55 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+        var kernelResult = await kernel.RunAsync(function1, function2, function3);
+
+        // Assert
+        Assert.NotNull(kernelResult);
+
+        var functionResult1 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function1" && l.PluginName == PluginName);
+        var functionResult2 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function2" && l.PluginName == PluginName);
+        var functionResult3 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function3" && l.PluginName == PluginName);
+
+        var function3Result = functionResult3.GetValue<string>();
```
dmytrostruk, line 606: Do we need this result in separate variable?

RogerBarreto:
Dropping this change for now, as I will be working in potentially other decision regarding the Streaming approach for v1

PR #3000, opened 2023-09-27
Use method to check char is Symbol

feiyun0112:
Use same method to check char is Symbol

for example:
````
switch (nextChar)
{
    case Symbols.VarPrefix:
        blocks.Add(new VarBlock(text, this._loggerFactory));
        break;

    case Symbols.DblQuote:
    case Symbols.SglQuote:
        blocks.Add(new ValBlock(text, this._loggerFactory));
        break;

    default:
        blocks.Add(new FunctionIdBlock(text, this._loggerFactory));
        break;
}
````
use method:
````
if (Block.IsVarPrefix(nextChar))
{
    blocks.Add(new VarBlock(text, this._loggerFactory));
}
else if (Block.IsQuote(nextChar))
{
    blocks.Add(new ValBlock(text, this._loggerFactory));
}
else
{
    blocks.Add(new FunctionIdBlock(text, this._loggerFactory));
}
````

### DIFF
--- a/dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/Block.cs
+++ b/dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/Block.cs
@@ -41,4 +41,24 @@ public abstract class Block
     /// <param name="errorMsg">Error message in case the content is not valid</param>
     /// <returns>True if the block content is valid</returns>
     public abstract bool IsValid(out string errorMsg);
+
+    internal static bool IsVarPrefix(char c)
+    {
+        return (c == Symbols.VarPrefix);
+    }
+
+    internal static bool IsBlankSpace(char c)
+    {
+        return c is Symbols.Space or Symbols.NewLine or Symbols.CarriageReturn or Symbols.Tab;
+    }
+
+    internal static bool IsQuote(char c)
+    {
+        return c is Symbols.DblQuote or Symbols.SglQuote;
+    }
+
+    internal static bool CanBeEscaped(char c)
+    {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/Block.cs [hunk 1]
```diff
@@ -41,4 +41,24 @@ private protected Block(string? content, ILoggerFactory? loggerFactory)
     /// <param name="errorMsg">Error message in case the content is not valid</param>
     /// <returns>True if the block content is valid</returns>
     public abstract bool IsValid(out string errorMsg);
+
+    internal static bool IsVarPrefix(char c)
```
dmytrostruk, line 45: Would it be better if we put these helper methods to static `Symbols` class?

dmytrostruk (approved):

dotnet/src/Extensions/TemplateEngine.PromptTemplateEngine/Blocks/Block.cs
SergeyMenshykh: +1 for moving these methods to the `Symbols` class. The `Block` class should not change every time a new symbol or helper method is added, removed, or changed.

markwallace-microsoft:
@feiyun0112 we've renamed this project as part of work we are doing to get ready for a release candidate. Can you please update the PR and we'll get it merged next week.

SergeyMenshykh (approved):

markwallace-microsoft:
@feiyun0112 I'm going to close this PR for now. Feel free to update and re-open and we will get it merged.

PR #3002, opened 2023-09-27
Python: Remapping of memory fields to custom ACS index schema at the constructor instead of global

awharrison-28:
### Motivation and Context
This PR addresses the scenario where a user wishes to map the fields of an Azure Cognitive Search index, created independently from SK, to SK memory. 

Before this work, these fields could be manipulated by setting semantic_kernel.connectors.memory.azure_cognitive_search.utils.SEARCHABLE_FIELD.... = "new_string"

However, this is a global string, and would not be scalable to having 2+ memory stores with different index schemas. This change allows the renaming/remapping of ACS index schemas to be isolated to a single memory store instance. 

Follow up work should be investigated to extend this functionality to other/all memory stores. 

### Description
ACS memory has a property called _schema to remember any of these overrides. The _schema property can be modified at the construction of the memory store. Some methods had to be moved from utils.py to the ACSMemoryStore class to access this new configuration. 

**Current Pattern**
```
from semantic_kernel.connectors.memory.azure_cognitive_search import utils as acs_utils

acs_utils.SEARCH_FIELD_EMBEDDING = "something_else"

memory_store = AzureCognitiveSearchMemoryStore(
        vector_size,
        AZURE_COGNITIVE_SEARCH_ENDPOINT,
        AZURE_COGNITIVE_SEARCH_ADMIN_KEY
)
```

**New Pattern**
```
from semantic_kernel.connectors.memory.azure_cognitive_search import utils as acs_utils

memory_store = AzureCognitiveSearchMemoryStore(
        vector_size,
        AZURE_COGNITIVE_SEARCH_ENDPOINT,
        AZURE_COGNITIVE_SEARCH_ADMIN_KEY,
        custom_schema={acs_utils.SEARCH_FIELD_EMBEDDING_KEY: "something_else"},
)
```

### Contribution Checklist


alliscode:
@dluc would you mind taking a look at this one please?

PR #3030, opened 2023-09-30
.Net: Bug + Regression Fix + Added UTs

RogerBarreto:
### Motivation and Context

1. One Bug was identified in the Pre/Post Handler examples where the result was not showing properly.
2. Regression introduced where changes made during FunctionInvoked event were not reflecting in the result.
3. Added Unit Tests to unsure expected behavior with events.

### Description

- In `Example57_FunctionEventHandlers.cs`, the `Console.WriteLine` statements that print the function results have been updated to use the `GetValue<string>()` method. This change ensures that the function results are correctly converted to strings before being printed.

- In `Kernel.cs`, a mechanism has been introduced to reflect any changes made to the `SKContext` by invoked handlers in the original function result. This is achieved by creating a new `FunctionResult` instance with the updated `SKContext` and its result.

- New test cases have been added to validate the behavior of the kernel when functions invoked including scenarios where Skip, Repead and Cancel can be triggered from `FunctionInvoking` or `FunctionInvoked` events.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Core/Kernel.cs
+++ b/dotnet/src/SemanticKernel.Core/Kernel.cs
@@ -209,9 +209,16 @@ repeat:
 
                 context = functionResult.Context;
 
+                var functionInvokedArgs = this.OnFunctionInvoked(functionDetails, functionResult);
+
+                if (functionInvokedArgs is not null)
+                {
+                    // All changes to the SKContext by invoked handlers may reflect in the original function result
+                    functionResult = new FunctionResult(functionDetails.Name, functionDetails.PluginName, functionInvokedArgs.SKContext, functionInvokedArgs.SKContext.Result);
+                }
+
                 allFunctionResults.Add(functionResult);
 
-                var functionInvokedArgs = this.OnFunctionInvoked(functionDetails, functionResult);
                 if (functionInvokedArgs?.CancelToken.IsCancellationRequested ?? false)
                 {
                     this._logger.LogInformation("Execution was cancelled on function invoked event of pipeline step {StepCount}: {PluginName}.{FunctionName}.", pipelineStepCount, skFunction.PluginName, skFunction.Name);
--- a/dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
+++ b/dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
@@ -345,16 +345,275 @@ public class KernelTests
 
         // Act
         var kernelResult = await kernel.RunAsync(function1, function2);
+        var functionResult1 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function1" && l.PluginName == PluginName);
+        var functionResult2 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function2" && l.PluginName == PluginName);
 
         // Assert
         Assert.NotNull(kernelResult);
         Assert.Equal("Result2", kernelResult.GetValue<string>());
-
-        var functionResult1 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function1" && l.PluginName == PluginName);
-        var functionResult2 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function2" && l.PluginName == PluginName);
-
         Assert.Equal("Result1", functionResult1.GetValue<string>());
         Assert.Equal("Result2", functionResult2.GetValue<string>());
+    }
+
+    [Fact]
+    public async Task ItReturnsChangedResultsFromFunctionInvokedEventsAsync()
+    {
+        var kernel = Kernel.Builder.Build();
+
+        // Arrange
+        [SKName("Function1")]
+        static string Function1() => "Result1";
+        var function1 = SKFunction.FromNativeMethod(Method(Function1), pluginName: "MyPlugin");
+        const string ExpectedValue = "new result";
+
+        kernel.FunctionInvoked += (object? sender, FunctionInvokedEventArgs args) =>
+        {
+            args.SKContext.Variables.Update(ExpectedValue);
+        };
+
+        // Act
+        var kernelResult = await kernel.RunAsync(function1);
+
+        // Assert
+        Assert.NotNull(kernelResult);
+        Assert.Equal(ExpectedValue, kernelResult.GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().Context.Result);
+    }
+
+    [Fact]
+    public async Task ItReturnsChangedResultsFromFunctionInvokingEventsAsync()
+    {
+        // Arrange
+        var kernel = Kernel.Builder.Build();
+
+        [SKName("Function1")]
+        static string Function1(SKContext context) => context.Variables["injected variable"];
+        var function1 = SKFunction.FromNativeMethod(Method(Function1), pluginName: "MyPlugin");
+        const string ExpectedValue = "injected value";
+
+        kernel.FunctionInvoking += (object? sender, FunctionInvokingEventArgs args) =>
+        {
+            args.SKContext.Variables["injected variable"] = ExpectedValue;
+        };
+
+        // Act
+        var kernelResult = await kernel.RunAsync(function1);
+
+        // Assert
+        Assert.NotNull(kernelResult);
+        Assert.Equal(ExpectedValue, kernelResult.GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().Context.Result);
+    }
+
+    [Theory]
+    [InlineData("Function1", 5)]
+    [InlineData("Function2", 1)]
+    public async Task ItRepeatsFunctionInvokedEventsAsync(string retryFunction, int numberOfRepeats)
+    {
+        // Arrange
+        var kernel = Kernel.Builder.Build();
+
+        [SKName("Function1")]
+        static string Function1(SKContext context) => "Result1";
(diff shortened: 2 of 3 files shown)


dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -209,9 +209,16 @@ public async Task<KernelResult> RunAsync(ContextVariables variables, Cancellatio
 
                 context = functionResult.Context;
 
+                var functionInvokedArgs = this.OnFunctionInvoked(functionDetails, functionResult);
+
+                if (functionInvokedArgs is not null)
+                {
+                    // All changes to the SKContext by invoked handlers may reflect in the original function result
+                    functionResult = new FunctionResult(functionDetails.Name, functionDetails.PluginName, functionInvokedArgs.SKContext, functionInvokedArgs.SKContext.Result);
```
dmytrostruk, line 217: In case of passing `functionInvokedArgs.SKContext.Result` as last parameter of `FunctionResult`, we will override the actual result from function with result from SKContext, which shouldn't be the case. I will put more description in next comment in unit tests.

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+        {
+            args.SKContext.Variables.Update(ExpectedValue);
+        };
+
+        // Act
+        var kernelResult = await kernel.RunAsync(function1);
+
+        // Assert
+        Assert.NotNull(kernelResult);
+        Assert.Equal(ExpectedValue, kernelResult.GetValue<string>());
```
dmytrostruk, line 379: This is invalid scenario. If `Function1` has return type `MyComplexType`, then Kernel should return `MyComplexType` independently from changes in `SKContext` in the middle of execution.

And probably we should not have the ability to change the function result using hooks, because it looks like a place for hacks and workarounds. If my function returns `MyComplexType.PropertyA = "a"`, we should not allow to change it at the end for `MyComplexType.PropertyA = "b"`. At least, I don't see the use cases for it at the moment, but maybe I'm missing something.

dmytrostruk (changes requested):

dotnet/src/SemanticKernel.Core/Kernel.cs
RogerBarreto: This might not be the best use case for it but a new design needs to be done to allow changes to the result.

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
RogerBarreto: This is a requirement for the current Hooks. 

A design needs to be thought on best way to internally be able to change the results thru Invoked hook.

dmytrostruk (approved):

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
-
-        var functionResult1 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function1" && l.PluginName == PluginName);
-        var functionResult2 = kernelResult.FunctionResults.First(l => l.FunctionName == "Function2" && l.PluginName == PluginName);
-
         Assert.Equal("Result1", functionResult1.GetValue<string>());
         Assert.Equal("Result2", functionResult2.GetValue<string>());
     }
 
+    [Fact]
+    public async Task ItReturnsChangedResultsFromFunctionInvokedEventsAsync()
```
markwallace-microsoft, line 359: Can we make the name of this test clearer e.g. `ItUsesFunctionInvokedHookToModifyTheFunctionResult`

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+
+        // Arrange
+        [SKName("Function1")]
+        static string Function1() => "Result1";
+        var function1 = SKFunction.FromNativeMethod(Method(Function1), pluginName: "MyPlugin");
+        const string ExpectedValue = "new result";
+
+        kernel.FunctionInvoked += (object? sender, FunctionInvokedEventArgs args) =>
+        {
+            args.SKContext.Variables.Update(ExpectedValue);
```
markwallace-microsoft, line 371: Can we introduce a pattern where we encapsulate whats going on here e.g. `args.ModifyResult(ExpectedValue)`

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+
+        // Assert
+        Assert.NotNull(kernelResult);
+        Assert.Equal(ExpectedValue, kernelResult.GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().Context.Result);
+    }
+
+    [Fact]
+    public async Task ItReturnsChangedResultsFromFunctionInvokingEventsAsync()
```
markwallace-microsoft, line 385: Suggested name change `ItUsesFunctionInvokingHookToAddAContextVariable`

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+        var kernel = Kernel.Builder.Build();
+
+        [SKName("Function1")]
+        static string Function1(SKContext context) => context.Variables["injected variable"];
+        var function1 = SKFunction.FromNativeMethod(Method(Function1), pluginName: "MyPlugin");
+        const string ExpectedValue = "injected value";
+
+        kernel.FunctionInvoking += (object? sender, FunctionInvokingEventArgs args) =>
+        {
+            args.SKContext.Variables["injected variable"] = ExpectedValue;
```
markwallace-microsoft, line 397: Would encapsulation be better here `args.ModifyVariable("injected variable", ExpectedValue)`

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+        Assert.NotNull(kernelResult);
+        Assert.Equal(ExpectedValue, kernelResult.GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().GetValue<string>());
+        Assert.Equal(ExpectedValue, kernelResult.FunctionResults.Single().Context.Result);
+    }
+
+    [Theory]
+    [InlineData("Function1", 5)]
+    [InlineData("Function2", 1)]
+    public async Task ItRepeatsFunctionInvokedEventsAsync(string retryFunction, int numberOfRepeats)
```
markwallace-microsoft, line 413: Name change "ItUsesFunctionInvokedHookToRepeatFunctionExecution"

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+        Assert.NotNull(kernelResult);
+        Assert.Equal(2 + numberOfRepeats, kernelResult.FunctionResults.Count);
+        Assert.Equal(2 + numberOfRepeats, numberOfInvocations);
+    }
+
+    [Theory]
+    [InlineData("Function1", "Result2 Result3")]
+    [InlineData("Function2", "Result1 Result3")]
+    [InlineData("Function3", "Result1 Result2")]
+    public async Task ItSkipsFunctionsFromFunctionInvokingEventsAsync(string skipFunction, string expectedResult)
```
markwallace-microsoft, line 453: Name change "ItUsesFunctionInvokingHookToSkipFunctionExecution"

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -345,18 +345,272 @@ public async Task ItReturnsFunctionResultsCorrectlyAsync()
...
+        Assert.Equal(ExpectedInvocations, numberOfInvocations);
+        Assert.Equal(ExpectedInvocations, kernelResult.FunctionResults.Count);
+    }
+
+    [Theory]
+    [InlineData(1, 0, 0)]
+    [InlineData(2, 0, 0)]
+    [InlineData(2, 1, 1)]
+    [InlineData(5, 2, 2)]
+    public async Task ItCancelsPipelineFromFunctionInvokingEventsAsync(int numberOfFunctions, int functionCancelIndex, int expectedInvocations)
```
markwallace-microsoft, line 505: Name change "ItUsesFunctionInvokingHookToCancelFunctionExecution"

dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -209,9 +209,16 @@ public async Task<KernelResult> RunAsync(ContextVariables variables, Cancellatio
 
                 context = functionResult.Context;
 
+                var functionInvokedArgs = this.OnFunctionInvoked(functionDetails, functionResult);
+
+                if (functionInvokedArgs is not null)
```
markwallace-microsoft, line 214: I'd expect to only create a new function result when needed e.g. the function invoked hook may have done nothing

```
If (functionInvokedArgs is not null && functionInvokedArgs.HasNewResult())
{
    functionResult = functionInvokedArgs.GetNewResult();
}
```

markwallace-microsoft (commented):
Sorry for the delay but some suggestions to improve the code

PR #3039, opened 2023-10-02
Python: Context Variables like dict

### LINKED ISSUE #2962: Python: change variables getter to be more pytonic
I just noticed that the `.get` method of the `context.variables` object gives back a tuple[bool, str], this is very unpytonic (clearly straight port from the .net world), would be good to change this to the same signature as all other collection `.get` methods in Python, with None being returned if nothing is present, and the actual object if it is, potentially including optionally passing in a default so that it is even more predictable for a python dev.

eavanvalkenburg:
### Motivation and Context

It's shame sometimes it shines through that some of the earliest code in the python code base was a direct port from C#, like with the .get method on the context variables, made the change here to have the exact some signature as the default python dict.
Also replaces contains_key with __contains__ so that 'in' keyword can be used, just like a normal dict.
Fixes #2962 

### Description

Changed the .get method
Replaced the contains_key method with __contains__
Made changes in samples, tests and planner to deal with that.

### Contribution Checklist


alliscode:
@dluc would you mind taking a look at this please?

python/notebooks/05-using-the-planner.ipynb
```diff
@@ -513,7 +513,7 @@
     "        description=\"The search query\",\n",
     "    )\n",
     "    async def search_async(self, query: str, context: SKContext) -> str:\n",
-    "        query = query or context.variables.get(\"query\")[1]\n",
+    "        query = query or context.variables.get(\"query\")\n",
```
dluc, line 516: the notebook runs with **0.3.12.dev0** where get still works the old way. Doesn't this break the notebook?

dluc:
Note: this is a breaking change. 

* change of return types
* change of logic to interact with the context

python/semantic_kernel/core_skills/text_memory_skill.py
```diff
@@ -130,19 +115,10 @@ async def save_async(self, text: str, context: "SKContext") -> None:
         if context.memory is None:
             raise ValueError("Context has no memory")
 
-        collection = (
-            context.variables[TextMemorySkill.COLLECTION_PARAM]
-            if context.variables.contains_key(TextMemorySkill.COLLECTION_PARAM)
-            else TextMemorySkill.DEFAULT_COLLECTION
-        )
-        if not collection:
```
dluc, line 138: this check is still required, in case the value is an empty string

python/semantic_kernel/core_skills/text_memory_skill.py
```diff
@@ -61,30 +61,15 @@ async def recall_async(self, ask: str, context: "SKContext") -> str:
...
             raise ValueError("Context has no memory")
 
-        collection = (
-            context.variables[TextMemorySkill.COLLECTION_PARAM]
-            if context.variables.contains_key(TextMemorySkill.COLLECTION_PARAM)
-            else TextMemorySkill.DEFAULT_COLLECTION
+        collection = context.variables.get(
+            TextMemorySkill.COLLECTION_PARAM, TextMemorySkill.DEFAULT_COLLECTION
         )
-        if not collection:
```
dluc, line 69: this check is still required, in case the value is an empty string

python/semantic_kernel/core_skills/text_memory_skill.py
```diff
@@ -61,30 +61,15 @@ async def recall_async(self, ask: str, context: "SKContext") -> str:
...
-            raise ValueError("Memory collection not defined for TextMemorySkill")
-
-        relevance = (
-            context.variables[TextMemorySkill.RELEVANCE_PARAM]
-            if context.variables.contains_key(TextMemorySkill.RELEVANCE_PARAM)
-            else TextMemorySkill.DEFAULT_RELEVANCE
+        relevance = context.variables.get(
+            TextMemorySkill.RELEVANCE_PARAM, TextMemorySkill.DEFAULT_RELEVANCE
         )
-        if relevance is None or str(relevance).strip() == "":
```
dluc, line 77: this check is still required, in case the value is an empty string

python/semantic_kernel/core_skills/text_memory_skill.py
```diff
@@ -61,30 +61,15 @@ async def recall_async(self, ask: str, context: "SKContext") -> str:
...
-            relevance = TextMemorySkill.DEFAULT_RELEVANCE
-
-        limit = (
-            context.variables[TextMemorySkill.LIMIT_PARAM]
-            if context.variables.contains_key(TextMemorySkill.LIMIT_PARAM)
-            else TextMemorySkill.DEFAULT_LIMIT
+        limit = context.variables.get(
+            TextMemorySkill.LIMIT_PARAM, TextMemorySkill.DEFAULT_LIMIT
         )
-        if limit is None or str(limit).strip() == "":
```
dluc, line 85: this check is still required, in case the value is an empty string

dluc (changes requested):
Please see comments inline, there are some bugs in logic used for memory and prompt rendering

python/notebooks/05-using-the-planner.ipynb
eavanvalkenburg: good call, didn't notice the hardcoded version, is there a plan for how and when the notebooks upgrade version?

dluc (approved):

alexchaomander (approved):
Looks good to me! Thanks Eduard!

PR #3040, opened 2023-10-02
.Net: ADR with proposal for the multiple LLM support use cases

markwallace-microsoft:
### Motivation and Context

- ADR describing the design for multiple LLM support. This will be expanded in the future to allow additional use cases such as selecting an AI service by LLM model id.
- The `IAIServiceSelector` has changed so that it is added to the `IKernel` instance, see example 62 for an example of a custom `IAIServiceSelector` implementation.

### Contribution Checklist


docs/decisions/0012-semantic-function-multiple-model-support.md
```diff
@@ -0,0 +1,260 @@
...
+    .WithOpenAIChatCompletionService(modelId: oai.ChatModelId,
+        serviceId: "OpenAIChat", apiKey: oai.ApiKey, setAsDefault: true)
+    .Build();
+
+// Configure semantic function with multiple LLM request settings
+string configPayload = @"{
+  ""schema"": 1,
+  ""description"": ""Hello AI, what can you do for me?"",
+  ""models"": [
+    { ""service_id"": ""AzureText"", ""max_tokens"": 60 },
```
RogerBarreto, line 57: IMO, configurations should be agnostic of the `Service Id` but not from the `Service Type`.

Reason, plugins can be imported from other sources (with configurations that should be common to everyone). Using `types` instead of `ids` defined by the dev.

```
models: [ 
    { 
        service_type: "IChatCompletion", "max_tokens": 400
        // This one will accept any IChatCompletion available
    }, 
    { 
        service_type: "AzureChatCompletion", "max_tokens": 200
        // This one will override the above if the IChatCompletion setup is a `AzureChatCompletion` type.
    }
]

```

dmytrostruk: I'm not sure if naming `service_type: "IChatCompletion"` or `service_type: "AzureChatCompletion"` will align with Python or Java implementations, since it looks like `IChatCompletion` and `AzureChatCompletion` are C# interfaces/types. The name should be compatible for all SK versions, either it's `Service Id` or `Service Type`.

While I understand the purpose of `Service Id`, it would be nice to learn more about `Service Type` approach and its benefits.

markwallace-microsoft: The goal of using `service_id` is to provide an unambiguous link between the request settings and the service i.e., these are the settings for service `foobar`. There are drawbacks with this approach e.g., if you want to create a prompt that works in multiple environments but for that use case there are other solutions in the ADR.

RogerBarreto: > I'm not sure if naming service_type: "IChatCompletion" or service_type: "AzureChatCompletion" will align with Python or Java implementations,

You are right, we might consider having a global ServiceType and ServiceName across platforms.

Completion example:
Category: IChatCompletion -> "chat_completion"
Name: AzureChatCompletion -> "azure_chat_completion"

Memory example:
Category: IMemoryStore -> "memory_store"
Name: AzureCognitiveSearchMemoryStore-> "azure_cognitive_search_memory_store"

Suggestion:
```csharp
public interface IAIService 
{
    string Category { get; }
    string Name { get; }
}
```

Json:
```
models: [ 
    { 
        service_category: "chat_completion", "max_tokens": 400
        // This one will accept any IChatCompletion available
    }, 
    { 
        service_name: "azure_chat_completion", "max_tokens": 200
        // This one will override the above if the IChatCompletion setup is a `AzureChatCompletion` type.
    }
]
```

markwallace-microsoft: What you have called "Service Name" is the same as "Service Id". This is existing functionality so there is no need to change the name.

The proposal is to use "Model Id" rather than service category because it is more specific e.g., the request settings for GPT 3.5 and GPT 4 may be different event though both at chat completion services.

This ADR is specific to Request Settings for AI Models and Memory and other services are out of scope.

docs/decisions/0012-semantic-function-multiple-model-support.md
```diff
@@ -0,0 +1,190 @@
...
+    { ""model_id"": ""gpt-35-turbo"", ""max_tokens"": 120 },
+    { ""model_id"": ""gpt-3.5-turbo"", ""max_tokens"": 180 }
+  ]
+}";
+var templateConfig = JsonSerializer.Deserialize<PromptTemplateConfig>(configPayload);
+var func = kernel.CreateSemanticFunction(prompt, config: templateConfig!, "HelloAI");
+
+// Semantic function is executed with AI Service returned by custom AIService factory
+var funcOptions = new FunctionOptions();
+funcOptions.AddFactory<IAIService>(myAiServiceFactory);
```
SergeyMenshykh, line 146: I believe the proposed approach of passing an AI service factory as an argument/option of the Kernel.Run async method should be reevaluated and designed properly, with a separate ADR outlining how the design fits/aligns with the existing approach of AI service resolution. For instance, today, it's possible to supply your own implementation of the `IAIServiceProvider` interface when creating the Kernel, so it can easily be used as a factory if needed.

markwallace-microsoft: Yeah this is just pseudo code to illustrate the concept. I'll remove from the ADR.

docs/decisions/0012-semantic-function-multiple-model-support.md
markwallace-microsoft: Switching based on model id and service type is out of scope for this ADR

matthewbolanos:
Could we also support a scenario where the kernel keys off of "Model ID" instead of "Service ID"? In most cases, the semantic function author doesn't know the names of all the services that the kernel invoker has added, they'll just know the standard model names that are out in the wild (e.g., gpt-4, gpt-3.5-turbo, etc.). Leveraging "Model ID" should likely be the main/predominant scenario.

markwallace-microsoft:
> Could we also support a scenario where the kernel keys off of "Model ID" instead of "Service ID"? In most cases, the semantic function author doesn't know the names of all the services that the kernel invoker has added, they'll just know the standard model names that are out in the wild (e.g., gpt-4, gpt-3.5-turbo, etc.). Leveraging "Model ID" should likely be the main/predominant scenario.

Right now you could make the service if the same as the model id and that would work.

Currently matching specifically on model id is called out of scope because we have no way to store arbitrary attributes for a service. I'll add this as an additional task and will address in a follow up PR.

SergeyMenshykh:
> Could we also support a scenario where the kernel keys off of "Model ID" instead of "Service ID"? In most cases, the semantic function author doesn't know the names of all the services that the kernel invoker has added, they'll just know the standard model names that are out in the wild (e.g., gpt-4, gpt-3.5-turbo, etc.). Leveraging "Model ID" should likely be the main/predominant scenario.

I'm not sure, but it seems we need both the Service ID and Model ID to cover scenarios in which one connector provides multiple models and when the same model is provided by multiple connectors. Otherwise, the connector will have to be registered many times – one time per model it supports for the former scenario, or it won't be possible to register many connectors supporting the same model for the latter scenario.

markwallace-microsoft:
> > Could we also support a scenario where the kernel keys off of "Model ID" instead of "Service ID"? In most cases, the semantic function author doesn't know the names of all the services that the kernel invoker has added, they'll just know the standard model names that are out in the wild (e.g., gpt-4, gpt-3.5-turbo, etc.). Leveraging "Model ID" should likely be the main/predominant scenario.
> 
> I'm not sure, but it seems we need both the Service ID and Model ID to cover scenarios in which one connector provides multiple models and when the same model is provided by multiple connectors. Otherwise, the connector will have to be registered many times – one time per model it supports for the former scenario, or it won't be possible to register many connectors supporting the same model for the latter scenario.

So right now we can do this:
```csharp
IKernel kernel = new KernelBuilder()
    .WithAzureChatCompletionService(serviceId: "azure-gpt-4", deploymentName: "gpt-4", endpoint: endpoint, apiKey: apiKey)
    .WithAzureChatCompletionService(serviceId: "azure-gpt-3.5", deploymentName: "gpt-3.5", endpoint: endpoint, apiKey: apiKey)
    .WithOpenAIChatCompletionService(serviceId: "openai-gpt-4", modelId: "gpt-4", apiKey: openAIApiKey)
    .WithOpenAIChatCompletionService(serviceId: "openai-gpt-3.5", modelId: "gpt-3.5", apiKey: openAIApiKey)
    .Build();
```
I can write an IAIServiceSelector that picks the right service by service id
 
In the future we would like to be able to support this:
```csharp
IKernel kernel = new KernelBuilder()
    .WithAzureChatCompletionService(modelId: "gpt-4", deploymentName: "foo", endpoint: endpoint, apiKey: apiKey)
    .WithAzureChatCompletionService(modelId: "gpt-3.5", deploymentName: "bar", endpoint: endpoint, apiKey: apiKey)
    .WithOpenAIChatCompletionService(modelId: "gpt-4", apiKey: openAIApiKey)
    .WithOpenAIChatCompletionService(modelId: "gpt-3.5", apiKey: openAIApiKey)
    .Build();
```
We need to solve how to write an IAIServiceSelector that picks the right service by model id and service type

dotnet/src/SemanticKernel.Core/Functions/SemanticFunction.cs
```diff
@@ -177,19 +176,18 @@ private void AddDefaultValues(ContextVariables variables)
...
-        Verify.NotNull(client);
-
         FunctionResult result;
 
         try
         {
             string renderedPrompt = await this._promptTemplate.RenderAsync(context, cancellationToken).ConfigureAwait(false);
-            IReadOnlyList<ITextResult> completionResults = await client.GetCompletionsAsync(renderedPrompt, requestSettings, cancellationToken).ConfigureAwait(false);
+            (var textCompletion, var defaultRequestSettings) = this._serviceSelector.SelectAIService<ITextCompletion>(renderedPrompt, context.ServiceProvider, this._modelSettings);
+            Verify.NotNull(textCompletion);
```
SergeyMenshykh, line 189: nit: we can throw exception here with more details.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Core/Functions/SemanticFunction.cs
markwallace-microsoft: I've added a task for this as I want to test this and also create unit tests for this behaviour

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -64,18 +64,25 @@ public CultureInfo Culture
     /// </summary>
     public IAIServiceProvider ServiceProvider { get; }
```
SergeyMenshykh, line 65: The service provided may no longer need to be available through the context, right?

docs/decisions/0015-semantic-function-multiple-model-support.md
```diff
@@ -0,0 +1,172 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: approved
+contact@ markwallace-microsoft
+date: 2023-10-26
+deciders: markwallace-microsoft, semenshi, rbarreto
```
RogerBarreto, line 6: My github tag is different 😊
```suggestion
deciders: markwallace-microsoft, semenshi, rogerbarreto
```

docs/decisions/0015-semantic-function-multiple-model-support.md
```diff
@@ -0,0 +1,172 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: approved
+contact@ markwallace-microsoft
```
RogerBarreto, line 4: typo

```suggestion
contact: markwallace-microsoft
```

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
markwallace-microsoft: Yes, I'm going to make this internal

RogerBarreto (approved):
This should be marked as a Breaking Change after making `ServiceProvider` internal.

Issue #2828, opened 2023-09-14
Kernel Extensions Registration Plan


nacharya1:
Where are extension instances stored?
i.e. TemplateEngine, Memory Connectors, etc.

markwallace-microsoft:
Removing this from the v1.0.0.RC1 milestone as there will be no associated code changes. The PR for the ADR is here: https://github.com/microsoft/semantic-kernel/pull/3062

PR #3062, opened 2023-10-03
.Net: ADR for Kernel Service Registration

dmytrostruk:
### Motivation and Context

Resolves: https://github.com/microsoft/semantic-kernel/issues/2828

This PR contains ADR for Kernel Service Registration functionality, as initial step to decouple services like `IPromptTemplateEngine` and `ISemanticTextMemory` from `IKernel` and make them as part of service collection, to resolve Plugin dependencies.

### Contribution Checklist


docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+var services = serviceCollection.BuildServiceProvider();
+
+var kernel = Kernel.Builder
+    .WithLoggerFactory(ConsoleLogger.LoggerFactory)
+    .WithOpenAITextEmbeddingGenerationService(modelId, apiKey)
+    .WithServices(services) // Pass all registered services from host application to Kernel
+    .Build();
+
+// Plugin Import - option #1
+var semanticTextMemory = kernel.Services.GetService<ISemanticTextMemory>();
```
RogerBarreto, line 158: Suggest avoiding exposing ServiceProvider `Services` property in the Kernel.

For this specific `ISemanticTextMemory` we might have an extension method added from specifics Memory Connector packages like: 

`ISemanticTextMemory.CreateWithOpenAITextEmbeddingGenerationService([...])`
or
`ISemanticTextMemory.CreateWith(IMemoryStore any [, ...])`

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+var memoryPlugin = new TextMemoryPlugin(semanticTextMemory);
+
+kernel.ImportFunctions(memoryPlugin);
+
+// Plugin Import - option #2
+kernel.ImportFunctions<TextMemoryPlugin>();
```
RogerBarreto, line 164: Add a new example for scenarios like factory/`keyed services`
```suggestion
kernel.ImportFunctions<TextMemoryPlugin>();
kernel.ImportFunctions<TextMemoryPlugin>(provider => 
    new TextMemoryPlugin(provider.GetService<T>("service-name"))
);
```

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+```csharp
+var serviceCollection = new ServiceCollection();
+
+serviceCollection.AddTransient<IMemoryStore, VolatileMemoryStore>();
+serviceCollection.AddTransient<ITextEmbeddingGeneration>(
+    (serviceProvider) => new OpenAITextEmbeddingGeneration(modelId, apiKey));
+
+serviceCollection.AddTransient<ISemanticTextMemory, SemanticTextMemory>();
+
+var services = serviceCollection.BuildServiceProvider();
```
RogerBarreto, line 149: Since this is the configuration phase, pass the service collection, not the provider.
Provider will be instantiated by the kernel when needed.

```suggestion
```

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+var semanticTextMemory = kernel.Services.GetService<ISemanticTextMemory>();
+var memoryPlugin = new TextMemoryPlugin(semanticTextMemory);
+
+kernel.ImportFunctions(memoryPlugin);
+
+// Use this
+kernel.ImportFunctions<TextMemoryPlugin>();
+```
+
+### Solution #3
```
SergeyMenshykh, line 136: Whatever solution we decide to go with, whether it's a custom DI container or the MS DI container, I would propose to use it internally and not passing it to SK. This will help maintain an isolation boundary between SK and its consumer code. SK consumer code developers will be able to explicitly control the services that need to be passed to SK, which should improve code maintenance and readability. Additionally, SK consumer code that does not use a DI container will easily be able to use SK without the need to create a container.

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+Pros:
+
+- No implementation is required for dependency resolution - just use already existing .NET library.
+- The possibility to inject all registered services at once in already existing applications and use them as Plugin dependencies.
+
+Cons:
+
+- Additional dependency for Semantic Kernel package - `Microsoft.Extensions.DependencyInjection`.
+- No possibility to include specific list of services (lack of isolation from host application).
+- Possibility of `Microsoft.Extensions.DependencyInjection` version mismatch and runtime errors (e.g. users have `Microsoft.Extensions.DependencyInjection` `--version 2.0`  while Semantic Kernel uses `--version 6.0`)
```
stephentoub, line 176: This is only a problem if there are breaking changes between the versions, which we try really hard to avoid; otherwise, package resolution will handle picking an appropriate version everyone's happy with.

This is also the case for every dependency SK has, whether it be LoggingFactory or System.Text.Json or whatever.  DI isn't really special here.

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+    (serviceProvider) => new OpenAITextEmbeddingGeneration(modelId, apiKey));
+
+serviceCollection.AddTransient<ISemanticTextMemory, SemanticTextMemory>();
+
+var services = serviceCollection.BuildServiceProvider();
+
+var kernel = Kernel.Builder
+    .WithLoggerFactory(ConsoleLogger.LoggerFactory)
+    .WithOpenAITextEmbeddingGenerationService(modelId, apiKey)
+    .WithServices(services) // Pass all registered services from host application to Kernel
```
stephentoub, line 154: There's a middle-ground, too:
- Use the Microsoft.Extensions.DI container internally as the implementation for handling all of this stuff
- Continue to provide helpers that make the easy case easy (e.g. a WithService extension on the kernel)
- But allow for the container to be passed in as well

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+var kernel = Kernel.Builder.Build();
+
+kernel.ImportFunctions(memoryPlugin);
+```
+
+Note: in similar way as Solution #1.1, this way should be supported out of the box. Users always can handle all the dependencies on their side and just provide required Plugins to Kernel.
+
+### Solution #2.1
+
+Custom service collection and service provider on Kernel level to simplify dependency resolution process, as addition to Solution #1.1 and Solution #1.2.
```
stephentoub, line 79: In the fullness of time, isn't it likely we're just going to find ourselves implementing a full-blown DI container, as more and more requests come in for this or that feature that needs to be supported?

dluc:
In case you haven't, I would suggest testing this approach in combination with an ASP.NET application, where the kernel uses it's own service collection and the ASP.NET application a separate service collection. There are scenarios where the builder needs to interact with both, e.g. injecting kernel dependencies into the ASP.NET service collection, because customers are reusing internal components. In this case you might want to consider an implementation to IServiceCollection that proxies some calls to a pool of collections.

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+var semanticTextMemory = kernel.Services.GetService<ISemanticTextMemory>();
+var memoryPlugin = new TextMemoryPlugin(semanticTextMemory);
+
+kernel.ImportFunctions(memoryPlugin);
+```
+
+Pros:
+
+- No dependency on specific DI container library.
+- Lightweight implementation.
```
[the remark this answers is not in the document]
dmytrostruk: It depends on what set of features we would like to support :) But I agree, as more and more features come in, it will be more complicated.

docs/decisions/0011-kernel-service-registration.md
```diff
@@ -0,0 +1,180 @@
...
+    .WithLoggerFactory(ConsoleLogger.LoggerFactory)
+    .WithOpenAITextEmbeddingGenerationService(modelId, apiKey)
+    .WithServices(services) // Pass all registered services from host application to Kernel
+    .Build();
+
+// Plugin Import - option #1
+var semanticTextMemory = kernel.Services.GetService<ISemanticTextMemory>();
+var memoryPlugin = new TextMemoryPlugin(semanticTextMemory);
+
+kernel.ImportFunctions(memoryPlugin);
```
stephentoub, line 161: Do we have a definition somewhere for what we expect the "kernel" to actually be? Is it ending up just as a tuple of a few kinds of state that can be passed around together, e.g. a list of functions and a list of registered services?  If the kernel is moving to be this super lightweight thing that's just aggregating other resources, that to me suggests we definitely don't want it to be a full-blown DI container and instead just a wrapper for one.

dluc:
About the note on ISemanticMemory in the PR description, I would avoid that argument in this conversation, because the plan is to remove memory from the kernel. I think the main reasons here are how multiple LLMs can be loaded in the kernel, and reused outside the kernel, without hardcoded solutions and without duplicating class instances.

docs/decisions/0011-kernel-service-registration.md
dmytrostruk: Closing this comment, since proposed approach won't be used as per team decision.

docs/decisions/0011-kernel-service-registration.md
dmytrostruk: Closing this comment, since proposed approach won't be used as per team decision.

docs/decisions/0011-kernel-service-registration.md
dmytrostruk: Closing this comment, since proposed approach won't be used as per team decision.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

Issue #2280, opened 2023-08-02
Sequential plans not able to run multiple ChatGPT functions together


matthewbolanos:
**Describe the bug**
When asking Semantic Kernel to run a plan with multiple ChatGPT functions together, the plan ends up failing because it passes incorrect input values. For example, if you use the [Math plugin](https://learn.microsoft.com/en-us/semantic-kernel/ai-orchestration/chatgpt-plugins), to answer the ask "what is 2 times 4 minus 3", the plan makes the two following API requests:

/multiply?number1=2&number2=4
/subtract?number1={"content": "8","contentType": "text/plain"}&number2=4

In the second API, instead of _just_ passing 8, it passes in a full JSON string

**To Reproduce**
1. Open Chat Copilot
2. Add the Math plugin
3. Ask a multi-step math problem
4. Run the plan
5. See error

**Expected behavior**
Instead of passing {"content": "8","contentType": "text/plain"} to the second API, it should just pass the plain text value.

**Screenshots**
![image](https://github.com/microsoft/semantic-kernel/assets/1409190/3bc42bc1-4aa6-4b4d-8a32-61489a5ad3db)

**Additional context**
This is caused by [RestApiOperationRunner](https://github.com/microsoft/semantic-kernel/blob/bfb8b9c28701bb0a512903e51a339906e1712db4/dotnet/src/Skills/Skills.OpenAPI/RestApiOperationRunner.cs#L119) returning a serialized version of RestApiOperationResponse. At some point, the plan runner needs to remove the extra JSON schema when invoking the next function.

glahaye:
@teresaqhoang  May or may not be related to changes you made recently. Just putting your name here for your visibility.

ramarnat:
I am having the same issue for running a custom plugin, as in this issue - https://github.com/microsoft/chat-copilot/issues/422#issue-1919837761

Is there a way to override this behavior?

PR #3070, opened 2023-10-04
.Net [OpenApi] Converting OpenApi plugins result to string

SergeyMenshykh:
### Motivation and Context
This change is necessary to align the handling of OpenAPI plugin results with the current approach for handling results from native plugins. In this approach, a plugin result is converted to a string to be saved to SKContext if such a conversion is possible. This conversion enables planners to use the result of the first plugin as input for the next one because the result of the first plugin is a string rather than a complex type like an object or JSON.

### Description
- The type of the `RestApiOperationResponse.Content` property has been changed from a string to an object to represent different content types such as JSON, XML, text, images, etc.
- The `RestApiOperationRunner` class has been modified to handle and serialize different content types differently. For example, JSON, text, and XML content are read and serialized using the `HttpContent.ReadAsStringAsync` method, while images are read and serialized using the `HttpContent.ReadAsByteArrayAsync` method.
- A new `RestApiOperationResponseConverter` class has been added to convert the `RestApiOperationResponse` to a string so that it can be used by any other plugin in the pipeline.
- A small regression in the `Plan` class has been fixed. This regression caused a failure in the `StepwisePlanner` if a plugin returned a non-string result.

Fixes the issue - https://github.com/microsoft/semantic-kernel/issues/2280

### Contribution Checklist


dotnet/src/Functions/Functions.UnitTests/OpenAPI/RestApiOperationResponseConverterTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    public RestApiOperationResponseConverterTests()
+    {
+        this._sut = new RestApiOperationResponseConverter();
+    }
+
+    [Fact]
+    public void ItShouldConvertStringContentToString()
+    {
+        //Arrange
+        var response = new RestApiOperationResponse("fake-content", "fake-content-type");
```
markwallace-microsoft, line 20: Should the content type be `text/plain` here?

markwallace-microsoft (commented):
I was expecting to see an integration test which reproduces the current failure that now passes. Can this be added?

dotnet/src/Functions/Functions.UnitTests/OpenAPI/RestApiOperationResponseConverterTests.cs
SergeyMenshykh: The content type does not matter with the current implementation of the converter, which uses the type of the value of the `Content` property to decide on a conversion mechanism, as opposed to doing so based on the content type. However, the converter implementation can and will be changed to perform conversion based on the content type when and if needed.

SergeyMenshykh:
> I was expecting to see an integration test which reproduces the current failure that now passes. Can this be added?
Yep, make sense, done.

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
```diff
@@ -29,6 +29,17 @@ internal sealed class RestApiOperationRunner
...
 
+    /// <summary>
+    /// A dictionary containing the content type as the key and the corresponding content serializer as the value.
+    /// </summary>
+    private static readonly Dictionary<string, HttpResponseContentSerializer> s_serializerByContentType = new()
+    {
+        { "image", async (content) => await content.ReadAsByteArrayAndTranslateExceptionAsync().ConfigureAwait(false) },
+        { "text", async (content) => await content.ReadAsStringWithExceptionMappingAsync().ConfigureAwait(false) },
+        { "application/json", async (content) => await content.ReadAsStringWithExceptionMappingAsync().ConfigureAwait(false)},
+        { "application/xml", async (content) => await content.ReadAsStringWithExceptionMappingAsync().ConfigureAwait(false)}
```
dmytrostruk, line 40: nit: Looks like there is no serialization difference between `application/json` and `application/xml`. There is also logic which fetches primary content type `image`, `text`, `application` and tries to get serializer by this type. Wondering, if we can leave just `application` primary content type in this dictionary.

dmytrostruk (approved):

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
SergeyMenshykh: Yeah, I initially considered registering a single serialization method using only the primary content-type 'application'. However, I later changed my mind and decided to register only the known ones for now, as there are many different 'application' subtypes for which it is not clear how to serialize - https://www.iana.org/assignments/media-types/media-types.xhtml#application."

dmytrostruk: Makes sense, thank you!

markwallace-microsoft (approved):

Issue #3081, opened 2023-10-05
.Net: Transition to Polly.Core


anthonypuppo:
Needs to be done after #3079.

The `Polly` package supports both v7 and v8 API's. The new `Polly.Core` package only supports v8 and is recommended for new projects (see [getting started](https://www.pollydocs.org/getting-started.html) and [migration guide](https://www.pollydocs.org/migration-v8.html)).

stephentoub:
The reliability support in SK has been replaced by relying on standard DI mechanisms, eg you can now use Microsoft.Extensions.Http.Resilience with SK, and it supports the latest versions of polly.

PR #3082, opened 2023-10-05
.Net: Add overloads for Polly v8 ResiliencePipeline and update associated tests/examples

### LINKED ISSUE #3079: .Net: Add overloads for Polly v8 ResiliencePipeline
Polly v8 was [officially released](https://www.thepollyproject.org/2023/09/28/polly-v8-officially-released/) with an associated [migration guide](https://www.pollydocs.org/migration-v8.html). The `Reliability.Polly` project should be updated to support the newly released `ResiliencePipeline`.

anthonypuppo:
### Motivation and Context
With the update of `Polly` to v8:
- The term Policy is now replaced with Strategy
- Introduction of Resilience Pipelines
- [Full announcement](https://www.thepollyproject.org/2023/09/28/polly-v8-officially-released/)

Fixes #3079.

### Description
Add support for the new v8 API:
- Update constructor/method overloads and examples
- Add v8 tests
- Add `Obsolete` to v7 constructors/methods with a link to the v8 migration guide
- Enables `Reliability.Polly` to be gracefully transitioned from `Polly` to `Polly.Core` (#3081).

While the `Polly` package supports both v7 and v8 `Polly.Core` only has support for v8. Due to this potential future transition, I marked the v7 related code as obsolete, but I can remove those attributes if need be.

### Contribution Checklist


dmytrostruk (commented):
@anthonypuppo Thank you for contribution! We will have to make sure that we don't introduce new warnings. 

We can:
- Remove the usage of obsolete code.
- Temporarily suppress warnings in places where obsolete code is used, until obsolete code will be removed in future versions of SK.

Also, next week we plan to release new version of SK and we will upgrade major version of package, which means that you can remove obsolete code now, and we will mark this PR as breaking change.

dmytrostruk:
> Sounds good. As far as v7 compatibility goes, if the goal is to continue supporting it alongside v8, I can just suppress the warnings for now. On the other hand, if there's no objection, since this PR is breaking already breaking I can just transition to `Polly.Core` (#3081)?

While v8 is recommended for new projects, I think we still should be backward-compatible, so it makes sense to support v7 and v8 for some period of time. If this is achievable, that would be awesome.

anthonypuppo:
> > Sounds good. As far as v7 compatibility goes, if the goal is to continue supporting it alongside v8, I can just suppress the warnings for now. On the other hand, if there's no objection, since this PR is breaking already breaking I can just transition to `Polly.Core` (#3081)?
> 
> While v8 is recommended for new projects, I think we still should be backward-compatible, so it makes sense to support v7 and v8 for some period of time. If this is achievable, that would be awesome.

Updated `PollyHttpRetryHandlerFactory` and `PollyHttpRetryHandlerTests` with the below to suppress the warnings for now.

```csharp
#pragma warning disable CS0618 // PollyHttpRetryHandler.PollyHttpRetryHandler(AsyncPolicy)' is obsolete: 'Use one of the constructors that takes a ResiliencePipeline instead (https://www.pollydocs.org/migration-v8.html).
```

Both v7 and v8 are usable and consumers still relying on the v7 `Policy` will get an obsolete warning.

Issue #3076, opened 2023-10-05
.Net: C# TextChunker customizability


rnortje-sygnia:
The current TextChunker uses the following hardcoded separators to split lines.  Defined as:

private static readonly string?[] s_plaintextSplitOptions = new[] { "\n\r", ".", "?!", ";", ":", ",", ")]}", " ", "-", null };
private static readonly string?[] s_markdownSplitOptions = new[] { ".", "?!", ";", ":", ",", ")]}", " ", "-", "\n\r", null };

The public methods does not allow one to specify your own separators. The above is problematic for instances like ours where the decimal point in numbers is a ".". Therefore given text like "The return of X is 0.5%" the chunker will see "The return of X is 0." as a line.  This causes some paragraphs to cut off at critical points. Though this can be solved to an extend by overlaps, we have various instances where this causes vectors on content to cut off like this. 

Is it possible that public functions be made available where these can be specified by the caller. What would be nice is if they remain strings and not char. we can then specify ". ". i.e a full stop followed by a space as a valid sentence breakpoint. This will make the TextChunker much more usable. It would also be nice if this can be specified as part of the runtime configuration settings.

alliscode:
Hi @rnortje-sygnia thanks for submitting this. Would you be interested in submitting a PR to fix this issue?

rnortje-sygnia:
Would love to but do not have the capacity to do this at this stage, at least not in a reasonable timeframe.

matthewbolanos:
We'll be removing the text chunker from v1.0.0 of Semantic Kernel, so I'm closing this issue.

matthewbolanos:
I may have spoken too soon! We're reopening this issue and hoping to find a new home for the TextChunker. In the meantime, we'll be marking as experimental.

markwallace-microsoft:
All .Net issues prior to 1-Dec-2023 are being closed. Please re-open, if this issue is still relevant to the .Net Semantic Kernel 1.x release. In the future all issues that are inactive for more than 90 days will be labelled as 'stale' and closed 14 days later.

PR #3091, opened 2023-10-06
.Net: C# TextChunker customizability

feiyun0112:
#3076 

Add splitOptions parameter

### DIFF
--- a/dotnet/src/SemanticKernel.Core/Text/TextChunker.cs
+++ b/dotnet/src/SemanticKernel.Core/Text/TextChunker.cs
@@ -32,12 +32,13 @@ public static class TextChunker
     /// <param name="text">Text to split</param>
     /// <param name="maxTokensPerLine">Maximum number of tokens per line.</param>
     /// <param name="tokenCounter">Function to count tokens in a string. If not supplied, the default counter will be used.</param>
+    /// <param name="splitOptions">Options to split string. If not supplied, the default options will be used.</param>
     /// <returns>List of lines.</returns>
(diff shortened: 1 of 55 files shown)


markwallace-microsoft (commented):
@feiyun0112  Thanks for this PR this looks like a good additional to the SK.

Could you add some unit tests to `TextChunkerTests.cs`, this will help us to ensure this functionality does not regress.

dotnet/src/SemanticKernel.Core/Text/TextChunker.cs
```diff
@@ -32,12 +32,13 @@ public static class TextChunker
     /// <param name="text">Text to split</param>
     /// <param name="maxTokensPerLine">Maximum number of tokens per line.</param>
     /// <param name="tokenCounter">Function to count tokens in a string. If not supplied, the default counter will be used.</param>
+    /// <param name="splitOptions">Options to split string. If not supplied, the default options will be used.</param>
     /// <returns>List of lines.</returns>
-    public static List<string> SplitPlainTextLines(string text, int maxTokensPerLine, TokenCounter? tokenCounter = null)
+    public static List<string> SplitPlainTextLines(string text, int maxTokensPerLine, TokenCounter? tokenCounter = null, string[]? splitOptions = null)
```
WeihanLi, line 37: This would be a kind of breaking change, if we're happy to accept breaking changes, think maybe we could keep the `tokenCounter` as the last parameter since it seems would be used less frequently

```suggestion
    public static List<string> SplitPlainTextLines(string text, int maxTokensPerLine, string[]? splitOptions = null, TokenCounter? tokenCounter = null)
```

PR #3092, opened 2023-10-06
.Net: Memory Plugin extraction - Part 2

dmytrostruk:
### Motivation and Context

This PR contains changes to mark Memory-related functionality in Kernel as Obsolete and create alternative methods in `Plugins.Memory` and Connectors projects. 

Note: Obsolete methods and properties will be removed in future releases.

ADR: [0011-memory-as-plugin.md](https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0011-memory-as-plugin.md)

**Main change**: helper methods for Memory were moved from `KernelBuilder` to `MemoryBuilder`. Previously, it was possible to configure only one Memory instance per Kernel instance. Now, Memory functionality is decoupled from Kernel and should be configured separately.

Previous Memory usage:
```csharp
var kernel = Kernel.Builder
    .WithOpenAITextEmbeddingGenerationService("text-embedding-ada-002", TestConfiguration.OpenAI.ApiKey)
    .WithMemoryStorage(new AzureCognitiveSearchMemoryStore(TestConfiguration.ACS.Endpoint, TestConfiguration.ACS.ApiKey))
    .Build();

kernel.Memory.SaveInformationAsync(...);
```

New Memory usage:
```csharp
var memory = new MemoryBuilder()
    .WithOpenAITextEmbeddingGenerationService("text-embedding-ada-002", TestConfiguration.OpenAI.ApiKey)
    .WithMemoryStore(new AzureCognitiveSearchMemoryStore(TestConfiguration.ACS.Endpoint, TestConfiguration.ACS.ApiKey))
    .Build();

memory.SaveInformationAsync(...);
```

### Description

1. Added `MemoryBuilder` as equivalent to `KernelBuilder` to initialize the logic related to Memory.
2. Added unit tests for `MemoryBuilder`.
3. Added extension methods for `MemoryBuilder` related to Memory and AI Connectors.
4. Marked Memory-related methods and properties in Kernel/KernelBuilder as obsolete.

### Contribution Checklist

### DIFF
--- a/dotnet/samples/KernelSyntaxExamples/Example12_SequentialPlanner.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example12_SequentialPlanner.cs
@@ -4,6 +4,7 @@ using System;
 using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Connectors.AI.OpenAI.TextEmbedding;
 using Microsoft.SemanticKernel.Diagnostics;
 using Microsoft.SemanticKernel.Memory;
 using Microsoft.SemanticKernel.Planners;
@@ -298,7 +299,7 @@ internal static class Example12_SequentialPlanner
     {
         var memoryStorage = new VolatileMemoryStore();
 
-        var textEmbeddingGenerator = new Microsoft.SemanticKernel.Connectors.AI.OpenAI.TextEmbedding.AzureTextEmbeddingGeneration(
+        var textEmbeddingGenerator = new AzureTextEmbeddingGeneration(
             modelId: TestConfiguration.AzureOpenAIEmbeddings.DeploymentName,
             endpoint: TestConfiguration.AzureOpenAIEmbeddings.Endpoint,
             apiKey: TestConfiguration.AzureOpenAIEmbeddings.ApiKey);
--- a/dotnet/samples/KernelSyntaxExamples/Example12_SequentialPlanner.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example12_SequentialPlanner.cs
@@ -149,7 +149,7 @@ internal static class Example12_SequentialPlanner
         Console.WriteLine("======== Sequential Planner - Find and Execute Saved Plan ========");
 
         // Save the plan for future use
-        var semanticMemory = GetMemory();
+        var semanticMemory = InitializeMemory();
         await semanticMemory.SaveInformationAsync(
             "plans",
             id: Guid.NewGuid().ToString(),
@@ -228,7 +228,8 @@ internal static class Example12_SequentialPlanner
     {
         Console.WriteLine("======== Sequential Planner - Create and Execute Plan using Memory ========");
 
-        var kernel = InitializeKernelWithMemory();
+        var kernel = InitializeKernel();
+        var memory = InitializeMemory();
 
         string folder = RepoFiles.SamplePluginsPath();
         kernel.ImportSemanticFunctionsFromDirectory(folder,
@@ -251,7 +252,7 @@ internal static class Example12_SequentialPlanner
         var goal = "Create a book with 3 chapters about a group of kids in a club called 'The Thinking Caps.'";
 
         // IMPORTANT: To use memory and embeddings to find relevant plugins in the planner, set the 'Memory' property on the planner config.
-        var planner = new SequentialPlanner(kernel, new SequentialPlannerConfig { SemanticMemoryConfig = new() { RelevancyThreshold = 0.5, Memory = kernel.Memory } });
+        var planner = new SequentialPlanner(kernel, new SequentialPlannerConfig { SemanticMemoryConfig = new() { RelevancyThreshold = 0.5, Memory = memory } });
 
         var plan = await planner.CreatePlanAsync(goal);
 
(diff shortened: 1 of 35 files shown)


dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -190,6 +167,19 @@ public interface IKernel
 
     #region Obsolete
 
+    /// <summary>
+    /// Semantic memory instance
+    /// </summary>
+    [Obsolete("Memory functionality will be placed in separate Microsoft.SemanticKernel.Plugins.Memory package. This will be removed in a future release.")]
```
markwallace-microsoft, line 173: We should include some information so a developer know what to do. Can we include a link to a kernel syntax example here?

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -201,5 +191,20 @@ public interface IKernel
...
+#pragma warning disable CS1591
+    IDictionary<string, ISKFunction> ImportSkill(object functionsInstance, string? pluginName = null);
+#pragma warning restore CS1591
+
+    /// <summary>
+    /// Set the semantic memory to use
+    /// </summary>
+    /// <param name="memory">Semantic memory instance</param>
+    /// <inheritdoc/>
+    [Obsolete("Memory functionality will be placed in separate Microsoft.SemanticKernel.Plugins.Memory package. This will be removed in a future release.")]
```
markwallace-microsoft, line 205: Again we need to provide sufficient information so a developer knows how to change from the obsolete API.

dotnet/src/Plugins/Plugins.Memory/MemoryBuilder.cs
```diff
@@ -0,0 +1,122 @@
...
+        this._memoryStoreFactory = () => store;
+        return this;
+    }
+
+    /// <summary>
+    /// Add memory store factory.
+    /// </summary>
+    /// <param name="factory">The store factory.</param>
+    /// <returns>Updated Memory builder including the memory store.</returns>
+    public MemoryBuilder WithMemoryStore<TStore>(Func<ILoggerFactory, TStore> factory) where TStore : IMemoryStore
```
markwallace-microsoft, line 79: Should this be named `WithMemoryStoreFactory`?

dotnet/samples/KernelSyntaxExamples/Example12_SequentialPlanner.cs
```diff
@@ -288,24 +289,22 @@ private static IKernel InitializeKernelWithMemory()
...
-    private static ISemanticTextMemory GetMemory(IKernel? kernel = null)
+    private static SemanticTextMemory InitializeMemory()
     {
-        if (kernel is not null)
-        {
-            return kernel.Memory;
-        }
         var memoryStorage = new VolatileMemoryStore();
+
         var textEmbeddingGenerator = new Microsoft.SemanticKernel.Connectors.AI.OpenAI.TextEmbedding.AzureTextEmbeddingGeneration(
```
markwallace-microsoft, line 301: Can we just use `new AzureTextEmbeddingGeneration(`?

markwallace-microsoft (commented):
I've added the breaking change tag to this PR as we expect developers to make a code change and this will get the PR listed in the release notes accordingly.

We want to make it easy for developers to understand and adopt the new pattern. Can you:

1. Update the PR description to that includes some instructions on how developers should change their code. It should read like a step by step migration guide. 
1. Add sufficient information to the obsolete message so developers know what they need to change. We don't have a good pattern here but maybe a link to a sample, we may even want dedicated samples that show how to change the code i.e. with before and after sample code.

dotnet/src/Plugins/Plugins.Memory/MemoryBuilder.cs
dmytrostruk: While we have `WithLoggerFactory(ILoggerFactory loggerFactory)`, where we pass `Factory` interface, I don't think it's the case for MemoryStore. The idea here is to create one `WithMemoryStore` method which will have 3 different overloads:
1. `IMemoryStore store` - pass MemoryStore instance as it is.
2. `Func<ILoggerFactory, TStore> factory` - factory for MemoryStore initialization with access to ILoggerFactory.
3. `Func<ILoggerFactory, IDelegatingHandlerFactory, TStore> factory` - factory for MemoryStore initialization with access to ILoggerFactory and IDelegatingHandlerFactory.

Not sure if different name for methods that accept delegate will provide some benefits.

dotnet/samples/KernelSyntaxExamples/Example14_SemanticMemory.cs
```diff
@@ -35,13 +36,19 @@ public static async Task RunAsync()
          * need to worry about embedding generation.
          */
 
-        var kernelWithACS = Kernel.Builder
-            .WithLoggerFactory(ConsoleLogger.LoggerFactory)
+        var kernel = Kernel.Builder
             .WithOpenAITextEmbeddingGenerationService("text-embedding-ada-002", TestConfiguration.OpenAI.ApiKey)
```
SergeyMenshykh, line 40: Does Kernel still need the WithOpenAITextEmbeddingGenerationService extension method?

SergeyMenshykh (approved):

dotnet/samples/KernelSyntaxExamples/Example14_SemanticMemory.cs
dmytrostruk: In theory Kernel doesn't need it, but it's still possible to register multiple embedding generation services and then get a specific service with `kernel.GetService<ITextEmbeddingGeneration>(name)`. Removing `.GetService` capability from Kernel and related extension methods should be part of separate PR I think.

markwallace-microsoft (approved):

Issue #3120, opened 2023-10-10
.Net: After upgrade to 1.0, InvokeAsync returns FunctionResult with no way to get SKContext?


wrharper-AASP:
![image](https://github.com/microsoft/semantic-kernel/assets/117128832/eac18f1e-3f55-4327-a067-c331d292bea5)
Shouldn't this be public? How do you get the SKContext or a ToString with this new way as of 1.0 beta now?

wrharper-AASP:
This is needed because of my token encoder:
How it was: SKContext result = await chatFunction.InvokeAsync(context);
tikToken.EncodeCount(result.ToString())

New way: FunctionResult fresult = await chatFunction.InvokeAsync(context);
tikToken.EncodeCount(fresult.?????)

wrharper-AASP:
Context should be public because then it would be possible to do:
New way: SKContext result = (await chatFunction.InvokeAsync(context)).Context;
tikToken.EncodeCount(result.ToString())
and things would still work....

wrharper-AASP:
Alternatively, create a (await chatFunction.InvokeAsync(context)).GetTokenCount or something like that?

chatFunction is a ISKFunction

wrharper-AASP:
also, what's up with this?
var semanticFunc = kernel.CreateSemanticFunction(message + memorySearch);
//var result = await semanticFunc.InvokeAsync();
var fresult = await semanticFunc.InvokeAsync(kernel);
Why can't you just keep it await semanticFunc.InvokeAsync(); ??? it should just use its own kernel that it is called from?

dmytrostruk:
Hi @wrharper-AASP !

In case your `chatFunction` returns string, then you can access your result as:
```csharp
FunctionResult fresult = await chatFunction.InvokeAsync(context);
tikToken.EncodeCount(fresult.GetValue<string>());
```

`SKContext` as a result type was replaced with `FunctionResult` in case you invoke function directly or `KernelResult` in case you invoke function through Kernel instance.

A couple of reasons for that:
- We started to see some patterns of `SKContext` usage which is different from initial `SKContext` purpose. That's why we try to make it as internal mechanism rather than public input/output object.
- `SKContext` supports only `string` as return data type. With `FunctionResult`, you can access not only strings, but any kind of data, including complex types as well as streaming (`IAsyncEnumerable<T>`).

By following link, you can find ADR with more details about this change:
[...]

PR #3170, opened 2023-10-12
.Net: Added ToString override for FunctionResult and KernelResult

dmytrostruk:
### Motivation and Context

Resolves: https://github.com/microsoft/semantic-kernel/issues/3120

This PR contains `ToString` method override for `FunctionResult` and `KernelResult` models. It will allow to work with result models in easier way when Value object has meaningful string representation.

For example, instead of this:
```csharp
var result = kernel.RunAsync(...);
Console.WriteLine(result.GetValue<string>());
```

It will be possible to do this (as previously it was possible with `SKContext`):
```csharp
var result = kernel.RunAsync(...);
Console.WriteLine(result);
```

This should simplify the process of experimentation, prototyping, logging and debugging.

### Description

1. Added `ToString` override for result models.
2. Added unit tests.

### Contribution Checklist


stephentoub (approved):
Thanks :)

Issue #3152, opened 2023-10-11
.Net: MinHeap and TopNCollection should not be public


stephentoub:
> [!IMPORTANT]  
> Labeled **Urgent** because it may require a breaking change.

https://github.com/microsoft/semantic-kernel/issues/615 was closed as completed with (I thought) the resolution to make MinHeap internal, but in the latest release it's still public.  Both it and TopNCollection aren't something that are really in the domain of SK: a developer can use `PriorityQueue<>` in .NET, or `OrderBy(...).Take(...)`, or any number of other solutions if they need the equivalent.

dmytrostruk:
@stephentoub We moved them to separate package `Microsoft.SemanticKernel.Plugins.Memory` and made them public as part of this package. One of the reasons is to be able to write some unit tests in separate test package. Otherwise, we will need to add `InternalsVisibleTo` attribute, and I'm not sure it's a good practice for testing internal classes/structures.

Do you have any recommendations here?

stephentoub:
What are the unit tests doing with MinHeap/TopNCollection that they can't do with PriorityQueue, OrderBy.Take, etc.?

dmytrostruk:
> What are the unit tests doing with MinHeap/TopNCollection that they can't do with PriorityQueue, OrderBy.Take, etc.?

Sorry, I meant unit tests for actual MinHeap/TopNCollection types :)
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Plugins/Plugins.UnitTests/Memory/Collections/MinHeapTests.cs
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Plugins/Plugins.UnitTests/Memory/Collections/TopNCollectionTests.cs

`TopNCollecton` is used in `VolatileMemoryStore` - our in-memory store implementation. Since `Plugins.Memory` package is based on `.NET Standard 2.0`, `PriorityQueue` is unavailable, unless we copy all its source code and place it in our repository.

dmytrostruk:
Makes sense, thank you! Will use `InternalsVisibleTo` flag then.

PR #3175, opened 2023-10-13
.Net: Hide Memory Collection types for internal usage only

dmytrostruk:
### Motivation and Context

Resolves: https://github.com/microsoft/semantic-kernel/issues/3152

This PR contains changes to make Memory Collection types as `internal`. As alternative, developers can use [PriorityQueue](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2?view=net-7.0) type, available in .NET 6, 7, 8.

### Description

1. Hide Memory Collection types.
2. Removed unit tests for hidden types, as they are covered with tests for public type that uses them (`VolatileMemoryStore`).
### Contribution Checklist


dotnet/src/Plugins/Plugins.UnitTests/Memory/Collections/MinHeapTests.cs
SergeyMenshykh, line 1: I would propose keeping the unit tests for the `MinHeap` and `TopNCollection` classes because tests for public functionality that use them may not cover all code execution paths in the classes.

dmytrostruk: If we add `InternalsVisibleTo` attribute, in this case methods from `InternalUtilities/Linq/AsyncEnumerable` will become available for test project as well. Taking into account that test project uses .NET 6 and it has its own `AsyncEnumerable` methods, there will be collision between our methods and methods from .NET 6. Both extension methods use the same `System.Linq` namespace, so there is no way how to differentiate them in the code.

Another way would be to access `internal` classes with reflection. If we are okay with this approach, I will add it then. Let me know what you think about it.

markwallace-microsoft (approved):

SergeyMenshykh (approved):

Issue #3165, opened 2023-10-12
.Net: Rename SKFunction.From methods


stephentoub:
SKFunction has two methods: FromNativeMethod and FromNativeFunction. The "Native" part here is confusing in the context of C#, where "native" is typically used in contrast with "managed".  Can we drop the "Native" part of these names? Also, can we just name them both "FromMethod" and have them be overloads of each other? Whether they're taking a Delegate or a MethodInfo (with optional object) doesn't seem particularly meaningful: in both cases, there's a .NET method we're wrapping an ISKFunction around.

PR #3178, opened 2023-10-14
Rename NativeFunction to MethodLevelFunction

feiyun0112:
#3165 

Rename NativeFunction to MethodLevelFunction

Rename FromNativeMethod and FromNativeFunction to FromMethod

### DIFF
--- a/dotnet/src/SemanticKernel.Core/Functions/MethodLevelFunction.cs
+++ b/dotnet/src/SemanticKernel.Core/Functions/MethodLevelFunction.cs
@@ -33,7 +33,7 @@ namespace Microsoft.SemanticKernel;
 /// with additional methods required by the kernel.
 /// </summary>
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
-internal sealed class NativeFunction : ISKFunction, IDisposable
+internal sealed class MethodLevelFunction : ISKFunction, IDisposable
 {
     /// <inheritdoc/>
     public string Name { get; }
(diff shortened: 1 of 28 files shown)


dotnet/src/SemanticKernel.Core/Functions/MethodLevelFunction.cs
```diff
@@ -33,7 +33,7 @@ namespace Microsoft.SemanticKernel;
 /// with additional methods required by the kernel.
 /// </summary>
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
-internal sealed class NativeFunction : ISKFunction, IDisposable
+internal sealed class MethodLevelFunction : ISKFunction, IDisposable
```
dmytrostruk, line 36: I would like to understand the reasons for this change, if possible. `Native Function` is domain naming which we currently use, and as far as I know, we don't have any plans at the moment to change it. This kind of renaming should be done in separate PR, powered by additional documentation (ADR). But in all cases, I'm not sure it's on our radar at the moment.

The issue, which is attached to this PR, is about methods `SKFunction.FromNativeMethod` and `SKFunction.FromNativeFunction`, which could be renamed to `FromMethod`.

dmytrostruk (changes requested):
@feiyun0112 Thank you for contribution!
The changes in this PR have a wider scope than linked issue. If possible, let's do just renaming for methods `SKFunction.FromNativeMethod` and `SKFunction.FromNativeFunction` to `FromMethod`. 

We can have separate conversations about `NativeFunction` class renaming, if this is something that we want. Thanks again!

SergeyMenshykh:
Hi @feiyun0112 thanks for helping us improve SK. Your assistance is greatly appreciated.

I believe that the renaming of the 'NativeFunction' class should be part of a broader design discussion that takes into account all the other 'SKFunction,' 'SemanticFunction,' and 'ISKFunctions' interfaces/classes. This is to ensure that the SK function model is simple, clear, and cohesive. I would suggest not renaming the class until this discussion occurs.

Issue #3169, opened 2023-10-12
.Net: Can we remove Kernel.Builder?


stephentoub:
The property is just defined as:
```C#
public static KernelBuilder Builder => new();
```

Two problems:
1) This creates two different ways to do the exact same thing: `Kernel.Builder` vs `new KernelBuilder()`.
2) Properties like this are generally expected to return singletons, which this is not nor can it be since it's mutable.

Can we just delete the static `Builder` property?

markwallace-microsoft:
We should obsolete this to start with

PR #3180, opened 2023-10-14
.Net: Remove Kernel.Builder

feiyun0112:
Resolves: #3169 
Use `new KernelBuilder()` instead of `Kernel.Builder`


dotnet/samples/KernelSyntaxExamples/Example42_KernelBuilder.cs
```diff
@@ -2,7 +2,7 @@
 
 // ==========================================================================================================
 // The easier way to instantiate the Semantic Kernel is to use KernelBuilder.
-// You can access the builder using either Kernel.Builder or KernelBuilder.
+// You can access the builder using either new KernelBuilder() or KernelBuilder.
```
dmytrostruk, line 5: I think now there will be only one way how to access `KernelBuilder` - it's `new KernelBuilder()`. I think this comment can be removed.
```suggestion
```

dotnet/src/Plugins/Plugins.Core/ConversationSummaryPlugin.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.SemanticKernel.Plugins.Core;
 /// </summary>
 /// <example>
 /// <code>
-/// var kernel Kernel.Builder.Build();
+/// var kernel new KernelBuilder().Build();
```
dmytrostruk, line 17:
```suggestion
/// var kernel = new KernelBuilder().Build();
```

dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -41,11 +41,6 @@ public sealed class Kernel : IKernel, IDisposable
     /// <inheritdoc/>
     public IPromptTemplateEngine PromptTemplateEngine { get; }
 
-    /// <summary>
-    /// Return a new instance of the kernel builder, used to build and configure kernel instances.
-    /// </summary>
-    public static KernelBuilder Builder => new();
```
dmytrostruk, line 47: This is definitely breaking change. I would propose to keep all changes in this PR as it is but mark this field as `Obsolete` for some period of time, so users will be able to migrate to another syntax smoothly. 
```csharp
/// <summary>
/// Return a new instance of the kernel builder, used to build and configure kernel instances.
/// </summary>
[Obsolete("This field will be removed in a future release. Initialize KernelBuilder through constructor instead (new KernelBuilder()).")]
public static KernelBuilder Builder => new();
```

dmytrostruk (commented):
@feiyun0112 Thanks a lot! 
I left a couple of comments to resolve.

samples/dotnet/OpenApiPluginsExample/Program.cs
```diff
@@ -51,7 +51,7 @@ private static async Task Main(string[] args)
         AIServiceOptions aiOptions = configuration.GetRequiredSection(AIServiceOptions.PropertyName).Get<AIServiceOptions>()
             ?? throw new InvalidOperationException($"Missing configuration for {AIServiceOptions.PropertyName}.");
 
-        KernelBuilder kb = Kernel.Builder
+        KernelBuilder kb = new KernelBuilder()
```
RogerBarreto, line 54:
```suggestion
        KernelBuilder builder = new KernelBuilder()
```

RogerBarreto (approved):
Fix conflicts, LGTM

dmytrostruk (approved):

Issue #3046, opened 2023-10-02
Python: Cannot connect to remote Chroma DB


snelson3:
**Describe the bug**
The ChromaMemoryStore class uses an outdated method of connecting to a remote chroma DB, which does not work with any recently released version of ChromaDB

**To Reproduce**
Run the following Python code with the most current versions of semantic_kernel and chromadb that are available on pypi
```
from semantic_kernel.connectors.memory.chroma import ChromaMemoryStore
from chromadb.config import Settings
mem = ChromaMemoryStore(client_settings=Settings(chroma_api_impl="rest", chroma_server_host="xxx.xxx.xxx.xxx", chroma_server_http_port="8000"))
```

See the following error

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'Settings' is not defined
>>> from chromadb.config import Settings
>>> mem = ChromaMemoryStore(client_settings=Settings(chroma_api_impl="rest", chroma_server_host="xxx.xxx.xxx.xxx", chroma_server_http_port="8000"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/snelson3/opt/anaconda3/lib/python3.9/site-packages/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py", line 73, in __init__
    self._client = chromadb.Client(self._client_settings)
  File "/Users/snelson3/opt/anaconda3/lib/python3.9/site-packages/chromadb/__init__.py", line 143, in Client
    api = system.instance(API)
  File "/Users/snelson3/opt/anaconda3/lib/python3.9/site-packages/chromadb/config.py", line 259, in instance
    fqn = self.settings.require(key)
  File "/Users/snelson3/opt/anaconda3/lib/python3.9/site-packages/chromadb/config.py", line 171, in require
    val = self[key]
  File "/Users/snelson3/opt/anaconda3/lib/python3.9/site-packages/chromadb/config.py", line 180, in __getitem__
    raise ValueError(LEGACY_ERROR)
ValueError: You are using a deprecated configuration of Chroma.

If you do not have data you wish to migrate, you only need to change how you construct
your Chroma client. Please see the "New Clients" section of https://docs.trychroma.com/migration.
________________________________________________________________________________________________

If you do have data you wish to migrate, we have a migration tool you can use in order to
migrate your data to the new Chroma architecture.
Please `pip install chroma-migrate` and run `chroma-migrate` to migrate your data and then
[...]

Issue #2050, opened 2023-07-18
Python: Update Chroma Connector for Python SK from 0.3.29 -> 0.4.0


awharrison-28:
** This should land Monday the 17th **

Chroma is upgrading from 0.3.29 to 0.4.0. 0.4.0 is easier to build, more durable, faster, smaller, and more extensible. This comes with a few changes:

1. A simplified and improved client setup. Instead of having to remember weird settings, users can just do EphemeralClient, PersistentClient or HttpClient (the underlying direct Client implementation is also still accessible)

1. We migrated data stores away from duckdb and clickhouse. This changes the api for the PersistentClient that used to reference chroma_db_impl="duckdb+parquet". Now we simply set is_persistent=true. is_persistent is set for you to true if you use PersistentClient.

1. Because we migrated away from duckdb and clickhouse - this also means that users need to migrate their data into the new layout and schema. Chroma is committed to providing extension notification and tooling around any schema and data migrations (for example - this PR!).

After upgrading to 0.4.0 - if users try to access their data that was stored in the previous regime, the system will throw an Exception and instruct them how to use the migration assistant to migrate their data. The migration assitant is a pip installable CLI: pip install chroma_migrate. And is runnable by calling chroma_migrate

Please reference the readme at [chroma-core/chroma-migrate](https://github.com/chroma-core/chroma-migrate) to see a full write-up of our philosophy on migrations as well as more details about this particular migration.

Please direct any users facing issues upgrading to our Discord channel called [#get-help](https://discord.com/channels/1073293645303795742/1129200523111841883). We have also created a [email listserv](https://airtable.com/shrHaErIs1j9F97BE) to notify developers directly in the future about breaking changes.

awharrison-28:
closed by https://github.com/microsoft/semantic-kernel/pull/2010

jeffchuber:
Commenting so I can be assigned 👋

jeffchuber:
thanks for shepherding this @awharrison-28 !

PR #3218, opened 2023-10-17
Python: Fix chromadb connector

GaiusRed:
### Motivation and Context

The ChromaDB connector works out of the box only for local data storage. This change is required to inform the user of the correct Setting to use as well as fix the `does_collection_exist_async` function that breaks only when its used on a remote data store.

This solves #3046, and also completes work that was started in #2050.

### Description

The change at line 49 updates the Settings for using a remote instance of ChromaDB to work with the changes for v0.4.0 (also mentioned in #2050).

The change starting at line 103 fixes an issue that only occurs when using a remote instance of ChromaDB, where the `does_collection_exist_async` function will always return True, because the `get_collection_async` function expects a ValueError if the collection does not exist. When using a remote instance, it will return an HTTPError instead of a ValueError.

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
+++ b/python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
@@ -46,7 +46,7 @@ class ChromaMemoryStore(MemoryStoreBase):
             # Create a ChromaMemoryStore with a custom Settings instance
             chroma_remote_data_store = ChromaMemoryStore(
                 client_settings=Settings(
-                    chroma_api_impl="chromadb.api.fastapi.FastAPI"",
+                    chroma_api_impl="chromadb.api.fastapi.FastAPI",
                     chroma_server_host="xxx.xxx.xxx.xxx",
                     chroma_server_http_port="8000"
                 )
--- a/python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
+++ b/python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
@@ -46,7 +46,7 @@ class ChromaMemoryStore(MemoryStoreBase):
             # Create a ChromaMemoryStore with a custom Settings instance
             chroma_remote_data_store = ChromaMemoryStore(
                 client_settings=Settings(
-                    chroma_api_impl="rest",
+                    chroma_api_impl="chromadb.api.fastapi.FastAPI"",
                     chroma_server_host="xxx.xxx.xxx.xxx",
                     chroma_server_http_port="8000"
                 )
--- a/python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
+++ b/python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
@@ -100,11 +100,14 @@ class ChromaMemoryStore(MemoryStoreBase):
         self, collection_name: str
     ) -> Optional["Collection"]:
         try:
-            # Current version of ChromeDB rejects camel case collection names.
-            return self._client.get_collection(
-                name=camel_to_snake(collection_name),
-                embedding_function=self._default_embedding_function,
-            )
+            if camel_to_snake(collection_name) in await self.get_collections_async():
+                # Current version of ChromeDB rejects camel case collection names.
+                return self._client.get_collection(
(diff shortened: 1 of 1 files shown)


python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
```diff
@@ -46,7 +46,7 @@ def __init__(
             # Create a ChromaMemoryStore with a custom Settings instance
             chroma_remote_data_store = ChromaMemoryStore(
                 client_settings=Settings(
-                    chroma_api_impl="rest",
+                    chroma_api_impl="chromadb.api.fastapi.FastAPI"",
```
moonbox3, line 49: Hi @GaiusRed, you mentioned in https://github.com/microsoft/semantic-kernel/issues/3046 that there is a "workaround" by configuring `chroma_api_impl` to be `chromadb.api.fastapi.FastAPI` instead of `rest`. Is that the correct way to configure it? I am seeing [ChromaDB's docs](https://docs.trychroma.com/migration ) specify that `chroma_api_impl` should be `rest`.

moonbox3, line 49: Let's remove the extra `"`, please.

python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
```diff
@@ -100,11 +100,14 @@ async def get_collection_async(
         self, collection_name: str
     ) -> Optional["Collection"]:
         try:
-            # Current version of ChromeDB rejects camel case collection names.
-            return self._client.get_collection(
-                name=camel_to_snake(collection_name),
-                embedding_function=self._default_embedding_function,
-            )
+            if camel_to_snake(collection_name) in await self.get_collections_async():
```
moonbox3, line 103: I see we're adding the preemptive check to see if the collection exists by calling `await self.get_collections_async()` to handle the issue where the remote instance returns an `HTTPError` instead of a `ValueError`. Now that we're making that call every time we want to get a single collection, is this introducing performance overhead? Thinking about the case where the collection is long or if this is called frequently, it could become a bottleneck, right?

moonbox3, line 103: Two further thoughts:
- it may be helpful to add a comment explaining why the new check is necessary, especially since it addresses a very specific issue related to using a remote instance of ChromaDB.
- is there any test coverage that needs to be updated?

python/semantic_kernel/connectors/memory/chroma/chroma_memory_store.py
GaiusRed: Hello @moonbox3! From the docs you linked:
```
### http client (to talk to server backend)

# before
import chromadb
from chromadb.config import Settings
client = chromadb.Client(Settings(chroma_api_impl="rest",
                                        chroma_server_host="localhost",
                                        chroma_server_http_port="8000"
                                    ))

# after
import chromadb
client = chromadb.HttpClient(host="localhost", port="8000")
```
If you are already using "rest" implementation, the correct way to migrate is to stop using the Client base object and use the HttpClient object instead.

If you check how the HttpClient object is implemented, it just configures the Client's Settings to use the `chromadb.api.fastapi.FastAPI` instead of `rest`:
```
def HttpClient(
    host: str = "localhost",
    port: str = "8000",
    ssl: bool = False,
    headers: Dict[str, str] = {},
    settings: Settings = Settings(),
) -> API:
    settings.chroma_api_impl = "chromadb.api.fastapi.FastAPI"
    settings.chroma_server_host = host
    settings.chroma_server_http_port = port
    settings.chroma_server_ssl_enabled = ssl
    settings.chroma_server_headers = headers
    return Client(settings)
```
Note that I just followed how #2050 solved the persistent client issue (use the correct Settings on the base Client object instead of using the PersistentClient as per the migration document).

GaiusRed: Fixed on ca58f2a.

moonbox3:
Closing this PR as there has been no action since Oct 2023. If this work is still necessary, please re-create the PR against the latest in main. Thank you.

Issue #3157, opened 2023-10-11
.Net: When running a plan with Kernel.RunAsync, the KernelResult.Metadata only contains information about the plan function.


alliscode:
**Describe the bug**
After running a plan that invokes plugins, the resulting KernelResult should contain metadata about the functions (plugins) called in the plan. Instead, it only contains metadata about the plan plugin.

**To Reproduce**
1. Register at least one plugin in the kernel.
2. Create a plan with a goal that will invoke the plugin.
3. Run the plan with `kernel.RunAsync(plan)`
4. Examine the KernelResult object returned from running the plan and notice that metadata is only present for 1 plugin which is the plan plugin.

**Expected behavior**
Metadata should exist for all of the plugins invoked in the plan execution.

**Platform**
 - OS: [e.g. Windows, Mac]
 - IDE: [e.g. Visual Studio, VS Code]
 - Language: [e.g. C#, Python]
 - Source: [e.g. NuGet package version 0.1.0, pip package version 0.1.0, main branch of repository]

**Additional context**

matthewbolanos:
Closing this since KernelResult no longer exists.

PR #3220, opened 2023-10-17
.Net: Added possibility to get Plan results 

dmytrostruk:
### Motivation and Context

Resolves: https://github.com/microsoft/semantic-kernel/issues/3157

After plan execution `var result = await kernel.RunAsync(plan)`, `result` is `KernelResult`, which contains `FunctionResults` collection. At the moment, in this collection there is only one `FunctionResult` which is original `Plan`, but there is no possibility to get function result from each plan step. 

This PR contains changes to add possibility to get Plan results after plan execution. Taking into account `Plan` tree-like structure (`Plan.Steps` are also of type `Plan`), and the fact that `Plan` implements `ISKFunction`, new structure was added - `PlanResult` which inherits `FunctionResult` and contains list of steps, which are also of type `PlanResult`. 

In order to get Plan results from `KernelResult`, additional extension method `GetPlanResults` was added for `KernelResult` which takes first `FunctionResult` in collection (which is original parent `Plan`) and returns step results.

Usage:
```csharp
 var result = await kernel.RunAsync(plan);
 var planResults = result.GetPlanResults().ToList();
```

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Core/Planning/KernelResultPlanExtensions.cs
+++ b/dotnet/src/SemanticKernel.Core/Planning/KernelResultPlanExtensions.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.SemanticKernel.Orchestration;
+
+namespace Microsoft.SemanticKernel.Planning;
+
+/// <summary>
+/// Extension methods for <see cref="KernelResult"/> with plan execution results.
+/// </summary>
+public static class KernelResultPlanExtensions
+{
+    /// <summary>
+    /// Returns plan execution results from <see cref="KernelResult"/>.
+    /// </summary>
+    /// <param name="kernelResult">Instance of <see cref="KernelResult"/>.</param>
+    public static IReadOnlyCollection<FunctionResult> GetPlanResults(this KernelResult kernelResult)
+    {
+        var functionResult = kernelResult.FunctionResults.FirstOrDefault();
+
+        if (functionResult is not null && functionResult is PlanResult planResult)
+        {
+            return planResult.StepResults;
+        }
+
+        return Array.Empty<FunctionResult>();
+    }
+}
--- a/dotnet/src/SemanticKernel.Core/Planning/Plan.cs
+++ b/dotnet/src/SemanticKernel.Core/Planning/Plan.cs
@@ -240,7 +240,7 @@ public sealed class Plan : IPlan
     {
         if (this.HasNextStep)
         {
-            var step = this.Steps[this.NextStepIndex];
+            var step = this.GetNextStep();
 
             // Merge the state with the current context variables for step execution
             var functionVariables = this.GetNextStepVariables(context.Variables, step);
@@ -325,6 +325,7 @@ public sealed class Plan : IPlan
         CancellationToken cancellationToken = default)
     {
         var result = new FunctionResult(this.Name, this.PluginName, context);
+        var stepResults = new List<PlanResult>();
 
         if (this.Function is not null)
         {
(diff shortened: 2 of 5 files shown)


dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
```diff
@@ -825,6 +828,41 @@ public async Task CanExecutePlanWithExpandedAsync()
...
+
+        [SKName("Function2")]
+        static string Function2() => "Result2";
+        functions.Add(SKFunction.FromNativeMethod(Method(Function2), pluginName: PluginName));
+
+        [SKName("Function3")]
+        static string Function3() => "Result3";
+        functions.Add(SKFunction.FromNativeMethod(Method(Function3), pluginName: PluginName));
+
+        var plan = new Plan("my goal", functions[0], functions[1], functions[2]);
```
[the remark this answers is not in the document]
dmytrostruk: Great test case, thank you! It appeared that previous implementation didn't support results for all height of plan tree, so I updated `Plan` to support that.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Core/Planning/KernelResultPlanExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+    public static IReadOnlyCollection<PlanResult> GetPlanResults(this KernelResult kernelResult)
+    {
+        var functionResult = kernelResult.FunctionResults.FirstOrDefault();
+
+        if (functionResult is not null && functionResult is PlanResult planResult)
+        {
+            return planResult.StepResults;
+        }
+
+        return Array.Empty<PlanResult>();
```
markwallace-microsoft, line 28: Do we end up with two cases where an empty collection is returned i.e.

1. First function result is a PlanResult and has zero StepResults
2. First function is not a PlanResult

Should the caller be able to differentiate between these two cases? E.g. should we throw an exception if the KernelResult is no associated with a plan?

dotnet/src/SemanticKernel.Core/Planning/Plan.cs
```diff
@@ -248,6 +254,11 @@ public async Task<Plan> InvokeNextStepAsync(SKContext context, CancellationToken
             // Execute the step
             var result = await context.Runner.RunAsync(step, functionVariables, cancellationToken).ConfigureAwait(false);
 
+            if (result is PlanResult planResult)
+            {
+                this.StepResults.Add(planResult);
+            }
```
markwallace-microsoft, line 260: What is the case where this is not a PlanResult and the result can be ignored?

markwallace-microsoft (approved):
LGTM but a few questions

dotnet/src/SemanticKernel.Core/Planning/KernelResultPlanExtensions.cs
dmytrostruk: Good catch! Although I think that instead of throwing the exception, we can simply return `null` if `KernelResult` is not associated with a plan. I pushed that change and updated XML documentation to describe this case. Thank you!

dotnet/src/SemanticKernel.Core/Planning/Plan.cs
dmytrostruk: This is the case when method `InvokeNextStepAsync` is invoked directly or through `kernel.StepAsync` extension method. It's the ability to control plan execution manually. In this case, `context.Runner` may run a function, which will return not `PlanResult`, but `FunctionResult`, which means that this Plan step doesn't have any child steps. In this case we should not add it to `StepResults` property and this collection will be empty.

lemillermicrosoft (approved):

dotnet/src/SemanticKernel.Core/Planning/Plan.cs
```diff
@@ -50,6 +50,12 @@ public sealed class Plan : IPlan
     [JsonPropertyName("outputs")]
     public IList<string> Outputs { get; set; } = new List<string>();
 
+    /// <summary>
+    /// Collection of step results
+    /// </summary>
+    [JsonIgnore]
+    public IList<FunctionResult> StepResults { get; set; } = new List<FunctionResult>();
```
RogerBarreto, line 57: Not fully sure about exposing this new property.

Plan has the "Object Model", "Execution", "State" and now the "Result" responsibity. :)

dmytrostruk:
For now let's keep this PR as draft.

PR #3226, opened 2023-10-18
Python: added parameter to control which context is used for the function call

eavanvalkenburg:
### Motivation and Context

Small improvement in the function calling setup, this allows you to control the context used during the function call, if you have functions that depend on static context (like text_memory_skill with the collection name) then you need the existing context with the additional parameters, this is now possible. The default is that a new context is created and used, same as before.

### Contribution Checklist


python/samples/kernel-syntax-examples/openai_function_calling.py
```diff
@@ -34,13 +32,6 @@
     ),
 )
 
-skills_directory = os.path.join(__file__, "../../../../samples/skills")
```
moonbox3, line 37: This is an intentional removal of this skill configuration?

python/semantic_kernel/connectors/ai/open_ai/utils.py
```diff
@@ -167,7 +174,17 @@ async def chat_completion_with_function_call(
     # if there is no function_call or if the content is not a FunctionCall object, return the context
     if function_call is None or not isinstance(function_call, FunctionCall):
         return context
-    result = await execute_function_call(kernel, function_call, log=log)
+    if function_call_with_new_context:
```
moonbox3, line 177: Are there any test cases that need to be added or updated related to this change?

python/samples/kernel-syntax-examples/openai_function_calling.py
eavanvalkenburg: yes, more of an oversight that it was in there in the first place, it is not usefull for the sample, so should be deleted to make sure the sample is as easy as possible. Could be a seperate PR...

python/semantic_kernel/connectors/ai/open_ai/utils.py
eavanvalkenburg: This utility func is currently not tested, I will have a look.

moonbox3: We can always create a separate item to track adding tests.

moonbox3 (approved):
The rest looks good to me.

python/semantic_kernel/connectors/ai/open_ai/utils.py
eavanvalkenburg: sounds good, will start that asap

alexchaomander (approved):
LGTM!

PR #3227, opened 2023-10-18
.Net: Introduce IAIServiceSelector interface to allow AI service and request settings to be selected during semantic function execution

markwallace-microsoft:
### Motivation and Context

Fix following issues for Semantic functions:

1. Allow semantic functions to dynamically retrieve the AI service and request settings to be used when executing the function

Previously semantic functions execution has the following limitations:

1. The AI service was set on the semantic function when it was created. This meant that an instance of a semantic function could not be used with multiple different Kernel instances.
1. The AI request settings were set on the semantic function when it was created. This meant that different request settings could not be selected when the function was executed.

### Description

This PR add's a new abstraction called `IAIServiceSelector`. An instance of this is added to a semantic function when it is created. The default implementation works as follows:

1. If the semantic function only has a single associated model request setting instance with no service id then the default AI service and the model request settings are used. This is consistent with the previous behaviour.
1. If the semantic function only has a single associated model request setting instance with a service id then the named AI service and the model request settings are used. If the named AI service is not available an exception will be thrown.  This is consistent with the previous behaviour.
1. If the semantic function only has multiple associated model request setting instances then:
    1. The model request setting instances are considered in order
    1. The first model request setting instance with no service id is considered the default
    1. For each model request setting instance that has a service id, if the service exists then it will be used with the associated model request settings
    1. If no matching service can be found and default request settings are provided the default service will be used
    1. If no matching service can be found and no default request settings are provided an exception will be thrown

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Functions/IAIServiceSelector.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    /// Return the AI service and requesting settings from the specified provider based on the model settings.
+    /// The returned value is a tuple containing instances of <see cref="IAIService"/> and <see cref="AIRequestSettings"/>
+    /// </summary>
+    /// <typeparam name="T">Type of AI service to return</typeparam>
+    /// <param name="serviceProvider">AI service provider</param>
+    /// <param name="modelSettings">Collection of model settings</param>
+    /// <returns></returns>
+    (T?, AIRequestSettings?) SelectAIService<T>(IAIServiceProvider serviceProvider, List<AIRequestSettings>? modelSettings) where T : IAIService;
```
SergeyMenshykh, line 24: nit: IReadOnlyList?

dotnet/src/SemanticKernel.Core/Functions/OrderedIAIServiceSelector.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                return (service, modelSettings?.FirstOrDefault<AIRequestSettings>());
+            }
+        }
+        else
+        {
+            AIRequestSettings? defaultRequestSettings = null;
+            foreach (var model in modelSettings)
+            {
+                if (model.ServiceId is not null)
```
SergeyMenshykh, line 34: Consider checking for not empty string as well.

dotnet/src/SemanticKernel.Core/Functions/OrderedIAIServiceSelector.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            if (defaultRequestSettings is not null)
+            {
+                return (serviceProvider.GetService<T>(null), defaultRequestSettings);
+            }
+        }
+
+        var names = string.Join("|", modelSettings.Select(model => model.ServiceId).ToArray());
+        throw new SKException($"Service of type {typeof(ITextCompletion)} and name {names ?? "<NONE>"} not registered.");
```
SergeyMenshykh, line 56: I believe you wanted to include type of the T into the exception message and not of ITextCompletion?

dotnet/src/SemanticKernel.Abstractions/Functions/ISKFunction.cs
```diff
@@ -34,9 +35,9 @@ public interface ISKFunction
     string Description { get; }
 
     /// <summary>
-    /// AI service settings
+    /// Model request settings.
     /// </summary>
-    AIRequestSettings? RequestSettings { get; }
+    List<AIRequestSettings>? ModelSettings { get; }
```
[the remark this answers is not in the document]
markwallace-microsoft: Yep, I have obsoleted the `RequestSettings` property and am not replacing it.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Core/Functions/OrderedIAIServiceSelector.cs
```diff
@@ -0,0 +1,57 @@
...
+/// <summary>
+/// Implementation of <see cref="IAIServiceSelector"/> that selects the AI service based on the order of the model settings.
+/// Uses the service id to select the preferred service provider and then returns the service and associated model settings.
+/// </summary>
+internal class OrderedIAIServiceSelector : IAIServiceSelector
+{
+    /// <inheritdoc/>
+    public (T?, AIRequestSettings?) SelectAIService<T>(IAIServiceProvider serviceProvider, IReadOnlyList<AIRequestSettings>? modelSettings) where T : IAIService
+    {
+        if (modelSettings == null || modelSettings.Count == 0)
```
RogerBarreto, line 20: nit
```suggestion
        if (modelSettings is null || modelSettings.Count == 0)
```

dotnet/src/SemanticKernel.Core/Functions/OrderedIAIServiceSelector.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            var service = serviceProvider.GetService<T>(null);
+            if (service != null)
```
RogerBarreto, line 23: nit
```suggestion
            if (service is not null)
```

dotnet/src/SemanticKernel.Core/Functions/OrderedIAIServiceSelector.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+        else
+        {
+            AIRequestSettings? defaultRequestSettings = null;
+            foreach (var model in modelSettings)
+            {
+                if (!string.IsNullOrEmpty(model.ServiceId))
+                {
+                    var service = serviceProvider.GetService<T>(model.ServiceId);
+                    if (service != null)
```
RogerBarreto, line 36:
```suggestion
                    if (service is not null)
```

dotnet/src/SemanticKernel.Core/Functions/OrderedIAIServiceSelector.cs
```diff
@@ -0,0 +1,57 @@
...
+                else
+                {
+                    // First request settings with empty or null service id is the default
+                    defaultRequestSettings ??= model;
+                }
+            }
+
+            if (defaultRequestSettings is not null)
+            {
+                return (serviceProvider.GetService<T>(null), defaultRequestSettings);
```
RogerBarreto, line 50: Do we allow the `SelectAIService` to return null from this `GetService<T>` call?

The other scenarios we throw, thinking about consistency suggest throwing in here as well.

dotnet/src/SemanticKernel.Core/Functions/SemanticFunction.cs
```diff
@@ -72,9 +72,12 @@ internal sealed class SemanticFunction : ISKFunction, IDisposable
             description: promptTemplateConfig.Description,
             pluginName: pluginName,
             functionName: functionName,
+            serviceSelector: serviceSelector,
             loggerFactory: loggerFactory
-        );
-        func.SetAIConfiguration(promptTemplateConfig.GetDefaultRequestSettings());
+        )
+        {
+            _modelSettings = promptTemplateConfig.ModelSettings
```
RogerBarreto, line 79: Why does a PromptTemplateConfiguration has model settings ? This configuration should be passed directly to the `FromSemanticConfig()`.

PromptTemplateEngine, should not have any logic or knowledge of the `Model` settings.

RogerBarreto (approved):
Nit comments.
Suggestion on changing `SelectAIService` to aways throw if no service is found. + Removing any logic that handles ITextCompletion being null in the SemanticFunction InvokeAsync impl.

Overall LGTM

dotnet/src/SemanticKernel.Core/Functions/SemanticFunction.cs
markwallace-microsoft: `PromptTemplateConfig.cs` is the model object for a config.json. We allow the model settings to be set in the config.json.

RogerBarreto (approved):

PR #3228, opened 2023-10-18
.Net PromptTemplateEngine using FunctionRunner.

RogerBarreto:
### Motivation and Context

1. Enable triggering of Hooks when calling functions references in the template. 
#2324 
2. One more step towards Kernel.RunAsync first.

### Description

- Move template engine logic to use `context.FunctionRunner.RunAsync` instead of `function.InvokeAsync`
- Removing responsibility to find functions from the PromptTemplateEngine.

### Contribution Checklist


dotnet/src/Extensions/TemplateEngine.Basic/Blocks/CodeBlock.cs
```diff
@@ -101,53 +101,25 @@ public async Task<string> RenderCodeAsync(SKContext context, CancellationToken c
...
-
-        if (!this.GetFunction(context.Functions!, fBlock, out ISKFunction? function))
-        {
-            var errorMsg = $"Function `{fBlock.Content}` not found";
-            this.Logger.LogError(errorMsg);
-            throw new SKException(errorMsg);
-        }
-
-        SKContext contextClone = context.Clone();
+        ContextVariables inputVariables = context.Variables.Clone();
```
SergeyMenshykh, line 104: 1. Please add a comment explaining the reason for the cloning.
2. nit: Consider putting the line into the `else` block to avoid double coning of context variables.

SergeyMenshykh (approved):

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/PromptTemplateEngineTests.cs
```diff
@@ -149,16 +151,18 @@ string MyFunctionAsync(SKContext context)
...
         }
 
-        ISKFunction func = SKFunction.FromNativeMethod(Method(MyFunctionAsync), this);
-        Assert.NotNull(func);
+        List<ISKFunction> functions = new()
+        {
+            SKFunction.FromNativeMethod(Method(MyFunctionAsync), this),
+        };
+
+        Assert.NotNull(functions[0]);
```
dmytrostruk, line 159: Looks like this change is present in a couple of places in tests (replacing one function with list of functions), but we still work only with one function and access it by index `functions[0]`. Is this change required? I think if we revert it, it should still work, but maybe I'm missing something, please let me know.

dotnet/src/SemanticKernel.Core/Functions/FunctionCollection.cs
```diff
@@ -76,6 +76,11 @@ public IFunctionCollection AddFunction(ISKFunction functionInstance)
     /// <inheritdoc/>
     public ISKFunction GetFunction(string pluginName, string functionName)
     {
+        if (string.IsNullOrWhiteSpace(pluginName))
+        {
+            return this.GetFunction(GlobalFunctionsPluginName, functionName);
+        }
```
dmytrostruk, line 82: nit: Maybe we can simplify and re-assign `pluginName` instead of calling same function again. In this case we won't increase execution stack and avoid potential infinite recursion (I know it's rare case to happen here, but still :) ).
```suggestion
        pluginName = !string.IsNullOrWhiteSpace(pluginName) ?  pluginName : GlobalFunctionsPluginName;
```

dmytrostruk (approved):

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/PromptTemplateEngineTests.cs
RogerBarreto: The previous implementation breaks when using the Invoke native reflection. 
Says that the target is not the same.

Not sure if this is a .Net bug or something, but when I use within a `List` it works.

@stephentoub do you have any insights what might be the problem here?

RogerBarreto: Interestingly enough, I retried with the previous logic and it worked. Not sure what might have happened in my last experiment, seems to be related to local execution. Still not sure what was it but can't replicate it now.

dmytrostruk (approved):

PR #3229, opened 2023-10-18
Python: strings don't have contains method in python

kevin-m-kent:
### Motivation and Context

I changed the substring in string check to the proper method for python. 

### Contribution Checklist

- [X ] The code builds clean without any errors or warnings
- [ X] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [ X] All unit tests pass, and I have added new tests where possible
- [ X] I didn't break anyone :smile:

### DIFF
--- a/python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
+++ b/python/semantic_kernel/planning/stepwise_planner/stepwise_planner.py
@@ -229,7 +229,7 @@ class StepwisePlanner:
 
         if thought_match:
             result.thought = thought_match.group(0).strip()
-        elif not input.contains(ACTION):
+        elif ACTION not in input:
             result.thought = input
         else:
             raise ValueError("Unexpected input format")
--- a/python/tests/unit/planning/stepwise_planner/test_stepwise_planner_parse_result.py
+++ b/python/tests/unit/planning/stepwise_planner/test_stepwise_planner_parse_result.py
@@ -30,5 +30,21 @@ def test_when_input_is_final_answer_returns_final_answer(input: str, expected: s
     assert result.final_answer == expected
 
 
+@pytest.mark.parametrize(
+    "input, expected",
+    [
+        ("My thought", "My thought"),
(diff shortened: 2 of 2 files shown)


dmytrostruk (approved):
@kevin-m-kent Thank you for fixing this! 
I tried to reproduce the problem, but I'm not sure at the moment how to achieve that second condition and verify how it's working. Looks like the test you added works with first condition only. Would be nice to add test at some point to check second condition as well.

Thanks again!

kevin-m-kent:
> @kevin-m-kent Thank you for fixing this! I tried to reproduce the problem, but I'm not sure at the moment how to achieve that second condition and verify how it's working. Looks like the test you added works with first condition only. Would be nice to add test at some point to check second condition as well.
> 
> Thanks again!

no problem! 

Are you referring to these conditions?

``` 
        final_answer_match = re.search(S_FINAL_ANSWER_REGEX, input)

        if final_answer_match:
            result.final_answer = final_answer_match.group(1).strip()
            return result

        # Extract thought
        thought_match = re.search(S_THOUGHT_REGEX, input)

        if thought_match:
            result.thought = thought_match.group(0).strip()
        elif ACTION not in input:
            result.thought = input
        else:
            raise ValueError("Unexpected input format")
```

I thought the 'My thought' input would trigger the second condition there since it doesn't have [FINAL ANSWER] or [THOUGHT] or [ACTION]?

dmytrostruk:
> Are you referring to these conditions?

Yes, that's correct.

> I thought the 'My thought' input would trigger the second condition

I tried to debug your test and it appeared that the second condition is not triggered, instead it triggers the first one. But I'm not sure at the moment, what would be the right test for second condition, need to investigate it further.

markwallace-microsoft (approved):

PR #3230, opened 2023-10-18
.Net StreamingResult derivation fix for Text and Chat Results

### LINKED ISSUE #2730: .Net: IChatStreamingResult should not derive from IChatResult and ITextStreamingResult should not derive from ITextResult
In my opinion, it is too easy to accidentally get non-streaming completions. A call to IChatCompletion.GetStreamingChatCompletionsAsync() returns an IChatStreamingResult and if you are just using intellisense it is easy to write code that would not stream. For example:
```C#
        var sb = new StringBuilder();
        await foreach (IChatStreamingResult result in results)
        {
            ChatMessageBase message = await result.GetChatMessageAsync();
            sb.Append(message.Content);
        }
```
However, you almost always want the code
```C#
        var sb = new StringBuilder();
        await foreach (IChatStreamingResult result in results)
        {
            await foreach (ChatMessageBase message in result.GetStreamingChatMessageAsync())
            {
                sb.Append(message.Content);
            }
        }
```

It is easy to mistakenly call the non-streaming API of the streaming interface when you want streaming, and I am not sure there is a valid use case where you would want to call that API.

If IChatStreamingResult just didn't implement IChatResult then callers would be driven to the "pit of success" of calling the intended API.

I am not suggesting to change the underlying implementation classes (ChatStreamingResult & TextCompletionStreamingResult). The class can still implement both interfaces since the implementation is not exposed to callers directly only the interfaces.

### LINKED ISSUE #3198: .Net: GetFunctionResponse doesn't work with IChatStreamingResult (and can't)
`IChatResult` has a `GetFunctionResponse` extension method that extracts the function response information from the result. `IChatStreamingResult` inherits `IChatResult`, but `GetFunctionResponse` doesn't work for it, and can't: the function information sent down as part of a function response can (and almost always does) span multiple streaming results, e.g. the first result might only say it's a function call, the second might only contain the name, and the arguments might be split across the third, fourth, fifth, etc.

RogerBarreto:
### Motivation and Context

It is easy to mistakenly call the non-streaming API of the streaming interface when you want streaming.
This also resolves problem where non streaming behavior/implementation is enforced upon streaming objects. 

Callers should be driven to the "pit of success" of calling the intended API. 

Resolves #2730 
Resolves Partially #3198

### Description

Removes the inheritance from ITextResult and IChatResult from streaming counterparts.

This will avoid Non-streaming objects to be derived as `Streaming capable` objects.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
+++ b/dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
@@ -3,7 +3,6 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
-using System.Runtime.CompilerServices;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
@@ -64,15 +63,13 @@ public sealed class HuggingFaceTextCompletion : ITextCompletion
     }
 
     /// <inheritdoc/>
-    public async IAsyncEnumerable<ITextStreamingResult> GetStreamingCompletionsAsync(
+    [Obsolete("Streaming capability is not supported, use GetCompletionsAsync instead")]
+    public IAsyncEnumerable<ITextStreamingResult> GetStreamingCompletionsAsync(
         string text,
         AIRequestSettings? requestSettings = null,
-        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        CancellationToken cancellationToken = default)
     {
-        foreach (var completion in await this.ExecuteGetCompletionsAsync(text, cancellationToken).ConfigureAwait(false))
-        {
-            yield return (ITextStreamingResult)completion;
-        }
+        throw new NotSupportedException("Streaming capability is not supported");
     }
 
     /// <inheritdoc/>
@@ -115,7 +112,7 @@ public sealed class HuggingFaceTextCompletion : ITextCompletion
             };
         }
 
-        return completionResponse.ConvertAll(c => new TextCompletionStreamingResult(c));
+        return completionResponse.ConvertAll(c => new TextCompletionResult(c));
     }
 
     /// <summary>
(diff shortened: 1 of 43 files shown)


dotnet/src/SemanticKernel.Abstractions/AI/IModelResult.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.SemanticKernel.Orchestration;
+
+namespace Microsoft.SemanticKernel.AI;
+
+/// <summary>
+/// Interface for model results
+/// </summary>
+public interface IModelResult
```
dmytrostruk, line 10: nit: I think there may be some confusion between:
- `IModelResult` interface, which is base for `IChatResult`, `ITextResult` etc.
- `ModelResult` class, which is container for results from AI.
- `ModelResult` property, which is property of `IModelResult` interface.

From the first sight, it looks like `ModelResult` class should implement `IModelResult` interface, but it's not the case.

Can we change the name of this interface to `IAIResult` or `IBaseResult` to clearly define that `ModelResult` implementation is unrelated to this interface and it's just part of it (as property).

stephentoub:
> Resolves https://github.com/microsoft/semantic-kernel/issues/3198

Only partially. It makes it so that GetFunctionResponse isn't available on IChatStreamingResult, but then there's also no exposed mechanism for accessing functions information when streaming.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextCompletion/HuggingFaceTextCompletion.cs
```diff
@@ -71,7 +71,7 @@ public HuggingFaceTextCompletion(string model, string? apiKey = null, HttpClient
     {
         foreach (var completion in await this.ExecuteGetCompletionsAsync(text, cancellationToken).ConfigureAwait(false))
         {
-            yield return completion;
+            yield return (ITextStreamingResult)completion;
```
dmytrostruk, line 74: nit: Does it mean that streaming is not supported here, and we just return completed result? I'm wondering if it makes sense to throw the exception instead to show end users that streaming capability is not supported in this connector (at least for now).

dmytrostruk (approved):
LGTM, left a couple of small comments.

dotnet/src/SemanticKernel.Abstractions/AI/IModelResult.cs
SergeyMenshykh: @RogerBarreto Can we visualize the interfaces, the result types their members return, and the classes that implement them as a class diagram in Mermaid or any other tool? The number of interfaces is increasing, and it's becoming quite difficult to follow.

RogerBarreto:
@stephentoub I found out that we can support the functioncalling during streaming. Will create another PR to support it after merging this one and closing #3198

SergeyMenshykh (approved):

PR #3241, opened 2023-10-19
Java: allow customising the reading of memory records. Fixes #3060

### LINKED ISSUE #3060: Java: Extend CognitiveSearchMemoryStore with custom MemoryRecord Mapper
- MemoryRecord mapping logic in AzureCognitiveSearchMemoryStore.getNearestMatchesAsync method can't be extended. Cognitive Search field names are hardcoded in [AzureCognitiveSearchMemoryRecord](https://github.com/microsoft/semantic-kernel/blob/experimental-java/java/connectors/semantickernel-connectors-memory-azurecognitivesearch/src/main/java/com/microsoft/semantickernel/connectors/memory/azurecognitivesearch/AzureCognitiveSearchMemoryRecord.java).

- I would suggest to provide an extensibility mechanism for custom cognitive search index field mapping while still preserving the MemoryRecord contract. An idea could be to add an overloaded getNearestMatchesAsync method which accepts a mapper function parameter like this: Function<SearchDocument,MemoryRecord> memoryRecordMapper. An example how you could implement it's [here](https://github.com/Azure-Samples/azure-search-openai-demo-java/blob/9c78923ada996d6d97736561334f62dfa906238d/app/backend/src/main/java/com/microsoft/openai/samples/rag/ask/approaches/semantickernel/memory/CustomAzureCognitiveSearchMemoryStore.java#L72)... and you can use it (+/-) like [this](https://github.com/Azure-Samples/azure-search-openai-demo-java/blob/9c78923ada996d6d97736561334f62dfa906238d/app/backend/src/main/java/com/microsoft/openai/samples/rag/ask/approaches/semantickernel/JavaSemanticKernelWithMemoryApproach.java#L118)

- This can be useful in scenario where you don't have full control (different teams, different implementations) of both vector store indexing and vector similarity search processes (which is my use case ).

johnoliver:
Fixes #3060

### DIFF
--- a/java/connectors/semantickernel-connectors-memory-azurecognitivesearch/src/main/java/com/microsoft/semantickernel/connectors/memory/azurecognitivesearch/AzureCognitiveSearchMemoryStore.java
+++ b/java/connectors/semantickernel-connectors-memory-azurecognitivesearch/src/main/java/com/microsoft/semantickernel/connectors/memory/azurecognitivesearch/AzureCognitiveSearchMemoryStore.java
@@ -35,8 +35,10 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.util.function.Tuple2;
@@ -54,6 +56,7 @@ public class AzureCognitiveSearchMemoryStore implements MemoryStore {
     private final SearchIndexAsyncClient _adminClient;
 
     private final Map<String, SearchAsyncClient> _clientsByIndex = new ConcurrentHashMap<>();
+    private final Function<SearchDocument, MemoryRecord> memoryRecordMapper;
 
     /// <summary>
     /// Create a new instance of memory storage using Azure Cognitive Search.
@@ -74,7 +77,8 @@ public class AzureCognitiveSearchMemoryStore implements MemoryStore {
                         .endpoint(endpoint)
                         .credential(new AzureKeyCredential(apiKey))
                         .clientOptions(clientOptions())
-                        .buildAsyncClient());
+                        .buildAsyncClient(),
+                null);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


java/connectors/semantickernel-connectors-memory-azurecognitivesearch/src/main/java/com/microsoft/semantickernel/connectors/memory/azurecognitivesearch/AzureCognitiveSearchMemoryStore.java
```diff
@@ -54,6 +56,7 @@ public class AzureCognitiveSearchMemoryStore implements MemoryStore {
     private final SearchIndexAsyncClient _adminClient;
 
     private final Map<String, SearchAsyncClient> _clientsByIndex = new ConcurrentHashMap<>();
+    private final Function<SearchDocument, MemoryRecord> memoryRecordMapper;
```
dsgrieve, line 59: Just a thought... if this was a BiFunction<SearchDocument, MemoryRecord, Boolean>, then you could default the memoryRecordMapper with  AzureCognitiveSearchMemoryRecord::toMemoryRecord and do away with the if block in getNearestMatchesAsync. But, as a supplier of the Function<SearchDocument, MemoryRecord> I can include or exclude embeddings, so it would only help to make a default. So, nevermind... 

Would it be better to pass the mapper in the call to getNearestMatch(es)Async? Maybe how you map the SearchDocument depends on the collection or some other factor. If so, you'd have to create a new AzCSMS for each unique handling of a collection.

dsgrieve (approved):

java/connectors/semantickernel-connectors-memory-azurecognitivesearch/src/main/java/com/microsoft/semantickernel/connectors/memory/azurecognitivesearch/AzureCognitiveSearchMemoryStore.java
johnoliver: adding it on the getNearestMatchesAsync would mean adding it to the MemoryStore interface, that would then mean leaking some ACS specific classes onto the MemoryStore interface, I am sort of ok with doing specific stuff in a constructor, but not so much on the interface

dsgrieve (approved):

PR #3246, opened 2023-10-20
.Net Issue Fix - Add FunctionResponse to ChatStreaming interfaces

### LINKED ISSUE #3198: .Net: GetFunctionResponse doesn't work with IChatStreamingResult (and can't)

RogerBarreto:
### Motivation and Context

Resolves #3198 

Today we can't call functions using streaming, this change allows it to be used.

The current approach will require the stream to be buffered while listening to a potential function call.

The example shows how that will be achieved.
```
StringBuilder chatContent = new();

// Non function result streaming will happen here.
await foreach (var message in chatResult.GetStreamingChatMessageAsync())
{
    if (message.Content is not null)
    {
        Console.Write(message.Content);
        chatContent.Append(message.Content);
    }
}
chatHistory.AddAssistantMessage(chatContent.ToString());

// After all stream was loaded check if a `FunctionResponse` was present.
var functionResponse = await chatResult.GetStreamingFunctionResponseAsync();
```

Remark: Calling `GetStreamingFunctionResponseAsync()` will buffer the stream to capture the full function call definition, use it after all the streaming messages were flushed out.

### Description

Allows streaming interfaces to get the function result.

This is a simplified approach (Buffer the streaming of a function call request until it's complete). Using the FunctionCall

### Contribution Checklist


SergeyMenshykh (approved):

dmytrostruk (approved):

alexminza:
@RogerBarreto it seems that when the chat completion call returns just text and no function call to be invoked - the current code will just throw a `NotSupportedException` when checking `GetOpenAIStreamingFunctionResponseAsync` because of https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs#L40

```
public FunctionCall FunctionCall
        => this._message?.FunctionCall ?? throw new NotSupportedException("Function call is not supported");
```

PR #3265, opened 2023-10-23
.Net Removing IPlan interface

RogerBarreto:
### Motivation and Context

IPlan interface has not definition and all Plan implementations actually implement ISKFunction members. 

Currently there is no need to have this as a `Flag interface`.

### Description

Removing `IPlan` interface and pointing all its dependencies to `ISKFunction` directly.

### Contribution Checklist


dotnet/src/SemanticKernel.Core/Planning/InstrumentedPlan.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.SemanticKernel.Planning;
 /// <summary>
 /// Standard Semantic Kernel callable plan with instrumentation.
 /// </summary>
-internal sealed class InstrumentedPlan : IPlan
+internal sealed class InstrumentedPlan : ISKFunction
```
markwallace-microsoft, line 20: Mark `IPlan` as obsolete and add a comment that users should extend `ISKFunction` instead.

markwallace-microsoft (commented):
We are using obsolescence to make the path to upgrade to newer versions of the SK easier

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Core/Planning/InstrumentedPlan.cs
```diff
@@ -34,16 +34,29 @@ internal sealed class InstrumentedPlan : IPlan
     /// <summary>
     /// Initialize a new instance of the <see cref="InstrumentedPlan"/> class.
     /// </summary>
-    /// <param name="plan">Instance of <see cref="IPlan"/> to decorate.</param>
+    /// <param name="plan">Instance of <see cref="Plan"/> to decorate.</param>
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
     public InstrumentedPlan(
-        IPlan plan,
+        Plan plan,
```
dmytrostruk, line 40: Would it be better to pass `ISKFunction plan` instead? In this case it should be easier to test `InstrumentedPlan` on our side, since we can mock `ISKFunction` (we still need to add more tests to classes with instrumentation).

dotnet/src/SemanticKernel.Core/Planning/InstrumentedPlan.cs
```diff
@@ -34,16 +34,29 @@ internal sealed class InstrumentedPlan : IPlan
...
         this._plan = plan;
         this._logger = loggerFactory is not null ? loggerFactory.CreateLogger(typeof(InstrumentedPlan)) : NullLogger.Instance;
     }
 
+    /// <summary>
+    /// Initialize a new instance of the <see cref="InstrumentedPlan"/> class.
+    /// </summary>
+    /// <param name="plan">Instance of <see cref="IPlan"/> to decorate.</param>
+    /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
+    [Obsolete("Use the other concrete class Plan constructor instead")]
```
dmytrostruk, line 52: This class is already internal, it's possible to initialize it only on our side. In this case, `Obsolete` flag is not required on constructor, and we can remove this constructor completely.

dotnet/src/SemanticKernel.Core/Planning/InstrumentedPlan.cs
```diff
@@ -70,7 +83,7 @@ public FunctionView Describe()
 
     #region private ================================================================================
 
-    private readonly IPlan _plan;
+    private readonly Plan _plan;
```
dmytrostruk, line 86: Same comment as above, using `ISKFunction` instead of `Plan` should make this class more testable.

dotnet/src/SemanticKernel.Core/Planning/PlanExtensions.cs
```diff
@@ -68,13 +69,24 @@ public static string ToPlanString(this Plan plan, string indent = " ")
...
     /// <summary>
     /// Returns decorated instance of <see cref="IPlan"/> with enabled instrumentation.
     /// </summary>
     /// <param name="plan">Instance of <see cref="IPlan"/> to decorate.</param>
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
+    [Obsolete("Use concrete class Plan WithInstrumentation instead")]
     public static IPlan WithInstrumentation(this IPlan plan, ILoggerFactory? loggerFactory = null)
     {
-        return new InstrumentedPlan(plan, loggerFactory);
+        throw new NotSupportedException("This method is obsolete, use concrete class Plan WithInstrumentation instead");
```
dmytrostruk, line 90: If we will replace `Plan` type with `ISKFunction` interface in `InstrumentedPlan` class, it should be possible to keep `return new InstrumentedPlan(plan, loggerFactory);` line instead of throwing an exception. Otherwise, there are chances that user will receive runtime error, so in theory it will be breaking change.

RogerBarreto: Can´t apply to this method, as InstrumentedPlan is not an `IPlan` anymore.

image

dotnet/src/SemanticKernel.Core/Planning/InstrumentedPlan.cs
RogerBarreto: I need to obsolete this one as this method need to be removed in the future, important to keep track of it as well due to the Warning it gives when I remove it.

image

dotnet/src/SemanticKernel.Core/Planning/PlanExtensions.cs
dmytrostruk: Got it, makes sense!

dotnet/src/SemanticKernel.Core/Planning/PlanExtensions.cs
```diff
@@ -68,13 +69,24 @@ public static string ToPlanString(this Plan plan, string indent = " ")
         return planString;
     }
 
+    /// <summary>
+    /// Returns decorated instance of <see cref="ISKFunction"/> with plan enabled instrumentation.
+    /// </summary>
+    /// <param name="plan">Instance of <see cref="Plan"/> to decorate.</param>
+    /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
+    public static ISKFunction WithInstrumentation(this ISKFunction plan, ILoggerFactory? loggerFactory = null)
```
dmytrostruk, line 77: While we want to work with `ISKFunction` in `InstrumentedPlan`, this extension method should still extend `Plan`, because we need some differentiation between Function and Plan, and there is already an extension method for Function:
https://github.com/microsoft/semantic-kernel/blob/0c1b669d234767c9a3b4839c8859e2ec548780cb/dotnet/src/SemanticKernel.Core/Extensions/SKFunctionExtensions.cs#L84-L87
```suggestion
    public static ISKFunction WithInstrumentation(this Plan plan, ILoggerFactory? loggerFactory = null)
```

RogerBarreto: Reverted

dmytrostruk (approved):

PR #3302, opened 2023-10-26
.Net: Function calling: Add function author role and name property

### LINKED ISSUE #3017: .Net: Function calling: Update ChatMessage to support 'name'
> [!IMPORTANT]  
> Labeled **Urgent** because it may require a breaking change. Please reduce the priority if this is not the case.

To fully support the function calling RAG pattern, we need to have a "name" field in our chat message objects. However, there are concerns about putting OpenAI specific fields in abstraction classes, so this may require a refactor of the chat message abstraction.
Here's an example of adding the "name" field to enable the full function calling scenario: https://github.com/gitri-ms/semantic-kernel/pull/2

gitri-ms:
### Motivation and Context

Resolves #3017 and enables the full function calling scenario as described in the [OpenAI function calling documentation](https://platform.openai.com/docs/guides/gpt/function-calling).

### Description

Adds the "function" author role to allow adding function messages to chat history.  Function messages require a "name" parameter, so this change also adds that to chat message model.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessageBase.cs
```diff
@@ -17,6 +17,11 @@ public abstract class ChatMessageBase
     /// </summary>
     public string Content { get; set; }
 
+    /// <summary>
+    /// If role is function, the name of the function that produced the content
+    /// </summary>
+    public string? Name { get; set; } = null;
```
gitri-ms, line 23: Surfacing @stephentoub 's suggestion from the issue:

> If we're not comfortable adding a name property to a chat message in support of this, we could also just add a dictionary to the message so that additional context can be provided and interpreted by the chat completion service however it likes:
https://github.com/microsoft/semantic-kernel/compare/main...stephentoub:semantic-kernel:fixfunctioncalling

leungkimming:
I found that your version has an issue. I have a CheckStock function and my prompt is "How many tables and chairs are there in stock?". Note that LLM should issue 2 function calling messages (I have added a do-while loop in Example59). However, with your version, LLM only called CheckStock once. In my original working version using native OpenAI, the Messages in the ChatCompletionsOptions keeps both the function calling assistance message and the function return user message. In your solution, only function return user
messages are kept in the abstracted ChatMessageBase. Hence, in ClientBase.cs, when CreateChatCompletionsOptions populate ChatCompletionsOptions using the abstracted ChatMessageBase, there are no function calling assistance message histories passing back to the LLM. Hence, LLM failed to relate the function result message with the original function calling message.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessageBase.cs
RogerBarreto: @gitri-ms I would not change Abstractions regarding FunctionCalling support, instead I would think of having a Specialized ChatHistory class for OpenAI where you can add that function. 
On the examples where we know you are dealing with a OpenAIModel, you can check if the ChatHistory is an OpenAIChatHistory and play with the detailed function data.

`SKChatMessage` of a `OpenAIChatHistory` will have the name.

RogerBarreto (changes requested):

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -93,26 +93,35 @@ private static async Task CompleteChatWithFunctionsAsync(string ask, ChatHistory
             // you can invoke it using the following code.
             if (kernel.Functions.TryGetFunctionAndContext(functionResponse, out ISKFunction? func, out ContextVariables? context))
             {
-                var kernelResult = await kernel.RunAsync(func, context);
+                var result = (await kernel.RunAsync(func, context)).GetValue<object>();
```
SergeyMenshykh, line 96: Do we intentionally skip adding an LLM function call request to the chat history before adding the function execution result?

image
Link - https://platform.openai.com/docs/guides/gpt/function-calling

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessageBase.cs
gitri-ms: I understand the concern here, but if we don't have a way to support additional fields like this in the base object, then devs will have to downcast the ChatHistory whenever they want to interact with OpenAI features, which makes for a really clunky experience. Talking with @alliscode, we thought that @stephentoub's approach would be most flexible and future-proof, so I updated this PR to add an optional dictionary to `ChatMessageBase`.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
gitri-ms: Good catch, I think i just overlooked that. However, now that you've pointed it out, it's not as simple as it seems to add the OpenAI message back into chat history. We can access the message via `IChatResult.ModelResult.GetResult<ChatModelResult>().Choice.Message`, which gives us back an `Azure.AI.OpenAI.ChatMessage` object. However, I dont believe we have the capability to add that back into chat history without losing some information in the translation (would convert to our `ChatMessage` to add to `ChatHistory`, and then back to `Azure.AI.OpenAI.ChatMessage` in `ClientBase.cs`)

gitri-ms: Actually.... I think we are able to handle this with the dictionary approach. When adding this message back into chat history, would need to add key/value for "function_call" into the dictionary and make sure that gets copied back into the OAI message. I will update the PR with this.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -403,7 +409,8 @@ private static ChatRole GetValidChatRole(AuthorRole role)
 
         if (validRole != ChatRole.User &&
             validRole != ChatRole.System &&
-            validRole != ChatRole.Assistant)
+            validRole != ChatRole.Assistant &&
+            validRole != ChatRole.Function)
         {
             throw new ArgumentException($"Invalid chat message author role: {role}");
         }
```
stephentoub, line 416: I think we should consider removing this role validation altogether. If the role is invalid according to OpenAI, it'll eventually tell us that. The problem with this validation is if OpenAI adds more valid roles, this validation will proactively block their use and will require changes to SK and picking up a new package in order to be able to use them, even if they'd otherwise work.

It's similar to how .NET used to proactively try to validate paths being passed to File/Path APIs, but now that validation has been removed and we just rely on the OS to say whether there was something wrong with the path.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIFunctionResponse.cs
```diff
@@ -27,6 +28,14 @@ public class OpenAIFunctionResponse
     /// </summary>
     public Dictionary<string, object> Parameters { get; set; } = new();
 
+    /// <summary>
+    /// Fully qualified name of the function. This is the concatenation of the plugin name and the function name,
+    /// separated by the value of <see cref="OpenAIFunction.NameSeparator"/>.
+    /// If there is no plugin name, this is the same as the function name.
+    /// </summary>
+    public string FullyQualifiedName =>
+        this.PluginName.IsNullOrEmpty() ? this.FunctionName : string.Join(OpenAIFunction.NameSeparator, this.PluginName, this.FunctionName);
```
stephentoub, line 37: `string.Join(..., params string[])` is unnecessary expense here.  This can instead be:
```suggestion
        this.PluginName.IsNullOrEmpty() ? this.FunctionName : $"{this.PluginName}{OpenAIFunction.NameSeparator}{this.FunctionName}";
```
or
```suggestion
        this.PluginName.IsNullOrEmpty() ? this.FunctionName : this.PluginName + OpenAIFunction.NameSeparator + this.FunctionName;
```
or
```suggestion
        this.PluginName.IsNullOrEmpty() ? this.FunctionName : string.Concat(this.PluginName, OpenAIFunction.NameSeparator, this.FunctionName);
```
all of which will compile down identically.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/AuthorRole.cs
```diff
@@ -31,6 +31,11 @@ namespace Microsoft.SemanticKernel.AI.ChatCompletion;
     /// </summary>
     public static readonly AuthorRole Tool = new("tool");
 
+    /// <summary>
+    /// The role that provides information about a function call result.
+    /// </summary>
+    public static readonly AuthorRole Function = new("function");
```
stephentoub, line 37: All of this System/Assistant/User/Tool/Function fields should be changed to properties, e.g.
```C#
public static AuthorRole Function { get; } = new("function");
```

stephentoub (commented):
Thank you for fixing this.

leungkimming:
@gitri-ms As mentioned above, we need to add LLM function call request to the chat history before adding the function execution result. Please consider my suggestions below:

1. Connectors.AI.OpenAI\AzureSdk\ClientBase.cs
```
{
    var azureMessage = new ChatMessage(GetValidChatRole(message.Role), message.Content);
    string functionName = "";
    if (message.AdditionalContext?.TryGetValue("Name", out string? name) is true)
    {
        azureMessage.Name = name;
        functionName = name;
    }

    if (message.AdditionalContext?.TryGetValue("Arguments", out string? arguments) is true)
    {
        azureMessage.FunctionCall = new FunctionCall(functionName, arguments);
    }

}
```
2. Connectors\Connectors.AI.OpenAI\ChatCompletion\ChatHistoryExtensions.cs
```
public static void AddAssistantMessage(this ChatHistory chatHistory, IChatResult chatResult)
{
    FunctionCall fc = chatResult.ModelResult.GetOpenAIChatResult().Choice.Message.FunctionCall;
    chatHistory.AddMessage(AuthorRole.Assistant, "", new Dictionary<string, string> {
        { "Name", fc.Name },
        { "Arguments", fc.Arguments }
    });
}
```
3. Example59_OpenAIFunctionCalling.cs
```
// Check for message response
var chatMessage = await chatResult.GetChatMessageAsync();
if (!string.IsNullOrEmpty(chatMessage.Content))
{
    Console.WriteLine(chatMessage.Content);

    // Add the response to chat history
    chatHistory.AddAssistantMessage(chatMessage.Content);
}
// Check for function response
OpenAIFunctionResponse? functionResponse = chatResult.GetFunctionResponse();
if (functionResponse is not null)
{
    chatHistory.AddAssistantMessage(chatResult);

```

gitri-ms:
@leungkimming - Thanks for the suggestions, I am incorporating them and will have updated changes soon!

leungkimming:
@gitri-ms I am sorry that there is a minor bug in my suggested coding. In the original OpenAI function calling assistance message, the Name field should be null but in my coding, the Name field contains the function name. Only the function execution result message should have a value in the Name field. Below is the correction in the ClientBase.cs file.
```
{
    string? functionName = null;

    if (message.AdditionalProperties?.TryGetValue("Name", out string? name) is true)
    {
        functionName = name;
    }

    if (message.AdditionalProperties?.TryGetValue("Arguments", out string? arguments) is true)
    {
        azureMessage.FunctionCall = new FunctionCall(functionName, arguments);
    } else
    {
        azureMessage.Name = functionName;
    }

}
```
In addition, to demonstrate multiple function calling, I would like to suggest below changes to the Example59_OpenAIFunctionCalling.cs file.
1. Make the second prompt to call both the TimePlugin-Date and KlarnaShoppingPlugin
2. Add a do loop until CompletionsFinishReason.Stopped
3. When passing function execution result, change to FunctionCallAuto instead of FunctionCallNone.
```
// Set FunctionCall to the name of a specific function to force the model to use that function.
requestSettings.FunctionCall = "TimePlugin-Date";

// Set FunctionCall to auto to let the model choose the best function to use.
await CompleteChatWithFunctionsAsync("What day is today and What computer tablets are available for under $200?", chatHistory, chatCompletion, kernel, requestSettings);

...

private static async Task CompleteChatWithFunctionsAsync(string ask, ChatHistory chatHistory, IChatCompletion chatCompletion, IKernel kernel, OpenAIRequestSettings requestSettings)
{
    Console.WriteLine($"User message: {ask}");
    chatHistory.AddUserMessage(ask);
    IChatResult chatResult;

    // Send request and add response to chat history
    chatResult = (await chatCompletion.GetChatCompletionsAsync(chatHistory, requestSettings))[0];
    chatHistory.AddAssistantMessage(chatResult);

    await PrintChatResultAsync(chatResult);
    do
    {
        // Check for function response
        OpenAIFunctionResponse? functionResponse = chatResult.GetFunctionResponse();
        if (functionResponse is not null)
        {
            // If the function returned by OpenAI is an SKFunction registered with the kernel,
            // you can invoke it using the following code.
            if (kernel.Functions.TryGetFunctionAndContext(functionResponse, out ISKFunction? func, out ContextVariables? context))
            {
                var result = (await kernel.RunAsync(func, context)).GetValue<object>();

                string? resultContent = null;
                if (result is RestApiOperationResponse apiResponse)
                {
                    resultContent = apiResponse.Content?.ToString();
                }
                else if (result is string str)
                {
                    resultContent = str;
                }

                if (!string.IsNullOrEmpty(resultContent))
                {
                    // Add the function result to chat history
                    chatHistory.AddFunctionMessage(resultContent, functionResponse.FullyQualifiedName);

                    //Get another completion
                    chatResult = (await chatCompletion.GetChatCompletionsAsync(chatHistory, requestSettings))[0];
                    chatHistory.AddAssistantMessage(chatResult);

                    await PrintChatResultAsync(chatResult);
                }
            }
            else
            {
                Console.WriteLine($"Error: Function {functionResponse.PluginName}.{functionResponse.FunctionName} not found.");
            }
        }
    } while (chatResult.ModelResult.GetOpenAIChatResult().Choice.FinishReason != CompletionsFinishReason.Stopped);
}
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/ChatHistoryExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>
+    /// Add an assistant message to the chat history.
+    /// </summary>
+    /// <param name="chatHistory">Chat history</param>
+    /// <param name="chatResult">Chat result from the model</param>
+    public static void AddAssistantMessage(this ChatHistory chatHistory, IChatResult chatResult)
+    {
+        var chatMessage = chatResult.ModelResult.GetOpenAIChatResult().Choice.Message;
+
+        if (!chatMessage.Content.IsNullOrEmpty())
```
SergeyMenshykh, line 35: Can a chat message have both text and a function call? The code suggests that it's possible.

leungkimming: @SergeyMenshykh As you have quoted from OpenAI documentation,
```
messages.append(response_message)  # extend conversation with assistant's reply
```
I believe that we should simply extend the response_message from LLM, regardless of its content to the chat history. The response_message may contain message or function call or both because OpenAI's API reference does not explicitly mention that they cannot co-exist. You are correct that if they both exist, the code will generate two messages instead of one. Therefore, I suggest below changes:
```
public static void AddAssistantMessage(this ChatHistory chatHistory, IChatResult chatResult)
{
    var chatMessage = chatResult.ModelResult.GetOpenAIChatResult().Choice.Message;

    var functionCall = chatMessage.FunctionCall;
    if (functionCall is not null)
    {
        chatHistory.AddMessage(AuthorRole.Assistant, chatMessage.Content, new Dictionary<string, string>
        {
            { "Name", functionCall.Name },
            { "Arguments", functionCall.Arguments }
        });
    } else
    {
        chatHistory.AddAssistantMessage(chatMessage.Content);
    }
}
```

markwallace-microsoft:
Created a new PR here: https://github.com/microsoft/semantic-kernel/pull/3421

markwallace-microsoft:
@leungkimming I have created a new PR here: https://github.com/microsoft/semantic-kernel/pull/3421 to continue this work. Could you try out the latest changes (just minor updates) and let me know if it works for your scenarios. Thanks

leungkimming:
> @leungkimming I have created a new PR here: #3421 to continue this work. Could you try out the latest changes (just minor updates) and let me know if it works for your scenarios. Thanks

@ markwallace-microsoft Thankyou. It works for my scenario. However, I have a minor suggestion. Using chatResult.GetOpenAIFunctionResponse to check if not null to execute functions may have problem if OpenAI set the CompletionsFinishReason other than "stop" or "function_call". Shall we add a "GetOpenAIFinishReason" in  ChatResultExtensions.cs and let us properly check below values to determine the next action to take?
        private const string StoppedValue = "stop";
        private const string TokenLimitReachedValue = "length";
        private const string ContentFilteredValue = "content_filter";
        private const string FunctionCallValue = "function_call"; 

I am using chatResult.ModelResult.GetOpenAIChatResult().Choice.FinishReason but not quite elegant!

PR #3304, opened 2023-10-26
.Net: Enable use of authentication info from Open AI plugin manifest

dehoward:
### Motivation and Context

- today the authentication info defined in `ai-plugin.json` is not considered.
- this PR enables the use of that info both when fetching the plugin's spec and when subsequent calls are made to the plugin's endpoint.
- Note: an example utilizing this functionality will be added in a follow-up PR.

### Description
1. split `ImportAIPluginFunctionsAsync` into two functions: `ImportOpenApiPluginFunctionsAsync` and `ImportOpenAIPluginFunctionsAsync`
2. added `OpenAIFunctionExecutionParameters` that makes use of `OpenAIAuthenticateRequestAsyncCallback`, which takes `OpenAIAuthentication` as an argument.

### Contribution Checklist


dotnet/src/Functions/Functions.UnitTests/OpenAPI/Extensions/KernelAIPluginExtensionsTests.cs
```diff
@@ -47,6 +52,39 @@ public KernelAIPluginExtensionsTests()
...
+            ContractResolver = new DefaultContractResolver
+            {
+                NamingStrategy = new SnakeCaseNamingStrategy()
+            }
+        };
+        using var reader = new StreamReader(ResourcePluginsProvider.LoadFromResource("ai-plugin.json"), Encoding.UTF8);
+        var openAiDocumentContent = JsonConvert.DeserializeObject<JObject>(await reader.ReadToEndAsync(), serializerSettings)!;
+        var actualAuthConfig = openAiDocumentContent["auth"]!.ToObject<OpenAIManifestAuthenticationConfig>(JsonSerializer.Create(serializerSettings))!;
+
+        var openAiDocument = ResourcePluginsProvider.LoadFromResource("ai-plugin.json");
```
dehoward, line 70: this unit test is failing because the stream is closed for some reason when the `SendAsync` method is called on the `HttpClient`. still trying to understand why.. any ideas would be appreciated.

SergeyMenshykh: Look at the test in my alternative solution. I slightly changed it but it works - https://github.com/microsoft/semantic-kernel/blob/db249e5836bb037932cf095aecc42cd92f0469d6/dotnet/src/Functions/Functions.UnitTests/OpenAPI/Extensions/KernelAIPluginExtensionsTests.cs#L55

dehoward: thanks. the test is passing now

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -15,9 +15,13 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.SemanticKernel.Diagnostics;
+using Microsoft.SemanticKernel.Functions.OpenAPI.Authentication;
 using Microsoft.SemanticKernel.Functions.OpenAPI.Model;
 using Microsoft.SemanticKernel.Functions.OpenAPI.OpenApi;
 using Microsoft.SemanticKernel.Orchestration;
+using Newtonsoft.Json;
```
dmytrostruk, line 22: Would it be possible to use `System.Text.Json` for this functionality?

dehoward: yes, updated

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -180,6 +182,17 @@ public static class KernelAIPluginExtensions
...
+            if (executionParameters?.AuthCallback != null &&
+                TryParseAIPluginForAuth(pluginContents, out var openAIManifestAuth) &&
+                openAIManifestAuth!.Type != OpenAIAuthenticationType.None)
+            {
+                var callback = executionParameters.AuthCallback;
+                executionParameters.AuthCallback = async (request, _) =>
+                {
+                    await callback(request, openAIManifestAuth).ConfigureAwait(false);
+                };
+            }
```
markwallace-microsoft, line 194: Ideally this code shouldn't have any OpenAI specific references.

Can this be implemented something like this:

```
if (executionParameters?.AuthCallback is not null)
{
     var callback = executionParameters.AuthCallback;
     executionParameters.AuthCallback = async (request, _) =>
     {
         await callback(request, pluginContents).ConfigureAwait(false);
     };
}
```

The idea is to push the OpenAI specific code i.e. `TryParseAIPluginForAuth(pluginContents, out var openAIManifestAuth)` into the callback.

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
```diff
@@ -322,6 +340,31 @@ private static bool TryParseAIPluginForUrl(string gptPluginJson, out string? ope
...
+            var serializerSettings = new JsonSerializerSettings
+            {
+                ContractResolver = new DefaultContractResolver
+                {
+                    NamingStrategy = new SnakeCaseNamingStrategy()
+                }
+            };
+
+            JObject? gptPlugin = JsonConvert.DeserializeObject<JObject>(gptPluginJson, serializerSettings);
+            openApiManifestAuth = gptPlugin?["auth"]?.ToObject<OpenAIManifestAuthenticationConfig>(JsonSerializer.Create(serializerSettings));
```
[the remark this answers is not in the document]
markwallace-microsoft: @SergeyMenshykh Should `KernelAIPluginExtensions` be named `KernelOpenAPIPluginExtensions`?

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
dehoward: removing the OpenAI specific references wasn't possible since a single method (i.e. `ImportAIPluginFunctionsAsync`) was being used to load both the OpenAI manifest and the OpenAPI spec, as it required the use of `TryParseAIPluginForUrl`.

due to this, I've separated out the OpenAI specific logic into a `ImportOpenAIPluginFunctionsAsync` function which calls the `ImportOpenApiPluginFunctionsAsync`. @markwallace-microsoft can you take another look at this approach?

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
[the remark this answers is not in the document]
dehoward: @markwallace-microsoft I think so, I've updated the naming to reflect this and have added a `KernelOpenAIPluginExtensions` file as well.

dotnet/src/Functions/Functions.OpenAPI/Authentication/AuthenticateRequestAsyncCallback.cs
```diff
@@ -9,5 +9,6 @@ namespace Microsoft.SemanticKernel.Functions.OpenAPI.Authentication;
 /// Represents a delegate that defines the method signature for asynchronously authenticating an HTTP request.
 /// </summary>
 /// <param name="request">The <see cref="HttpRequestMessage"/> to authenticate.</param>
+/// <param name="authConfig">The <see cref="OpenAIManifestAuthenticationConfig"/> used to authenticate.</param>
 /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
-public delegate Task AuthenticateRequestAsyncCallback(HttpRequestMessage request);
+public delegate Task AuthenticateRequestAsyncCallback(HttpRequestMessage request, OpenAIManifestAuthenticationConfig? authConfig = null);
```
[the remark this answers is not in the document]
dehoward: @SergeyMenshykh I've updated the code to remove this coupling and separated out the OpenAI specific paths into their own `ImportOpenAIPluginFunctionsAsync` function which makes use of the newly defined `OpenAIAuthenticateRequestAsyncCallback`.

SergeyMenshykh (approved):

dotnet/src/Functions/Functions.OpenAPI/OpenAI/OpenAIAuthenticationManifest.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel.Functions.OpenAPI.OpenAI;
+
+/// <summary>
+/// Represents the authentication configuration for an OpenAI plugin.
+/// </summary>
+public record OpenAIManifestAuthentication
```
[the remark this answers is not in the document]
dehoward: I named it this way because it's the authentication from the manifest. but I think we can drop the "Manifest" from this object entirely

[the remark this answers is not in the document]
SergeyMenshykh: Actually, I think that the `OpenAIAuthenticationConfig` would be a proper name since it says that is configuration related to authentication.

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelAIPluginExtensions.cs
markwallace-microsoft: LGTM

markwallace-microsoft (approved):

alliscode (approved):

PR #3320, opened 2023-10-26
Java: Set Useragent for connectors. Extract settings classes to their own module

johnoliver:
- Set Useragent for connectors to Semantic-Kernel-Java. 
- Extract settings classes to their own optional module


java/extensions/semantickernel-settings-loader/src/main/java/com/microsoft/semantickernel/SemanticKernelHttpSettings.java
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft. All rights reserved.
+package com.microsoft.semantickernel;
+
+import com.azure.core.util.ClientOptions;
+import com.azure.core.util.Header;
+import java.util.Collections;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+public class SemanticKernelHttpSettings {
```
dsgrieve, line 10: I'm okay with this as a temporary solution, but there are a lot of options in ClientOptions and it seems that it would be better to inject ClientOptions into OpenAIClientProvider than to rely on configured settings. If we are going to rely on configured settings, then it seems like it would be better to have Map<Settings, String> where Settings is an enum (or enum-ish) of the allowed settings.

dsgrieve (approved):

java/extensions/semantickernel-settings-loader/src/main/java/com/microsoft/semantickernel/SemanticKernelHttpSettings.java
johnoliver: I think the issue is that this is for more than just OpenAI clients, its on AzureCognitiveSearchMemoryStore  and may be on any further clients that make http calls

milderhc (approved):

brunoborges (approved):
LGTM

PR #3332, opened 2023-10-27
Update ADR Kernel Hooks Phase 1

RogerBarreto:
### Motivation and Context

Updated decision documentation.


markwallace-microsoft (approved):

docs/decisions/0005-function-execution-handlers.md
```diff
@@ -3,58 +3,62 @@
 status: accepted
```
markwallace-microsoft, line 3: We use the term "hooks" most places but not in the ADR or Kernel Syntax Sample. Can we rename this to `005-function-hooks.md` and change the title to `# Kernel Hooks (aka Function Handlers) - Phase 1`

markwallace-microsoft (approved):

dmytrostruk (approved):

PR #3342, opened 2023-10-30
Java: add documentation for the various packages

johnoliver:
### Contribution Checklist

### DIFF
--- a/java/PACKAGES.md
+++ b/java/PACKAGES.md
@@ -62,7 +62,7 @@ Provides a memory connector that can be used to interact with a JDBC database.
 : Implementations of various planners that can be used in the execution of semantic functions.
 
 `semantickernel-plugin-core`
-: Several example plugins, many of which are used in the samples.
+: Several example plugins, many of which are used in the [Java Semantic Kernel samples](/java-development/java/samples).
 
 `semantickernel-gpt3-tokenizer`
 : A tokenizer that can be used to tokenize text for use with GPT-3. Can be used to estimate cost.
--- a/java/PACKAGES.md
+++ b/java/PACKAGES.md
@@ -70,7 +70,9 @@ Provides a memory connector that can be used to interact with a JDBC database.
 
 ## Example Configurations
 
-### A simple project that uses Open AI for text/chat completion with a SQLite database as memory
+### Example: OpenAI + SQLite
+
+POM XML for a simple project that uses OpenAI for text or chat completion with an SQLite database as the memory provider.
 
 ```xml
 <dependencyManagement>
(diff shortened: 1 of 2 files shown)


java/PACKAGES.md
```diff
@@ -0,0 +1,147 @@
...
+## Connectors
+
+`semantickernel-connectors-ai-openai`
+: Provides a connector that can be used to interact with an OpenAI API.
+
+### Memory Connectors
+
+#### JDBC Memory Connectors
+
+Provides a memory connector that can be used to interact with a JDBC database.
```
d3r3kk, line 51: Do we want to make a brief mention of the JDBC package, and how one might want to  use it to extend a JDBC-based connection that we don't provide out of the box?

java/PACKAGES.md
```diff
@@ -0,0 +1,147 @@
...
+- `semantickernel-connectors-memory-sqlite`
+- `semantickernel-connectors-memory-postgresql`
+- `semantickernel-connectors-memory-mysql`
+
+`semantickernel-connectors-memory-azurecognitivesearch`
+: Provides a memory connector that allows using Azure Cognitive Search as a memory.
```
d3r3kk, line 57:
```suggestion
: Provides a memory connector for using Azure Cognitive Search as the memory provider for an application.
```

java/PACKAGES.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Miscellaneous Packages
+
+`semantickernel-planners`
+: Implementations of various planners that can be used to plan the execution of semantic functions.
```
d3r3kk, line 62:
```suggestion
: Implementations of various planners that can be used in the execution of semantic functions.
```

java/PACKAGES.md
```diff
@@ -0,0 +1,147 @@
...
+
+`semantickernel-plugin-core`
+: Several example plugins, many of which are used in the samples.
```
d3r3kk, line 65:
```suggestion
: Several example plugins, many of which are used in the [Java Semantic Kernel samples](/java-development/java/samples).
```

d3r3kk (approved):
Wrote a bunch of stuff that may have value, but I'm not going to be put out if you don't agree with my flowery prose 😁.

java/PACKAGES.md
brunoborges: Yes, eventually we will need detailed documentation for each concept, similarly to how Spring Project is documented.

Example of Spring Data JDBC https://docs.spring.io/spring-data/jdbc/docs/current/reference/html/

java/PACKAGES.md
brunoborges: "Semantic Kernel for Java" is what we use at all times.

d3r3kk (approved):

milderhc (approved):

brunoborges (approved):
This is a good start!

PR #3348, opened 2023-10-31
Replace IPromptTemplateEngine with IPromptTemplateFactory

markwallace-microsoft:
### Motivation and Context

Making this change to enable support for multiple prompt template formats
See https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0014-custom-prompt-template-formats.md

### Contribution Checklist


dotnet/src/Extensions/TemplateEngine.Basic/BasicPromptTemplateFactory.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /// <inheritdoc/>
+    public IPromptTemplate CreatePromptTemplate(string templateString, PromptTemplateConfig promptTemplateConfig)
+    {
+        if (promptTemplateConfig.TemplateFormat.Equals(PromptTemplateConfig.SEMANTICKERNEL, System.StringComparison.Ordinal))
+        {
+            return new BasicPromptTemplate(templateString, promptTemplateConfig, this._loggerFactory);
+        }
+        else if (this._promptTemplateFactory != null)
```
RogerBarreto, line 50: nit. Change where it applies.
```suggestion
        else if (this._promptTemplateFactory is not null)
```

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
```diff
@@ -285,4 +313,89 @@ public static class KernelSemanticFunctionExtensions
...
+        if (kernel.PromptTemplateEngine is not null)
+        {
+            return new PromptTemplateFactory(kernel.PromptTemplateEngine);
+        }
+#pragma warning restore CS0618 // Type or member is obsolete
+
+        if (!s_promptTemplateFactoryInitialized)
+        {
+            s_promptTemplateFactoryType = GetPromptTemplateFactoryType();
+            s_promptTemplateFactoryInitialized = true;
```
RogerBarreto, line 343: You can use just check if the type is null, not sure you need also a boolean flag for it.

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
```diff
@@ -285,4 +313,89 @@ public static class KernelSemanticFunctionExtensions
             kernel.LoggerFactory
         );
     }
+
+    private const string BasedTemplateFactoryAssemblyName = "Microsoft.SemanticKernel.TemplateEngine.Basic";
+    private const string BasedTemplateFactoryTypeName = "BasicPromptTemplateFactory";
```
RogerBarreto, line 318:
```suggestion
    private const string BasicTemplateFactoryTypeName = "BasicPromptTemplateFactory";
```

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
RogerBarreto, line 1: Not supper fan of storing static state in an `static extensions class` potentially this also adds `Basic` package knowledge to the Kernel.Core.

Potentially we could have a class that stores the Default in core, and when the package is initialized it register the basic in this core class, so we would not necessarily need to reflect or look for an assembly with the basic.

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
```diff
@@ -285,4 +313,89 @@ public static class KernelSemanticFunctionExtensions
...
+        }
+
+        throw new SKException($"Unable to create default prompt template factory. Please provide an implementation of IPromptTemplateFactory or depend on {BasedTemplateFactoryAssemblyName}");
+    }
+
+    /// <summary>
+    /// Get the prompt template engine type if available
+    /// </summary>
+    /// <returns>The type for the prompt template engine if available</returns>
+    private static Type? GetPromptTemplateFactoryType()
```
RogerBarreto, line 368: I'm not fully sure we should auto-discover template engines from the assembly, theorically the package could register itself in a core class like mentioned above...

dotnet/samples/KernelSyntaxExamples/Example56_TemplateNativeFunctionsWithMultipleArguments.cs
RogerBarreto, line 1: Consider adding an example of multiple Template Engine Factories in composition.

new Factory1(new Factory2)

Getting the templates registered on both.

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/PromptTemplateEngineTests.cs
```diff
@@ -21,6 +21,8 @@
 
 namespace SemanticKernel.Extensions.UnitTests.TemplateEngine.Prompt;
 
+[Obsolete("Use BasicPromptTemplateFactory instead. This will be removed in a future release.")]
+[EditorBrowsable(EditorBrowsableState.Never)]
 public sealed class PromptTemplateEngineTests
```
RogerBarreto, line 26: Tests in the Basic template engine should still be applicable to the new code, can we alter them to be `Factory` friendly?

Make sure to test also the composition of different factories.

RogerBarreto (changes requested):

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
markwallace-microsoft: `GetPromptTemplateFactoryType` could return null which would mean the code would keep trying to get the type. The `s_promptTemplateFactoryInitialized` flag ensures the code only runs once.

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
markwallace-microsoft: I don't like this code either. I think we should move the basic prompt template support back into the core.

RogerBarreto: Added some comments in the issue.
I think we can keep in another package, but use another approach.

RogerBarreto (approved):
Thanks for the changes, 

nit. Could not find the Factory.Factory test.

Found! 😊

dotnet/src/SemanticKernel.Abstractions/TemplateEngine/PromptTemplateConfig.cs
```diff
@@ -14,6 +14,11 @@ namespace Microsoft.SemanticKernel.TemplateEngine;
 /// </summary>
 public class PromptTemplateConfig
 {
+    /// <summary>
+    /// Semantic Kernel template format.
+    /// </summary>
+    public static readonly string SEMANTICKERNEL = "semantic-kernel";
```
RogerBarreto, line 20: Is there a special reason for Upper casing this "constant" ?

Follow the other examples in PascalCasing.

```suggestion
    public const string SemanticKernelTemplateFormat = "semantic-kernel";
```

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
```diff
@@ -285,4 +313,89 @@ public static class KernelSemanticFunctionExtensions
...
+        {
+            return null;
+        }
+    }
+
+    #endregion
+}
+
+[Obsolete("IPromptTemplateEngine is being replaced with IPromptTemplateFactory. This will be removed in a future release.")]
+internal sealed class PromptTemplateFactory : IPromptTemplateFactory
```
RogerBarreto, line 388: Move to another file or add this as a private inner class of the extension class.

dotnet/src/SemanticKernel.Core/TemplateEngine/AggregatorPromptTemplateFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            try
+            {
+                var promptTemplate = promptTemplateFactory.CreatePromptTemplate(templateString, promptTemplateConfig);
+                if (promptTemplate != null)
+                {
+                    return promptTemplate;
+                }
+            }
+            catch (SKException)
```
RogerBarreto, line 41: Lets not hide exceptions.
Remove try catch block, or throw a SKException.

dotnet/src/SemanticKernel.UnitTests/TemplateEngine/AggregatorPromptTemplateFactoryTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        // Act
+        var result1 = target.CreatePromptTemplate(templateString, new PromptTemplateConfig() { TemplateFormat = "my-format-1" });
+        var result2 = target.CreatePromptTemplate(templateString, new PromptTemplateConfig() { TemplateFormat = "my-format-2" });
+
+        // Assert
+        Assert.Throws<SKException>(() => target.CreatePromptTemplate(templateString, new PromptTemplateConfig() { TemplateFormat = "unknown-format" }));
+    }
+}
+
+public class MyPromptTemplateFactory1 : IPromptTemplateFactory
```
RogerBarreto, line 50: If those classes are only used in this test, change them as inner classes, otherwise we should move those `Fakes` to a Fakes folder.

dotnet/src/SemanticKernel.Core/TemplateEngine/AggregatorPromptTemplateFactory.cs
RogerBarreto, line 1: As we changed the approach to combine multiple engines, I think is lacking an example in our Samples project to the community where and how this one can be used, (combining f-string with handlebars or basic).

RogerBarreto (changes requested):
Thanks for the update Mark.

nit comments and small changes suggested.

dotnet/src/SemanticKernel.Core/TemplateEngine/AggregatorPromptTemplateFactory.cs
markwallace-microsoft: I need to catch this one and ignore it so I can try the next factory

dotnet/src/SemanticKernel.Core/TemplateEngine/AggregatorPromptTemplateFactory.cs
markwallace-microsoft: I'll add this when I add the Handlebars support

dotnet/src/SemanticKernel.Core/TemplateEngine/AggregatorPromptTemplateFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+public class AggregatorPromptTemplateFactory : IPromptTemplateFactory
+{
+    private readonly IPromptTemplateFactory[] _promptTemplateFactories;
+
+    /// <summary>
+    /// Constructor for <see cref="AggregatorPromptTemplateFactory"/>.
+    /// </summary>
+    /// <param name="promptTemplateFactories">List of <see cref="IPromptTemplateFactory"/> instances</param>
+    public AggregatorPromptTemplateFactory(params IPromptTemplateFactory[] promptTemplateFactories)
+    {
```
stephentoub, line 19:
```suggestion
    {
        Verify.NotNull(promptTemplateFactories);
```

dotnet/src/Extensions/TemplateEngine.Basic/BasicPromptTemplate.cs
```diff
@@ -0,0 +1,167 @@
...
+/// </summary>
+public sealed class BasicPromptTemplate : IPromptTemplate
+{
+    /// <summary>
+    /// Constructor for PromptTemplate.
+    /// </summary>
+    /// <param name="templateString">Prompt template string.</param>
+    /// <param name="promptTemplateConfig">Prompt template configuration</param>
+    /// <param name="loggerFactory">Logger factory</param>
+    public BasicPromptTemplate(string templateString, PromptTemplateConfig promptTemplateConfig, ILoggerFactory? loggerFactory = null)
```
stephentoub, line 35: In what situation does someone create one of these directly? Could this whole type be internal instead?

dotnet/src/Extensions/TemplateEngine.Basic/BasicPromptTemplateFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.SemanticKernel.Diagnostics;
+
+namespace Microsoft.SemanticKernel.TemplateEngine.Basic;
+
+/// <summary>
+/// Default implementation of <see cref="IPromptTemplateFactory"/> for the semantic-kernel prompt template format.
+/// </summary>
+public sealed class BasicPromptTemplateFactory : IPromptTemplateFactory
```
stephentoub, line 12: Is there a name of the SK prompt template format? Seems like this is _the_ factory for that format, so it should be named after that format, rather than "Basic".

dotnet/src/Extensions/TemplateEngine.Basic/Extensions/BasicPromptTemplateFactoryExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+#pragma warning disable IDE0130
+// ReSharper disable once CheckNamespace - Using the namespace of IKernel
+namespace Microsoft.SemanticKernel.TemplateEngine.Basic;
+#pragma warning restore IDE0130
+
+/// <summary>
+/// Class for extensions methods associated with the the basic prompt template factory.
+/// </summary>
+public static class BasicPromptTemplateFactoryExtensions
```
stephentoub, line 15: BasicPromptTemplateFactory is public.  Why have this as an extension method on BasicPromptTemplateFactory rather than just making it an instance method on the type?

Or should this instead be an extension method for any IPromptTemplateFactory rather than being specific to BasicPromptTemplateFactory?

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
```diff
@@ -240,34 +266,36 @@ public static class KernelSemanticFunctionExtensions
                 if (!File.Exists(promptPath)) { continue; }
 
                 // Load prompt configuration. Note: the configuration is optional.
-                var config = new PromptTemplateConfig();
+                var promptTemplateConfig = new PromptTemplateConfig();
                 var configPath = Path.Combine(dir, ConfigFile);
                 if (File.Exists(configPath))
                 {
-                    config = PromptTemplateConfig.FromJson(File.ReadAllText(configPath));
+                    promptTemplateConfig = PromptTemplateConfig.FromJson(File.ReadAllText(configPath));
```
stephentoub, line 273:
```C#
var configPath = Path.Combine(dir, ConfigFile);
var promptTemplateConfig = File.Exists(configPath) ?
    PromptTemplateConfig.FromJson(File.ReadAllText(configPath)) :
    new PromptTemplateConfig();
```
to avoid the unnecessary allocation when the file doesn't exist.

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
```diff
@@ -285,4 +313,91 @@ public static class KernelSemanticFunctionExtensions
...
+#pragma warning disable CS0618 // Type or member is obsolete
+        if (kernel.PromptTemplateEngine is not null)
+        {
+            return new PromptTemplateFactory(kernel.PromptTemplateEngine);
+        }
+#pragma warning restore CS0618 // Type or member is obsolete
+
+        if (!s_promptTemplateFactoryInitialized)
+        {
+            s_promptTemplateFactoryType = GetPromptTemplateFactoryType();
```
stephentoub, line 342: All this reflection suggests a problem in layering.  If this core assembly needs access to the default template factory, shouldn't the default template factory be in this assembly?

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
```diff
@@ -110,24 +112,27 @@ public static class KernelSemanticFunctionExtensions
...
+        string? pluginName = null,
+        IPromptTemplateFactory? promptTemplateFactory = null)
     {
         functionName ??= RandomFunctionName();
         Verify.ValidFunctionName(functionName);
         if (!string.IsNullOrEmpty(pluginName)) { Verify.ValidPluginName(pluginName); }
 
-        var template = new PromptTemplate(promptTemplate, promptTemplateConfig, kernel.PromptTemplateEngine);
+        var factory = promptTemplateFactory ?? CreateDefaultPromptTemplateFactory(kernel);
+        IPromptTemplate promptTemplateInstance = factory.CreatePromptTemplate(promptTemplate, promptTemplateConfig);
```
SergeyMenshykh, line 130: Just as an observation, no action required: As far as I understand it, the reason why the variable `promptTemplateInstance` is named the way it is, instead of being named `promptTemplate`, is to avoid a name collision between the parameter and the variable names. I think the root cause of it lies in the name of the class we decided to use to encapsulate prompt template rendering functionality - 'PromptTemplate.' The class is not strictly a prompt template, even though it accepts a prompt template in its constructor. It functions more as a prompt template rendering engine that accepts a prompt and can render it. So, I think the name 'PromptTemplateEngine' would be more appropriate for the class, as it better communicates what this class does and is responsible for.

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
markwallace-microsoft: Done

dotnet/src/SemanticKernel.UnitTests/TemplateEngine/AggregatorPromptTemplateFactoryTests.cs
markwallace-microsoft: Done

dotnet/src/Extensions/TemplateEngine.Basic/BasicPromptTemplate.cs
markwallace-microsoft: If I'm using the agregator to combine support for the basic (default) format plus Handlebars, plus Fluid I would do something like this:

```
var target = new AggregatorPromptTemplateFactory(
    new BasicPromptTemplateFactory(), 
    new HandlebarsPromptTemplateFactory(), 
    new FluidPromptTemplateFactory());
```

dotnet/src/Extensions/TemplateEngine.Basic/BasicPromptTemplateFactory.cs
markwallace-microsoft: This factory will support the semantic-kernel and f-string formats. I can change the name to `DefaultPromptTemplateFactory` or something else later. We need to agree on what the naming strategy will be as it impacts the reliability package also.

Tracking issue: https://github.com/microsoft/semantic-kernel/issues/3386

dotnet/src/SemanticKernel.Core/Extensions/KernelSemanticFunctionExtensions.cs
markwallace-microsoft: I think you may be right. I have a follow up task to remove this code. Roger has suggested an approach but if that doesn't work I may move the code back to the core package.

Tracking task: https://github.com/microsoft/semantic-kernel/issues/3353

dotnet/src/SemanticKernel.Abstractions/TemplateEngine/IPromptTemplateFactory.cs
```diff
@@ -0,0 +1,21 @@
...
+public interface IPromptTemplateFactory
+{
+    /// <summary>
+    /// Create an instance of <see cref="IPromptTemplate"/> from a template string and a configuration. Return null if the specified template format is not supported.
+    /// </summary>
+    /// <param name="templateString">Template string using the format associated with this <see cref="IPromptTemplateFactory"/></param>
+    /// <param name="promptTemplateConfig">Prompt template configuration</param>
+    /// <returns>Instance of <see cref="IPromptTemplate"/></returns>
+    /// <throws><see cref="SKException"/> if template format is not supported</throws>
+    IPromptTemplate CreatePromptTemplate(string templateString, PromptTemplateConfig promptTemplateConfig);
```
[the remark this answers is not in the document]
markwallace-microsoft: Fixed

dotnet/src/Extensions/TemplateEngine.Basic/Extensions/BasicPromptTemplateFactoryExtensions.cs
markwallace-microsoft: Taking Sergey's feedback below and I am removing this

dotnet/src/Planners/Planners.Core/Stepwise/StepwisePlanner.cs
```diff
@@ -350,14 +350,14 @@ private async Task<string> GetUserManualAsync(string question, SKContext context
     {
         var descriptions = await this._kernel.Functions.GetFunctionsManualAsync(this.Config, question, this._logger, cancellationToken).ConfigureAwait(false);
         context.Variables.Set("functionDescriptions", descriptions);
-        return await this._promptRenderer.RenderAsync(this._manualTemplate, context, cancellationToken).ConfigureAwait(false);
+        return await this._promptTemplateFactory.RenderAsync(this._manualTemplate, context, cancellationToken).ConfigureAwait(false);
     }
 
     private Task<string> GetUserQuestionAsync(SKContext context, CancellationToken cancellationToken)
-        => this._promptRenderer.RenderAsync(this._questionTemplate, context, cancellationToken);
+        => this._promptTemplateFactory.RenderAsync(this._questionTemplate, context, cancellationToken);
```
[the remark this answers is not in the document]
markwallace-microsoft: Will remove this

dotnet/src/Extensions/TemplateEngine.Basic/BasicPromptTemplateFactory.cs
RogerBarreto: @markwallace-microsoft might work keeping the single responsibility principle, and keep one "factory" for each format, and having the DefaultPromptTemplateFactory as an specialization of the Aggregate.

RogerBarreto (approved):

dotnet/src/Extensions/TemplateEngine.Basic/BasicPromptTemplateFactory.cs
markwallace-microsoft: @RogerBarreto Will do, the next template factory will be Handlebars which will be a seperate package as it introduces some new dependencies. After that I'll work on f-string and will follow the pattern you've suggested.

SergeyMenshykh (approved):

PR #3354, opened 2023-10-31
.Net: Reduce visibility of two fields from public to private

markwallace-microsoft:
### Motivation and Context

Two properties were inadvertently made public and should be private

### Contribution Checklist


dmytrostruk (approved):

RogerBarreto (approved):
If those are not obsoleted, this PR should be flagged as a breaking change.

PR #3374, opened 2023-11-02
.Net: Reduce TextChunker allocation with default token counter

stephentoub:
### Motivation and Context

There's still a lot of opportunity for improvement, but this cleans up some low-hanging fruit, in particular when the default token counter is used.

### Description

```C#
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using Microsoft.SemanticKernel.Text;

BenchmarkSwitcher.FromAssembly(typeof(Tests).Assembly).Run(args);

[HideColumns("Error", "StdDev", "Median", "RatioSD")]
[MemoryDiagnoser(displayGenColumns: false)]
public unsafe partial class Tests
{
    private static readonly string s_text = new HttpClient().GetStringAsync("https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/").Result;

    [Benchmark]
    public List<string> Split() => TextChunker.SplitPlainTextParagraphs(TextChunker.SplitPlainTextLines(s_text, 100), 1000);
}
```

Default Counter:

| Method | Mean     | Allocated |
|------- |---------:|----------:|
| Split_Before  | 28.55 ms |  87.17 MB |
| Split_After  | 11.44 ms |  18.78 MB |

Custom Counter (that does the same calculation as default):

| Method | Mean     | Allocated |
|------- |---------:|----------:|
| Split_Before  | 28.10 ms |  87.17 MB |
| Split_After  | 23.92 ms |  58.44 MB |

### Contribution Checklist


markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Core/Text/TextChunker.cs
```diff
@@ -240,12 +237,17 @@ private static (List<string>, bool) Split(List<string> input, int maxTokens, Rea
         return (result, inputWasSplit);
     }
 
-    private static (List<string>, bool) Split(ReadOnlySpan<char> input, string? inputString, int maxTokens, ReadOnlySpan<char> separators, bool trim, TokenCounter tokenCounter)
+    private static (List<string>, bool) Split(ReadOnlySpan<char> input, string? inputString, int maxTokens, ReadOnlySpan<char> separators, bool trim, TokenCounter? tokenCounter)
```
WeihanLi, line 240: Is it possible to use a named tuple for better readability
```suggestion
    private static (List<string> splitResult, bool hasSplit) Split(ReadOnlySpan<char> input, string? inputString, int maxTokens, ReadOnlySpan<char> separators, bool trim, TokenCounter? tokenCounter)
```

stephentoub: It could use names.  Wasn't changing that here, though.

SergeyMenshykh (approved):

PR #3380, opened 2023-11-03
.Net: Implementation of FunctionResultExtended with kernel example

joslat:
### Motivation and Context
I wanted a strongly type as the kernel result that provides me an easy to access result class with validated output from the model (or plan) execution.

One that I can associate the time, the Tokens and if I have a complex type returned, like a JSON structure in a string, validate and parse it into a ready to use strongly typed structure, right at our KernelResultExtended custom class.

I implemented the KernelResultExtended.cs next to KernelResult, in Microsoft.SemanticKernel.Orchestration, and an example Example64_SemanticFunctionWithKernelResultExtended where all the KernelSyntaxExamples are located.

For convenience I created a simple semantic function summarization execution with a custom KernelResultExtended class, which is declared inline for ease of understanding the sample.
It is below the sample, named KernelResultExtSummarize.

It helps to access all the "inner" parts of the KernelResult and provides a way to transform our Result into a strongly type very easily, all in the same strongly typed object.

It is a convenient change, not really needed but I feel more comfortable laying the parsing and validation of the result at this level, where I can even trigger a retry.

Makes all the internals of KernelResult easily accessible in a strongly typed way. 
Avoids repeating code every time we get the result and want the tokens, or want to parse the result.

Better architecture, helps providing a more strongly typed and is extensible to our output type customizations, validations and parsings that we want to implement.

### Description

A simple lightweight record that extends and exposes the KernelResult, which accepts on its constructor.
It is meant to be used directly or to be inherited and augmented to provide Tokens, duration and custom output type parsing and validation, to provide a verified strongly typed result .

### Contribution Checklist

- [ X] The code builds clean without any errors or warnings
- [ X] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [ X] All unit tests pass, and I have added new tests where possible
- [X ] I didn't break anyone :smile:


joslat:

joslat:
Hi, unsure of what can I do to improve this PR... tried to build in local and executing the unit tests and that works fine. For the integration, I lack the configuration for this to work properly, as this is probably hard to mock so far.
I reviewed and saw a double line so cleaned that up.
More than that, happy to help if you tell me how can improve this PR...

dmytrostruk:
> Hi, unsure of what can I do to improve this PR... tried to build in local and executing the unit tests and that works fine. For the integration, I lack the configuration for this to work properly, as this is probably hard to mock so far. I reviewed and saw a double line so cleaned that up. (that is the recent change added => one line removal) More than that, happy to help if you tell me how can improve this PR...

Hi @joslat , thank you for your contribution! 
We will review this PR shortly and provide our feedback. Thanks again!

dotnet/src/SemanticKernel.Abstractions/Orchestration/KernelResultExtended.cs
```diff
@@ -0,0 +1,62 @@
...
+public record KernelResultExtended
+{
+    /// <summary>
+    /// Represents the result of the kernel execution.
+    /// </summary>
+    public string Result { get; init; }
+    /// <summary>
+    /// Represent the usage in Tokens.
+    /// </summary>
+    public TokenCounts? TokenCounts { get; set; }
```
dmytrostruk, line 29: The reason why we didn't add things like `TokenCounts` or `ModelResults` to `FunctionResult`/`KernelResult` under `Microsoft.SemanticKernel.Orchestration` namespace is because these properties are related to AI execution. But there may be situations, when Kernel works only with native functions, so tokenization logic or model result is not applicable.

To make it more abstract, we added `Dictionary<string, object> Metadata` to `FunctionResult`, which can hold any kind of data, related to semantic or native function execution. We also provided `AIFunctionResultExtensions.GetModelResults(this FunctionResult result)` extension method under `Microsoft.SemanticKernel.AI` namespace, so it should help to get `ModelResults` in easier way. 

But in general, our goal was to provide minimal API, that is required to get complex results from Kernel. There are a lot of ways how to simplify it, by implementing wrappers, static classes, extension methods etc. But this is something that can be done on user side and users can choose what syntax works better in their specific scenarios. 

While we can keep the file with example to show how the result can be handled, I would try to keep `SemanticKernel.Abstractions` as lightweight as possible. 

Please let me know if that makes sense. Thank you!

dmytrostruk (commented):
@joslat Thank you for your contribution! 
I provided my feedback, happy to discuss it further in case of any questions. Thanks again!

dotnet/src/SemanticKernel.Abstractions/Orchestration/KernelResultExtended.cs
joslat: It does, thanks for the detailed explanation. And yes, it makes sense. Should I simplify this and put it as just an example on how to parse the KernelResult?

dmytrostruk: @joslat Yes, you can put it as an example, I think it should be useful for developers. Thanks a lot!

markwallace-microsoft:
@joslat Thanks for your contribution. We are refactoring all of our kernel syntax examples so they can be run as integration tests. We would like to add this sample, could you take a look at this PR https://github.com/microsoft/semantic-kernel/pull/4526 and update to match that pattern and we'll get this sample merged. Thanks.

joslat:
Hi Mark! Thanks for this :)
I understand that I should simply "take a number"  - If the latest is "https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example75_AgentTools.cs" add an Example76_KernelResult.cs or some name to imply the strongly typed result and put all the code on this file, right?

joslat:
@markwallace-microsoft It's done - I put it as example 76 and added a proper usage to the wrapper class to FunctionResult (prompt returns a JSON with synthetic data and parses it into a strongly typed structure, ready to work with and verified, as well as the tokenresult - which now uses CompletionsUsage and not ModelResult, that has been deprecated it seems).

joslat:
Please let me know if I am missing anything - so I can do it :) 
I updated the branch with the latest code, tested the changes locally - everything works. Maybe the Prompt only works with GPT4 which I tested it with... or needs some hardening but so far the 10 or so times I tested it manually - it works flawlessly.

dotnet/samples/KernelSyntaxExamples/Example76_StronglyTypedKernelResult.cs
```diff
@@ -0,0 +1,126 @@
...
+
+public class TestCompany
+{
+    public string Id { get; set; }
+    public string Name { get; set; }
+    public string Description { get; set; }
+}
+
+// The FunctionResult custom wrapper to parse the result and the tokens
+public record FunctionResultTestDataGen : FunctionResultExtended
```
RogerBarreto, line 69: Suggest avoiding usage of `records` as they add unneeded complexity at the compiler time and there's no actual benefit of using it in the example.

joslat: changing it in 1, 2...

joslat: Done :)

joslat: Hi @RogerBarreto and @markwallace-microsoft I think it is clean and ready...

joslat:
And adapted to the "new style" Xunit test inheriting from BaseTest ;)

joslat:
And I just realized of the "unresolved" conversation... which was resolved...

dotnet/samples/KernelSyntaxExamples/Example76_StronglyTypedKernelResult.cs
```diff
@@ -0,0 +1,148 @@
...
+using System.Diagnostics;
+using System.Collections.Generic;
+using Xunit;
+using Xunit.Abstractions;
+using Examples;
+
+// The following example shows how to receive the results from the kernel in a strongly typed object
+// wich stores the usage in tokens and converts the JSON result to a strongly typed object, where a validation can also
+// be performed
+public class Example76_StronglyTypedKernelResult : BaseTest
```
dmytrostruk, line 17: Seems like this should be Example77 now :) Also, since we don't have `KernelResult` anymore, I believe we should rename this example to `Example77_StronglyTypedFunctionResult`.

dotnet/samples/KernelSyntaxExamples/Example76_StronglyTypedKernelResult.cs
```diff
@@ -0,0 +1,148 @@
...
+        this.WriteLine($"Total Tokens: {functionResultTestDataGen.TokenCounts!.TotalTokens} \n");
+    }
+
+    public Example76_StronglyTypedKernelResult(ITestOutputHelper output) : base(output)
+    {
+    }
+}
+
+// The structure to put the JSON result in a strongly typed object
+public class RootObject
```
dmytrostruk, line 66: Since all classes belong to this specific example, let's make them `private sealed`.

dmytrostruk (approved):
@joslat Sorry for long delay, let's merge this to main as soon as possible. I left small comments there. Also, build pipeline shows that there is a typo and formatting issues. if you would be able to fix these errors, that would be awesome. Thanks a lot!

dotnet/samples/KernelSyntaxExamples/Example76_StronglyTypedKernelResult.cs
joslat: Alrighty!

dotnet/samples/KernelSyntaxExamples/Example76_StronglyTypedKernelResult.cs
joslat: agree, except FunctionResultExtended ;)

joslat:
@dmytrostruk - all done! - thx for the feedback :)  💯 🥇

markwallace-microsoft:
> Hi Mark! Thanks for this :) I understand that I should simply "take a number" - If the latest is "https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example75_AgentTools.cs" add an Example76_KernelResult.cs or some name to imply the strongly typed result and put all the code on this file, right?

We are going to group all of the samples and get rid of the numbering soon to make everything more discoverable

markwallace-microsoft (approved):

PR #3384, opened 2023-11-05
Java: optionally allow/disallow missing functions in sequential planner

### LINKED ISSUE #2855: Java: Enable plan validation in the SequentialPlanner
We've been using the experimental java branch with the SequentialPlanner. Sometimes, the LLM will return a plan that uses some example skills we don't implement in our project. 

For example, the plan might look like

```
Steps:
  - CustomSkill.MyCustomSemanticFunction ...
  - Emailer.EmailTo ...
```

where we have implemented `CustomSkill.MyCustomSemanticFunction` but not `Emailer.EmailTo`

It happens infrequently (we have the temperature set to 0 and see it <5% of the time in our tests), so I tried to implement logic in our project to inspect the resulting `Plan` to check if any of the steps weren't available to the kernel, but the steps on the plan are not accessible. 

I assume this will be a common problem; I could see a few different ways to approach it:
* Make the steps accessible (even a read-only list or copied list) so that the consumer can do the validation
* Add some validation method on the `Plan` itself for a user to check if it's executable and retry creation if it is not 
* Perform some validate plan and/or retry creation as part of the plan creation to reduce the likelihood of consumers getting back a plan that can't be executed.

lnavarette:
### Motivation and Context

This is to resolve #2855 -- when using the java `SequentialPlanner`, the planner will return a plan with missing skills. This enables a configuration setting to allow/disallow the missing functions. 

### Description

This PR adds
* The configuration setting to allow/disallow missing functions. I set the default to false (throw on a missing function), technically a breaking change from what was previously implemented. Given this branch is still prerelease, I decided that's the better default configuration, and the breaking change should be okay. 
* The logic in the plan parser; to throw when `allowMissingFunctions=true` and the plan has a missing function. 
* Unit tests for the SequentialPlanner that bring its tests to parity with the other Planner unit tests and test the desired behavior for this setting when it is on or off. 

### Contribution Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -472,6 +472,9 @@ java/.mvn/wrapper/maven-wrapper.jar
 # Java settings
 conf.properties
 
+# Intellij configuration
+*.iml
+
 # Playwright
 playwright-report/
 
--- a/java/extensions/semantickernel-planners/src/main/java/com/microsoft/semantickernel/planner/sequentialplanner/SequentialPlanParser.java
+++ b/java/extensions/semantickernel-planners/src/main/java/com/microsoft/semantickernel/planner/sequentialplanner/SequentialPlanParser.java
@@ -48,6 +48,26 @@ public class SequentialPlanParser {
     // result for a plan.
     private static final String AppendToResultTag = "appendToResult".toLowerCase(Locale.ROOT);
 
+
+
+    /**
+     * Convert a plan xml string to a plan
+     *
+     * @param xmlString The plan xml string
+     * @param goal The goal for the plan
+     * @param skills The skills to use
+     * @return The plan
+     * @throws PlanningException If the plan xml is invalid
+     */
+    public static Plan toPlanFromXml(
+        String xmlString,
+        String goal,
+        ReadOnlySkillCollection skills
+    ) throws PlanningException {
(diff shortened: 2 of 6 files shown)


lnavarette:

johnoliver:
@lnavarette thank you very much for the pr.

We have moved development over to the `java-development` branch, you can change the target branch of this PR with the edit button at the top of this page.

Other than that, looks good, approved, thanks

johnoliver (approved):

lnavarette:
@johnoliver can you hit approve again? changing the base branch dismissed your last review.

johnoliver:
@lnavarette looks like you have a compile failure in the integration tests:

https://github.com/microsoft/semantic-kernel/blob/java-development/java/api-test/integration-tests/src/test/java/com/microsoft/semantickernel/planner/sequentialplanner/SequentialPlanTest.java#L182

Might be worth adding a override of 

```
public static Plan toPlanFromXml(String xmlString, String goal, ReadOnlySkillCollection skills) throws PlanningException {
  return toPlanFromXml(xmlString, goal, skills, false);
}
```

to maintain backwards compatibility (other than the exception)

johnoliver (approved):

johnoliver:
@lnavarette Thank you for the contribution

PR #3387, opened 2023-11-06
.Net: Chat Prompts - handling in Connector

dmytrostruk:
### Motivation and Context

This PR contains changes to support complex prompts that can contain not only user ask in text representation, but also instructions for AI models (e.g. `system` message), examples how AI model should respond (e.g. `assistant` message) and potentially other type of prompts in the future (e.g. images, video, audio etc).

ADR (option 1): https://github.com/microsoft/semantic-kernel/pull/3338
The logic in this approach performs prompt parsing on Connector level. 

Pros:
 - The `SemanticFunction` won't need to change to support the mapping of a new prompt syntax when new completion type connectors (audio, video, etc.) are added.

 - Prompts can be run by
    - Kernel.RunAsync 
    - Completion connectors

Cons:
 - Every new completion connector, whether of an existing type or a new type, will have to implement the mapping functionality

Demo:
```
Text Prompt:
What is Seattle?
Text Prompt Result:
Seattle is a city located in the state of Washington in the United States...

Chat Prompt:
<message role="user">What is Seattle?</message>
<message role="system">Respond with JSON.</message>

Chat Prompt Result:
{
  "Seattle": {
    "Description": "Seattle is a city located in the state of Washington, in the United States...",
    "Population": "Approximately 753,675 as of 2019",
    "Area": "142.5 square miles",
    ...
  }
}
```

### Description
- `XmlPromptParser` - public class, converts prompt string in XML format to collection of `PromptNode` instances.
- `PromptNode` - public class, intermediate prompt representation. It will allow Connector authors to work with the same prompt model in case of other prompt parsers as well (e.g. `JsonPromptParser` that will produce same `PromptNode` instances) and avoid coupling with XML classes.
- `ChatPromptParser` - internal class in OpenAI connector, that processes `PromptNode` in unique way (i.e. fetching `message` tags with `role` attributes and `content`.

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -292,6 +302,12 @@ await foreach (var chatCompletionStreamingResult in chatCompletionStreamingResul
     private static OpenAIChatHistory PrepareChatHistory(string text, AIRequestSettings? requestSettings, out OpenAIRequestSettings settings)
     {
         settings = OpenAIRequestSettings.FromRequestSettings(requestSettings);
+
+        if (ChatPromptParser.TryGetChatHistory(text, out var chatHistory) && chatHistory is not null)
```
SergeyMenshykh, line 306: I think the additional check of the `chatHsitory` for null is not necessary here. The `TryGetChatHistory` should always initialize chat history if it returns true.

SergeyMenshykh (approved):

RogerBarreto (approved):
LGTM

markwallace-microsoft (approved):

SergeyMenshykh (approved):

PR #3389, opened 2023-11-06
Adds Intellij IDEA files to .gitignore

agoncal:
### Motivation and Context

Intellij IDEA is heavy used in the Java ecosystem, so it would be good to have default files in the .gitignore

### Description

Add the default Intellij IDEA working files and directory to the .gitignore

### Contribution Checklist


johnoliver (dismissed):

johnoliver:
@agoncal I have had to move the pr to be against `java-development` unfortunately needs you to rebase against that branch

johnoliver (approved):

dsgrieve (approved):

.gitignore
```diff
@@ -409,14 +409,19 @@ FodyWeavers.xsd
 *.key
 *.pem
 
+# JetBrains IntelliJ
+.idea
```
WeihanLi, line 413: Should it be `.idea/`?

dsgrieve: Having no trailing separator char is fine. 

From https://git-scm.com/docs/gitignore

- If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories.

- For example, a pattern doc/frotz/ matches doc/frotz directory, but not a/doc/frotz directory; however frotz/ matches frotz and a/frotz that is a directory (all paths are relative from the .gitignore file).

markwallace-microsoft (approved):

PR #3390, opened 2023-11-06
Java - Changes constant settings visibility to public

agoncal:
### Motivation and Context

When writing external APIs that enhance the configuration, we need to have access to the settings. This PR turns all the constant settings to public.

AbstractOpenAIClientSettings.KEY_SUFFIX and OpenAISettings.OPEN_AI_ORGANIZATION_SUFFIX are already public, but the other settings are private, which makes it impossible to be used by external APIs.

### Description

External APIs need to have access to all the constant settings to enhance connection.

### Contribution Checklist

### DIFF
--- a/java/extensions/semantickernel-settings-loader/src/main/java/com/microsoft/semantickernel/connectors/ai/openai/util/AzureOpenAISettings.java
+++ b/java/extensions/semantickernel-settings-loader/src/main/java/com/microsoft/semantickernel/connectors/ai/openai/util/AzureOpenAISettings.java
@@ -7,9 +7,9 @@ import javax.annotation.Nullable;
 
 /** Settings for Azure OpenAI client */
 public class AzureOpenAISettings extends AbstractOpenAIClientSettings {
-    public static final String DEFAULT_SETTINGS_PREFIX = "client.azureopenai";
-    public static final String AZURE_OPEN_AI_ENDPOINT_SUFFIX = "endpoint";
-    public static final String AZURE_OPEN_AI_DEPLOYMENT_NAME_SUFFIX = "deploymentname";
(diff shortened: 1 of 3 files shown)


java/extensions/semantickernel-settings-loader/src/main/java/com/microsoft/semantickernel/connectors/ai/openai/util/AzureOpenAISettings.java
```diff
@@ -7,9 +7,9 @@
 
 /** Settings for Azure OpenAI client */
 public class AzureOpenAISettings extends AbstractOpenAIClientSettings {
-    private static final String DEFAULT_SETTINGS_PREFIX = "client.azureopenai";
-    private static final String AZURE_OPEN_AI_ENDPOINT_SUFFIX = "endpoint";
-    private static final String AZURE_OPEN_AI_DEPLOYMENT_NAME_SUFFIX = "deploymentname";
+    public static final String DEFAULT_SETTINGS_PREFIX = "client.azureopenai";
```
dsgrieve, line 10: My recommendation here is to have a static getter method for these. The issue is, if you reference DEFAULT_SETTINGS_PREFIX in your code, the compiler will do a straight-up string replacement. If the connectors library later changes the value, your code is not going to work with the newer library. If you use a getter, your code will always have the correct value from the connectors library.  

e.g., `public static final String getDefaultSettingsPrefix() { return DEFAULT_SETTINGS_PREFIX; }`

dsgrieve (changes requested):

brunoborges (changes requested):
As per David's comments, please add getters to these properties. The getters methods may be `public static`

johnoliver (approved):

dsgrieve (approved):

PR #3392, opened 2023-11-06
.Net: Handlebars prompt template support

markwallace-microsoft:
### Contribution Checklist


dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        var handlebars = HandlebarsDotNet.Handlebars.Create();
+
+        var functionViews = executionContext.Functions.GetFunctionViews();
+
+        foreach (FunctionView functionView in functionViews)
+        {
+            var skfunction = executionContext.Functions.GetFunction(functionView.PluginName, functionView.Name);
+#pragma warning disable VSTHRD101 // Avoid unsupported async delegates
+            handlebars.RegisterHelper($"{functionView.PluginName}_{functionView.Name}", async (writer, hcontext, parameters) =>
```
SergeyMenshykh, line 46: It seems that Handlebars.NET doesn't support [async helpers](https://github.com/Handlebars-Net/Handlebars.Net/issues/166). This means it doesn't wait for the result of async functions provided to the RegisterHelper method. It is not clear how Handlebars will handle the situation when it compiles subsequent template blocks before completing the compilation of previous ones. Another question is what will happen if InvokeAsync throws an exception. The expected behavior would be to have the exception propagated from the lambda to the Handlebars engine and subsequently up the call stack. The actual behavior is unclear and may lead to a corrupted compiled template.

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -0,0 +1,101 @@
...
+    private readonly ILoggerFactory _loggerFactory;
+    private readonly ILogger _logger;
+    private readonly string _templateString;
+    private readonly PromptTemplateConfig _promptTemplateConfig;
+    private readonly Lazy<IReadOnlyList<ParameterView>> _parameters;
+
+    private List<ParameterView> InitParameters()
+    {
+        // Parameters from prompt template configuration
+        Dictionary<string, ParameterView> result = new(this._promptTemplateConfig.Input.Parameters.Count, StringComparer.OrdinalIgnoreCase);
```
SergeyMenshykh, line 71: It seems we don't need dictionary because we use only values and don't use keys. Using List<ParameterView> should be enough.

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
markwallace-microsoft: Fixed using `var result = skfunction.InvokeAsync(executionContext).GetAwaiter().GetResult();`

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
markwallace-microsoft: Fixed

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Handlebars/HandlebarsPromptTemplateFactoryTests.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace SemanticKernel.Extensions.UnitTests.TemplateEngine.Handlebars;
+
+public sealed class HandlebarsPromptTemplateFactoryTests
+{
+    [Fact]
+    public void ItCreatesHandlebarsPromptTemplate()
+    {
+        // Arrange
+        var templateString = "{{$input}}";
```
RogerBarreto, line 16: nit. Handlerbars template isn't `{{input}}` ? There's an extra `$` there...

RogerBarreto (approved):

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Handlebars/HandlebarsPromptTemplateFactoryTests.cs
markwallace-microsoft: Fixed

SergeyMenshykh (approved):

PR #3394, opened 2023-11-06
ADR for JSON-serializable custom types

dehoward:
### Motivation and Context

this PR contains an ADR for utilizing the native serialization methods from `System.Text.Json` to serialize custom types. 
it will be one in a series of changes needed in supporting complex types within planners.

### Contribution Checklist


docs/decisions/0016-json-serializable-custom-types.md
```diff
@@ -0,0 +1,71 @@
...
+deciders: alliscode, markwallace-microsoft
+consulted:
+informed:
+---
+
+# JSON Serializable Custom Types
+
+## Context and Problem Statement
+
+Today, use of custom types within Semantic Kernel requires developers to implement a custom `TypeConverter` to convert to/from the string representation of the type. This is demonstrated in [Example60_AdvancedNativeFunctions](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example60_AdvancedNativeFunctions.cs#L202C44-L202C44) as seen below:
```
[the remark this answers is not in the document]
dehoward: I added a line for this after the code snippet. I was going to add code, but not sure it is necessary

docs/decisions/0021-json-serializable-custom-types.md
```diff
@@ -0,0 +1,72 @@
...
+        public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+        {
+            return JsonSerializer.Serialize(value);
+        }
+    }
+```
+
+## Considered Options
+
+**1. Use native serialization methods**
```
dmytrostruk, line 46: In scope of this work, it would be nice to describe in what part of SK the serialization/deserialization logic will be performed. There are a couple of options:
1. As current approach with TypeConverter, we can do serialization/deserialization in NativeFunction. This solution should temporarily resolve the problem, but in NativeFunction and other places in SK we will still be coupled to the string.
2. Do serialization only in place where it's actually needed - in SemanticFunction, to prepare a request and call AI. In this case, we will avoid multiple serialization/deserialization actions in data processing, and objects will be passed between native functions as-is. This will require more changes and design, because there are other open questions, how to represent object as string in meaningful way, how to convert it back to complex type (TypeChat?) etc.

So, it would be great to have more information in this ADR about high-level implementation. Please let me know if that makes sense. Thank you!

dmytrostruk: Tagging @SergeyMenshykh to verify if this part of work doesn't overlap with current refactoring process.

dehoward: @dmytrostruk the idea proposed was going to lean more toward number 1, by to updating `NativeFunction` like you mentioned and likely making updates to the other places as needed to allow the object to be passed as-is. 
number 2 is interesting tho. for your open questions, isn't this what we would use the JSON serialization for? I'm not sure I understand what you mean there.

dmytrostruk: > isn't this what we would use the JSON serialization for? I'm not sure I understand what you mean there.

Let's take the case, when we have following functions in pipeline:
1. `NativeFunction1` - it returns `ComplexType1`.
2. `SemanticFunction1` - it should accept `ComplexType1`, pass it to AI, and return `string` or `ComplexType2` (depends on how we will implement handling of objects instead of strings).
3. `NativeFunction2` - it accepts `ComplexType2` and returns some result.

In case of JSON serialization, in `SemanticFunction1` you will take `ComplexType1`, serialize it to JSON and pass it to AI. Then, you will get some response, and in order to pass that response to `NativeFunction2`, you will need to convert it to `ComplexType2`.

The open question is - how to do that properly? If we use JSON deserialization for that, the response from `SemanticFunction1` should be always in JSON format. We will have to force model to respond in JSON only to achieve that. Otherwise, if you will do `JsonSerializer.Deserialize<ComplexType2>("response_from_AI")`, where `response_from_AI` is plain string, you will receive `System.Text.Json.JsonException: ''response_from_AI' is an invalid JSON literal.`, and `JsonConverter` won't handle that. While with `TypeConverter` it's still possible to do it, because `converter.ConvertFromString("response_from_AI")` will pass this value to your converter, and you will be able to map it to your complex type correctly.

The other way would be to replace `ComplexType2` parameter in `NativeFunction2` with `string` and delegate data processing to the user, but in this case, we won't support complex types natively, it's the same `string` processing, so it would be nice to avoid that.

dehoward: after some talks following this comment, it was decided that we're ok with constraining the model to always return JSON responses given that we will know how to deserialize the response. I've added cases in the ADR where serialization/deserialization will be necessary, specifically when we're going from a native fn response to a semantic fn input vice versa.

the updated proposed approach that's stemmed from the discussions is: **Use TypeConverter for primitive types and native serialization methods for custom types**

docs/decisions/0021-json-serializable-custom-types.md
```diff
@@ -0,0 +1,163 @@
...
+
+```
+
+If a developer wants to customize the serialization and/or deserialization of a custom type, they can implement their own `JsonConverter` and apply the `JsonConverterAttribute` their custom type as seen below.
+
+**Example usage of the `JsonConverter`** class and attribute:
+
+```csharp
+    [JsonConverter(typeof(MyCustomTypeConverter))]
+    private sealed class MyCustomType
```
markwallace-microsoft, line 127: The example above uses a `TypeConverter` which is implemented using JSON so it does two/from strings but it is flexible enough to support any type. Are we changing that?

docs/decisions/0021-json-serializable-custom-types.md
```diff
@@ -0,0 +1,163 @@
...
+            JsonSerializer.Serialize(writer, value, options);
+        }
+    }
+```
+
+_When is serialization/deserialization required?_
+
+Required
+
+- **Native to Semantic:** Passing variables from Native to Semantic **will** require serialization of the output of the Native Function from complex type to string so that it can be passed to the LLM.
```
markwallace-microsoft, line 152: If we have an LLM which accepts something other than strings how will this work?

docs/decisions/0021-json-serializable-custom-types.md
```diff
@@ -0,0 +1,163 @@
...
+        public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+        {
+            return JsonSerializer.Serialize(value);
+        }
+    }
+```
+
+## Considered Options
+
+**1. Use TypeConverter for primitive types and native serialization methods for custom types**
```
markwallace-microsoft, line 47: Can the summary be improved, is the following accurate

1. Primitive types will be handled using JSON serialization
2. Complex types will be handled by the registered `TypeConverter` is available
3. If no `TypeConverter` is registered for a complex type, JSON serialization/deserialization is used
    - This option may fail in which case a clear error message is provided explaining a  `TypeConverter` is required

docs/decisions/0021-json-serializable-custom-types.md
```diff
@@ -0,0 +1,163 @@
...
+        if (targetType == typeof(decimal)) { return new JsonConverterWrapper(new DecimalConverter()); }
+        if (targetType == typeof(TimeSpan)) { return new JsonConverterWrapper(new TimeSpanConverter()); }
+        if (targetType == typeof(DateTime)) { return new JsonConverterWrapper(new DateTimeConverter()); }
+        if (targetType == typeof(DateTimeOffset)) { return new JsonConverterWrapper(new DateTimeOffsetConverter()); }
+        if (targetType == typeof(Uri)) { return new JsonConverterWrapper(new UriTypeConverter()); }
+        if (targetType == typeof(Guid)) { return new JsonConverterWrapper(new GuidConverter()); }
+
+        if (targetType.GetCustomAttribute<JsonConverterAttribute>() is JsonConverterAttribute tca &&
+            Type.GetType(tca.ConverterType.Name, throwOnError: false) is Type converterType &&
+            Activator.CreateInstance(converterType) is JsonConverter converter)
```
markwallace-microsoft, line 78: Can you explain why this must be a JsonConverter?

docs/decisions/0021-json-serializable-custom-types.md
dehoward: yes, this is accurate. I've updated the desription to better reflect this.

docs/decisions/0021-json-serializable-custom-types.md
dehoward: it felt like the more natural choice since we're now defaulting to serialization, but you're right it doesn't have to be. 
thinking more about it, it will be less disruptive if we leave this alone as `TypeConverter` and instead just default to a `SimpleTypeConverter` like the one shown in the above example.

docs/decisions/0021-json-serializable-custom-types.md
dehoward: updated to use a `TypeConverter` but then return a default, JSON-serializing one if no custom converter is found

docs/decisions/0021-json-serializable-custom-types.md
dehoward: can you give an example of this? the assumption here is that semantic functions will always be working with strings, is that not a safe assumption to make?

markwallace-microsoft: If we have a native function which loads an image and returns a stream and then we want to pass it to a an LLM that can do image to text how would this work?

markwallace-microsoft (approved):

docs/decisions/0021-json-serializable-custom-types.md
dehoward: ah ok yes, it would work in the way you've described. image representation came up in the discussion for this ADR, and we agreed that a base64 string representation of the image included in the prompt would be the best approach for this.

markwallace-microsoft (approved):

Issue #3330, opened 2023-10-27
Python: prepare for OpenAI v1.0 package


eavanvalkenburg:
OpenAI is working on a 1.0 version of their package that we rely on heavily to call openai models, the interfaces are changing (for the better), current work can be seen here: https://github.com/openai/openai-python/tree/v1

alexchaomander:
@eavanvalkenburg agree that this should be worked on! Other projects like AutoGen are already preparing their latest releases for this change

PR #3172, opened 2023-10-13
Python: improve ai service usability for text/chat completion

joowon-dm-snu:
### Motivation and Context

fix #3171

### Contribution Checklist

- [ ] The code builds clean without any errors or warnings
- [ ] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [ ] All unit tests pass, and I have added new tests where possible
- [ ] I didn't break anyone :smile:

Issue #3171, opened 2023-10-13
Python: improve ai service usability for chat/text completions


joowon-dm-snu:
1) As `AzureChatCompletion` inherits `OpenAIChatCompletion`, its error message below is confusing because it always saying OpenAI service not Azure Service

```python
    async def _send_completion_request(
        ...
        try:
            response: Any = await openai.Completion.acreate(
                ...

        except Exception as ex:
            raise AIException(
                AIException.ErrorCodes.ServiceError,
                "OpenAI service failed to complete the prompt",
                ex,
            )
```

2) endpoint auto rstrip would be good as Azure OpenAI service provide endpoint with "/" at last. 
This raises error while executing `_send_completion_request`
<img width="697" alt="image" src="https://github.com/microsoft/semantic-kernel/assets/69464062/d5ce1f8a-113c-47a8-b287-d6c03212f895">

PR #3417, opened 2023-11-08
Python: Update SK Python to be compatible with openai-python 1.0

alexchaomander:
### Motivation and Context

Need to update SK Python to integrate the breaking changes from openai-python 1.0

https://github.com/microsoft/semantic-kernel/issues/3330


python/semantic_kernel/connectors/ai/open_ai/services/open_ai_text_embedding.py
```diff
@@ -66,21 +66,22 @@ async def generate_embeddings_async(
...
                 )
                 # make numpy arrays from the response
-                raw_embeddings.extend([array(x["embedding"]) for x in response["data"]])
+                raw_embeddings.extend([array(x.embedding) for x in response.data])
             return array(raw_embeddings)
         except Exception as ex:
             raise AIException(
                 AIException.ErrorCodes.ServiceError,
-                f"{self.__class__.__name__} failed to generate embeddings",
+                "OpenAI service failed to generate embeddings",
```
dmytrostruk, line 85: I think `f"{self.__class__.__name__}` was added previously because when developers used `AzureTextEmbedding` class which inherits `OpenAITextEmbedding` class, in case of exception, the error message will say `OpenAI service failed to...`, but it was actually `Azure OpenAI` service, so it was confusing.

Issue: https://github.com/microsoft/semantic-kernel/issues/3171
PR: https://github.com/microsoft/semantic-kernel/pull/3172

eavanvalkenburg:
Could we merge this into a separate branch so that multiple people can contribute to these changes, lot's of larger and smaller issues we will need to iron out for this! Might be good to consider combining with this: #2843 because there I made some improvements in the classes which will reduce the amount of doubled code.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
```diff
@@ -267,7 +262,13 @@ async def _send_chat_request(
                 model_args["functions"] = functions
 
         try:
-            response: Any = await openai.ChatCompletion.acreate(**model_args)
+            client = AsyncOpenAI(
```
eavanvalkenburg, line 265: we should probably move this client into the class attributes (including the api details) so that we do not reinitialise the creation, would also make differentiating the init between the openai and azureopenai version easier, both inits end with a initialized asyncopenai client as self.client and then the completions can just use that.

alexchaomander:
> Could we merge this into a separate branch so that multiple people can contribute to these changes, lot's of larger and smaller issues we will need to iron out for this! Might be good to consider combining with this: #2843 because there I made some improvements in the classes which will reduce the amount of doubled code.

Just created https://github.com/microsoft/semantic-kernel/pull/3470 as its own separate branch! @eavanvalkenburg can you merge over the code that you have in #2843 and we can iterate fully on that branch? I have merge rights for it, so we can move more quickly

alexchaomander:
Closing this PR because we will be moving this work over to this branch instead https://github.com/microsoft/semantic-kernel/pull/3470

PR #3421, opened 2023-11-08
.Net: Function calling: Add function author role and name property

### LINKED ISSUE #3017: .Net: Function calling: Update ChatMessage to support 'name'

markwallace-microsoft:
### Motivation and Context

Replaces https://github.com/microsoft/semantic-kernel/pull/3302

Resolves #3017 and enables the full function calling scenario as described in the [OpenAI function calling documentation](https://platform.openai.com/docs/guides/gpt/function-calling).

### Description

Adds the "function" author role to allow adding function messages to chat history.  Function messages require a "name" parameter, so this change also adds that to chat message model.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -37,12 +37,14 @@ public static async Task RunAsync()
...
-        await StreamingCompleteChatWithFunctionsAsync("What day is today?", chatHistory, chatCompletion, kernel, requestSettings);
+
+        // Uncomment the samples and run them one at a time
+        //await StreamingCompleteChatWithFunctionsAsync("What day is today?", chatHistory, chatCompletion, kernel, requestSettings);
 
         // Set FunctionCall to auto to let the model choose the best function to use.
         requestSettings.FunctionCall = OpenAIRequestSettings.FunctionCallAuto;
-        await CompleteChatWithFunctionsAsync("What computer tablets are available for under $200?", chatHistory, chatCompletion, kernel, requestSettings);
-        await StreamingCompleteChatWithFunctionsAsync("What computer tablets are available for under $200?", chatHistory, chatCompletion, kernel, requestSettings);
+        //await CompleteChatWithFunctionsAsync("What computer tablets are available for under $200?", chatHistory, chatCompletion, kernel, requestSettings);
```
RogerBarreto, line 46: nit. IMHO I prefer to keep all code that works as non-commented and to improve clarity move it to a specific method.

See: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example57_KernelHooks.cs

RogerBarreto (approved):
CLG

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
markwallace-microsoft: I'll fix this in a follow up PR

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -390,25 +390,22 @@ private static ChatCompletionsOptions CreateChatCompletionsOptions(OpenAIRequest
...
-        return options;
-    }
+            if (message.AdditionalProperties?.TryGetValue("Name", out string? name) is true)
+            {
+                azureMessage.Name = name;
 
-    private static ChatRole GetValidChatRole(AuthorRole role)
-    {
-        var validRole = new ChatRole(role.Label);
+                if (message.AdditionalProperties?.TryGetValue("Arguments", out string? arguments) is true)
```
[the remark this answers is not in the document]
markwallace-microsoft: Done

dmytrostruk (approved):

leungkimming:
@markwallace-microsoft I found a bug in beta6 that will cause OpenAI to return 500 Internal Server Error. To reproduce, change and execute Example59 as below:
```
        IKernel kernel = await InitializeKernelAsync();
        var chatCompletion = kernel.GetService<IChatCompletion>();
        var chatHistory = chatCompletion.CreateNewChat();

        OpenAIRequestSettings requestSettings = new()
        {
            Functions = kernel.Functions.GetFunctionViews().Select(f => f.ToOpenAIFunction()).ToList(),
        };
        await CompleteChatWithFunctionsAsync("What day is one month later?", chatHistory, chatCompletion, kernel, requestSettings);
``` 
I am using "gpt-3.5-turbo-0613" LLM model.

In fact, I have reported this issue in PR 3302 before. When LLM returns the function calling assistant message to us (record 2 below), the Name field is null and the FunctionCall field contains the function to be called. I am using native OpenAI API call here.
Screenshot 2023-11-10 211552
The name field is used only when returning the function result to LLM via the function message (record 3).
Hence, when we insert the function calling assistant message to chat history and pass it back to LLM, the Name field should also be null. Please consider my version of the foreach loop inside the CreateChatCompletionsOptions method in the ClientBase.cs file. This will fix the bug and LLM would not return 500 server error.
```
{
    if (message.AdditionalProperties?.TryGetValue(NameProperty, out string? name) is true)
    {
        if (message.AdditionalProperties?.TryGetValue(ArgumentsProperty, out string? arguments) is true)
        {
            azureMessage.FunctionCall = new FunctionCall(name, arguments);
        }
        else
        {
            azureMessage.Name = name;
        }
    }
}
```

PR #3422, opened 2023-11-08
.Net: Add ReturnParameterView for native functions

momuno:
### Motivation and Context

More specific information on the output of native functions is needed to enable planners.
This change provides a way to describe an output for a native function using an attribute, and a way to view this description along with the output return type via FunctionView.

### Description

- MathPlugin.cs - Adds return Description usage as an example.
- FunctionView.cs - Adds ReturnParameterView to FunctionView
- ReturnParameterView.cs - New ReturnParameterView. Has a description. Waiting on type until design decision made.
- NativeFunction.cs - Adds ReturnParameterView to NativeFunction. Adds ReturnParameterView to MethodDetails and updates logic in GetMethodDetails to access the return parameter's Description.
- SKFunction.cs - Adds ReturnParameterView parameter.
- FunctionViewTests.cs - Adds unit tests for FunctionView and logic around attributes in GetMethodDetails.


dotnet/src/SemanticKernel.Abstractions/Functions/FunctionView.cs
```diff
@@ -15,14 +15,21 @@ namespace Microsoft.SemanticKernel;
...
 {
     /// <summary>
     /// List of function parameters
     /// </summary>
     public IReadOnlyList<ParameterView> Parameters { get; init; } = Parameters ?? Array.Empty<ParameterView>();
+
+    /// <summary>
+    /// Function output
+    /// </summary>
+    public OutputView Output { get; init; } = Output ?? new OutputView();
```
markwallace-microsoft, line 34: `ReturnParameterView` might be a better name, following the pattern here: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.methodinfo.returnparameter?view=net-7.0

Or `ReturnValueView`

dotnet/src/SemanticKernel.Abstractions/Functions/OutputView.cs
```diff
@@ -0,0 +1,15 @@
...
+#pragma warning restore IDE0130
+
+/// <summary>
+/// Class used to copy and export data about function output for planner and related scenarios.
+/// </summary>
+/// <param name="Description">Function output description</param>
+/// <param name="Type">Function output return type</param>
+public sealed record OutputView(
+    string? Description = null,
+    System.Type? Type = null);
```
markwallace-microsoft, line 15: We should have a consistent way to specify types for parameters and return values

markwallace-microsoft (commented):
Like to here others thoughts on the naming

dotnet/src/SemanticKernel.Abstractions/Functions/FunctionView.cs
momuno: Good idea.  I use ReturnParameter in the NativeFunction.cs GetMethodDetails()

lemillermicrosoft (approved):

dotnet/src/Plugins/Plugins.Core/MathPlugin.cs
```diff
@@ -21,6 +21,7 @@ public sealed class MathPlugin
     /// <param name="amount">The amount to add as a string.</param>
     /// <returns>The resulting sum as a string.</returns>
     [SKFunction, Description("Adds an amount to a value")]
+    [return: Description("Summation of two values.")]
```
TaoChenOSU, line 24: nit: the description sounds more like the function description. I would expect the description of the return parameter to be something like "The resulting value".

dotnet/src/Plugins/Plugins.Core/MathPlugin.cs
```diff
@@ -33,6 +34,7 @@ public sealed class MathPlugin
     /// <param name="amount">The amount to subtract as a string.</param>
     /// <returns>The resulting subtraction as a string.</returns>
     [SKFunction, Description("Subtracts an amount from a value")]
+    [return: Description("Difference that results from subtracting an amount from a value")]
```
TaoChenOSU, line 37: nit: Same as above.

TaoChenOSU (approved):

dotnet/src/SemanticKernel.Core/Functions/NativeFunction.cs
```diff
@@ -293,12 +310,16 @@ private void ThrowNotSemantic()
             }
         }
 
-        string? description = method.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description;
+        string? methodDescription = method.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description;
```
SergeyMenshykh, line 313: nit: Consider renaming this variable to `functionDescription` to align it with the one above, `functionName`.

SergeyMenshykh (approved):

dotnet/src/Plugins/Plugins.Core/MathPlugin.cs
momuno: I believe "The resulting value" is too generic. If someone told me the return of a function was "the resulting value", I would say "of course" and then ask, "the resulting value of doing what?"  On the other hand, "summation" is the technical term for the result of adding two values.

Additionally, (yes a pun), this can be updated later to whatever works best for the planners.

dotnet/src/Plugins/Plugins.Core/MathPlugin.cs
momuno: i can just put "the difference"

momuno:
Thanks everyone for your feedback.  Definitely think the naming and descriptions have improved as a result!

lemillermicrosoft (approved):

Issue #3083, opened 2023-10-05
.Net: OpenAI plugins have hardcoded values (payload, server URL, and content type) for body requests


matthewbolanos:
When creating an OpenAI plugin, the request body is called "Payload" and the content type is also requested by the planner.

<img width="1437" alt="image" src="https://github.com/microsoft/semantic-kernel/assets/1409190/9c476d6e-a564-4e17-993b-2488f463cbcc">

To allow a developer to create plugins that behave functionally identical to plugins written with semantic/native functions, the server URL and content type should not be required (this might be possible today) and instead of "Payload" the underlying properities should be exposed to the LLM.

At the end of this work, a user should be able to author the same function as a native function and as an API and have the same rendered prompt created for planners.

matthewbolanos:
Closed since this was completed

PR #3427, opened 2023-11-08
.Net: Remove the artificially created server-url parameter from OpenAPI function parameter view

TaoChenOSU:
### Motivation and Context

The kernel will artificially add a parameter named `serverl-url` to the parameter view of an OpenAPI function when the function is registered to the kernel. The optional `server-url` parameter allows planners to dynamically provide an alternative server url to override the one that is registered with the plugin. The flexibility offered by this parameter can occasionally lead to confusion among planners, causing them to create plans that fail to execute.

Local plugins and OpenAPI plugins should be indifferent to the planners. We are removing this artificially created parameter to improve the performance of planners in creating valid function calls regardless of the function type.

We are aware that there are OpenAPI plugins that require `server-url` override, such as the Jira plugin and the Azure Key Vault plugin. For these plugins, it's still possible to provide a server url at import/registration time, through the optional `OpenAIFunctionExecutionParameters` parameter (please refer to the updated kernel syntax example 22 and 24).

Tracked by: https://github.com/microsoft/semantic-kernel/issues/3083

### Description
1. Remove the `server-url` from the parameter view of OpenAPI functions.
2. Update unit tests.
3. Update kernel syntax example 22 and 24.

### Contribution Checklist

### DIFF
--- a/dotnet/samples/KernelSyntaxExamples/Example22_OpenApiPlugin_AzureKeyVault.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example22_OpenApiPlugin_AzureKeyVault.cs
@@ -49,11 +49,14 @@ public static class Example22_OpenApiPlugin_AzureKeyVault
         var plugin = await kernel.ImportOpenApiPluginFunctionsAsync(
             PluginResourceNames.AzureKeyVault,
             stream!,
-            new OpenApiFunctionExecutionParameters { AuthCallback = authenticationProvider.AuthenticateRequestAsync });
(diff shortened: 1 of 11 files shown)


TaoChenOSU:
Still need to update kernel syntax example 24.

SergeyMenshykh (approved):

glahaye (approved):
Thank you for this.  The code was way too convoluted before and absolutely confused the planners (and software engineers, too)!

teresaqhoang (approved):

SergeyMenshykh (approved):

PR #3434, opened 2023-11-09
.Net: Function Calling Stepwise Planner

gitri-ms:
### Motivation and Context
Stepwise planner built on OpenAI function calling

### Contribution Checklist


dotnet/src/Planners/Planners.OpenAI/Stepwise/InitialPlanPrompt.txt
```diff
@@ -0,0 +1,13 @@
+Create a step-by-step plan to satisfy the user's goal, using only the available functions or your own capabilities.
+
+[AVAILABLE FUNCTIONS]
+
+{{$available_functions}}
```
lemillermicrosoft, line 5: So the functions are given both in the prompt and through the `Functions` parameter of the RequestSettings? How come?

dotnet/src/Planners/Planners.Core/Extensions/FunctionViewExtensions.cs
```diff
@@ -9,14 +9,14 @@ namespace Microsoft.SemanticKernel.Planners;
...
-internal static class FunctionViewExtensions
+public static class FunctionViewExtensions
 {
     /// <summary>
     /// Create a manual-friendly string for a function.
     /// </summary>
     /// <param name="function">The function to create a manual-friendly string for.</param>
     /// <returns>A manual-friendly string for a function.</returns>
-    internal static string ToManualString(this FunctionView function)
+    public static string ToManualString(this FunctionView function)
```
gitri-ms, line 19: Alternatively, if we don't want to make this public, I can add the `InternalsVisibleTo` property in the `Planners.Core` project file.

dotnet/src/Planners/Planners.OpenAI/Stepwise/InitialPlanPrompt.txt
gitri-ms: This planner works in two "phases": (1) ask the model for a set of steps to solve the goal; (2) repeatedly ask the model to solve the next step (using function calling) until a final answer is found.

This is the prompt for phase 1. This request does NOT leverage function calling, as we don't want a function response yet -- just the set of steps we will need to follow. But it is important that the AI know what functions/capabilities are available to use, so that it can return a set of steps that is doable. Hence, including the function manual here.

The second phase is where function calling (and the Functions parameter in request settings) comes in. We begin the chat history with `StepPrompt.txt` as the system message, then repeatedly ask the model to fulfill the next step. Each of these requests includes `Functions` in the request settings, but does not include the function manual in the chat history.

Does that make sense?

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,254 @@
...
+            }
+
+            // Look up SKFunction
+            if (!this._kernel.Functions.TryGetFunctionAndContext(functionResponse, out ISKFunction? pluginFunction, out ContextVariables? funcContext))
+            {
+                throw new SKException($"Function {functionResponse.PluginName}.{functionResponse.FunctionName} not found in kernel.");
+            }
+
+            // Execute function and add to chat history
+            var result = (await this._kernel.RunAsync(funcContext, cancellationToken, pluginFunction).ConfigureAwait(false)).GetValue<object>();
```
lemillermicrosoft, line 121: Did you consider catching this if execution fails and adding that as a user message? I'm not positive what the OpenAI guidance might be here.

dotnet/src/Planners/Planners.OpenAI/Stepwise/InitialPlanPrompt.txt
lemillermicrosoft: Yeah this helps, thank you!

I'm wondering if you considered implementing `CreatePlan` in such a way that the initial set of steps that will be attempted could be visible as part of plan creation.

dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,58 @@
...
+    public static async Task RunAsync()
+    {
+        string[] questions = new string[]
+        {
+            "Write a limerick and send it via email to John"
+        };
+
+        var kernel = InitializeKernel();
+
+        var config = new FunctionCallingStepwisePlannerConfig
```
TaoChenOSU, line 24: nit: without the `{}`
```suggestion
        var config = new FunctionCallingStepwisePlannerConfig();
```

dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        IKernel kernel = new KernelBuilder()
+            .WithLoggerFactory(ConsoleLogger.LoggerFactory)
+            .WithAzureOpenAIChatCompletionService(
+                ChatModelOverride ?? TestConfiguration.AzureOpenAI.ChatDeploymentName,
+                TestConfiguration.AzureOpenAI.Endpoint,
+                TestConfiguration.AzureOpenAI.ApiKey)
+            .Build();
+
+        kernel.ImportFunctions(new EmailPlugin(), "EmailPlugin");
```
TaoChenOSU, line 54: How come we don't need to import a function to write a limerick here (I am thinking something like the joke function)? If the step wise planner can perform certain tasks just by invoking the model, would be nice to document it here.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,254 @@
...
+        this._logger = this._kernel.LoggerFactory.CreateLogger(this.GetType());
+    }
+
+    /// <summary>
+    /// Execute a plan
+    /// </summary>
+    /// <param name="question">The question to answer</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Result containing the model's response message and chat history.</returns>
+    /// <exception cref="SKException"></exception>
```
TaoChenOSU, line 61: nit: this doesn't contain a description.

dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs
TaoChenOSU: Or would be nice to show a sample that requires multiple functions.

alliscode (approved):

dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs
gitri-ms: I wanted to try something that used both the model's own capabilities and external functions. I'll add a comment to highlight that.

This sample does actually require/call multiple functions (get email address and send email), but I can explain that in a comment and also add a couple more examples.

dotnet/src/Planners/Planners.OpenAI/Stepwise/InitialPlanPrompt.txt
```diff
@@ -0,0 +1,13 @@
...
+
+[END AVAILABLE FUNCTIONS]
+
+To create the plan, follow these steps:
+0. Each step should be something that is capable of being done by the list of available functions.
+1. Each step should describe the action, inputs, and outputs of the step.
```
gitri-ms, line 11: Remove - seems to be causing more harm than good

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,254 @@
...
+        var requestSettings = this.PrepareOpenAIRequestSettingsWithFunctions();
+        return (await this._chatCompletion.GetChatCompletionsAsync(chatHistory, requestSettings, cancellationToken).ConfigureAwait(false))[0];
+    }
+
+    private async Task<string> GetFunctionsManualAsync(CancellationToken cancellationToken)
+    {
+        var availableFunctions = await this._kernel.Functions.GetFunctionsAsync(this.Config, null, this._logger, cancellationToken).ConfigureAwait(false);
+
+        return string.Join("\n\n", availableFunctions.Select(x => x.ToManualString()));
+    }
```
gitri-ms, line 161: This manual should match the format of the function descriptions sent to OpenAI - it currently does not.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
gitri-ms: I don't believe OpenAI provides guidance on what to do in this case (as it's on the app to handle execution)... but this is a great idea! I'm already seeing improvements in success rate by giving the planner a chance to try something different in case of failure. Thanks!

dotnet/src/Planners/Planners.Core/Extensions/FunctionViewExtensions.cs
gitri-ms, line 19: Revert to internal

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIFunctionResponse.cs
```diff
@@ -43,24 +43,31 @@ public class OpenAIFunctionResponse
     /// <returns>Instance of <see cref="OpenAIFunctionResponse"/>.</returns>
     public static OpenAIFunctionResponse FromFunctionCall(FunctionCall functionCall)
     {
-        OpenAIFunctionResponse response = new();
-        if (functionCall.Name.Contains(OpenAIFunction.NameSeparator))
+        try
```
gitri-ms, line 46: Remove unnecessary try/catch

dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Planners;
+using Microsoft.SemanticKernel.Plugins.Core;
+using Plugins;
+using RepoUtils;
+
+// ReSharper disable once InconsistentNaming
+public static class Example66_FunctionCallingStepwisePlanner
+{
+    internal static string? ChatModelOverride = null; //"gpt-35-turbo-0613"
```
markwallace-microsoft, line 14: Should this ne null?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIFunction.cs
```diff
@@ -77,7 +77,7 @@ public class OpenAIFunction
     /// <summary>
     /// Separator between the plugin name and the function name
     /// </summary>
-    public const string NameSeparator = "-";
+    public const string NameSeparator = "_";
```
markwallace-microsoft, line 80: Glad we made this change as using `-` made it look like a matchs operation. Was there a use case that caused the old separator to change?

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,309 @@
...
+
+        // Initialize prompt renderer
+        this._promptTemplateFactory = new BasicPromptTemplateFactory(this._kernel.LoggerFactory);
+
+        // Set up Config with default values and excluded plugins
+        this.Config = config ?? new();
+        this.Config.ExcludedPlugins.Add(RestrictedPluginName);
+
+        this._initialPlanPrompt = EmbeddedResource.Read("Stepwise.InitialPlanPrompt.txt");
+        this._stepPrompt = EmbeddedResource.Read("Stepwise.StepPrompt.txt");
```
markwallace-microsoft, line 50: Should we allow these prompts to be overridden via the config object?

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,309 @@
...
+    /// <summary>
+    /// Execute a plan
+    /// </summary>
+    /// <param name="question">The question to answer</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Result containing the model's response message and chat history.</returns>
+    /// <exception cref="SKException">If no goal found or failed to retrieve function from kernel.</exception>
+    public async Task<FunctionCallingStepwisePlannerResult> ExecuteAsync(
+        string question,
+        CancellationToken cancellationToken = default)
```
markwallace-microsoft, line 65: This is a very long method and hard to grok without reading in a lot of detail so it will be hard to maintain. Can you consider extracting some methods to make the logic easier to follow.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,309 @@
...
+    {
+        string planResponse = string.Empty;
+
+        if (string.IsNullOrEmpty(question))
+        {
+            throw new SKException("Goal not found.");
+        }
```
markwallace-microsoft, line 72: Use `Verify.NotNull` so the correct exception is thrown.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,309 @@
...
+        // We've completed the max iterations, but the model hasn't returned a final answer.
+        return new FunctionCallingStepwisePlannerResult
+        {
+            FinalAnswer = string.Empty,
+            ChatHistory = chatHistoryForSteps,
+            Iterations = this.Config.MaxIterations,
+        };
+    }
+
+    private async Task<IChatResult> GetCompletionWithFunctionsAsync(
```
markwallace-microsoft, line 193: Private stuff outside the private region

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,309 @@
...
+    private const string AvailableFunctionsKey = "available_functions";
+    private const string InitialPlanKey = "initial_plan";
+    private const string GoalKey = "goal";
+
+    #endregion private
+
+    /// <summary>
+    /// Contains helper functions used by the <see cref="FunctionCallingStepwisePlanner"/>
+    /// </summary>
+    public class UserInteraction
```
markwallace-microsoft, line 296: Should this be internal and/or sealed?

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -0,0 +1,309 @@
...
+
+#pragma warning disable IDE0130
+// ReSharper disable once CheckNamespace - Using NS of Plan
+namespace Microsoft.SemanticKernel.Planners;
+#pragma warning restore IDE0130
+
+/// <summary>
+/// A planner that uses OpenAI function calling in a stepwise manner to fulfill a user goal or question.
+/// </summary>
+public class FunctionCallingStepwisePlanner
```
markwallace-microsoft, line 27: Please seal this

dotnet/src/Planners/Planners.OpenAI/Stepwise/InitialPlanPrompt.txt
```diff
@@ -0,0 +1,12 @@
+You are Gina, an expert at generating plans from a given GOAL. Think step by step and determine a plan to satisfy the specified GOAL using only the FUNCTIONS provided to you. You can also make use of your own knowledge while forming an answer but you must not use functions that are not provided. Once you have come to a final answer, use the UserInteraction_SendFinalAnswer function to communicate this back to the user.
```
markwallace-microsoft, line 1: Please remove the name "Gina"

dotnet/src/Planners/Planners.OpenAI/Utils/EmbeddedResource.cs
```diff
@@ -0,0 +1,26 @@
...
+#pragma warning restore IDE0130
+
+internal static class EmbeddedResource
+{
+    private static readonly string? s_namespace = typeof(EmbeddedResource).Namespace;
+
+    internal static string Read(string name)
+    {
+        var assembly = typeof(EmbeddedResource).GetTypeInfo().Assembly;
+        if (assembly == null) { throw new SKException($"[{s_namespace}] {name} assembly not found"); }
```
markwallace-microsoft, line 18: Can you check is there a standard .Net exception we should use here

dotnet/src/Planners/Planners.OpenAI/Utils/EmbeddedResource.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        using Stream? resource = assembly.GetManifestResourceStream($"{s_namespace}." + name);
```
markwallace-microsoft, line 20: Can you check is there a standard .Net exception we should use here

markwallace-microsoft (commented):
I will need to do another review of the long function once it is broken into smaller parts

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIFunction.cs
gitri-ms: @alliscode made this change - not sure if it was in response to a particular use case or not.

dotnet/src/Planners/Planners.OpenAI/Utils/EmbeddedResource.cs
gitri-ms: `FileNotFoundException` seems like the closest for this and the comment below.

I copied the `EmbeddedResource` class from `Planners.Core` (and I see there's also one in `Experimental.Orchestration.Flow` so I will update the exceptions in those files as well.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
gitri-ms: Yes to sealed, but it has to be public to be registered with the kernel / included in the function list.

dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs
gitri-ms: I added a comment to clarify this

markwallace-microsoft (approved):

matthewbolanos (approved):

markwallace-microsoft (approved):

Issue #3129, opened 2023-10-11
Add OpenAPI skill return type schema to the skill output object


matthewbolanos:
> [!IMPORTANT]  
> Labeled **High** because it will not require a breaking change, but it's important to complete by v1.0.0 to improve the quality of planners.

In planners, the output schema of a function should be provided so the planner knows how to use it. 

Ideally, the output schema (and input schema) should leverage Type TypeScript so it works with Typechat in the future.

matthewbolanos:
Closed because this is now possible

Issue #1978, opened 2023-07-12
.Net: Allow native functions to have complex types for inputs/outputs


nschuessler:
Currently only primitive types are supported for native code skills (i.e. string, int, double).
Would a PR be allowed to add checks to `SkFunction.cs` to allow generic collections of primitive types (short term), until a more general solution is available?

i.e. ideally you would test if the return type is serializable to JSON (if that is what the concern is), even if it a customer provided class/struct.

Example:

```
+        if (targetType.IsGenericType &&
+            targetType.GetGenericTypeDefinition().Equals(typeof(List<>)) &&
+            GenericTypeConverter.TryGetParameterConverters(targetType, out TypeConverter[]? parameterConverters))
+        {
+            return new ListTypeConverter(targetType, parameterConverters);
+        }
```

```
public static bool TryGetParameterConverters(Type targetType, [NotNullWhen(true)] out TypeConverter[]? converters)
    {
        converters = null;

        Type[] typeParameters = targetType.GetGenericArguments();
        if (typeParameters != null && typeParameters.Length > 0)
        {
            int count = 0;
            TypeConverter[] tempConverters = new TypeConverter[typeParameters.Length];
            for (int i = 0; i < typeParameters.Length; i++)
            {
                TypeConverter? temp = SKFunction.GetTypeConverter(typeParameters[i]);
                if (temp == null)
                {
                    break;
                }

                tempConverters[i] = temp;
                count++;
            }

            if (count == typeParameters.Length)
            {
                converters = tempConverters;
            }
        }

        return converters != null;
    }
```

nschuessler:
@shawncal What is the usual triage turn around time? thx

nacharya1:
@shawncal  - can you take a look into this?

ambrose-leung:
This may be related to: https://github.com/microsoft/semantic-kernel/issues/2081

SergeyMenshykh:
@nschuessler there have been many improvements introduced for functions since the issue was created. Today, generic collections with primitives are supported.
<img width="678" alt="image" src="https://github.com/microsoft/semantic-kernel/assets/68852919/6783631e-aecf-4a7c-8362-e6a4d30fdd9e">
Please get latest version of SK [RC3](https://www.nuget.org/packages/Microsoft.SemanticKernel/) ATM and let us know if it covers your scenario,

SergeyMenshykh:
Closing the issue for now, assuming the problem is resolved. Feel free to reopen it if that's not the case.

PR #3450, opened 2023-11-10
Add support for OpenAPI responses and schema validation

lemillermicrosoft:
## Summary
This pull request adds support for OpenAPI responses and schema validation to the RestApiOperationRunner and RestApiOperationResponse classes. Specifically, it adds a new parameter to the SendAsync method in RestApiOperationRunner to allow for passing in a dictionary of expected responses. If a matching response is found, the response schema is updated with the schema from the expected response. Additionally, the CreateRestApiOperationResponses method was added to parse the OpenAPI responses and create a dictionary of expected responses. RestApiOperationResponse now includes a new property 'Schema' to specify the schema against which the response body should be validated, and a new method 'IsValid' to validate the response content against the schema. RestApiOperation now includes a new property 'Responses' to specify possible operation responses. The KernelOpenApiPluginExtensions was modified to use the correct response when multiple responses are defined in an operation. RestApiOperationResponseTests were added to test RestApiOperationResponse objects with and without schemas, and RestApiOperationRunnerTests were updated to use the new RestApiOperationResponse.IsValid() method. Additionally, test cases were added for RestApiOperationRunner and OpenApiPluginFunctions, and a test was added for the Instacart plugin.

## Changes
- Added a new parameter to the SendAsync method in RestApiOperationRunner to allow for passing in a dictionary of expected responses
- Added the CreateRestApiOperationResponses method to parse the OpenAPI responses and create a dictionary of expected responses
- Added a new property 'Schema' to RestApiOperationResponse to specify the schema against which the response body should be validated
- Added a new method 'IsValid' to RestApiOperationResponse to validate the response content against the schema
- Added a new property 'Responses' to RestApiOperation to specify possible operation responses
- Modified KernelOpenApiPluginExtensions to use the correct response when multiple responses are defined in an operation
- Added RestApiOperationResponseTests to test RestApiOperationResponse objects with and without schemas
- Updated RestApiOperationRunnerTests to use the new RestApiOperationResponse.IsValid() method
- Added test cases for RestApiOperationRunner and OpenApiPluginFunctions
- Added a test for the Instacart plugin

---
Related:
#3129
#1978

### Contribution Checklist


dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -285,6 +286,74 @@ private static List<RestApiOperationParameter> CreateRestApiOperationParameters(
...
+            Type = schema.Type
+            // Add other properties as needed
+        };
+
+        if (schema.Type == "object" && schema.Properties != null)
+        {
+            foreach (var property in schema.Properties)
+            {
+                // Recursively parse nested schemas
+                var parsedSchema = ParseSchema(property.Value);
```
lemillermicrosoft, line 340: @alliscode Here is where I did some recursion to get things working for me. Maybe similar is needed for the `JsonSchema.Net` stuff.

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperationResponseSchema.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Json.Schema;
+
+namespace Microsoft.SemanticKernel.Functions.OpenAPI.Model;
+
+/// <summary>
+/// The REST API operation response schema
+/// </summary>
+public sealed class RestApiOperationResponseSchema
```
alliscode, line 12: I'm curious why have `RestApiOperationResponseSchema` to hold the JsonDocument with title and type, instead of just a JsonDocument on `RestApiOperationResponse`?

lemillermicrosoft: Yup, good call. I had the same question as I was reducing things. I thought perhaps it'd be nice having the top level title/type available still but yeah for now will just reduce the scope of things.

lemillermicrosoft: Removed.

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperationResponse.cs
```diff
@@ -20,14 +22,49 @@ public sealed class RestApiOperationResponse
...
+    /// </summary>
+    /// <returns>True if the response is valid, false otherwise.</returns>
+    /// <remarks>
+    /// If the schema is not specified, the response is considered valid.
+    /// </remarks>
+    public bool IsValid()
+    {
+        if (this.Schema is null)
+        {
+            return true; // todo add tests
```
lemillermicrosoft, line 54: TODO

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -287,6 +287,36 @@ private static List<RestApiOperationParameter> CreateRestApiOperationParameters(
         return new RestApiOperationPayload(mediaType, payloadProperties, requestBody.Description, mediaTypeMetadata?.Schema?.ToJsonDocument());
     }
 
+    private static IEnumerable<(string, RestApiOperationResponse)> CreateRestApiOperationResponses(OpenApiResponses responses)
+    {
+        if (responses == null)
+        {
+            yield break; // todo add tests
```
lemillermicrosoft, line 294: Do this.

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -287,6 +287,36 @@ private static List<RestApiOperationParameter> CreateRestApiOperationParameters(
...
+        }
+
+        foreach (var response in responses)
+        {
+            if (response.Value == null)
+            {
+                yield break; // todo add tests
```
lemillermicrosoft, line 301: Do this.

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -287,6 +287,36 @@ private static List<RestApiOperationParameter> CreateRestApiOperationParameters(
...
+            if (mediaType == null)
+            {
+                // TODO -- use text/plain or application/json as default?
+                yield return (response.Key, new RestApiOperationResponse(string.Empty, string.Empty));
+            }
+            else
+            {
+                var matchingSchema = response.Value.Content[mediaType].Schema;
+
+                // TODO - Kind of overloading `content` and `contentType` properties here.
```
lemillermicrosoft, line 314: Do this.

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -287,6 +287,36 @@ private static List<RestApiOperationParameter> CreateRestApiOperationParameters(
...
+        {
+            if (response.Value == null)
+            {
+                yield break; // todo add tests
+            }
+
+            var mediaType = s_supportedMediaTypes.FirstOrDefault(smt => response.Value.Content.ContainsKey(smt));
+            if (mediaType == null)
+            {
+                // TODO -- use text/plain or application/json as default?
```
lemillermicrosoft, line 307: Do this.

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
```diff
@@ -171,7 +173,25 @@ internal sealed class RestApiOperationRunner
...
+        var response = await SerializeResponseContentAsync(responseMessage.Content).ConfigureAwait(false);
+
+        if (responses is not null)
+        {
+            var statusCodeKey = $"{(int)responseMessage.StatusCode}";
+            var matchingResponse = responses.FirstOrDefault(r => r.Key == statusCodeKey).Value;
+
+            if (matchingResponse is null)
+            {
+                matchingResponse = responses.FirstOrDefault(r => r.Key == "default").Value; // todo add tests
```
lemillermicrosoft, line 185: Do this.

dotnet/src/Functions/Functions.UnitTests/OpenAPI/RestApiOperationResponseTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        //Act
+        var result = response.IsValid();
+
+        //Assert
+        Assert.True(result);
+    }
+
+    [Theory]
+    [InlineData("{\"products\": [{\"id\": \"1234\", \"name\": \"Laptop\"}]}", "application/json", "{\"type\": \"object\"}")] // TODO -- Is this good enough/complete?
```
TaoChenOSU, line 67: Is the TODO at the end of this line intentional?

alliscode (approved):

dotnet/src/Functions/Functions.UnitTests/OpenAPI/RestApiOperationResponseTests.cs
lemillermicrosoft: Thank you! It's not. I do think I should add tests (and functionality) for the possible `ContentType` here.

lemillermicrosoft: Also above on line 47 -- curious what folks think the correct behavior should be there.

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelOpenApiPluginExtensions.cs
```diff
@@ -268,9 +268,27 @@ async Task<RestApiOperationResponse> ExecuteAsync(SKContext context)
             })
             .ToList();
 
+        RestApiOperationResponse? restOperationResponse = null;
+        foreach (var response in operation.Responses)
+        {
+            if (response.Key == "200" || response.Key == "2XX" || (response.Key == "default" && restOperationResponse == null))
```
markwallace-microsoft, line 274: Can we check if the key starts with `2` or use a regex to match e.g. `^3\d\d$`
Do we need to worry about the case of "default"?

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
```diff
@@ -171,7 +173,34 @@ internal sealed class RestApiOperationRunner
...
+        {
+            var statusCodeKey = $"{(int)responseMessage.StatusCode}";
+
+            // Exact Match
+            var matchingResponse = responses.FirstOrDefault(r => r.Key == statusCodeKey).Value;
+
+            // Wildcard match e.g. 2XX
+            if (matchingResponse is null)
+            {
+                matchingResponse = responses.FirstOrDefault(r => r.Key.EndsWith("X", StringComparison.InvariantCultureIgnoreCase) && statusCodeKey.StartsWith(r.Key.Substring(0, 1), StringComparison.InvariantCultureIgnoreCase)).Value;
```
markwallace-microsoft, line 188: This only matches first and last characters. Can we convert the key to a regex and use that to match?

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
```diff
@@ -171,7 +173,34 @@ internal sealed class RestApiOperationRunner
...
+            }
+
+            // Default
+            if (matchingResponse is null)
+            {
+                matchingResponse = responses.FirstOrDefault(r => r.Key == "default").Value;
```
markwallace-microsoft, line 194: Define "default" as const somewhere

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
```diff
@@ -171,7 +173,34 @@ internal sealed class RestApiOperationRunner
...
         using var responseMessage = await this._httpClient.SendWithSuccessCheckAsync(requestMessage, cancellationToken).ConfigureAwait(false);
 
-        return await SerializeResponseContentAsync(responseMessage.Content).ConfigureAwait(false);
+        var response = await SerializeResponseContentAsync(responseMessage.Content).ConfigureAwait(false);
+
+        if (responses is not null)
+        {
+            var statusCodeKey = $"{(int)responseMessage.StatusCode}";
+
+            // Exact Match
```
markwallace-microsoft, line 182: Suggestion: create a private method to do the matching stuff

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -287,6 +287,21 @@ private static List<RestApiOperationParameter> CreateRestApiOperationParameters(
...
+    {
+        foreach (var response in responses)
+        {
+            var mediaType = s_supportedMediaTypes.FirstOrDefault(smt => response.Value.Content.ContainsKey(smt));
+            if (mediaType is not null)
+            {
+                var matchingSchema = response.Value.Content[mediaType].Schema;
+                var description = response.Value.Description ?? matchingSchema?.Description ?? string.Empty;
+
+                yield return (response.Key, new RestApiOperationResponse(description, mediaType, matchingSchema?.ToJsonDocument()));
```
SergeyMenshykh, line 300: Consider creating and using a dedicated model class with three properties - description, schema, and contentType, to represent a response schema, rather than reusing the response class. 
   
This class was designed to represent a REST API operation call result, rather than representing a model class meant to pass information from the parser to the runner. It may also be extended in future when more data (e.g a correlation vector from response headers) need to be exposed to the OpenAPI skills callers and the change should only impact RestApiOperationRunner and not the parser.

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperation.cs
```diff
@@ -59,6 +59,11 @@ public sealed class RestApiOperation
     /// </summary>
     public IList<RestApiOperationParameter> Parameters { get; }
 
+    /// <summary>
+    /// The list of possible operation responses.
+    /// </summary>
+    public IDictionary<string, RestApiOperationResponse> Responses { get; }
```
SergeyMenshykh, line 65: ResponseSchemas?

dotnet/src/Functions/Functions.UnitTests/OpenAPI/RestApiOperationResponseTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        //Act
+        var result = response.IsValid();
+
+        //Assert
+        Assert.True(result);
+    }
+
+    [Theory]
+    [InlineData("{\"p\": [{\"id\": \"1234\", \"name\": \"Laptop\"}", "application/json", "{\"title\":\"ProductResponse\",\"type\":\"object\",\"properties\":{\"products\":{\"type\":\"array\",\"items\":{\"title\":\"Product\",\"type\":\"object\",\"properties\":{\"attributes\":{\"type\":\"array\",\"items\":{\"type\":\"string\"}},\"name\":{\"type\":\"string\"},\"price\":{\"type\":\"string\"},\"url\":{\"type\":\"string\"}}}}}}")]
```
SergeyMenshykh, line 82: The inline JSONs are unreadable. Consider putting them into resource files with descriptive names and use the names in the InlineData attributes instead.

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelOpenApiPluginExtensions.cs
SergeyMenshykh, line 274: nit: .StartsWith("2") ?

dotnet/src/Functions/Functions.UnitTests/OpenAPI/RestApiOperationRunnerTests.cs
```diff
@@ -1009,6 +1010,77 @@ public async Task ItShouldThrowExceptionForUnsupportedContentTypeAsync()
...
+    {
+        public IEnumerator<object[]> GetEnumerator()
+        {
+            yield return new object[] {
+                    "default",
+                    new (string, RestApiOperationResponse)[] {
+                        ("400", new RestApiOperationResponse("fake-content", "fake-content-type", JsonDocument.Parse("{\"title\":\"FakeResponse\",\"type\":\"object\",\"properties\":{\"fakeItems\":{\"type\":\"array\",\"items\":{\"title\":\"Item\",\"type\":\"object\",\"properties\":{\"attributes\":{\"type\":\"array\",\"itemName\":{\"type\":\"string\"}},\"name\":{\"type\":\"string\"}}}}}}"))),
+                        ("default", new RestApiOperationResponse("Default response content", "application/json", JsonDocument.Parse("{\"title\":\"DefaultResponse\",\"type\":\"object\",\"properties\":{\"code\":{\"type\":\"integer\",\"format\":\"int32\"},\"message\":{\"type\":\"string\"}}}"))),
+                    },
+            };
```
SergeyMenshykh, line 1023: This code is read-only, making it very difficult to support in the future if any adjustments are required. Consider either to read JSONs from resource file(s) or build C# object and serialize it to JSON.

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
lemillermicrosoft: Yeah originally I did have similar. Can you look at the `prev` in the following commit and let me know if that aligns with what you are thinking? Thanks!

https://github.com/microsoft/semantic-kernel/pull/3450/commits/86e91300867a8c702b945fd4ea513f346b80ec3e#diff-e3dd20c9a18b37db4537970a29d06152ab4792660943d6779f4697a9ec6886c0

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelOpenApiPluginExtensions.cs
lemillermicrosoft: There are options here. The most significant impact here will be on Planning. 

1. Return any response that is 200-299 or 2XX.
2. Return the first response that is 200-299 or 2XX, in that order.
3. Return all responses.

I do believe we need to worry about default - https://spec.openapis.org/oas/latest.html#fixed-fields-13

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
lemillermicrosoft: The intent is to match with the first digit of the response code and XX. Could change to compare with `$"{statusCodeKey.Substring(0,1)}XX"`. If the response code is 403 and 403 is not defined we should not match with any other 4** code other than the literal string "4XX".

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
lemillermicrosoft: Also, the goal here is for the Schema to be exposed to the OpenAPI skill callers. Both so the object can validate itself or callers implement their own validation. As well as planners can represent the expectations in their manuals.

However, I do see the concern. If I understand correctly, the commit previous I shared could be improved to have the RestApiOperationResponseSchema be a member of a new/different class than RestApiOperationResponse. Will take a stab at this.

dotnet/src/Functions/Functions.OpenAPI/RestApiOperationRunner.cs
```diff
@@ -139,13 +139,15 @@ internal sealed class RestApiOperationRunner
...
     /// <param name="payload">HTTP request payload.</param>
+    /// <param name="responses">The dictionary of expected responses.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>Response content and content type</returns>
     private async Task<RestApiOperationResponse> SendAsync(
         Uri url,
         HttpMethod method,
         IDictionary<string, string>? headers = null,
         HttpContent? payload = null,
+        IDictionary<string, RestApiOperationResponse>? responses = null,
```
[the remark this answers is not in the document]
lemillermicrosoft: Changed this to just be the JsonDocument for now and updated naming.

dotnet/src/Functions/Functions.OpenAPI/OpenApi/OpenApiDocumentParser.cs
SergeyMenshykh: > Yeah originally I did have similar. Can you look at the `prev` in the following commit and let me know if that aligns with what you are thinking? Thanks! See usage of `RestApiOperationResponseSchema`
> 
> [86e9130#diff-e3dd20c9a18b37db4537970a29d06152ab4792660943d6779f4697a9ec6886c0](https://github.com/microsoft/semantic-kernel/commit/86e91300867a8c702b945fd4ea513f346b80ec3e#diff-e3dd20c9a18b37db4537970a29d06152ab4792660943d6779f4697a9ec6886c0)

Yes, the RestApiOperationResponseSchema class is what we need here. However, there are a few small changes required - rename the Title property to Description, the Type property to MediaType, and remove the IsValid method. Additionally, it is unclear whether we need the Json* attributes applied to its properties.

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelOpenApiPluginExtensions.cs
lemillermicrosoft: I've refactored things here. Feedback appreciated.

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperation.cs
lemillermicrosoft: For now I've named the new class `RestApiOperationExpectedResponse`. Would appreciate feedback on this naming. I like `Responses` still just because that matches the OpenAPI definitions but could see renaming to `ExpectedResponses`. Thoughts?

dotnet/src/Functions/Functions.OpenAPI/Extensions/KernelOpenApiPluginExtensions.cs
lemillermicrosoft: Refactored this a bit. Feedback appreciated.

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperation.cs
lemillermicrosoft: Resolving as-is for now.

markwallace-microsoft (approved):

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #3452, opened 2023-11-11
.Net: Allow to exclude operations when importing and OpenAPI swagger

dluc:
### Motivation and Context

When importing OpenAPI plugins, there can be operations not supported by SK, triggering an exception.

For instance, SK currently supports only "application/json" and "text/plain" media types, and `kernel.ImportOpenApiPluginFunctionsAsync` throws an exception when encountering an operation that uses "multipart/form-data" (`SKException($"Neither of the media types of {operationId} is supported.")`).

### Description

Add new option to `OpenApiFunctionExecutionParameters` allowing to exclude some operations, e.g. those that would otherwise cause an exception.


markwallace-microsoft (approved):
LGTM

alliscode (approved):

SergeyMenshykh (approved):

PR #3453, opened 2023-11-11
Better manuals

alliscode:
### Motivation and Context

To give planners more information about how to use plugin functions, we need to retain more information from native functions and remote OpenAPI specs.

### Description
- Adds a Shema to FunctionParameterVIew and FunctionReturnParameterView that is populated when loading remote plugins.
- Adds a native Type to FunctionParameterVIew and FunctionReturnParameterView that is populated when loading native plugins.
- Adds a minimal Json Schema model to core abstractions to be used for function manuals.

Still todo:
- Update Connectors.AI.OpenAI function calling to use minimal Json Schemas from core abstractions.
- Propagate output schemas from OpenAPI spec when loading remote plugins.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Functions/ParameterView.cs
```diff
@@ -13,9 +16,13 @@ namespace Microsoft.SemanticKernel;
...
+/// <param name="NativeType">The native type. Null if this parameter did not come from a native function.</param>
+/// <param name="Schema">The JSON Schema of the type. May be null for native function parameters.</param>
 public sealed record ParameterView(
     string Name,
     string? Description = null,
     string? DefaultValue = null,
     ParameterViewType? Type = null,
-    bool? IsRequired = null);
+    bool? IsRequired = null,
+    Type? NativeType = null,
```
markwallace-microsoft, line 27: How about using `ParameterType` to be consistent with reflection API.

https://learn.microsoft.com/en-us/dotnet/api/system.reflection.parameterinfo.parametertype?view=net-7.0

dotnet/src/Functions/Functions.OpenAPI/Extensions/OpenApiSchemaExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Gets a JSON serialized representation of an <see cref="OpenApiSchema"/>
+    /// </summary>
+    /// <param name="schema">The schema.</param>
+    /// <returns></returns>
+    internal static JsonDocument ToJsonDocument(this OpenApiSchema schema)
+    {
+        var schemaBuilder = new StringBuilder();
+        schema.SerializeAsV3WithoutReference(new OpenApiJsonWriter(new StringWriter(schemaBuilder)));
```
lemillermicrosoft, line 20: If the schema is an array of objects, this doesn't seem to do as named.

> {
  "type": "array",
  "items": {
    "$ref": "#/components/schemas/Product"
  }
}

or
> {"title":"ProductResponse","type":"object","properties":{"products":{"type":"array","items":{"$ref":"#/components/schemas/Product"}}}}

alliscode: Hmm, let me look into this...

alliscode: I pushed up a change that appears to fix this, can you give it another shot?

dotnet/src/SemanticKernel.Core/Functions/NativeFunction.cs
```diff
@@ -306,7 +306,9 @@ ValueTask<FunctionResult> Function(ITextCompletion? text, AIRequestSettings? req
             Name = functionName!,
             Description = method.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description ?? "",
             Parameters = stringParameterViews,
-            ReturnParameter = new ReturnParameterView(method.ReturnParameter.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description ?? ""),
+            ReturnParameter = new ReturnParameterView(
+                Description: method.ReturnParameter.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description ?? "",
```
SergeyMenshykh, line 310: nit: unnecessary empty line.

SergeyMenshykh (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/CustomClient/OpenAIClientBase.cs
```diff
@@ -122,7 +121,7 @@ private protected async Task<T> ExecutePostRequestAsync<T>(string url, string re
 
     private protected T JsonDeserialize<T>(string responseJson)
     {
-        var result = Json.Deserialize<T>(responseJson);
+        var result = Microsoft.SemanticKernel.Text.Json.Deserialize<T>(responseJson);
```
markwallace-microsoft, line 124: What's the reason for the longhand here and elsewhere?

dotnet/src/SemanticKernel.Abstractions/Schema/JsonSchemaFunctionParameters.cs
```diff
@@ -0,0 +1,34 @@
...
+/// <summary>
+/// A class to describe the parameters of an SKFunction in a Json Schema friendly way.
+/// </summary>
+public sealed class JsonSchemaFunctionParameters
+{
+    /// <summary>
+    /// The type of schema which is always "object" when describing function parameters.
+    /// </summary>
+    [JsonPropertyName("type")]
+    public string Type => "object";
```
markwallace-microsoft, line 21: Why is this needed if it never changes?

If it stays should it be called `ParameterType`?

dotnet/src/Connectors/Connectors.AI.OpenAI/Connectors.AI.OpenAI.csproj
```diff
@@ -26,5 +26,6 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.AI.OpenAI" />
+    <PackageReference Include="JsonSchema.Net.Generation" />
```
markwallace-microsoft, line 29: Should this be a dependency here?

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Schema/JsonSchemaFunctionParameters.cs
alliscode: This is fixed value because we're describing the function as an object and its parameters as properties on the object. This is the defined format that OpenAI expects. The json name needs to be type per the OpenAI spec, but I could change the C# property name if you think that would make more sense.


Example:
```
{
        "name": "search_hotels",
        "description": "Retrieves hotels from the search index based on the parameters provided",
        "parameters": {
            "type": "object",
            "properties": {
                "location": {
                    "type": "string",
                    "description": "The location of the hotel (i.e. Seattle, WA)"
                },
                "max_price": {
                    "type": "number",
                    "description": "The maximum price for the hotel"
                },
                "features": {
                    "type": "string",
                    "description": "A comma separated list of features (i.e. beachfront, free wifi, etc.)"
                }
            },
            "required": ["location"]
        }
    }
```

dotnet/src/Connectors/Connectors.AI.OpenAI/Connectors.AI.OpenAI.csproj
alliscode: Yes the OpenAI connector needs to be able to generate the Schemas so it needs this library.

stephentoub:
I've been discussing this offline with @alliscode. We're going to cull back how much is exposed publicly here, at least for now.

PR #3467, opened 2023-11-13
.Net: OpenAI Connector Internals Update Azure Beta 9 - Option 1

### LINKED ISSUE #3432: .Net: Support Latest Changes in Azure SDK Streaming Interfaces (Beta-9)
## Azure.AI.OpenAI_1.0.0-beta.9 introduces breaking changes to Streaming interfaces

https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.OpenAI_1.0.0-beta.9/sdk/openai/Azure.AI.OpenAI/CHANGELOG.md#100-beta9-2023-11-06

This issue represents the request to support the new interfaces in Semantic Kernel OpenAI connector introduced in `beta-9`.

### LINKED ISSUE #3492: .Net: Accessing SKChatMessage.FunctionCall throws exception when the OpenAI model does not return a function call
**Describe the bug**
When streaming the chat response and checking if the OpenAI model returned a function call accessing `SKChatMessage.FunctionCall` throws `NotSupportedException`.

```
Exception has occurred: CLR/System.NotSupportedException
Exception thrown: 'System.NotSupportedException' in System.Private.CoreLib.dll: 'Function call is not supported'
   at Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk.SKChatMessage.get_FunctionCall()
   at Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk.ChatStreamingResultExtensions.<GetOpenAIStreamingFunctionResponseAsync>d__0.MoveNext()
   at Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk.ChatStreamingResultExtensions.<GetOpenAIStreamingFunctionResponseAsync>d__0.MoveNext()
   ...
```

Reference: https://github.com/microsoft/semantic-kernel/pull/3246

**To Reproduce**
Code similar to [Example59_OpenAIFunctionCalling.cs](https://github.com/RogerBarreto/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs#L160)

```csharp
var chatCompletionResults = this.Chat.GetStreamingChatCompletionsAsync(
    chat: chatHistory,
    requestSettings: requestSettings,
    cancellationToken: cancellationToken);

await foreach (var chatCompletionResult in chatCompletionResults)
{
    await foreach (var chatMessageStream in chatCompletionResult.GetStreamingChatMessageAsync(cancellationToken))
    {
        if (!string.IsNullOrEmpty(chatMessageStream.Content))
            yield return chatMessageStream.Content;
    }

    var functionResponse = await chatCompletionResult.GetOpenAIStreamingFunctionResponseAsync();
    yield break;
}
```

**Expected behavior**
`SKChatMessage.FunctionCall` should return `null` instead.

**Platform**
 - OS: macOS Sonoma 14.1.1
 - IDE: VS Code 1.84.2
 - Language: C# 7.0.403
 - Source: Microsoft.SemanticKernel 1.0.0-beta6

RogerBarreto:
### Motivation and Context

Resolves #3432 
Resolves #3492

In this approach we'll keep the same behavior as it was expected by the users. The changes will be limited to internal changes in components which will allow multiple streaming choices to be handled per choice basis by the users.

With the new flattened streaming API we'll need to keep track of the choices and their indexes as they arrive in the streaming data. This will allow us to provide the correct choice to the user when they request it. To be able to provide the choice abstraction back, the changes demanded a background working approach where the stream is managed internally changing the choices details by reference as the data arrives.

### Description

Changing internal implementations to keep the same behavior and interfaces using the new streaming API constructs.

### Contribution Checklist


docs/decisions/0022-azuresdk-streaming-api-update.md
```diff
@@ -0,0 +1,251 @@
...
+deciders: SergeyMenshykh, markwallace-microsoft, rogerbarreto
+consulted:
+informed:
+---
+
+# How to handle streaming API Changes in Azure SDK beta 9
+
+## Context and Problem Statement
+
+Azure Sdk changes their interface for streaming API in beta 9. This change is not backward compatible and requires changes in our code.
```
stephentoub, line 14: The old APIs aren't still available?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatStreamingResult.cs
```diff
@@ -15,41 +16,100 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
...
+        {
+            if (string.IsNullOrEmpty(role))
+            {
+                role = message.Role.ToString();
+            }
+
+            fullMessage.Append(message.Content);
+        }
+
+        if (fullMessage.Length == 0)
```
stephentoub, line 44: It's not possible that we got back an empty message?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatStreamingResult.cs
```diff
@@ -15,41 +16,100 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
...
 
     /// <inheritdoc/>
     public async IAsyncEnumerable<ChatMessageBase> GetStreamingChatMessageAsync([EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
-        await foreach (var message in this._choice.GetMessageStreaming(cancellationToken))
+        string role = string.Empty;
+        int i = 0;
+        CompletionsFinishReason? finishReason = null;
+        string? messageId = null;
+        DateTimeOffset? created = null;
```
stephentoub, line 59: Why `DateTimeOffset? create` rather than `DateTimeOffset`?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatStreamingResult.cs
```diff
@@ -15,41 +16,100 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
...
+            {
+                created = message.Created;
+            }
+
+            if (message.ContentUpdate is { Length: > 0 })
+            {
+                yield return new SKChatMessage(role, message.ContentUpdate, new() { { nameof(message.ContentUpdate), message.ContentUpdate } });
+            }
+
+            // This part may change to expose the function name and arguments as new properties of a message
```
stephentoub, line 91: What does this comment mean?  The Azure SDK may change?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatStreamingResult.cs
```diff
@@ -15,41 +16,100 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
...
+            }
+
+            i++;
+
+            // Wait for next choice update...
+            while (!this._isStreamEnded && i >= this._chatUpdates.Count)
             {
-                yield return new SKChatMessage(message);
+                await Task.Delay(50, cancellationToken).ConfigureAwait(false);
             }
```
stephentoub, line 108: This isn't thread-safe nor is it efficient.  We need a better way to actually await for the next message to arrive.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -140,15 +140,159 @@ private protected ClientBase(ILoggerFactory? loggerFactory = null)
...
+            {
+                result.EndOfStream();
+            }
+        }, cancellationToken);
+
+        // Wait for streaming to start
+        while (!streamingStarted)
+        {
+            await Task.Delay(10, cancellationToken).ConfigureAwait(false);
+        }
```
stephentoub, line 194: This code needs to be rewritten to not depend on such delays or concurrent access to types being mutated by the async reading above.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -140,15 +140,159 @@ private protected ClientBase(ILoggerFactory? loggerFactory = null)
...
+                    }
+                }
+            }
+
+            // Updates all results to flag end of stream
+            streamingEnded = true;
+            foreach (var result in results)
+            {
+                result.EndOfStream();
+            }
```
stephentoub, line 187: What happens if this throws an exception? How is that failure communicated back to the user?  Does the system hang because the results were never fully populated?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -140,15 +140,159 @@ private protected ClientBase(ILoggerFactory? loggerFactory = null)
...
+        {
+            yield return results[i];
+
+            i++;
+
+            // Wait until any new choices were created before the end of the stream
+            while (!streamingEnded && i >= results.Count)
+            {
+                await Task.Delay(50, cancellationToken).ConfigureAwait(false);
+            }
```
stephentoub, line 295: Same concerns as earlier.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
```diff
@@ -20,7 +30,23 @@ public class SKChatMessage : ChatMessageBase
...
     {
+        if (message.Name is not null)
+        {
+            this.AdditionalProperties![nameof(message.Name)] = message.Name;
+        }
+
+        if (message.FunctionCall?.Name is not null)
+        {
+            this.AdditionalProperties![FunctionCallNameKey] = message.FunctionCall.Name;
+        }
```
stephentoub, line 41: What's the difference between message.Name and message.FunctionCall.Name?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
```diff
@@ -33,9 +59,32 @@ public SKChatMessage(string role, string content)
...
+    /// <param name="functionCall">Function call definition.</param>
+    /// <param name="arguments">Arguments of the message.</param>
+    public SKChatMessage(string role, string content, FunctionCall? functionCall, Dictionary<string, string> arguments)
+        : base(new AuthorRole(role), content, arguments)
+    {
+        this.FunctionCall = functionCall;
+    }
+
     /// <summary>
     /// Exposes the underlying OpenAI SDK function call chat message representation
```
stephentoub, line 87: Why is a property specific to OpenAI on SKChatMessage?

stephentoub: Oh, nevermind, this is the Azure OpenAI connector's ChatMessageBase-derived type.  Seems like SKChatMessage is the wrong name for the type.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/TextStreamingResult.cs
```diff
@@ -12,29 +13,47 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
...
+
+            i++;
+
+            // Wait for next choice update...
+            while (!this._isStreamEnded && i >= this._choices.Count)
+            {
+                await Task.Delay(50, cancellationToken).ConfigureAwait(false);
+            }
+        }
+    }
```
stephentoub, line 50: Same concerns as earlier.  This producer/consumer implementation should be done differently.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessageBase.cs
```diff
@@ -34,6 +34,6 @@ protected ChatMessageBase(AuthorRole role, string content, IDictionary<string, s
     {
         this.Role = role;
         this.Content = content;
-        this.AdditionalProperties = additionalProperties;
+        this.AdditionalProperties = additionalProperties ?? new Dictionary<string, string>();
```
stephentoub, line 37: This is going to allocate a new dictionary for every chat message even if the dictionary is never needed.  It should instead be done lazily as part of the getter.

docs/decisions/0022-azuresdk-streaming-api-update.md
RogerBarreto: Nope, they were removed, new classes were added representing the new streaming behavior.

stephentoub: Do any other LLM SDKs expose the equivalent of the old behavior?  If the Azure SDK moved away from it, why do we think it's still right for SK consumers?

RogerBarreto:
Halting this PR as pooling is not an optimal approach, new streaming interfaces will come that align better with the current new Azure SDK's APIs.

Creating another one to handle the FunctionCall resolved issue specifically.

RogerBarreto:
Dropping this, as Streaming interfaces will change and be more aligned with the actuall output of the SDK. 

After streaming is merged, a new PR will be created to support the latest changes.

PR #3470, opened 2023-11-13
Python: Make SK compatible with OpenAI 1.0

alexchaomander:
Need to update SK Python to integrate the breaking changes from openai-python 1.0

https://github.com/microsoft/semantic-kernel/issues/3330

This PR replaces the old one that was merging to main https://github.com/microsoft/semantic-kernel/pull/3417


python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
```diff
@@ -3,7 +3,7 @@
 from logging import Logger
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
-import openai
+from openai import AsyncOpenAI
```
moonbox3, line 6: I see that we used to directly import the `openai` module and now that is not being done. In the unit tests, an AttributeError is raised because the attribute `openai` cannot be found in the target module.

I see: AttributeError: <module 'semantic_kernel.connectors.ai.open_ai.services.open_ai_chat_completion' from '/home/runner/work/semantic-kernel/semantic-kernel/python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py'> does not have the attribute 'openai'

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_config_base.py
```diff
@@ -0,0 +1,52 @@
...
+from semantic_kernel.connectors.ai.open_ai.services.open_ai_handler import (
+    OpenAIHandler,
+)
+from semantic_kernel.connectors.ai.open_ai.services.open_ai_model_types import (
+    OpenAIModelTypes,
+)
+
+
+class OpenAIConfigBase(OpenAIHandler):
+    def __init__(
```
moonbox3, line 17: This code looks good. Is it necessary to have any further error handling here? e.g., an invalid API key is provided, or an invalid model is provided?

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
eavanvalkenburg: Indeed tests still need to be fixed

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_config_base.py
eavanvalkenburg: I think those are runtime checks, since you might have proxies in between that do different key formats etc, or like AzureOpenAI that can use AAD auth instead of a key. Instantiating a class is unlikely to do a actual check since that would be a async call and that shouldn't happen in a init. We could decide to do some checks with pydantic, similar to what happens in the AIServiceClientBase, where the model_id is at least 1 character.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
rewrlution: @moonbox3 and @eavanvalkenburg , yes, we will need to change the mock target from `openi` to `AsyncOpenAI` and update the way it is mocked.

Here's an example:
https://github.com/microsoft/semantic-kernel/pull/3486

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
```diff
@@ -2,67 +2,91 @@
...
 
     def __init__(
         self,
         deployment_name: str,
-        endpoint: Optional[str] = None,
+        endpoint: str,
+        api_version: str = "2022-12-01",
         api_key: Optional[str] = None,
-        api_version: str = "2023-03-15-preview",
+        ad_token: Optional[str] = None,
```
moonbox3, line 34: Should this be updated to reflect that it's now called "Microsoft Entra" and not "AAD?" Similar for the comment below.

PR #3474, opened 2023-11-13
.Net: Add IPlugin and overhaul to use it

stephentoub:
Fair warning, this is a massive breaking change (the first of a few to address key concerns about the shape of the solution).

- There is now a first-class concept of a plugin, represented as an ISKPlugin. Plugins and functions now have a many-to-many relationship. ISKFunctions aren't inherently tied to a particular plugin; rather, plugins can contain any number of functions, and a given function can be any in any number of plugins: a plugin is effectively just a named collection of functions. As a result of this, ISKFunction no longer has a PluginName property, and a plugin name is no longer required to construct one. A concrete SKPlugin : ISKPlugin is introduced to make it easy to create plugins that group any set of ISKFunctions.
- SKPluginCollection, which implements ISKPluginCollection, which inherits IReadOnlySKPluginCollection, provides a collection of those IPlugin instances.
- IKernel.Functions becomes IKernel.Plugins. Functions was an IReadOnlyFunctionCollection, but the read-onlyness was a lie as Register functions on the kernel would mutate that collection. Plugins is an IPluginCollection (not read-only), such that new plugins can just be added directly with Add.
- A variety of extensions are provided as helpers on plugins and plugin collections.
- The concept of a global function collection is removed. If you try to find a function without a plugin name, a helper will search all plugins for the first with the specified name. (We can choose to remove this behavior if we don't like it.)
- The ability to create functions and plugins, add functions to plugins, and import plugins into the kernel is now standardized.  A static method provides the Create behavior that's not dependent on a kernel.  Then extensions on Plugin support AddFunction'ing those to simplify the combination of Add+Create. Extensions on kernel support Create'ing functions and plugins as well as importing plugins.


dotnet/src/SemanticKernel.Core/Functions/SKPlugin.cs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// <inheritdoc/>
+    public string Name { get; }
+
+    /// <inheritdoc/>
+    public string Description { get; }
+
+    /// <summary>Gets the number of functions in this plugin.</summary>
+    public int FunctionCount { get; }
```
SergeyMenshykh, line 65: It seems the property was forgotten :) - `public int FunctionCount => this._functions.Count; ` ?

dotnet/src/SemanticKernel.Core/Functions/SKFunctionFromMethod.cs
```diff
@@ -22,65 +22,75 @@
...
 namespace Microsoft.SemanticKernel;
-#pragma warning restore IDE0130
 
 /// <summary>
-/// <see cref="ISKFunction"/> implementation backed by a delegate.
+/// Provides factory methods for creating <see cref="ISKFunction"/> instances backed by a .NET method.
 /// </summary>
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
-internal sealed class NativeFunction : ISKFunction
+public sealed class KernelFunctionFromMethod : ISKFunction
```
SergeyMenshykh, line 32: Is this the final name?

dotnet/src/SemanticKernel.Core/Functions/SKPlugin.cs
stephentoub: Oops! Thanks.

dotnet/src/SemanticKernel.Core/Functions/SKFunctionFromMethod.cs
stephentoub: @matthewbolanos, @markwallace-microsoft, what would you like to do here?  This is what we'd discussed, but then Mark suggested we instead consolidate down to having a single factory class per assembly, e.g. `SKFunction.CreateFunctionFromMethod`.  What would you prefer?

markwallace-microsoft: We want to minimise the dependencies of SK.Core i.e. it shouldn't know anything about any YAML packages. 

So we would have:

**SK.Core**
```csharp
SKFunction.CreateFromPrompt(string prompt)
SKFunction.CreateFromPrompt(string prompt, PromptTemplateConfig promptTemplateConfig) // Support for current format
SKFunction.CreateFromPrompt(PromptModel promptModel)
SKFunction.CreateFromPromptJson(string json) // Deserialises json to an instance of PromptModel
SKFunction.CreateFromPromptJsonResource(string resourceName)
```

**SK.Functions.Yaml**
```csharp
SKFunctionYaml.CreateFromPromptYaml(string yaml) // Deserialises yaml to an instance of PromptModel
SKFunctionYaml.CreateFromPromptYamlResource(string resourceName)
```

**SK.Functions.Markdown**
```csharp
SKFunctionMarkdown.CreateFromPromptMarkdown(string markdown) // Creates PromptModel from code blocks in markdown
SKFunctionMarkdown.CreateFromPromptMarkdownResource(string resourceName)
```

If someone wants to support a new serialised format which could be json or xml they would do

**Contoso.Functions.Foo**
```csharp
SKFunctionFoo.CreateFromPromptFooJson(string json) // Deserialises custom json to an instance of PromptModel
SKFunctionFoo.CreateFromPromptFooXml(string xml) // Deserialises custom xml to an instance of PromptModel
```

markwallace-microsoft (approved):
We need to finalise naming but this can be done later

dotnet/src/SemanticKernel.Core/Functions/SKFunctionFromMethod.cs
markwallace-microsoft: Added this to our chat and OK to merge and rename if needed later

SergeyMenshykh (approved):
Approving this PR to proceed with the next one - https://github.com/microsoft/semantic-kernel/pull/3489 All outstanding issues in the PR, if found, can be addressed in a separate PRs against the V1.0 branch.

Issue #3471, opened 2023-11-13
Python: Prompt Template Loading not loading template due to unknown character on Windows.


LukasDauterman:
**Describe the bug**

I have several prompt templates with very noisy natural language in them, and semantic kernel fails to load the template. 

I give me an error saying " UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 21425: character maps to <undefined>"

The trace points to line 793 in kernel.py:

`            
            # Load Prompt Template

            with open(prompt_path, "r") as prompt_file:

                template = PromptTemplate(

                    prompt_file.read(), self.prompt_template_engine, config

                )
`

This looks like its caused by reading the prompt file using the default windows charact mapping functionality. 

**My Fix**
I can resolve the issue by modifying the open statements in kernel.py on lines 792 and 788 by adding a  encoding='utf-8' argument, and it runs as expected.

**To Reproduce**

1. Create a template with utf-8 characters that are not in default windows open file parser character mapping.

2. Try and load it using import_semantic_skill_from_directory.

**Expected behavior**

I expect the template to load given a potentially wide range of input characters.

**Screenshots**
![image](https://github.com/microsoft/semantic-kernel/assets/3969402/52ec9659-2291-4029-9106-33db126f03c5)

**Platform**
 - OS: Windows 11
 - IDE: Visual studio Code
 - Language: Python
 - Source: pip package

**Additional context**

matthewbolanos:
Thank you for identifying the problem and the proposed fix. Would you be able to add your fix as a PR so that we can get it merged in?

LukasDauterman:
Yes I can, let me work on getting it in.

alliscode:
Closing issue as it's no longer relevant.

afrancoc2000:
This is also happening to me loading a plugin, the change in this case is in ln 334 of kernel_plugin.py, it's also fixed by setting the encoding to utf-8. @matthewbolanos I can add the fix, should I also create a PR?

sakesun:
This problem still exists

sakesun:
On Windows, set environment variable "PYTHONUTF8" = "1" to workaround this problem.

PR #3477, opened 2023-11-13
Python: Windows Text Parsing of Prompt Templates kernel.py

LukasDauterman:
fixing issue with reading

### Motivation and Context

  By default on windows, the character mapping for reading the prompt templates does not cover a wide range of characters, causing parsing failures when trying to read templates on windows when characters that utf-8 encodings handle.

By stetting it so that the encoding is set to utf-8 it makes it so that by default the package can read a larger set of characters without throwing an error.

Allows for a wide range of templates and kinds of text are in the template without throwing an error when using semantic kernel python on windows.

Usage of Semantic Kernel on windows machines that covers a large number of potential characters in the template. Especially important for few-shot examples when the input text is natural language. 

https://github.com/microsoft/semantic-kernel/issues/3471

### Description

Change the opening of the template and config file to by default use utf-8 encoding by default 

### Contribution Checklist

- [Yes] The code builds clean without any errors or warnings
- [ Small change, I can't share internal examples, but can find something simple examples if needed.] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [Yes ] All unit tests pass, and I have added new tests where possible
- [Should not have broken anyone] I didn't break anyone :smile:

If you need me to make changes or revisions let me know.

### DIFF
--- a/python/semantic_kernel/kernel.py
+++ b/python/semantic_kernel/kernel.py
@@ -863,11 +863,11 @@ class Kernel:
 
             config = PromptTemplateConfig()
             config_path = os.path.join(directory, CONFIG_FILE)
-            with open(config_path, "r") as config_file:
+            with open(config_path, "r", encoding='utf-8') as config_file:
                 config = config.from_json(config_file.read())
 
             # Load Prompt Template
-            with open(prompt_path, "r") as prompt_file:
+            with open(prompt_path, "r", encoding='utf-8') as prompt_file:
                 template = PromptTemplate(
                     prompt_file.read(), self.prompt_template_engine, config
                 )


moonbox3:
Other locales may also require a different encoding config. For example, see the following PR: #4409. We should figure out how the encoding can be specified once in some type of config, and then made available for use throughout the project.

moonbox3:
Hi @LukasDauterman, thank you for your contribution. Is this work still required? Per my previous comment, there looks to be something related in #4409. If this work is still required, please re-open the PR. Thank you.

PR #3489, opened 2023-11-14
.Net: Kernel as a state/property bag

SergeyMenshykh:
This PR is the first in a series of follow-up PRs aimed at making the SK Kernel a central component of the SK API
1. The IKernel interface is now obsolete and no longer used by the Kernel class to avoid breaking changes when adding new members. This makes it easier to extend Kernel without breaking classes that inherit from it.
2. The RunAsync method has been changed from an instance method to an extension method, as the primary purpose of the Kernel is no longer to execute functions. Now, RunAsync is just one of several operations, modeled as extension methods, that can be performed on the Kernel.


dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
```diff
@@ -26,7 +26,7 @@ namespace Microsoft.SemanticKernel;
 /// <summary>
 /// Provides extension methods for the <see cref="KernelBuilder"/> class to configure OpenAI and AzureOpenAI connectors.
 /// </summary>
-public static class OpenAIKernelBuilderExtensions
+public static class OpenAKernelBuilderExtensions
```
markwallace-microsoft, line 29: I don't think you meant to make this change

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelExtensions.cs
```diff
@@ -5,14 +5,14 @@
...
+// ReSharper disable once CheckNamespace - Using the namespace of Kernel
 namespace Microsoft.SemanticKernel;
 #pragma warning restore IDE0130
 
 /// <summary>
-/// Class for extension methods for <see cref="IKernel"/> using OpenAI request settings.
+/// Class for extension methods for <see cref="Kernel"/> using OpenAI request settings.
 /// </summary>
-public static class OpenAIKernelExtensions
+public static class OpenAKernelExtensions
```
markwallace-microsoft, line 15: Same here

markwallace-microsoft (commented):
Couple of search and replace issues

dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIKernelBuilderExtensions.cs
SergeyMenshykh: Yep, right, it was not intentional change. Thanks.

dotnet/src/Planners/Planners.Core.UnitTests/Action/ActionPlannerTests.cs
```diff
@@ -158,37 +165,39 @@ public async Task ListOfFunctionsExcludesExcludedFunctionsAsync()
...
     {
-        plugins ??= new SKPluginCollection();
-        var functionRunner = new Mock<IFunctionRunner>();
-        var serviceProvider = new Mock<IAIServiceProvider>();
-        var serviceSelector = new Mock<IAIServiceSelector>();
-        var kernel = new Mock<IKernel>();
+        if (plugins is null)
+        {
+            plugins = new SKPluginCollection();
+        }
```
stephentoub, line 173: This can just be:
```C#
plugin ??= new SKPluginCollection();
```

dotnet/src/Planners/Planners.Core.UnitTests/Sequential/SequentialPlanParserTests.cs
```diff
@@ -358,4 +346,34 @@ public void CanCreatePlanWithIgnoredNodes(string goalText, string planText)
         Assert.Equal("MockPlugin", plan.Steps[1].PluginName);
         Assert.Equal("Echo", plan.Steps[1].Name);
     }
+
+    private Kernel CreateKernel(string testPlanString, SKPluginCollection? plugins = null)
+    {
+        if (plugins is null)
+        {
+            plugins = new SKPluginCollection();
+        }
```
stephentoub, line 355:
```suggestion
        plugins ??= new SKPluginCollection();
```

dotnet/src/Planners/Planners.Core.UnitTests/Sequential/SequentialPlannerTests.cs
```diff
@@ -21,174 +19,122 @@ public sealed class SequentialPlannerTests
...
-    // Method to create Mock<IKernel> objects
-    private static Mock<IKernel> CreateMockKernel()
+    private Kernel CreateKernel(string testPlanString, SKPluginCollection? plugins = null)
     {
-        var kernel = new Mock<IKernel>();
-        kernel.Setup(x => x.LoggerFactory).Returns(NullLoggerFactory.Instance);
+        if (plugins is null)
+        {
+            plugins = new SKPluginCollection();
+        }
```
stephentoub, line 96:
```suggestion
        plugins ??= new SKPluginCollection();
```

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
```diff
@@ -44,18 +43,6 @@ public interface IKernel
     [Obsolete("This will be removed in a future release. Use RegisterPlugin instead.")]
     ISKFunction RegisterCustomFunction(ISKFunction customFunction);
```
stephentoub, line 44: I thought I deleted this.  Given all the breaking changes we're making, probably makes sense to delete all the Obsolete stuff that's in the repo. @markwallace-microsoft? That can obviously be a separate PR.

dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -87,6 +93,7 @@ public sealed class Kernel : IKernel, IDisposable
     /// <param name="httpHandlerFactory">HTTP handler factory</param>
     /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
     /// <param name="serviceSelector">AI Service selector</param>
+    [Obsolete("This constructor is obsolete and will be removed in one of the next version of SK. Please use one of the available overload.")]
```
stephentoub, line 96: Related to my previous comment, should we instead just delete these?

dotnet/src/SemanticKernel.Core/Kernel.cs
```diff
@@ -107,70 +114,53 @@ public sealed class Kernel : IKernel, IDisposable
...
+        ILoggerFactory? loggerFactory = null)
     {
-        Verify.NotNull(customFunction);
-
-        this.Plugins.Add(new SKPlugin($"plugin{Guid.NewGuid():N}", new[] { customFunction }));
+        this.LoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
+        this.HttpHandlerFactory = httpHandlerFactory ?? NullHttpHandlerFactory.Instance;
+        this._aiServiceProvider = aiServiceProvider;
+        this.Plugins = plugins ?? new SKPluginCollection();
+        this._aiServiceSelector = serviceSelector ?? new OrderedIAIServiceSelector();
```
stephentoub, line 136: We could make these cheaper by creating them lazily; that way they won't be instantiated unless used.  But if we expect not using them to be rare, it's fine and simpler to create them here in the ctor.

stephentoub (approved):

dotnet/src/SemanticKernel.Abstractions/IKernel.cs
markwallace-microsoft: Agreed, let's clean out obsolete stuff as we go

markwallace-microsoft (approved):

dotnet/src/Planners/Planners.Core.UnitTests/Action/ActionPlannerTests.cs
SergeyMenshykh: Yep, fixed.

dotnet/src/SemanticKernel.Core/Kernel.cs
SergeyMenshykh: A good idea that I would suggest deferring for a later stage after we have completed cloning.

PR #3493, opened 2023-11-14
.Net: Move static factory methods to SKFunction / SKPlugin

stephentoub:
Moves the recently added CreateFunctionFromXx and CreatePluginFromXx methods to SKFunction and SKPlugin and renames them as discussed.

No implementations changed here; the code just moved around.


stephentoub:
@markwallace-microsoft, @SergeyMenshykh, please take a look when you have a moment.

dotnet/src/Experimental/Orchestration.Flow.UnitTests/FlowExtensionsTests.cs
```diff
@@ -8,6 +8,8 @@
 using Microsoft.SemanticKernel.Experimental.Orchestration.Abstractions;
 using Xunit;
 
+#pragma warning disable SKEXP0001
```
SergeyMenshykh, line 11: @stephentoub what are this warnings are about? and why are we disabling them?

SergeyMenshykh (approved):

dotnet/src/Experimental/Orchestration.Flow.UnitTests/FlowExtensionsTests.cs
stephentoub: The experimental types being used here are marked [Experimental]. This triggers a warning to let folks know they're depending on something that might change. The tests need to suppress that warning.

SergeyMenshykh: Got it, thanks.

PR #3502, opened 2023-11-15
.Net: Handlebars planner complex types

teresaqhoang:
### Motivation and Context

Support for complex types in Handlebars Planner

### Description

Native Plugin with Complex Types
image

Remote Plugin with Complex Types
image

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Program.cs
```diff
@@ -21,7 +21,7 @@ public static async Task Main(string[] args)
         using CancellationTokenSource cancellationTokenSource = new();
         CancellationToken cancelToken = cancellationTokenSource.ConsoleCancellationToken();
 
-        string? defaultFilter = null; // Modify to filter examples
+        string? defaultFilter = "Example66"; // Modify to filter examples
```
markwallace-microsoft, line 24: Revert this change

teresaqhoang: Done!

dotnet/samples/KernelSyntaxExamples/Example66_HandlebarsPlannerWithComplexTypes.cs
```diff
@@ -0,0 +1,232 @@
...
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Functions.OpenAPI.Extensions;
+using Microsoft.SemanticKernel.Planners.Handlebars;
+using RepoUtils;
+
+/**
+ * This example shows how to use the Handlebars sequential planner with complex types.
+ */
+public static class Example66_HandlebarsPlannerWithComplexTypes
```
teresaqhoang, line 20: Currently working to merge this with Example65, had been waiting on the other PR to merge first

dotnet/src/Planners/Planners.Core/Handlebars/skPrompt.handlebars
```diff
@@ -2,19 +2,23 @@
 Explain how to achieve the user's goal with the available helpers with a Handlebars template.
```
markwallace-microsoft, line 2: The naming of this file can be improved, instead of `skPrompt` use a name that describes the purpose of the prompt. We won't to avoid having multiple files with the same name.

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlanner.cs
```diff
@@ -98,11 +101,55 @@ public async Task<HandlebarsPlan> CreatePlanAsync(string goal, CancellationToken
...
+        }
+
+        return functionsView;
+    }
+
+    // Extract any complex schemas for isolated render in prompt template
+    private ParameterView SetComplexTypeDefinition(ParameterView parameter)
+    {
+        // TODO (@teresaqhoang): handle case when schema and ParameterType can exist i.e., when ParameterType = RestApiResponse
+        if (parameter.ParameterType != null)
```
markwallace-microsoft, line 136: Use `is not`

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
```diff
@@ -69,9 +89,15 @@ private static async Task RunSampleAsync(string goal, params string[] pluginDire
         }
 
         // The gpt-35-turbo model does not handle loops well in the plans.
-        var allowLoopsInPlan = chatDeploymentName.Contains("gpt-35-turbo", StringComparison.OrdinalIgnoreCase) ? false : true;
+        var allowLoopsInPlan = !chatDeploymentName.Contains("gpt-35-turbo", StringComparison.OrdinalIgnoreCase);
```
TaoChenOSU, line 92: nit: 
1. Be nice if this explicitly mentions what models to use if loops are desired. Will gpt-4 work better, or it needs to be gpt4-turbo?
2. What is expected to happen if a goal requires loops but the planner is using 35 turbo?

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlanner.cs
```diff
@@ -54,55 +58,108 @@ public HandlebarsPlanner(IKernel kernel, HandlebarsPlannerConfig? config = defau
...
 
     private List<FunctionView> GetAvailableFunctionsManual(CancellationToken cancellationToken = default)
     {
-        return this._kernel.Functions.GetFunctionViews()
+        var availableFunctions = this._kernel.Functions.GetFunctionViews()
             .Where(s => !this._config.ExcludedPlugins.Contains(s.PluginName, StringComparer.OrdinalIgnoreCase)
                 && !this._config.ExcludedFunctions.Contains(s.Name, StringComparer.OrdinalIgnoreCase)
                 && !s.Name.Contains("Planner_Excluded"))
-            .ToList();
+                .ToList();
```
TaoChenOSU, line 111: nit: is this extra tab necessary?

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlanner.cs
```diff
@@ -54,55 +58,108 @@ public HandlebarsPlanner(IKernel kernel, HandlebarsPlannerConfig? config = defau
...
+        if (parameter.ParameterType is not null)
+        {
+            // Async return type - need to extract the actual return type and override ParameterType property
+            var type = parameter.ParameterType;
+            if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Task<>))
+            {
+                parameter = parameter with { ParameterType = type.GenericTypeArguments[0] }; // Actual Return Type
+            }
+
+            this._parametersTypeView.UnionWith(parameter.ParameterType.ToHandlebarsParameterTypeView());
```
TaoChenOSU, line 148: With this class member, it looks like the planner object of this class is expected to be discarded after a plan is created. Is this correct?

dotnet/samples/KernelSyntaxExamples/Example66_HandlebarsPlannerWithComplexTypes.cs
teresaqhoang: Merged

dotnet/src/Planners/Planners.Core/Handlebars/skPrompt.handlebars
teresaqhoang: Changed to `CreatePlanPrompt.handlebars`

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlanner.cs
teresaqhoang: Changes in all instances

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlanner.cs
teresaqhoang: nope, fixed now!

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
teresaqhoang: It still works w 35-turbo just not as well

Changed comments to:
```
// Use gpt-4 or newer models if you want to test with loops. 
// Older models like gpt-35-turbo are able handle loops but is more prone to syntax errors.
```

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlanner.cs
teresaqhoang: Yep! It's only needed for prompt rendering

teresaqhoang: Synced offline -- removed complex type definitions as class properties. These views now only exist in request scope of CreatePlanAsync

TaoChenOSU (approved):

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
```diff
@@ -3,42 +3,51 @@
...
-        await RunDictionarySampleAsync();
+        await RunDictionaryWithBasicTypesSampleAsync();
         await RunPoetrySampleAsync();
         await RunBookSampleAsync();
+
+        // Using Complex Types as inputs and outputs
+        await RunLocalDictionaryWithComplexTypesSampleAsync(shouldPrintPrompt);
+
+        // TODO (@teresaqhoang): Enable this sample once complex types are supported for remote plugins return type.
+        // await RunRemoteDictionaryWithComplexTypesSampleAsync();
```
markwallace-microsoft, line 42: Can we leave this out until it's working?

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
```diff
@@ -59,9 +68,20 @@ private static async Task RunSampleAsync(string goal, params string[] pluginDire
...
 
-        if (pluginDirectoryNames[0] == DictionaryPlugin.PluginName)
+        if (pluginDirectoryNames[0] == BasicDictionaryPlugin.PluginName)
+        {
+            kernel.ImportPluginFromObject(new BasicDictionaryPlugin(), BasicDictionaryPlugin.PluginName);
+        }
+        else if (pluginDirectoryNames[0] == ComplexDictionaryPlugin.PluginName)
+        {
+            kernel.ImportPluginFromObject(new ComplexDictionaryPlugin(), ComplexDictionaryPlugin.PluginName);
+        }
```
markwallace-microsoft, line 78: The naming of the two dictionary plugins isn't super obvious, something like the following would be clearer

1. `StringParamsDictionaryPlugin`
1. `ComplexParamsDictionaryPlugin`

dotnet/src/Planners/Planners.Core/Handlebars/CreatePlanPrompt.handlebars
```diff
@@ -127,7 +182,7 @@ IMPORTANT: You can only use the helpers that are listed above. Do not use any ot
 - Do not make up values. Use the helpers to generate the data you need or extract it from the goal.
 - Keep data well-defined. Each variable should have a unique name. Create and assign each variable only once.
 - Be extremely careful about types. For example, if you pass an array to a helper that expects a number, the template will error out.
-- Avoid using loops. Try a solution without loops before you deploy a loop.
+{{#if allowLoops}}- Avoid using loops. Try a solution without before you deploy a loop.{{/if}}
 - There is no need to check your results in the template.
 - Do not nest sub-expressions or helpers because it will cause the template to error out.
 - Each step should contain only one helper call.
```
markwallace-microsoft, line 188: We definitely need to figure out how to make this more maintainable

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/ParameterViewExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        // Add nested complex types
+        foreach (var property in properties)
+        {
+            parameterTypes.UnionWith(property.PropertyType.ToHandlebarsParameterTypeView());
+        }
+    }
+
+    public static bool isPrimitiveOrStringType(string type) =>
+        type == "string" || type == "number" || type == "integer" || type == "boolean" || type == "null";
```
markwallace-microsoft, line 80: Looking below `null` is an object and not a string or primitive, so is the above correct?

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterView.cs
```diff
@@ -0,0 +1,98 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel.Planners.Handlebars.Models;
+
+internal class HandlebarsParameterTypeView
```
markwallace-microsoft, line 8: We are changing our naming from `View` to `Metadata`, see https://github.com/microsoft/semantic-kernel/pull/3545

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/ParameterViewExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Planners.Handlebars.Models;
+
+namespace Microsoft.SemanticKernel.Planners.Handlebars.Extensions;
+
+internal static class ParameterViewExtensions
```
markwallace-microsoft, line 13: Changing naming from `View` to `Metadata` see https://github.com/microsoft/semantic-kernel/pull/3545

markwallace-microsoft (commented):
There is a rename coming which this PR should adopt

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/ParameterViewExtensions.cs
teresaqhoang: Refactored everything from `View` -> `Metadata`

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterView.cs
teresaqhoang: Refactored everything from `View` -> `Metadata`

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/ParameterViewExtensions.cs
teresaqhoang: You're right, I moved the "null" type check from here to the conditional below

dotnet/src/Planners/Planners.Core/Handlebars/CreatePlanPrompt.handlebars
teresaqhoang: Yeah, I'm thinking we can split it with partials, and it'll be more digestible.

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
teresaqhoang: Done

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
teresaqhoang: Done

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterTypeMetadata.cs
```diff
@@ -0,0 +1,115 @@
...
+    public override bool Equals(object obj)
+    {
+        // Check if the obj is null or of a different type
+        if (obj == null || obj.GetType() != this.GetType())
+        {
+            return false;
+        }
+
+        // Cast the obj to HandlebarsParameterTypeMetadata
+        var other = (HandlebarsParameterTypeMetadata)obj;
```
SergeyMenshykh, line 34:
```suggestion
        if (obj is not HandlebarsParameterTypeMetadata other)
        {
            return false;
        }
```

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterTypeMetadata.cs
```diff
@@ -0,0 +1,115 @@
...
+        for (int i = 0; i < list1.Count; i++)
+        {
+            if (!list1[i].Name.Equals(list2[i].Name, System.StringComparison.Ordinal) || !list1[i].ParameterType!.Equals(list2[i].ParameterType))
+            {
+                return false;
+            }
+        }
+
+        // If all elements are equal, return true
+        return true;
```
SergeyMenshykh, line 65: The code can be replaced by `return this.Properties.SequenceEqual(other.Properties);` if SKParameterMetadata class implements Equals method and if sequential comparison is what we need here.

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterTypeMetadata.cs
```diff
@@ -0,0 +1,115 @@
...
+    // A helper method to compute the hash code of a list of SKParameterMetadata
+    private static int GetPropertiesHashCode(List<SKParameterMetadata> list)
+    {
+        // Use a prime number to combine the hash codes of the elements
+        int hash = 19;
+
+        // Use the default SKParameterMetadata hash code for each element
+        foreach (var item in list)
+        {
+            hash = (hash * 37) + GetPropertyHashCode(item);
```
SergeyMenshykh, line 95: No need in this method if SKParameterMetadata implements GetHashCode method. @stephentoub Is there an easier way in .NET to perform hash calculations?

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/SKParameterMetadataExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+namespace Microsoft.SemanticKernel.Planning.Handlebars.Extensions;
+#pragma warning restore IDE0130
+
+internal static class SKParameterMetadataExtensions
+{
+    /// <summary>
+    /// Checks if type is primitive or string
+    /// </summary>
+    public static bool isPrimitiveOrStringType(Type type) => type.IsPrimitive || type == typeof(string);
```
SergeyMenshykh, line 20:
```suggestion
    public static bool IsPrimitiveOrStringType(Type type) => type.IsPrimitive || type == typeof(string);
```

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/SKParameterMetadataExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+    private static void AddNestedComplexTypes(this HashSet<HandlebarsParameterTypeMetadata> parameterTypes, PropertyInfo[] properties)
+    {
+        // Add nested complex types
+        foreach (var property in properties)
+        {
+            parameterTypes.UnionWith(property.PropertyType.ToHandlebarsParameterTypeMetadata());
+        }
+    }
+
+    public static bool isPrimitiveOrStringType(string type) =>
```
SergeyMenshykh, line 81:
```suggestion
    public static bool IsPrimitiveOrStringType(string type) =>
```

SergeyMenshykh, line 81: Do we need this overload of the `isPrimitiveOrStringType` method taking into account we already declared one above? 
Can we keep the two methods next to each other?

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlan.cs
```diff
@@ -14,15 +16,23 @@ public sealed class HandlebarsPlan
...
     /// <summary>
     /// Initializes a new instance of the <see cref="HandlebarsPlan"/> class.
     /// </summary>
-    /// <param name="kernel">The kernel.</param>
-    /// <param name="template">The Handlebars template.</param>
-    public HandlebarsPlan(Kernel kernel, string template)
+    /// <param name="kernel">Kernel instance.</param>
+    /// <param name="template">A Handlebars template representing the plan.</param>
+    /// <param name="createPlanPromptTemplate">Prompt template used to generate the plan.</param>
+    public HandlebarsPlan(Kernel kernel, string template, string createPlanPromptTemplate)
```
SergeyMenshykh, line 31: Consider adding more details about `template` and `createPlanPromptTemplate` parameters to clearly describe the difference between the two.

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsTemplateEngineExtensions.cs
```diff
@@ -66,11 +69,11 @@ internal sealed class HandlebarsTemplateEngineExtensions
...
         IHandlebars handlebarsInstance,
-        SKFunctionMetadata functionView,
+        SKFunctionMetadata functionMetadata,
         Dictionary<string, object?> variables,
         CancellationToken cancellationToken = default)
     {
-        string fullyResolvedFunctionName = functionView.PluginName + ReservedNameDelimiter + functionView.Name;
+        string fullyResolvedFunctionName = functionMetadata.PluginName + ReservedNameDelimiter + functionMetadata.Name;
 
         handlebarsInstance.RegisterHelper(fullyResolvedFunctionName, (in HelperOptions options, in Context context, in Arguments arguments) =>
```
SergeyMenshykh, line 78: Consider splitting the `RegisterHelper` callback in a few smaller private methods to it readable.

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterTypeMetadata.cs
teresaqhoang: We do need sequential property comparison but only need to check Name and PropertyType from the SKParameterMetadata object, since those are the only 2 properties we use in the HandlebarsParameterTypeMetadata object, 
image

Also, SKParameterMetadata doesn't override the Equals method.

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/SKParameterMetadataExtensions.cs
teresaqhoang: Yes one takes a `Type` parameter and the other takes a stringified type (type serialized in JSON Schema). Moved them to be next to each other

image

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterTypeMetadata.cs
stephentoub: > Also, SKParameterMetadata doesn't override the Equals method.

We could make it do so if desirable.

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsPlan.cs
teresaqhoang: I added some comments as well as changed the parameter name in the constructor:
image

I didn't want to change the class `_template` property name because ``HandlebarsPlan._template` makes sense

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsTemplateEngineExtensions.cs
teresaqhoang: Done, split into `ProcessHashArguments`, `ProcessPositionalArguments`, `InitializeContextVariables`, and `InvokeSKFunction`

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterTypeMetadata.cs
teresaqhoang: This scenario is pretty specific to the Planners and how we render complex types for the function manual, but I think it could be beneficial to centralize this in the SKParameterMetadata object anyway (for future planners or data checks when sharing variables across multiple functions of the same plugin)

dotnet/src/Planners/Planners.Core/Handlebars/Models/HandlebarsParameterTypeMetadata.cs
stephentoub: System.HashCode is the recommended way to combine multiple hash codes into a single one, e.g.
```C#
public override int GetHashCode()
{
    HashCode hc = default;
    hc.Add(this.Name);
    hc.Add(this.IsComplex);
    foreach (var item in this.Properties)
    {
        hc.Add(item?.Name);
        hc.Add(item?.ParameterType);
    }
    return hc.ToHashCode();
}
```
That said, also keep in mind that while equality does need to factor in all relevant data, hash code doesn't have to. It just needs to factor in enough to provide a reasonably good-enough distribution.  You could, for example, just have the implementation be:
```C#
public override int GetHashCode() => this.Name?.GetHashCode() ?? 0;
```
and depending on the circumstances, that might be "good enough".

teresaqhoang: Refactored to use HashCode

SergeyMenshykh (approved):

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/HandlebarsPlannerExtensions.cs
```diff
@@ -4,7 +4,9 @@
 using System.IO;
 using System.Reflection;
 
-namespace Microsoft.SemanticKernel.Planning.Handlebars;
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+namespace Microsoft.SemanticKernel.Planning.Handlebars.Extensions;
+#pragma warning restore IDE0130
```
markwallace-microsoft, line 9: I'd expect the extensions to be in th same namespace as the class they are extending

Nit: there is no need to do warning restore here

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/SKParameterMetadataExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Planning.Handlebars.Models;
+
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+namespace Microsoft.SemanticKernel.Planning.Handlebars.Extensions;
```
markwallace-microsoft, line 12: Is this the best namespace?

markwallace-microsoft (approved):
Couple of minor things.

Next PR we should move this to it's own package.

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/HandlebarsPlannerExtensions.cs
teresaqhoang: This was an intentional change by Stephen to consolidate everything into the `Planning` namespace: https://github.com/microsoft/semantic-kernel/pull/3564

And I see this error if I remove the warning ignore:
image

stephentoub: But why is `.Extensions` needed?

teresaqhoang: Ah I see what you mean, I removed it

dotnet/src/Planners/Planners.Core/Handlebars/Extensions/SKParameterMetadataExtensions.cs
teresaqhoang: Removed `.Extensions`, just `namespace Microsoft.SemanticKernel.Planning.Handlebars` now

markwallace-microsoft (approved):

TaoChenOSU (approved):

TaoChenOSU (approved):

PR #3510, opened 2023-11-15
.Net: Fix nullable reference type warnings

stephentoub:
### Motivation and Context

I temporarily retargeted all of the projects to net8.0 and fixed all of the NRT warnings, some of which were noise, and some of which were legit.

### Contribution Checklist


dotnet/src/Connectors/Connectors.Memory.Weaviate/WeaviateMemoryStore.cs
```diff
@@ -563,11 +563,11 @@ private static MemoryRecordMetadata ToMetadata(WeaviateObject weaviateObject)
 
         return new(
             false,
-            string.Empty,
-            weaviateObject.Properties["sk_id"].ToString(),
-            weaviateObject.Properties["sk_description"].ToString(),
-            weaviateObject.Properties["sk_text"].ToString(),
-            weaviateObject.Properties["sk_additional_metadata"].ToString()
+            "",
```
RogerBarreto, line 566: nit. IMHO I like to keep `string.Empty`, as it is easy to spot and gives clarity of the intention. For me it also make me think twice as more when adding those.

RogerBarreto (approved):

markwallace-microsoft (approved):
Nice, we should do this again once we get to the v1.0 release candidate

stephentoub:
> Nice, we should do this again once we get to the v1.0 release candidate

At some point, we'll likely also want to multitarget (e.g. add a net8.0 target in addition to the netstandard2.0 target), at which point the validation will always be there.

RogerBarreto (approved):

PR #3516, opened 2023-11-15
.Net: Kernel class is moved abstractions

SergeyMenshykh:
### Motivation and Context
The primary goal of the PR is to move the Kernel class to the `SemanticKernel.Abstractions` project. This allows it to be added as a parameter to the `ISKFunction.InvokeAsync` method and to other methods of classes/interfaces residing in the abstraction project that require the Kernel as a parameter.

### Description
1. The `Kernel` class has been moved to the `SemanticKernel.Abstractions` project.  
2. The `SKPluginCollection` class has also been moved to the Abstractions project. The Kernel uses it by default internally if no external collection is provided.  
3. The `OrderedIAIServiceSelector` class has been moved as well. The Kernel uses it by default internally if no external one is provided.  
4. The `FunctionRunner` class and its interface have been removed. 
5. The `SKContext` class has been refactored to use the Kernel as the function runner.


markwallace-microsoft (approved):

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/BasicPromptTemplateTests.cs
```diff
@@ -379,32 +367,17 @@ private static MethodInfo Method(Delegate method)
...
-                var context = new SKContext(this._functionRunner.Object, this._serviceProvider.Object, this._serviceSelector.Object, variables);
-                var function = functions.First(f => f.Name == functionName);
-
-                return await function.InvokeAsync(context, null, cancellationToken);
-            });
-    }
+        if (functions?.Length > 0)
+        {
+            plugins.Add(new SKPlugin("plugin", functions));
+        }
```
stephentoub, line 377: I love how much simpler the tests get when the mocking is replaced by the use of the simple types.

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/Blocks/CodeBlockTests.cs
```diff
@@ -19,51 +18,40 @@ namespace SemanticKernel.Extensions.UnitTests.TemplateEngine.Prompt.Blocks;
...
+        var context = new SKContext(this._kernel, this._serviceProvider.Object, this._serviceSelector.Object, plugins: this._kernel.Plugins);
+
         var function = new Mock<ISKFunction>();
+        function.Setup(x => x.Name).Returns("function");
         function
             .Setup(x => x.InvokeAsync(
                 It.IsAny<SKContext>(),
                 It.IsAny<AIRequestSettings?>(),
                 It.IsAny<CancellationToken>()))
             .Throws(new RuntimeWrappedException("error"));
```
stephentoub, line 50: Presumably all of this mocking could also be replaced by an `SKFunction.FromMethod(() => throw new RuntimeWrappedException("error"))` or something similar?  Just checking; I realize the purpose of this PR isn't to clean up such tests, of which I expect there are a lot.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -49,7 +49,7 @@ public CultureInfo Culture
     /// <summary>
     /// Executes functions using the current resources loaded in the context
     /// </summary>
-    public IFunctionRunner Runner { get; }
+    public Kernel Runner { get; }
```
stephentoub, line 52: I assume we're not renaming this because we expect the whole class to disappear soon.

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -553,6 +554,29 @@ public async Task ItCancelsPipelineFromFunctionInvokedEventsAsync(int numberOfFu
...
+    {
+        //Arrange
+        var serviceProvider = new Mock<IAIServiceProvider>();
+        var serviceSelector = new Mock<IAIServiceSelector>();
+
+        var context = new SKContext(new Kernel(serviceProvider.Object), serviceProvider.Object, serviceSelector.Object, new ContextVariables());
+
+        var function = new Mock<ISKFunction>();
+        function.Setup(f => f.Name).Returns("function");
+        function.Setup(f => f.InvokeAsync(It.IsAny<SKContext>(), It.IsAny<AIRequestSettings>(), It.IsAny<CancellationToken>())).ReturnsAsync(new FunctionResult("function", context));
```
stephentoub, line 568: This could easily just be an SKFunction.FromMethod, right?

stephentoub (approved):

RogerBarreto (approved):
FunctionRunner is gone, are we exposing all the kernel back I suppose...

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/Blocks/CodeBlockTests.cs
SergeyMenshykh: Yep, sure, it can be done this way. I did not pay attention on this when refactored the tests. I will do it in the following PRs when changing unit tests.

By the way, when I tried to use the real function - `KernelFunctionFromMethod.Create((Action)(() => throw new RuntimeWrappedException("error")), "function");`, I received a `TargetInvocationException` exception that had a `RuntimeWrappedException` as an inner exception, instead of a `RuntimeWrappedException` itself. I wonder what the right behavior here would be. Should we keep throwing the `TargetInvocationException`, or should we rethrow its inner exception?

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
SergeyMenshykh: Yep, right - no need to rename since it'll be removed soon.

dotnet/src/Extensions/Extensions.UnitTests/TemplateEngine/Prompt/Blocks/CodeBlockTests.cs
stephentoub: We should use ExceptionDispatchInfo to throw the inner. On core there's an option to pass to invoke to tell it to not wrap the exception, but that's not available in netstandard2.0.

Issue #2443, opened 2023-08-14
.Net: Microsoft.SemanticKernel.AI.ChatCompletion.ChatHistory is not deserializable


alexminza:
**Describe the bug**

`ChatHistory` is not deserializable due to `ChatMessageBase` being an [abstract class](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessageBase.cs):

```
[2023-08-14T17:33:25.624Z] fail: Microsoft.SemanticKernel.IKernel[0]
[2023-08-14T17:33:25.624Z]       Function call fail during pipeline step 0: RecommendProductPlugin.RecommendProduct. Error: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Microsoft.SemanticKernel.AI.ChatCompletion.ChatMessageBase'. Path: $[0] | LineNumber: 0 | BytePositionInLine: 2.
```

**To Reproduce**
```
var chatHistory = JsonSerializer.Deserialize<ChatHistory>(chatHistoryString);
```

Fails with both `System.Text.Json` and `Newtonsoft.Json`.

**Expected behavior**
In the scenario where chat history is not stored on the server side it is important to easily pass the object back and forth serialized as a string.

**Platform**
 - OS: Mac
 - IDE:  VS Code
 - Language: C#
 - Source: Microsoft.SemanticKernel, 0.19.230804.2-preview

**Additional context**
Building a chatbot with the chat history stored on the client side.

alexminza:
Had to come with something like this to be able to unblock the development progress:

```csharp
public class ChatHistoryMessageSurrogate
{
    public AuthorRoleSurrogate Role { get; set; }

    public string Content { get; set; }

    public ChatHistoryMessageSurrogate(AuthorRoleSurrogate role, string content)
    {
        this.Role = role;
        this.Content = content;
    }
}

public class AuthorRoleSurrogate
{
    public string Label { get; set; }

    public AuthorRoleSurrogate(string label)
    {
        this.Label = label!;
    }
}

public static ChatHistory? DeserializeChatHistory(string chatHistoryString, ILogger? logger)
{
    ChatHistory? chatHistory = null;

    try
    {
        var chatHistoryList = JsonSerializer.Deserialize<List<ChatHistoryMessageSurrogate>>(chatHistoryString)!;

[...]

Issue #2687, opened 2023-09-01
ChatHistory should not derive from `List<ChatMessageBase>`


stephentoub:
(1)
It's confusing for a consumer, as you then see methods on `ChatHistory` like `Add(ChatMessageBase)`, but you can't new-up a ChatMessageBase as it's abstract.  `ChatHistory` should expose on it whatever API surface area it needs and implementing whatever collection interfaces make sense, and the storage for it should be an implementation detail contained in the class.

(2)
(It's extra strange that not only does `ChatHistory` derive from `List<ChatMessageBase>`, it also exposes a `List<ChatMessageBase> Messages` property that just returns `this`.)

(3)
Also, why is the type called "ChatMessageBase"?  If it's the type that consumers are intended to work with, and connectors are supposed to create their own private implementations to hand out, it should likely just be "ChatMessage".

(4)
Why is there an "SKChatMessage" that's public in Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk?

nacharya1:
@shawncal - can you take a look into this question/request.

PR #3517, opened 2023-11-15
.Net: Add serialisation and deserialisation of ChatHistory

markwallace-microsoft:
### Motivation and Context

Resolves https://github.com/microsoft/semantic-kernel/issues/2443

### Description

- Base class for `ChatMessage` is not abstract

### Contribution Checklist


SergeyMenshykh (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
```diff
@@ -7,11 +7,11 @@
 namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
 
 /// <summary>
-/// Chat message representation from Semantic Kernel ChatMessageBase Abstraction
+/// Chat message representation from Semantic Kernel ChatMessage Abstraction
 /// </summary>
-public class SKChatMessage : ChatMessageBase
+public class SKChatMessage : SemanticKernel.AI.ChatCompletion.ChatMessage
```
stephentoub, line 12: Can we rename this as well? e.g. AzureOpenAIChatMessage?  Calling it SKChatMessage when it's actually specific to the Azure OpenAI connector, and when the base type that is specific to SK is called just ChatMessage feels wrong.

dotnet/samples/KernelSyntaxExamples/Example34_CustomChatModel.cs
```diff
@@ -89,13 +89,13 @@ public MyChatResult(MyRoles role, string content)
...
 
-    public Task<ChatMessageBase> GetChatMessageAsync(CancellationToken cancellationToken = default)
+    public Task<ChatMessage> GetChatMessageAsync(CancellationToken cancellationToken = default)
     {
         return Task.FromResult(this._message);
     }
 }
 
-public class MyChatMessage : ChatMessageBase
+public class MyChatMessage : ChatMessage
```
stephentoub, line 98: Is this needed at all now that ChatMessage's ctor is public?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
```diff
@@ -7,11 +7,11 @@
...
 {
-    private readonly ChatMessage? _message;
+    private readonly Azure.AI.OpenAI.ChatMessage? _message;
```
stephentoub, line 14: Would it make sense to replace the `public FunctionCall FunctionCall { get; }` property with `public Azure.AI.OpenAI.ChatMessage Message { get; }`? i.e. expose the whole message rather than just a piece of it, in order to provide any addditional further details a developer might want to see?

dotnet/src/SemanticKernel.UnitTests/AI/ChatCompletion/ChatHistoryTests.cs
```diff
@@ -0,0 +1,73 @@
...
+/// Unit tests of <see cref="ChatHistory"/>.
+/// </summary>
+public class ChatHistoryTests
+{
+    [Fact]
+    public void ItCanBeSerialised()
+    {
+        // Arrange
+        var options = new JsonSerializerOptions();
+        options.Converters.Add(new AuthorRoleConverter());
```
stephentoub, line 21: Should the AuthorRole property instead be tagged with [JsonConverter(typeof(AuthorRoleConverter))] so that this isn't necessary?

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessage.cs
```diff
@@ -17,20 +17,27 @@ public abstract class ChatMessageBase
     /// <summary>
     /// Content of the message
     /// </summary>
-    public string Content { get; set; }
+    public string Content { get; set; } = string.Empty;
```
stephentoub, line 20: Do we want to add any validation code to the setters to prevent Content/Role from being set to null?

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessage.cs
```diff
@@ -17,20 +17,27 @@ public abstract class ChatMessageBase
...
+    /// Creates a new instance of the <see cref="ChatMessage"/> class
     /// </summary>
     /// <param name="role">Role of the author of the message</param>
     /// <param name="content">Content of the message</param>
     /// <param name="additionalProperties">Dictionary for any additional message properties</param>
-    protected ChatMessageBase(AuthorRole role, string content, IDictionary<string, string>? additionalProperties = null)
+    public ChatMessage(AuthorRole role, string content, IDictionary<string, string>? additionalProperties = null)
     {
         this.Role = role;
         this.Content = content;
```
stephentoub, line 43: You added `= string.Empty` to the property initializer.  If someone passes in a null role / content, what do you want to do?  Should they still end up as empty? Throw?

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -9,19 +9,12 @@ namespace Microsoft.SemanticKernel.AI.ChatCompletion;
...
-        public ChatMessage(AuthorRole authorRole, string content, IDictionary<string, string>? additionalProperties) : base(authorRole, content, additionalProperties)
-        {
-        }
-    }
-
     /// <summary>
     /// List of messages in the chat
     /// </summary>
-    public List<ChatMessageBase> Messages => this;
+    public List<ChatMessage> Messages => this;
```
stephentoub, line 17: Why is this property necessary? As I noted in https://github.com/microsoft/semantic-kernel/issues/2687, it's strange to me that ChatHistory both derives from `List<ChatMessage>` and exposes a `List<Chat> Messages`; seems like we should have one or the other, but not both.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessage.cs
RogerBarreto: If a ChatMessage has just function calling the content can be null.

IMO this should be

```suggestion
    public string? Content { get; set; };
```

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessage.cs
```diff
@@ -17,20 +17,27 @@ public abstract class ChatMessageBase
...
     /// <summary>
     /// Dictionary for any additional message properties
     /// </summary>
     public IDictionary<string, string>? AdditionalProperties { get; set; }
 
     /// <summary>
-    /// Creates a new instance of the <see cref="ChatMessageBase"/> class
+    /// Creates a new instance of the <see cref="ChatMessage"/> class
+    /// </summary>
+    public ChatMessage()
```
RogerBarreto, line 30: Isn't this enough to get the Issue fixed? 

Why `ChatMessageBase` can't be abstract?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ChatResult.cs
```diff
@@ -23,8 +23,8 @@ public ChatResult(ChatCompletions resultData, ChatChoice choice)
 
     public ModelResult ModelResult { get; }
 
-    public Task<ChatMessageBase> GetChatMessageAsync(CancellationToken cancellationToken = default)
-        => Task.FromResult<ChatMessageBase>(new SKChatMessage(this._choice.Message));
+    public Task<SemanticKernel.AI.ChatCompletion.ChatMessage> GetChatMessageAsync(CancellationToken cancellationToken = default)
```
RogerBarreto, line 26: +1 to the reasons why moving the Base abstraction class to non abstract (clashing the names within other packages)

RogerBarreto (approved):
Some comments.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessage.cs
markwallace-microsoft: The problem is that `ChatHistory` was a list of `ChatMessageBase` so deserialization fail. Once I make `ChatMessageBase` becomes non abstract it's the same as `ChatMessage`.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
markwallace-microsoft: We'd have to move the logic to throw the exception also.  I've created a task to follow up on this: https://github.com/microsoft/semantic-kernel/issues/3521

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessage.cs
RogerBarreto: We can resolve the Deserialization problem but indeed won't be able to serialize it back. 

Make sense

dotnet/src/SemanticKernel.UnitTests/AI/ChatCompletion/ChatHistoryTests.cs
markwallace-microsoft: I can use `[JsonConstructor]` instead

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKChatMessage.cs
markwallace-microsoft: Done

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
markwallace-microsoft: Removed `List<Chat> Messages`

Created  task for the design of `ChatHistory` https://github.com/microsoft/semantic-kernel/issues/3522

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatMessage.cs
markwallace-microsoft: I am reverting this change and i have a follow up task https://github.com/microsoft/semantic-kernel/issues/3522

RogerBarreto (approved):

SergeyMenshykh (approved):

PR #3526, opened 2023-11-16
.Net: Propagate original unwrapped exception from NativeMethod

stephentoub:
### Motivation and Context

Native method invocation happens via MethodInfo.Invoke, which by default wraps the underlying exception in a TargetInvocationException. This is an implementation detail that should not be exposed to the consumer. This fixes it to propagate the original exception instead.

### Contribution Checklist


SergeyMenshykh (approved):

stephentoub:
The CI format leg is failing because of warnings in main.  I've fixed them in https://github.com/microsoft/semantic-kernel/pull/3536.

markwallace-microsoft (approved):

PR #3530, opened 2023-11-16
.Net: Make SemanticKernel.UnitTests run on both .NET 6 and .NET Framework 4.8

stephentoub:
In order to get some validation on .NET Framework.


dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -120,47 +120,49 @@ public async Task RunAsyncHandlesPreInvocationAsync(int pipelineCount)
     {
         // Arrange
         var sut = new KernelBuilder().Build();
-        var myPlugin = new Mock<MyPlugin>();
```
stephentoub, line 123: While the changes in this file are nice anyway, the impetus for changing them was because there appears to be an issue on .NET Framework with getting attributes on return parameters when the target object is coming from Mock.

dotnet/src/InternalUtilities/src/Http/HttpClientExtensions.cs
```diff
@@ -23,27 +24,34 @@ internal static class HttpClientExtensions
...
-
-            return response;
         }
         catch (HttpRequestException e)
         {
-            string? responseContent = null;
+            throw new HttpOperationException(HttpStatusCode.BadRequest, null, e.Message, e);
+        }
 
+        if (!response.IsSuccessStatusCode)
```
stephentoub, line 36: Without these changes, the response content on .NET Framework was always empty.

markwallace-microsoft (approved):

SergeyMenshykh (approved):

PR #3531, opened 2023-11-16
.Net: Kernel is passed to SKFunction.InvokeAsync method

SergeyMenshykh:
### Motivation and Context
The goal of the PR is to add the Kernel class as a parameter to the ISKFunction.InvokeAsync method.


dotnet/samples/KernelSyntaxExamples/Example06_TemplateLanguage.cs
```diff
@@ -54,7 +54,7 @@ public static async Task RunAsync()
         Console.WriteLine("--- Rendered Prompt");
         var promptTemplateFactory = new BasicPromptTemplateFactory();
         var promptTemplate = promptTemplateFactory.Create(FunctionDefinition, new PromptTemplateConfig());
-        var renderedPrompt = await promptTemplate.RenderAsync(kernel.CreateNewContext());
+        var renderedPrompt = await promptTemplate.RenderAsync(kernel, kernel.CreateNewContext());
```
RogerBarreto, line 57: Do we need to pass this? 

Can't we just get SKContext (while still exists) from within the promptTemplateRender method?

RogerBarreto (approved):

dotnet/samples/KernelSyntaxExamples/Example06_TemplateLanguage.cs
RogerBarreto: Never mind, looking forward to see the upcoming changes on those APIs 🚀

PR #3535, opened 2023-11-16
.Net: Fix warnings causing CI format leg to fail

stephentoub:


SergeyMenshykh (approved):

markwallace-microsoft:
We should fix all of the namespaces, these are the old ones prior to moving the code

stephentoub:
> We should fix all of the namespaces, these are the old ones prior to moving the code

These were the ones causing https://github.com/microsoft/semantic-kernel/pull/3526 to fail CI.

markwallace-microsoft:
@stephentoub I'm going to replace this with an equivalent PR to fix the namespaces, https://github.com/microsoft/semantic-kernel/pull/3539

PR #3536, opened 2023-11-16
.Net: Delete / implement remaining [Obsolete] APIs from feature-api-v1.0

stephentoub:
Deleted most of the remaining ones, but for the HuggingFaceTextCompletion streaming implementation, rather than throwing, implemented it in terms of the non-streaming implementation so that it "just works".


RogerBarreto (approved):

stephentoub:
Will close and re-open against main once https://github.com/microsoft/semantic-kernel/pull/3543 lands.

PR #3543, opened 2023-11-17
.Net Kernel improvements for V1

SergeyMenshykh:
### Motivation and Context
This PR introduces a few new improvements to make SK simpler, flexible and developer friendly:
- Plugins support - https://github.com/microsoft/semantic-kernel/pull/3474
- Kernel as a state/property bag - https://github.com/microsoft/semantic-kernel/pull/3489, https://github.com/microsoft/semantic-kernel/pull/3516
- SKFunction / SKPlugin factory methods - https://github.com/microsoft/semantic-kernel/pull/3493
- Kernel is passed to SKFunction.InvokeAsync - https://github.com/microsoft/semantic-kernel/pull/3531


stephentoub:
> This PR introduces a few new improvements to make SK simpler, flexible and developer friendly:

For anyone following along, this effort is not yet complete; there are more pieces necessary to complete the story, so this is an intermediate state. :)

stephentoub (approved):

dotnet/src/SemanticKernel.Core/Functions/SKPlugin.cs
```diff
@@ -0,0 +1,182 @@
...
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Provides an <see cref="ISKPlugin"/> implementation around a collection of functions
+/// as well as static factory methods for creating commonly-used plugin implementations.
+/// </summary>
+[DebuggerDisplay("Name = {Name}, Functions = {FunctionCount}")]
+[DebuggerTypeProxy(typeof(SKPlugin.TypeProxy))]
+public sealed class SKPlugin : ISKPlugin
```
dmytrostruk, line 24: I still prefer to avoid `SK` prefixing in naming, since namespace already contains product name. 
In my opinion, `Microsoft.SemanticKernel.Plugin` looks better and more readable than `Microsoft.SemanticKernel.SKPlugin`.

If we decide to avoid prefixing in all places, it's better to avoid it in new types now, rather than add it now with prefix and rename it later.

Also, currently there are some inconsistencies in names like `Plan`, `Kernel`, `SKFunction`, `SKPlugin`. I think we should make it either:
- `Plan`, `Kernel`, `Function`, `Plugin` - this one is my preferred option, or:
- `SKPlan`, `SKKernel`, `SKFunction`, `SKPlugin`.

stephentoub: As noted in https://github.com/microsoft/semantic-kernel/pull/3543#issuecomment-1817103587, this is not done; there are multiple subsequent rounds of changes necessary to complete this work. One of those is naming consistency.

> I think we should make it either

There are basically three options:
1. No prefix.
2. SK prefix
3. Kernel prefix

The problem with (1) is some of these words are super common. If you import multiple namespaces, there's a high likelyhood of conflict.  Even if you don't, there's conceptual conflict; if you're using SK with Azure Functions, what is a Function... is that related to SK or related to Azure Functions?  If you're using SK with Office, is a Plugin an SK plugin or an Office plugin?  Etc.

(2) would be what you enumerate, e.g. SKPlan, ISKFunction, ISKPlugin, though I don't know about Kernel itself. That is a rarer word with less chance of conflict, and whereas "SKFunction" is expanded to "semantic kernel function", "SKKernel" would be "semantic kernel kernel", which is duplicative.  Current thinking on this option is SK-prefix for all the core types, except for Kernel.

(3) would be Kernel, IKernelFunction, IKernelPlugin, etc.  It's a bit longer, but it also ties everything back to "Kernel".

markwallace-microsoft (approved):
Changes LGTM and have already been reviewed in detail. Just need action on the CodeQL issues.

markwallace-microsoft (approved):

markwallace-microsoft (approved):

markwallace-microsoft (approved):

PR #3545, opened 2023-11-17
.Net: Revise core "View" types

stephentoub:
*** NOTE: This is based on https://github.com/microsoft/semantic-kernel/pull/3543, so ignore all commits except for the last one. **
https://github.com/microsoft/semantic-kernel/commit/67de187291a8a8be37b7c2111c7f325ca4cf059a

These "View" types are C# records. C# records are incredibly handy for internal implementation detail. However, they're less desirable for public surface area of a long-lived library:
- Argument names to the ctor don't follow common casing convention
- Equality semantics for records that contain collections are wonky; the default equality check is looking for whether the reference identity of the collections is the same, not a deep compare, so these view types are using an intermediate form of equality comparison. (It Since view comparison isn't commonly needed, better to just use reference identity all-up.
- Exposing all properties as part of the constructor leads to a design where future added properties either are lesser in that they're not part of the constructor, or it requires adding new constructors. A design where everything is just properties means all properties today and tomorrow are on equal footing. (This can be done with records as well, but it's not the common approach.)
- Exposing the properties normally means we can do validation in the initers.

On top of that, the naming is a little weird here. There's "SKFunction" but then the view of it is "FunctionView" rather than "SKFunctionView". And these "views" aren't just about subsetting, or surfacing a readonly view of writable data; it's surfacing data in an entirely different form than the underlying representation.  Plus you get these via a "Describe" method, which doesn't make it discoverable how to get one of these instances from a function.

Also, the IsRequired property on the parameter view was a `bool?`. It's not clear why it was nullable or what that third state means; I made it just `bool`.

This commit moves these from being records to normal classes and also cleans up the naming.  It also renames the ParameterViewType to ParameterJsonType, as it's really just a strongly-typed string for a JSON data type. However, I'm hopeful this type will go away entirely shortly.


dotnet/src/SemanticKernel.Core/Extensions/FunctionViewExtensions.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.SemanticKernel.Extensions;
 /// <summary>
 /// Extensions for functions views.
 /// </summary>
-public static class FunctionViewExtensions
+public static class SKFunctionMetadataExtensions
```
SergeyMenshykh, line 12: nit: the file name still has old name - 'FunctionViewExtensions.cs'

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Core/Extensions/FunctionViewExtensions.cs
stephentoub: Oops, I didn't see this. Will rename it.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Core/Extensions/FunctionViewExtensions.cs
stephentoub: Actually, let me do this in a follow-up PR, since it's minor and I have two signoffs :)

stephentoub: Nevermind, I needed to fix a merge conflict anyway; fixed.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #3550, opened 2023-11-18
.Net: Update repo to LangVersion 11

stephentoub:
C# 11 shipped a year ago; there's no reason not to be using it at this point, and it brings with it support for raw string literals, which are particularly valuable when working with lots of JSON, as this repo does.

.NET 8 and C# 12 shipped last week.  We should update the repo to LangVersion 12 in the near future, but we can wait to do so until after v1.0 and give folks some time to update their toolsets.


stephentoub:
Ugh, CI is still using a .NET 6 SDK for some of the legs, in which case this won't work. 😦

PR #3551, opened 2023-11-18
.Net: Fix up some issues related to JSON Schema

stephentoub:
**Commit 1: Use SKJsonSchema instead of JsonDocument for Schema properties**

System.Text.Json doesn't currently have JSON schema support. As such, SK is currently using JsonDocument, however this is problematic because JsonDocuments are IDisposable due to holding on to ArrayPool resources; if you don't dispose them, you take away arrays from the pool that could have been used by others, leading to process-wide performance issues, and even if that wasn't a problem, the lack of ownership transfer with an IDisposable in these object models makes it confusing.

Ideally it would just use a System.Text.Json schema type, but none exists, and this needs to ship before any such thing might materialize in a hypothetical future. So we need an alternative, which could include using the type from the underlying JSON schema library being used, using string, using JsonElement, or using a custom type. We don't want to use the underlying library's type as that would bake that dependency into the API forever rather than it just being an implementation detail that may come and go. string is viable, but it means the JSON won't necessarily be validated as it flows through the system, and if it's serialized out, it won't respect whatever settings are part of the JsonSerializerOptions (since it's already rendered JSON). JsonElement would work, but it's a bit odd to have one in an object model, since it's logically part of a JsonDocument, and even then we don't necessarily want to force the schema to be stored as a JsonElement, e.g. if in the future S.T.J did add a schema type, it'd be nice to use that as the underlying representation.

So, I've gone with adding a new SKJsonSchema type, that's just a wrapper around a JsonElement as an implementation detail.  It exposes that element as a property, but if its implementation changes in the future, the implementation of that property could also change to lazily materialize that JsonElement.  Everywhere JsonDocument was being used for a schema in the public surface area is now using SKJsonSchema instead.

**Commit 2: Move JsonSchemaFunctionXx types to be internal in planners**

These can be exposed in the future as needed, but for now we can keep all of the function manual work internal in planners until it's proved out further.  This required removing the dependency on JsonSchemaFunctionView from OpenAI, but that's a good thing, regardless, as the implementation was doing much more than was needed cost-wise.

cc: @eiriktsarpalis


markwallace-microsoft (approved):

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Functions/SKJsonSchema.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    /// <summary>The schema stored as a string.</summary>
+    private string? _schemaAsString;
+
+    /// <summary>Parses a JSON Schema for a parameter type.</summary>
+    /// <param name="jsonSchema">The JSON Schema as a string.</param>
+    /// <returns>A parsed <see cref="SKJsonSchema"/>.</returns>
+    /// <exception cref="ArgumentException"><paramref name="jsonSchema"/> is null.</exception>
+    /// <exception cref="JsonException">The JSON is invalid.</exception>
+    public static SKJsonSchema Parse(string jsonSchema) => new(JsonSerializer.Deserialize<JsonElement>(jsonSchema));
```
eiriktsarpalis, line 23: Technically, these JsonSerializer methods require reflection to work (even though the particular type uses a built-in converter). At some point this should be updated to use a JsonSerializerContext that includes `JsonElement`.

dotnet/src/SemanticKernel.Abstractions/Functions/SKJsonSchema.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// <summary>Initializes a new instance from the specified <see cref="JsonElement"/>.</summary>
+    /// <param name="jsonSchema">The schema to be stored.</param>
+    private SKJsonSchema(JsonElement jsonSchema) => this.RootElement = jsonSchema;
+
+    /// <summary>Gets a <see cref="JsonElement"/> representing the root element of the schema.</summary>
+    public JsonElement RootElement { get; }
+
+    /// <summary>Gets the JSON Schema as a string.</summary>
+    public override string ToString() => this._schemaAsString ??= JsonSerializer.Serialize(this.RootElement);
```
eiriktsarpalis, line 45: Is this being used at all? If this is meant for user-readability perhaps it makes sense to enable indentation?

dotnet/src/SemanticKernel.Abstractions/Functions/SKJsonSchema.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <param name="jsonSchema">The JSON Schema as a sequence of UTF16 chars.</param>
+    /// <returns>A parsed <see cref="SKJsonSchema"/>.</returns>
+    /// <exception cref="JsonException">The JSON is invalid.</exception>
+    public static SKJsonSchema Parse(ReadOnlySpan<char> jsonSchema) => new(JsonSerializer.Deserialize<JsonElement>(jsonSchema));
+
+    /// <summary>Parses a JSON Schema for a parameter type.</summary>
+    /// <param name="utf8JsonSchema">The JSON Schema as a sequence of UTF8 bytes.</param>
+    /// <returns>A parsed <see cref="SKJsonSchema"/>.</returns>
+    /// <exception cref="JsonException">The JSON is invalid.</exception>
+    public static SKJsonSchema Parse(ReadOnlySpan<byte> utf8JsonSchema) => new(JsonSerializer.Deserialize<JsonElement>(utf8JsonSchema));
```
eiriktsarpalis, line 35: These methods will at the moment accept any valid JSON document (or even just plain values such as `null` or `[]`). Perhaps some validation (or a debug assertion) from JsonSchema.NET might be warranted here?

eiriktsarpalis (approved):
SKJsonSchema changes LGTM overall.

dotnet/src/SemanticKernel.Abstractions/Functions/SKJsonSchema.cs
stephentoub: > At some point this should be updated to use a JsonSerializerContext that includes JsonElement.

JsonSerializer is used extensively with reflection throughout SK.  At one point I'd gone through and updated it to use source gen, but it didn't get merged as it was too painful for too little gain at the time.  I'm sure we'll want to revisit it at some point.

dotnet/src/SemanticKernel.Abstractions/Functions/SKJsonSchema.cs
stephentoub: Makes sense.

stephentoub: Ah, I can't here.  This core assembly doesn't reference that library.

markwallace-microsoft (approved):

markwallace-microsoft (approved):

Issue #3514, opened 2023-11-15
Python: Fix broken notebooks and update integration tests


markwallace-microsoft:
Following notebooks are only partially working

1. `10-multiple-results-per-prompt.ipynb`
1. `11-streaming-completions.ipynb`

As part of this task integration tests must be extended to cover the scenarios covered by the notebooks.

Also configuration to use Azure should be set once in the .env file

john0isaac:
I'm working on this

PR #3557, opened 2023-11-19
Python: Fix notebooks 10 and 11

john0isaac:
### Motivation and Context

fix: #3514, I didn't know which integration tests @markwallace-microsoft is referring to but I fixed both notebooks by introducing a check before running the code that users have to define at the beginning of the notebook.

### Description

Whichever service the user chooses (OpenAI - Azure OpenAI - Huggine Face) the code adjusts accordingly without the need to change anything later other than the .env file (obviously) and the Service name at the second cell of the two notebooks.


moonbox3:
Hi @john0isaac, thank you for your help with this. We've had a lot of (breaking) changes over the last few weeks/months after this PR was created. There have been updates on these notebooks since then and I believe they're working fine. If you have time, could you verify, please?

john0isaac:
@moonbox3 Thank you for checking this PR.
You are correct, I have fixed multiple things in this PR which have been addressed by other separate PRs that were already merged.

But the main problem still exists in these notebooks

Running code that calls hugging face modules, openai modules, and azure openai modules in the same cells gives you errors:
image

The main feature that this PR adds which has not been fixed in other PRs is a variable at the beginning of the notebook to choose between (openai, azure, hugging face)
set it one time
then run the whole notebook and you won't get any errors as there are checks to see which one did you choose and run the code related to it.

I will fix the merge conflicts one last time if you approve of the idea you can merge it, if not please close this PR.

john0isaac:
Another thing I noticed `text-davinci-003` is no longer available for deployment from Azure OpenAI See MSFT reference https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/legacy-models

It was replaced by `gpt-35-turbo-instruct` - I added a note to the comment

john0isaac:
@moonbox3 Done ✅.
Please check and let me know what you think about this..

python/notebooks/10-multiple-results-per-prompt.ipynb
```diff
@@ -1,362 +1,391 @@
...
+                "        deployment_name=\"text-davinci-003\", endpoint=endpoint, api_key=api_key\n",
+                "    )  # set the deployment name to the value of your text model (e.g. text-davinci-003 or gpt-35-turbo-instruct)\n",
+                "    azure_chat_service = AzureChatCompletion(\n",
+                "        deployment_name=\"gpt-35-turbo\", endpoint=endpoint, api_key=api_key\n",
+                "    )  # set the deployment name to the value of your chat model\n",
+                "\n",
+                "# Configure OpenAI service\n",
+                "if selectedService == Service.OpenAI:\n",
+                "    api_key, org_id = sk.openai_settings_from_dot_env()\n",
+                "    oai_text_service = OpenAITextCompletion(ai_model_id=\"text-davinci-003\", api_key=api_key, org_id=org_id)\n",
```
moonbox3, line 103: Let's replace with `gpt-35-turbo-instruct`, please. I am even getting a 404 today when trying to use this model since it was deprecated on 1/4/24.

python/notebooks/11-streaming-completions.ipynb
```diff
@@ -1,305 +1,334 @@
...
+                "        deployment_name=\"text-davinci-003\", endpoint=endpoint, api_key=api_key\n",
+                "    )  # set the deployment name to the value of your text model (e.g. text-davinci-003 or gpt-35-turbo-instruct)\n",
+                "    azure_chat_service = AzureChatCompletion(\n",
+                "        deployment_name=\"gpt-35-turbo\", endpoint=endpoint, api_key=api_key\n",
+                "    )  # set the deployment name to the value of your chat model\n",
+                "\n",
+                "# Configure OpenAI service\n",
+                "if selectedService == Service.OpenAI:\n",
+                "    api_key, org_id = sk.openai_settings_from_dot_env()\n",
+                "    oai_text_service = OpenAITextCompletion(ai_model_id=\"text-davinci-003\", api_key=api_key, org_id=org_id)\n",
```
moonbox3, line 97: Same replacement here?

moonbox3 (changes requested):
LGTM but two small changes requested, please. Thank you!

python/notebooks/11-streaming-completions.ipynb
john0isaac: Yep, it's the same for both

python/notebooks/10-multiple-results-per-prompt.ipynb
john0isaac: done

moonbox3 (approved):
LGTM

juliomenendez (changes requested):
Would it make sense to refactor the `Service` enum to a module that can be used by all the notebooks instead of repeated in each of them? I see this being used not just in those 2 notebooks in the future.
Otherwise, LGTM

juliomenendez (approved):
Thanks! LGTM

moonbox3:
@john0isaac it just needs a quick `ruff . --fix` and you should be good to go.

john0isaac:
@moonbox3 Done ✅ 
Thank you so much!

moonbox3:
We should file an issue to update the other notebooks from using the bool to determine which service to use to this new service enum you've added.

moonbox3 (approved):

Issue #3556, opened 2023-11-19
.Net: Why are there interfaces for IActionPlanner / ISequentialPlanner / IStepwisePlanner?


stephentoub:
We expect there to be other implementations of these exact planning schemes and for code to be written in terms of these abstractions?

If we need a planning abstraction, shouldn't it be one interface used by all of these?
```C#
public interface IPlanner
{
    Task<Plan> CreatePlanAsync(string goal, CancellationToken cancellationToken = default);
}
```
or something like that?

PR #3558, opened 2023-11-20
.Net: Clean up planner interfaces and instrumentation

stephentoub:
Fixes https://github.com/microsoft/semantic-kernel/issues/3556
Contributes to https://github.com/microsoft/semantic-kernel/issues/3490

This removes IActionPlanner, ISequentialPlanner, and IStepwisePlanner, and replaces them all with just IPlanner. Each of the interfaces also had a WithInstrumentation extension method and dedicated implementations for each; those have all been consolidated down to just a single extension method and implementation.  While doing that, I cleaned up the metrics for both the planners and functions to better follow OpenTelemetry semantic conventions. (Subsequently, we'll want to delete WithInstrumentation entirely and just have the metrics always there rather than being opt-in per function.)


dotnet/src/Planners/Planners.Core/InstrumentedPlanner.cs
```diff
@@ -0,0 +1,80 @@
...
+        catch (Exception ex)
+        {
+            this._logger.LogError(ex, "Plan creation failed: {Message}", ex.Message);
+            throw;
+        }
+        finally
+        {
+            TimeSpan duration = new((long)((Stopwatch.GetTimestamp() - startingTimestamp) * (10_000_000.0 / Stopwatch.Frequency)));
+            this._logger.LogInformation("Plan creation finished in {ExecutionTime}ms.", duration.TotalMilliseconds);
+            s_createPlanDuration.Record(duration.TotalSeconds, new KeyValuePair<string, object?>("planner", plannerName));
```
JamesNK, line 77: Could log exception as error.type here.

dotnet/src/Planners/Planners.Core/IPlanner.cs
```diff
@@ -11,9 +11,9 @@ namespace Microsoft.SemanticKernel.Planners;
 #pragma warning restore IDE0130
 
 /// <summary>
-/// Interface for planner that uses a set of semantic functions to create a sequential plan.
+/// Interface for a planner that creates a plan to achieve a goal.
 /// </summary>
-public interface ISequentialPlanner
+public interface IPlanner
```
markwallace-microsoft, line 16: Any reason not to move this to ms.sk.core? Or move both `IPlanner` and `Plan` to ms.sk.abstractions?

dotnet/src/SemanticKernel.Core/Functions/InstrumentedSKFunction.cs
```diff
@@ -18,153 +18,96 @@
...
+    private static readonly Counter<int> s_successfulInvocations = s_meter.CreateCounter<int>(
+        name: "sk.function.invocations",
+        unit: "{invocation}",
+        description: "Number of successful function invocations");
+
+    /// <summary><see cref="Counter{T}"/> to track the number of failed function invocations.</summary>
+    private static readonly Counter<int> s_errors = s_meter.CreateCounter<int>(
+        name: "sk.function.errors",
+        unit: "{error}",
+        description: "Number of function invocation errors");
```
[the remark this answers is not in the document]
stephentoub: You tell me :-)

I was following examples from opentelemetry, like for FaaS. What do we do in similar situations?

Will Aspire show all the relevant information in the dashboard of we just use duration?

dotnet/src/SemanticKernel.Core/Functions/InstrumentedSKFunction.cs
```diff
@@ -18,153 +18,96 @@
 namespace Microsoft.SemanticKernel;
 #pragma warning restore IDE0130
 
-/// <summary>
-/// Standard Semantic Kernel callable function with instrumentation.
-/// </summary>
+/// <summary>Instrumented function that surrounds the invocation of another function with loggic and metrics.</summary>
```
markwallace-microsoft, line 21: Typo `loggic`

markwallace-microsoft (approved):
Couple of minor comments

dotnet/src/SemanticKernel.Core/Functions/InstrumentedSKFunction.cs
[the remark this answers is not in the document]
JamesNK: > I was following examples from opentelemetry, like for FaaS. What do we do in similar situations?

OTEL semantic conventions for HTTP only have a duration histogram and active_requests up/down counter. It doesn't have total counters. The reason is that the histogram tracks values and counts.

> Will Aspire show all the relevant information in the dashboard if we just use duration? or if we just use invocations, will it easily show a chart of errors?

That is a good point. In Grafana you can access histogram values as counts and then create charts to display counts. Aspire always displays histograms as P50/P90/P99 charts without counts.

The problem here is Aspire. It should have the option to switch histograms between displaying percentiles and counts.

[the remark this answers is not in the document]
stephentoub: Ok, I can consolidate this to just duration.

> That is a good point. [...] The problem here is Aspire. It should have the option to switch histograms between displaying percentiles and counts.

Sounds like you're on it?  And I assume it'll enable filtering / displaying based on tags / attributes so that you can see the difference between success / errors?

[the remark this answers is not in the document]
stephentoub: I read https://github.com/dotnet/aspire/issues/965; I don't see any mention of tags / attributes.  Will someone be able to differentiate successes from failures in a view where the histogram is rendered as counts?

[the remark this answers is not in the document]
JamesNK: Yes. It will have all the same filtering capability.

dotnet/src/SemanticKernel.Core/Functions/InstrumentedSKFunction.cs
```diff
@@ -18,153 +18,92 @@
...
             throw;
         }
         finally
         {
-            stopwatch.Stop();
-            this._executionTotalCounter.Add(1);
-            this._executionTimeHistogram.Record(stopwatch.ElapsedMilliseconds);
+            TimeSpan duration = new((long)((Stopwatch.GetTimestamp() - startingTimestamp) * (10_000_000.0 / Stopwatch.Frequency)));
+            this._logger.LogInformation("Function invocation duration: {Duration}ms", duration.TotalMilliseconds);
+            s_invocationDuration.Record(duration.TotalSeconds, in tags);
```
TaoChenOSU, line 103: Do we also want to add the tags to the activity?

dotnet/src/SemanticKernel.Core/Planning/InstrumentedPlanner.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            this._logger.LogError(ex, "Plan creation failed: {Message}", ex.Message);
+            tags.Add("error.type", ex.GetType().FullName);
+            throw;
+        }
+        finally
+        {
+            TimeSpan duration = new((long)((Stopwatch.GetTimestamp() - startingTimestamp) * (10_000_000.0 / Stopwatch.Frequency)));
+            this._logger.LogInformation("Plan creation duration: {Duration}ms.", duration.TotalMilliseconds);
+            s_createPlanDuration.Record(duration.TotalSeconds, in tags);
```
TaoChenOSU, line 88: Do we also want to add the tags to the activity?

dotnet/src/SemanticKernel.Core/Functions/InstrumentedSKFunction.cs
```diff
@@ -18,153 +18,92 @@
...
-/// <summary>
-/// Standard Semantic Kernel callable function with instrumentation.
-/// </summary>
+/// <summary>Instrumented function that surrounds the invocation of another function with logging and metrics.</summary>
 internal sealed class InstrumentedSKFunction : ISKFunction
 {
-    /// <inheritdoc/>
-    public string Name => this._function.Name;
+    /// <summary><see cref="ActivitySource"/> for function-related activities.</summary>
+    private static readonly ActivitySource s_activitySource = new(typeof(ISKFunction).FullName!);
```
CodeBlanch, line 25: I feel like "Microsoft.SemanticKernel" would be a better name for the `ActivitySource` as well. To listen in OTel you would do something like...

```csharp
using var tracerProvider = Sdk.CreateTracerProviderBuilder()
   .AddSource("Microsoft.SemanticKernel")
   .AddMyFavoriteExporter()
   .Build();
```

As opposed to...

```csharp
using var tracerProvider = Sdk.CreateTracerProviderBuilder()
   .AddSource("Microsoft.SemanticKernel.ISKFunction")
   .AddMyFavoriteExporter()
   .Build();
```

dotnet/src/SemanticKernel.Core/Functions/InstrumentedSKFunction.cs
```diff
@@ -18,153 +18,92 @@
...
-        using var activity = s_activitySource.StartActivity($"{this.PluginName}.{this.Name}");
-
-        this._logger.LogInformation("{PluginName}.{FunctionName}: Function execution started.", this.PluginName, this.Name);
-
-        var stopwatch = new Stopwatch();
-        stopwatch.Start();
-
-        FunctionResult result;
+        this._logger.LogInformation("Function invoking");
+        TagList tags = new() { { "function.name", this.Name } };
```
CodeBlanch, line 75: Might be best to prefix the tags eg `sk.function.name`, `sk.planner.name`, etc.

dotnet/src/SemanticKernel.Core/Functions/InstrumentedSKFunction.cs
stephentoub: The tags are passed to StartActivity, right?  The only tags being used with the metrics here are the function name, and if it fails, the exception type.  The latter isn't available when StartActivity is called, and the name of the function is being used as the name of the activity, so I'm not sure there's anything else to use as a tag?

stephentoub: Though addressing Mikel's feedback below, then it would no longer be the name, so the name could be used as a tag, I guess.

SergeyMenshykh (approved):

Issue #3151, opened 2023-10-11
.Net: Revisit namespaces and type placement


stephentoub:
> [!IMPORTANT]  
> Labeled **Urgent** because it may require a breaking change.

(Metapoint: Try rewriting all the examples without using `var` and seeing how many namespaces need to be imported to successfully compile.)

- [x] (1)
To do anything meaningful with memory now I need both:
```C#
using Microsoft.SemanticKernel.Memory; // for ISemanticTextMemory, unless I rely exclusively on var
using Microsoft.SemanticKernel.Plugins.Memory; // for MemoryBuilder and friends
```
plus a using for whatever store I'm using it if it's not the VolatileMemoryStore, e.g.
```C#
using Microsoft.SemanticKernel.Connectors.Memory.AzureCognitiveSearch;
```
That's three `using`s just to create an `ISemanticTextMemory`.  Can these be consolidated?

- [x] (2)
To do anything meaningful with planning, I now need both:
```C#
using Microsoft.SemanticKernel.Planners; // for the actual planner
using Microsoft.SemanticKernel.Planning; // for Plan
```
The difference between "Planners" and "Planning" is also very subtle.  Can these be consolidated?

- [x] (3)
It looks like Microsoft.SemanticKernel.Diagnostics exists now just for `SKException`, `HttpOperationException`, and `Telemetry`.  Do we need this namespace at all?  Can we move these types somewhere else, like moving `SKException` to the root `Microsoft.SemanticKernel` namespace?

- [x] (4)
Similarly, can we consolidate Microsoft.SemanticKernel.Events with something else? "Events" is pretty broad; the types here are specific to pre/post hooks with plan execution, right?

- [ ] (5) 
Why do we need Microsoft.SemanticKernel.Plugins.Core? Can't those just move down into Plugins?

- [x] (6)
We should also move "extensions" types to the namespace with the types they extend. For example, it looks like ImageGenerationServiceExtensions, TextEmbeddingServiceExtensions, TextCompletionServiceExtensions, only have extension methods and those only apply to `IAIService` in Microsoft.SemanticKernel.Services... can they move there?  And can they be consolidated into an AIServicesExtensions class?  And `OpenAIModelResultExtensions` has methods that operate only on `Microsoft.SemanticKernel.Orchestration.ModelResult`... can it move there?

- [x] (7)
What is MemoryConfiguration and why does it live in the root namespace?

- [x] (8)
[...]

PR #3565, opened 2023-11-20
.Net: Remove Microsoft.SemanticKernel.Diagnostics namespace

stephentoub:
It contained:
- SKException. As the exception that can be thrown from any component in the system and is the base of all SK-related exceptions, it should be in the root namespace.
- HttpOperationException. This is specific to HTTP.  There's already a Microsoft.SemanticKernel.Http namespace; it should be there.
- Telemetry. This type was problematic for a variety of reasons. It exposed a public IsTelemetryEnabled property that was actually specific to Azure; if someone needs to know if Azure telemetry is enabled/disabled, it should use an Azure-specific mechanism to do so. It was also only used to set OpenAIClientOptions.Diagnostics.IsTelemetryEnabled, but that property defaults to using the exact same environment variable IsTelemetryEnabled was itself reading from, making it a nop.  The other thing exposed from Telemetry was HttpUserAgent, which is not something that needs to be public; it's now on an internal class that's referenced where it's needed.
- Verify. This is internal and has been moved to the root namespace.  It also conflicted with a duplicate version of the type in the Qdrant connector; that duplicate has been removed, with a few additional things it needed brought into the shared internal type.
- HttpStatusCodeType. This is almost entirely duplicative of HttpStatusCode; deleted.

Contributes to https://github.com/microsoft/semantic-kernel/issues/3151


markwallace-microsoft (approved):
Just need to fix the format issue

SergeyMenshykh (approved):

SergeyMenshykh (approved):

SergeyMenshykh (approved):

stephentoub:
> note: removing the Telemetry logic means that telemetry can't be turned off using the `AZURE_TELEMETRY_DISABLED` variable anymore

What telemetry was it controlling that wasn't already controlled by AZURE_TELEMETRY_DISABLED?

PR #3567, opened 2023-11-20
.Net: Switch from ISKFunction to abstract SKFunction

markwallace-microsoft:
### Contribution Checklist

### DIFF
--- a/docs/decisions/0022-skfunction.md
+++ b/docs/decisions/0022-skfunction.md
@@ -37,7 +37,7 @@ The goal of the ADR is decide if this is the best abstraction to use to meet the
 ### `SKFunction` Case Class
 
 - Good, because the changes to the interface are **not** breaking changes for consumers.
-- Good, because class construct can be made `internal` so we can prevent extensions until we know there are valid use cases.
+- Good, because class constructor can be made `internal` so we can prevent extensions until we know there are valid use cases.
 - Good, because we can change the default implementation easily in future.
 - Bad, because implementations can only extend `SKFunction`.
 
--- a/docs/decisions/0022-skfunction.md
+++ b/docs/decisions/0022-skfunction.md
@@ -1,0 +1,47 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: proposed
+contact: markwallace-microsoft
+date: 2023-11-20
+deciders: SergeyMenshykh, markwallace, rbarreto, mabolan, stephentoub
+consulted: 
+informed: 
+---
+
+# Semantic Kernel Functions are defined using Interface or Abstract Base Class
+
+## Context and Problem Statement
+
+The Semantic Kernel must define an abstraction to represent a Function i.e. a method that can be called as part of an AI orchestration.
+Currently this abstraction is the `ISKFunction` interface.
+The goal of the ADR is decide if this is the best abstraction to use to meet the long term goals of Semantic Kernel.
+
+## Decision Drivers
+
+- The abstraction **must** extensible so that new functionality can be added later.
+- Changes to the abstraction **must not** result in breaking changes for consumers.
+- It is not clear at this time if we need to allow consumers to provide their own `SKFunction` implementations. If we do we this may cause problems as we add new functionality to the Semantic Kernel e.g. what if we define a new hook type?
+
+## Considered Options
+
+- `ISKFunction` interface
+- `SKFunction` base class
+
+### `ISKFunction` Interface
+
+- Good, because implementations can extend any arbitrary class
+- Bad, because we can only change the default behavior of our implementations and customer implementations may become incompatible.
+- Bad, because we cannot prevent customers for implementing this interface.
+- Bad, because changes to the interface are breaking changes for consumers.
+
+### `SKFunction` Case Class
+
+- Good, because the changes to the interface are **not** breaking changes for consumers.
+- Good, because class construct can be made `internal` so we can prevent extensions until we know there are valid use cases.
+- Good, because we can change the default implementation easily in future.
+- Bad, because implementations can only extend `SKFunction`.
+
+## Decision Outcome
+
+Chosen option: "{title of option 1}", because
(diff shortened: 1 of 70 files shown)


dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -90,7 +90,7 @@ private static async Task CompleteChatWithFunctionsAsync(string ask, ChatHistory
         OpenAIFunctionResponse? functionResponse = chatResult.GetOpenAIFunctionResponse();
         if (functionResponse is not null)
         {
-            // If the function returned by OpenAI is an SKFunction registered with the kernel,
+            // If the function returned by OpenAI is an SKFunctionHelper registered with the kernel,
```
stephentoub, line 93: Is the rename here because if SKFunction becomes an abstract base class it needs to be in abstractions which means it can't reference everything necessary for these FromMethod/FromPrompt factories?

We might want to reconsider how valuable the abstractions / core split actually is.

But if we do still like it and need a different name for this, "Helper" doesn't sound right to me.  I'd prefer "Factory".

docs/decisions/0022-skfunction.md
```diff
@@ -0,0 +1,47 @@
...
+- Good, because implementations can extend any arbitrary class
+- Bad, because we can only change the default behavior of our implementations and customer implementations may become incompatible.
+- Bad, because we cannot prevent customers for implementing this interface.
+- Bad, because changes to the interface are breaking changes for consumers.
+
+### `SKFunction` Case Class
+
+- Good, because the changes to the interface are **not** breaking changes for consumers.
+- Good, because class construct can be made `internal` so we can prevent extensions until we know there are valid use cases.
+- Good, because we can change the default implementation easily in future.
```
stephentoub, line 41: We can also have the exposed invoke method be separate from a protected abstract invoke method, which means the base implementation can provide wrapper code for every SKFunction that will be invoked whenever SKFunction.InvokeAsync is called, e.g.
```C#
public abstract class SKFunction
{
    public FunctionResult InvokeAsync(...)
    {
        ... // do whatever, including instrumentation, invoking pre-hooks, etc.
        try
        {
            return InvokeCoreAsync(...);
        }
        finally
        {
            ... // do whatever
        }
    }

    protected abstract FunctionResult InvokeCoreAsync(...);
}
```

docs/decisions/0022-skfunction.md
```diff
@@ -0,0 +1,47 @@
...
+
+- Good, because implementations can extend any arbitrary class
+- Bad, because we can only change the default behavior of our implementations and customer implementations may become incompatible.
+- Bad, because we cannot prevent customers for implementing this interface.
+- Bad, because changes to the interface are breaking changes for consumers.
+
+### `SKFunction` Case Class
+
+- Good, because the changes to the interface are **not** breaking changes for consumers.
+- Good, because class construct can be made `internal` so we can prevent extensions until we know there are valid use cases.
```
stephentoub, line 40: It can... but if one of those use cases is mocking, that _could_ be problematic.

docs/decisions/0022-skfunction.md
```diff
@@ -0,0 +1,47 @@
...
+### `ISKFunction` Interface
+
+- Good, because implementations can extend any arbitrary class
+- Bad, because we can only change the default behavior of our implementations and customer implementations may become incompatible.
+- Bad, because we cannot prevent customers for implementing this interface.
+- Bad, because changes to the interface are breaking changes for consumers.
+
+### `SKFunction` Case Class
+
+- Good, because the changes to the interface are **not** breaking changes for consumers.
```
stephentoub, line 39: Additions to the interface aren't breaking.  Changes to existing APIs could be, depending on what they are. Just clarifying.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
markwallace-microsoft: Agreed naming here needs some thought. If we use `Factory` then we may want to rename the methods.

docs/decisions/0022-skfunction.md
markwallace-microsoft: Agreed, that is a nice additional advantage.

docs/decisions/0022-skfunction.md
markwallace-microsoft: In the last commit I have all of the tests passing. I added an `SKFunctionMock` implementation to the unit tests (which already have access to internals) and made upstream unit test projects depend on the core unit tests.

docs/decisions/0022-skfunction.md
stephentoub: If I've understood correctly, the challenge with the pre-hook is needing to be able to include the rendered prompt.  We should consider splitting the pre-hook into two: one with all the general stuff, and one that's just for surfacing a rendered prompt.  Then this base implementation could handle the pre/post, and the base could expose a protected method for raising the rendered-prompt event.  Having such members be protected means they won't clutter up the surface area / won't be callable by anything other than SKFunction implementations, which is also a nice benefit.

markwallace-microsoft: Yeah, I really like having a hook dedicated to access the rendered prompt. In fact we may want to have pre/post hooks for the render operation so hooks can be used to influence how the prompt gets rendered.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
markwallace-microsoft: Changed to `KernelFunctionFactory`

docs/decisions/0022-skfunction.md
markwallace-microsoft: Will implement this as part of the Hooks refactor

PR #3568, opened 2023-11-20
.Net: ADR to add more planner related telemetry items

TaoChenOSU:
### Motivation and Context

As we pivot to more capable planning features, we need to collect more telemetries for ourselves, and for developers using SK.

Screenshots of the planning telemetry sent to Application Insights by running our telemetry example with log level set to Trace:
image
image
image

Follow up work item will update the telemetry sample to include a ReadMe and show metrics examples. 
### Description

1. An ADR that proposes options to collect more planner telemetries.
2. Implementation of the ADR.

### Contribution Checklist


stephentoub:
@tarekgh, would you mind taking a look at this and sharing any relevant wisdom?

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+5. Contoso wants to be able to see the token usage of a particular planner run.
+6. Contoso wants to be able to see the time taken to create a plan of a particular planner run.
+7. Contoso wants to be able to see the steps in a plan.
+8. Contoso wants to be able to see the inputs&outputs of each plan step.
+9. Contoso wants to change a few settings that may affect the performance of the planners. They would like to know how the performance will be affected before committing the changes.
+10. Contoso wants to update to a new model that is cheaper and faster. They would like to know how the new model performs in planning tasks.
+
+## Out of scope
+
+1. We focus on Application Insights. Although other telemetry service options are supported technically, we will not cover possible ways of setting them up in this ADR.
```
tarekgh, line 33: > We focus on Application Insights. [](http://example.com/codeflow?start=2&length=34)

I am wondering if you talked to Azure observability team about that and get their feedback? I am cc'ing a couple of contacts who will be good to sync up about that. 

@reyang @cijothomas @CodeBlanch @utpilla

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+TBD
+
+## Validation
+
+Tests can be added to make sure that all the expected telemetry items are in place and of the correct format.
+
+## Description the Options
+
+### Function hooks
```
tarekgh, line 65: >Function hooks [](http://example.com/codeflow?start=4&length=14)

Would be better if you can allow enriching the telemetry through the configuration instead? Will be easier to the users to enable more telemetry if needed by just configuring the app.

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+Pros:
+
+1. Everything has to a string.
+2. Developers may have to creation a new container for each activity depending on the service used.
+
+#### Log
+
+```csharp
+this._logger.LogTrace("{PlannerType}: Created plan with details: \n {Plan}", PlannerType, plan.ToPlanString());
```
tarekgh, line 205: may consider using the LoggerMessage attribute which can generate optimized code. I know this is implementation details here :-)

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+// In internal GetTextResults and GetChatResults calls
+using var activity = s_activitySource.StartActivity($"{this.GetType().Name}.GetTextResults");
+if (activity?.GetBaggageItem("Operation") is not null)
+{
+    activity?.AddTag("Operation", activity.GetBaggageItem("Operation"));
+}
+```
+
+In an application that uses SK and Application Insights:
```
tarekgh, line 145: I mentioned before would be good if can sync up with the observability team. I am wondering here is it necessary the app need to manually to listen to the activity events? App Insights can't be enabled to automatically do this listening and collection?

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Considered Options
+
+- Function hooks
+  - Inject logic to functions that will get executed before or after a function is invoked.
+- Distributed tracing
+  - Activity tags & Metrics
+    - Use activities to record individual operations.
+    - Use tags in activities to record custom information.
+    - Use baggage in activities to more easily correlate dependencies.
```
cijothomas, line 49: baggage is probably the wrong thing to use here...

tarekgh:
CC @noahfalk

docs/decisions/0022-planner-telemetry-enhancement.md
cijothomas: My suggestion is to leverage .NET's built-in APIs for logs,metrics,traces, which is also the OpenTelemetry's recommended API. These APIs are vendor-neutral, and final destination of telemetry can be chosen and changed without any re-instrumentation. [ApplicationInsights officially support this](https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-enable?tabs=aspnetcore), but so is almost all other Observability vendors!

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+For more information, please refer to this document: [.Net distributed tracing](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/)
+
+Overall pros:
+
+1. Native .Net support and the kernel has set this up.
+
+Overall cons:
+
+1. Less flexibility for app developers to add custom traces and metrics.
```
cijothomas, line 97: why is that so?

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+In an application that uses SK and Application Insights:
+
+```csharp
+var operations = new ConcurrentDictionary<string, IOperationHolder<DependencyTelemetry>>();
+
+// For more detailed tracing we need to attach Activity entity to Application Insights operation manually.
+void activityStarted(Activity activity)
+{
+  var operation = telemetryClient.StartOperation<DependencyTelemetry>(activity);
```
cijothomas, line 153: strongly recommend against using ApplicationInsights specific APIs. See https://github.com/microsoft/semantic-kernel/pull/3568/files#r1400954697

docs/decisions/0022-planner-telemetry-enhancement.md
cijothomas: OpenTelemetry should be used to do this automatically. And yes, AppInsights supports OpenTelemetry already, so indirectly - this is supported in ApplicationInsights without these manual listeners.

docs/decisions/0022-planner-telemetry-enhancement.md
tarekgh: agree with @cijothomas here. Using the native .NET APIs will allow using ApplicationInsights but not be restricted to it.

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Validation
+
+Tests can be added to make sure that all the expected telemetry items are in place and of the correct format.
+
+## Description the Options
+
+### Function hooks
+
+Function hooks allow developers to inject logic to kernel that will be executed before or after a function is invoked. Example use cases include logging the function input before a function is invoked, and logging results after the function returns.
```
TaoChenOSU, line 67: This option will be nice for devs if they want additional stuff, provided they have the right information in the hooks.

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+ActivitySource.AddActivityListener(activityListener);
+```
+
+The above produces dependency information similar to the screenshot below:
+
+<img src="./images/0022-dependency-in-application-insights-create-plan.png" alt="Example create plan dependency generated in Application Insights" width="400"/>
+
+<img src="./images/0022-dependency-in-application-insights-client-base.png" alt="Example model calling dependency generated in Application Insights" width="400"/>
+
+Cons:
```
TaoChenOSU, line 191: This should be Pros

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+1. Easy to add custom data.
+2. Semi-structured data organized by tags (later saved as custom dimensions for Application Insights).
+3. Automatically correlated from parent to chid items. Baggage makes it even easier for querying.
+
+Pros:
```
TaoChenOSU, line 197: Cons

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+1. Everything has to a string.
+2. Developers may have to creation a new container for each activity depending on the service used.
```
tarekgh, line 200: >Developers may have to creation a new container for each activity depending on the service used. [](http://example.com/codeflow?start=3&length=96)

Can you elaborate more on that?

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+#### Metrics
+
+Pros:
+
+1. Easy to use for continuous monitoring.
+2. Has been set up in SK.
+
+Cons: No additional cons identified given the scope of this document.
+
+#### Activity tags & baggage
```
TaoChenOSU, line 108: Need to consider sampling

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Scenarios
+
+Contoso is a company that is developing an AI application using SK.
+
+1. Contoso needs to continuously monitor the token usage of a particular planner, including prompt tokens, completion tokens, and the total tokens.
+2. Contoso needs to continuously monitor the time it takes for a particular planner to create a plan.
+3. Contoso needs to continuously monitor the success rate of a particular planner in creating a valid plan.
+4. Contoso needs to continuously monitor the success rate of a particular plan type being executed successfully.
+5. Contoso wants to be able to see the token usage of a particular planner run.
```
TaoChenOSU, line 24: Token metrics is being collected at the moment but they are not categorized.

docs/decisions/0022-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+```csharp
+// In Connectors.AI.OpenAI ClientBase.cs
+private static readonly ActivitySource s_activitySource = new(typeof(ClientBase).FullName);
+
+// In internal GetTextResults and GetChatResults calls
+using var activity = s_activitySource.StartActivity($"{this.GetType().Name}.GetTextResults");
+if (activity?.GetBaggageItem("Operation") is not null)
+{
+    activity?.AddTag("Operation", activity.GetBaggageItem("Operation"));
```
TaoChenOSU, line 141: Find out if we can tag the metrics with the operation.

cijothomas (changes requested):
I am not from semantic-kernel team, so feel free to do what is best for this repo. I want to call out that the proposal in this PR is not aligned with our direction.

As one of the maintainers/owners of https://github.com/Microsoft/ApplicationInsights-dotnet and https://github.com/open-telemetry/opentelemetry-dotnet, I strongly recommend to leverage OpenTelemetry, and *do not* code against ApplicationInsights specific APIs. AppInsights will be deprecating its custom API in the future, given the overall direction to use OpenTelemetry. 

Other libraries owned by Microsoft, like Azure SDKs already follow the direction of leveraging OpenTelemetry, and *NOT* tied to ApplicationInsights in anyway.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: Yes, we are planning to leverage the System.Diagnostics APIs. The item here just states that we will only show (in an example) how to configure SK to send telemetry to Application Insights but not all destinations.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: Agreed! We will think about how to do this through just configuration.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: Thanks! This is taken from our current example application. We will update that application as well!

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: Could you elaborate?

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: This is no longer true.
This PR addresses the issue and aligns our sample app to the direction of OpenTelemetry. Will update this statement.

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,218 @@
...
+deciders: alliscode, dmytrostruk, markwallace, SergeyMenshykh, stephentoub
+consulted: {}
+informed: {}
+---
+
+# Planner Telemetry Enhancement
+
+## Context and Problem Statement
+
+It would be extremely beneficial for applications using Semantic Kernel' planning features to be able to continuously monitor the performance of planners and plans.
```
teresaqhoang, line 14:
```suggestion
It would be extremely beneficial for applications using Semantic Kernel's planning features to be able to continuously monitor the performance of planners and plans.
```

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,218 @@
...
+
+Contoso is a company that is developing an AI application using SK.
+
+1. Contoso needs to continuously monitor the token usage of a particular planner, including prompt tokens, completion tokens, and the total tokens.
+2. Contoso needs to continuously monitor the time it takes for a particular planner to create a plan.
+3. Contoso needs to continuously monitor the success rate of a particular planner in creating a valid plan.
+4. Contoso needs to continuously monitor the success rate of a particular plan type being executed successfully.
+5. Contoso wants to be able to see the token usage of a particular planner run.
+6. Contoso wants to be able to see the time taken to create a plan of a particular planner run.
+7. Contoso wants to be able to see the steps in a plan.
```
teresaqhoang, line 26:
```suggestion
7. Contoso wants to be able to see the steps and functions used in a plan.
```

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,218 @@
...
+2. This ADR does not seek to modify the current instrumentation design in SK.
+
+## Decision Drivers
+
+- The framework should be telemetry service agnostic.
+- Enabling and disabling specific telemetry items should require minimum effort.
+
+## Difficulties
+
+- SK currently produces metrics for token usage. However, they are not categorized. Thus, there is no way for developers to know the toke usage used by different operations.
```
teresaqhoang, line 43:
```suggestion
- SK currently produces metrics for token usage. However, they are not categorized. Thus, there is no way for developers to know the token usage used by different operations.
```

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Considered Options
+
+- Function hooks
+  - Inject logic to functions that will get executed before or after a function is invoked.
+- Distributed tracing
+  - Activity tags & Metrics
+    - Use activities to record individual operations.
```
teresaqhoang, line 51: What's considered an operation? Create Plan and Execute Plan or each function execution (i.e., each step is one operation with Execute Plan as a parent operation)?

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,218 @@
...
+- Enabling and disabling specific telemetry items should require minimum effort.
+
+## Difficulties
+
+- SK currently produces metrics for token usage. However, they are not categorized. Thus, there is no way for developers to know the toke usage used by different operations.
+
+## Considered Options
+
+- Function hooks
+  - Inject logic to functions that will get executed before or after a function is invoked.
```
teresaqhoang, line 48: Should specify what kind of data we want to collect with these function hooks (on pre and post execution)

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,218 @@
...
+  activity?.AddTag("Status", "Failed");
+}
+```
+
+```csharp
+// In Connectors.AI.OpenAI ClientBase.cs
+private static readonly ActivitySource s_activitySource = new(typeof(ClientBase).FullName);
+
+// In internal GetTextResults and GetChatResults calls
+using var activity = s_activitySource.StartActivity($"{this.GetType().Name}.GetTextResults");
```
teresaqhoang, line 147: Nit: add `Completion` here to avoid confusion with string results. (Same with `GetChatResults` -> `GetChatCompletionResults` if you're using it elsewhere

```suggestion
using var activity = s_activitySource.StartActivity($"{this.GetType().Name}.GetTextCompletionResults");
```

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,218 @@
...
+1. Easy to use for continuous monitoring.
+2. Has been set up in SK.
+
+Cons: No additional cons identified given the scope of this document.
+
+#### Activity tags & baggage
+
+```csharp
+// In the instrumented planner
+activity?.AddBaggage("Operation", $"{PlannerType}.CreatePlan");
```
teresaqhoang, line 121: Are there unique IDs associated with each operation?

docs/decisions/0022-planner-telemetry-enhancement.md
teresaqhoang: Will developers have the option to override the hook completely or are we exposing a callback that can be optionally called in addition to whatever we implement in the function hooks?

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: Discussed offline.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: No, if we inject some default hooks behind the scene during function registration, I don't think devs can override that.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: No longer considering baggage as explained in the latest version of the ADR.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: Addressed in the latest ADR.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: See the latest ARD.

docs/decisions/0023-planner-telemetry-enhancement.md
TaoChenOSU: It's up to us to define an operation in the code. A SKFunction can be an operation as well as a call to the model, in which case the SKFucntion will be a parent operation to the model call operation.

docs/decisions/0023-planner-telemetry-enhancement.md
TaoChenOSU: What are the differences between steps and functions?

docs/decisions/0023-planner-telemetry-enhancement.md
TaoChenOSU: We are no longer going with this approach. Updated in the latest version.

docs/decisions/0023-planner-telemetry-enhancement.md
TaoChenOSU: We are no longer considering baggage as it will raise security concerns. Updated in the latest version.

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,267 @@
...
+
+- Bottom-up: Propagate token usage information from connectors back to the functions, along with model results.
+- Top-down: Propagate function information down to the connectors, enabling them to tag metric items with function information.
+
+We have decided to implement the bottom-up approach for the following reasons:
+
+1. SK is already configured to propagate token usage information from connectors via `ModelResult`. We simply need to extend the list of items that get propagated, such as model information.
+2. Currently, SK does not have a method for passing function information down to the connector level. Although we considered using [baggage](https://opentelemetry.io/docs/concepts/signals/baggage/#:~:text=In%20OpenTelemetry%2C%20Baggage%20is%20contextual%20information%20that%E2%80%99s%20passed,available%20to%20any%20span%20created%20within%20that%20trace.) as a means of propagating information downward, experts from the OpenTelemetry team advised against this approach due to security concerns.
```
stephentoub, line 119: > Currently, SK does not have a method for passing function information down to the connector level.

Do you mean passing ISKFunctions into connectors? That'll work imminently; passing the kernel into all connector operations is part of the refactoring work @SergeyMenshykh is doing.

docs/decisions/0023-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Function hooks
+
+Function hooks allow developers to inject logic to the kernel that will be executed before or after a function is invoked. Example use cases include logging the function input before a function is invoked, and logging results after the function returns.
+For more information, please refer to the following ADRs:
+
+1. [Kernel Hooks Phase 1](./0005-kernel-hooks-phase1.md)
+2. [Kernel Hooks Phase 2](./0018-kernel-hooks-phase2.md)
+
+We can inject, during function registration, default callbacks to log critical information for all functions.
```
stephentoub, line 159: All functions now derivs from KernelFunction, and we can put arbitrary logic into the base implementation. That's where metrics are now tracked for function duration, errors, etc.

docs/decisions/0023-planner-telemetry-enhancement.md
TaoChenOSU: We believe that's not what we are looking for in this case. At the connector level, we don't have any information about the calling function, even with a kernel instance.

The approach we are leaning towards is to return richer information from connectors to functions.

docs/decisions/0023-planner-telemetry-enhancement.md
TaoChenOSU: This makes it even better! Thanks!

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+```csharp
+/// <summary>
+/// Interface for model results
+/// </summary>
+public interface IResultBase
+{
+  /// <summary>
+  /// Model name or Id
+  /// </summary>
+  string ModelId { get; }
```
SergeyMenshykh, line 132: Every time we add a new member to an interface, we break all its implementations, as they must implement it. For telemetry purposes, we need a more extensible solution for passing information up the stack. Today, we need to track the number of tokens, and tomorrow we may need to track the prompt-response relevance score.

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+
+## To optionally allow the kernel to send sensitive information
+
+In some environments, it is not inappropriate to store sensitive information, such as PII, in any services. Nonetheless, those intermediate results (generated using non-customer data) can be extremely useful for debugging purposes. Thus, we need a switch that can turn on and off telemetries that contain sensitive information.
+
+```csharp
+// In a new InstrumentationOptions.cs
+public sealed class InstrumentationOptions
+{
+  public bool IncludeSensitiveInformation { set; get; } = false;
```
SergeyMenshykh, line 220: I wonder how this property will be used. Do we plan to wrap code for writing logs/traces within if statements throughout the code? Or will it be done only in one location?

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+It would be extremely beneficial for applications using Semantic Kernel's planning features to be able to continuously monitor the performance of planners and plans.
+
+## Scenarios
+
+Contoso is a company that is developing an AI application using SK.
+
+1. Contoso needs to continuously monitor the token usage of a particular planner, including prompt tokens, completion tokens, and the total tokens.
+2. Contoso needs to continuously monitor the time it takes for a particular planner to create a plan.
+3. Contoso needs to continuously monitor the success rate of a particular planner in creating a valid plan.
+4. Contoso needs to continuously monitor the success rate of a particular plan type being executed successfully.
```
markwallace-microsoft, line 23: It will also be important to be able to categorise failures in all cases

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+    - Description: A function can consist of multiple prompts. The execution time of a function is the duration from start to end of a function's `invoke` call.
+    - Dimensions: ComponentType, ComponentName, Service ID, Model ID
+    - Example:
+      | ComponentType | ComponentName | Value |
+      |---|---|---|
+      | Function | WritePoem | 1m
+      | Function | TellJoke | 1m
+      | Function | WriteAndTellJoke | 1.5m
+      | Planner | CreateHandlebarsPlan | 2m
+  - Success count for planners
```
markwallace-microsoft, line 83: Success/failure

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+    - Description: A planner run is considered successful when it generates a valid plan.
+    - Dimensions: ComponentType, ComponentName, Service ID, Model ID
+    - Example:
+      | ComponentType | ComponentName | Fail | Success
+      |---|---|---|---|
+      | Planner | CreateHandlebarsPlan | 5 | 95
+      | Planner | CreateHSequentialPlan | 20 | 80
+  - Success count for plans
```
markwallace-microsoft, line 91: Success/failure

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+      | Function | WriteAndTellJoke | 1.5m
+      | Planner | CreateHandlebarsPlan | 2m
+  - Success count for planners
+    - Description: A planner run is considered successful when it generates a valid plan.
+    - Dimensions: ComponentType, ComponentName, Service ID, Model ID
+    - Example:
+      | ComponentType | ComponentName | Fail | Success
+      |---|---|---|---|
+      | Planner | CreateHandlebarsPlan | 5 | 95
+      | Planner | CreateHSequentialPlan | 20 | 80
```
markwallace-microsoft, line 90: When a planner fails how will we figure out why it failed?

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+    - Example:
+      | ComponentType | ComponentName | Value |
+      |---|---|---|
+      | Function | WritePoem | 1m
+      | Function | TellJoke | 1m
+      | Function | WriteAndTellJoke | 1.5m
+      | Planner | CreateHandlebarsPlan | 2m
+  - Success count for planners
+    - Description: A planner run is considered successful when it generates a valid plan.
+    - Dimensions: ComponentType, ComponentName, Service ID, Model ID
```
markwallace-microsoft, line 85: Is there a Plan name we can include also so we know if it's just specific plans or all plans that are failing

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+    - Example:
+      | ComponentType | ComponentName | Fail | Success
+      |---|---|---|---|
+      | Planner | CreateHandlebarsPlan | 5 | 95
+      | Planner | CreateHSequentialPlan | 20 | 80
+  - Success count for plans
+    - Description: A plan execution is considered successful when all steps in the plan are executed successfully.
+    - Dimensions: ComponentType, ComponentName, Service ID, Model ID
```
markwallace-microsoft, line 93: As above we need to know which plan failed. 

It would also be useful to know which step in the plan failed.

docs/decisions/0024-planner-telemetry-enhancement.md
markwallace-microsoft: Yes, I `Dictionary` would be a more flexible approach and we can add extension methods for well known entries

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+}
+var modelId = result.GetModelId();
+
+if (result.GetModelResults() is null)
+{
+  logger.LogInformation("Model results not found in function result.");
+  return;
+}
+var modelResult = result.GetModelResults().FirstOrDefault();
+var modelResultJson = modelResult?.GetJsonResult();
```
markwallace-microsoft, line 190: Is this an OpenAI specific solution?

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+catch (Exception ex) when (ex is KeyNotFoundException)
+{
+  logger.LogInformation("Usage details not found in model result.");
+  return;
+}
+catch (Exception ex)
+{
+  logger.LogError(ex, "Error while parsing usage details from model result.");
+  throw;
+}
```
markwallace-microsoft, line 209: How are the values being used? This snippet looks incomplete.

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+
+It would be extremely beneficial for applications using Semantic Kernel's planning features to be able to continuously monitor the performance of planners and plans.
+
+## Scenarios
+
+Contoso is a company that is developing an AI application using SK.
+
+1. Contoso needs to continuously monitor the token usage of a particular planner, including prompt tokens, completion tokens, and the total tokens.
+2. Contoso needs to continuously monitor the time it takes for a particular planner to create a plan.
+3. Contoso needs to continuously monitor the success rate of a particular planner in creating a valid plan.
```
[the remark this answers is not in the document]
TaoChenOSU: A plan is valid when the model response can be parsed into a plan of desired format and it contains one or more steps.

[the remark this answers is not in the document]
TaoChenOSU: I will add the criteria to the ADR.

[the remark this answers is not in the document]
TaoChenOSU: It's documented in the tables.

docs/decisions/0024-planner-telemetry-enhancement.md
TaoChenOSU: We will wrap code for writing logs/traces within if statements. Please see an example in the implementation detail.

docs/decisions/0024-planner-telemetry-enhancement.md
TaoChenOSU: Wouldn't that require some predefined error categories/codes?

docs/decisions/0024-planner-telemetry-enhancement.md
TaoChenOSU: By plan name, do you mean an identifier of a specific plan instance of say the Handlebars Plan?

docs/decisions/0024-planner-telemetry-enhancement.md
TaoChenOSU: We need to go into the trace detail to figure out the failure details. Metrics are not good medium to do that.

docs/decisions/0024-planner-telemetry-enhancement.md
TaoChenOSU: More implementation details added.

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+    - Description: A prompt is the smallest unit that consumes tokens. A function can consist of multiple prompts.
+    - Dimensions: ComponentType, ComponentName, Service ID, Model ID
+    - Example:
+      | ComponentType | ComponentName | Service ID | Model ID | Value |
+      |---|---|---|---|---|
+      | Function | WritePoem | | GPT-3.5-Turbo | 0.5
+      | Function | TellJoke | | GPT-4 | 0.5m
+      | Function | WriteAndTellJoke | | GPT-3.5-Turbo | 0.3m
+      | Planner | CreateHandlebarsPlan | | GPT-3.5-Turbo | 1m
+  - Aggregated execution time for functions
```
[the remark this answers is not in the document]
TaoChenOSU: image
It's up to the application to capture the http request dependencies. 

We can add more logs to show the host, method, and status in the connectors.

docs/decisions/0024-planner-telemetry-enhancement.md
TaoChenOSU: We can add a dimension that is the exception type.

docs/decisions/0024-planner-telemetry-enhancement.md
```diff
@@ -0,0 +1,366 @@
...
+2. Distributed tracing is already implemented in SK. We just need to add more telemetry.
+3. Telemetry service agnostic with [OpenTelemetry](https://opentelemetry.io/docs/what-is-opentelemetry/).
+
+Overall cons:
+
+1. Less flexibility for app developers consuming SK as a library to add custom traces and metrics.
+
+#### Logging
+
+Logs will be used to record interesting events while the code is running.
```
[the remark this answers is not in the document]
cijothomas: Yes, using Exemplars : https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/exemplars#using-exemplars-in-opentelemetry-net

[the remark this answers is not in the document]
TaoChenOSU: Tracing (creates logs) and metering (creates metrics) are two operations that are not correlated. If we want to see the token usage of a particular function, we will have to use logs. 

Metrics are more for tracking some value overtime.

docs/decisions/0022-planner-telemetry-enhancement.md
TaoChenOSU: This PR partially addresses this comment for `KenrnelFunction` and `PlannerInstrumentation`: https://github.com/microsoft/semantic-kernel/pull/3859. Other part of the kernel will follow soon.

docs/decisions/0024-planner-telemetry-enhancement.md
TaoChenOSU: Updated the ADR with the option to carry the model id in `ModelResult`.

dotnet/src/InternalUtilities/planning/PlanInstrumentation.cs
```diff
@@ -0,0 +1,108 @@
...
+            name: "sk.planning.run_plan.success",
+            unit: "{invocation}",
+            description: "Measures the number of successful plan executions");
+
+    /// <summary><see cref="Counter{T}"/> to record planner execution failure counts.</summary>
+    private static readonly Counter<int> s_executionFailure =
+        PlanningInstrumentation.Meter.CreateCounter<int>(
+            name: "sk.planning.run_plan.failure",
+            unit: "{invocation}",
+            description: "Measures the number of failed plan executions");
```
stephentoub, line 34: Do we need these? This information should all be available via the s_planExecutionDuration histogram: we can just add an error.type tag when the counter is failure.  This is what we do for function invocation.

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -12,17 +12,26 @@ namespace Microsoft.SemanticKernel.Planning;
...
+            name: "sk.planning.create_plan.success",
+            unit: "{invocation}",
+            description: "Measures the number of successful planner executions");
+
+    /// <summary><see cref="Counter{T}"/> to record planner invocation failure counts.</summary>
+    private static readonly Counter<int> s_invocationFailure =
+        PlanningInstrumentation.Meter.CreateCounter<int>(
+            name: "sk.planning.create_plan.failure",
+            unit: "{invocation}",
+            description: "Measures the number of failed planner executions");
```
stephentoub, line 34: Same comment: we shouldn't need these. All of the relevant information should be available via the single s_createPlanDuration: if a recording is tagged as error.type, then it was a failure, otherwise it was a success.

dotnet/src/InternalUtilities/planning/PlanInstrumentation.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.SemanticKernel.Planning;
+
+/// <summary>Surrounds the invocation of a planner with logging and metrics.</summary>
+internal static partial class PlanInstrumentation
```
stephentoub, line 13: I'd suggest just merging this with the PlannerInstrumentation class.

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlan.cs
```diff
@@ -46,11 +48,28 @@ public override string ToString()
...
-    public string Invoke(
+    public async Task<string> InvokeAsync(
         Kernel kernel,
         KernelArguments arguments,
         CancellationToken cancellationToken = default)
     {
-        return HandlebarsTemplateEngineExtensions.Render(kernel, this._template, arguments, cancellationToken);
+        var logger = kernel.LoggerFactory.CreateLogger(typeof(HandlebarsPlan));
+
+        return await PlanInstrumentation.InvokePlanAsync(
```
stephentoub, line 58:
```suggestion
        return PlanInstrumentation.InvokePlanAsync(
```

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlan.cs
```diff
@@ -46,11 +48,28 @@ public override string ToString()
...
+    private async Task<string> InvokeCoreAsync(
+        Kernel kernel,
+        KernelArguments arguments,
+        CancellationToken cancellationToken = default)
+    {
+        Task<string> task = Task.Run(
+            () => HandlebarsTemplateEngineExtensions.Render(kernel, this._template, arguments, cancellationToken),
+            cancellationToken);
+
+        return await task.ConfigureAwait(false);
```
stephentoub, line 73: This shouldn't use Task.Run.  If the core of this is synchronous, why make everything Task-returning?

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -42,6 +42,21 @@ public sealed class FunctionCallingStepwisePlanner
         Kernel kernel,
         string question,
         CancellationToken cancellationToken = default)
+    {
+        var logger = kernel.LoggerFactory.CreateLogger(this.GetType());
+
+        return await PlanInstrumentation.InvokePlanAsync(
```
stephentoub, line 48: Same comments here: you can remove the async/await/ConfigureAwait(false)

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -228,5 +245,59 @@ private async Task<(ITextGenerationService, string, PromptRenderedEventArgs?)> R
...
+        var completionTokens = 0;
+        try
+        {
+            var jsonObject = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.Serialize(usageObject));
+            promptTokens = jsonObject.GetProperty("PromptTokens").GetInt32();
+            completionTokens = jsonObject.GetProperty("CompletionTokens").GetInt32();
+        }
+        catch (Exception ex) when (ex is KeyNotFoundException)
+        {
+            logger.LogInformation("Usage details not found in model result.");
```
stephentoub, line 281: Shouldn't this return?

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -228,5 +245,59 @@ private async Task<(ITextGenerationService, string, PromptRenderedEventArgs?)> R
...
+        }
+        catch (Exception ex)
+        {
+            logger.LogError(ex, "Error while parsing usage details from model result.");
+            throw;
```
stephentoub, line 286: Shouldn't this return rather than throw? It's bad for diagnostics to actual introduce failure modes.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -228,5 +245,59 @@ private async Task<(ITextGenerationService, string, PromptRenderedEventArgs?)> R
...
+            return;
+        }
+
+        var promptTokens = 0;
+        var completionTokens = 0;
+        try
+        {
+            var jsonObject = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.Serialize(usageObject));
+            promptTokens = jsonObject.GetProperty("PromptTokens").GetInt32();
+            completionTokens = jsonObject.GetProperty("CompletionTokens").GetInt32();
```
stephentoub, line 277: Can we avoid roundtripping through JSON to get at this information? It adds a lot of overhead. And relying on exceptions if it can't parse makes it super expensive if we're getting back different objects.

dotnet/src/InternalUtilities/planning/PlanInstrumentation.cs
TaoChenOSU: You're right. We can filter the measurements in Azure monitor. Removed.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
TaoChenOSU: Unfortunately, the way the token information is returned would require us to perform the round trip. The object that contains the token usage is of type `Azure.AI.OpenAI.CompletionsUsage`. We don't want to take a dependency on OpenAI at the function level, thus we have to rely on Json.

TaoChenOSU: I have changed it to use `SerializeToElement`, which simplifies the code but I believe it performs the same operation underneath.

dotnet/src/InternalUtilities/planning/PlanInstrumentation.cs
TaoChenOSU: Done

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -232,5 +249,65 @@ private async Task<(ITextGenerationService, string, PromptRenderedEventArgs?)> R
...
+        }
+
+        var promptTokens = 0;
+        var completionTokens = 0;
+#pragma warning disable CA1031 // Do not catch general exception types
+        try
+        {
+            var jsonObject = JsonSerializer.SerializeToElement(usageObject);
+            promptTokens = jsonObject.GetProperty("PromptTokens").GetInt32();
+            completionTokens = jsonObject.GetProperty("CompletionTokens").GetInt32();
```
stephentoub, line 290: How about TryGetProperty and TryGetInt32 rather than relying on exceptions?

alliscode (approved):

dotnet/samples/KernelSyntaxExamples/Example48_GroundednessChecks.cs
```diff
@@ -146,7 +146,7 @@ public static async Task PlanningWithGroundednessAsync()
         Console.WriteLine($"======== Goal: ========\n{ask}");
         Console.WriteLine($"======== Plan ========\n{plan}");
 
-        var result = plan.Invoke(kernel, new KernelArguments(), CancellationToken.None);
+        var result = await plan.InvokeAsync(kernel, new KernelArguments(), CancellationToken.None);
```
glahaye, line 149: This doesn't look like changes you made but rather, changes that got in when you merged to main...

Something seems wrong that this appears in your PR...

TaoChenOSU: No, this is something I changed. The plan invoke function is now async.

glahaye (approved):

Issue #3544, opened 2023-11-17
.Net: ParameterView has too many forms of "type"


stephentoub:
ParameterView now has _three_ properties that each expose their own distinct notion of type information:
- the .NET type
- a JSON data type
- a JSON schema

The middle one was added a few months ago, and the last one was just added last week.  Now that the last one exists, we shouldn't also need the middle one.

Assuming that's correct, it should be removed prior to shipping 1.0.

cc: @alliscode, @markwallace-microsoft

PR #3576, opened 2023-11-21
.Net: Remove SKParameterMetadata.Type and ParameterJsonType

stephentoub:
It's now duplicative of information in the schema.
Fixes https://github.com/microsoft/semantic-kernel/issues/3544


dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsTemplateEngineExtensions.cs
```diff
@@ -378,14 +358,20 @@ private static double CastToNumber(object? number)
      * Handlebar argument type is an object (this covers complex types).
      * Function parameter is a generic type.
      */
-    private static bool IsExpectedParameterType(string functionViewType, string handlebarArgumentType, object handlebarArgValue)
+    private static bool IsExpectedParameterType(SKParameterMetadata parameterType, object argument)
```
stephentoub, line 361: I made a best guess on this code, but I'm not sure if/how the previous code was actually correct, and it doesn't look like tests are exercising any of this.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKFunctionMetadataExtensions.cs
```diff
@@ -12,34 +13,51 @@ public static class SKFunctionMetadataExtensions
...
+            SKJsonSchema? schema = param.Schema ?? OpenAIFunction.GetJsonSchema(param.ParameterType, param.Description);
+
+            // Read the "type" property from the schema, if it exists.
+            // If it doesn't, that means we lack any type information for the property.
+            string? type = null;
+            if (schema?.RootElement.TryGetProperty("type", out JsonElement prop) == true)
+            {
+                type = prop.GetString();
+            }
+            type ??= "string";
```
alliscode, line 34: Is `type` used for anything?

alliscode (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/SKFunctionMetadataExtensions.cs
stephentoub: Hah, it was, and then I subsequently deleted OpenAIFunctionParameter.Type, at which point this is dead.

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsTemplateEngineExtensions.cs
markwallace-microsoft: I will follow up on test coverage for the Handlebars planner

markwallace-microsoft (approved):

PR #3584, opened 2023-11-21
.Net Remove KernelResult class

SergeyMenshykh:
### Motivation and Context
Taking into account that the Kernel.RunAsync methods for running pipeline(sequence of functions) are being moved to samples the KernelResult class becomes redundant. 

### Description
1. The KernelResult class has been removed.
2. The Kernel.RunAsync methods are refactored to return FunctionResult instead.
3. The Kernel.RunAsync methods that execute pipeline are refactored to return result of the last function in the pipeline.

### Contribution Checklist


dotnet/src/SemanticKernel.Core/Planning/Plan.cs
```diff
@@ -400,15 +400,13 @@ internal string ExpandFromVariables(ContextVariables variables, string input)
 
             // Execute the step
             var result = await kernel.RunAsync(step, functionVariables, cancellationToken).ConfigureAwait(false);
-
-            var stepResult = result.FunctionResults.FirstOrDefault();
-            if (stepResult is null)
+            if (result is null)
             {
                 // Step was cancelled
                 return null;
```
RogerBarreto, line 406: If a function was cancelled or skipped it render's no result.

@matthewbolanos @SergeyMenshykh If we stop using Skip we actually get the benefit of not having the current problem where a plan step don't know when to skip.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
```diff
@@ -855,7 +855,7 @@ void FunctionInvoked(object? sender, FunctionInvokedEventArgs e)
         var result = await sut.RunAsync("PlanInput", plan);
 
         // Assert
-        Assert.NotNull(result);
+        Assert.Null(result);
```
stephentoub, line 858: This looks strange. Why the flip?

stephentoub (approved):

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsTemplateEngineExtensions.cs
```diff
@@ -143,7 +143,7 @@ internal sealed class HandlebarsTemplateEngineExtensions
             ISKFunction function = kernel.Plugins.GetFunction(functionView.PluginName, functionView.Name);
 
 #pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
-            KernelResult result = kernel.RunAsync(executionContext.Variables, cancellationToken, function).GetAwaiter().GetResult();
+            FunctionResult result = kernel.RunAsync(executionContext.Variables, cancellationToken, function).GetAwaiter().GetResult();
```
RogerBarreto, line 146: Can we promote this static to be an async to avoid this diable TSTHRD002?

dotnet/src/SemanticKernel.UnitTests/KernelTests.cs
```diff
@@ -65,7 +65,7 @@ public async Task RunAsyncRunsWhenNotCancelledAsync()
         using CancellationTokenSource cts = new();
 
         // Act
-        KernelResult result = await kernel.RunAsync(cts.Token, kernel.Plugins.GetFunction("mySk", "GetAnyValue"));
+        FunctionResult result = await kernel.RunAsync(cts.Token, kernel.Plugins.GetFunction("mySk", "GetAnyValue"));
```
RogerBarreto, line 68:
```suggestion
        var result = await kernel.RunAsync(cts.Token, kernel.Plugins.GetFunction("mySk", "GetAnyValue"));
```

RogerBarreto (approved):
Nit comments, LGTM.

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
RogerBarreto: KernelResults would be returned with 0 functions.

When the function is the result, I agree don't make sense to return a FunctionResult without result...

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Core/Planning/Plan.cs
SergeyMenshykh: Yep, will be done separately in the PR related to removing the Skip action.

dotnet/src/Planners/Planners.Core/Handlebars/HandlebarsTemplateEngineExtensions.cs
SergeyMenshykh: Nope, today it's impossible - https://github.com/microsoft/semantic-kernel/pull/3392#discussion_r1386317690

dotnet/src/SemanticKernel.UnitTests/Planning/PlanTests.cs
SergeyMenshykh: Agreed offline to keep it this way taking into account that the method is going to be moved to samples. However, it should be fixed in Function.InvokeAsync in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3586

RogerBarreto (approved):

PR #3594, opened 2023-11-21
.Net: SKContext.Culture property removed

SergeyMenshykh:
In this PR:
1. The Kernel.Data property bag added to store arbitrary/ambient data on kernel.
2. SKContext.Culture property is removed to prepare SKContext for removal.
3. Code required culture is refactored to use the one associated with kernel.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -120,6 +117,11 @@ public sealed class Kernel
         throw new SKException($"Service of type {typeof(T)} and name {name ?? "<NONE>"} not registered.");
     }
 
+    /// <summary>
+    /// Dictionary for arbitrary/ambient data associated with the kernel.
+    /// </summary>
+    public IDictionary<string, object> Data { get; } = new Dictionary<string, object>();
```
stephentoub, line 123: The `object` value should be nullable, e.g. `object?`. Also, I don't know how frequently this will be used... might make sense to lazily initialize it (in a thread-safe way), e.g.
```C#
private Dictionary<string, object?>? _data;

public IDictionary<string, object?> Data =>
    _data ??
    Interlocked.CompareExchange(ref _data, new Dictionary<string, object?>(), null) ??
    _data;
```

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
```diff
@@ -74,25 +61,6 @@ public SKContext Clone(ContextVariables? variables)
...
-            this.Culture);
-    }
-
-    /// <summary>
-    /// The culture currently associated with this context.
-    /// </summary>
-    private CultureInfo _culture;
-
-    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
-    private string DebuggerDisplay
```
stephentoub, line 87: If you're deleting DebuggerDisplay, you should also delete the `[DebuggerDisplay("{DebuggerDisplay,nq}")]` on the class.

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
```diff
@@ -363,6 +364,42 @@ public static ISKPlugin ImportPluginFromObject(this Kernel kernel, object target
...
+    /// Associate a culture with the kernel.
+    /// </summary>
+    /// <param name="kernel">The kernel.</param>
+    /// <param name="culture">The culture.</param>
+    public static void SetCulture(this Kernel kernel, CultureInfo culture)
+    {
+        Verify.NotNull(kernel);
+        Verify.NotNull(culture);
+
+        kernel.Data[KernelCultureDataKey] = culture;
```
stephentoub, line 381: I don't think we should store this in the dictionary. Having reserved keys like this that anyone can stomp over (accidentally or on purpose) just leads to problems.  We can just make this a private field.

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
```diff
@@ -363,6 +364,42 @@ public static ISKPlugin ImportPluginFromObject(this Kernel kernel, object target
...
+    #region Kernel.Data helper methods
+
+    private const string KernelCultureDataKey = "kernel.culture";
+
+    /// <summary>
+    /// Associate a culture with the kernel.
+    /// </summary>
+    /// <param name="kernel">The kernel.</param>
+    /// <param name="culture">The culture.</param>
+    public static void SetCulture(this Kernel kernel, CultureInfo culture)
```
stephentoub, line 376: Why are these extension methods?  Why not just have a `public CultureInfo Culture { get; set; }` on Kernel?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
SergeyMenshykh: Right, fixed.

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
SergeyMenshykh: make sense.

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
SergeyMenshykh: Just modeled it through the Data dictionary. The Culture class property is a better alternative.

dotnet/src/SemanticKernel.Abstractions/Orchestration/SKContext.cs
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
```diff
@@ -46,7 +46,7 @@ public Kernel Build()
         );
 #pragma warning restore CS8604 // Possible null reference argument.
 
-        instance.SetCulture(this._culture ?? instance.GetCulture());
+        instance.Culture = this._culture ?? instance.Culture;
```
stephentoub, line 49: Is the `?? instance.Culture` part necessary?

stephentoub (approved):

RogerBarreto (approved):

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
SergeyMenshykh: nope, not really, just wanted to avoid if operator and not making Kernel.Culture property nullable.

SergeyMenshykh: addressed with if operator

PR #3639, opened 2023-11-22
.Net: Make several members of Kernel lazily-initialized

stephentoub:
Also clean up some of the XML comments.


stephentoub:
Batched into https://github.com/microsoft/semantic-kernel/pull/3641

PR #3641, opened 2023-11-23
.Net: Several Kernel-related improvements / fixes

stephentoub:
Batching these into one PR due to possible conflicts between them (and Thanksgiving absences that will slow things down with such conflicts)...

**Commit 1:** Make several members of Kernel lazily-initialized and clean up its XML comments.

**Commit 2:** Remove ISKPluginCollection. This interface served simply to unify `ICollection<ISKPlugin>` and `IReadOnlySKPluginCollection`. However, this leads to problems due to members with the same name on both interfaces, e.g. if you have:
```C#
ISKPluginCollection c = ...;
int count = c.Count;
```
that will fail to compile due to an ambiguity error, with the compiler not knowing which of the base interfaces that comes from.  The fix is to remove `ISKPluginCollection` and just have `SKPluginCollection` implement both interfaces directly, similar to how `List<T>` implements both `IList<T>` and `IReadOnlyList<T>`. This then means that `Kernel.Plugins` is an `SKPluginCollection` rather than an interface. General philosophy now:
- If you only care about being able to enumerate plugins, just use `IEnumerable<ISKPlugin>`
- If you only care about reading but want to be able to query rather than just enumerate, use `IReadOnlySKPluginCollection`.
- If you want the full set of mutable capabilities, use `SKPluginCollection`.
 
**Commit 3:** Add Kernel.Clone


dotnet/src/Experimental/Assistants/AssistantBuilder.cs
```diff
@@ -155,7 +155,10 @@ public AssistantBuilder WithPlugin(ISKPlugin plugin)
     /// <returns><see cref="AssistantBuilder"/> instance for fluid expression.</returns>
     public AssistantBuilder WithPlugins(IEnumerable<ISKPlugin> plugins)
     {
-        this._plugins.AddRange(plugins);
+        foreach (ISKPlugin plugin in plugins)
+        {
+            this._plugins.Add(plugin);
+        }
```
markwallace-microsoft, line 161: Nit: Can we not use `SKPluginCollection.AddRange` here?

markwallace-microsoft (approved):
Just one minor nit that can be fixed later

SergeyMenshykh (approved):

dotnet/src/Experimental/Assistants/AssistantBuilder.cs
stephentoub: > Can we not use SKPluginCollection.AddRange here?

We can. I'd replaced this when I removed the extension method and neglected to add it back when I put the method onto the concrete type.

PR #3644, opened 2023-11-23
.Net: Remove event handler wrapper from SKContext

markwallace-microsoft:
### Contribution Checklist


dotnet/src/SemanticKernel.Core/Planning/Plan.cs
```diff
@@ -308,7 +312,7 @@ protected override SKFunctionMetadata GetMetadataCore()
                 // Return the last result state of the plan.
                 if (stepResult is null)
```
RogerBarreto, line 313: With your change I believe this code won't be hit. Since we return the functionresult for cancellations and skip, we can now in the Plan code, check for the new properties you added to the functionResult.

If a PlanStep result was cancelled, break the loop (Trumpts over Skip)
If a PlanStep result was skipped, continue (skip)

RogerBarreto (approved):

dotnet/src/SemanticKernel.Core/Planning/Plan.cs
markwallace-microsoft: I need to change this line https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Core/KernelExtensions.cs#L536 to ensure we always return a `FunctionResult`, right now this causes a null return for the one function cancelled or skipped case. I have a task to remove the code to run a pipeline so I'll have this change as part of that work.

SergeyMenshykh (approved):

PR #3647, opened 2023-11-23
.Net: Remove support for passing multiple functions to Kernel.RunAsync

markwallace-microsoft:
### Motivation and Context

- `Kernel.RunAsync` will only support running a single function
- Plans can be use dto run multiple functions if needed or
- Developers can create their own strategy to do this

### Contribution Checklist


stephentoub:
> Kernel.RunAsync will only support running a single function

Are you thinking this is temporary until the hooks support is moved into KernelFunction.InvokeAsync?

Or are you thinking we'd ship with this? Once the hooks support is removed, and like this would just be a one-liner whose sole purpose was to swap the order of arguments.

SergeyMenshykh (approved):

markwallace-microsoft:
> > Kernel.RunAsync will only support running a single function
> 
> Are you thinking this is temporary until the hooks support is moved into KernelFunction.InvokeAsync?
> 
> Or are you thinking we'd ship with this? Once the hooks support is removed, seems like this would just be a one-liner whose sole purpose was to swap the order of arguments?

The intent is to ship without this specific functionality.

We will provide a sample demonstrating how to implement a custom pipeline.

stephentoub:
Ok, so the single function support is just temporary?

(If we were to keep it, I'd want to rename it to InvokeAsync, to match both InvokeAsync on a function and InvokePromptAsync.)

markwallace-microsoft:
> Ok, so the single function support is just temporary?
> 
> (If we were to keep it, I'd want to rename it to InvokeAsync, to match both InvokeAsync on a function and InvokePromptAsync.)

No single function support will be the permanent. My understanding is that we wanted to simplify things so that the Kernel extension methods could be used as a convenience to run a single kernel function.

Renaming so we only have one name for this makes sense. That avoids the question as to what the difference between the two methods is.

stephentoub:
> Renaming so we only have one name for this makes sense. That avoids the question as to what the difference between the two methods is.

Yeah, and functionally we'll want to keep them identical, with the kernel method just delegating to the function method.

It also doesn't need to be an extension anymore; it can just be an instance method on kernel.

markwallace-microsoft:
> > Renaming so we only have one name for this makes sense. That avoids the question as to what the difference between the two methods is.
> 
> Yeah, and functionally we'll want to keep them identical, with the kernel method just delegating to the function method.
> 
> It also doesn't need to be an extension anymore; it can just be an instance method on kernel.

I'm going to make this change in a seperate PR as there are a lot of files to be changed and it will be easier to code review on it's own.

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
```diff
@@ -364,108 +363,68 @@ public static ISKPlugin ImportPluginFromObject(this Kernel kernel, object target
...
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <param name="pipeline">List of functions</param>
-    /// <returns>Result of the function composition</returns>
+    /// <returns>Result of the function</returns>
     public static Task<FunctionResult> RunAsync(
         this Kernel kernel,
-        CancellationToken cancellationToken,
-        params KernelFunction[] pipeline)
+        KernelFunction function,
+        ContextVariables variables,
```
RogerBarreto, line 419: This will allow all to use 
RunAsync(function)
RunAsync(function, variables)
RunAsync(function, variables, cancellationToken)

```suggestion
        ContextVariables? variables = null,
```

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
```diff
@@ -364,108 +363,68 @@ public static ISKPlugin ImportPluginFromObject(this Kernel kernel, object target
...
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>Result of the function</returns>
     public static Task<FunctionResult> RunAsync(
         this Kernel kernel,
         KernelFunction function,
-        ContextVariables? variables = null,
         CancellationToken cancellationToken = default)
     {
         Verify.NotNull(kernel);
+        Verify.NotNull(function);
```
RogerBarreto, line 380: This one is not needed, see my comment below.

RogerBarreto (approved):
Small suggestions

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
markwallace-microsoft: The code already supports these

```
RunAsync(function)
RunAsync(function, input)
RunAsync(function, input, cancellationToken)
RunAsync(function, variables)
RunAsync(function, variables, cancellationToken)
```

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -62,31 +40,31 @@ public static async Task RunAsync()
...
-        await kernel.RunAsync(variables, kernel.Plugins["test"]["type06"]);
+        await kernel.RunAsync(kernel.Plugins["test"]["type06"], variables);
 
         await kernel.RunAsync(plugin["type07"], variables);
-        await kernel.RunAsync(variables, kernel.Plugins["test"]["type07"]);
+        await kernel.RunAsync(kernel.Plugins["test"]["type07"], variables);
 
-        await kernel.RunAsync("", plugin["type08"]);
-        await kernel.RunAsync("", kernel.Plugins["test"]["type08"]);
+        await kernel.RunAsync(plugin["type08"], "");
```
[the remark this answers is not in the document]
markwallace-microsoft: Fixed

RogerBarreto (approved):

PR #3648, opened 2023-11-23
.Net: Move code to invoke pre and post hooks down to KernelFunction

markwallace-microsoft:
### Contribution Checklist


dotnet/src/SemanticKernel.Core/Planning/Plan.cs
```diff
@@ -292,28 +291,19 @@ protected override SKFunctionMetadata GetMetadataCore()
...
             while (this.HasNextStep)
             {
                 AddVariablesToContext(this.State, context);
 
                 var stepResult = await this.InternalInvokeNextStepAsync(kernel, context, cancellationToken).ConfigureAwait(false);
 
-                if (stepResult.IsCancellationRequested)
+                // If a step was cancelled before invocation
+                // Return the last result state of the plan.
+                if (stepResult is null || stepResult.IsCancellationRequested)
```
RogerBarreto, line 303: What is the scenario where stepResult is null?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -63,11 +64,14 @@ public abstract class KernelFunction
     /// <param name="name">Name of the function.</param>
     /// <param name="description">Function description.</param>
     /// <param name="modelSettings">Model request settings.</param>
-    internal KernelFunction(string name, string description, IEnumerable<AIRequestSettings>? modelSettings = null)
+    /// <param name="invokeEventHandlers">Flad to control whether or not event handlers are invoked</param>
+    internal KernelFunction(string name, string description, IEnumerable<AIRequestSettings>? modelSettings = null, bool invokeEventHandlers = true)
```
RogerBarreto, line 68: Why `inovkeEventHandlers` flag?

IMO all functions should invoke...

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -87,19 +91,58 @@ internal KernelFunction(string name, string description, IEnumerable<AIRequestSe
...
+        }
 
         cancellationToken.ThrowIfCancellationRequested();
 
         TagList tags = new() { { "sk.function.name", this.Name } };
         long startingTimestamp = Stopwatch.GetTimestamp();
         try
         {
+            // Invoke pre hook, and stop if skipping requested.
+            if (this._invokeEventHandlers)
```
RogerBarreto, line 106: Avoid the extra ifs removing the flag.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -212,6 +222,28 @@ internal bool OnFunctionInvoked(FunctionInvokedEventArgs eventArgs)
         }
         return handled;
     }
+
+    internal bool OnPromptRendering(PromptRenderingEventArgs eventArgs)
+    {
+        bool handled = false;
+        if (this.PromptRendering != null)
```
RogerBarreto, line 229: nit, apply to other places.
```suggestion
        if (this.PromptRendering is not null)
```

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -212,6 +222,28 @@ internal bool OnFunctionInvoked(FunctionInvokedEventArgs eventArgs)
...
+    internal bool OnPromptRendered(PromptRenderedEventArgs eventArgs)
+    {
+        bool handled = false;
+        if (this.PromptRendered != null)
+        {
+            this.PromptRendered.Invoke(this, eventArgs);
+            handled = true;
+        }
+        return handled;
+    }
```
RogerBarreto, line 246: bool return is not being used, this method can be simplified to
```suggestion
    internal void OnPromptRendered(PromptRenderedEventArgs eventArgs)
    {
        this.PromptRendered?.Invoke(this, eventArgs);
    }
```

Same for `OnPromptRendering`

RogerBarreto (approved):
Some suggestions, 

This change make to code a lot simpler, very good.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
markwallace-microsoft: There is because Plan wraps steps (which are KernelFunctions) in Plans (which are KernelFunctions) and I have to avoid triggering the hooks twice. The proper fix would be to change Plan to had a collection of Plans | KernelFunctions but that's too involved and the future of Plan is uncertain.

dotnet/src/SemanticKernel.Abstractions/Events/PromptRenderingEventArgs.cs
```diff
@@ -0,0 +1,28 @@
...
+/// </summary>
+public class PromptRenderingEventArgs : SKEventArgs
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PromptRenderingEventArgs"/> class.
+    /// </summary>
+    /// <param name="metadata">Function metadata</param>
+    /// <param name="context">Context related to the event</param>
+    /// <param name="requestSettings">request settings used by the AI service</param>
+    public PromptRenderingEventArgs(SKFunctionMetadata metadata, SKContext context, AIRequestSettings? requestSettings) : base(metadata, context)
```
stephentoub, line 19: Is there a reason for all these events that we pass the metadata rather than the function itself?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
stephentoub: > and I have to avoid triggering the hooks twice

Why? Wouldn't we want any function invocation to trigger the events whether it wraps other functions or not?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -87,19 +91,58 @@ internal KernelFunction(string name, string description, IEnumerable<AIRequestSe
         using var activity = s_activitySource.StartActivity(this.Name);
         ILogger logger = kernel.LoggerFactory.CreateLogger(this.Name);
 
-        logger.LogInformation("Function invoking.");
+        if (logger.IsEnabled(LogLevel.Trace))
+        {
+            logger.LogTrace("Function {Name} invoking.", this.Name);
```
stephentoub, line 96: Isn't the function name here baked into the logger? Why does the name also need to be part of the message?

dotnet/src/SemanticKernel.Core/Functions/SKFunctionFromPrompt.cs
```diff
@@ -237,16 +234,11 @@ private void AddDefaultValues(ContextVariables variables)
...
+    private PromptRenderingEventArgs CallPromptRendering(Kernel kernel, SKContext context, AIRequestSettings? requestSettings)
     {
-        var eventArgs = new FunctionInvokingEventArgs(this.GetMetadata(), context)
-        {
-            Metadata = {
-                [SKEventArgsExtensions.RenderedPromptMetadataKey] = renderedPrompt
-            }
-        };
-        kernel.OnFunctionInvoking(eventArgs);
+        var eventArgs = new PromptRenderingEventArgs(this.GetMetadata(), context, requestSettings);
```
stephentoub, line 240: It'd be better to only allocate the event args if there's a handler to invoke.

stephentoub (commented):
can review in more depth when I'm not on my phone

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
markwallace-microsoft: In this case I don't think we should be wrapping the individual steps. I'll undo the change and update the tests the account for the extra events.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
markwallace-microsoft: Removed this

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
markwallace-microsoft: @RogerBarreto I used the bool return value so that I know when to create a new `FunctionResult` as I don't wnat to do that unless needed.

```c#
private (FunctionInvokedEventArgs, FunctionResult) CallFunctionInvoked(Kernel kernel, SKContext context, FunctionResult result)
{
    var eventArgs = new FunctionInvokedEventArgs(this.GetMetadata(), result);
    if (kernel.OnFunctionInvoked(eventArgs))
    {
        // Apply any changes from the event handlers to final result.
        result = new FunctionResult(this.Name, eventArgs.SKContext, eventArgs.SKContext.Variables.Input)
        {
            // Updates the eventArgs metadata during invoked handler execution
            // will reflect in the result metadata
            Metadata = eventArgs.Metadata
        };
    }

    return (eventArgs, result);
}
```

markwallace-microsoft: Changed to the pattern Stephen suggested

dotnet/src/SemanticKernel.Core/Planning/Plan.cs
markwallace-microsoft: This is no longer needed will remove it

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Events/PromptRenderedEventArgs.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <param name="renderedPrompt">Rendered prompt</param>
+    public PromptRenderedEventArgs(KernelFunction function, SKContext context, string renderedPrompt) : base(function, context)
+    {
+        this.RenderedPrompt = renderedPrompt;
+    }
+
+    /// <summary>
+    /// Rendered prompt.
+    /// </summary>
+    public string RenderedPrompt { get; set; }
```
stephentoub, line 26: This needs a setter?

dotnet/src/SemanticKernel.Abstractions/Events/SKEventArgs.cs
```diff
@@ -14,22 +14,27 @@ public abstract class SKEventArgs : EventArgs
...
+    /// <summary>
+    /// Kernel function
+    /// </summary>
+    public KernelFunction Function { get; }
+
     /// <summary>
     /// Function metadata
     /// </summary>
-    public SKFunctionMetadata FunctionMetadata { get; }
+    public SKFunctionMetadata FunctionMetadata => this.Function.GetMetadata();
```
stephentoub, line 37: Is this still needed? A caller that needs it can just do e.Function.GetMetadata().

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -87,21 +88,53 @@ internal KernelFunction(string name, string description, IEnumerable<AIRequestSe
         using var activity = s_activitySource.StartActivity(this.Name);
         ILogger logger = kernel.LoggerFactory.CreateLogger(this.Name);
 
-        logger.LogInformation("Function invoking.");
+        if (logger.IsEnabled(LogLevel.Trace))
```
stephentoub, line 91: This is ok. I've personally only done the check when there's additional work being done at the call site, like passing arguments that'll result in a params array allocation.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -191,26 +202,52 @@ public CultureInfo Culture
...
+        }
+
+        var eventArgs = new FunctionInvokingEventArgs(function, context);
+        this.FunctionInvoking.Invoke(this, eventArgs);
+        return eventArgs;
+    }
+
+    internal FunctionInvokedEventArgs? OnFunctionInvoked(KernelFunction function, FunctionResult result)
+    {
+        if (this.FunctionInvoked is null)
```
stephentoub, line 219: There's a race condition here: if code concurrently unsubscribes their handler and it's the last one, the invocation could null ref, as you'll check and see it's not null but then it'll become null after you check. That's handled for you with the ?. syntax, as it makes a local copy before the null check. Wanting to return the event args complicates things, though. So you'll want to store it into a local first and then do all operations (null check and invocation) with that local.

stephentoub (approved):

dotnet/src/SemanticKernel.Abstractions/Events/PromptRenderedEventArgs.cs
stephentoub: Got it.

stephentoub: What happens if a hook sets it to null or whitespace?

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -144,14 +168,22 @@ internal KernelFunction(string name, string description, IEnumerable<AIRequestSe
 
         logger.LogInformation("Function streaming invoking.");
 
-        cancellationToken.ThrowIfCancellationRequested();
+        // Invoke pre hook, and stop if skipping requested.
```
RogerBarreto, line 171: Thanks for adding this, I added in my current UT branch, will use yours when merging.

RogerBarreto (commented):
Merged already.

PR #3655, opened 2023-11-24
.Net: Eliminate the MS.SK.TemplateEngine namespace

markwallace-microsoft:
### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/PromptTemplate/IPromptTemplate.cs
```diff
@@ -5,7 +5,9 @@
 using System.Threading.Tasks;
 using Microsoft.SemanticKernel.Orchestration;
 
-namespace Microsoft.SemanticKernel.TemplateEngine;
+#pragma warning disable IDE0130 // Namespace does not match folder structure
```
stephentoub, line 8: This analyzer is doing more harm than good for how the solution is structured... seems like the majority of files are having to disable it. How about we just turn it off all up?

stephentoub (approved):

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/IPromptTemplate.cs
markwallace-microsoft: Good suggestion, I'll handle this separately

SergeyMenshykh (approved):

PR #3671, opened 2023-11-24
.Net: KernelFunctionArguments class has been added

SergeyMenshykh:
### Motivation and Context
The class will be used to represent arguments for both Kernel.InvokeAsync and KernelFunction.InvokeAsync methods. 

- Signature of the KernelFunction.InvokeAsync methods will change from the current one:
_InvokeAsync(Kernel kernel, **SKContext context, AIRequestSettings? requestSettings,** CancellationToken cancellationToken)_
to the new one:
_InvokeAsync(Kernel kernel, **KernelFunctionArguments arguments**, CancellationToken cancellationToken)_

- Similarly, signature of the Kernel.InvokeAsync methods will change from the current one:
_InvokeAsync(this Kernel kernel, KernelFunction function, **ContextVariables variables**, CancellationToken cancellationToken)_ 
to the new one:
_InvokeAsync(this Kernel kernel, KernelFunction function, **KernelFunctionArguments arguments**,  CancellationToken cancellationToken)_ 

Currently, SK works with string arguments 99% of the time. In the future, the argument value type will change from string to object? in scope of the complex-type support feature.
 


dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
```diff
@@ -0,0 +1,28 @@
...
+/// <summary>
+/// Represents arguments of <see cref="KernelFunction.InvokeAsync"/> and "Kernel.InvokeAsync" methods.
+/// </summary>
+public sealed class KernelFunctionArguments : Dictionary<string, string>
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="KernelFunctionArguments"/> class.
+    /// </summary>
+    /// <param name="requestSettings">The AI request settings.</param>
+    public KernelFunctionArguments(AIRequestSettings? requestSettings = null)
```
stephentoub, line 19: This should initialize the base with OrdinalIgnoreCase, right?  Or are we switching to make argument matching case-sensitive?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma warning disable IDE0130
+namespace Microsoft.SemanticKernel;
+#pragma warning restore IDE0130
+
+/// <summary>
+/// Represents arguments of <see cref="KernelFunction.InvokeAsync"/> and "Kernel.InvokeAsync" methods.
+/// </summary>
+public sealed class KernelFunctionArguments : Dictionary<string, string>
+{
```
stephentoub, line 14: I suggest there should also be a constructor that accepts a KernelFunctionArguments and another that accepts an `IDictionary<>`.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
RogerBarreto, line 1: QQ, do we have plans in doing this a Dictionary<string, object> in a future iteration?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
SergeyMenshykh: Right, completely forgot about this aspect. Fixed.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
SergeyMenshykh: I had both in my POC but wanted to add them along the way if they are needed. Anyway, they are added.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
SergeyMenshykh: Yep, we have plans to support complex types, so we'll change the value type to - Dictionary<string, object?>

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// <param name="other">The <see cref="KernelFunctionArguments"/> whose AI request setting and elements are copied to the new <see cref="KernelFunctionArguments"/>.</param>
+    public KernelFunctionArguments(KernelFunctionArguments other) : this(other as IDictionary<string, string>)
+    {
+        this.RequestSettings = other.RequestSettings;
+    }
+
+    /// <summary>
+    /// Gets or sets the AI request settings.
+    /// </summary>
+    public AIRequestSettings? RequestSettings { get; set; }
```
markwallace-microsoft, line 46: `public PromptExecutionSettings? ExecutionSettings { get; set; }`

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// Initializes a new instance of the <see cref="KernelFunctionArguments"/> class that contains AI request settings and elements copied from the specified <see cref="KernelFunctionArguments"/>
+    /// </summary>
+    /// <param name="other">The <see cref="KernelFunctionArguments"/> whose AI request setting and elements are copied to the new <see cref="KernelFunctionArguments"/>.</param>
+    public KernelFunctionArguments(KernelFunctionArguments other) : this(other as IDictionary<string, string>)
+    {
+        this.RequestSettings = other.RequestSettings;
+    }
+
+    /// <summary>
+    /// Gets or sets the AI request settings.
```
markwallace-microsoft, line 44: Gets or sets the prompt execution settings

markwallace-microsoft (commented):
There is a rename needed

markwallace-microsoft (approved):

PR #3673, opened 2023-11-24
.Net: Remove dependency on planners.core

markwallace-microsoft:
### Motivation and Context

New planners must not depend on planners core so they can be taken to v1.0 independently

### Description

Move the shared code to internal utilities

### Contribution Checklist


dotnet/src/SemanticKernel.Core/Planning/PlannerConfigBase.cs
RogerBarreto, line 1: Shouldn't this file keep in Planners.Core ?

dotnet/src/SemanticKernel.Core/Planning/PlannerInstrumentation.cs
RogerBarreto, line 1: Same for this?

RogerBarreto (approved):
LGTM, Not sure if we should keep any Plan file in Kernel.Core.

dotnet/src/SemanticKernel.Core/Planning/PlannerConfigBase.cs
markwallace-microsoft: Moving to the internal utilities as it's shared with the openai planner

dotnet/src/SemanticKernel.Core/Planning/PlannerInstrumentation.cs
markwallace-microsoft: Moving to the internal utilities as it's shared with the openai planner

SergeyMenshykh (approved):

Issue #3571, opened 2023-11-20
.Net: Revise / remove SK's custom IDelegatingHandlerFactory / reliability libraries


stephentoub:
SK currently has its own IDelegatingHandlerFactory type, along with public implementations like NullHttpHandlerFactory. It also has public types like PollyHttpRetryHandlerFactory for implementing retries. And extension methods on KernelBuilder like WithHttpHandlerFactory, which is confusingly close to the existing HttpClientFactory.

Can we get rid of all of this, get rid of the Microsoft.SemanticKernel.Reliability.Basic library, get rid of the Microsoft.SemanticKernel.Reliability.Polly library, and simply rely on existing Microsoft.Extensions.* libraries (or other means) for getting an HttpClient that includes retry support?

PR #3676, opened 2023-11-25
.Net: Remove SK's custom resilience mechanisms

stephentoub:
Better to just use established conventions. This removes SK's custom IDelegatingHandlerFactory and all implementations of it, including the two "Basic" and Polly-based libraries. It replaces WithHttpHandlerFactory (which wasn't related to either HttpClientFactory or HttpClientHandler) with a WithHttpClient method that puts a specific HttpClient into the Kernel and then makes that available to any services that want it. The OpenAIClent's built-in retry support is now only suppressed if an HttpClient is supplied.

(My aim is that after this PR, we're able to replace the custom service provider in Kernel with an actual IServiceProvider and then remove the custom support for things like ILoggingFactory and HttpClient with just having those resolved from that IServiceProvider. But for now, WithHttpClient is a step forward from what we had.)

Fixes https://github.com/microsoft/semantic-kernel/issues/3571


markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -71,14 +70,14 @@ public CultureInfo Culture
...
     /// </summary>
     /// <remarks>
-    /// This is typically only used as part of creating plugins and functions, as that is typically
-    /// when such clients are constructed.
+    /// This is typically only used as part of creating services, plugins, and functions, as that is typically
+    /// when such clients are constructed. This instance may be ignored if an <see cref="HttpClient"/> was provided
+    /// to the relevant component directly or via other means.
     /// </remarks>
-    public IDelegatingHandlerFactory HttpHandlerFactory { get; }
+    public HttpClient? HttpClient { get; }
```
RogerBarreto, line 80: What happens when we have different services with different clients? Shoudn't the HttpClient be provided at the service configuration not in the Kernel?

RogerBarreto (approved):
One question... 
A handler we can inject behavior in httpclients from services, but providing a client not sure how that will work as we might have multiple services using different clients.

stephentoub:
> One question... A handler we can inject behavior in httpclients from services, but providing a client not sure how that will work as we might have multiple services using different clients.

Can you elaborate? What's an example where passing around an HttpMessageHandler works but passing around an HttpClient doesn't? All of the SK services today that let you customize the transport accept an HttpClient.

RogerBarreto:
I don't have a scenario for now, but let's say I have a multi model project where I have hugging face models and openai models, if I specify a client with my baseAddress (for a hugging face API) this will break the OpenAI connector, or am I missing something?

stephentoub:
> if I specify a client with my baseAddress (for a hugging face API) this will break the OpenAI connector, or am I missing something?

That's true, but there's no base address to even specify if the handler is what's part of the kernel. So is the concern here:
- whether the supplied HttpClient is configured inappropriately for some use?
- whether it's an HttpClient or an HttpMessageHandler?
- whether either of them is here at all?
- something else?

I have a subsequent set of changes I'm working on and hoping to get up a PR for today that will basically just make Kernel around an arbitrary IServiceProvider, such that it can contain any services.  At that point, this dedicated method would just go away.

stephentoub:
Superceded by https://github.com/microsoft/semantic-kernel/pull/3714

PR #3677, opened 2023-11-25
.Net: Add Update Assistant

sultanlive:
### Motivation and Context

1. Add Update Assistant method


crickman:
Thanks for this contribution.  Now that the Thanksgiving holiday is over, we are taking second pass at this API.

I originally excluded the modify/update functionality to avoid rationizing how a definition change to assistant affects the chat-lifecycle (across a variety of cases).

I will aim to consider this over the next week in coordination with our product manager.

crickman (changes requested):
Thanks for this contribution. Now that the Thanksgiving holiday is over, we are taking second pass at this API.

I originally excluded the modify/update functionality to avoid rationizing how a definition change to assistant affects the chat-lifecycle (across a variety of cases).

I will aim to consider this over the next week in coordination with our product manager.

crickman:
@sultanlive - You may have noticed that this initial version was built around a hand-built rest client.  Moving forward, we want to invest in a general agent pattern and re-implement the assistant version of "agent" using the Azure `AssistantClient`. 

- https://github.com/microsoft/semantic-kernel/pull/5245
- https://techcommunity.microsoft.com/t5/ai-azure-ai-services-blog/azure-openai-service-announces-assistants-api-new-models-for/ba-p/4049940

As such, we have basically frozen this current experimental version.  While we are taking critical fixes to this code, extending life-cycle contracts around Agent definition might be best addressed within our current investment.

PR #3678, opened 2023-11-25
.Net: Move Plan to the ms.sk.planners.core package

markwallace-microsoft:
### Motivation and Context

- Change Plan so it is no longer a KernelFunction because we want to keep constructor for KernelFunction internal
- Move Plan (and associated classes and tests) to planners core

### Contribution Checklist


stephentoub:
Is the motivation for this because KernelFunction's ctor is internal and Plan is moving to an assembly that doesn't have friend access?  Is that motivation to reconsider having the ctor be internal? Or we're still confident in that decision for now?

dotnet/src/Planners/Planners.Core/Plan.cs
```diff
@@ -371,9 +412,9 @@ private async Task<FunctionResult> InternalInvokeNextStepAsync(Kernel kernel, Co
             var functionVariables = this.GetNextStepVariables(variables, step);
 
             // Execute the step
-            var result = await kernel.InvokeAsync(step, functionVariables, cancellationToken).ConfigureAwait(false);
+            var result = await step.InvokeAsync(kernel, functionVariables, null, cancellationToken).ConfigureAwait(false);
 
-            var resultValue = result.Variables.Input.Trim();
+            var resultValue = result.GetValue<string>()?.Trim();
```
SergeyMenshykh, line 417: The line will throw the `InvalidCastException` for values of non-string type. For example, if the function returned integer - 23 the GetValue method will throw - System.InvalidCastException: 'Cannot cast System.Int32 to System.String'

markwallace-microsoft: Will change so that the code does something that is functionally equivalent.

dotnet/src/Planners/Planners.Core/KernelPlanExtensions.cs
```diff
@@ -15,6 +15,45 @@ namespace Microsoft.SemanticKernel;
 /// </summary>
 public static class KernelPlanExtensions
 {
+    /*
```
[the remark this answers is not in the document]
markwallace-microsoft: Removed

stephentoub:
>  I think composition more accurately reflects the relationship between a Plan and a KernelFunction.

Ok. We do expect functions to compose other functions, but if logically a plan isn't one, then changing the inheritance makes sense.

SergeyMenshykh (approved):

matthewbolanos (approved):

PR #3680, opened 2023-11-25
.Net: Rename SK prefix to Kernel

markwallace-microsoft:
### Contribution Checklist


SergeyMenshykh (approved):

matthewbolanos (approved):

stephentoub (approved):
I spot checked. LGTM.

Two thoughts:
- I think it should be KernelNameAttribute rather than KernelFunctiomNameAttribute.
- I think we should consider removing Parameter from KernelParameterJsonSchema.

Issue #3682, opened 2023-11-26
.Net: Gpt-3.5-turbo and other LLMs do not follow prompt breaking planer


MaxAkbar:
**Describe the bug**
When using other language models and gpt-3.5-turbo, the planner breaks due to invalid XML. This occurs because the prompt is not being followed, resulting in text describing the XML. Here is an example response from gpt-3.5-turbo:

Here is the code being executed (The code example is from [11-Planner](https://github.com/MicrosoftDocs/semantic-kernel-docs/tree/main/samples/dotnet/11-Planner)).
When the code is executed against gpt-3.5-turbo it generates the following response. The code throws an exception as the regex can't parse the response into valid XML. The following is the response from gpt-3.5-turbo.

```
To create a plan to satisfy the given goal, we can follow these steps:

1. Start by creating a <plan> element.
2. Since the goal involves calculating the increase in investment, we can use the MathPlugin.Multiply function to calculate the increase amount.
3. We need to multiply the investment amount by the percentage increase. To do this, we can use the MathPlugin.Multiply function with the inputs being the investment amount and the percentage increase.
4. Save the result of the multiplication using the setContextVariable attribute.
5. Next, we need to subtract the amount spent on the latte from the increased investment amount. We can use the MathPlugin.Subtract function with the inputs being the increased investment amount and the amount spent on the latte.
6. Save the result of the subtraction using the setContextVariable attribute.
7. Finally, append the result of the subtraction to the plan result using the appendToResult attribute.

Here is the XML plan to satisfy the given goal:

<plan>
    <!-- Calculate the increase amount -->
    <function.MathPlugin.Multiply input="2130.23" number2="1.23" setContextVariable="INCREASE_AMOUNT"/>
    <!-- Subtract the amount spent on the latte -->
    <function.MathPlugin.Subtract input="$INCREASE_AMOUNT" number2="5" setContextVariable="FINAL_AMOUNT" appendToResult="RESULT__FINAL_AMOUNT"/>
</plan>
```

**To Reproduce**

1. Run the following code against gpt-3.5-turbo:
```
using System.Text.Json;
using Microsoft.Extensions.Logging;
using Microsoft.SemanticKernel;
using Microsoft.SemanticKernel.Planning;

using ILoggerFactory loggerFactory = LoggerFactory.Create(builder =>
{
    builder
        .SetMinimumLevel(0)
        .AddDebug();
});

// Create kernel
Kernel kernel = new KernelBuilder()
[...]

PR #3683, opened 2023-11-26
.Net: Update the regex to extract the xml when additional text also exists.…

MaxAkbar:
Update the unit test to match the correct exception. This allows gpt-3.5-turbo response of the following to be extracted:

```
To create a plan to satisfy the given goal, we can follow these steps:

1. Start by creating a <plan> element.
2. Since the goal involves calculating the increase in investment, we can use the MathPlugin.Multiply function to calculate the increase amount.
3. We need to multiply the investment amount by the percentage increase. To do this, we can use the MathPlugin.Multiply function with the inputs being the investment amount and the percentage increase.
4. Save the result of the multiplication using the setContextVariable attribute.
5. Next, we need to subtract the amount spent on the latte from the increased investment amount. We can use the MathPlugin.Subtract function with the inputs being the increased investment amount and the amount spent on the latte.
6. Save the result of the subtraction using the setContextVariable attribute.
7. Finally, append the result of the subtraction to the plan result using the appendToResult attribute.

Here is the XML plan to satisfy the given goal:

<plan>
    <!-- Calculate the increase amount -->
    <function.MathPlugin.Multiply input="2130.23" number2="1.23" setContextVariable="INCREASE_AMOUNT"/>
    <!-- Subtract the amount spent on the latte -->
    <function.MathPlugin.Subtract input="$INCREASE_AMOUNT" number2="5" setContextVariable="FINAL_AMOUNT" appendToResult="RESULT__FINAL_AMOUNT"/>
</plan>
```

### Motivation and Context
When using gpt-3.5-turbo with the planner, the prompt is not followed, and an exception is raised. This fixes the issue.
As noted in [.Net: Gpt-3.5-turbo and other LLMs do not follow prompt breaking planer](https://github.com/microsoft/semantic-kernel/issues/3682).

### Description

Updated the regex to account for additional text in the chat completion response prior to planner execution.

### Contribution Checklist


markwallace-microsoft:
@MaxAkbar Thanks for your contribution. We have put the older planners on hold and introduced two new planners.

1. [Handlebars Planner](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs)
2. [Stepwise Function Calling Planner](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs)

Can you try these instead.

I'm going to close this PR as we are not currently making changes to the old planners.

PR #3709, opened 2023-11-27
.Net: Add experimental attributes to packages

matthewbolanos:
### Motivation and Context

Adding the experimental attribute to packages and classes in the .NET SDK of Semantic Kernel to mark things that won't be ready be v1.0.0 of Semantic Kernel.

### Contribution Checklist


markwallace-microsoft (approved):

dotnet/docs/EXPERIMENTS.md
```diff
@@ -0,0 +1,62 @@
+# Experiments
```
stephentoub, line 1: Nice summary.

dotnet/src/SemanticKernel.Core/Memory/SemanticTextMemory.cs
```diff
@@ -33,6 +35,7 @@ public sealed class SemanticTextMemory : ISemanticTextMemory
     }
 
     /// <inheritdoc/>
+    [Experimental("SKEXP0003")]
```
stephentoub, line 38: Marking the type as experimental should result in everything on it being considered experimental as well. You shouldn't need to mark the individual methods when you've also annotated the type.

matthewbolanos: Yeah... I think I went overboard on that one. Should be fixed now. I did, however, need to add it to any of the static methods.

SergeyMenshykh (approved):

terrajobst:
> @terrajobst, how fine-grained would you recommend different experimental diagnostic ids be? Does this look reasonable?

In general, my recommendation is to have one diagnostic per experimental feature such that a user has a single suppression per experimental feature. It looks like the diagnostic IDs were based on area. If the entire area is experimental that is probably fine.

However you'll want to think about how you're going to evolve this moving forward. For example, in V1 all of "image services" (`SKEXP0002`) is going to be experimental. In V2 you decide to ship it stable. In V3 you're adding a new Doodad feature that is part of image services. Are you going to use `SKEXP0002` for Doodad as well? What are you going to do for cases where a new feature is cross cutting and needs supporting APIs across multiple areas? In this case, I think you'll want a net-new diagnostic ID for the Doodad feature.

Does this make sense?

matthewbolanos:
> Does this make sense?

@terrajobst, yes it does. Based on my understanding of what you said, once we make an area non-experimental, we likely will retire the ID of the area. New IDs should then be created for each feature that can ship independently.

markwallace-microsoft (approved):

Issue #3659, opened 2023-11-24
Investigate replacing custom service container


stephentoub:
SK has its own limited services container. It should really be using IServiceProvider, and for keys, the new IKeyedServiceProvider. It could still have its own implementation of these interfaces, or it could use the built-in implementation in M.E.DependencyInjection (now or in the future). This would mean Kernel would no longer need to special-case one-off services, like ILoggerFactory; they'd be queried from the provider just like with the AI services.

Issue #3191, opened 2023-10-15
.Net: Does IKernel / Kernel need to store an IPromptTemplateEngine?


stephentoub:
There are 0 calls to KernelBuilder.WithPromptTemplateEngine anywhere in the code base... including tests.

The only direct use of Kernel's ctor that accepts a prompt template engine is in a test that then completely ignores the constructed kernel.

Kernel itself does not use the PromptTemplateEngine (though its constructor apparently thinks it's so important it stores it into the same field twice).

PromptTemplate has a ctor `public PromptTemplate(string template, PromptTemplateConfig promptTemplateConfig, IKernel kernel)` that's used 0 times in the whole codebase, and all it does anyway is just delegate to the ctor that takes an IPromptTemplateEngine instead of an IKernel.

In the entire repo there are only 5 call sites that use `.PromptTemplateEngine`, 1 of which is the above ctor and 2 of which are tests of RegisterSemanticFunction. The remaining 2 uses are in the CreateSemanticFunction and ImportSemanticFunctionsFromDirectory extension methods, which take an IKernel but could easily accept an IPromptTemplateEngine as well.

What's the benefit of having this top-level in IKernel?  Why can't it be just another entry in the ServiceCollection?

I'm looking at IKernel from the perspective of trying to understand what it really is. And once you look past the cruft, it seems like it's really just:
- a holder for services, which is suspect: https://github.com/microsoft/semantic-kernel/issues/3151#issuecomment-1763229144
- a holder for functions, for which there's already a dedicated FunctionsCollection
- a method for invoking functions

and I'm wondering if maybe IKernel/Kernel as types should go away: code can just pass around a ServiceCollection and a FunctionCollection as needed, and have a static RunAsync method somewhere that's purely about the invocation.

SergeyMenshykh:
That's exactly what I thought about all the time. Here's the same idea from our latest design review meting:
![image](https://github.com/microsoft/semantic-kernel/assets/68852919/5cc44db9-10d1-4040-a420-d2355d1f6ba4)

Issue #2463, opened 2023-08-15
Extend IKernel.GetService<> to non AI services


yan-li:
Today IKernel.GetService<> only supports AI services. 
```
/// <summary>
/// Get one of the configured services. Currently limited to AI services.
/// </summary>
/// <param name="name">Optional name. If the name is not provided, returns the default T available</param>
/// <typeparam name="T">Service type</typeparam>
/// <returns>Instance of T</returns>
T GetService<T>(string? name = null) where T : IAIService;
```

Extending it as well as registration process would make skill authoring more friendly to derive dependencies from IKernel.

yan-li:
Hi @nacharya1 here are some pseudo code for expected dependency injection experience. 

Current
```
IAPIClient client = ...
var skillA = new ASkill(client)
var skillB = new BSkill(client)
kernel.ImportSkills(...)
```

Expected
```
IAPIClient client = ...
kernel.RegisterService<IAPIClient>(client)
var skillA = new ASkill(kernel)
var skillB = new BSkill(kernel)
kernel.ImportSkills(...)

public class ASkill
{
  IAPIClient client;
  public ASkill(IKernel kernel)
  {
     this.client = kernel.GetService<IAPIClient>();
  }
}
```

The implementation could be extending current GetService interface on IKernel so devs could have similar experience as other dotnet apps.

Also for reference: https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection

matthewbolanos:
@stephentoub, I believe your most recent PR addresses this.

stephentoub:
> @stephentoub, I believe your most recent PR addresses this.

yup

PR #3714, opened 2023-11-27
.Net: Replace SK's IAIServiceProvider with IServiceProvider

stephentoub:
Closes https://github.com/microsoft/semantic-kernel/issues/3694
Closes https://github.com/microsoft/semantic-kernel/issues/3693
Closes https://github.com/microsoft/semantic-kernel/issues/3659
Closes https://github.com/microsoft/semantic-kernel/issues/3571
Closes https://github.com/microsoft/semantic-kernel/issues/3191
Closes https://github.com/microsoft/semantic-kernel/issues/2463

(This replaces https://github.com/microsoft/semantic-kernel/pull/3676, including that as a first commit. That resiliency functionality is no longer necessary after this change, as developers can simply use libraries like Microsoft.Extensions.Http.Resilience to achieve the desired support.)

SK has had a custom IAIServiceProvider interface that's represented a provider of AI related services. Kernel has then wrapped one of these, but also wrapped other individual non-AI services, like ILoggerFactory and HttpClient.  This leads to a variety of problems, such as it being more complicated to construct a Kernel instance, it not integrating as nicely with a broader dependency injection system, and not being as future proof because there isn't a good way to flow additional non-AI-related services through the Kernel to anywhere that might want access. On the DI front, it's also more difficult to integrate because all of the KernelBuilder extension methods are specific to KernelBuilder, and even though they're creating services, the helpers aren't usable with IServiceCollection and thus with the rest of the DI system; a developer needs to go directly to the underlying types and register those directly.

There are also a handful of features that are of little value and in some cases exposed but not actually implemented, such as setAsDefault and alsoAsTextCompletion.  And other cases where there's something valuable but missing, such as the ability to create certain kinds of Azure-related service instances with an existing OpenAIClient.

This PR attempts to address all of these issues and set up SK to evolve more easily:
- Kernel now wraps an IServiceProvider, and all services (AI and non-AI) are retrieved from it. This has a variety of benefits, including future-proofing for wanting to flow more state around, better integration with other systems (e.g. Microsoft.Extensions.DependencyInjection), simpler construction, etc.
- KernelBuilder still has dedicated WithXx methods for first-class features, but they're all layered on top of a new ConfigureServices method, which lets you add anything to an IServiceCollection. Because everything is based on IServiceCollection/IServiceProvider, dependencies are automatically satisfied from the container. For example, for services that accept an optional HttpClient, if one isn't specified to the WithXx method, the system will pull one out of the DI container if it can, which means all existing support for adding configured HttpClient instances to an IServiceCollection immediately apply to Kernel{Builder} as well.
- KernelBuilder also exposes a ConfigurePlugins method, largely for parity with ConfigureServices, as fundmantally a Kernel is just the combination of a collection of services and a collection of plugins (plus some additional transient data, event handlers, etc.) This also integrates with IServiceProvider, with one passed into the callback, so that if the plugin itself needs access to something like logging, it can simply get it from the container.  These ConfigureServices and ConfigurePlugins methods can be called any number of times.
- KernelBuilder is now just a simplified, opinionated way to achieve something you can achieve in two other ways: construct the Kernel directly, or construct the Kernel via DI.  For the latter, you can simply add Kernel as a DI resource, and it'll automatically be composed out of what's available. For example, if in ASP.NET you AddAzureOpenChatCompletion, AddLogging, AddHttp, etc., and then ask for a Kernel to be injected, that Kernel will be constructed from the IServiceProvider that contains all of those other resources, and thus anyone using the Kernel will get that logger, will get that HttpClient, etc.; the HttpClient will also be used by the Azure OpenAI chat completion.

@matthewbolanos, @markwallace-microsoft, @SergeyMenshykh, please review from an SK perspective to let me know if you agree with the direction, whether there are specific things you'd like to see changed, etc. There are some straggling things we should still do after this (like add a few more extension methods for adding plugins to a plugin collection, adding more tests, revamping XML comments, cleaning up more code, etc.), but this is a sweeping change and I wanted to get it in.

@eerhardt, @halter73, I'd appreciate it if you could review this from a DI perspective, e.g. are there variations in the patterns I should be following, any best-practices I'm violating, whether I'm using singleton when I should be using transient, etc. etc.

Note that I ran into one issue as part of this: the Microsoft.Extensions.DependencyInjection support for keyed services doesn't let you enumerate all services for a particular T, regardless of key. That's something SK depends on. To workaround that, KernelBuilder tracks all of the type/keys and injects a dictionary as a service with that information; then places the Kernel needs it, it can query for that dictionary and use it.  This means that a Kernel constructed with KernelBuilder will fully support keys, whereas a Kernel constructed with DI or directly won't work as nicely with named services. My hope is that M.E.DI can fix this asap (offline conversation) in a way that will allow SK to upgrade to a patched version, delete its hack, and have those broader scenarios "just work".  In the meantime, though, KernelBuilder behaves as desired.

I also stopped short of adding Add{Azure}OpenAIXx methods that resolve the OpenAIClient from the service provider. That'll be an important thing to do once there's an Aspire component for Azure.AI.OpenAI, but in the meantime, I didn't want there to be an attractive Add{Azure}OpenAIXx method that looked like you didn't need to supply the necessary information even though you do.


dotnet/src/SemanticKernel.Core/SemanticKernel.Core.csproj
```diff
@@ -26,6 +26,11 @@
     <InternalsVisibleTo Include="SemanticKernel.Extensions.UnitTests" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
```
eerhardt, line 30: Do you have to reference `Microsoft.Extensions.DependencyInjection`? It would be best if you could just reference the abstractions.

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
```diff
@@ -55,110 +90,84 @@ public Kernel Build()
...
-    /// Add a logger to the kernel to be built.
+    /// Configures the services to be available through the <see cref="Kernel"/>.
     /// </summary>
-    /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
-    /// <returns>Updated kernel builder including the logger.</returns>
-    public KernelBuilder WithLoggerFactory(ILoggerFactory loggerFactory)
+    /// <param name="configureServices">Callback invoked as part of this call. It's passed the service collection to manipulate.</param>
+    /// <returns>This <see cref="KernelBuilder"/> instance.</returns>
+    /// <remarks>The callback will be invoked synchronously as part of the call to <see cref="ConfigureServices"/>.</remarks>
+    public KernelBuilder ConfigureServices(Action<IServiceCollection> configureServices)
```
eerhardt, line 98: Why `ConfigureServices` instead of just exposing a `Services` collection property?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -99,155 +152,148 @@ public CultureInfo Culture
...
+        T? service = null;
 
-        this.ServiceProvider = aiServiceProvider;
-        this._plugins = plugins is not null ? new KernelPluginCollection(plugins) : null;
-        this._serviceSelector = serviceSelector;
-        this.HttpHandlerFactory = httpHandlerFactory ?? NullHttpHandlerFactory.Instance;
-        this.LoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
+        if (serviceId is not null)
+        {
+            if (this.Services is IKeyedServiceProvider)
```
eerhardt, line 179: Should we throw a better exception if a `serviceId` is provided, but the service provider isn't an `IKeyedServiceProvider`? Right now it would just say `"Service of type '{typeof(T)}' and key '{serviceId}' not registered."`

dotnet/src/SemanticKernel.Core/SemanticKernel.Core.csproj
stephentoub: Yes, that's where BuildServiceProvider lives.

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
stephentoub: A combination of a) that seemed to be the pattern employed across the various M.E libraries and ASP.NET, and b) it keeps the APIs on KernelBuilder fluent for KernelBuilder.  You think it should change?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
stephentoub: I can tweak the message for that case.

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
```diff
@@ -1,50 +1,87 @@
...
+            foreach (ServiceDescriptor serviceDescriptor in services)
+            {
+                if (!typeToKeyMappings.TryGetValue(serviceDescriptor.ServiceType, out List<object?>? keys))
+                {
+                    typeToKeyMappings[serviceDescriptor.ServiceType] = keys = new();
+                }
+
+                keys.Add(serviceDescriptor.ServiceKey);
+            }
+            services.AddKeyedSingleton(Kernel.ServiceTypeToKeyMappingsKey, typeToKeyMappings);
```
markwallace-microsoft, line 68: Could we do this every time `Kernel.GetAllServices` is called?

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -99,155 +152,148 @@ public CultureInfo Culture
...
-        };
+            // M.E.DI doesn't support querying for a service without a key, and it also doesn't
+            // support AnyKey currently: https://github.com/dotnet/runtime/issues/91466
+            // As a workaround, KernelBuilder injects a service containing the type-to-all-keys
+            // mapping. We can query for that service and and then use it to try to get a service.
+            if (this.Services.GetKeyedService<Dictionary<Type, List<object?>>>(ServiceTypeToKeyMappingsKey) is { } typeToKeyMappings &&
+                typeToKeyMappings.TryGetValue(typeof(T), out List<object?> keys))
+            {
+                return keys.Select(key => this.Services.GetKeyedService<T>(key)).Where(s => s is not null)!;
+            }
```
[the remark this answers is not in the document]
stephentoub: > Are these correct

Yes

> are there any other limitations?

Not that I'm aware of. And the limitation should go away once this absorbs https://github.com/dotnet/runtime/pull/95308. 

> No additional services added to the Kernel after it was built

That's generally not an issue. IServiceProvider typically represents a closed set of services: you add a bunch to a service collection and then build a service provider from what's there.

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
stephentoub: Why?

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
```diff
@@ -39,7 +40,7 @@ public static class KernelExtensions
     {
         Verify.NotNull(kernel);
 
-        return KernelFunctionFactory.CreateFromMethod(method.Method, method.Target, functionName, description, parameters, returnParameter, kernel.LoggerFactory);
+        return KernelFunctionFactory.CreateFromMethod(method.Method, method.Target, functionName, description, parameters, returnParameter, kernel.GetService<ILoggerFactory>());
```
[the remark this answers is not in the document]
stephentoub: We could if you'd like. I was aiming to keep custom things on Kernel to a minimum, but if you want to promote it back to being in the public surface area, we can.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
markwallace-microsoft: Answered above

dotnet/src/SemanticKernel.Core/SemanticKernel.Core.csproj
```diff
@@ -26,6 +26,11 @@
...
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
```
eerhardt, line 31:
```suggestion
```

You don't need both. This comes transitively both through `Microsoft.Extensions.DependencyInjection` and `SemanticKernel.Abstractions.csproj`.

dotnet/src/SemanticKernel.Core/KernelBuilder.cs
eerhardt: > that seemed to be the pattern employed across the various M.E libraries and ASP.NET

The past couple releases have gone away with some of that. See https://learn.microsoft.com/en-us/aspnet/core/migration/50-to-60-samples?view=aspnetcore-8.0#add-services.

> b) it keeps the APIs on KernelBuilder fluent for KernelBuilder

This point makes sense, if you want to keep it fluent.

> You think it should change?

I'm not sure. The part that I always found confusing when doing these `ConfigureXXX(action)` methods is "when does my action run? Even here between `ConfigureServices` and `ConfigurePlugins`, the action runs at different times - services runs inline and plugins don't run until `Build` is called.

Following the .NET 6+ ASP.NET / Hosting pattern, this could look more like:

```csharp
class KernelBuilder
{
    IServiceCollection Services { get; }
    PluginBuilder Plugins { get; }
}
```

The usage wouldn't be fluent anymore. And we'd need a special method to handle the `ConfigurePlugins(Action<IServiceProvider, ICollection<IKernelPlugin>> configurePlugins)` case. That one has to be deferred for sure.

stephentoub: Thanks, @eerhardt. I'll leave it as ConfigureServices/Plugins for this PR.  One the one hand, I like the simplicity of just .Services/.Plugins. On the other hand, for these use cases I think there's some value in the fluency, in particular when just appending .Build(), we'd still need something special for the ConfigurePlugins (as you say), and there's something to be said for keeping the services stuff a bit more buried.  If we keep them, I'm also wondering if we should name them WithServices and WithPlugins.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

SergeyMenshykh (approved):

PR #3722, opened 2023-11-28
.Net: Throw InvalidOperationException from GetResult when FunctionResult cancelled or skipped

### LINKED ISSUE #3586: .Net: SKFunction cancellation
Considering that the pre/post hook functionality will be implemented by the SKFunction class, it is essential to ensure a suitable model for handling function execution cancellation scenarios.

The function caller requires a method to determine if the function was cancelled due to a pre/post hook. This can be represented as an IsCancelled property of the FunctionResult class, or the function can throw an OperationCancelledException.

If we choose to represent it using the OperationCancelledException and need to differentiate between natural cancellations, such as a timeout, and user-initiated cancellations, we can consider using a custom exception called SKFunctionCancelledException.

We should avoid modeling the cancellation by simply returning null from InvokeAsync or returning an instance of the FunctionResult class with a null value, as this would not allow for distinguishing between cancellations and cases where the function naturally returns null.

CC: @markwallace-microsoft @RogerBarreto @stephentoub

markwallace-microsoft:
Closes #3586

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -81,6 +81,16 @@ public FunctionResult(string functionName, ContextVariables variables, object? v
     /// <exception cref="InvalidCastException">Thrown when it's not possible to cast result value to <typeparamref name="T"/>.</exception>
     public T? GetValue<T>()
     {
+        if (this.IsCancellationRequested)
+        {
+            throw new InvalidOperationException("Cannot get result value from cancelled function.");
```
stephentoub, line 86: OperationCanceledException?

stephentoub: > OperationCanceledException is for the act of cancellation, if this happens after the cancellation make sense to be an InvalidOperationException

This feels similar to me as async/await.  If you do:
```C#
Task t = SomethingAsync();
...
var result = await t;
```
and SomethingAsync is canceled, the cancellation exception propagates out when you go to get the result.

SergeyMenshykh (approved):

PR #3723, opened 2023-11-28
Python: Upgrade to Pydantic 2+

eavanvalkenburg:
### Motivation and Context

Upgraded pydantic to 2.0+, much cleaner, less complexity in different types of base models.

### Contribution Checklist


python/tests/unit/test_serialization.py
```diff
@@ -270,16 +271,15 @@ def test_usage_as_required_field(
...
 
 def assert_serializable(obj: _Serializable, obj_type) -> None:
-    """Assert that an object is serializable."""
+    """Assert that an object is serializable, uses both dump and dump_json methods."""
     assert obj is not None
-    serialized = obj.json()
+    serialized = obj.model_dump_json()
     assert isinstance(serialized, str)
-    deserialized = obj_type.parse_raw(serialized)
-    assert deserialized == obj
```
rewrlution, line 285: Any reason why we don't do this anymore?

python/semantic_kernel/template_engine/code_tokenizer.py
```diff
@@ -12,6 +15,8 @@
 from semantic_kernel.template_engine.blocks.var_block import VarBlock
 from semantic_kernel.utils.null_logger import NullLogger
 
+# from semantic_kernel.utils.null_logger import NullLogger
```
rewrlution, line 18: if it is not in use, can we remove it?

rewrlution (commented):
Looks good overall, just have some minor questions.

python/tests/unit/test_serialization.py
eavanvalkenburg: Was replaced by a slightly different approach to comparing, see the note on __eq__ method here: https://docs.pydantic.dev/latest/migration/, so we are now checking equality by looking at the model_dumps (dict) of the original and one created based on json dump.

rewrlution:
Looks good to me!

rewrlution (approved):
Looks good to me!

alexchaomander (approved):
LGTM!

python/pyproject.toml
```diff
@@ -16,7 +16,7 @@ python-dotenv = "1.0.0"
 regex = "^2023.6.3"
 openapi_core = "^0.18.0"
 prance = "^23.6.21.0"
-pydantic = "<2"
+pydantic = ">2"
```
moonbox3, line 19: I can't comment on it above, but can we please upgrade the `semantic-kernel` version to "0.4.1.dev"?

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
```diff
@@ -60,7 +60,7 @@ def from_dict(cls, settings: Dict[str, str]) -> "OpenAIChatCompletion":
         """
 
         return OpenAIChatCompletion(
-            model_id=settings["model_id"],
+            model_id=settings["ai_model_id"],
```
moonbox3, line 63: Would it be better to standardize `ai_model_id` across all code instead of having a mix and match or `ai_model_id` and `model_id`?

python/pyproject.toml
moonbox3: We chatted offline. No need to update now.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion.py
eavanvalkenburg: agreed, bit bigger breaking change but much easier to read and understand what's going on.

moonbox3 (approved):
LGTM, thank you @eavanvalkenburg!

PR #3727, opened 2023-11-28
.Net BugFix #3715 Exception Streaming OpenAI Chat WithData (Example 54)

### LINKED ISSUE #3715: .Net: Exception during streaming of Azure OpenAI Chat Completion With Data
**Describe the bug**
When running C# Example54_AzureChatCompletionWithData, GetChatCompletionAsync() succeeds, but GetStreamingContentAsync() throws an exception: `System.NotSupportedException: 'Type System.String is not supported'`

This seems to be coming from AzureOpenAIChatCompletionWithData.cs line 232. The type (string) should actually be supported in the conditional on line 219, but seems like the awaiting logic isn't behaving correctly and it proceeds to throw this error.

**To Reproduce**
1. Go to dotnet KernelSyntaxExamples and run example 54. 
2. May need to manually set the stop sequences parameter to null (instead of empty list) to get past a separate validation error.
3. See NotSupportedException thrown after the second ask/response.

**Expected behavior**
Streaming should proceed with console outputs as the response comes in.

**Screenshots**
N/A

**Platform**
 - OS: Windows
 - IDE:  Visual Studio
 - Language: C#
 - Source: main branch of repository

**Additional context**
N/A

RogerBarreto:
### Motivation and Context

This fixes two bugs 

- While using WithData apis if no stopsequence is provided a BadRequest happens.
- Resolves #3715

### Description

Fix for the above bugs.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIPromptExecutionSettings.cs
+++ b/dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIPromptExecutionSettings.cs
@@ -68,7 +68,7 @@ public class OpenAIPromptExecutionSettings : PromptExecutionSettings
     /// Sequences where the completion will stop generating further tokens.
     /// </summary>
     [JsonPropertyName("stop_sequences")]
-    public IList<string> StopSequences { get; set; } = Array.Empty<string>();
+    public IList<string>? StopSequences { get; set; } = null;
 
     /// <summary>
     /// How many completions to generate for each prompt. Default is 1.
--- a/dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIPromptExecutionSettings.cs
+++ b/dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIPromptExecutionSettings.cs
@@ -68,7 +68,7 @@ public class OpenAIPromptExecutionSettings : PromptExecutionSettings
     /// Sequences where the completion will stop generating further tokens.
     /// </summary>
     [JsonPropertyName("stop_sequences")]
-    public IList<string>? StopSequences { get; set; } = null;
+    public IList<string>? StopSequences { get; set; } = Array.Empty<string>();
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -156,6 +156,26 @@ public static OpenAIPromptExecutionSettings FromRequestSettings(PromptExecutionS
...
+    /// <returns>An instance of OpenAIPromptExecutionSettings</returns>
+    public static OpenAIPromptExecutionSettings FromRequestSettingsWithData(PromptExecutionSettings? executionSettings, int? defaultMaxTokens = null)
+    {
+        var requestSettings = FromRequestSettings(executionSettings, defaultMaxTokens);
+
+        if (requestSettings.StopSequences?.Count == 0)
+        {
+            // Azure OpenAI WithData API does not allow to send empty array of stop sequences
+            // Gives back "Validation error at #/stop/str: Input should be a valid string\nValidation error at #/stop/list[str]: List should have at least 1 item after validation, not 0"
+            requestSettings.StopSequences = null;
```
markwallace-microsoft, line 173: Should this be fixed in the OPenAI specific version of PromptExecutionSettings so that it applies to all calls to OpenAI?

RogerBarreto: The normal APIs (Non WithData) validation don't complain about the empty array.

I'm applying this change to the WithData only to minimize any impact on the current code.

image

RogerBarreto: Analyzing further, in the ClientBase, using the Non WithData endpoints, Azure SDK model don't have nullable StopSequences for its `ChatCompletionsOptions` class.

image

markwallace-microsoft (approved):

PR #3738, opened 2023-11-28
.Net: Rename AzureTextCompletion to AzureOpenAITextCompletion

stephentoub:
There are other AI-related services on Azure beyond Azure OpenAI.

(Is this type even still needed, with AzureOpenAIChatCompletion implementing both ITextCompletion and IChatCompletion?)


RogerBarreto:
@stephentoub Yes, our current implementation still uses the `ITextCompletion` in the `KernelFunctionsFromPrompt` this will change as soon the multi modality work I'm doing is implemented.

RogerBarreto (approved):

stephentoub:
> @stephentoub Yes, our current implementation still uses the `ITextCompletion` in the `KernelFunctionsFromPrompt` this will change as soon the multi modality work I'm doing is implemented.

Right, but why is this class needed for that? AzureOpenAIChatCompletion also inplements ITextCompletion.

RogerBarreto:
@stephentoub When Chat completion became a thing, this was a way to be backward compatible with current implementation that actually uses Chat completion models. So when a sk developer used the service directly it would have the benefits of the new interface and when using the Kernel, the prompt is internally converted to a CHatHistory.

stephentoub:
Yup, I get it. I'm wondering what benefit this type _currently_ has. It looks completely unnecessary / duplicatuve _now_. Can we just delete it?

SergeyMenshykh (approved):

PR #3748, opened 2023-11-29
.Net: Remove KernelFunctionMetadata.PluginName

### LINKED ISSUE #3704: .Net: Do we need to keep KernelFunctionMetadata.PluginName

markwallace-microsoft:
### Motivation and Context

Close #3704 

### Description

- Remove `KernelFunctionMetadata.PluginName`
- Introduce `KernelPluginFunctionMetadata`
- Change planner code to manage plugin name association with `KernelFunctionMetadata`

### Contribution Checklist


dotnet/src/InternalUtilities/planning/Functions/KernelPluginFunctionMetadata.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>Gets the metadata for the parameters to the function.</summary>
+    /// <remarks>If the function has no parameters, the returned list will be empty.</remarks>
+    public IReadOnlyList<KernelParameterMetadata> Parameters => this._metadata.Parameters;
+
+    /// <summary>Gets parameter metadata for the return parameter.</summary>
+    /// <remarks>If the function has no return parameter, the returned value will be a default instance of a <see cref="KernelReturnParameterMetadata"/>.</remarks>
+    public KernelReturnParameterMetadata ReturnParameter => this._metadata.ReturnParameter;
+
+    /// <summary>Gets the <see cref="KernelFunctionMetadata"/> that this instance was created from.</summary>
+    public KernelFunctionMetadata Metadata => this._metadata;
```
stephentoub, line 49: Did you consider something more like:
```C#
public KernelPluginFunctionMetadata(string pluginName, KernelFunctionMetadata functionMetadata)
{
    Verify.NotNull(pluginName);
    Verify.NotNull(function);
    this.PluginName = pluginName;
    this.FunctionMetadata = functionMetadata;
}

public string PluginName { get; }
public KernelFunctionMetadata FunctionMetadata { get; }
```
?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelPluginMetadata.cs
```diff
@@ -0,0 +1,64 @@
...
+    [AllowNull]
+    public string Description
+    {
+        get => this._description;
+        init => this._description = value ?? string.Empty;
+    }
+
+    /// <summary>Gets the metadata for the functions to the plugin.</summary>
+    /// <remarks>If the plugin has no functions, the returned list will be empty.</remarks>
+    public IReadOnlyList<KernelFunctionMetadata> FunctionsMetadata
```
stephentoub, line 55: Shouldn't these be KernelPluginFunctionMetadata?

dotnet/src/InternalUtilities/planning/Functions/KernelPluginFunctionMetadata.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+#pragma warning disable IDE0130
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Provides read-only metadata for an <see cref="KernelFunction"/>.
+/// </summary>
+public sealed class KernelPluginFunctionMetadata
```
RogerBarreto, line 14: Should we consider using `record` instead?

```suggestion
public sealed record KernelPluginFunctionMetadata
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelPluginMetadata.cs
```diff
@@ -0,0 +1,64 @@
...
+
+#pragma warning disable IDE0130
+// ReSharper disable once CheckNamespace - Using the main namespace
+namespace Microsoft.SemanticKernel;
+#pragma warning restore IDE0130
+
+/// <summary>
+/// Provides read-only metadata for an <see cref="IKernelPlugin"/>.
+/// </summary>
+public sealed class KernelPluginMetadata
```
RogerBarreto, line 15: Should we consider using record instead?

```suggestion
public sealed record KernelPluginMetadata
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelPluginMetadata.cs
RogerBarreto: @markwallace-microsoft That's a good point.

What's the difference between `KernelFunctionMetadata` and `KernelPluginFunctionMetadata`

dotnet/src/InternalUtilities/planning/Extensions/IKernelPluginExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+// ReSharper disable once CheckNamespace - Using the main namespace
+namespace Microsoft.SemanticKernel;
+
+/// <summary>Provides extension methods for working with <see cref="IKernelPlugin"/>s and collections of them.</summary>
+internal static class IKernelPluginExtensions
+{
+    /// <summary>Gets a collection of <see cref="KernelPluginFunctionMetadata"/> instances, one for every function in every plugin in the plugins collection.</summary>
+    /// <param name="plugins">The plugins collection.</param>
+    /// <returns>A list of metadata over every function in the plugins collection</returns>
+    public static IList<KernelPluginFunctionMetadata> GetFunctionsMetadata(this IEnumerable<IKernelPlugin> plugins)
```
stephentoub, line 16: I think we should also have a `public static KernelPluginFunctionMetadata GetFunctionMetadata(this IReadOnlyKernelPluginCollection plugins, string pluginName, string functionName)` that gets the KernelPluginFunctionMetadata for a specific function, with the plugin name appropriately configured.

We should also need an overload of ToOpenAIFunction that takes a KernelPluginFunctionMetadata.  That would then be used in places like:
https://github.com/microsoft/semantic-kernel/pull/3748/files#diff-e1c9cdaa9f8f5449a7cac7aab6c7f6abb673b04e3bea5de3ea16635423053f6dR170
such that the plugin name would be appropriately propagated.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelPluginMetadata.cs
markwallace-microsoft: `KernelPluginFunctionMetadata` is an internal class that will only be used by the planners. I'm adding it becase the planner implementation is very tightly coupled to having the plugin name as part of the `KernelFunctionMetadata`.

I started trying to refactor the planner code to unwind this but it requires a lot of changes so I went for the simple approach. This PR allows us to clean up the public API and doesn't overly disrupt the planner implementation.

dotnet/src/InternalUtilities/planning/Functions/KernelPluginFunctionMetadata.cs
markwallace-microsoft: I'll adopt some of those changes but the shortcut properties are handy

dotnet/samples/KernelSyntaxExamples/Example10_DescribeAllPluginsAndFunctions.cs
```diff
@@ -86,7 +89,7 @@ private static void PrintFunction(KernelFunctionMetadata func)
 /** Sample output:
 
 *****************************************
-****** Native plugins and functions ******
+****** Native plugins and pluginsMetadata ******
```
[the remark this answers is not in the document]
markwallace-microsoft: The current description is good we are reading the plugin and function metadata to see the descriptions for the plugins and functions

Issue #3750, opened 2023-11-29
.Net: Bug: dotnet README Instructions Invalid


JasonMCreighton:
**Describe the bug**
The OpenAIPromptExecutionSettings object is not available. 

**To Reproduce**
1. Install Microsoft.SemanticKernel from nuget (defaults to 1.0.0-beta8)
2. Copy code into IDE
3. See error

**Expected behavior**
The README code should execute with minimal changes

**Platform**
 - OS: Windows
 - IDE: Visual Studio
 - Language: C#
 - Source: 1.0.0-beta8

**Additional context**
Pull request attached

matthewbolanos:
Everything is up-to-date in the repo. We'll be publishing this class later today when we publish our release candidate to NuGet

PR #3751, opened 2023-11-29
.Net: Update README.md

JasonMCreighton:
OpenAIPromptExecutionSettings is no longer available in Microsoft.SemanticKernel 1.0.0-beta8, instead OpenAIRequestSettings is available.

### Motivation and Context
The README file does not execute as expected. As I was working through it, I had to figure out that the original object is no longer in use. Changing it to the updated OpenAIRequestSettings will allow users to use the README file easier.

https://github.com/microsoft/semantic-kernel/issues/3750

### Contribution Checklist

### DIFF
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -42,7 +42,7 @@ var prompt = @"{{$input}}
 
 One line TLDR with the fewest words.";
 
-var summarize = kernel.CreateSemanticFunction(prompt, executionSettings: new OpenAIPromptExecutionSettings { MaxTokens = 100 });
+var summarize = kernel.CreateSemanticFunction(prompt, requestSettings: new OpenAIRequestSettings() { MaxTokens = 100 });
 
 string text1 = @"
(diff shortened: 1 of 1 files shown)


dotnet/README.md
```diff
@@ -42,7 +42,7 @@ var prompt = @"{{$input}}
 
 One line TLDR with the fewest words.";
 
-var summarize = kernel.CreateSemanticFunction(prompt, executionSettings: new OpenAIPromptExecutionSettings { MaxTokens = 100 });
+var summarize = kernel.CreateSemanticFunction(prompt, requestSettings: new OpenAIRequestSettings() { MaxTokens = 100 });
```
markwallace-microsoft, line 45: The class was renamed since we release 1.0.0-beta8 so the readme is correct

See: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.AI.OpenAI/OpenAIPromptExecutionSettings.cs

Issue #3690, opened 2023-11-27
.Net: Change Kernel.RunAsync extension methods to use KernelFunctionArguments.


markwallace-microsoft:

stephentoub:
This was done

Issue #3689, opened 2023-11-27
.Net: **(WIP)**(Sergey) Add KernelFunctionsArgument arg to ISKFunction.InvokeAsync, remove SKContext.ContextVariables. Delete ContextVariables.


markwallace-microsoft:

stephentoub:
This was done

PR #3752, opened 2023-11-29
.Net ContextVariables is replaced by KernelFunctionArguments

SergeyMenshykh:
Closes - https://github.com/microsoft/semantic-kernel/issues/3689 and https://github.com/microsoft/semantic-kernel/issues/3690

The PR introduces the following changes:
- The 'ContextVariables' class has been removed. Previously, this class was used by kernel functions as a property bag, where functions could resolve their arguments and store their outputs. By removing this class and introducing the 'KernelFunctionArguments' class, the function calling pattern for SK functions now resembles a more familiar and convenient approach, where functions accept a list of arguments and return a result, rather than working with and mutating a property bag.
- The functionality for converting function results from objects to strings has been moved from the 'KernelFunctionFromMethod' class to the 'FunctionResult' class. This change is a step towards enabling SK to work with complex types, rather than just strings. Currently, SK can only work with strings, and it is not possible to save the string representation of results in context variables anymore. Therefore, the conversion functionality has been moved to the 'FunctionResult' class until support for complex types is added to SK.


dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -95,109 +82,45 @@ public async Task<string> Type03Async()
...
     }
 
     [KernelFunction]
-    public async Task Type18Async()
+    public async Task Type09Async()
     {
         await Task.Delay(0);
-        Console.WriteLine("Running function type 18");
+        Console.WriteLine("Running function type 09");
     }
```
markwallace-microsoft, line 125: Should there also be

```csharp
public async FunctionResult Type10()
public async Task<FunctionResult> Type11Async()
```

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -25,7 +24,7 @@ public HandlebarsPromptTemplate(PromptTemplateConfig promptConfig, ILoggerFactor
     }
 
     /// <inheritdoc/>
-    public async Task<string> RenderAsync(Kernel kernel, ContextVariables variables, CancellationToken cancellationToken = default)
+    public async Task<string> RenderAsync(Kernel kernel, IDictionary<string, string> arguments, CancellationToken cancellationToken = default)
```
markwallace-microsoft, line 27: Can you pass in `KernelFunctionArguments` because we want to access the execution settings also and if anything new gets added we may want the template rendering to have access to that.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -95,18 +99,12 @@ public FunctionResult(string functionName, ContextVariables variables, object? v
...
     {
-        if (this.Variables.TryGetValue(key, out string? valueObject))
-        {
-            value = valueObject;
-            return true;
-        }
-
-        value = default!;
-        return false;
+        this.Value = value;
```
markwallace-microsoft, line 107: Why do we need to do this?

I'd expect the `FunctionResult` to be immutable

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/IPromptTemplate.cs
```diff
@@ -17,8 +17,8 @@ public interface IPromptTemplate
     /// Render the template using the information in the context
     /// </summary>
     /// <param name="kernel">The Kernel.</param>
-    /// <param name="variables">The variables.</param>
+    /// <param name="arguments">The arguments.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>Prompt rendered to string</returns>
-    public Task<string> RenderAsync(Kernel kernel, ContextVariables variables, CancellationToken cancellationToken = default);
+    public Task<string> RenderAsync(Kernel kernel, IDictionary<string, string> arguments, CancellationToken cancellationToken = default);
```
markwallace-microsoft, line 23: Pass `KernelFunctionArguments` so prompt has access to the execution settings

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
```diff
@@ -260,9 +257,9 @@ private static bool IsAsyncMethod(MethodInfo method)
     /// <summary>
     /// Gets a delegate for handling the marshaling of a parameter.
     /// </summary>
-    private static (Func<Kernel, ContextVariables, CancellationToken, object?>, KernelParameterMetadata?) GetParameterMarshalerDelegate(
+    private static (Func<Kernel, KernelFunctionArguments, CancellationToken, object?>, KernelParameterMetadata?) GetParameterMarshalerDelegate(
```
markwallace-microsoft, line 260: Mentioned above also, should we support `Task<FunctionResult>` and `FunctionResult` as supported return types?

I think we supported `SKContext` previously.

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/CodeBlock.cs
```diff
@@ -71,7 +70,7 @@ public override bool IsValid(out string errorMsg)
     }
 
     /// <inheritdoc/>
-    public async Task<string> RenderCodeAsync(Kernel kernel, ContextVariables variables, CancellationToken cancellationToken = default)
+    public async Task<string> RenderCodeAsync(Kernel kernel, IDictionary<string, string> arguments, CancellationToken cancellationToken = default)
```
markwallace-microsoft, line 73: I will change this to take `KernelFunctionArguments`

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
SergeyMenshykh: They can easily be be added. We will have to extend existing marshaling for function results to prevent function results be wrapped into another function result. I'm happy to update it.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
SergeyMenshykh: The change is needed to allow the function result to be altered by the invoked hook. Before the changes, the result would be indirectly modified through the context variables within the hook by assigning a new value to the input variable:
image
After that, the 'CallFunctionInvoked' will create a new instance of 'FunctionResult' and use the input variable from the context as a result.
image

We can consider an alternative design such as adding a new 'UpdateResult(object? newResult)' method to the 'FunctionInvokedEventArgs' class. This would initialize an internal property that would eventually be used when creating a new instance of 'FunctionResult' in the 'CallFunctionInvoked' method.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
SergeyMenshykh: I don't see any reason why we shouldn't. It has already been implemented as part of another comment.

dotnet/samples/KernelSyntaxExamples/Example03_Variables.cs
```diff
@@ -21,10 +20,11 @@ public static async Task RunAsync()
...
         var textPlugin = kernel.ImportPluginFromObject<StaticTextPlugin>();
 
-        var variables = new ContextVariables("Today is: ");
-        variables.Set("day", DateTimeOffset.Now.ToString("dddd", CultureInfo.CurrentCulture));
+        var arguments = new KernelFunctionArguments();
+        arguments["input"] = "Today is: ";
+        arguments["day"] = DateTimeOffset.Now.ToString("dddd", CultureInfo.CurrentCulture);
 
-        var result = await kernel.InvokeAsync(textPlugin["AppendDay"], variables);
+        var result = await kernel.InvokeAsync(textPlugin["AppendDay"], arguments);
```
stephentoub, line 27: Just to confirm, syntactically this could also be written as:
```C#
var result = await kernel.InvokeAsync(textPlugin["AppendDay"], new()
{
    { "input", "Today is: " },
    { "day", DateTimeOffset.Now.ToString("dddd", CultureInfo.CurrentCulture) }
});
```
and it would compile and work, right?  Just want to make sure our design of KernelFunctionArguments doesn't wall that off, as it's relatively natural and will become even more so once collection expressions support dictionaries in C# 13.

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -35,15 +34,15 @@ public async Task<string> RenderAsync(Kernel kernel, ContextVariables variables,
             {
                 handlebars.RegisterHelper($"{plugin.Name}_{function.Name}", (writer, hcontext, parameters) =>
                 {
-                    var result = function.InvokeAsync(kernel, variables).GetAwaiter().GetResult();
-                    writer.WriteSafeString(result.GetValue<string>());
+                    var result = function.InvokeAsync(kernel, new(arguments)).GetAwaiter().GetResult();
```
stephentoub, line 37: Why are we cloning it? Are we trying to avoid mutations propagating from one invocation to another or something?

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -95,109 +86,60 @@ public async Task<string> Type03Async()
...
-        return "";
-    }
-
-    [KernelFunction]
-    public async Task<string> Type06Async(Kernel kernel)
+    public async Task<string> Type04Async(Kernel kernel)
     {
-        var summary = await kernel.InvokeAsync(kernel.Plugins["SummarizePlugin"]["Summarize"], new ContextVariables("blah blah blah"));
-        Console.WriteLine($"Running function type 6 [{summary?.GetValue<string>()}]");
+        var summary = await kernel.InvokeAsync(kernel.Plugins["SummarizePlugin"]["Summarize"], new KernelFunctionArguments() { { "input", "blah blah blah" } });
```
stephentoub, line 91: If we're passing this to more than just KernelFunction.InvokeAsync, I wonder if it should be KernelArguments rather than KernelFunctionArguments?  Not for this PR, just something to think about as a possible refactor immediately after.

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -95,109 +86,60 @@ public async Task<string> Type03Async()
...
+        Console.WriteLine($"Running function type 4 [{summary?.GetValue<string>()}]");
```
stephentoub, line 92: Could this just be:
```suggestion
        Console.WriteLine($"Running function type 4 [{summary}]");
```
and rely on FunctionResult's ToString?

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -95,109 +86,60 @@ public async Task<string> Type03Async()
...
+    public async Task Type09Async()
     {
         await Task.Delay(0);
-        Console.WriteLine("Running function type 16");
+        Console.WriteLine("Running function type 09");
     }
 
     [KernelFunction]
-    public async Task Type17Async(string x, ContextVariables variables)
+    public FunctionResult Type10()
```
stephentoub, line 132: Is FunctionResult special-cased, or does this end up with a FunctionResult wrapping a FunctionResult?

dotnet/src/IntegrationTests/Plugins/PluginTests.cs
```diff
@@ -63,7 +62,7 @@ public class PluginTests
             new Uri(pluginEndpoint),
             new OpenApiFunctionExecutionParameters(httpClient));
 
-        var contextVariables = new ContextVariables();
+        var contextVariables = new KernelFunctionArguments();
```
stephentoub, line 65: Also search/replace `contextVariables`?

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsTemplateEngineExtensions.cs
```diff
@@ -27,16 +27,14 @@ internal sealed class HandlebarsTemplateEngineExtensions
...
-    /// <param name="variables">The dictionary of variables to pass to the Handlebars template engine.</param>
+    /// <param name="arguments">The dictionary of arguments to pass to the Handlebars template engine.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>The rendered Handlebars template.</returns>
     public static string Render(
         Kernel kernel,
-        ContextVariables contextVariables,
         string template,
-        Dictionary<string, object?> variables,
+        Dictionary<string, object?> arguments,
```
stephentoub, line 37: Should some arguments that are currently dictionaries becomes KernelFunctionArguments?

dotnet/src/Plugins/Plugins.Core/ConversationSummaryPlugin.cs
```diff
@@ -59,50 +58,43 @@ public ConversationSummaryPlugin()
...
     {
         List<string> lines = TextChunker.SplitPlainTextLines(input, MaxTokens);
         List<string> paragraphs = TextChunker.SplitPlainTextParagraphs(lines, MaxTokens);
 
         string[] results = new string[paragraphs.Count];
         for (int i = 0; i < results.Length; i++)
         {
-            variables.Update(paragraphs[i]);
-            results[i] = (await func.InvokeAsync(kernel, variables).ConfigureAwait(false)).GetValue<string>() ?? "";
+            results[i] = (await func.InvokeAsync(kernel, new KernelFunctionArguments { { KernelFunctionArguments.InputParameterName, paragraphs[i] } }).ConfigureAwait(false)).GetValue<string>() ?? "";
```
stephentoub, line 97: Construct the dictionary once and overwrite the value on each iteration?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
```diff
@@ -13,6 +13,11 @@ namespace Microsoft.SemanticKernel;
 /// </summary>
 public sealed class KernelFunctionArguments : Dictionary<string, string>
 {
+    /// <summary>
+    /// The main input parameter name.
+    /// </summary>
+    public const string InputParameterName = "input";
```
stephentoub, line 19: We'll consider removing this subsequently?

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -1,7 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel;
+using System.Globalization;
+using System.Reflection;
 
 namespace Microsoft.SemanticKernel.Orchestration;
```
stephentoub, line 10: We should move this to Microsoft.SemanticKernel. That's where Kernel lives, KernelFunction lives, and every invocation on those gives you back one of these.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -47,31 +51,31 @@ public sealed class FunctionResult
...
     /// <summary>
     /// Initializes a new instance of the <see cref="FunctionResult"/> class.
     /// </summary>
     /// <param name="functionName">Name of executed function.</param>
-    /// <param name="variables">Instance of <see cref="ContextVariables"/> to pass in function pipeline.</param>
     /// <param name="value">Function result object.</param>
-    public FunctionResult(string functionName, ContextVariables variables, object? value)
-        : this(functionName, variables)
+    /// <param name="culture">The culture configured on the Kernel that executed the function.</param>
+    public FunctionResult(string functionName, object? value, CultureInfo culture)
```
stephentoub, line 74: Rather than pulling out the culture specifically, how about just passing in and storing the Kernel?

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -127,5 +125,78 @@ public bool TryGetMetadataValue<T>(string key, out T value)
     }
 
     /// <inheritdoc/>
-    public override string ToString() => this.Value?.ToString() ?? base.ToString();
+    public override string ToString()
+    {
+        var stringResult = ConvertToString(this.Value, this.Culture);
+
+        return string.IsNullOrEmpty(stringResult) ? string.Empty : stringResult!;
```
stephentoub, line 132:
```suggestion
        return ConvertToString(this.Value, this.Culture) ?? string.Empty;
```

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -127,5 +125,78 @@ public bool TryGetMetadataValue<T>(string key, out T value)
...
+                return (input, cultureInfo) =>
+                {
+                    return converter.ConvertToString(context: null, cultureInfo, input);
+                };
+            }
+
+            return null;
+        });
+
+    private static TypeConverter? GetTypeConverter(Type sourceType)
```
stephentoub, line 169: Can we consolidate this into a shared file used by both this code and the marshaling logic in KernelFunctionFromMethod?

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
SergeyMenshykh: In this case, it's a type conversion from 'IDictionary<string, string>' to 'KernelFunctionArguments' because the 'RenderAsync' method uses a dictionary instead of function arguments, as the 'InvokeAsync' method does.

We can think about using 'KernelFunctionArguments' everywhere instead of dictionaries - the planner, prompt template mentioned by @markwallace-microsoft in a few comments above.

dotnet/src/SemanticKernel.Core/Extensions/KernelFunctionExtensions.cs
```diff
@@ -19,32 +20,46 @@ public static class KernelFunctionExtensions
...
         CancellationToken cancellationToken = default)
     {
-        return function.InvokeAsync(kernel, variables ?? new ContextVariables(), executionSettings, cancellationToken);
+        var arguments = new KernelFunctionArguments(requestSettings);
+        if (!string.IsNullOrEmpty(input))
+        {
+            arguments.Add(KernelFunctionArguments.InputParameterName, input);
+        }
+
+        return function.InvokeAsync(kernel, arguments, cancellationToken);
```
stephentoub, line 39: We allow arguments to be null, right? If yes, could this instead be:
```C#
KernelFunctionArguments? arguments = requestSettings is non null ? new(requestSettings) : null;
if (!string.IsNullOrEmpty(input))
{
    (arguments ??= new()).Add(KernelFunctionArguments.InputParameterName, input);
}

return function.InvokeAsync(kernel, arguments, cancellationToken);
```

dotnet/src/SemanticKernel.Core/Extensions/KernelFunctionExtensions.cs
```diff
@@ -19,32 +20,46 @@ public static class KernelFunctionExtensions
...
-    public static Task<FunctionResult> InvokeAsync(this KernelFunction function,
+    /// <returns>A asynchronous list of streaming result chunks</returns>
+    public static IAsyncEnumerable<T> InvokeStreamingAsync<T>(this KernelFunction function,
         Kernel kernel,
         string input,
         PromptExecutionSettings? executionSettings = null,
         CancellationToken cancellationToken = default)
-        => function.InvokeAsync(kernel, new ContextVariables(input), executionSettings, cancellationToken);
+    {
+        var arguments = new KernelFunctionArguments(executionSettings);
```
stephentoub, line 57: Ditto

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
SergeyMenshykh: Yep, good idea, in this case the class could be used in a broader context.

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
SergeyMenshykh: I've made it special case to prevent the wrapping.

dotnet/samples/KernelSyntaxExamples/Example06_TemplateLanguage.cs
```diff
@@ -53,7 +53,7 @@ public static async Task RunAsync()
         Console.WriteLine("--- Rendered Prompt");
         var promptTemplateFactory = new KernelPromptTemplateFactory();
         var promptTemplate = promptTemplateFactory.Create(new PromptTemplateConfig(FunctionDefinition));
-        var renderedPrompt = await promptTemplate.RenderAsync(kernel, new ContextVariables());
+        var renderedPrompt = await promptTemplate.RenderAsync(kernel, new Dictionary<string, string>());
```
RogerBarreto, line 56: I think now we can add optional to this parameter...

```suggestion
        var renderedPrompt = await promptTemplate.RenderAsync(kernel);
```

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -39,35 +39,26 @@ public static async Task RunAsync()
...
         await kernel.InvokeAsync(kernel.Plugins["test"]["type03"]);
 
-        await kernel.InvokeAsync(plugin["type04"], variables);
-        await kernel.InvokeAsync(kernel.Plugins["test"]["type04"], variables);
+        await kernel.InvokeAsync(plugin["type04"]);
+        await kernel.InvokeAsync(kernel.Plugins["test"]["type04"]);
 
-        await kernel.InvokeAsync(plugin["type05"], variables);
-        await kernel.InvokeAsync(kernel.Plugins["test"]["type05"], variables);
+        await kernel.InvokeAsync(plugin["type05"], arguments);
```
RogerBarreto, line 45: IF there are no arguments we shouldn't need to pass this, same for others.

This might be here due to the old `SKContext`.

```suggestion
        await kernel.InvokeAsync(plugin["type05"]);
```

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -95,109 +86,60 @@ public async Task<string> Type03Async()
...
     }
 
     [KernelFunction]
-    public async Task Type17Async(string x, ContextVariables variables)
+    public FunctionResult Type10()
     {
-        await Task.Delay(0);
-        Console.WriteLine("Running function type 17");
+        Console.WriteLine("Running function type 10");
+        return new FunctionResult("");
```
RogerBarreto, line 135:
```suggestion
        return new FunctionResult(string.Empty);
```

dotnet/samples/KernelSyntaxExamples/Example15_TextMemoryPlugin.cs
```diff
@@ -204,7 +203,7 @@ private static async Task RunWithStoreAsync(IMemoryStore memoryStore, Cancellati
 
         // Save a memory with the Kernel
         Console.WriteLine("Saving memory with key 'info5': \"My family is from New York\"");
-        await kernel.InvokeAsync(memoryFunctions["Save"], new ContextVariables()
+        await kernel.InvokeAsync(memoryFunctions["Save"], new KernelFunctionArguments()
```
RogerBarreto, line 206: Suggest changing the name of the variable to Plugin.

```suggestion
        await kernel.InvokeAsync(memoryPlugin["Save"], new KernelFunctionArguments()
```

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsTemplateEngineExtensions.cs
SergeyMenshykh: I believe so, as mentioned in several other comments, except for this specific case, which will be refactored later when we have support for complex types.

dotnet/samples/KernelSyntaxExamples/Example21_ChatGPTPlugins.cs
```diff
@@ -28,11 +27,11 @@ private static async Task RunChatGptPluginAsync()
         var plugin = await kernel.ImportPluginFromOpenAIAsync("<plugin name>", new Uri("<chatGPT-plugin>"), new OpenAIFunctionExecutionParameters(httpClient));
 
         //Add arguments for required parameters, arguments for optional ones can be skipped.
-        var contextVariables = new ContextVariables();
-        contextVariables.Set("<parameter-name>", "<parameter-value>");
+        var arguments = new KernelFunctionArguments();
+        arguments["<parameter-name>"] = "<parameter-value>";
```
RogerBarreto, line 31:
```suggestion
        var arguments = new KernelFunctionArguments { ["<parameter-name>"] = "<parameter-value>" };
```

dotnet/src/Plugins/Plugins.Core/ConversationSummaryPlugin.cs
SergeyMenshykh: Yep

dotnet/samples/KernelSyntaxExamples/Example22_OpenApiPlugin_AzureKeyVault.cs
```diff
@@ -83,15 +82,15 @@ public static async Task AddSecretToAzureKeyVaultAsync(InteractiveMsalAuthentica
...
-        contextVariables.Set("secret-name", "<secret-name>");
-        contextVariables.Set("api-version", "7.0");
-        contextVariables.Set("value", "<secret-value>");
-        contextVariables.Set("enabled", "<enabled>");
+        var arguments = new KernelFunctionArguments();
+        arguments["server-url"] = TestConfiguration.KeyVault.Endpoint;
+        arguments["secret-name"] = "<secret-name>";
+        arguments["api-version"] = "7.0";
+        arguments["value"] = "<secret-value>";
+        arguments["enabled"] = "<enabled>";
```
RogerBarreto, line 90: Use index initializers when possible. Less verbose.

```suggestion
        var arguments = new KernelFunctionArguments 
        {
           ["server-url"] = TestConfiguration.KeyVault.Endpoint,
           ["secret-name"] = "<secret-name>",
           ["api-version"] = "7.0",
           ["value"] = "<secret-value>",
           ["enabled"] = "<enabled>",
        };
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -108,29 +106,34 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
 
         cancellationToken.ThrowIfCancellationRequested();
 
+        //Cloning the arguments to prevent mutation of the original ones
```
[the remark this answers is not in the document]
SergeyMenshykh: I thought about scenarios when the same instance of the arguments is used in more than one function - two run in parallel or sequentially, and there's a pre/post hook that could mutate the arguments.

On the other hand, callers themselves can clone the arguments per function call, if necessary.

I'll remove the cloning functionality.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
markwallace-microsoft: Yeah, let's plan on doing that, think it would be better to restrict when and how the function result can be changed.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionAttribute.cs
```diff
@@ -30,9 +27,7 @@ namespace Microsoft.SemanticKernel;
 /// by an LLM or embedding.
 /// </para>
 /// <para>
-/// Functions may have any number of parameters. Parameters of type <see cref="ILogger"/> and
-/// <see cref="CancellationToken"/> are filled in from the corresponding members of the <see cref="ContextVariables"/>;
-/// <see cref="ContextVariables"/> itself may also be a parameter. A given native function may declare at
+/// Functions may have any number of parameters. A given native function may declare at
```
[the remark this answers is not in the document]
SergeyMenshykh: The comments is updated.

dotnet/src/IntegrationTests/Plugins/PluginTests.cs
SergeyMenshykh: done

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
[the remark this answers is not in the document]
SergeyMenshykh: removed

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
SergeyMenshykh: Done

dotnet/samples/KernelSyntaxExamples/Example07_BingAndGooglePlugins.cs
```diff
@@ -134,13 +134,14 @@ private static async Task Example2Async(Kernel kernel)
...
-        Console.WriteLine(questions);
+        var question = "Who is the most followed person on TikTok right now? What's the exchange rate EUR:USD?";
+        Console.WriteLine(question);
 
         var oracle = kernel.CreateFunctionFromPrompt(SemanticFunction, new OpenAIPromptExecutionSettings() { MaxTokens = 150, Temperature = 0, TopP = 1 });
 
-        var answer = await kernel.InvokeAsync(oracle, new ContextVariables(questions)
+        var answer = await kernel.InvokeAsync(oracle, new KernelFunctionArguments()
         {
+            ["input"] = question,
```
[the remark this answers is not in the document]
SergeyMenshykh: No, we don't, it just was there as in many other places. Fixed.

dotnet/src/SemanticKernel.Core/Extensions/KernelFunctionExtensions.cs
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Core/Extensions/KernelFunctionExtensions.cs
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
SergeyMenshykh: yep, simple and elegant.

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
SergeyMenshykh: Yep, I'll do that shortly after this PR in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
SergeyMenshykh: Yep, I'll fix this shortly after this PR in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/CodeBlock.cs
SergeyMenshykh: Yep, it will be done shortly after this PR in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761

dotnet/samples/KernelSyntaxExamples/Example03_Variables.cs
SergeyMenshykh: It should be possible; however, there is ambiguity: "The call is ambiguous between the following methods or properties: 
'KernelExtensions.InvokeAsync(Kernel, KernelFunction, string, CancellationToken)' and 'KernelExtensions.InvokeAsync(Kernel, KernelFunction, KernelFunctionArguments?, CancellationToken)'."

It can easily be fixed by specifying parameter name:

    var result = await kernel.InvokeAsync(textPlugin["AppendDay"], arguments: new()
    {
        ["input"] = "Today is: ",
        [ "day"] = DateTimeOffset.Now.ToString("dddd", CultureInfo.CurrentCulture),
    })

Please let me know if this solution is acceptable. If not, I will try to address it within the scope of the issue - https://github.com/microsoft/semantic-kernel/issues/3761

dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
SergeyMenshykh: This will be addressed in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761 shortly after the PR is merged.

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
SergeyMenshykh: This will be addressed in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761 shortly after the PR is merged.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionArguments.cs
SergeyMenshykh: This will be addressed in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761 shortly after the PR is merged.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
SergeyMenshykh: Yep, absolutely. It will be addressed in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761 shortly after the PR is merged.

dotnet/src/SemanticKernel.Abstractions/Orchestration/FunctionResult.cs
```diff
@@ -47,31 +51,31 @@ public sealed class FunctionResult
...
     /// </summary>
     /// <param name="functionName">Name of executed function.</param>
-    /// <param name="variables">Instance of <see cref="ContextVariables"/> to pass in function pipeline.</param>
-    public FunctionResult(string functionName, ContextVariables variables)
+    public FunctionResult(string functionName)
     {
         this.FunctionName = functionName;
-        this.Variables = variables;
+        this.Culture = CultureInfo.InvariantCulture;
     }
```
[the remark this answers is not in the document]
SergeyMenshykh: Thanks Roger, the design is not relevan anymore and will be changes in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761.

dotnet/src/SemanticKernel.Abstractions/Functions/FunctionResult.cs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        throw new InvalidCastException($"Cannot cast {this.Value.GetType()} to {typeof(T)}");
+    }
+
+    /// <summary>
+    /// Sets function result.
+    /// </summary>
+    /// <param name="value">The result value.</param>
+    public void SetValue(object? value)
```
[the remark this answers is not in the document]
SergeyMenshykh: Yes, I'm considering a similar design in which the 'FunctionInvokedEventArgs' class would have an UpdateResult or SetResult method. This method would update the internal property of 'FunctionInvokedEventArgs', which would then be passed to the 'FunctionResult' constructor when creating a new instance in the 'CallFunctionInvoked' method. Anyway, I'll handle this in scope of this issue - https://github.com/microsoft/semantic-kernel/issues/3761 shortly after the PR is merged.

dotnet/samples/KernelSyntaxExamples/Example06_TemplateLanguage.cs
SergeyMenshykh: Tracking this by this issue - https://github.com/microsoft/semantic-kernel/issues/3761

RogerBarreto (approved):

PR #3758, opened 2023-11-29
.Net: Switch to NetAnalyzers 8.0.0 and fix warnings

markwallace-microsoft:
### Contribution Checklist


dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsTemplateEngineExtensions.cs
```diff
@@ -231,7 +231,7 @@ internal sealed class HandlebarsTemplateEngineExtensions
             var parameters = arguments[0] as IDictionary<string, object>;
 
             // Verify that the message has a role
-            if (!parameters!.ContainsKey("role"))
+            if (!parameters!.TryGetValue("role", out _))
```
stephentoub, line 234: The warning is because there are three dictionary lookups being done when only one is needed: here it's using ContainsKey, and then below it's using parameters["role"] twice, which is the same lookup.  So TryGetValue should be used as you've done, but instead of `out _`, it should be `out var role`, and then use `role` below instead of `parameters["role"]`.

PR #3760, opened 2023-11-30
add temp ADR to capture multi-modal requirements

matthewbolanos:
### Motivation and Context

Capturing the requirements for multi-modal; do not merge in.

### DIFF
--- a/docs/decisions/0024-multi-modal-requirements.md
+++ b/docs/decisions/0024-multi-modal-requirements.md
@@ -1,0 +1,304 @@
+# Requirements for long term multi-modal support
+
+## The core scenarios
+
+There are two main scenarios we need to support:
+
+1. A prompt engineer wants to author a prompt YAML file that can support _any_ multi-modal scenario.
+2. A professional developer wants to use a connector directly for strongly typed access to the models.
+
+For #2, we believe that our current interfaces for ITextCompletion and IChatCompletion are sufficient for professional developers.
+We believe this is true because both the input and output are already strongly typed.
+
+What's not supported today is #1. This is because the our Prompt function implementation only can return FunctionResults with `ITextResult`
+content. To ensure we don't "boil the ocean" it's _ok_ (and expected) that this implementation does not change, however we need to ensure
+our interfaces are sufficient to support additional types in the future.
+
+The following sections describe the requirements for #1 and provide scenarios that demonstrate the need for each requirement.
+
+### Requirement #1: Enable "Strongly typed" output for prompt functions
+
+Today, native functions can specify the return type of the function, but this is not possible for prompt functions.
+
+#### The problem
+
+Because prompts can't define their output, it is ambiguous what the return type of the prompt should be.
+
+```yaml
+template: |
+  <message role="system">
+  You are a helpful bot.
+  </message>
+  <message role="user">
+  Can you get me an image of a {{topic}}?
+  </message>
+template_format: handlebars
+description: Creates an image of the specified topic
+name: CreateImage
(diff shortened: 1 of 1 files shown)


docs/decisions/0024-multi-modal-requirements.md
```diff
@@ -0,0 +1,296 @@
...
+whenever the output is specified as an `Image`.
+
+```csharp
+private sealed class CustomAIServiceSelector : IAIServiceSelector
+{
+    public (T?, PromptExecutionSettings?) SelectAIService<T>(Kernel kernel, ContextVariables variables, KernelFunction function) where T : class, IAIService
+    {
+        if (function.Output.Type == type(Image))
+        {
+            foreach (var service in kernel.GetAllServices<T>())
```
markwallace-microsoft, line 112: We should really be getting all `IImageGeneration` services

docs/decisions/0024-multi-modal-requirements.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+#### Scoping for V1.0.0
+
+To keep V1.0.0 scoped, we just need to ensure the v1.0.0 interfaces have the methods that require a single string as input so they
+can be called by prompt functions. The interfaces that must be updated are:
+
+- ITextCompletion
+- IChatCompletion
+- And _maybe_ IAIService
```
markwallace-microsoft, line 221: Anything we add to `IAIService` we need to be 100% sure we will want this forever and it is reasonable for all services to implement.
Out current approach of having specializations of `IAIService` and allowing connectors to pick which ones to implement gives us good flexibility and the ability to adapt to future changes.

docs/decisions/0024-multi-modal-requirements.md
```diff
@@ -0,0 +1,296 @@
...
+
+_All other services (e.g., IImageGeneration) are marked experimental, so they can be ignored for right now._
+
+### Requirement #3: Make it easy to select a service with a required return type
+
+Today, each of our service interfaces have a unique return type:
+
+- `ITextCompletion`: `ITextResult`
+- `IChatCompletion`: `IChatResult`
+- `IImageGeneration`: `IImageResult`
```
markwallace-microsoft, line 231: We are missing `IImageResult` but we should add this. Created https://github.com/microsoft/semantic-kernel/issues/3768

docs/decisions/0024-multi-modal-requirements.md
```diff
@@ -0,0 +1,296 @@
...
+It is unrealistic to expect the `IAIServiceSelector` to enumerate all of these interfaces _just_ to find a service that returns text.
+
+#### Ideal developer experience
+
+Ideally, the `IAIServiceSelector` should be able to easily deduce the supported output types of a service in a single line of code.
+
+```csharp
+private sealed class CustomAIServiceSelector : IAIServiceSelector
+{
+    public (T?, PromptExecutionSettings?) SelectAIService<T>(Kernel kernel, ContextVariables variables, KernelFunction function) where T : class, IAIService
```
markwallace-microsoft, line 266: We still have the problem of how to determine the correct `T` to use.

If we want we can add a new abstraction e.g. `IAIMultiModelServiceSelector` in the future which determines the type and then the service and request settings. It could wrap the existing service selector to be compatible.

markwallace-microsoft (commented):
Thanks @matthewbolanos this helps to understand the requirements.

I think we can tackle this post v1.0 without introducing breaking changes. We are more likely to introduce issues by adding last minute changes to `IAIService`

PR #3762, opened 2023-11-30
.Net: Instrument the HandleBars planner

TaoChenOSU:
### Motivation and Context

Instrument the Handlebars planner.

### Description

Use the PlannerInstrumention class to wrap the HandleBars planner for instrumentation.
Run using the telemetry sample in the repository:
image
image

### Contribution Checklist

- [ ] The code builds clean without any errors or warnings
- [ ] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [ ] All unit tests pass, and I have added new tests where possible
- [ ] I didn't break anyone :smile:

PR #3763, opened 2023-11-30
.Net: Make PlannerInstrumentation internal rather than public

stephentoub:


markwallace-microsoft (approved):

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.SemanticKernel.Planning;
 #pragma warning restore IDE0130
 
 /// <summary>Surrounds the invocation of a planner with logging and metrics.</summary>
-public static class PlannerInstrumentation
+internal static class PlannerInstrumentation
```
markwallace-microsoft, line 16: Does this should up in the VS Solution Explorer? 

The other classes from this folder are listed in the main solution file.

markwallace-microsoft (commented):
This can be just a comment on Tao's PR https://github.com/microsoft/semantic-kernel/pull/3762

RogerBarreto (approved):

PR #3778, opened 2023-11-30
.Net: Remove implementation detail from public {Azure} OpenAI client surface area

stephentoub:
- Custom ".AzureSdk" and ".CustomClient" namespaces were removed
- Base class hierarchies were removed; these are implementation detail that needn't surface


markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/OpenAIFunction.cs
```diff
@@ -8,7 +8,9 @@
 using Json.Schema;
 using Json.Schema.Generation;
 
-namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.AzureSdk;
+#pragma warning disable IDE0130 // Namespace does not match folder structure
```
RogerBarreto, line 11: I think we can remove those pragmas as the new PR is getting merged.

markwallace-microsoft (approved):

PR #3791, opened 2023-11-30
.Net: Add streaming equivalent to InvokePromptAsync

stephentoub:
For the simple getting-started scenarios, InvokePromptAsync is a handy one-liner. We should have the same for streaming. This adds InvokePromptStreamingAsync.

Related, it's not clear to me why InvokePromptAsync takes a functionName or description. What will end up using those in this case (other than the function name being used in logging)? Can they be removed? We should keep the shape in sync between InvokePromptAsync and InvokePromptStreamingAsync, so we'd do the same thing for both.


dotnet/src/SemanticKernel.Core/KernelExtensions.cs
```diff
@@ -388,12 +388,45 @@ public static IKernelPlugin AddPluginFromObject(this ICollection<IKernelPlugin>
         string promptTemplate,
         PromptExecutionSettings? executionSettings = null,
         string? functionName = null,
-        string? description = null) =>
-        kernel.InvokeAsync((KernelFunction)KernelFunctionFactory.CreateFromPrompt(
+        string? description = null)
+    {
+        Verify.NotNull(kernel);
+        Verify.NotNull(promptTemplate);
```
RogerBarreto, line 394:
```suggestion
        Verify.NotNullOrWhiteSpace(promptTemplate);
```

RogerBarreto (approved):
Small suggestion

markwallace-microsoft (approved):

PR #3847, opened 2023-11-30
.Net: Add tests for KernelBuilder (and tweak the DI workaround)

stephentoub:


dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -234,12 +234,22 @@ public CultureInfo Culture
...
+                }
+
+                return Enumerable.Empty<T>();
             }
+        }
 
+        if (this.Services is EmptyServiceProvider)
+        {
+            // If the Kernel is created without a service provider, it uses a simple empty implementation.
+            // But he GetServices extension relies on the service provider special-casing enumerables.
```
markwallace-microsoft, line 251: Typo `But he`

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
stephentoub, line 251:
```suggestion
            // But the GetServices extension relies on the service provider special-casing enumerables.
```

PR #3848, opened 2023-11-30
.Net: Add back Kernel.LoggerFactory shortcut

stephentoub:
@markwallace-microsoft, this is something you previously suggested you wanted so I still had it on my todo list.  Do you still want this? I'm fine either way.


dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -120,6 +120,14 @@ public CultureInfo Culture
         set => this._culture = value ?? CultureInfo.CurrentCulture;
     }
 
+    /// <summary>
+    /// Gets the <see cref="ILoggerFactory"/> associated with this <see cref="Kernel"/>.
+    /// </summary>
+    /// <remarks>
+    /// This behaves the same as calling <see cref="GetService{ILoggerFactory}"/>.
+    /// </remarks>
+    public ILoggerFactory LoggerFactory => this.GetService<ILoggerFactory>();
```
RogerBarreto, line 129: Should we add this as an extension Method instead of a property?

```
public static ILoggerFactory GetLoggerFactory(this Kernel kernel) { ... } 
```

stephentoub: Why? If it's an extension method in the same assembly in the same namespace, it may as well just be an instance member.

RogerBarreto: Make sense, I just see that adding this as a non nullable property into the Kernel gives the strong message that this would be expected to be always there.

What happens when we don't add a loggerfactory in the KernelBuilder or in the DI, I believe this will this be null

stephentoub: > What happens when we don't add a loggerfactory in the KernelBuilder or in the DI, I believe this will this be null

This code ensures that `kernel.GetService<ILoggerFactory>()` always returns one:
https://github.com/microsoft/semantic-kernel/blob/1d51d5d1eb93f40fe888676c101d8cd5799b22f7/dotnet/src/SemanticKernel.Abstractions/Kernel.cs#L196-L202

markwallace-microsoft (approved):

RogerBarreto (approved):

PR #3849, opened 2023-11-30
.Net: Net: Tiny fixes found during bug bash

SergeyMenshykh:


markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -117,7 +117,7 @@ public sealed class Kernel
     public CultureInfo Culture
     {
         get => this._culture;
-        set => this._culture = value ?? CultureInfo.CurrentCulture;
+        set => this._culture = value ?? CultureInfo.InvariantCulture;
```
RogerBarreto, line 120: Suggest dropping the private field and use this property directly.

The field default is `CurrentCulture` ...

SergeyMenshykh: Thank you. The default value is now fixed. We need the field to include the logic in the setter - 'value ?? CultureInfo.InvariantCulture'. This prevents the field from being set to null, which is quite convenient because it eliminates the need for null checks in other component where the culture is being used.

RogerBarreto: Yep, sounds good to me.

RogerBarreto (approved):

PR #3854, opened 2023-11-30
.Net BugBash Bugfix Example 59 - FunctionCalling (Added default serialization to Method Function results)

### LINKED ISSUE #3829: Example59_OpenAIFunctionCalling Bug While Executing MethodFunctions
Error while using the new Connector Function calling impl.

System.InvalidCastException: 'Cannot cast Microsoft.SemanticKernel.Functions.OpenAPI.Model.RestApiOperationResponse to System.String'

A change to the code will be needed to handle <object> instead of strings.

RogerBarreto:
### Motivation and Context

Resolves #3829 

### Description

An error happened when attepting to convert Method functions to string while sending back to the LLM. 
To resolve the issue the value regardless of the type is converted to a JsonString and used as response.

Another bug was found while detecting the StreamingRole where only one chunk will have the role, not all.

### Contribution Checklist


dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -286,7 +287,10 @@ await foreach (Completions completions in response)
             }
 
             // Otherwise, invoke the function.
-            string functionResult = (await function.InvokeAsync(kernel, functionArgs, cancellationToken: cancellationToken).ConfigureAwait(false)).GetValue<string>() ?? string.Empty;
+            var functionResult = (await function.InvokeAsync(kernel, functionArgs, cancellationToken: cancellationToken).ConfigureAwait(false))
+                .GetValue<object>() ?? string.Empty;
+
+            var serializedFunctionResult = JsonSerializer.Serialize(functionResult);
```
stephentoub, line 292: @SergeyMenshykh, this ties in with the conversation we were having today.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -340,12 +344,12 @@ await foreach (Completions completions in response)
             StringBuilder? contentBuilder = null;
             string? functionName = null;
             StringBuilder? functionArgumentsBuilder = null;
-            ChatRole streamedRole = default;
+            ChatRole? streamedRole = default;
```
stephentoub, line 346: Why is this being made nullable?

RogerBarreto: To make it easier to use the less verbose `??=` operator. If this should be avoided I'm ok in adding an if statement in the foreach.

stephentoub: I'd done this matching the README example at https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/openai/Azure.AI.OpenAI#use-chat-functions (the last one with streaming).  Is it wrong, or did I copy it wrong? My understanding was that the SDK ensures that once Role shows up, subsequent updates include that same Role, so it's ok to always use the last. Is that not the case?

RogerBarreto: Seems to be wrong in the their example too.

This breaks when we try to reuse it for the next client.Call
Basically, the role only comes one chunk, 

`streamdRole = update.Role ?? default`
(will update the role to null again, erasing the chunk that has it set)

[This is a working sample of them](https://github.com/Azure/azure-sdk-for-net/blob/bfc10208f4abfb0d44463d64808132c2f68e5915/sdk/openai/Azure.AI.OpenAI/tests/Samples/Sample04_StreamingChat.cs#L33) on how to handle Roles.

[The function calling sample does not do the same](https://github.com/Azure/azure-sdk-for-net/blob/bfc10208f4abfb0d44463d64808132c2f68e5915/sdk/openai/Azure.AI.OpenAI/tests/Samples/Sample07_ChatFunctions.cs#L117C17-L117C55)

Will file an issue there. Thanks for letting me know.

stephentoub: Ok, thanks. I really thought the Azure SDK was propagating the role from update to update. Maybe that got dropped when they changed the streaming model? (Or maybe I'm completely misremembering.)

PR #3858, opened 2023-11-30
[Fix] Issuing multiple function callings will cause OpenAI to return 500 Internal Server Error

### LINKED ISSUE #3515: .Net: Issuing multiple function callings will cause OpenAI to return 500 Internal Server Error
**Describe the bug**
Issuing a single funciton  call is ok but if I issue multiple function calls to Open AI, Open AI will return 500 Internal Server Error.

**To Reproduce**
1. Set Example59 to use LLM model gpt-3.5-turbo-0613
2. Run Example59 and confirm that it is successful
3. Then, amend Example59 as below and execute again
4. 500 Internal Server Error is returned
```
        IKernel kernel = await InitializeKernelAsync();
        var chatCompletion = kernel.GetService<IChatCompletion>();
        var chatHistory = chatCompletion.CreateNewChat();

        OpenAIRequestSettings requestSettings = new()
        {
            Functions = kernel.Functions.GetFunctionViews().Select(f => f.ToOpenAIFunction()).ToList(),
        };
        await CompleteChatWithFunctionsAsync("What day is one month later?", chatHistory, chatCompletion, kernel, requestSettings);
```

**Expected behavior**
Both Function calls should be successful without 500 Internal Server Error.

**Platform**
 - OS: Windows
 - IDE: Visual Studio
 - Language: C#
 - Source: NuGet package version Microsoft.SemanticKernel 1.0.0-beta6
 - LLM Model: LLM model gpt-3.5-turbo-0613

**Reason of error and posible fix**
I repeated Example59 using direct native OpenAI API call and no 500 Internal Server Error was returned. Therefore, it is NOT an OpenAI bug. Below is the screen shot of the chat messages:
282069939-4d24c3d1-5e4f-4c2b-b5ff-e5f769e3b0fe
When OpenAI returns the function calling assistant message to us (record 2 above), the Name field is null and the FunctionCall field contains the function to be called. The name field is used only when returning the function result to OpenAI via the function message (record 3).

However, in SemanticKernel, when we insert the function calling assistant message to chat history and pass it back to OpenAI, the Name field was NOT null because in the foreach loop inside the CreateChatCompletionsOptions method in the ClientBase.cs file, the Name field was set to the function name instead, causing OpenAI to return 500 Internal Server Error.

With the below fix of the foreach loop in ClientBase.cs, OpenAI will NOT return 500 Internal Server Error:
```
{
    if (message.AdditionalProperties?.TryGetValue(NameProperty, out string? name) is true)
    {
        if (message.AdditionalProperties?.TryGetValue(ArgumentsProperty, out string? arguments) is true)
        {
            azureMessage.FunctionCall = new FunctionCall(name, arguments);
        }
        else
        {
            azureMessage.Name = name;
        }
    }
}
```

gitri-ms:
### Description

Fixes #3515 

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
+++ b/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
@@ -629,11 +629,13 @@ public abstract class ClientBase
 
             if (message.AdditionalProperties?.TryGetValue(NameProperty, out string? name) is true)
             {
-                azureMessage.Name = name;
-
                 if (message.AdditionalProperties?.TryGetValue(ArgumentsProperty, out string? arguments) is true)
                 {
                     azureMessage.FunctionCall = new FunctionCall(name, arguments);
+                }
+                else
+                {
+                    azureMessage.Name = name;
                 }
             }
 


dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientBase.cs
```diff
@@ -629,12 +629,14 @@ private static CompletionsOptions CreateCompletionsOptions(string text, OpenAIPr
 
             if (message.AdditionalProperties?.TryGetValue(NameProperty, out string? name) is true)
             {
-                azureMessage.Name = name;
-
                 if (message.AdditionalProperties?.TryGetValue(ArgumentsProperty, out string? arguments) is true)
                 {
                     azureMessage.FunctionCall = new FunctionCall(name, arguments);
                 }
+                else
```
alliscode, line 636: Is there a test case that failed before this change and now succeeds?

gitri-ms: Looks like our test case was made to match the incorrect behavior, which is why it was succeeding but now failing with this PR. I need to update the test.

stephentoub (approved):

alliscode (approved):

PR #3859, opened 2023-11-30
.Net: Establish pattern to use LoggerMessage attribute in KernelFunction and Instrumented Planner

TaoChenOSU:
### Motivation and Context

It has been recommended by folks from the OpenTelemetry team to use the `LoggerMessage` attribute to produce more optimized code. As we are improving observability for the kernel, it's also important that we achieve better observability with optimized performance.

For more information, please refer to this link: https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

### Description

1. Create a new extension for `KernelFunction` named `KernelFunctionLogExtensions` that contains methods to perform optimized logging.
2. Turn logging calls in `PlannerInstrumentation` to calls to source-generated logging code.

Other parts of the kernel will also adopt `LoggerMessage` in the future.

### Contribution Checklist


dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -39,39 +39,62 @@ internal static class PlannerInstrumentation
...
         long startingTimestamp = Stopwatch.GetTimestamp();
         try
         {
             var plan = await createPlanAsync(planner, kernel, goal, cancellationToken).ConfigureAwait(false);
-
-            if (logger.IsEnabled(LogLevel.Information))
-            {
-                logger.LogInformation("Plan created. Plan:\n{Plan}", planToString(plan));
-            }
+            logger.LogPlanCreatedWithPlan(planToString(plan));
```
stephentoub, line 50: This is actually a deoptimization. The source generator unfortunately leads folks into a false sense of "it takes care of everything for me", but it doesn't here: the argument `planToString(plan)` will be computed even if the associated log level is disabled. The check for LogLevel.Information should be put back.

TaoChenOSU: The `LoggerMessage` attribute has a hardcoded log level. If we put the log level check back, we'd have to make sure the log level in the check matches the one in the `LoggerMessage`. This feels a bit error prone. Let me see if I think of something else.

TaoChenOSU: We could consider parameterizing the log level.

TaoChenOSU: The PlannerInstrumentation will not longer accept a lambda that convert a plan to a string. A Plan object should override the `ToString` method instead if they want to instrument the planner result.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -49,4 +50,10 @@ public KernelArguments(KernelArguments other) : this(other as IDictionary<string
     /// Gets or sets the prompt execution settings
     /// </summary>
     public PromptExecutionSettings? ExecutionSettings { get; set; }
+
+    /// <summary>
+    /// Returns a comma-separated string that represents the kernel argument.
+    /// </summary>
+    /// <returns>A string that represents the kernel argument.</returns>
+    public override string ToString() => string.Join(", ", this.Select(v => $"{v.Key}:{v.Value}"));
```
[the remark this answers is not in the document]
alliscode: @TaoChenOSU would it work to create a ToLogMessage() method or something similar and call that explicitly in the attribute e.g. `Message = "Function {FunctionName} invoking with arguments {Arguments.ToLogMessage()}.")`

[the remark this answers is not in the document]
TaoChenOSU: Yes, this is a good idea too.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionLogExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// Logs invocation of a <see cref="KernelFunction"/>.
+    /// </summary>
+    [LoggerMessage(
+        EventId = 0,
+        Level = LogLevel.Trace,
+        Message = "Function {FunctionName} invoking with arguments: {Arguments}.")]
+    public static partial void LogFunctionInvokingWithArguments(
+        this ILogger logger,
+        string functionName,
+        KernelArguments arguments);
```
dmytrostruk, line 25: Taking into account that `KernelArguments` doesn't have `ToString` override, there is a chance that logging output for `Arguments` will be always `Microsoft.SemanticKernel.KernelArguments`. Maybe it worth to check it. Same applies to method `LogFunctionStreamingInvokingWithArguments`.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionLogExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    /// <summary>
+    /// Logs cancellation of a <see cref="KernelFunction"/>.
+    /// </summary>
+    [LoggerMessage(
+        EventId = 1,
+        Level = LogLevel.Trace,
```
dmytrostruk, line 32: I think that this message should have `Information` log level, since it doesn't contain sensitive information and should be useful to understand application flow, in this case - whether function was canceled or not. Same applies to method `LogFunctionComplete`.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionLogExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.SemanticKernel.Functions;
+
+/// <summary>
+/// Extensions for logging <see cref="KernelFunction"/> invocations.
+/// This extension uses the <see cref="LoggerMessageAttribute"/> to
+/// generate logging code at compile time to achieve optimized code.
+/// </summary>
+internal static partial class KernelFunctionLogExtensions
```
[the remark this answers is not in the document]
TaoChenOSU: I agree with you. I think `{ClassName}LogMessages` is more descriptive that `{ClassName}LogExtensions`.

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -36,39 +35,69 @@ internal static class PlannerInstrumentation
...
+    [LoggerMessage(
+        EventId = 4,
+        Level = LogLevel.Error,
+        Message = "Plan creation failed. Error: {Message}")]
+    static partial void LogPlanCreationError(this ILogger logger, Exception exception, string message);
+
+    [LoggerMessage(
+        EventId = 5,
+        Level = LogLevel.Information,
+        Message = "Plan creation duration: {Duration}ms.")]
```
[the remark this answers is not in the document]
TaoChenOSU: Good call!

dmytrostruk (approved):

alliscode (approved):

PR #3876, opened 2023-12-01
Make KernelArguments _wrap_ rather than _be_ a Dictionary<>

stephentoub:
Having it just implement the interfaces makes it more future proof should we ever want to do anything special in any of the methods in the implementation. It also lets us change "key" to "name". Also fix the nullability of its value and propagate those changes to address resulting warnings.

### DIFF
--- a/dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
+++ b/dotnet/src/Extensions/TemplateEngine.Handlebars/HandlebarsPromptTemplate.cs
@@ -64,14 +64,15 @@ internal sealed class HandlebarsPromptTemplate : IPromptTemplate
             }
         }
 
-        if (arguments == null)
+        if (arguments is not null)
         {
-            return result;
-        }
-
-        foreach (var kvp in arguments)
-        {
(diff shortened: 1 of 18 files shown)


markwallace-microsoft (approved):

SergeyMenshykh (approved):

matthewbolanos:
Never mind, I see that it still is IDictionary, so I’m assuming the collection initializer syntax will still work.

SergeyMenshykh:
> Does this make it harder to easily create arguments when invoking the kernel? Had trouble finding an example of creating arguments when invoking to see what it looks like

Nope, the arguments class creation semantic is the same: 
`kernel.InvokeAsync(function, new() {["p1"] = "v1"})`/`function.InvokeAsync(new() {["p1"] = "v1"}))`
or
`kernel.InvokeAsync(function, new() {{"p1", "v1"}})`/`function.InvokeAsync(new() {{"p1", "v1"}}))`

note: there's small issue, at the moment, with the kernel.InvokeAsync method - https://github.com/microsoft/semantic-kernel/pull/3752#discussion_r1410020677 that will be addressed separately.

RogerBarreto (approved):

PR #3852, opened 2023-11-30
Fix Assistant managment of Kernel completion service.

crickman:
### Motivation and Context
Experimental assistant API broken

### Description
Recent changes made the ITextCompletion service unreachable.  Creating service in Assistant constructor an overall simplicitation anyway.

https://github.com/microsoft/semantic-kernel/issues/3787

Also addressed JSON serialization that is no longer required (performance).

### Contribution Checklist
- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #3879, opened 2023-12-01
.Net: Remove KernelBuilder.ConfigurePlugins and WithCulture

stephentoub:
We'd added it for symmetry with ConfigureServices, but it actually leads to messier code (in its brief lifetime we've already seen it cause some confusion), and the asymmetry just reflects the existing asymmetry that Kernel.Plugins is mutable, so while you use KernelBuilder to compose what becomes an immutable set of services, plugins can be added once the Kernel has been constructed. I considered just having a KernelBuilder.Plugins property, but that brings other complications, namely that we want to just transfer the built up collection to the kernel, but because we want each built Kernel to have its own unique copy, we'd have to do more complicated tracking to ensure we didn't pass the same collection to two different builds while also avoiding unnecessary copying of the collection when it was only built up once.

I also removed WithCulture. It's also mutable on Kernel and can thus be changed after the Kernel is built, but it'll likely be rare to do so. And by getting rid of it, the only things that remain on KernelBuilder are about building up the immutable set of services.  All the builder was doing was setting the property after the Kernel was constructed, anyway.

We can always add back such things later.  Keeping it simpler for now.

### DIFF
--- a/dotnet/samples/KernelSyntaxExamples/Example42_KernelBuilder.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example42_KernelBuilder.cs
@@ -10,9 +10,9 @@ using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.SemanticKernel;
 using RepoUtils;
+using ILogger = Microsoft.Extensions.Logging.ILogger;
 
 // ReSharper disable once InconsistentNaming
 public static class Example42_KernelBuilder
@@ -43,19 +43,13 @@ public static class Example42_KernelBuilder
             })
             .Build();
 
-        // Plugins may also be configured via the corresponding ConfigurePlugins method. There are multiple
-        // overloads of ConfigurePlugins, one of which provides access to the services registered with the
-        // builder, such that a plugin can resolve and use those services.
-        Kernel kernel3 = new KernelBuilder()
-            .ConfigurePlugins((serviceProvider, plugins) =>
-            {
-                ILogger logger = serviceProvider.GetService<ILogger<KernelBuilder>>() ?? (ILogger)NullLogger.Instance;
-                plugins.Add(new KernelPlugin("Example", new[]
-                {
-                    KernelFunctionFactory.CreateFromMethod(() => logger.LogInformation("Function invoked"), functionName: "LogInvocation")
-                }));
-            })
-            .Build();
+        // Plugins may also be configured on the Kernel after it's been built. The Kernel's Plugins collection is mutable.
+        Kernel kernel3 = new KernelBuilder().Build();
+        ILogger logger = kernel3.GetService<ILoggerFactory>().CreateLogger<KernelBuilder>();
+        kernel3.Plugins.Add(new KernelPlugin("Example", new[]
+        {
+            KernelFunctionFactory.CreateFromMethod(() => logger.LogInformation("Function invoked"), functionName: "LogInvocation")
+        }));
 
         // Every call to KernelBuilder.Build creates a new Kernel instance, with a new service provider
         // and a new plugin collection.
--- a/dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
@@ -29,12 +29,9 @@ public static class Example59_OpenAIFunctionCalling
             {
                 services.AddLogging(services => services.AddConsole().SetMinimumLevel(LogLevel.Trace));
             })
-            .ConfigurePlugins(plugins =>
-            {
-                plugins.AddPluginFromObject<TimePlugin>();
-                plugins.AddPluginFromObject<WidgetPlugin>();
-            })
             .Build();
+        var time = kernel.ImportPluginFromObject<TimePlugin>();
+        var widget = kernel.ImportPluginFromObject<WidgetPlugin>();
 
         // Load additional functions into the kernel
         await kernel.ImportPluginFromOpenAIAsync("KlarnaShoppingPlugin", new Uri("https://www.klarna.com/.well-known/ai-plugin.json"));
@@ -44,7 +41,7 @@ public static class Example59_OpenAIFunctionCalling
         var executionSettings = new OpenAIPromptExecutionSettings();
 
         // Set FunctionCall to the result of FunctionCallBehavior.RequireFunction with a specific function to force the model to use that function.
-        executionSettings.FunctionCallBehavior = FunctionCallBehavior.RequireFunction(kernel.Plugins["TimePlugin"]["Date"].Metadata.ToOpenAIFunction(), autoInvoke: true);
+        executionSettings.FunctionCallBehavior = FunctionCallBehavior.RequireFunction(time["Date"].Metadata.ToOpenAIFunction(), autoInvoke: true);
         await CompleteChatWithFunctionsAsync("What day is today?", chatHistory, chatCompletion, kernel, executionSettings);
 
         // Set FunctionCall to FunctionCallBehavior.ProvideKernelFunctions to let the model choose the best function to use from all available in the kernel.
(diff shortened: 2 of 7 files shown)


markwallace-microsoft:
I think removing these makes creating a `Kernel` more awkward because you can only do some things with the `KernelBuilder` but people do have to use `KernelBuilder`  to ensure `Kernel.GetAllServices` works properly. What about adding a `KernelBuilder.WithPlugins` method?

Below is the reference sample that @matthewbolanos is sharing and I've updated it to use `WithPlugins`

``` csharp
// Define the LightPlugin class
class LightPlugin {
    public bool IsOn { get; set; } = false;

    [KernelFunction("GetState")]
    [Description("Gets the state of the light.")]
    public string GetState() => IsOn ? "on" : "off";

    [KernelFunction("Light.ChangeState")]
    [Description("Changes the state of the light.")]
    public string ChangeState(bool newState) => {
        IsOn = newState;
        return IsOn;
    };
}

// Create the kernel
Kernel kernel = new KernelBuilder()
    .WithOpenAIChatCompletion("gpt-3.5-turbo-1106", apiKey)
    .WithPlugins(KernelPluginFactory.CreateFromObject<LightPlugin>())
    .Build();

// Invoke the prompt
OpenAIPromptExecutionSettings settings = new() {
    FunctionCallBehavior = FunctionCallBehavior.AutoInvokeKernelFunctions
};
Console.WriteLine(await kernel.InvokePromptAsync("""
    <message role="user">Can you toggle the light?</message>
    <message role="system">The lights are currently {{LightPlugin.GetState}}.</message>
"""},
settings));
```

matthewbolanos:
@stephentoub, are you able to share an example of what the messier code looks like? Is it because you have to retrieve the function from the kernel?

if so, could we instead use overloads that accept a plugin name and function name?

```
executionSettings.FunctionCallBehavior = FunctionCallBehavior.RequireFunction("Time", "Date", autoInvoke: true);
```

stephentoub:
> are you able to share an example of what the messier code looks like? Is it because you have to retrieve the function from the kernel?

e.g. https://github.com/microsoft/semantic-kernel/pull/3852#pullrequestreview-1758503927

> What about adding a KernelBuilder.WithPlugins method?

We _could_. But it means:
a) The various Import/Add extension methods aren't available.
b) It'll make every use of KernelBuilder with plugins more expensive, as we'll need to build up the collection and then clone it to pass the clone to the Kernel, even though the vast majority use case is building only one Kernel.

Also, neither it nor the current ConfigurePlugins methods support asynchronous plugin creation (we have a case of that with one or more of the OpenAPI-based plugins).

> could we instead use overloads that accept a plugin name and function name to clean things up where necessary?

I'm not following. What would using that look like end-to-end?

stephentoub:
> Below is the reference sample that @matthewbolanos is sharing and I've updated it to use WithPlugins

Yes, instead of:
```C#
Kernel kernel = new KernelBuilder()
    .WithOpenAIChatCompletion("gpt-3.5-turbo-1106", apiKey)
    .WithPlugins(KernelPluginFactory.CreateFromObject<LightPlugin>())
    .Build();
```
it's:
```C#
Kernel kernel = new KernelBuilder()
    .WithOpenAIChatCompletion("gpt-3.5-turbo-1106", apiKey)
    .Build();
kernel.ImportPluginFromObject<LightPlugin>();
```

stephentoub:
If we really want to keep plugins on the builder, I suggest we drop the callbacks for both services and plugins and just have properties, e.g.
```C#
KernelBuilder builder = new KernelBuilder()
    .WithOpenAIChatCompletion("gpt-3.5-turbo-1106", apiKey);
builder.Plugins.AddFromObject<LightPlugin>();
Kernel kernel = builder.Build();
```
We lose the fluency, so you need the builder local, and we'll need to clone the plugins collection when building (or say that Build can be used only once per KernelBuilder), but it keeps plugins and services largely on the same footing and allows you to do everything through the builder.

I'm still of the mind that just keeping the builder about the immutable services is the right thing. Everything else on Kernel is mutable (plugins, event handlers, data, culture), and the expectation then is the developer is mutating the post-built kernel for such mutable things.

Just let me know which direction you'd like to go.

matthewbolanos:
### Could we have both?
Regarding https://github.com/microsoft/semantic-kernel/pull/3852#pullrequestreview-1758503927,

Could we not support both? If you want to use `ConfigurePlugins` and/or the Builder, you could. Or if you wanted to use `kernel.AddPlugins` since the plugin collection is mutable, you'd be able to do that as well.

The reason I'm a fan of `ConfigurePlugins` and the Builder methods is that they...
- Encourages defining your plugins once and then injecting the kernel everywhere you need it
- Reinforces that a kernel is a property bag of AI services and plugins

This update makes plugins feel like a second-class citizen from the kernel perspective when it's one of the most important aspects / value adds.

```csharp
Kernel kernel = new KernelBuilder()
    .WithOpenAIChatCompletion("gpt-3.5-turbo-1106", apiKey)
    .Build();
kernel.ImportPluginFromObject<LightPlugin>();
```

### Matthew's crazy hack to make plugin name / function name work...
> What would using that look like end-to-end?

For `FunctionCallBehavior.RequireFunction`, I'd imagine it could look something like the following (not a fan of the implementation though...)

1. Instead requiring a `FunctionDefinition`, have two methods: one that accepts a `KernelFunction` and another that accepts a plugin name and function name.
    ```csharp
    public static FunctionCallBehavior RequireFunction(KernelFunction function, bool autoInvoke = false)
    {
        Verify.NotNull(function);
        return new RequiredFunction(function, autoInvoke);
    }
    
    public static FunctionCallBehavior RequireFunction(string pluginName, string functionName, bool autoInvoke = false)
    {
        Verify.NotNull(function);
        return new RequiredFunction(pluginName, functionName, autoInvoke);
    }
    ```
2. The internal `RequiredFunction` class would need to store this information.
    ```csharp
    internal sealed class RequiredFunction : FunctionCallBehavior
    {
        private AmbiguousFunction[]? _functionArray;

        public RequiredFunction(KernelFunction function, bool autoInvoke) : base(autoInvoke)
        {
            this.Function = new AmbiguousFunction(function);
        }

        public RequiredFunction(string pluginName, string functionName, bool autoInvoke) : base(autoInvoke)
        {
            this.Function = new AmbiguousFunction(pluginName, functionName);
        }

        public AmbiguousFunction Function { get; }

        /// <summary>Gets the target function wrapped in an array.</summary>
        public AmbiguousFunction[] FunctionArray => this._functionArray ??= new[] { this.Function };

        public override string ToString() => $"RequiredFunction(autoInvoke:{this.AutoInvoke}): {this.Function.Name}";
    }

    /// <summary>contains the necessary information to resolve to a specific kernel function.</summary>
    internal sealed class AmbiguousFunction
    {
        public KernelFunction? Function { get; }
        public string? PluginName { get; }
        public string Name { get; }

        public AmbiguousFunction(KernelFunction function)
        {
            this.Function = function;
            this.Name = function.Name;
        }

        public AmbiguousFunction(string pluginName, string functionName)
        {
            this.PluginName = pluginName;
            this.Name = functionName;
        }
    }
    ```
4. Finally, in the chat completion, use the available kernel to resolve the functions where only the plugin name and functino names were provided.
    ```csharp
    case FunctionCallBehavior.RequiredFunction rufb:
        if (rufb is null)
        {
            throw new ArgumentNullException(nameof(executionSettings.FunctionCallBehavior));
        }
    
        // 1. Resolve the Kernel Function if only plugin name and function name are provided
        // 2. Turn the Kernel Function into a FunctionDefinition
        // 3. Set the options

        break;
    ````

stephentoub:
> Could we not support both?

That's what we have today in main. My primary concern is because we have this more convoluted mechanism on the KernelBuilder, folks are encouraged to do that even though there's a simpler mechanism post-Build().

> Encourages defining your plugins once and then injecting the kernel everywhere you need it with DI

I don't think that's really affected. If you're using KernelBuilder in DI, then you're doing something like this:
```C#
services.AddSingleton<Kernel>(serviceProvider =>
{
    return new KernelBuilder()
        .WithStuff(...);
        .ConfigurePlugins(plugins => plugins.Add(...));
        .Build();
});
```
and with post-Build() it's instead:
```C#
services.AddSingleton<Kernel>(serviceProvider =>
{
    var builder = new KernelBuilder()
        .WithStuff(...);
    builder.Plugins.Add(...);
    return builder.Build();
});
```

> This update makes plugins feel like a second-class citizen from the kernel perspective

That's fair.  Though they are fundamentally different, in that the services are the one immutable thing on the kernel, whereas everything else (including plugins) are mutable. So the builder ends up being about building the things that can't be mutated later.

> Matthew's crazy hack to make plugin name / function name work

I'm confused. I thought you were proposing a mechanism for getting the relevant plugins into the kernel. Are you instead talking about getting functions from the kernel as part of automatic function calling? I think it'd be fine to augment the FunctionCallBehavior stuff with such a mechanism, but that's orthogonal to the discussion in this thread.

> Reinforces that a kernel is a property bag of AI services and plugins

Ok, I'll close this PR. I'll open a new one that tries to address a few of my smaller concerns.

PR #3895, opened 2023-12-01
.Net: FunctionMetadata isn't populating PluginName 

### LINKED ISSUE #3855: .Net: FunctionMetadata isn't populating PluginName
**Describe the bug**

PluginName isn't populated in Function.Metadata in kernel object (see screenshot below) though Plugin.Name is populated at the top level.

**To Reproduce**
1. Run Example65_HandlebarsPlanner in KernelSyntaxExamples
2. Put a breakpoint at line 95 (after plugin import)
3. Inspect the kernel.Plugins[0].Functions[0]
4. Look at the `Metadata` property

**Expected behavior**

PluginName should be populated in Function.Metadata

It's populated when you explicitly call `GetFunctionsMetadata`, but I would expect this would be initialized when the function is created.
image

**Screenshots**

image

**Platform**
 - OS: Windows
 - IDE:VS Code
 - Language: C#
 - Source: V1 Release Candidate

**Additional context**

markwallace-microsoft:
### Motivation and Context

Closes #3855 

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Functions/IKernelPluginExtensions.cs
```diff
@@ -112,7 +112,7 @@ public static IList<KernelFunctionMetadata> GetFunctionsMetadata(this IEnumerabl
         {
             foreach (KernelFunction function in plugin)
             {
-                metadata.Add(new KernelFunctionMetadata(function.Metadata) { PluginName = plugin.Name });
+                metadata.Add(function.Metadata);
```
stephentoub, line 115: Shouldn't the metadata we add here contain the plugin name?

dotnet/src/SemanticKernel.Core/Functions/KernelPlugin.cs
```diff
@@ -80,6 +79,7 @@ public KernelPlugin(string name, string? description, IEnumerable<KernelFunction
     public void AddFunction(KernelFunction function)
     {
         Verify.NotNull(function);
+        function.Metadata.PluginName = this.Name;
```
stephentoub, line 82: This seems problematic. What happens if the same KernelFunction is added to two different plugins?

dotnet/src/SemanticKernel.Abstractions/Functions/IKernelPluginExtensions.cs
markwallace-microsoft: Will revert this change

dotnet/src/SemanticKernel.Core/Functions/KernelPlugin.cs
markwallace-microsoft: @matthewbolanos The issue #3855 here is that when we add functions to a plugin the function meteadata doesn't get assigned the plugin name i.e. it only happens if you call `GetFunctionsMetadata` and this is an extension method.

The proposed fix would be to set the plugin name when the function is added to the plugin but this causes the issue Stephen identified. So we can

1. Leave things as they are but then the function metadata is inconsistent
2. Prevent the same function being added to two plugins i.e. throw an exception if someone attempts this

Which behaviour should we aim for?

matthewbolanos: Let’s stick with option 1. I can then call it out in docs as a limitation. The preferred route for calling functions should probably be through the `string pluginName, string functionName` `InvokeAsync` method since we get the plugin name at that point.

can we confirm the metadata is populated for that code path?

PR #3909, opened 2023-12-03
Change how event handlers cancel, and clean up FunctionResult

stephentoub:
- Previously, an event handler canceling the function would result in the function still returning a FunctionResult, but that object would have IsCancellationRequested returning true.  That has a couple of problems. First, it means a developer consuming a FunctionResult now has two different ways to care about cancellation (an OperationCanceledException getting thrown, and that property being set), which means effectively every function invocation needs to check that property. To partially remediate this, the FunctionResult.GetValue method was made to check this property and throw an OperationCanceledException if it's true, but that's not a reliable way to indicate cancellation if the result is never consumed, and it means a developer can't observe the result of the function that caused a post-hook to cancel (possibly based on observing that very result). To fix all of that, this commit removes that IsCancellationRequested property from FunctionResult (as well as two other Is properties that were apparently accidentally left over when repeating/skipping was removed), and instead models cancellation from event handlers as OCEs as well. Further, all OCEs coming out of a function invocation are now actually a derived KernelFunctionCanceledException, which contains the result of the operation if cancellation occurred after the result was produced. This gives a developer a way to observe it if they want to, while still making it so a developer doesn't have to check IsCancellationRequested after every invocation. The exception carries with it additional information to help in diagnosing a failure.
- FunctionResult is now simplified: all data is passed into its constructor, now internal setters. TryGetMetadata is also removed: if someone wants to examine the metadata, they just examine the Metadata, which is now a nullable dictionary, making it easy to see whether there is any metadata without forcing the dictionary into existence.
- Remove from KernelFunctionFromPrompt.InvokeCoreAsync the try/catch; that logging is handled by the base implementation.
- Changed KernelFunction.InvokeCoreAsync to return a ValueTask. We know all consumers (there's only the one in InvokeCoreAsync), and it's common for KernelFunctionFromMethod methods to complete synchronously with a value. This then lets us change KernelFunctionFromMethod's implementation to not await and instead just return the result of invoking the implementation delegate.
- Updated XML comments


RogerBarreto (approved):
LGTM

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -113,13 +113,12 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
     /// <param name="arguments">The function arguments.</param>
     /// <returns>The updated context, potentially a new one if context switching is implemented.</returns>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <exception cref="OperationCanceledException">The <see cref="KernelFunction"/>'s invocation was canceled.</exception>
```
markwallace-microsoft, line 116: @stephentoub should we also list `KernelFunctionCanceledException` here?

PR #3911, opened 2023-12-03
.Net: Preparation for non-string arguments support

SergeyMenshykh:
This PR changes SK public API surface to lay the foundation for the long-awaited feature - non-string arguments. As part of this PR the following major changes were done:

- The `KernelArguments` class is changed to support non-string arguments. Now, it's a dictionary of nullable objects rather than strings. 
- The `KernelFunctionFromMethod` class is refactored to <br> **a)** marshal function arguments as they are, without changing their type. The existing logic of converting arguments represented as strings to target parameter types is retained to provide more flexibility for SK consumer code to supply arguments as strings. <br> **b)** remove support for default values provided via `DefaultValueAttribute` in favor of native C# mechanism of optional parameters that allows default values to be provided as - `...., string p1 = "abc"`.
-  The `KernelPromptTemplate`, `RestApiOperationRunner`, `GrpcOperationRunner` were modified to include an argument type check that only allows string arguments and rejects arguments of other types. These temporary safeguards will be removed when these classes are refactored to support non-string arguments.


stephentoub:
> a) marshal function arguments as they are, without changing their type. The existing logic of converting arguments represented as strings to target parameter types is retained for backward compatibility.

Aren't we going to always need this, and in fact expand it? We'll want to convert from whatever is in the arguments collection to the parameter type.

stephentoub:
> enables implicit type conversion (e.g., int to long, float to double) performed by runtime, which would otherwise require additional code to support.

Is it actually meaningfully more code? As noted above, I was expecting we'd _always_ need marshaling code to go from argument to parameter, and we'd get that effectively for free for these default values then as well.

dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
```diff
@@ -45,6 +45,9 @@ public KernelPromptTemplate(PromptTemplateConfig promptConfig, ILoggerFactory? l
     /// <inheritdoc/>
     public async Task<string> RenderAsync(Kernel kernel, KernelArguments? arguments = null, CancellationToken cancellationToken = default)
     {
+        //Make sure all arguments are of string type. This is temporary check until non-string arguments are supported.
```
markwallace-microsoft, line 48: nit: space after `//`

dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
```diff
@@ -136,5 +139,29 @@ internal IList<Block> RenderVariables(IList<Block> blocks, KernelArguments? argu
...
+        }
+
+        foreach (var argument in arguments)
+        {
+            if (argument.Value is string)
+            {
+                continue;
+            }
+
+            throw new KernelException($"Kernel prompt template arguments must be of type string. Argument '{argument.Key}' is of type '{argument.Value?.GetType()}'");
```
markwallace-microsoft, line 162: Let's explain in the message (and the others) that this is a temporary state. 

E.g. `Non string kernel prompt template arguments are not supported in release candidate 1. This feature is coming soon but for now please make all arguments strings.`

@matthewbolanos Can you help here with the exact message to use?

Would this be a better exception: https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-8.0

dotnet/src/Functions/Functions.Grpc/GrpcOperationRunner.cs
```diff
@@ -73,14 +77,34 @@ public async Task<JsonObject> RunAsync(GrpcOperation operation, KernelArguments
...
+    private static Dictionary<string, string> CastToStringArguments(KernelArguments arguments, GrpcOperation operation)
+    {
+        return arguments.ToDictionary(item => item.Key, item =>
+        {
+            if (item.Value is string stringValue)
+            {
+                return stringValue;
+            }
+
+            throw new KernelException($"gRPC operation '{operation.Name}' arguments must be of type string. Argument '{item.Key}' is of type '{item.Value?.GetType()}'.");
```
markwallace-microsoft, line 104: Explain this is a temporary limitation

markwallace-microsoft (commented):
Change exception message to explain this is a temporary limitation

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperation.cs
```diff
@@ -133,7 +133,7 @@ public Uri BuildOperationUrl(IDictionary<string, string?> arguments, Uri? server
             var headerValue = header.Value;
 
             //A try to resolve header value in arguments.
-            if (arguments.TryGetValue(headerName, out string? value) && !string.IsNullOrEmpty(value))
+            if (arguments.TryGetValue(headerName, out string value) && !string.IsNullOrEmpty(value))
```
stephentoub, line 136: Does anything here guarantee that the value string is non-null? The juxtaposition between `out string value` but then checking `string.IsNullOrEmpty` is strange to me. I'd expect that either we'd keep it `out string? value` or we'd change the latter condition to be `value.Length != 0` if it's somehow impossible the value could be null.

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperation.cs
```diff
@@ -177,14 +177,14 @@ public Uri BuildOperationUrl(IDictionary<string, string?> arguments, Uri? server
...
-    private string ReplacePathParameters(string path, IDictionary<string, string?> arguments)
+    private string ReplacePathParameters(string path, IDictionary<string, string> arguments)
     {
         string ReplaceParameter(Match match)
         {
             var parameterName = match.Groups[1].Value;
 
             //A try to find parameter value in arguments
-            if (arguments.TryGetValue(parameterName, out string? value) && value is not null)
+            if (arguments.TryGetValue(parameterName, out string value) && value is not null)
```
stephentoub, line 187: Same question for all of these.

(If the arguments comes in via a public API where we haven't validated non-nullness of the arguments, then it's still possible they're null.)

dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
```diff
@@ -136,5 +139,29 @@ internal IList<Block> RenderVariables(IList<Block> blocks, KernelArguments? argu
...
+        {
+            return;
+        }
+
+        foreach (var argument in arguments)
+        {
+            if (argument.Value is string)
+            {
+                continue;
+            }
```
stephentoub, line 160: Shouldn't this go through the same TypeConverter logic as with KernelFunctionFromMethod, just with a destination type for now of string?

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/VarBlock.cs
```diff
@@ -73,9 +73,9 @@ public string Render(KernelArguments? arguments)
             throw new KernelException(ErrMsg);
         }
 
-        if (arguments.TryGetValue(this.Name, out string? value))
+        if (arguments.TryGetValue(this.Name, out object? value))
         {
-            return value ?? string.Empty;
+            return (string?)value ?? string.Empty;
```
stephentoub, line 78: This is guaranteed to be null or a string?

dotnet/src/Functions/Functions.OpenAPI/Builders/QueryStringBuilder.cs
```diff
@@ -24,15 +24,15 @@ internal static class QueryStringBuilder
...
+    public static string BuildQueryString(this RestApiOperation operation, IDictionary<string, string> arguments)
     {
         var segments = new List<string>();
 
         var parameters = operation.Parameters.Where(p => p.Location == RestApiOperationParameterLocation.Query);
 
         foreach (var parameter in parameters)
         {
-            if (!arguments.TryGetValue(parameter.Name, out string? argument) || argument is null)
+            if (!arguments.TryGetValue(parameter.Name, out string argument) || argument is null)
```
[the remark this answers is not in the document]
SergeyMenshykh: I don't see any way that nulls can slip through the null checks before OpenAPI function execution [here](https://github.com/microsoft/semantic-kernel/blob/0f23dc986d243624b44549c5426cb87f0f6d16c2/dotnet/src/Functions/Functions.OpenAPI/Extensions/OpenApiKernelExtensions.cs#L286C12-L286C67). However, I have rolled back the change for now and will revise it later in the scope of supporting non-string arguments.

dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperation.cs
SergeyMenshykh: I don't see any way that nulls can slip through the null checks before OpenAPI function execution [here](https://github.com/microsoft/semantic-kernel/blob/0f23dc986d243624b44549c5426cb87f0f6d16c2/dotnet/src/Functions/Functions.OpenAPI/Extensions/OpenApiKernelExtensions.cs#L286C12-L286C67). However, I have rolled back the change for now and will revise it later in the scope of supporting non-string arguments.

dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
SergeyMenshykh: Yes, certainly, it can be done this way. However, I have not done it for two reasons - I would like to keep this PR relatively short, and I plan to add non-string support to the kernel prompt template in one of the following PRs.

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/VarBlock.cs
SergeyMenshykh: Together with the temporary safeguard below it can't be anything except string. The explicit - (string?) casting needed to make the compiler happy. The code will be refactored later is scope of supporting non-string arguments by the kernel prompt template.
image

SergeyMenshykh:
> > enables implicit type conversion (e.g., int to long, float to double) performed by runtime, which would otherwise require additional code to support.
> 
> Is it actually meaningfully more code? As noted above, I was expecting we'd _always_ need marshaling code to go from argument to parameter, and we'd get that effectively for free for these default values then as well.

As per our offline discussion, custom marshaling of method arguments is necessary to respect the TypeConverterAttribute on complex types and support type conversions, such as Guid to string, which are not supported by the runtime out of the box. This is in addition to implicit conversions like int/float to double, and int to long, etc..

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
```diff
@@ -295,110 +295,72 @@ private static bool IsAsyncMethod(MethodInfo method)
...
 
-        if (!type.IsByRef && GetParser(type) is Func<string, CultureInfo, object> parser)
-        {
-            string name = SanitizeMetadataName(parameter.Name ?? "");
-            bool nameIsInput = name.Equals(KernelArguments.InputParameterName, StringComparison.OrdinalIgnoreCase);
-            ThrowForInvalidSignatureIf(name.Length == 0, method, $"Parameter {parameter.Name}'s attribute defines an invalid name.");
-            ThrowForInvalidSignatureIf(sawFirstParameter && nameIsInput, method, "Only the first parameter may be named 'input'");
+        string name = SanitizeMetadataName(parameter.Name ?? "");
+        bool nameIsInput = name.Equals(KernelArguments.InputParameterName, StringComparison.OrdinalIgnoreCase);
+        ThrowForInvalidSignatureIf(name.Length == 0, method, $"Parameter {parameter.Name}'s attribute defines an invalid name.");
```
RogerBarreto, line 302:
```suggestion
        ThrowForInvalidSignatureIf(string.IsNullOrWhiteSpace(name), method, $"Parameter {parameter.Name}'s attribute defines an invalid name.");
```

RogerBarreto (approved):
LGTM

markwallace-microsoft (approved):

PR #3919, opened 2023-12-04
Python: Add python connector for Azure OpenAI on your data

sarah-widder:
### Motivation and Context

This change adds a python connector for Azure OpenAI chat completions with data, which was already supported in the .NET SDK. 

### Description

Add connector for Azure OpenAI chat completions with data and associated sample code for how to use it. Some minor changes were made to the SK function to handle the "tool" message that contains the citations and other information returned from the chat completions extensions API.

### Contribution Checklist


moonbox3:
Hi @sarah-widder, you'll want to take a look at the `Merge Gatekeeper` and `Python Lint / ruff` failed steps to see some errors to fix. If you haven't installed the Python package called `black` you should do so. You can then run `black .` in the `semantic-kernel/python` directory and it'll help modify some of the files that are failing checks. You can also install `ruff` to run some similar checks. That'll help remove some references in your files to imports that aren't used.

python/semantic_kernel/semantic_functions/chat_with_data_prompt_template.py
```diff
@@ -0,0 +1,35 @@
...
+if TYPE_CHECKING:
+    from semantic_kernel.orchestration.sk_context import SKContext
+
+ChatMessageT = TypeVar("ChatMessageT", bound=ChatMessage)
+
+
+class ChatWithDataPromptTemplate(ChatPromptTemplate):
+    _messages: List[ChatMessageT]
+
+    def __init__(
```
moonbox3, line 24: nit: could we add a doctstring here, please? It's helpful to provide context to readers/users.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion_with_data.py
```diff
@@ -0,0 +1,303 @@
...
+            api_version=api_version,
+            api_key=api_key,
+            log=log,
+            ad_token=ad_token,
+            ad_token_provider=ad_token_provider,
+        )
+
+        self._data_source_settings = data_source_settings
+
+    async def complete_chat_async(
```
moonbox3, line 86: Is there any way we can re-use the code that exists in `open_ai_chat_completion_base.py`? Most of the code shown here like `complete_chat_async`, `complete_chat_stream_async`, `_parse_choices` and others exist there.

python/semantic_kernel/connectors/ai/open_ai/models/chat/azure_chat_with_data_response.py
```diff
@@ -0,0 +1,72 @@
...
+            message += choice.delta.content
+
+        return message, choice.index
+
+    async def __aiter__(self):
+        """Iterate through the message stream and yield the assistant response."""
+        if not self._tool_message:
+            await self._iterate_to_assistant_message()
+
+        async for chunk in self._assistant_message:
```
moonbox3, line 59: Similar to a previous comment, we have this code in `async def complete_chat_stream_async` as part of `OpenAIChatCompletionBase`.

python/semantic_kernel/connectors/ai/open_ai/services/azure_config_base.py
```diff
@@ -41,7 +41,7 @@ def __init__(
             )
         if base_url:
             client = AsyncAzureOpenAI(
-                base_url=base_url,
+                base_url=str(base_url),
```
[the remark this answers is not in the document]
sarah-widder: the OpenAI SDK was throwing an error with the pydantic HttpUrl type, so I cast it to a string. Maybe this is not expected?

python/semantic_kernel/connectors/ai/open_ai/models/chat/azure_chat_with_data_response.py
sarah-widder: In this case, the stream returned from the extensions chat completions API isn't just the assistant message deltas - the first message will be the "tool" message, then a message indicating role "assistant", and then the assistant message deltas start from the next message. So this class is intended to wrap the stream so that the first few events are processed to get out the tool message, and then the assistant message deltas can be consumed by the client in an `async for`. If we use the base class there's no way for the client to access the tool message.

python/semantic_kernel/semantic_functions/chat_with_data_prompt_template.py
```diff
@@ -0,0 +1,35 @@
...
+)
+
+if TYPE_CHECKING:
+    from semantic_kernel.orchestration.sk_context import SKContext
+
+ChatMessageT = TypeVar("ChatMessageT", bound=ChatMessage)
+
+
+class ChatWithDataPromptTemplate(ChatPromptTemplate):
+    _messages: List[ChatMessageT]
```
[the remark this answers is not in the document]
sarah-widder: Not needed - removed!

python/semantic_kernel/connectors/ai/open_ai/services/azure_config_base.py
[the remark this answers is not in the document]
eavanvalkenburg: I saw this, added a test and this same fix into a seperate PR.

python/semantic_kernel/orchestration/sk_function.py
```diff
@@ -152,12 +152,19 @@ async def _local_func(client, request_settings, context: "SKContext", **kwargs):
...
-                        completion,
-                        function_call,
-                    ) = await client.complete_chat_with_functions_async(
+                    outputs = await client.complete_chat_with_functions_async(
                         messages, functions, request_settings
                     )
+                    if function_config.has_chat_with_data_prompt:
+                        completion, tool_message, function_call = outputs
+                        if tool_message:
+                            context.objects["tool_message"] = tool_message
```
eavanvalkenburg, line 161: This looks good for now, for v1 we plan to get rid of the current way of working with the context, but for now this works.

python/semantic_kernel/semantic_functions/chat_with_data_prompt_template.py
```diff
@@ -0,0 +1,32 @@
...
+    PromptTemplateConfig,
+)
+from semantic_kernel.template_engine.protocols.prompt_templating_engine import (
+    PromptTemplatingEngine,
+)
+
+ChatMessageT = TypeVar("ChatMessageT", bound=ChatMessage)
+
+
+class ChatWithDataPromptTemplate(ChatPromptTemplate):
```
eavanvalkenburg, line 18: This would have to be inside the connectors.ai.openai folder as it is Azure OpenAI specific

python/semantic_kernel/connectors/ai/open_ai/models/chat/azure_chat_with_data_settings.py
```diff
@@ -0,0 +1,36 @@
...
+from enum import Enum
+from typing import Any, Dict, Optional
+
+class DataSourceType(Enum):
+    """Enum to hold data source type."""
+
+    AZURE_AI_SEARCH = "AzureCognitiveSearch"
+
+@dataclass
+class DataSourceParametersBase:
```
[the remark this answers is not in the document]
eavanvalkenburg: Indeed, prefered is a pydantic class, using semantic_kernel.sk_pydantic.SKBaseModel

python/semantic_kernel/connectors/ai/open_ai/models/chat/azure_chat_with_data_settings.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class DataSourceType(Enum):
+    """Enum to hold data source type."""
+
+    AZURE_AI_SEARCH = "AzureCognitiveSearch"
+
+
+@dataclass
+class DataSourceParametersBase:
```
eavanvalkenburg, line 15: Why is this base class empty? then what does it add? I prefer not to add classes that are only there because in the future there might be others, unless you already add a second subclass.

eavanvalkenburg (changes requested):
I would really prefer to not have an additional AzureChatCompletion class, instead a azureopenai specific request setting and output parser should be enough (combined with proper base_url support), I'm currently working on revamping the request settings and that is where this becomes much easier to add, then we only need a way to check the result when parsing in the openai_chat_completion_base file to handle the references, in the api definition there are also other extensions being mentioned (for instance OCR) so looks like the structure of the result might change with the new api (2023-12 and onward. In general the service does not need to know what it supports, if you send in a request that uses the extra_body it is used, and then the output should just be properly parsed based on what's in there.

sarah-widder:
> I would really prefer to not have an additional AzureChatCompletion class, instead a azureopenai specific request setting and output parser should be enough (combined with proper base_url support), I'm currently working on revamping the request settings and that is where this becomes much easier to add, then we only need a way to check the result when parsing in the openai_chat_completion_base file to handle the references, in the api definition there are also other extensions being mentioned (for instance OCR) so looks like the structure of the result might change with the new api (2023-12 and onward. In general the service does not need to know what it supports, if you send in a request that uses the extra_body it is used, and then the output should just be properly parsed based on what's in there.

@eavanvalkenburg how would you recommend to handle the output parsing? According to the request setting, e.g. knowing this is a 'with data' call? My concern is that for the streaming case, there wouldn't be a way to access the initial context message (which for data, returns a "tool" message with citations). This is why I wrapped the output to make the initial message processed and accessible and then iterate through the assistant messages.

eavanvalkenburg:
@sarah-widder I think we need a version of the openai_chat_completion_base that has those specifically for Azure, since the extension endpoint requires that, and @moonbox3 is also looking at adding support for assistants which at this time are OpenAI specific, so splitting that will cause some code duplication in the short term but will allow this to work with the same endpoint and just a additional set of request settings, there is currently a complete_chat_with_function_async method, you can follow that pattern to do the parsing there, I plan to merge that into complete_chat_async in the future so if the code is already there then that works!

sarah-widder:
@eavanvalkenburg I see, so we could create methods like `complete_chat_with_data_async`, `complete_chat_with_data_and_functions`, and `complete_chat_stream_with_data_async` - and have the streaming one return the AzureChatWithDataResponse wrapper I mentioned that will process the tool message? Seems reasonable I think. We'll work on this.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion_with_data.py
```diff
@@ -0,0 +1,369 @@
...
+            Union[str, List[str]] -- The completion result(s).
+        """
+        response = await self._send_request(
+            messages=messages, request_settings=settings, stream=True
+        )
+
+        return AzureChatWithDataStreamResponse(response, settings)
+
+    @staticmethod
+    def _parse_choices(choice) -> Tuple[str, int]:
```
johnliu55-msft, line 212: This method looks exactly like the original implementation in `open_ai_chat_completion_base.OpenAIChatCompletionBase` which this class inherited. Do we need to reimplement this here?

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion_with_data.py
```diff
@@ -0,0 +1,369 @@
...
+            ) from ex
+        if not isinstance(response, AsyncStream):
+            if response.usage:
+                self.log.info(f"OpenAI usage: {response.usage}")
+                self.prompt_tokens += response.usage.prompt_tokens
+                self.completion_tokens += response.usage.completion_tokens
+                self.total_tokens += response.usage.total_tokens
+        return response
+
+    def _create_model_args(
```
johnliu55-msft, line 265: nit: I think most of the codes here duplicates with the original implementation, maybe using a `super()._create_model_args()` would be better?

python/semantic_kernel/connectors/ai/open_ai/models/chat/azure_chat_with_data_settings.py
sarah-widder: There will be others in the future - actually the AOAI with data API already supports several others, but they are in preview and lower priority than AI search for this change. I can remove the base class for now.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion_with_data.py
sarah-widder: Yeah, I'm moving all of this logic into `open_ai_chat_completion_base.py` since much is reused.

sarah-widder:
Hi @eavanvalkenburg , @johnliu55-msft thanks for your reviews. I refactored to move the with-data handling into the base handler instead of creating a separate service. Please take another look when you have some time and let me know your feedback. I want to add integration tests at some point as well but this will require setting up a search service accessible from the Github workflow, still working on that.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion_with_data.py
sarah-widder: Refactored to use the base class.

python/semantic_kernel/orchestration/sk_function.py
sarah-widder: From the with-data side, the main thing we need is for the tool message to be in the chat message list (it contains the intent, which is used in future retrieval calls). I added it to the context in case it would need to be used there, but I can take it out if it's not needed.

python/semantic_kernel/connectors/ai/open_ai/services/azure_config_base.py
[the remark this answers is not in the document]
sarah-widder: Thanks!

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_handler.py
```diff
@@ -149,6 +161,32 @@ def _create_model_args(
                 ]
             else:
                 model_args["functions"] = functions
+
+        if request_settings.data_source_settings is not None:
+            model_args["extra_body"] = {
```
eavanvalkenburg, line 166: could you make this part of the data source settings class? and maybe you can then use the model_dump setting of pydantic, that way different versions of this (for AI search and for Cosmos/Mongo can work the same way more easily

python/semantic_kernel/orchestration/sk_function.py
eavanvalkenburg: Understood, at some point I would like to introduce a class to hold responses so that we do not have to deal with this kind of ugly tuple, but that is for later!

python/semantic_kernel/orchestration/sk_function.py
```diff
@@ -166,9 +183,24 @@ async def _local_func(client, request_settings, context: "SKContext", **kwargs):
                     if function_call is not None:
                         context.objects["function_call"] = function_call
                 else:
-                    completion = await client.complete_chat_async(
-                        messages, request_settings
-                    )
+                    if request_settings.data_source_settings is not None and hasattr(
+                        client, "complete_chat_with_data_async"
```
eavanvalkenburg, line 187: What happens with OpenAI Client classes with this?

python/semantic_kernel/orchestration/sk_function.py
```diff
@@ -189,11 +221,28 @@ async def _local_stream_func(client, request_settings, context):
...
+                        client, "complete_chat_stream_with_data_async"
                     ):
-                        completion += partial_content
-                        yield partial_content
+                        response = await client.complete_chat_stream_with_data_async(
+                            messages, request_settings
+                        )
+                        async for partial_content in response:
+                            completion += partial_content
+                            yield partial_content
```
eavanvalkenburg, line 234: this will loop through the content, but then you will not have the sources in the front-end until the full response is in, are the sources also streamed?

python/semantic_kernel/utils/settings.py
```diff
@@ -246,3 +246,33 @@ def azure_aisearch_settings_from_dot_env() -> Tuple[str, str]:
     assert api_key is not None, "Azure AI Search API key not found in .env file"
 
     return api_key, url
+
+
+def azure_aisearch_datasource_settings_from_dot_env() -> Tuple[str, str, str]:
```
eavanvalkenburg, line 251: there already is a azure_aisearch_settings_from_dot_env method, why not extend that with a parameter to also get the index?

python/semantic_kernel/utils/settings.py
```diff
@@ -246,3 +246,33 @@ def azure_aisearch_settings_from_dot_env() -> Tuple[str, str]:
...
+
+def azure_aisearch_datasource_settings_from_dot_env_as_dict() -> Dict[str, str]:
+    """
+    Reads the Azure AI Search environment variables including index name from the .env file.
+
+    Returns:
+        Dict[str, str]: the Azure AI search environment variables
+    """
+    api_key, url, index_name = azure_aisearch_datasource_settings_from_dot_env()
+    return {"key": api_key, "endpoint": url, "indexName": index_name}
```
eavanvalkenburg, line 278: why not index_name?

python/samples/kernel-syntax-examples/azure_chat_gpt_with_data_api.py
```diff
@@ -0,0 +1,95 @@
...
+
+# For example, AI Search index may contain the following document:
+
+# Emily and David, two passionate scientists, met during a research expedition to Antarctica.
+# Bonded by their love for the natural world and shared curiosity, they uncovered a
+# groundbreaking phenomenon in glaciology that could potentially reshape our understanding of climate change.
+
+
+chat_service = sk_oai.AzureChatCompletion(
+    base_url=f"{str(endpoint).rstrip('/')}/openai/deployments/{deployment}/extensions",
```
eavanvalkenburg, line 31: maybe we should move this into a constructur with a flag for the extentions (`if use_extentions: base_url = ...`)

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion_base.py
```diff
@@ -85,6 +88,43 @@ async def complete_chat_with_functions_async(
                 _parse_message(choice.message, self.log) for choice in response.choices
             ]
 
+    async def complete_chat_with_data_async(
```
eavanvalkenburg, line 91: we need to move this new set of functions into the `azure_chat_completion` class since they are azure specific...

eavanvalkenburg (changes requested):
Getting there, some changes still needed and I hope to get the first version out of a major PR that changes the request settings setup which will make this easier...

python/semantic_kernel/utils/settings.py
sarah-widder: the extensions with data API expects camel casing (snake case support is in progress)

python/semantic_kernel/connectors/ai/chat_request_settings.py
```diff
@@ -19,7 +22,12 @@ class ChatRequestSettings:
     max_tokens: int = 256
     token_selection_biases: Dict[int, int] = field(default_factory=dict)
     stop_sequences: List[str] = field(default_factory=list)
-    function_call: Optional[str] = None
+    function_call: Optional[str] = (None,)
```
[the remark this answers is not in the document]
sarah-widder: whoops! fixed!!

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_handler.py
sarah-widder: refactored the settings to have the highest level contain dataSources, and AI search contains type and parameters. I think this should be nicely extensible but let me know if you have further thoughts on it.

python/semantic_kernel/orchestration/sk_function.py
sarah-widder: Switched the order so the check for with_data method is first, I think this should be safe for OpenAI clients but I don't have credentials to test one manually

python/semantic_kernel/orchestration/sk_function.py
sarah-widder: Moved the tool message processing before the loop!

python/semantic_kernel/utils/settings.py
sarah-widder: good point. added to that one, also added a method to directly deserialize to the AI search data source class that can be used in the azure chat with data settings.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion_base.py
sarah-widder: moved to `azure_chat_completion`

sarah-widder:
Hi @eavanvalkenburg if you can take another look I’d appreciate another review! Also I wonder if you could help with why the Python integration tests are being skipped, I want to make sure they do run in case there are any issues before merging. Thanks for your help!

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
```diff
@@ -124,6 +141,45 @@ def __init__(
...
+            the Azure portal. You can use either KEY1 or KEY2.
+        api_version: The API version to use. (Optional)
+            The default value is "2023-05-15".
+        ad_auth: Whether to use Azure Active Directory authentication. (Optional)
+            The default value is False.
+        default_headers: The default headers mapping of string keys to
+            string values for HTTP requests. (Optional)
+        log: The logger instance to use. (Optional)
+        use_extensions: Whether to use extensions, for example when chatting with data. (Optional)
+            The default value is False.
```
eavanvalkenburg, line 180: please add a note on what happens when this is set

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
```diff
@@ -171,6 +228,10 @@ def __init__(
         """
         if logger:
             logger.warning("The 'logger' argument is deprecated, use 'log' instead.")
+
+        if use_extensions:
```
eavanvalkenburg, line 232: either note in the docstring that when use_extensions is set, that base_url is ignored, or vice versa, when base_url is set, in theory someone could create an instance with a base_url with /extensions and use_extension = False, what happens then?

python/semantic_kernel/connectors/ai/open_ai/utils.py
```diff
@@ -189,7 +189,7 @@ async def chat_completion_with_function_call(
 
 
 def _parse_message(
-    message: ChatCompletion, logger: Optional[Logger] = None
+    message: ChatCompletion, logger: Optional[Logger] = None, with_data: bool = False
 ) -> Tuple[Optional[str], Optional[FunctionCall]]:
```
eavanvalkenburg, line 193: please update the output typing

python/semantic_kernel/utils/settings.py
```diff
@@ -231,7 +235,7 @@ def redis_settings_from_dot_env() -> str:
     return connection_string
 
 
-def azure_aisearch_settings_from_dot_env() -> Tuple[str, str]:
+def azure_aisearch_settings_from_dot_env() -> Tuple[str, str, str]:
```
eavanvalkenburg, line 238: I would prefer the index to be driven by a include_index_name variable, and then returning 2 or 3 items in the Tuple, so that this is not a breaking change.

eavanvalkenburg (changes requested):
some minor things, starting to look good!

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
sarah-widder: Noted in the docstring. If use_extensions is false, but base_url has /extensions, it will still call the extensions API so it's the same as use_extensions=True.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
sarah-widder: Updated docstring

python/semantic_kernel/utils/settings.py
sarah-widder: Good call. Updated this.

python/samples/kernel-syntax-examples/azure_chat_gpt_with_data_api.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Load Azure OpenAI Settings
+deployment, api_key, endpoint = sk.azure_openai_settings_from_dot_env()
+
+# Load Azure OpenAI with data settings
+azure_aisearch_datasource = sk.azure_aisearch_settings_from_dot_env_as_datasource()
+
+azure_chat_with_data_settings = (
+    sk.PromptTemplateWithDataConfig.AzureChatWithDataSettings(
+        dataSources=[azure_aisearch_datasource]
```
eavanvalkenburg, line 18: I would really like to see the field mapping in here (or a seperate example)

python/semantic_kernel/semantic_functions/prompt_template_with_data_config.py
```diff
@@ -0,0 +1,103 @@
...
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from semantic_kernel.semantic_functions.prompt_template_config import (
+    PromptTemplateConfig,
+)
+
+
+@dataclass
+class PromptTemplateWithDataConfig(PromptTemplateConfig):
```
eavanvalkenburg, line 12: Please move this class and file into the connectors/ai/open_ai folders since it is OpenAI specific.

eavanvalkenburg: and maybe it should even be a subclass of OpenAIChatPromptTemplate, since it is a chat feature and then Azure specific.

python/semantic_kernel/utils/settings.py
```diff
@@ -245,4 +251,43 @@ def azure_aisearch_settings_from_dot_env() -> Tuple[str, str]:
...
+    Returns:
+        Dict[str, str]: the Azure AI search environment variables
+    """
+    api_key, url, index_name = azure_aisearch_settings_from_dot_env(
+        include_index_name=True
+    )
+    return {"key": api_key, "endpoint": url, "indexName": index_name}
+
+
+def azure_aisearch_settings_from_dot_env_as_datasource() -> (
```
eavanvalkenburg, line 277: I would prefer if this class does not start including classes from the rest of SK.

eavanvalkenburg (changes requested):
definately getting there, some changes still

python/semantic_kernel/semantic_functions/prompt_template_with_data_config.py
sarah-widder: I don't think it can be a subclass of OpenAIChatPromptTemplate, since it's a PromptTemplateConfig and not a PromptTemplate

python/semantic_kernel/utils/settings.py
sarah-widder: Fair enough. Removed

python/semantic_kernel/semantic_functions/prompt_template_with_data_config.py
sarah-widder: Moved to the open_ai folders and renamed to OpenAIChatPromptTemplateWithDataConfig (nice and long name...). Still inherits from PromptTemplateConfig.

python/samples/kernel-syntax-examples/azure_chat_gpt_with_data_api.py
sarah-widder: Good idea - added field mapping to this example to demonstrate it

eavanvalkenburg (approved):
LGTM!

moonbox3 (approved):

moonbox3 (approved):

PR #3932, opened 2023-12-04
.Net: Quick fix for Azure DallE2 + 3

dluc:
### Motivation and Context

Some syntax errors, passing the API key as the model ID, or attempting to pass "dall-e3" model ID, leading to exceptions.  DallE3 uses a different HTTP flow compared to Dalle2, so the existing code could not be reused. Azure AI SDK for .NET doesn't support DallE3 either.

### Description

Refactor Azure Image generator to use Azure AI SDK for DallE2.
Add explicit check for DallE3 which is not supported yet by Azure AI SDK.

Future TODO:
* Refactor image generation to use Azure AI SDK also for OpenAI
* Update width/height checks, which are different for DallE3
* Support Dall-E 3 for OpenAI Client Calls.


dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,173 @@
...
+        ILoggerFactory? loggerFactory = null,
+        int? maxRetryCount = null,
+        string? apiVersion = null)
     {
         Verify.NotNullOrWhiteSpace(apiKey);
-        Verify.NotNull(modelId);
+
+        if (!string.IsNullOrEmpty(modelId))
+        {
+            throw new NotSupportedException($"Specifying a model ID, such as Dall-E-3, is not supported yet. Please provide an empty '{nameof(modelId)}' to use Dall-E-2.");
```
dmytrostruk, line 52: If DallE2 is the only model, which is supported by this class, I think we should remove `modelId` parameter from constructor, because with this implementation parameter `modelId` should be always set as empty, which would be nice to avoid. Since DallE3 uses a different HTTP flow compared to Dalle2, maybe this logic will be implemented in separate class, so `modelId` wouldn't be required in the future for this class.

Another option would be to set default value for `modelId` parameter as `dall-e-2` and in case of other Model ID - temporarily throw `NotSupportedException` exception until we add required implementation for DallE3. 

The main point here is to avoid accepting empty value in required constructor parameter as the only valid option.

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,173 @@
...
-        var operationId = await this.StartTextToImageAsync(description, width, height, cancellationToken).ConfigureAwait(false);
-        var result = await this.GetTextToImageResultAsync(operationId, cancellationToken).ConfigureAwait(false);
-
-        if (result.Result is null)
+        Verify.NotNull(description);
+        if (width != height)
         {
-            throw new KernelException("Azure OpenAI Text To Image null response");
+            throw new ArgumentOutOfRangeException("width,height", $"{width}x{height}",
+                "OpenAI can generate only square images");
```
dmytrostruk, line 84: A little improvement for exception message as below. 
```suggestion
                "OpenAI can generate only square images of size 256x256, 512x512, or 1024x1024");
```

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,173 @@
...
         }
-
-        var requestBody = JsonSerializer.Serialize(new TextToImageRequest
+        catch (RequestFailedException e) when (e.Status == 404)
         {
-            Prompt = description,
-            Size = $"{width}x{height}",
-            Count = 1
-        });
+            this._logger.LogError("Image generation failed with status code 404. This error can occur also when attempting to use Dall-E-3 which is still not supported");
```
dmytrostruk, line 116: Small improvement 😄 

```suggestion
            this._logger.LogError("Image generation failed with status code 404. This error can occur also when attempting to use Dall-E-3 which is not supported yet.");
```

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,173 @@
...
 
-        var uri = this.GetUri(GenerationImageOperation);
-        var result = await this._core.ExecutePostRequestAsync<AzureOpenAITextToImageResponse>(uri, requestBody, cancellationToken).ConfigureAwait(false);
+        if (!imageGenerations.HasValue)
+        {
+            throw new KernelException("The response does not contain an image result");
+        }
 
-        if (result == null || string.IsNullOrWhiteSpace(result.Id))
+        if (imageGenerations.Value.Data.Count == 0)
```
dmytrostruk, line 125: Can we add this condition to the previous one and have just:
```csharp
if (!imageGenerations.HasValue || imageGenerations.Value.Data.Count == 0)
{
    throw new KernelException("The response does not contain an image result");
}
```

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,173 @@
...
+                break;
+            case "2023-07-01-preview":
+                version = OpenAIClientOptions.ServiceVersion.V2023_07_01_Preview;
+                break;
+            case "2023-08-01-preview":
+                version = OpenAIClientOptions.ServiceVersion.V2023_08_01_Preview;
+                break;
+            case "2023-09-01-preview":
+                version = OpenAIClientOptions.ServiceVersion.V2023_09_01_Preview;
+                break;
```
dmytrostruk, line 155: VS highlights this case as not required, since it's the same as default option. Maybe we can remove it:
```suggestion
```

markwallace-microsoft (commented):
Just a few minor comments

dotnet/samples/KernelSyntaxExamples/KernelSyntaxExamples.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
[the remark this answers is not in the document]
dluc: yes, it copies appsettings.Development.json into the bin folder automatically if it exists, nothing else

dotnet/samples/KernelSyntaxExamples/Program.cs
```diff
@@ -75,6 +82,7 @@ private static async Task RunExamplesAsync(string? filter, CancellationToken can
     private static void LoadUserSecrets()
     {
         IConfigurationRoot configRoot = new ConfigurationBuilder()
+            .AddJsonFile("appsettings.Development.json")
```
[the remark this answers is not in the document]
dluc: it makes it so much easier to run the examples... is it a problem?

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
dluc: personally I prefer being explicit, given that new versions will appear in time, but I don't care too much tbh, seems very subjective, unless there are build warnings...

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,173 @@
...
+using Azure.AI.OpenAI;
+using Azure.Core.Pipeline;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.SemanticKernel.AI.TextToImage;
-using Microsoft.SemanticKernel.Services;
+using Microsoft.SemanticKernel.Http;
 
-namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.TextToImage;
+namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.ImageGeneration;
```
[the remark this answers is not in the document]
dluc: shouldn't this trigger a build warning?

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
dluc: actually I think the existing form with "yet" is better. The use of "still" could imply that it was expected to be supported by now but it isn't.

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
dluc: aren't they 2 different errors? one is about the result type, the other about the number of images? I'd prefer specific error messages to make debugging easier

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
dmytrostruk: In this case, as I mentioned, we can set `modelId` default value to `dall-e-2` and throw exception if passed ID is different. We don't know when support for DallE3 will be added, so there is a chance that empty parameter will be required to pass for a long period of time.

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
dmytrostruk: They are different, but from the user perspective, there is no difference between `The response does not contain an image result` exception message and `The response does not contain any image` exception message. I'm not sure if there is a case when user needs to handle these exceptions in different ways.

dotnet/samples/KernelSyntaxExamples/KernelSyntaxExamples.csproj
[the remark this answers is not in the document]
dluc: when using Sdk.Web files like `appsettings.Development.json` are automatically deployed to the bin folder, which is ideal for this project, and doesn't impact anything else, it's still a console app.

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,173 @@
...
+        int? maxRetryCount = null,
+        string? apiVersion = null)
     {
         Verify.NotNullOrWhiteSpace(apiKey);
-        Verify.NotNull(modelId);
+
+        if (!string.IsNullOrEmpty(modelId))
+        {
+            throw new NotSupportedException($"Specifying a model ID, such as Dall-E-3, is not supported yet. Please provide an empty '{nameof(modelId)}' to use Dall-E-2.");
+        }
```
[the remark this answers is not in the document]
dluc: I'm not familiar with the new telemetry framework, could you make that change?

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
dluc: "dall-e-2" exists only on OpenAI though, not on Azure where the model ID is undefined. Under Azure Studio Deployments, dall-e-2 doesn't show, it's more like a special endpoint. With dall-e-3 instead the models appears in the list.

Thoughts?

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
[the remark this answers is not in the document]
dmytrostruk: Which one exactly? If you mean `IDE0130 Namespace does not match folder structure` warning, as far as I know, we don't follow this rule anymore.

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
dluc: screenshot for DallE3

image

dmytrostruk: In the future when we add support for DallE3 in the same class, how users will differentiate between DallE2 and DallE3? If DallE3 will require `dall-e-3` model id, then intuitively user will try to set `dall-e-2` model id for DallE2. Even if `dall-e-2` is not part of Azure Studio Deployments, we still have to make it easy for SK users. 

Another way, as mentioned in previous comment, would be to implement DallE3 in scope of separate class, especially if this model has different HTTP flow than DallE2. In this case, for DallE2 implementation we can remove `modelId` parameter completely. We can also rename the class from `AzureOpenAITextToImageService` to `AzureOpenAIDalle2Service` or something like that, to clearly specify that this class works with special endpoint for DallE2. cc: @markwallace-microsoft 

But I still think that empty parameter is not clear identification that this is for DallE2.

dluc: change the code, now it let's everything through except for "dall-e-3"

dmytrostruk: Thank you!

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,161 @@
...
-            if (response.Headers.TryGetValues("retry-after", out var afterValues) && long.TryParse(afterValues.FirstOrDefault(), out var after))
-            {
-                await Task.Delay(TimeSpan.FromSeconds(after), cancellationToken).ConfigureAwait(false);
-            }
-
-            // increase retry count
-            retryCount++;
-        }
-    }
+            RetryPolicy = new RetryPolicy(maxRetries: Math.Max(0, maxRetryCount.Value)),
```
SergeyMenshykh, line 152: Thank you for refactoring the class. Since you have introduced the Azure AI SDK usage here, let's ensure that it's configured in the same way as in [the other places where it's used](https://github.com/microsoft/semantic-kernel/blob/e905aa9d3fc3e2f654f4f9a25729d063068b8004/dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientCore.cs#L479) to maintain consistent behavior across our AI services and connectors.

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,161 @@
...
+        }
 
-        this._endpoint = !string.IsNullOrEmpty(endpoint) ? endpoint! : httpClient!.BaseAddress!.AbsoluteUri;
-        this._apiKey = apiKey;
-        this._maxRetryCount = maxRetryCount.Value;
-        this._apiVersion = apiVersion;
-        this._core.AddAttribute(AIServiceExtensions.EndpointKey, endpoint);
-        this._core.AddAttribute(AIServiceExtensions.ModelIdKey, modelId);
-        this._core.AddAttribute(AIServiceExtensions.ApiVersionKey, apiVersion);
+        endpoint = !string.IsNullOrEmpty(endpoint) ? endpoint! : httpClient!.BaseAddress!.AbsoluteUri;
```
RogerBarreto, line 66:
```diff
-        if (httpClient?.BaseAddress == null && string.IsNullOrEmpty(endpoint))
-        {
-           throw new ArgumentException($"The {nameof(httpClient)}.{nameof(HttpClient.BaseAddress)} and {nameof(endpoint)} are both null or empty. Please ensure at least one is provided.");
-        }
-
-        endpoint = !string.IsNullOrEmpty(endpoint) ? endpoint! : httpClient!.BaseAddress!.AbsoluteUri;
+        endpoint = !string.IsNullOrEmpty(endpoint) ? endpoint! : httpClient?.BaseAddress?.AbsoluteUri;
+        
+        if (endpoint is null)
+        {
+            throw new ArgumentException($"The {nameof(httpClient)}.{nameof(HttpClient.BaseAddress)} and {nameof(endpoint)} are both null or empty. Please ensure at least one is provided.");
+        }
```

dotnet/src/Connectors/Connectors.AI.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,209 +3,161 @@
...
+                break;
+            case 512:
+                size = ImageSize.Size512x512;
+                break;
+            case 1024:
+                size = ImageSize.Size1024x1024;
+                break;
+            default:
+                throw new ArgumentOutOfRangeException(nameof(width), SizeErr);
         }
```
RogerBarreto, line 106: Suggestoin for simplification:

```diff
- var size = ImageSize.Size256x256;
- switch (width)
- {
-     case 256:
-         size = ImageSize.Size256x256;
-         break;
-     case 512:
-         size = ImageSize.Size512x512;
-         break;
-     case 1024:
-         size = ImageSize.Size1024x1024;
-         break;
-     default:
-         throw new ArgumentOutOfRangeException(nameof(width), SizeErr);
- }
+ var size = width switch
+ {
+     256 => ImageSize.Size256x256,
+     512 => ImageSize.Size512x512,
+     1024 => ImageSize.Size1024x1024,
+     _ => throw new ArgumentOutOfRangeException(nameof(width), SizeErr)
+ };

File: dotnet/src/Connectors/Connectors.UnitTests/OpenAI/TextToImage/AzureOpenAITextToImageTests.cs [hunk 13]
```diff
@@ -13,17 +14,60 @@
...
+        const string ValidModel3 = "";
+
+        // Act
+        Assert.Throws<NotSupportedException>(() => new AzureOpenAITextToImageService(modelId: InvalidModel, endpoint: "https://az.com", apiKey: "abc"));
+
+        // No exceptions for these
+        var _ = new AzureOpenAITextToImageService(modelId: ValidModel1, endpoint: "https://az.com", apiKey: "abc");
+        _ = new AzureOpenAITextToImageService(modelId: ValidModel2, endpoint: "https://az.com", apiKey: "abc");
+        _ = new AzureOpenAITextToImageService(modelId: ValidModel3, endpoint: "https://az.com", apiKey: "abc");
+    }
```
RogerBarreto, line 38: This is a good example for `[Theory]`
```suggestion
    [Theory]
    [InlineData("dall-e-3", false)]
    [InlineData("dall-e-2", true)]
    [InlineData("Dall-E-2", true)]
    [InlineData("", true)]
    public void ItValidatesTheModelId(modelId, shouldBeValid)
    {
        // Act
        if (!shouldBeValid) 
        {
            Assert.Throws<NotSupportedException>(() => new AzureOpenAITextToImageService(modelId: modelId, endpoint: "https://az.com", apiKey: "abc"));
        }
        else 
        {
            // No exceptions for these
            var _ = new AzureOpenAITextToImageService(modelId: modelId, endpoint: "https://az.com", apiKey: "abc");
        }
    }
```

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,215 +3,151 @@
...
+
+        this._deploymentName = deploymentName;
 
-        if (httpClient?.BaseAddress == null && string.IsNullOrEmpty(endpoint))
+        this._logger = loggerFactory?.CreateLogger(typeof(AzureOpenAITextToImageService)) ?? NullLogger.Instance;
+        if (string.Equals(modelId, "dall-e-2", StringComparison.OrdinalIgnoreCase))
         {
-            throw new ArgumentException($"The {nameof(httpClient)}.{nameof(HttpClient.BaseAddress)} and {nameof(endpoint)} are both null or empty. Please ensure at least one is provided.");
+            this._logger.LogWarning($"{nameof(AzureOpenAITextToImageService)} supports only Dall-E-3. .");
+            throw new NotSupportedException("Dall-E-2 support was deprecated in Azure Open AI latest SDK. Please use a 'dall-e-3' deployment.");
```
dmytrostruk, line 57: I still think that this condition is not required because `modelId` property is not used in this class. Whichever `modelId` is passed, it doesn't impact the functionality, which means that we should not throw an exception and probably `modelId` property can be removed from constructor. In case if developer uses Azure deployment, that is not supported, Azure SDK will throw appropriate exception and we will pass it to the user.

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,215 +3,151 @@
...
+        string description,
+        int width,
+        int height,
+        Kernel? kernel = null,
+        CancellationToken cancellationToken = default)
     {
-        var operationId = await this.StartTextToImageAsync(description, width, height, cancellationToken).ConfigureAwait(false);
-        var result = await this.GetTextToImageResultAsync(operationId, cancellationToken).ConfigureAwait(false);
+        Verify.NotNull(description);
+        const string SizeErr = "Dall-E 2 can generate only square images of size 256x256, 512x512, or 1024x1024";
```
dmytrostruk, line 83: Since this class supports only Dall-E 3, this should be updated:
```suggestion
        const string SizeErr = "Dall-E 3 can generate only images of size 1024x1024, 1792x1024, or 1024x1792";
```

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,215 +3,151 @@
...
 
-        if (result.Result is null)
+        if (width != height)
```
dmytrostruk, line 85: This condition is not valid as well, since Dall-E 3 supports non-square images as well.

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,215 +3,151 @@
...
+            throw new ArgumentOutOfRangeException("width,height", $"{width}x{height}", SizeErr);
         }
 
-        if (result.Result.Images.Count == 0)
+        var size = width switch
         {
-            throw new KernelException("Azure OpenAI Text To Image result not found");
+            256 => ImageSize.Size256x256,
+            512 => ImageSize.Size512x512,
+            1024 => ImageSize.Size1024x1024,
```
dmytrostruk, line 94: This should be updated to handle `ImageSize.Size1024x1024`, `ImageSize.Size1792x1024`, `ImageSize.Size1024x1792`.

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,215 +3,151 @@
...
-    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <returns> The operationId that identifies the original text to image request. </returns>
-    private async Task<string> StartTextToImageAsync(string description, int width, int height, CancellationToken cancellationToken = default)
-    {
-        Verify.NotNull(description);
-        if (width != height || width != 256 && width != 512 && width != 1024)
+        catch (RequestFailedException e) when (e.Status == 404)
         {
-            throw new ArgumentOutOfRangeException(nameof(width), width, "OpenAI can generate only square images of size 256x256, 512x512, or 1024x1024.");
+            this._logger.LogError("Image generation failed with status code 404. This error can occur also when attempting to use Dall-E-3 which is still not supported");
```
dmytrostruk, line 111: This error message may not be applicable with new implementation.

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,215 +3,151 @@
...
-    private string GetUri(string operation, params string[] parameters)
-    {
-        var uri = new Azure.Core.RequestUriBuilder();
-        uri.Reset(new Uri(this._endpoint));
-        uri.AppendPath(operation, false);
-        foreach (var parameter in parameters)
+        if (httpClient != null)
         {
-            uri.AppendPath("/" + parameter, false);
+            options.Transport = new HttpClientTransport(httpClient);
```
SergeyMenshykh, line 148: We need to disable default Azure SDK retry policy to have only one source of truth - SK to avoid double retries. 

`options.RetryPolicy = new RetryPolicy(maxRetries: 0); // Disable Azure SDK retry policy if and only if a custom HttpClient is provided.`

dotnet/src/Connectors/Connectors.UnitTests/OpenAI/TextToImage/AzureOpenAITextToImageTests.cs
```diff
@@ -13,17 +14,60 @@
...
+        }
+        else
+        {
+            // No exceptions for these
+            new AzureOpenAITextToImageService(deploymentName: deploymentName, modelId: modelId, endpoint: "https://az.com", apiKey: "abc");
+        }
+    }
+
+    [Fact(Skip = "Needs refactoring, decouple from Azure SDK implementation details")]
+    public async Task ItGeneratesImagesAsync()
```
dmytrostruk, line 41: Should we remove the test in this case?

RogerBarreto: Removed this and added other for the NotSupportedException scenarios.

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
RogerBarreto: Thanks, added the suggestion

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
RogerBarreto: Removed, the current ClientCore don't log at this stage, removing to keep it similar.

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
RogerBarreto: You are right, updated on the conditional code in the client to expect only viable Dall-E-3 scenarios

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
RogerBarreto: Removed

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
```diff
@@ -1120,12 +1123,11 @@ public static class OpenAIServiceCollectionExtensions
```
markwallace-microsoft, line 1123: Given we pass in a modelId we  should pass it to the service and have set it in the service attributes and it will be used by the service selector

RogerBarreto: @markwallace-microsoft I followed Dmytro suggestion of removing the ModelId, not sure if this comment applies anymore.

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs
```diff
@@ -3,215 +3,136 @@
...
     /// <param name="httpClient">Custom <see cref="HttpClient"/> for HTTP requests.</param>
     /// <param name="loggerFactory">The ILoggerFactory used to create a logger for logging. If null, no logging will be performed.</param>
-    /// <param name="maxRetryCount"> Maximum number of attempts to retrieve the text to image operation result.</param>
     /// <param name="apiVersion">Azure OpenAI Endpoint ApiVersion</param>
     public AzureOpenAITextToImageService(
-        string? endpoint,
+        string deploymentName,
+        string endpoint,
         string apiKey,
-        string? modelId = null,
```
markwallace-microsoft, line 70: We should keep the `modelId` for telemetry purposes in the short term.

Longer term if Azure OpenAI supports both Dalle 3&3 then we will want to allow developers to specify which model to use and use the service selector to select the correct service.

RogerBarreto: Brought it back

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
RogerBarreto: Brought back the ModeId

markwallace-microsoft (approved):

markwallace-microsoft (approved):

PR #3964, opened 2023-12-05
.Net: Remove CreateFunctionFromPromptXXXResource methods

### LINKED ISSUE #3951: CreateFunctionFromPromptYamlResource doesn't get the resource from the right assembly
If you try to use CreateFunctionFromPromptYamlResource, it throws an error because it looks for the resource in the Functions.Yaml assembly (not the developers current project)

markwallace-microsoft:
### Motivation and Context

Close #3951 

### Contribution Checklist

### DIFF
--- a/dotnet/src/Functions/Functions.Yaml/KernelFunctionYaml.cs
+++ b/dotnet/src/Functions/Functions.Yaml/KernelFunctionYaml.cs
@@ -1,7 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
-using System.IO;
-using System.Reflection;
 using Microsoft.Extensions.Logging;
 using YamlDotNet.Serialization;
 using YamlDotNet.Serialization.NamingConventions;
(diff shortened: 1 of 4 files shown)


RogerBarreto (approved):

SergeyMenshykh (approved):

dotnet/src/Functions/Functions.Yaml/KernelFunctionYaml.cs
```diff
@@ -13,26 +11,6 @@ namespace Microsoft.SemanticKernel;
...
-    /// <param name="resourceName">Resource containing the YAML representation of the <see cref="PromptTemplateConfig"/> to use to create the prompt function</param>
-    /// <param name="promptTemplateFactory">>Prompt template factory.</param>
-    /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging. If null, no logging will be performed.</param>
-    /// <returns>The created <see cref="KernelFunction"/>.</returns>
-    public static KernelFunction FromPromptYamlResource(
-        string resourceName,
-        IPromptTemplateFactory? promptTemplateFactory = null,
-        ILoggerFactory? loggerFactory = null)
-    {
-        using StreamReader reader = new(Assembly.GetExecutingAssembly().GetManifestResourceStream(resourceName));
```
stephentoub, line 28: How commonly needed were these CreateFromXxResource methods?  If it's a really common scenario, we could add back CreateFromXx methods that take a Stream instead of a string, and then it's just the caller's responsibility to produce that Stream (which they can get from their own assembly's resources).

PR #3974, opened 2023-12-05
.Net Adding Kernel + Function.InvokeAsync<T>

### LINKED ISSUE #3863: .Net: Should the FunctionResult class be made generic instead of using GetResult<T>
If we changed `FunctionResult` to `FunctionResult<T>`, in the future, we could enable scenarios described in [option 3 of Roger's ADR](https://github.com/RogerBarreto/semantic-kernel/blob/baadfba005d92ccc76ef61a206c33a2860b81dbd/docs/decisions/0024-multi-modal-service-support.md#option-3---gives-back-what-you-asked). It would also drive further parallelism with `IAsyncEnumerable<T>` in the streaming code paths.

The new class would look something like this...

```csharp
public sealed class FunctionResult<T> : FunctionResult
{
    internal T? Value { get; private set; } = null;
    public T? GetValue()
    {
        if (this.Value is null)
        {
            return default;
        }

        return typedResult;
    }
    public FunctionResult(string functionName, T? value, CultureInfo culture)
        : this(functionName)
    {
        this.Value = value;
        this.Culture = culture;
    }
    {
        return ConvertToString(this.Value, this.Culture) ?? string.Empty;
    }
}
```

As a consequence (benefit?), the `InvokeAsync()` methods would need to be updated to `InvokeAsync<T>()` to match `InvokeStreamingAsync<T>()`.

In the future, the `InvokeAsync<T>()` method could use the generic to determine which type (if multiple are supported) to return back to the user.

RogerBarreto:
### Motivation and Context

Proposal to resolve the below issue.

Resolves #3863


dotnet/src/SemanticKernel.Core/Functions/KernelFunctionExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// Invokes the<see cref="KernelFunction"/>.
+    /// </summary>
+    /// <param name="kernelFunction">The <see cref="KernelFunction"/> to invoke.</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="arguments">The arguments to pass to the function's invocation, including any <see cref="PromptExecutionSettings"/>.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The provided generic typed result value of the function's execution.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="kernel"/> is null.</exception>
+    /// <exception cref="KernelFunctionCanceledException">The <see cref="KernelFunction"/>'s invocation was canceled.</exception>
+    public static async Task<T?> InvokeAsync<T>(
```
stephentoub, line 25: In all of these, I suggest renaming T to be TResult

dotnet/samples/KernelSyntaxExamples/Example03_Variables.cs
```diff
@@ -25,8 +25,20 @@ public static async Task RunAsync()
...
+        // ** Different ways of executing function with arguments **
 
-        Console.WriteLine(result.GetValue<string>());
+        // Specify and get the value type as generic parameter
+        var resultValue = await kernel.InvokeAsync<string>(textPlugin["AppendDay"], arguments);
+        Console.WriteLine($"string -> {resultValue}");
+
+        // If you need to access the result metadata, you can use the non-generic version to get the FunctionResult
+        var functionResult = await kernel.InvokeAsync(textPlugin["AppendDay"], arguments);
+        var metadata = functionResult.Metadata;
```
markwallace-microsoft, line 36: Do we need this?

markwallace-microsoft (approved):

dotnet/samples/KernelSyntaxExamples/Example03_Variables.cs
RogerBarreto: Added just to justify the `functionResult` usage according to the comment above.

I added all different ways you can retrieve the result, from common to most unlikely. We can remove if this is too much.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #3979, opened 2023-12-05
.Net Reorder Example Numbers

### LINKED ISSUE #3978: .Net: [Sample organization] Kernel Examples Have Numbering Gaps
# Problem

Current examples numbers have gaps, 

# Solution 

Reorder the existing ones for V1.

RogerBarreto:
### Motivation and Context

Resolves #3978


dotnet/samples/KernelSyntaxExamples/Example03_InlineFunctionDefinition.cs
```diff
@@ -7,7 +7,7 @@
 using RepoUtils;
 
 // ReSharper disable once InconsistentNaming
-public static class Example05_InlineFunctionDefinition
+public static class Example03_InlineFunctionDefinition
```
stephentoub, line 10: At a minimum the files should be renamed as well.

RogerBarreto: I'm going to put a hold on this PR until we are very close to v1 and have removed any invalid examples that may be still lurking around

markwallace-microsoft (commented):
@RogerBarreto I'm working on a PR to drop the numbering and instead categorise all of the samples. This is the plan agreed with Matthew to try make the samples easier to navigate. So I think we can close this PR.

RogerBarreto:
Dropping this as we will restructure the Examples.

PR #3984, opened 2023-12-05
.Net Adding Breaking Glass support for Completion/Generation Services

### LINKED ISSUE #3182: .Net: Need to be able to "break glass" with AzureChatCompletion and friends
Abstractions are great, but they're invariably the lowest common denominator of the things they're abstracting / wrapping, and at some point you often need to "break glass" and get at functionality from the thing being wrapped (especially if new features are added to the underlying thing and not yet exposed through the wrapper).  Today, AzureChatCompletion and OpenAIChatCompletion don't make that possible as they don't expose the underlying OpenAIClient and model ID they're wrapping.  That means in a situation where you have an IChatCompletion and want to test whether it's an AzureChatCompletion and do something special in that case, you have no good way to get at that underlying information, and instead need to also pass around the OpenAIClient.

We should expose both the OpenAIClient and model ID used to construct the instance publicly on these types. The properties are already there (`Client` and `ModelId`), they're just private; we can make them public.  Then code that really needs them can do:
```C#
IChatCompletion completion = ...;
if (completion is AzureChatCompletion azureCompletion)
{
    Use(azureCompletion.Client, azureCompletion.ModelId);
}
```
Etc.

(There's an argument to be made that exposing these ties the implementation to using OpenAIClient, but it's already tied in that manner because it accepts it into the constructor. If we were really concerned, such that maybe in the future the completion object would only use an OpenAIClient if one were provided but otherwise would use a different mechanism, we could make the Client property nullable.)

RogerBarreto:
### Motivation and Context

Resolves #3182


dotnet/src/Connectors/Connectors.AI.HuggingFace/TextGeneration/HuggingFaceTextGenerationService.cs
```diff
@@ -29,6 +29,21 @@ public sealed class HuggingFaceTextGenerationService : ITextGenerationService
...
+    /// </summary>
+    public string? ModelId => this.GetModelId();
+
+    /// <summary>
+    /// Http Client used for the service call.
+    /// </summary>
+    /// <remarks>
+    /// The usage of this property is considered "unsafe". Use it only if strictly necessary.
+    /// </remarks>
+    public HttpClient Client => this._httpClient;
```
stephentoub, line 44: Does this actually provide a benefit? It doesn't always have the endpoint built in, it doesn't add any headers that might be necessary. What can you do with this that you can't do with another HttpClient instance?

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureOpenAIChatCompletionService.cs
```diff
@@ -19,6 +19,21 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.ChatCompletion;
...
 public sealed class AzureOpenAIChatCompletionService : IChatCompletionService, ITextGenerationService
 {
+    #region BreakingGlass Properties
+    /// <summary>
+    /// Model Id for OpenAI or Deployment Name when using Azure OpenAI.
+    /// </summary>
+    public string DeploymentOrModelName => this._core.DeploymentOrModelName;
+
+    /// <summary>
+    /// OpenAI / Azure OpenAI Client
```
stephentoub, line 29: At this point it'd be good to be writing the XML comments that we expect developers to actually see, e.g.
```suggestion
    /// Gets the underlying <see cref="OpenAIClient"/> used by this <see cref="AzureOpenAIChatCompletionService"/> instance.
```

stephentoub (commented):
Thanks.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureOpenAIChatCompletionService.cs
```diff
@@ -19,6 +19,21 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.ChatCompletion;
...
+    /// </summary>
+    /// <remarks>
+    /// The usage of this property is considered "unsafe". Use it only if strictly necessary.
```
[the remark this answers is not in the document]
RogerBarreto: We have no control on the implimentation of this client. Is a word that can change, I'm using the same remark description for all "breaking glasses" impl.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureOpenAIChatCompletionService.cs
RogerBarreto: Resolved, applied commit.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureOpenAIChatCompletionService.cs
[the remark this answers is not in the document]
markwallace-microsoft: How about:

`The Semantic Kernel will be kept up-to-date with the latest Azure.AI.OpenAI package. As this package changes you may see changes to the <see cref="OpenAIClient"/>. Use this only if strictly necessary.`

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextGeneration/HuggingFaceTextGenerationService.cs
```diff
@@ -29,6 +29,21 @@ public sealed class HuggingFaceTextGenerationService : ITextGenerationService
     private readonly string? _apiKey;
     private readonly Dictionary<string, object?> _attributes = new();
 
+    #region BreakingGlass Properties
+    /// <summary>
+    /// Model Id.
+    /// </summary>
+    public string? ModelId => this.GetModelId();
```
markwallace-microsoft, line 36: This isn't breaking glass, this get's set by the client.

The value are available via `IAIService.Attributes` already so no need to add these here.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextGeneration/HuggingFaceTextGenerationService.cs
markwallace-microsoft: Let's remove this to keep the public API surface minimal. We can add it back if needed.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureOpenAIChatCompletionService.cs
```diff
@@ -19,6 +19,21 @@ namespace Microsoft.SemanticKernel.Connectors.AI.OpenAI.ChatCompletion;
 /// </summary>
 public sealed class AzureOpenAIChatCompletionService : IChatCompletionService, ITextGenerationService
 {
+    #region BreakingGlass Properties
+    /// <summary>
+    /// Model Id for OpenAI or Deployment Name when using Azure OpenAI.
+    /// </summary>
+    public string DeploymentOrModelName => this._core.DeploymentOrModelName;
```
markwallace-microsoft, line 26: This is not breaking glass as it's set by the developer.

These values are available via `IAIService.Attributes` already so no need to add these here.

Also we should not be exposing `DeploymentorModelName` as a property. We have separate properties for DeploymentName and ModelId.

markwallace-microsoft (changes requested):
I don't think we need these changes. We can get access to model id, deploy name etc from IAIService.Attributes and we can inject the various clients.

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletion/AzureOpenAIChatCompletionService.cs
RogerBarreto: Removed.

dotnet/src/Connectors/Connectors.AI.HuggingFace/TextGeneration/HuggingFaceTextGenerationService.cs
RogerBarreto: Removed.

RogerBarreto:
Dropping this for V1.

PR #4000, opened 2023-12-05
.Net: Add default JSON-serializable TypeConverter

dehoward:
### Motivation and Context

Allows for custom types to be JSON-serializable by default by having a `TypeConverter` returned if a custom one is not already defined for the given type.

### Description
- Changes `TypeConverterFactory` to return a `JsonSerializationTypeConverter` 
- add tests for the factory
- updates example 60 to mention the converter is optional unless some sort of custom serialization is needed

### Contribution Checklist


dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
```diff
@@ -51,6 +53,21 @@ internal static class TypeConverterFactory
...
+        public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType) => true;
+
+        public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+        {
+            return JsonSerializer.Deserialize<object>((string)value);
+        }
+
+        public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+        {
+            return JsonSerializer.Serialize(value);
```
stephentoub, line 70: Both ConvertFrom and ConvertTo here are making assumptions about the source and destination types. This is going to require some re-work as part of @SergeyMenshykh's work to support marshaling with types other than string.

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
```diff
@@ -51,6 +53,21 @@ internal static class TypeConverterFactory
...
+        return new JsonSerializationTypeConverter();
+    }
+
+    internal sealed class JsonSerializationTypeConverter : TypeConverter
+    {
+        public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType) => true;
+
+        public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+        {
+            return JsonSerializer.Deserialize<object>((string)value);
```
stephentoub, line 65: Does specifying `<object>` here work? How does the serializer know the actual destination type to use?

dehoward: no it doesn't. updated to pass the correct type as a 2nd argument to the `JsonSerializer`

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
dehoward: I've setup a sync with him for the morning to get on the same page about this.

dotnet/samples/KernelSyntaxExamples/Example60_AdvancedMethodFunctions.cs
```diff
@@ -75,8 +75,9 @@ public static MyCustomType Function2()
     #region Custom Type
 
     /// <summary>
-    /// In order to use custom types, <see cref="TypeConverter"/> should be specified,
-    /// that will convert object instance to string representation.
+    /// In order to use custom types, a <see cref="TypeConverter"/> can be specified,
+    /// but is not required if the native serialiation provided by default is sufficient.
```
stephentoub, line 79:
```suggestion
    /// but is not required if the native serialization provided by default is sufficient.
```

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
```diff
@@ -51,6 +53,28 @@ internal static class TypeConverterFactory
...
+
+        public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+        {
+            return JsonSerializer.Deserialize((string)value, this.type);
+        }
+
+        public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+        {
+            return JsonSerializer.Serialize(value, this.type);
+        }
```
stephentoub, line 78: I think you said you're going to sync with @SergeyMenshykh; I expect this is going to need to be changed based on his efforts around arbitrary object marshaling. As written, this expects everything is strings.

SergeyMenshykh:
@dehoward you will need the PR in https://github.com/microsoft/semantic-kernel/pull/4044 to have the JSON converter working.

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
```diff
@@ -16,7 +18,7 @@ internal static class TypeConverterFactory
     /// </summary>
     /// <param name="type">The Type of the object to convert.</param>
     /// <returns>A TypeConverter instance if a suitable converter is found, otherwise null.</returns>
-    internal static TypeConverter? GetTypeConverter(Type type)
+    internal static TypeConverter GetTypeConverter(Type type)
```
SergeyMenshykh, line 21: Please add unit tests to the `KernelFunctionFromMethodTests1` class for testing object-to-JSON and JSON-to-object marshaling.

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
```diff
@@ -51,6 +53,33 @@ internal static class TypeConverterFactory
             return converter;
         }
 
-        return null;
+        return new JsonSerializationTypeConverter(type);
+    }
+
+    internal sealed class JsonSerializationTypeConverter : TypeConverter
+    {
+        private Type type { get; }
```
SergeyMenshykh, line 61: There is no need to declare a property if a field can be used instead.

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
dehoward: resolving in favor of the other comment

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
dehoward: had a sync with him and after some discussion we determined that this was fine. specifically because we should only ever be going to/from a string when marshalling a complex type.

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
dehoward: I've added them, can you take a look?

dotnet/src/SemanticKernel.UnitTests/Functions/KernelFunctionFromMethodTests1.cs
```diff
@@ -1104,6 +1104,94 @@ public async Task ItPropagatesOriginalExceptionTypeAsync()
...
+        {
+            await function.InvokeAsync(this._kernel, arguments: new() { { "complexType", new ComplexType() { Foo = 42 } } });
+            Assert.Fail("Should have thrown an ArgumentException since complex type cannot be marshalled to int");
+        }
+        catch (ArgumentException)
+        {
+            count++;
+        }
+
+        Assert.Equal(1, count);
```
SergeyMenshykh, line 1166: All these lines can be replaced by this one - await Assert.ThrowsAsync<ArgumentException>(() => fawait function.InvokeAsync(this._kernel, arguments: new() { { "complexType", new ComplexType() { Foo = 42 } } }));

dotnet/src/SemanticKernel.UnitTests/Functions/KernelFunctionFromMethodTests1.cs
```diff
@@ -1104,6 +1104,94 @@ public async Task ItPropagatesOriginalExceptionTypeAsync()
...
+        {
+            result.GetValue<double>();
+            Assert.Fail("Should have thrown an InvalidCastException since complex type cannot be cast to double");
+        }
+        catch (InvalidCastException)
+        {
+            count++;
+        }
+
+        Assert.Equal(1, count);
```
SergeyMenshykh, line 1192: Same as comment above.

dotnet/src/InternalUtilities/src/System/TypeConverterFactory.cs
```diff
@@ -49,6 +51,41 @@ internal static class TypeConverterFactory
...
+        public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType) => sourceType == typeof(string);
+
+        public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+        {
+            if (value is not string)
+            {
+                throw new ArgumentOutOfRangeException(nameof(value), value, "Cannot convert from non-string value.");
+            }
+
+            return JsonSerializer.Deserialize(value.ToString()!, this._type);
```
SergeyMenshykh, line 78: no need to call `ToString` method here. Just add variable name it the if block above - `if (value is not string json)`

SergeyMenshykh (approved):

alliscode (approved):

markwallace-microsoft (approved):

PR #4007, opened 2023-12-05
.Net With the new dedicated Rendering events rederedPrompt metadata is not necessary.

### LINKED ISSUE #4005: Drop KernelEventArgsExtensions and RenderedPrompt from Kernel Events
Since the addition of the new `PromptRendering` and `PromptRendered` events, it is not necessary anymore to have a metadata with the rendered prompt available to the Kernel `FunctionInvoking` `FunctionInvoked` events.

Removing those implementations for simplicity.

RogerBarreto:
Resolves #4005

Also fixes a potential bug where metadata references from multiple choices and chunks shares the seame reference.


dmytrostruk (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -130,10 +130,7 @@ internal sealed class KernelFunctionFromPrompt : KernelFunction
 
         var textContent = await textGeneration.GetTextContentAsync(renderedPrompt, arguments.ExecutionSettings, kernel, cancellationToken).ConfigureAwait(false);
 
-        IDictionary<string, object?> metadata = textContent.Metadata ?? new Dictionary<string, object?>();
-        metadata.Add(KernelEventArgsExtensions.RenderedPromptMetadataKey, renderedPrompt);
-
-        return new FunctionResult(this, textContent.Text, kernel.Culture, new Dictionary<string, object?>(metadata));
+        return new FunctionResult(this, textContent.Text, kernel.Culture, new Dictionary<string, object?>(textContent.Metadata ?? new Dictionary<string, object?>()));
```
stephentoub, line 133:
```suggestion
        return new FunctionResult(this, textContent.Text, kernel.Culture, textContent.Metadata);
```

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -120,7 +120,7 @@ internal ClientCore(ILogger? logger = null)
 
         this.CaptureUsageDetails(responseData.Usage);
         var metadata = GetResponseMetadata(responseData);
-        return responseData.Choices.Select(choice => new TextContent(choice.Text, this.DeploymentOrModelName, choice, Encoding.UTF8, metadata)).ToList();
+        return responseData.Choices.Select(choice => new TextContent(choice.Text, this.DeploymentOrModelName, choice, Encoding.UTF8, new Dictionary<string, object?>(metadata))).ToList();
```
stephentoub, line 123: Why is it we need a new dictionary and can't just pass through the original?

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -367,7 +367,7 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
                     }
                 }
 
-                yield return new OpenAIStreamingChatMessageContent(update, update.ChoiceIndex ?? 0, this.DeploymentOrModelName, metadata);
+                yield return new OpenAIStreamingChatMessageContent(update, update.ChoiceIndex ?? 0, this.DeploymentOrModelName, new(metadata));
```
stephentoub, line 370: Why does this dictionary need to be wrapped?

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
RogerBarreto: Actually this was good, I just looked and fix a potential reference leak between different choices and chunks. 

Knowing that each choice and chunk now will have its own dictionary, we can pass the reference to the FunctionResult.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientCore.cs
RogerBarreto: The metadata when is changed in an FunctionResult or a ContentBase or a StreamingContentBase, if thats not done will have the same reference, and changes to it will reflect on all chunks and choices also.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientCore.cs
RogerBarreto: Answered above...

stephentoub: Where else is this dictionary instance used other than this one return?

RogerBarreto: This is a yield return, so, changing the dictionary of this chunk will eventually change the same for all other chunks.

If you have a logic like, add key in the `foreach` this would break with key already exists in the following chunk.

stephentoub: How about just having:
```C#
metadata ??= new ReadOnlyDictionary<string, object?>(GetResponseMetadata(update));
```
?

RogerBarreto: @stephentoub Currently StreamingContentBase Metadata is a `IDictionary<string, object?>`.

This is read-write to allow changes to the values as contents can be created, serialized and deserialized. 

This is analogous to the `ContentBase` -> `ChatMessageContent` to be able to serialize and deserialize from a ChatHistory.

dotnet/src/Connectors/Connectors.AI.OpenAI/AzureSdk/ClientCore.cs
stephentoub: How common will it be to have multiple Choices?  That's pretty rare, right? Might be nice to avoid the unnecessary copy if Choices.Count == 1.  This could be something like:
```C#
var results = new TextContent[responseData.Choices.Count];
for (int i = 0; i < results.Length; i++)
{
    results[i] = new TextContent(choice.Text, this.DeploymentOrModelName, choice, Encoding.UTF8,
        i == 0 ? metadata : new Dictionary<string, object?>(metadata));
}
return results;
```

SergeyMenshykh (approved):

Issue #3692, opened 2023-11-27
.Net: Change KernelFunctionsArgument from string values to object values. Fix marshaling accordingly.


markwallace-microsoft:
```[tasklist]
### Tasks
- [x] 1 Prepare public API surface(Change KernelArguments value type to object? + add string value checks)
- [x] 2 Support object? type by OpenAPI(RestAPIOperationRunner)
- [x] 3 Support object? type by KernelPromptTemplate
```

PR #4013, opened 2023-12-05
.Net: Support for primitives and complex types by Kernel prompt template

SergeyMenshykh:
### Motivation and Context

Closes a corresponding task in the - https://github.com/microsoft/semantic-kernel/issues/3692 issue.

This update is the next step in making it possible to use primitive and complex data types in SK. Currently, the kernel prompt template can only work with strings. This PR makes it possible to use other basic types like int, decimal, bool, object, etc... and complex types as well.

For an easy review experience, this PR does not rename the `ITextRendering` interface and introduces duplication for object-to-string conversion functionality. These changes will be addressed in follow-up PRs shortly afterward.


dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
```diff
@@ -123,42 +122,42 @@ internal async Task<string> RenderAsync(List<Block> blocks, Kernel kernel, Kerne
...
 
-    /// <summary>
-    /// Validates that all the KernelArguments are of string type.
-    /// </summary>
-    /// <param name="arguments">The collection of KernelArguments to validate.</param>
-    /// <exception cref="KernelException">Thrown when an argument is not of type string.</exception>
-    private static void AssertArgumentOfStringType(KernelArguments? arguments)
-    {
-        if (arguments == null)
+    private static Func<object?, CultureInfo, string?>? GetTypeConverterDelegate(Type sourceType) =>
```
stephentoub, line 138: This is the duplicated logic you talk about cleaning up subsequently?

stephentoub (approved):

dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
SergeyMenshykh: Right, both `ConvertToString` and `GetTypeConverterDelegate` method will be moved to a separate class and reused by the `KErnelPromptTemplate` and `FunctionResult` classes.

Issue #3947, opened 2023-12-04
.Net: Isolate Handlebars Kernel System helpers in their own class + add Categories filter for KernelSystemHelpers (i.e., KernelHelperCategories) 


teresaqhoang:
When registering helpers we should follow the same pattern as is used here https://github.com/Handlebars-Net/Handlebars.Net.Helpers. 

We don't want to put everything into a single class as indicated by using private static helper methods.

When registering these helpers with the template engine, there should be calls that look something like this:
```
KernelPromptHelpers.Register(handlebarsContext);
KernelPluginHelpers.Register(handlebarsContext);
KernelContextHelpers..Register(handlebarsContext);
```

Can probably also use `RegisterCustomHelpers` from `Handlebars.Net.Helpers` library for their built-in argument handling

Other notes
+ add option to filter KernelSystemHelpers (i.e., KernelHelperCategories) 
+ Should also consider standardizing the naming conventions for these helpers, i.e., 'Message' instead of 'message' since the built in Handlebars library helpers use PascalCase

markwallace-microsoft:
Closing as obsolete

PR #4017, opened 2023-12-06
.Net: Handlebars Prompt Template Helpers

teresaqhoang:
### Motivation and Context

This PR extends the Handlebars Template Factory with built-in and custom helpers to provide a consistent and convenient way for prompt and planner engineers to write Handlebars prompt templates.

These helpers bake in functionality for:
- Marking a block of text as a message with a role for chat completion connectors.
- Invoking functions from the kernel and passing parameters to them.
- Setting and getting variables in the template context.
- Performing common operations such as concatenation, arithmetic, comparison, and JSON serialization.
- Supporting different output types and formats for the rendered template.

ADR here: https://github.com/microsoft/semantic-kernel/pull/3600/files

### Description

Effectively, there are four buckets of helpers enabled in the Handlebars Template Engine:

1. Default helpers from the Handlebars library, including:
   - [Built-in helpers](https://handlebarsjs.com/guide/builtin-helpers.html) that enable loops and conditions (#if, #each, #with, #unless)
   - [Handlebars.Net.Helpers](https://github.com/Handlebars-Net/Handlebars.Net.Helpers/wiki)
2. Functions in the kernel
3. Helpers helpful to prompt engineers (i.e., message, or)
4. Utility helpers that can be used to perform simple logic or transformations on the template data or arguments (i.e., set, get, json, concat, equals, range, array)

When importing kernel functions as helpers, parameters are checked both positionally and as hash arguments to ensure expected types match and all required parameters are present.

This PR also refactors the Handlebars Planner to leverage the new Handlebars Prompt Template Factory.

Some follow-up (marked with TODOs)
- Isolate Handlebars Kernel System helpers in their own class (Issue #3947)
- Support override of default options
- Add Categories filter for KernelSystemHelpers (i.e., KernelHelperCategories)
- Add support for complex types (in parameter type checks)

### Contribution Checklist


dotnet/Directory.Packages.props
```diff
@@ -9,13 +9,15 @@
...
     <PackageVersion Include="Azure.Search.Documents" Version="11.5.0-beta.5" />
+    <PackageVersion Include="Handlebars.Net.Helpers" Version="2.4.1" />
     <PackageVersion Include="Markdig" Version="0.33.0" />
     <PackageVersion Include="Handlebars.Net" Version="2.1.4" />
     <PackageVersion Include="JsonSchema.Net.Generation" Version="3.5.0" />
     <PackageVersion Include="Microsoft.Azure.Kusto.Data" Version="11.3.5" />
     <PackageVersion Include="Microsoft.Bcl.HashCode" Version="[1.1.0, )" />
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Bcl.TimeProvider" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.SemanticKernel.PromptTemplate.Handlebars" Version="1.0.0-beta8" />
```
glahaye, line 20: Wait...  What?  Ingesting this code through a NuGet instead of as a project reference?  Is this what we want to do?

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
```diff
@@ -127,7 +127,9 @@ private static async Task PlanNotPossibleSampleAsync(bool shouldPrintPrompt = fa
             // Load additional plugins to enable planner but not enough for the given goal.
             await RunSampleAsync("Send Mary an email with the list of meetings I have scheduled today.", shouldPrintPrompt, "SummarizePlugin");
         }
-        catch (KernelException e)
+        catch (KernelException ex) when (
+            ex.Message.Contains("The plan references hallucinated helpers", StringComparison.CurrentCultureIgnoreCase)
+            || ex.Message.Contains("Unable to create plan for goal with available functions", StringComparison.CurrentCultureIgnoreCase))
```
glahaye, line 132: We just swallow these errors and do nothing when we get them?

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,72 @@
...
+/// Represents a Handlebars prompt template.
+/// </summary>
+public class HandlebarsPromptTemplate : IPromptTemplate
 {
     /// <summary>
-    /// Constructor for PromptTemplate.
+    /// Default options for built-in Handlebars helpers.
+    /// </summary>
+    /// TODO [@teresaqhoang]: Support override of default options
+    private readonly HandlebarsPromptTemplateOptions _options = new();
```
glahaye, line 20: Nit: initialization not needed since it will always be done in the constructor

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,72 @@
...
     public async Task<string> RenderAsync(Kernel kernel, KernelArguments? arguments = null, CancellationToken cancellationToken = default)
-#pragma warning restore CS1998
     {
-        var handlebars = HandlebarsDotNet.Handlebars.Create();
+        arguments = this.GetVariables(arguments);
+        var handlebarsInstance = HandlebarsDotNet.Handlebars.Create();
 
-        foreach (IKernelPlugin plugin in kernel.Plugins)
+        // Add helpers for kernel functions
+        KernelFunctionHelpers.Register(handlebarsInstance, kernel, arguments, this._options.PrefixSeparator, cancellationToken);
```
glahaye, line 43: Can kill this empty line...

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,72 @@
...
+            this._options.RegisterCustomHelpers(handlebarsInstance, arguments);
         }
 
-        var template = handlebars.Compile(this._promptModel.Template);
+        var template = handlebarsInstance.Compile(this._promptModel.Template);
+
+        var prompt = template(arguments);
 
-        return template(this.GetVariables(arguments));
+        return await Task.FromResult(prompt).ConfigureAwait(true);
```
glahaye, line 58: As a rule of thumb, you want this to be ConfigureAwait(false) in libraries, including here.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,226 @@
...
+        string fullyResolvedFunctionName = functionMetadata.PluginName + nameDelimiter + functionMetadata.Name;
+
+        handlebarsInstance.RegisterHelper(fullyResolvedFunctionName, (in HelperOptions options, in Context context, in Arguments handlebarsArguments) =>
+        {
+            // Get the parameters from the template arguments
+            if (handlebarsArguments.Any())
+            {
+                if (handlebarsArguments[0].GetType() == typeof(HashParameterDictionary))
+                {
+                    ProcessHashArguments(functionMetadata, executionContext, handlebarsArguments[0] as IDictionary<string, object>, nameDelimiter);
```
glahaye, line 60: This superfluous empty line passed our style checker?

dotnet/src/Extensions/Extensions.UnitTests/PromptTemplate/Handlebars/HandlebarsPromptTemplateTests.cs
```diff
@@ -40,67 +39,115 @@ public async Task ItRendersVariablesAsync()
...
-        });
-        promptConfig.InputVariables.Add(new InputVariable()
-        {
-            Name = "baz",
-            Description = "Baz",
-            Default = "Baz"
-        });
-        promptConfig.Template = "Foo {{bar}} {{baz}}";
+
+        this._factory = new(TestConsoleLogger.LoggerFactory, options);
```
TaoChenOSU, line 135: Is this going to affect other tests? The factory is used in other tests that may or may not run after this one.

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public RegisterCustomHelpersCallback? RegisterCustomHelpers { get; set; } = null;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HandlebarsPromptTemplateOptions"/> class.
+    /// </summary>
+    public HandlebarsPromptTemplateOptions()
+    {
+        this.PrefixSeparator = this.DefaultNameDelimiter;
+        this.Categories = new Category[] { Category.Math, Category.String };
```
TaoChenOSU, line 41: nit: could use more documentation on what `Category.Math` and `Category.String` are.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,226 @@
...
+/// Extension class to register Kernel functions as helpers.
+/// </summary>
+public static class KernelFunctionHelpers
+{
+    /// <summary>
+    /// Register all (default) or specific categories.
+    /// </summary>
+    /// <param name="handlebarsInstance">The <see cref="IHandlebars"/>-context.</param>
+    /// <param name="kernel">Kernel instance.</param>
+    /// <param name="executionContext">Kernel arguments maintained as the executing context..</param>
```
TaoChenOSU, line 25:
```suggestion
    /// <param name="executionContext">Kernel arguments maintained as the executing context.</param>
```

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,226 @@
...
+    /// <param name="executionContext">Arguments maintained in the executing context.</param>
+    /// <param name="handlebarsArguments">Arguments passed to the Handlebars helper.</param>
+    /// <param name="nameDelimiter">The character used to delimit the plugin name and function name in a Handlebars template.</param>
+    /// <exception cref="KernelException">Thrown when a required parameter is missing.</exception>
+    private static void ProcessHashArguments(
+        KernelFunctionMetadata functionMetadata,
+        KernelArguments executionContext,
+        IDictionary<string, object>? handlebarsArguments,
+        string nameDelimiter
+    )
```
TaoChenOSU, line 120: nit: closing parenthesis on the same line with the last parameter.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,226 @@
...
+        foreach (var param in functionMetadata.Parameters)
+        {
+            var fullyQualifiedParamName = functionMetadata.Name + nameDelimiter + param.Name;
+            var value = handlebarsArguments is not null && (handlebarsArguments.TryGetValue(param.Name, out var val) || handlebarsArguments.TryGetValue(fullyQualifiedParamName, out val)) ? val : null;
+
+            if (value is not null && (handlebarsArguments?.ContainsKey(param.Name) == true || handlebarsArguments?.ContainsKey(fullyQualifiedParamName) == true))
+            {
+                if (IsExpectedParameterType(param, value))
+                {
+                    executionContext[param.Name] = value;
```
TaoChenOSU, line 132: Should we check for duplicated keys?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,226 @@
...
+    /// Processes the positional arguments passed to a Handlebars helper function.
+    /// </summary>
+    /// <param name="functionMetadata">KernelFunctionMetadata for the function being invoked.</param>
+    /// <param name="executionContext">Arguments maintained in the executing context.</param>
+    /// <param name="handlebarsArguments">Arguments passed to the Handlebars helper.</param>
+    /// <exception cref="KernelException">Thrown when a required parameter is missing.</exception>
+    private static void ProcessPositionalArguments(KernelFunctionMetadata functionMetadata, KernelArguments executionContext, Arguments handlebarsArguments)
+    {
+        var requiredParameters = functionMetadata.Parameters.Where(p => p.IsRequired).ToList();
+        if (handlebarsArguments.Length >= requiredParameters.Count && handlebarsArguments.Length <= functionMetadata.Parameters.Count)
```
TaoChenOSU, line 156:
```suggestion
        if (requiredParameters.Count <= handlebarsArguments.Length && handlebarsArguments.Length <= functionMetadata.Parameters.Count)
```

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelperUtils.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <param name="type">The type to check.</param>
+    /// <returns>True if the type is a numeric type; otherwise, false.</returns>
+    public static bool IsNumericType(Type? type)
+    {
+        return type is not null &&
+            Type.GetTypeCode(type) is
+            TypeCode.SByte or TypeCode.Int16 or TypeCode.Int32 or TypeCode.Int64 or
+            TypeCode.Byte or TypeCode.UInt16 or TypeCode.UInt32 or TypeCode.UInt64 or
+            TypeCode.Double or TypeCode.Single or
+            TypeCode.Decimal;
```
TaoChenOSU, line 27:
```
            return type is not null &&
                Type.GetTypeCode(type) is
                    TypeCode.SByte or
                    TypeCode.Int16 or
                    TypeCode.Int32 or
                    TypeCode.Int64 or
                    TypeCode.Byte or
                    TypeCode.UInt16 or
                    TypeCode.UInt32 or
                    TypeCode.UInt64 or
                    TypeCode.Double or
                    TypeCode.Single or
                    TypeCode.Decimal;
```
may be more readable.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            options.PrefixSeparator = helperOptions.PrefixSeparator;
+            options.Categories = helperOptions.Categories;
+            options.UseCategoryPrefix = helperOptions.UseCategoryPrefix;
+            options.CustomHelperPaths = helperOptions.CustomHelperPaths;
+        });
+    }
+
+    /// <summary>
+    ///  Register all system helpers.
```
TaoChenOSU, line 44:
```suggestion
    /// Register all system helpers.
```

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,177 @@
...
+    /// </summary>
+    /// <param name="handlebarsInstance"></param>
+    /// <param name="variables"></param>
+    /// <exception cref="KernelException"></exception>
```
TaoChenOSU, line 48: xml docs has no content.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,177 @@
...
+    private static void RegisterSystemHelpers(
+        IHandlebars handlebarsInstance,
+        KernelArguments variables
+    )
```
TaoChenOSU, line 52: nit: closing parenthesis on the same line with the last parameter.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        handlebarsInstance.RegisterHelper("concat", (in HelperOptions options, in Context context, in Arguments arguments) =>
+        {
+            List<string?> strings = arguments.Select((var) =>
+            {
+                if (var == null)
+                {
+                    return null;
+                }
+                return var!.ToString();
```
TaoChenOSU, line 124:
```
return var == null ? null : var.ToString();
```

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlan.cs
```diff
@@ -9,6 +12,11 @@ namespace Microsoft.SemanticKernel.Planning.Handlebars;
 /// </summary>
 public sealed class HandlebarsPlan
 {
+    /// <summary>
+    /// Error message for hallucinated helpers.
+    /// </summary>
+    public const string HallucinatedHelpersErrorMessage = "Template references a helper that cannot be resolved.";
```
TaoChenOSU, line 18: Does this have to be public?

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
```diff
@@ -16,15 +20,41 @@ namespace Microsoft.SemanticKernel.Planning.Handlebars;
 /// </summary>
 public sealed class HandlebarsPlanner
 {
+    /// <summary>
+    /// Error message if kernel does not contain sufficient functions to create a plan.
+    /// </summary>
+    public const string InsufficientFunctionsError = "Additional helpers may be required";
```
TaoChenOSU, line 26: Does this have to be public?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
dmytrostruk: This should work as well: `return var?.ToString();`
All block in `RegisterHelper` could look like `return string.Concat(arguments.Select(var => var?.ToString()));`

dotnet/src/Extensions/Extensions.UnitTests/PromptTemplate/Handlebars/HandlebarsPromptTemplateTests.cs
teresaqhoang: Should be a non-effect if any. This just registers the custom helper, and none of the other templates used in other tests reference the helper, so if anything, it's just an extra function that's registered in the kernel.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
teresaqhoang: Love the optimizations suggestions, have been looking at this code for too long. Thank you!

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
teresaqhoang: no, changed

dotnet/Directory.Packages.props
teresaqhoang: somehow this was auto-added, removing

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelperUtils.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <returns>The double representation of the object.</returns>
+    public static double CastToNumber(object number)
+    {
+        try
+        {
+            return Convert.ToDouble(number, CultureInfo.CurrentCulture);
+        }
+        catch (FormatException)
+        {
+            return Convert.ToDouble(number, CultureInfo.InvariantCulture);
```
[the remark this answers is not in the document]
glahaye: Ah, I see.  Maybe put a short comment to that effect.

Also, consider renaming the method to CastToDouble

dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs
```diff
@@ -126,7 +126,9 @@ private static async Task PlanNotPossibleSampleAsync(bool shouldPrintPrompt = fa
...
```
stephentoub, line 131: We generally discourage developers from parsing exception messages. It's very fragile, both to tweaks / improvements over time and to localization (improving an exception message could break such code). If it's important to be able to differentiate these cases, we should consider throwing a more derived exception type and/or differentiating these cases with an enum error code on the exception.

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,72 @@
...
 using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.SemanticKernel.PromptTemplate.Handlebars.Helpers;
 
 namespace Microsoft.SemanticKernel.PromptTemplate.Handlebars;
 
-internal sealed class HandlebarsPromptTemplate : IPromptTemplate
+/// <summary>
+/// Represents a Handlebars prompt template.
+/// </summary>
+public class HandlebarsPromptTemplate : IPromptTemplate
```
stephentoub, line 14: This needs to be public for the 1.0 release?

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,72 @@
...
-            foreach (KernelFunction function in plugin)
-            {
-                handlebars.RegisterHelper($"{plugin.Name}_{function.Name}", (writer, hcontext, parameters) =>
-                {
-                    var result = function.InvokeAsync(kernel, arguments).GetAwaiter().GetResult();
-                    writer.WriteSafeString(result.ToString());
-                });
-            }
+            this._options.RegisterCustomHelpers(handlebarsInstance, arguments);
         }
```
stephentoub, line 52: Nit: this can be:
```C#
this._options.RegisterCustomHelpers?.Invoke(handlebarsInstance, arguments);
```

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateFactory.cs
```diff
@@ -14,11 +15,26 @@ public class HandlebarsPromptTemplateFactory : IPromptTemplateFactory
     /// <summary>The name of the Handlebars template format.</summary>
```
stephentoub, line 15: Can we seal this?

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// Delegate for registering custom helpers.
+    /// </summary>
+    /// <param name="handlebarsInstance">The Handlebars instance.</param>
+    /// <param name="executionContext">Arguments maintained in the template execution context.</param>
+    public delegate void RegisterCustomHelpersCallback(IHandlebars handlebarsInstance, KernelArguments executionContext);
+
+    /// <summary>
+    /// Callback for registering custom helpers.
+    /// </summary>
+    public RegisterCustomHelpersCallback? RegisterCustomHelpers { get; set; } = null;
```
stephentoub, line 33: The `= null;` here isn't necessary.

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// Default character to use for delimiting plugin name and function name in a Handlebars template.
+    /// </summary>
+    public string DefaultNameDelimiter { get; set; } = "-";
+
+    /// <summary>
+    /// Delegate for registering custom helpers.
+    /// </summary>
+    /// <param name="handlebarsInstance">The Handlebars instance.</param>
+    /// <param name="executionContext">Arguments maintained in the template execution context.</param>
+    public delegate void RegisterCustomHelpersCallback(IHandlebars handlebarsInstance, KernelArguments executionContext);
```
stephentoub, line 28: Do we need a custom delegate, or could we just use `Action<IHandlebars, KernelArguments>`?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,220 @@
...
+using HandlebarsDotNet.Compiler;
+
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+namespace Microsoft.SemanticKernel.PromptTemplate.Handlebars.Helpers;
+#pragma warning restore IDE0130
+
+/// <summary>
+/// Extension class to register Kernel functions as helpers.
+/// </summary>
+public static class KernelFunctionHelpers
```
stephentoub, line 18: Is this intended to be extension methods as the comment indicates? If so, it should be named with an "Extensions" suffix, and the public method(s) should actually be extensions, e.g. it should be `Register(this IHandlebars handlebars, ...`.  Also, the name of the type starts with "KernelFunction" but this actually extends "IHandlebars"... should it be "Handlebarsextensions"?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,220 @@
...
+        IHandlebars handlebarsInstance,
+        Kernel kernel,
+        KernelArguments executionContext,
+        string nameDelimiter,
+        CancellationToken cancellationToken = default)
+    {
+        kernel.Plugins.GetFunctionsMetadata().ToList()
+            .ForEach(function =>
+                RegisterFunctionAsHelper(kernel, executionContext, handlebarsInstance, function, nameDelimiter, cancellationToken)
+            );
```
stephentoub, line 38: This should just be a foreach loop, rather than allocating a List and ForEach'ing that, e.g.
```C#
foreach (var function in kernel.Plugins.GetFunctionsMetadata())
{
    RegisterFunctionAsHelper(kernel, executionContext, handlebarsInstance, function, nameDelimiter, cancellationToken);
}
```

Also, this method is taking a CancellationToken but it's not async... that's an unusual combination outside of certain domains.  What's the purpose of the CancellationToken here?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelperUtils.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+
+#pragma warning disable IDE0130 // Namespace does not match folder structure
```
stephentoub, line 5: We shouldn't need these IDE0130 suppressions anymore.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelperUtils.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Microsoft.SemanticKernel.PromptTemplate.Handlebars.Helpers;
+#pragma warning restore IDE0130
+
+/// <summary>
+/// Extension class to register additional helpers as Kernel System helpers.
+/// </summary>
+public static class KernelHelpersUtils
```
stephentoub, line 12: This needs to be public? If yes, wee should come up with a different class name, but better to keep it internal if at all possible.

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlan.cs
```diff
@@ -46,11 +54,28 @@ public override string ToString()
...
+            TemplateFormat = HandlebarsPromptTemplateFactory.HandlebarsTemplateFormat,
+            Name = "ExecutePlan",
+        };
+
+        var handlebarsTemplate = templateFactory.Create(promptTemplateConfig) as HandlebarsPromptTemplate;
+        try
+        {
+            return await handlebarsTemplate!.RenderAsync(kernel, arguments, cancellationToken).ConfigureAwait(false);
+        }
+        catch (HandlebarsRuntimeException ex) when (ex.Message.Contains(HallucinatedHelpersErrorMessage))
```
stephentoub, line 75: This is the only way to check for what we need? What happens if handlebars fixes a typo or otherwise tweaks this message?

teresaqhoang: Screenshot 2023-12-07 at 8 37 25 AM

Doesn't look like there's an error code or something else that we can check.

This is more or less an optimization to make the error scenario clearer to the user (`Helper that cannot be resolved` vs explicitly calling out `Hallucinated helpers`), so I can remove this conditional, and it would more or less have the same outcome if that's preferred.

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
teresaqhoang: Changed to internal

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
teresaqhoang: Changed to `Action<IHandlebars, KernelArguments>`

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
teresaqhoang: Removed

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using HandlebarsDotNet;
+using HandlebarsDotNet.Helpers.Enums;
+using HandlebarsDotNet.Helpers.Options;
+
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+namespace Microsoft.SemanticKernel.PromptTemplate.Handlebars.Helpers;
```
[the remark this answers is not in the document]
teresaqhoang: Removed the .Helpers

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,220 @@
...
+        KernelFunctionMetadata functionMetadata,
+        string nameDelimiter,
+        CancellationToken cancellationToken = default)
+    {
+        string fullyResolvedFunctionName = functionMetadata.PluginName + nameDelimiter + functionMetadata.Name;
+
+        handlebarsInstance.RegisterHelper(fullyResolvedFunctionName, (in HelperOptions options, in Context context, in Arguments handlebarsArguments) =>
+        {
+            // Get the parameters from the template arguments
+            if (handlebarsArguments.Any())
```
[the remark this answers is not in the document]
teresaqhoang: Heard, changed! I believe this is the only place we check, lmk if I missed elsewhere

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
teresaqhoang: `RegisterFunctionAsHelper` calls `InvokeKernelFunctionAsync`, which is where the cancellation token is actually used. 

We use a GetAwaiter there since Handlebars doesn't support async ops yet

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
teresaqhoang: Opted for the former for now

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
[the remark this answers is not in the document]
matthewbolanos: Yeah, we can go with PromptTemplates since it aligns with the other namespaces we have (e.g., Connectors, Functions, Plugins, etc.)

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelFunctionHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,220 @@
...
+            executionContext.Add(v.Key, varString);
+        }
+
+        return executionContext;
+    }
+
+    /// <summary>
+    /// Invokes an SK function and returns a typed result, if specified.
+    /// </summary>
+    private static object? InvokeKernelFunctionAsync(
```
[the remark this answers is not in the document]
teresaqhoang: Technically, it contains a synchronous wait when invoking the function since Handlebars doesn't support async operations

But I changed this back to `InvokeKernelFunction`

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,72 @@
...
 
-        var template = handlebars.Compile(this._promptModel.Template);
+        var template = handlebarsInstance.Compile(this._promptModel.Template);
 
-        return template(this.GetVariables(arguments));
+        var prompt = template(arguments);
+
+        return await Task.FromResult(prompt).ConfigureAwait(false);
```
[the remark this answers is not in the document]
stephentoub: > This method has to be asynchronous to match the interface

It has to be Task returning to match the Interface, but it doesn't need to use async; even though async is written in the method signature, it's just an implementation detail for how that task is implemented.

Regardless, you can keep the method as async, not have any awaits, and just suppress the resulting warning.

dotnet/src/InternalUtilities/src/Type/TypeExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+/// </summary>
+internal static class TypeExtensions
+{
+    /// <summary>
+    /// Tries to get the result type from a generic parameter of type Task, Nullable, or ValueTask.
+    /// </summary>
+    /// <param name="returnType">Return type.</param>
+    /// <param name="resultType">The result type of the Nullable generic parameter.</param>
+    /// <returns><c>true</c> if the result type was successfully retrieved; otherwise, <c>false</c>.</returns>
+    public static bool TryGetGenericResultType(this Type? returnType, out Type resultType)
```
teresaqhoang, line 19: @stephentoub is this the best way to do this?

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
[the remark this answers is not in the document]
teresaqhoang: Ahhhh I see, thanks for explaining! Removed and suppressed warning

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,68 @@
...
 
-        var template = handlebars.Compile(this._promptModel.Template);
+        // Add built-in system helpers
+        KernelSystemHelpers.Register(handlebarsInstance, arguments, this._options);
 
-        return template(this.GetVariables(arguments));
+        // Register any custom helpers
+        this._options.RegisterCustomHelpers?.Invoke(handlebarsInstance, arguments);
+
+        var template = handlebarsInstance.Compile(this._promptModel.Template);
```
stephentoub, line 53: We have to do this "Compile" per RenderAsync call because the compilation depends on functions that have been registered with it and those can change?

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+/// Configuration for Handlebars helpers.
+/// </summary>
+public sealed class HandlebarsPromptTemplateOptions : HandlebarsHelpersOptions
+{
+    // TODO [@teresaqhoang]: Issue #3947 Add Categories filter for KernelSystemHelpers (i.e., KernelHelperCategories)
+
+    /// <summary>
+    /// Default character to use for delimiting plugin name and function name in a Handlebars template.
+    /// </summary>
+    public string DefaultNameDelimiter { get; set; } = "-";
```
stephentoub, line 20: Is this intended to be an instance member rather than static? This appears to only be read in the ctor, which will happen before someone gets a chance to set this.  Is this even needed?

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
teresaqhoang: Yep exactly, we spin up a new Hb instance per prompt template so the functions should change on each render

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelperUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+                    TypeCode.Int16 or
+                    TypeCode.Int32 or
+                    TypeCode.Int64 or
+                    TypeCode.Byte or
+                    TypeCode.UInt16 or
+                    TypeCode.UInt32 or
+                    TypeCode.UInt64 or
+                    TypeCode.Double or
+                    TypeCode.Single or
+                    TypeCode.Decimal;
```
stephentoub, line 31: This should probably also include `type is nint || type is nuint`

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+        KernelArguments variables)
+    {
+        // TODO [@teresaqhoang]: Issue #3947 Isolate Handlebars Kernel System helpers in their own class
+        // Should also consider standardizing the naming conventions for these helpers, i.e., 'Message' instead of 'message'
+        handlebarsInstance.RegisterHelper("message", (writer, options, context, arguments) =>
+        {
+            var parameters = arguments[0] as IDictionary<string, object>;
+
+            // Verify that the message has a role
+            if (!parameters!.ContainsKey("role"))
```
stephentoub, line 60: Casting with `as` followed by immediate usage is an anti-pattern.  Better to instead use a normal cast, e.g.
```C#
var parameters = (IDictionary<string, object>)arguments[0];

if (!parameters!.ContainsKey("role"))
```
If you use `as` and the type isn't of the right type, it'll result in the failure being a NullReferenceException. If you use a normal cast and the type isn't of the right type, it'll result in a cast exception that includes more helpful information.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            if (variables.TryGetValue(name, out var argVal))
+            {
+                variables[name] = value;
+            }
+            else
+            {
+                variables.Add(name, value);
+            }
```
stephentoub, line 94:
```suggestion
            variables[name] = value;
```
?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        handlebarsInstance.RegisterHelper("raw", (writer, options, context, arguments) =>
+        {
+            options.Template(writer, null);
+        });
+
+        handlebarsInstance.RegisterHelper("array", (in HelperOptions options, in Context context, in Arguments arguments) =>
+        {
+            // Convert all the arguments to an array
+            var array = arguments.Select(a => a).ToList();
```
stephentoub, line 129: `return arguments.ToList();`?

Also, the comment says "to an array", but this is converting it to a `List<T>` not a `T[]`?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            return array;
+        });
+
+        handlebarsInstance.RegisterHelper("range", (in HelperOptions options, in Context context, in Arguments arguments) =>
+        {
+            var start = int.Parse(arguments[0].ToString(), CultureInfo.InvariantCulture);
+            var end = int.Parse(arguments[1].ToString(), CultureInfo.InvariantCulture) + 1;
```
stephentoub, line 137: Should this be using the CultureInfo from the Kernel? Or is the Kernel not available here?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            var count = end - start;
+
+            // Create array from start to end
+            var array = Enumerable.Range(start, count).ToList();
+
+            return array;
```
stephentoub, line 144: Same question about "array"

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+                var parameters = arguments[0] as IDictionary<string, object>;
+                return variables[(string)parameters!["name"]];
+            }
+
+            return variables[arguments[0].ToString()];
+        });
+
+        handlebarsInstance.RegisterHelper("json", (in HelperOptions options, in Context context, in Arguments arguments) =>
+        {
+            object objectToSerialize = arguments[0];
```
stephentoub, line 110: In some of these helpers, the value from the arguments is checked for null, which suggests they can be null. In other places, though, like this one, we're just dereferencing.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
teresaqhoang: Ohhh good to know, changed it. Thank you!

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelperUtils.cs
teresaqhoang: Seeing this warning when I add this:
`[CS8121](classifiedtextelementview-invokeaction:0): An expression of type 'Type' cannot be handled by a pattern of type 'nint'.`

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
teresaqhoang: No, I removed it altogether, set `"-"` as the default delimiter in the constructor

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelperUtils.cs
stephentoub: Sorry, I was typing too quickly, it should be:
```C#
return
    type == typeof(nuint) ||
    type == typeof(nint) ||
    ... // what you currently have
```

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
teresaqhoang: If I remember correctly, this threw an error during testing if an entry with that key already existed in variables dict, (or maybe I was just using `.Add` before) -- will have to double test this

stephentoub: > this threw an error during testing if an entry with that key already existed in variables dict

It shouldn't. It should overwrite an existing entry if it's already there or add a new entry if it doesn't.

teresaqhoang: Awesome, changed

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+            string json = objectToSerialize.GetType() == typeof(string) ? (string)objectToSerialize : JsonSerializer.Serialize(objectToSerialize);
+
+            return json;
+        });
+
+        handlebarsInstance.RegisterHelper("concat", (in HelperOptions options, in Context context, in Arguments arguments) =>
+        {
+            return string.Concat(arguments.Select(var => var?.ToString()));
```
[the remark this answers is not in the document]
teresaqhoang: Heard, changed `string.Concat(arguments)`

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
teresaqhoang: Good catch, changed to array since it's a fixed size

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
teresaqhoang: Drilled down the kernel to use `kernel.Culture`

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
teresaqhoang: Changed to ToArray

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/SystemHelpers/KernelSystemHelpers.cs
teresaqhoang: Added null and empty argument checks for the "json" and "equals" helpers

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplateOptions.cs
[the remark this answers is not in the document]
teresaqhoang: Reverted to `PromptTemplate` (without s) to avoid all making all the same changes in sk abstractions and core per @markwallace-microsoft rec, @matthewbolanos signed off in standup

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlan.cs
```diff
@@ -53,21 +59,36 @@ public override string ToString()
         KernelArguments arguments,
         CancellationToken cancellationToken = default)
     {
-        var logger = kernel.LoggerFactory.CreateLogger(typeof(HandlebarsPlan));
+        var logger = kernel.LoggerFactory.CreateLogger(typeof(HandlebarsPlan).ToString());
```
teresaqhoang, line 62: @TaoChenOSU I had to add this bc it was throwing a Casting error here, something like "could not convert from object to string"

stephentoub: > @TaoChenOSU I had to add this bc it was throwing a Casting error here, something like "could not convert from object to string"

That should not throw. Was this in a particular test where the ILoggerFactory might have been customized to something unexpected?

markwallace-microsoft (commented):
I still have some files to review, but adding the feedback I have so far.

dotnet/src/Extensions/PromptTemplate.Handlebars/PromptTemplate.Handlebars.csproj
```diff
@@ -2,8 +2,8 @@
 
   <PropertyGroup>
     <!-- THIS PROPERTY GROUP MUST COME FIRST -->
-    <AssemblyName>Microsoft.SemanticKernel.PromptTemplate.Handlebars</AssemblyName>
-    <RootNamespace>Microsoft.SemanticKernel.PromptTemplate.Handlebars</RootNamespace>
+    <AssemblyName>Microsoft.SemanticKernel.PromptTemplates.Handlebars</AssemblyName>
+    <RootNamespace>Microsoft.SemanticKernel.PromptTemplates.Handlebars</RootNamespace>
```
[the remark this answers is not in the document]
teresaqhoang: I think the files may have been outdated when you were reviewing -- I pushed a change that reverted this to `Microsoft.SemanticKernel.PromptTemplate` (without the s).

Let's continue this discussion on the thread with Stephen in the [Example64_MultiplePromptTemplates.cs](https://github.com/microsoft/semantic-kernel/pull/4017/files/0bb965397e0ea233c80cca2845c9c16795eda552#diff-2b768740c500845069d753c45bb72f28f74784bc58bade3b1277e6496df1094a) file

dotnet/src/Extensions/PromptTemplate.Handlebars/Extensions/HandlebarsKernelExtensions.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System.Threading.Tasks;
-using Microsoft.SemanticKernel.PromptTemplate.Handlebars;
+using Microsoft.SemanticKernel.PromptTemplates.Handlebars;
```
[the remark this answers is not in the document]
teresaqhoang: I think the files may have been outdated when you were reviewing -- I pushed a change that reverted this to `Microsoft.SemanticKernel.PromptTemplate` (without the s).

Let's continue this discussion on the thread with Stephen in the [Example64_MultiplePromptTemplates.cs](https://github.com/microsoft/semantic-kernel/pull/4017/files/0bb965397e0ea233c80cca2845c9c16795eda552#diff-2b768740c500845069d753c45bb72f28f74784bc58bade3b1277e6496df1094a) file

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,68 @@
...
+    private readonly HandlebarsPromptTemplateOptions _options;
+
+    /// <summary>
+    /// Constructor for Handlebars PromptTemplate.
     /// </summary>
     /// <param name="promptConfig">Prompt template configuration</param>
     /// <param name="loggerFactory">Logger factory</param>
-    public HandlebarsPromptTemplate(PromptTemplateConfig promptConfig, ILoggerFactory? loggerFactory = null)
+    /// <param name="options">Handlebars promnpt template options</param>
+    public HandlebarsPromptTemplate(PromptTemplateConfig promptConfig, ILoggerFactory? loggerFactory = null, HandlebarsPromptTemplateOptions? options = null)
```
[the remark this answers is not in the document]
teresaqhoang: Dropped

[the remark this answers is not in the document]
teresaqhoang: Removed

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,68 @@
...
+        // Add helpers for kernel functions
+        KernelFunctionHelpers.Register(handlebarsInstance, kernel, arguments, this._options.PrefixSeparator, cancellationToken);
 
-        var template = handlebars.Compile(this._promptModel.Template);
+        // Add built-in system helpers
+        KernelSystemHelpers.Register(handlebarsInstance, arguments, this._options);
 
-        return template(this.GetVariables(arguments));
+        // Register any custom helpers
+        this._options.RegisterCustomHelpers?.Invoke(handlebarsInstance, arguments);
```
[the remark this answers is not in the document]
teresaqhoang: Added `RegisterHelperSafe` to KernelHelpersUtils that will throw an error on conflicts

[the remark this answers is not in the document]
teresaqhoang: Need to update tests

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+
+    #region private
+
+    private static void RegisterFunctionAsHelper(
+        Kernel kernel,
+        KernelArguments executionContext,
+        IHandlebars handlebarsInstance,
+        KernelFunctionMetadata functionMetadata,
+        string nameDelimiter,
+        CancellationToken cancellationToken = default)
```
SergeyMenshykh, line 48: As a good practice to follow, use default cancellation tokens only for public APIs and not for internal ones. This ensures proper passing of the token from the public API down to lower levels, preventing bugs. The API consumer code may decide whether to pass it or not, while internally, it will be passed through all levels.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+                    }
+                    else
+                    {
+                        ProcessPositionalArguments(functionMetadata, executionContext, handlebarsArguments);
+                    }
+                }
+                else if (functionMetadata.Parameters.Any(p => p.IsRequired))
+                {
+                    throw new KernelException($"Invalid parameter count for function {functionMetadata.Name}. {handlebarsArguments.Length} were specified but {functionMetadata.Parameters.Count} are required.");
+                }
```
SergeyMenshykh, line 72:
```suggestion
                    throw new ArgumentException($"No arguments are provided for {fullyResolvedFunctionName}.");
```

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+        KernelFunctionMetadata functionMetadata,
+        KernelArguments executionContext,
+        IDictionary<string, object>? handlebarsArguments,
+        string nameDelimiter)
+    {
+        // Prepare the input parameters for the function
+        foreach (var param in functionMetadata.Parameters)
+        {
+            var fullyQualifiedParamName = functionMetadata.Name + nameDelimiter + param.Name;
+            var value = handlebarsArguments is not null && (handlebarsArguments.TryGetValue(param.Name, out var val) || handlebarsArguments.TryGetValue(fullyQualifiedParamName, out val)) ? val : null;
```
SergeyMenshykh, line 124: Shouldn't we first use a specific parameter like "fullyQualifiedParamName" to resolve the argument, and if not found, use a less specific one "param.Name"?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            if (value is not null && (handlebarsArguments?.ContainsKey(param.Name) == true || handlebarsArguments?.ContainsKey(fullyQualifiedParamName) == true))
```
SergeyMenshykh, line 126: Is this additional check necessary: `(handlebarsArguments?.ContainsKey(param.Name) == true || handlebarsArguments?.ContainsKey(fullyQualifiedParamName) == true)`? It seems that it's already done in the line above.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// Checks if handlebars argument is a valid type for the function parameter.
+    /// Must satisfy one of the following:
+    /// Types are an exact match.
+    /// Argument is any kind of numeric type if function parameter requires a numeric type.
+    /// Argument type is an object (this covers complex types).
+    /// Function parameter is a generic type.
+    /// </summary>
+    /// <param name="parameterType">Function parameter type.</param>
+    /// <param name="argument">Handlebar argument.</param>
+    private static bool IsExpectedParameterType(KernelParameterMetadata parameterType, object argument)
```
SergeyMenshykh, line 91: It's not type, right? metadata?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        var actualParameterType = parameterType.ParameterType.TryGetGenericNullableType(out var nullableType) ? nullableType : parameterType.ParameterType;
+
+        bool parameterIsNumeric = KernelHelpersUtils.IsNumericType(actualParameterType)
+            || (parameterType.Schema?.RootElement.TryGetProperty("type", out JsonElement typeProperty) == true && typeProperty.GetString() == "number");
+
+        bool argIsNumeric = KernelHelpersUtils.IsNumericType(argument?.GetType())
```
SergeyMenshykh, line 98: The `?` in the call `argument?.GetType()` assumes that the argument is nullable, but the method signature does not indicate this.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            if (value is not null && (handlebarsArguments?.ContainsKey(param.Name) == true || handlebarsArguments?.ContainsKey(fullyQualifiedParamName) == true))
+            {
+                if (IsExpectedParameterType(param, value))
+                {
+                    executionContext[param.Name] = value;
+                }
+                else
+                {
+                    throw new KernelException($"Invalid parameter type for function {functionMetadata.Name}. Parameter {param.Name} expects type {param.ParameterType ?? (object?)param.Schema} but received {value.GetType()}.");
```
SergeyMenshykh, line 134: It's the argument, which type is not valid, not the parameter's one, right?

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -1,60 +1,98 @@
...
-    /// <param name="loggerFactory">Logger factory</param>
-    public HandlebarsPromptTemplate(PromptTemplateConfig promptConfig, ILoggerFactory? loggerFactory = null)
+    /// <param name="options">Handlebars prompt template options</param>
+    public HandlebarsPromptTemplate(PromptTemplateConfig promptConfig, HandlebarsPromptTemplateOptions options)
     {
-        this._loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
+        this._loggerFactory ??= NullLoggerFactory.Instance;
         this._logger = this._loggerFactory.CreateLogger(typeof(HandlebarsPromptTemplate));
         this._promptModel = promptConfig;
+        this._options = options ?? new();
```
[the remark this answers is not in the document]
teresaqhoang: Yes good call, changed!

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+{
+    /// <summary>
+    /// Register all (default) or specific categories.
+    /// </summary>
+    /// <param name="handlebarsInstance">The <see cref="IHandlebars"/>-context.</param>
+    /// <param name="kernel">Kernel instance.</param>
+    /// <param name="executionContext">Kernel arguments maintained as the executing context.</param>
+    /// <param name="nameDelimiter">The character used to delimit the plugin name and function name in a Handlebars template.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    public static void Register(
```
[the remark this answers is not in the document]
teresaqhoang: Can do, but this was implemented this way to follow the syntax set by the [Handlebars.Helpers](https://github.com/Handlebars-Net/Handlebars.Net.Helpers) library:
`HandlebarsHelpers.Register(handlebarsContext, Category.Math, Category.String);`

Originally recommended by @markwallace-microsoft, thoughts?

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            RegisterFunctionAsHelper(kernel, executionContext, handlebarsInstance, function, nameDelimiter, cancellationToken);
+        }
+    }
+
+    #region private
+
+    private static void RegisterFunctionAsHelper(
+        Kernel kernel,
+        KernelArguments executionContext,
```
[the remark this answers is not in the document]
teresaqhoang: Yeah, a property bag of variables that's maintained by the Handlebars render context (which includes any initial arguments and function outputs / variables). Should I name it `handlebarsVariables` instead? I want to avoid something like `handlebarsArguments` because there's also a set of arguments that come in with each function call when rendering the template (See `InvokeKernelFunction` in `KernelFunctionHelpers`.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
teresaqhoang: heard, removed all ` = default` on cancellation tokens here

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
teresaqhoang: Good catch, I consolidated it within the if statement
```
if (handlebarsArguments is not null && (handlebarsArguments.TryGetValue(fullyQualifiedParamName, out var value) || handlebarsArguments.TryGetValue(param.Name, out var value)))
```

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
teresaqhoang: touché, this method only checks the ParameterType property of `KernelParameterMetadata` -- I changed the name of the parameter to `parameterMetadata`

```suggestion
    private static bool IsExpectedParameterType(KernelParameterMetadata parameterMetadata, object argument)
```

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
teresaqhoang: removed the `?`

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
teresaqhoang: yes, changed to

```suggestion
                    throw new KernelException($"Invalid argument type for function {functionMetadata.Name}. Parameter {param.Name} expects type {param.ParameterType ?? (object?)param.Schema} but received {value.GetType()}.");
```

dotnet/samples/KernelSyntaxExamples/Example64_MultiplePromptTemplates.cs
```diff
@@ -3,7 +3,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.SemanticKernel;
-using Microsoft.SemanticKernel.PromptTemplate.Handlebars;
+using Microsoft.SemanticKernel.PromptTemplates.Handlebars;
```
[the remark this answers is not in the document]
teresaqhoang: Changed everything to `Microsoft.SemanticKernel.PromptTemplates`. @markwallace-microsoft said he would look after deprecating the old package post V1

dotnet/src/Extensions/PromptTemplate.Handlebars/HandlebarsPromptTemplate.cs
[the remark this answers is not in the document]
teresaqhoang: Added unit tests in `KernelHelperUtilsTests`

dotnet/src/InternalUtilities/src/Type/TypeExtensions.cs
teresaqhoang: Chatted with Stephen offline -- Will leave as is for now since this doesn't impact the public API surface area - It's an implementation detail. 

He suggested either temporarily living with the overhead, or putting a cache in place, e.g. a static readonly ConcurrentDictionary that maps the input to the output, so that you only need to compute the answer for a given type once. Issue #4202 to track this

dotnet/src/Extensions/PromptTemplates.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.SemanticKernel.PromptTemplates.Handlebars.Helpers;
+
+namespace Microsoft.SemanticKernel.PromptTemplates.Handlebars;
+
+/// <summary>
+/// Represents a Handlebars prompt template.
+/// </summary>
+internal class HandlebarsPromptTemplate : IPromptTemplate
```
teresaqhoang, line 16: @markwallace-microsoft should this be sealed?

markwallace-microsoft: Yes, it should be

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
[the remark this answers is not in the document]
SergeyMenshykh: Ok, if that's agreed pattern that let's keep it.

dotnet/src/Extensions/PromptTemplate.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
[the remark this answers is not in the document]
SergeyMenshykh: Got it, in this case context make sense. Let's keep it.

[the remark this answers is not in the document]
teresaqhoang: Sounds good, this is just an implementation detail so we can change it in the future with little effect.

SergeyMenshykh (approved):

alliscode (approved):

markwallace-microsoft (approved):

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlan.cs
```diff
@@ -61,13 +68,28 @@ public override string ToString()
...
+            TemplateFormat = HandlebarsPromptTemplateFactory.HandlebarsTemplateFormat,
+            Name = "InvokeHandlebarsPlan",
+        };
+
+        var handlebarsTemplate = templateFactory.Create(promptTemplateConfig);
+        try
+        {
+            return await handlebarsTemplate!.RenderAsync(kernel, arguments, cancellationToken).ConfigureAwait(false);
+        }
+        catch (HandlebarsRuntimeException ex) when (ex.Message.Contains(HallucinatedHelpersErrorMessage))
```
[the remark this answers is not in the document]
teresaqhoang: Handlebars, added a new integration tests that checks for this error case under `HandlebarsPlanTests.cs`

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlan.cs
```diff
@@ -11,6 +13,11 @@ namespace Microsoft.SemanticKernel.Planning.Handlebars;
 /// </summary>
 public sealed class HandlebarsPlan
 {
+    /// <summary>
+    /// Error message for hallucinated helpers.
```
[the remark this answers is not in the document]
teresaqhoang: helpers that are not built-in system helpers or registered kernel functions, added more details

PR #4023, opened 2023-12-06
.Net: Address API review feedback about DI integration

stephentoub:
- Replaces WithServices/WithPlugins callback-based methods with Services/Plugins properties. It makes it less fluent, but it brings it into line with the current best-practices patterns used by ASP.NET, it reduces nesting, it avoids confusion about when the callbacks are invoked, it avoids async-related issues, and it makes the pre-Build syntax for adding plugins to the builder much closer to the post-Build syntax for adding plugins to the Kernel.
- Plugins added as services are now automatically resolved as part of Kernel construction. The Kernel will prefer to use a collection it's explicitly handed, but if it isn't given one, it'll try to fetch a collection from the services, and if it can't find that, it'll try to fetch all IKernelPlugins from the services.
- Renamed With methods to Add methods to improve parity both with current ASP.NET practices as well as between when doing `builder.AddXx` and `builder.Services.AddXx`.
- Renamed `AddPluginFromObject<T>` to `AddFromType<T>`.
- Remove `new()` constraint from `AddFromType<T>` and instead use ActivatorUtilities.CreateInstance. The KernelBuilder.Plugins property contains the same IServiceCollection as the builder, and so `AddFromType<T>` has access to it and can use it with CreateInstance to resolve any plugin object ctor arguments from the service provider. For example, HttpPlugin has a ctor that takes an HttpClient, so now doing `builder.Plugins.AddFromType<HttpPlugin>()` will find an HttpClient in the container and use it automatically when the plugins are built as part of Build.
- Removed WithLoggerFactory and WithCulture. The latter is basically never used, and it's much more like the Data property and event handlers on the Kernel, where it can easily be mutated on the Kernel instance and doesn't need any builder support. WithLoggerFactory was used a lot, but mainly because of the patterns employed in the tests, and it's easily replacable with AddLogging.
- Added IKernelBuilder interface. It's implemented by KernelBuilder and returned from a new AddKernel extension method on IServiceCollection, adding a singleton KernelPluginCollection and a transient Kernel to the container. IKernelBuilder has both the Services and Plugins properties, and also has Build, but Build throws an exception if it's used on the instance returned from AddKernel, per feedback that it can lead to folks accidentally building the container multiple times.
- Changed all of the extensions on KernelBuilder to instead by on IKernelBuilder.

@davidfowl, @halter73, please ensure this is in line with your feedback. Thanks!


dotnet/src/Experimental/Orchestration.Flow.IntegrationTests/FlowOrchestratorTests.cs
```diff
@@ -104,7 +104,7 @@ private KernelBuilder InitializeKernelBuilder()
         Assert.NotNull(azureOpenAIConfiguration);
 
         return new KernelBuilder()
-            .WithLoggerFactory(this._logger)
+            .his._logger)
```
WeihanLi, line 107: Maybe want to remove this line of code?
```suggestion

```

dotnet/src/Plugins/Plugins.Core/HttpPlugin.cs
```diff
@@ -31,8 +32,9 @@ public HttpPlugin() : this(HttpClientProvider.GetHttpClient())
     /// <remarks>
     /// <see cref="HttpPlugin"/> assumes ownership of the <see cref="HttpClient"/> instance and will dispose it when the plugin is disposed.
     /// </remarks>
-    public HttpPlugin(HttpClient client) =>
-        this._client = client;
+    [ActivatorUtilitiesConstructor]
+    public HttpPlugin(HttpClient? client = null) =>
```
WeihanLi, line 36: Is there any benefit for adding a default value? Wondering that if we had `HttpClient` registed as a service in the serviceProvider, would the client be `null` or be the service we injected?

stephentoub: Without a default value, it makes the HttpClient required as part of service resolution, meaning trying to get one of these via DI will throw an exception of there's no HttpClient configured in the container.

markwallace-microsoft (commented):
Not fully done reviewing but some comments

For code like this 

```csharp
        KernelBuilder builder = new();
        builder.AddOpenAIChatCompletion(this._model.Model, this._restContext.ApiKey);
        this.Kernel = builder.Build();
```

Is there a more fluent approach we can use?

stephentoub:
> Not fully done reviewing but some comments
> 
> For code like this
> 
> ```cs
>         KernelBuilder builder = new();
>         builder.AddOpenAIChatCompletion(this._model.Model, this._restContext.ApiKey);
>         this.Kernel = builder.Build();
> ```
> 
> Is there a more fluent approach we can use?

That can be:
```C#
this.Kernel = new KernelBuilder()
    .AddOpenAIChatCompletion(this._model.Model, this._restContext.ApiKey);
    .Build();
```
Fluency there is fine. It just stops being fluent once you need to use .Services directly or choose to use .Plugins directly. The only solution I have for that is to add back a Configure/WithServices method or Configure/WithPlugins method, e.g.
```C#
public static IKernelBuilder ConfigureServices(this IKernelBuilder builder, Action<IServiceCollection> configure)
{
    configure(builder.Services);
    return builder;
}
```
That can be layered on top by anyone,  including but not limited to us.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -52,7 +52,16 @@ public sealed class Kernel
         KernelPluginCollection? plugins = null)
     {
         this.Services = services ?? EmptyServiceProvider.Instance;
-        this._plugins = plugins;
+
+        this._plugins = plugins ?? this.Services.GetService<KernelPluginCollection>();
+        if (this._plugins is null)
+        {
+            IEnumerable<IKernelPlugin> e = this.Services.GetServices<IKernelPlugin>();
+            if (e is not ICollection<IKernelPlugin> c || c.Count != 0)
```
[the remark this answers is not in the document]
stephentoub: It's just an optimization to avoid allocating the plugins collection if there aren't any plugins directly registered in DI. I'll add a comment.

SergeyMenshykh (approved):

dotnet/src/Plugins/Plugins.Core/HttpPlugin.cs
WeihanLi: Should we consider removing the parameter-less constructor because this constructor has a default value, if could be removed, maybe the `ActivatorUtilitiesConstructor` can be removed since there's only one constructor then

stephentoub: We could... I kept it so that anyone with a `new()` constraint would still work.

markwallace-microsoft (approved):

markwallace-microsoft (approved):

PR #4055, opened 2023-12-06
.Net: Nulls for function arguments

SergeyMenshykh:
### Description

Missing arguments or arguments with null values are passed to the method call as null arguments, rather than empty strings. The behavior for regular prompt variables, such as {{$foo}}, remains the same, with an empty string being inserted into the prompt. This change ensures that the behavior of the kernel prompt template is consistent with the handlebars prompt template.


dotnet/src/SemanticKernel.UnitTests/TemplateEngine/Blocks/VarBlockTests.cs
```diff
@@ -37,33 +37,33 @@ public void ItIgnoresSpacesAround()
...
+        Assert.Null(result);
     }
 
     [Fact]
-    public void ItRendersToEmptyStringIfArgumentIsMissing()
+    public void ItRendersToNullWithNullArgument()
     {
         // Arrange
-        var target = new VarBlock("  $var \n ");
+        var target = new VarBlock("$var ");
```
RogerBarreto, line 56: Can't understand what this test is really doing.

`$var` isnt supposed to be a parameter named `var` ? 

If you create an argument `foo` and pass to the target, it means that if you have a variable foo, it won't be case to var, is that it?

SergeyMenshykh: Right, it was typo that is fixed now. Thanks for catching.

markwallace-microsoft (approved):

PR #4059, opened 2023-12-06
.Net: Add KernelArguments methods to simplify working with arguments

alexminza:
### Motivation and Context
After the recent breaking changes in RC1 https://github.com/microsoft/semantic-kernel/pull/3911 it has become more difficult and cumbersome to work with the usual string parameters as before.

### Description

This PR adds extensions methods `TryGetStringValue` and `GetStringInput` to facilitate writing easily readable code for the most frequent use cases.

Examples:
```csharp
variables.TryGetStringValue(ChatHistoryKey, out string? chatHistoryString);
string question = variables.GetInput()!;
```

### Contribution Checklist


alexminza:

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArgumentsExtensions.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>Provides extension methods for working with <see cref="KernelArguments"/> string values.</summary>
+public static class KernelArgumentsExtensions
```
stephentoub, line 6: _If_ it was deemed important to support something here, my preference would be:
1. Not extension methods: if the extension methods class is in the same namespace and assembly as the class it's extending, the members could just be instance methods
2. The methods not be specific to `string` but instead be generic.

markwallace-microsoft:
@stephentoub Thoughts on this PR?

stephentoub:
> @stephentoub Thoughts on this PR?

My personal opinion is I'd rather not have an API like this in SK. It's not SK-specific, and is instead could just be a helper in someone's project or in whatever nuget package as an extension method for any `IDictionary<TKey, object?>`, not just KernelArguments.

PR #4075, opened 2023-12-07
.Net: Remove implicit operator string from base content types

stephentoub:
These base types shouldn't implicitly cast to string. It might be ok for derived types that wrap a string, but it doesn't make sense for other content types that don't have a natural string representation or that would need to allocate to create one.


dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletionWithData/AzureOpenAIChatCompletionWithDataService.cs
```diff
@@ -66,7 +66,7 @@ public IAsyncEnumerable<StreamingChatMessageContent> GetStreamingChatMessageCont
     public async Task<IReadOnlyList<TextContent>> GetTextContentsAsync(string prompt, PromptExecutionSettings? executionSettings = null, Kernel? kernel = null, CancellationToken cancellationToken = default)
     {
         return (await this.GetChatMessageContentsAsync(prompt, executionSettings, kernel, cancellationToken).ConfigureAwait(false))
-            .Select(chat => new TextContent(chat.Content, chat, chat.ModelId, Encoding.UTF8, chat.Metadata))
+            .Select(chat => new TextContent(chat.Content, chat.ModelId, chat.Content, Encoding.UTF8, chat.Metadata))
```
RogerBarreto, line 69: You need to pass the ChatMessage as the innercontent in here otherwise it will break the InvokeAsync<T> when asking for a `ChatMessageCOntent` when the Service supports `IChatCompletion` and `ITextGeneration`.

```suggestion
            .Select(chat => new TextContent(chat.Content, chat.ModelId, chat, Encoding.UTF8, chat.Metadata))
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletionWithData/AzureOpenAIChatCompletionWithDataService.cs
```diff
@@ -79,7 +79,7 @@ public async Task<IReadOnlyList<TextContent>> GetTextContentsAsync(string prompt
     {
         await foreach (var streamingChatContent in this.InternalGetChatStreamingContentsAsync(new ChatHistory(prompt), executionSettings, kernel, cancellationToken).ConfigureAwait(false))
         {
-            yield return new StreamingTextContent(streamingChatContent.Content, streamingChatContent.ChoiceIndex, streamingChatContent, streamingChatContent.ModelId, Encoding.UTF8, streamingChatContent.Metadata);
+            yield return new StreamingTextContent(streamingChatContent.Content, streamingChatContent.ChoiceIndex, streamingChatContent.Content, streamingChatContent.ModelId, Encoding.UTF8, streamingChatContent.Metadata);
```
RogerBarreto, line 82: Spotted a bug in here (modelId and innerobject arguments are inverted), I will create UT to ensure we are using the correct arguments for all contents.

For same reasons above, we need to pass the streamingChat as the innerobject.
```suggestion
            yield return new StreamingTextContent(streamingChatContent.Content, streamingChatContent.ChoiceIndex, streamingChatContent.ModelId, streamingChatContent, Encoding.UTF8, streamingChatContent.Metadata);
```

dotnet/src/Connectors/Connectors.AI.OpenAI/ChatCompletionWithData/AzureOpenAIChatCompletionWithDataService.cs
RogerBarreto: Proposed a big change to this as I just realized the InvokeASync<T> today doesn't have access to the Connector output, only the `.Text` string.

stephentoub: Fixed

markwallace-microsoft (approved):

SergeyMenshykh (approved):

PR #4084, opened 2023-12-07
.Net: Change IKernelPlugin to be abstract class KernelPlugin

markwallace-microsoft:
### Motivation and Context

Changing `IKernelPlugin` to be abstract class `KernelPluginBase` so that we have the flexibility to extend that we get from abstract classes.

### Contribution Checklist


stephentoub:
What's the reason for this change?

If we want to convert IKernelPlugin to be a base class instead, I'd suggest a few changes:
- Just have it be called KernelPlugin rather than ReadOnlyKernelPlugin.
- Change the existing KernelPlugin to be named something more like ComposedKernelPlugin, DynamicKernelPlugin, ManualKernelPlugin, etc.  The number of folks that will need to consume / work with the base class is much, much larger than the number of folks that need to manually build up a new KernelPlugin from individual functions, so we should use the "good" name for the base class.
- Any extension method we currently have in Abstractions (and possibly in Core, depending on what it is) around an IKernelPlugin should move to being an instance member of the base KernelPlugin.

stephentoub:
> Change the existing KernelPlugin to be named something more like ComposedKernelPlugin, DynamicKernelPlugin, ManualKernelPlugin, etc. The number of folks that will need to consume / work with the base class is much, much larger than the number of folks that need to manually build up a new KernelPlugin from individual functions, so we should use the "good" name for the base class.

Actually, the current KernelPlugin (the mutable one) probably doesn't even need to be public. You just have a KernelPluginFactory method that takes a name and an IEnumerable<KernelFunction> and it returns an internal type we can name whatever we want.

stephentoub:
> > Change the existing KernelPlugin to be named something more like ComposedKernelPlugin, DynamicKernelPlugin, ManualKernelPlugin, etc. The number of folks that will need to consume / work with the base class is much, much larger than the number of folks that need to manually build up a new KernelPlugin from individual functions, so we should use the "good" name for the base class.
> 
> Actually, the current KernelPlugin (the mutable one) probably doesn't even need to be public. You just have a KernelPluginFactory method that takes a name and an `IEnumerable<KernelFunction>` and it returns an internal type we can name whatever we want. Then we just use `KernelPlugin` as the good name for the abstract base class.

I like this for the additional reason that it means b someone can't downcast to the mutable type and start adding functions to some random plugin.

markwallace-microsoft:
> > Change the existing KernelPlugin to be named something more like ComposedKernelPlugin, DynamicKernelPlugin, ManualKernelPlugin, etc. The number of folks that will need to consume / work with the base class is much, much larger than the number of folks that need to manually build up a new KernelPlugin from individual functions, so we should use the "good" name for the base class.
> 
> Actually, the current KernelPlugin (the mutable one) probably doesn't even need to be public. You just have a KernelPluginFactory method that takes a name and an `IEnumerable<KernelFunction>` and it returns an internal type we can name whatever we want. Then we just use `KernelPlugin` as the good name for the abstract base class.

For tooling scenarios we want to allow a developer to add functions to the Plugin. This is what Diego does in his .Net Interactive tool. So if we don't make the implementation that allows this public, do we add extension methods for `AddFunction` and `AddFunctions` that only work when the implementation is ours?

stephentoub:
> For tooling scenarios we want to allow a developer to add functions to the Plugin. This is what Diego does in his .Net Interactive tool

I see. Add functions one by one? In that case, yes, _someone_ would need a mutable implementation, though it doesn't necessarily need to be in the core. With most of the functionality in the base class, someone who needs that interactivity would write the class which would be pretty short, right?

dotnet/src/SemanticKernel.Abstractions/Functions/ReadOnlyKernelPlugin.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>Represents a plugin that may be registered with a <see cref="Kernel"/>.</summary>
+/// <remarks>
+/// A plugin is named read-only collection of functions. There is a many-to-many relationship between
+/// plugins and functions: a plugin may contain any number of functions, and a function may
+/// exist in any number of plugins.
+/// </remarks>
+public abstract class ReadOnlyKernelPlugin : IEnumerable<KernelFunction>
```
RogerBarreto, line 18: THis class can be extended to not be readonly, I would avoid setting this as a `ReadONly` in the name, maybe `KernelPluginBase` ?

markwallace-microsoft: Using `KernelPlugin`

dotnet/src/SemanticKernel.Core/KernelExtensions.cs
```diff
@@ -612,15 +612,17 @@ public static IKernelBuilder AddKernel(this IServiceCollection services)
         ILoggerFactory? loggerFactory = null)
     {
         Verify.NotNull(plugin);
+        var defaultPlugin = plugin as DefaultKernelPlugin;
```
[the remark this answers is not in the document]
markwallace-microsoft: Added a sample showing how to create a mutable KernelPlugin

SergeyMenshykh (approved):

dotnet/samples/KernelSyntaxExamples/Resources/Example73_MutableKernelPlugin.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+            this._functions.Add(function.Name, function);
+        }
+
+        /// <summary>Adds all of the functions in the specified <paramref name="functions"/> collection to this plugin.</summary>
+        /// <param name="functions">The functions to add.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="functions"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">A function in <paramref name="functions"/>'s has a null <see cref="KernelFunction.Name"/>.</exception>
+        /// <exception cref="ArgumentException">A function with the same <see cref="KernelFunction.Name"/> already exists in this plugin.</exception>
+        public void AddFunctions(IEnumerable<KernelFunction> functions)
```
stephentoub, line 87: Consider removing this method, just to simplify

dotnet/samples/KernelSyntaxExamples/Resources/Example73_MutableKernelPlugin.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// <param name="function">The function to add.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="function"/> is null.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="function"/>'s <see cref="KernelFunction.Name"/> is null.</exception>
+        /// <exception cref="ArgumentException">A function with the same <see cref="KernelFunction.Name"/> already exists in this plugin.</exception>
+        public void AddFunction(KernelFunction function)
+        {
+            if (function is null)
+            {
+                throw new ArgumentNullException(nameof(function));
+            }
```
stephentoub, line 78: These samples build with. NET 6, right? This could be ArgumentNullException.ThrowIfNull

dotnet/samples/KernelSyntaxExamples/Resources/Example73_MutableKernelPlugin.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            this._functions = new Dictionary<string, KernelFunction>(StringComparer.OrdinalIgnoreCase);
+            if (functions is not null)
+            {
+                foreach (KernelFunction f in functions)
+                {
+                    if (f is null)
+                    {
+                        throw new ArgumentNullException(nameof(functions));
+                    }
```
stephentoub, line 55: This could also be ArgumentNullException.ThrowIfNull

PR #4089, opened 2023-12-07
.Net Upgrade to Azure SDK Beta 11

### LINKED ISSUE #3972: Upgrade to Azure AI SDK Beta 10
- [ ] Updated to the latest Azure AI SDK
- [ ] Add support to use Tools instead of Functions

RogerBarreto:
# Upgrades to Azure SDK Beta 11

Azure SDK Beta 10 had a regression with the FunctionCalling behavior and we step to 11 as soon that was fixed.

Resolves #3972 

There's still a minor bug happening in their side (Resetting functions on the callback) where we needed to apply a workround.


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -290,10 +290,13 @@ await foreach (Completions completions in response)
 
             string fqn = functionCallResponse.FullyQualifiedName;
 
-            chatOptions.Messages.Add(resultChoice.Message);
-            chatOptions.Messages.Add(new ChatMessage(ChatRole.Function, serializedFunctionResult) { Name = fqn });
+            if (resultChoice.Message.Content is { Length: > 0 })
```
stephentoub, line 293: Why is this being agreed only conditionally? I thought it was always required and without it subsequent calls will fail... is that not our no longer the case?

RogerBarreto: At first I thought it was the problem (adding an empty assistant message to the history) in the Beta 10 bug, and I just added the message when the assistant actually render a message. 

I kept this change in place to see if it was actually impacting the results and running the Example 59, it worked as expected.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -612,28 +633,122 @@ private static CompletionsOptions CreateCompletionsOptions(string text, OpenAIPr
...
+    {
+        if (message.Role == ChatRole.System)
+        {
+            return new ChatRequestSystemMessage(message.Content);
+        }
+
+        if (message.Role == ChatRole.Assistant)
+        {
+            return new ChatRequestAssistantMessage(message.Content);
+        }
```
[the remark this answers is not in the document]
RogerBarreto: This is not part of assistant API, this is the assistant role for a chat message.

markwallace-microsoft (approved):

SergeyMenshykh (approved):

dmytrostruk (approved):
Small comments, feel free to skip.

dmytrostruk (approved):

PR #4091, opened 2023-12-07
.Net: Change IAIServiceSelector.SelectAIService to be a Try method and add non-Try extension

stephentoub:
And add an extension method to provide the required semantics. This allows a consumer to try to select multiple service types without dealing with exceptions. It also puts the responsibility for constructing a common exception and exception message into one place.


dotnet/src/SemanticKernel.Abstractions/Services/IAIServiceSelector.cs
```diff
@@ -9,15 +11,26 @@ namespace Microsoft.SemanticKernel;
...
-    /// <param name="function">Semantic Kernel callable function interface</param>
+    /// <param name="function">The function.</param>
     /// <param name="arguments">The function arguments.</param>
-    /// <returns></returns>
+    /// <param name="service">The selected service, or null if none was selected.</param>
+    /// <param name="serviceSettings">The settings associated with the selected service. This may be null even if a service is selected.</param>
+    /// <returns>true if a matching service was selected; otherwise, false.</returns>
 #pragma warning disable CA1716 // Identifiers should not match keywords
-    (T?, PromptExecutionSettings?) SelectAIService<T>(Kernel kernel, KernelFunction function, KernelArguments arguments) where T : class, IAIService;
+    bool TrySelectAIService<T>(
```
RogerBarreto, line 29: With this change, the introduced `try/catch` block from KernelFunctionFromPrompt` can be removed in favor of this try check.

markwallace-microsoft (approved):

Issue #4033, opened 2023-12-06
.Net: Add Kernel.CreateBuilder()


matthewbolanos:
This item is to allow a user to use the KernelBuilder in DI.

stephentoub:
_If_ we want to do this, I expect it would look like this:
https://github.com/microsoft/semantic-kernel/compare/main...stephentoub:semantic-kernel:createbuilder

PR #4096, opened 2023-12-08
.Net: Replace KernelBuilder with Kernel.CreateBuilder

stephentoub:
@matthewbolanos, @markwallace-microsoft, entirely up to you whether you want to merge or close this; fine with me either way, I'm just opening this to make it easy to decide. Fixes https://github.com/microsoft/semantic-kernel/issues/4033.

This removes `KernelBuilder` and adds a static `CreateBuilder()` such that instead of doing:
```C#
Kernel kernel = new KernelBuilder().Whatever().Build();
```
one would now do:
```C#
Kernel kernel = Kernel.CreateBuilder().Whatever().Build();
```
Everything else should be functionally the same.


RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/KernelBuilder.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>Provides a builder for constructing instances of <see cref="Kernel"/>.</summary>
+internal sealed class KernelBuilder : IKernelBuilder
+{
+    /// <summary>Whether to allow a call to Build.</summary>
+    /// <remarks>As a minor aid to help avoid misuse, we try to prevent Build from being called on instances returned from AddKernel.</remarks>
+    internal bool _allowBuild;
```
dmytrostruk, line 12: nit: Since it's not private and we use it outside, should we rename it from `_allowBuild` to `AllowBuild`? Convert it to property?

dotnet/src/SemanticKernel.Abstractions/Services/KernelServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿
```
dmytrostruk, line 1:
```suggestion
```

markwallace-microsoft (approved):
LGTM @matthewbolanos do you agree?

matthewbolanos (approved):

qihangnet:
What is the purpose of making such destructive changes repeatedly during the RC phase?
Do you have any misconceptions about RC?

strikene:
> > What is the purpose of making such destructive changes repeatedly during the RC phase? Do you have any misconceptions about RC?
> 
> Hi @qihangnet apologies for the additional breaking changes. This one was made after an internal review of the API surface and was deemed important for consistency with other .Net API's. The team is working very hard to make sure we provide the best experience for developers. We are on track to have a stable 1.0 API very soon and end the cycle of breaking changes.

After I upgraded from RC3 to RC4, I couldn't even build the kernel properly
CS7069	Reference to type 'IKernelBuilder' claims it is defined in 'Microsoft.SemanticKernel.Core', but it could not be found 

image

stephentoub:
> > > What is the purpose of making such destructive changes repeatedly during the RC phase? Do you have any misconceptions about RC?
> > 
> > 
> > Hi @qihangnet apologies for the additional breaking changes. This one was made after an internal review of the API surface and was deemed important for consistency with other .Net API's. The team is working very hard to make sure we provide the best experience for developers. We are on track to have a stable 1.0 API very soon and end the cycle of breaking changes.
> 
> After I upgraded from RC3 to RC4, I couldn't even build the kernel properly CS7069 Reference to type 'IKernelBuilder' claims it is defined in 'Microsoft.SemanticKernel.Core', but it could not be found
> 
> image

It sounds like something in your solution is still pulling in the RC3 version, such that some assemblies think the interface is defined in the abstractions assembly and some think it's defined in the core assembly.

qihangnet:
@strikene, With the release of RC4, there have been significant changes to the namespace and package names related to Connectors. I suggest confirming if you are using the latest version. In my experience, the use of outdated RC3 packages is often the reason for this situation. PR #4230 provides a comprehensive overview of these updates.

I also shared your concern about the significant changes in naming conventions without sufficient documentation or guidance. It seems that there is a difference in understanding between the Semantic Kernel team and the more conservative changes we expect at this stage of the “Release Candidate” (RC); the extent of these changes feels more like the **Beta** stage rather than the **RC** stage.

So far, almost all the community-related articles, blogs, video, including official Microsoft documents, have become obsolete, making the experiences of all relevant developers outdated. This reminds me of some things from the Windows Phone era. Many in the community are very concerned about these arbitrary changes, even angry, and some have lost patience and turned to LangChain.

Unfortunately, it looks like these renaming and disruptive changes will continue to occur. This could significantly damage SK's efforts to engage with the community.

markwallace-microsoft:
> Unfortunately, it looks like these renaming and disruptive changes will continue to occur. This could significantly damage SK's efforts to engage with the community.

There will be no further disruptions in the API surface once we reach v1.0.0 (in the next few days). Our goal is to provide a stable Semantic Kernel  that our partners and internal teams can rely on.

matthewbolanos:
I want to plus one what Mark has shared. The last round of changes during our RC were made explicitly because of feedback from the community and from the .NET SDK experts at Microsoft. Has we lock on v1.0 next week, all non-experimental APIs **will not change**.

I also want to share that we were able to republish all of the content in learn.Microsoft.com to reflect the latest in RC4. We expect few (if any) breaking changes from now and until V1.0 next week

qihangnet:
Oh,one more thing. Recently, the access level of some practical classes in the Semantic Kernel has been changed from public to internal. This change has affected some developers who used these classes and methods directly before. Due to this change, developers are now forced to extract and duplicate the necessary code from the source code of the Semantic Kernel repo, or create equivalent functionality on their own. If your schedule permits, I request that you evaluate this issue and consider possible solutions before the release of version 1.0. Thank you again.🤭

stephentoub:
>  the access level of some practical classes in the Semantic Kernel has been changed from public to internal. This change has affected some developers who used these classes and methods directly before.

It'd be helpful for future reference to know which types/members you're referring to and how they were being used / the scenario. 

In the incubation of SK, there have been many experiments, many approaches tried, etc., and as is commonly the case with such efforts, that left a lot of things in an inconsistent state, many things that weren't necessarily still useful, etc.  When preparing for a stable 1.0 release, one really needs to remove as much of that as possible, as once you're stable, you can effectively never remove anything, since you're signing up to avoid breaking changes.  Since it's always possible to add later but effectively never possible to remove later, the general philosophy is to get rid of things that aren't deemed valuable, as they can be brought back later if proven otherwise.

Issue #3312, opened 2023-10-26
Python: Refactor to support generic LLM request settings


matthewbolanos:
This issue is to capture any work necessary for the Python SDK to conform to this ADR: https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0008-support-generic-llm-request-settings.md#refactor-to-support-generic-llm-request-settings

If this is already complete, please close this issue.

eavanvalkenburg:
I started working on this.

eavanvalkenburg:
@markwallace-microsoft @dmytrostruk @lemillermicrosoft @stephentoub question to you all on this one, this is implementing the pattern described in decision 0008, one thing I ran into is that for Azure different API versions require/accept different arguments and will throw an error when a "unknown" argument is given, my thinking now is to add the control of that to the specific request settings class (and not for instance the client class) and to do that by adding api_version to the AzureChatRequestSettings class and then having a dict that tells the (tentativally named, have to check if there is a equivalent func in dotnet) `prepare_settings_dict` function which fields to exclude or include, three questions: 1) what do you guys think? 2) is there something like this in dotnet and if so where? and 3) can this be added to de decision?

eavanvalkenburg:
This is merged.

PR #4097, opened 2023-12-08
Python: Implement AI request settings

eavanvalkenburg:
### Motivation and Context

This PR implements the new approach to AI Request settings as described here: https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0008-support-generic-llm-request-settings.md and completes #3312 

### Description

Does this:
- Creates AI Request Settings with service_id (referring to the registered AI services in a kernel) and extension_data, a dict that can hold any number of settings
- has several methods:
  - `from_ai_request_settings` -> constructor to create a new request_settings object based on another, supports creating specific ai_request_settings from generic ones.
  - `update_from_ai_request_settings` -> updating the fields of a request setting from another ones.
  - `prepare_settings_dict` -> method that is used when actually creating the dict that get's passed to the handler of the AI service.
- Created subclasses like OpenAIChatRequestSettings which has all the required fields of a OpenAI Chat completion, same for OpenAIText, AzureChat, GooglePalm, HuggingFace, etc. 
- added `get_request_settings_class` to ai_services to return the class used for the settings for that service.

The goal is that when running, this settings object get's updated and a single operations (called `prepare_settings_dict`) is used  to return a dict that can be passed to the api call directly, thereby reducing the need for custom logic everywhere, this will also negate the need for things like `complete_chat_with_functions_async` and others. So also removed complete_chat_..._async methods from openai and azure openai.

It has also changed the prompt_template_config with the generic type for the AIRequestSettings or subclasses, and some other changes.

### Contribution Checklist


python/semantic_kernel/connectors/ai/ai_request_settings.py
```diff
@@ -0,0 +1,41 @@
+from typing import Any, Dict, Optional
+
+from pydantic import Field
+
+from semantic_kernel.sk_pydantic import SKBaseModel
+
+
+class AIRequestSettings(SKBaseModel):
+    service_id: Optional[str] = Field(None, min_length=1)
```
moonbox3, line 9: Could we add some comments here about what the service_id and extension_data is, how they're used, etc?

python/semantic_kernel/connectors/ai/google_palm/services/gp_text_completion.py
```diff
@@ -86,24 +50,27 @@ async def _send_completion_request(
...
-                stop_sequences=(
-                    request_settings.stop_sequences
-                    if request_settings.stop_sequences is not None
-                    and len(request_settings.stop_sequences) > 0
-                    else None
-                ),
-                candidate_count=request_settings.number_of_responses,
-                top_p=request_settings.top_p,
-            )
+            response = palm.generate_text(**request_settings.prepare_settings_dict())
```
moonbox3, line 53: Love the de-cluttering here. Amazing.

moonbox3:
What an awesome PR. Thank you for doing this. I love the direction we're heading and how much you're simplifying config related items.

moonbox3:
Have we gone through all Kernel examples and updated them with these new settings? Does the example `chat_gpt_api_function_calling` need any update regarding the prompt template config.

python/samples/kernel-syntax-examples/azure_chat_gpt_api.py
```diff
@@ -25,11 +26,14 @@
     **azure_openai_settings_from_dot_env_as_dict(include_api_version=True)
 )
 kernel.add_chat_service("chat-gpt", chat_service)
-
-prompt_config = sk.PromptTemplateConfig.from_completion_parameters(
-    max_tokens=2000, temperature=0.7, top_p=0.8
+req_settings = AIRequestSettings(
```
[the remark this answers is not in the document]
eavanvalkenburg: I need to add some notes on the logic here, but it is also possible to get the request settings class from the service, which is a nice way of doing it when just using one service, the advantage of the whole approach is that you can define the parameters for multiple types of services and they can parsed and matched to the specific settings dynamically during runtime.

python/semantic_kernel/connectors/ai/open_ai/services/open_ai_text_completion_base.py
```diff
@@ -57,15 +65,15 @@ async def complete_stream_async(
...
+            settings {OpenAIRequestSettings} -- The settings to use for the completion request.
             logger {Optional[Logger]} -- The logger instance to use. (Optional)
 
         Returns:
             Union[str, List[str]] -- The completion result(s).
         """
-        response = await self._send_request(
-            prompt=prompt, request_settings=settings, stream=True
-        )
+        settings.prompt = prompt
```
moonbox3, line 74: The incoming settings of type `OpenAIRequestSettings` doesn't look to be of an OpenAITextRequestSettings instance, so it doesn't have a prompt attribute to set.

moonbox3: Even with that update, I see this error:

TypeError("Missing required arguments; Expected either ('messages' and 'model') or ('messages', 'model' and 'stream') arguments to be given")

python/samples/kernel-syntax-examples/azure_chat_gpt_api.py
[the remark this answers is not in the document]
eavanvalkenburg: So I added a bit of text to that sample with the different ways you can setup the request settings and when to choose which.

python/semantic_kernel/connectors/ai/open_ai/request_settings/azure_chat_request_settings.py
```diff
@@ -0,0 +1,79 @@
...
+
+from pydantic import Field
+from pydantic.dataclasses import dataclass
+
+from semantic_kernel.connectors.ai.open_ai.request_settings.open_ai_request_settings import (
+    OpenAIChatRequestSettings,
+)
+from semantic_kernel.sk_pydantic import SKBaseModel
+
+logger = logging.getLogger(__name__)
```
juliomenendez, line 12: If this is not being used in this module, do you need it?

python/semantic_kernel/connectors/ai/open_ai/semantic_functions/open_ai_chat_prompt_template.py
```diff
@@ -43,19 +42,26 @@ def add_message(
...
+                        role=role,
+                        fixed_content=message,
+                        name=name,
+                        function_call=function_call,
+                    )
                 )
-                function_call = None
-        self._messages.append(
+                return
+            self._log.warning(
```
juliomenendez, line 56: This should be `logger.warning`

python/semantic_kernel/connectors/ai/open_ai/semantic_functions/open_ai_chat_prompt_template_with_data_config.py
```diff
@@ -1,106 +1,99 @@
 # Copyright (c) Microsoft. All rights reserved.
```
juliomenendez, line 1: Given that the entire file is commented out, should it just be deleted?

python/tests/unit/ai/google_palm/services/test_palm_chat_completion.py
```diff
@@ -65,17 +64,16 @@ async def test_google_palm_text_completion_complete_chat_async_call_with_paramet
             ai_model_id=ai_model_id,
             api_key=api_key,
         )
-        settings = ChatRequestSettings()
+        settings = GooglePalmChatRequestSettings()
         response = await gp_chat_completion.complete_chat_async(prompt, settings)
         assert isinstance(response.result(), str) and len(response.result()) > 0
-
+        print(mock_gp.chat)
```
juliomenendez, line 70: Should this `print()` statement be here?

juliomenendez (changes requested):
Looking good, just a couple of questions

python/semantic_kernel/connectors/ai/google_palm/services/gp_text_embedding.py
```diff
@@ -12,6 +14,11 @@
 from semantic_kernel.connectors.ai.embeddings.embedding_generator_base import (
     EmbeddingGeneratorBase,
 )
+from semantic_kernel.connectors.ai.google_palm.request_settings.gp_request_settings import (
+    GooglePalmEmbeddingRequestSettings,
+)
+
+logger = logging.getLogger(__name__)
```
juliomenendez, line 21: Do you need this if it's not used in this module?

python/semantic_kernel/connectors/ai/hugging_face/__init__.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) Microsoft. All rights reserved.
 
-from semantic_kernel.connectors.ai.hugging_face.hf_request_settings import (
+from connectors.ai.hugging_face.request_settings.hf_request_settings import (
```
juliomenendez, line 3: All the other imports are using the absolute name with `semantic_kernel.`, shouldn't this follow the same?

python/semantic_kernel/connectors/ai/hugging_face/services/hf_text_completion.py
```diff
@@ -9,7 +9,7 @@
 
 from semantic_kernel.connectors.ai.ai_exception import AIException
 from semantic_kernel.connectors.ai.ai_service_client_base import AIServiceClientBase
-from semantic_kernel.connectors.ai.hugging_face.hf_request_settings import (
```
juliomenendez: All the other imports are using the absolute name with `semantic_kernel.`, shouldn't this follow the same?

juliomenendez (changes requested):

python/semantic_kernel/connectors/ai/hugging_face/__init__.py
eavanvalkenburg: yeah, that is a mistake

juliomenendez (approved):
Looks good!

alexchaomander (approved):

moonbox3 (approved):

PR #4107, opened 2023-12-08
.Net: Trace log rendered prompts in KernelFunctionFromPrompt

stephentoub:


dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -268,8 +268,21 @@ private async Task<(IAIService, string, PromptRenderedEventArgs?)> RenderPromptA
 
         var renderedPrompt = await this._promptTemplate.RenderAsync(kernel, arguments, cancellationToken).ConfigureAwait(false);
 
+        if (this._logger.IsEnabled(LogLevel.Trace))
+        {
+            this._logger.LogTrace("Rendered prompt: {Prompt}", renderedPrompt);
```
dmytrostruk, line 273: In `KernelPromptTemplate.cs` we already have one trace log for rendered prompts:
https://github.com/microsoft/semantic-kernel/blob/37f2760fdd146d49780e5fff859a2b79c39b7661/dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs#L123-L127

It looks like if `_promptTemplate` in line 269 is `KernelPromptTemplate`, we will add trace log for rendered prompt twice. Should we remove it from `KernelPromptTemplate` in this case to avoid duplication?

stephentoub: Hmm, you're right. I'm realizing I'm not seeing that tracing because I'm not passing a logger factory into the construction of the function.

I think we either remove the logging from here, or we replace this PR with one that teaches KernelPromptTemplate how to use the logger from the kernel instead or in addition.

Is it common to use a prompt template outside of a prompt function?

dmytrostruk: > Is it common to use a prompt template outside of a prompt function?

I don't think so, probably in case if somebody wants to use `KernelPromptTemplate` directly. Now I think maybe we can make `KernelPromptTemplate` internal at all, if it's possible. I like the approach to do the logging on prompt function level.

dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
```diff
@@ -112,21 +107,12 @@ private async Task<string> RenderAsync(List<Block> blocks, Kernel kernel, Kernel
                     break;
 
                 default:
-                    const string Error = "Unexpected block type, the block doesn't have a rendering method";
-                    this._logger.LogError(Error);
-                    throw new KernelException(Error);
+                    Debug.Fail($"Unexpected block type {block?.GetType()}, the block doesn't have a rendering method");
+                    break;
```
dmytrostruk, line 111: Is `Debug.Fail` accidental change?
```suggestion
                    const string Error = "Unexpected block type, the block doesn't have a rendering method";
                    this._logger.LogError(Error);
                    throw new KernelException(Error);
```

stephentoub: No, it's on purpose. This isn't reachable if the code is correct, right?

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Core/PromptTemplate/KernelPromptTemplate.cs
dmytrostruk: Yeah, just wondering if we should throw an exception, in case we introduce incorrect code in the future. Looks like `Debug.Fail` causes the debugger to break at the line of code, so I was double-checking if this is something that could impact user experience.

dmytrostruk (approved):

dmytrostruk (approved):

Issue #3874, opened 2023-12-01
.Net: Content in ChatMessage is a string; the latest version of OpenAI APIs treats it as a string or array


matthewbolanos:

RogerBarreto:
Content in ChatMessage is still a string.

![image](https://github.com/microsoft/semantic-kernel/assets/19890735/8b9077d8-5d3b-456b-9c3d-d2f08348a42b)

What became an array was the new property `tool_calls` over the deprecated single `function_call`

matthewbolanos:
When you send a multi-modal payload, it becomes an array:

![Image](https://github.com/microsoft/semantic-kernel/assets/1409190/27870485-4d49-4690-b937-f6a875fb21e5)

Issue #3522, opened 2023-11-16
Revisit design of ChatHistory and ChatMessage


markwallace-microsoft:
Comments from this PR https://github.com/microsoft/semantic-kernel/pull/3517

I also question whether this should instead be an IList<ChatMessage> instead of a List<ChatMessage>, so that we can add extra validation and behavior specific to ChatHistory, e.g. ensuring that Add(null) throws.

You added = string.Empty to the property initializer. If someone passes in a null role / content, what do you want to do? Should they still end up as empty? Throw?

stephentoub:
Today, every time a ChatHistory is submitted to the {Azure} OpenAI service, the entire history is duplicated into the corresponding object model for the Azure SDK.  What would it look like instead if the SK ChatMessage objects wrapped the corresponding Azure SDK ChatMessage types, such that the object model didn't need to be recreated each time?  Presumably that would mean ChatHistory couldn't just be a `List<ChatMessage>`, because that forces a particular storage mechanism, and requires that the base `List<T>` APIs all work. It would likely need to change to be an `IList<T>`.  And what about creating the new ChatMessage instances... if a ChatHistory is created via a call to IChatCompletion.CreateNewChat, should a chat message be created via a ChatHistory.CreateNewMessage or some such new virtual method?  Otherwise, how would it wrap the appropriate instance from the service?

It might be worth prototyping this out.

kcadduk:
I'm hoping this is the correct place, I'll open a separate issue if not, if you're revisiting a the design, it would be helpful if the data models needed like ChatHistory and ChatMessage were in some abstraction layer or abstracted in some other way, potentially implementing some interface.

As a consumer of the package, I'm tightly coupling my application to your models, where-as it would be nice to simply implement some interface and not have that tight coupling.

markwallace-microsoft:
We also need to ensure that our chat history and message abstractions will support GPT Vision https://platform.openai.com/docs/guides/vision i.e. the ability for a message to have multiple different types of chat content

kcadduk:
To be able to run `IChatCompletion.GenerateMessageStreamAsync` i need to pass in the `ChatHistory` which is a list of `ChatMessage`.

[...]

PR #4113, opened 2023-12-08
.Net: Updates for ChatMessageContent to support complex content model

dmytrostruk:
### Motivation and Context

Resolves: https://github.com/microsoft/semantic-kernel/issues/3874
Resolves: https://github.com/microsoft/semantic-kernel/issues/3522
Resolves: #3768 

This PR contains changes to prepare public API of `ChatMessageContent` to support extensible content model in the future. With latest OpenAI API it's possible to have content as `string` or `array` of content objects (scenario for GPT vision):
image

### Description

1. Added `ChatMessageContentItemCollection` to store multiple contents of different types in chat message.
2. Added `ChatMessageContentItemCollection? Items` property to `ChatMessageContent`, made `string? Content` property nullable.
3. Added `ImageContent` class that stores image URI.
4. Added ADR and diagram of chat and text content models.
5. Added `Example74_GPTVision` to demonstrate the functionality with GPT Vision model.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/TextToImage/ImageContent.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.SemanticKernel.TextToImage;
+
+/// <summary>
+/// Represents image content.
+/// </summary>
+public class ImageContent : ContentBase
```
markwallace-microsoft, line 12: This should be sealed. If a developer was a customized `ImageContent` they should extend `ContentBase` directly.

markwallace-microsoft (approved):
One small issue

dotnet/src/SemanticKernel.Abstractions/AI/TextToImage/ImageContent.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// The URI of image.
+    /// </summary>
+    public Uri Uri { get; set; }
```
SergeyMenshykh, line 17: Sooner or later, we will need to support images represented as byte arrays. It would be good to consider whether this will be modeled either through a new `byte[] Bytes {get; set;}` property of the class or as an `ImageByteContent` subclass of `ContentBase`. This exercise will be useful in shaping the SK public API surface now to prevent potential breaking changes in the future.

dotnet/samples/KernelSyntaxExamples/Example49_LogitBias.cs
```diff
@@ -44,14 +44,14 @@ public static async Task RunAsync()
         await MessageOutputAsync(chatHistory);
 
         var replyMessage = await chatCompletionService.GetChatMessageContentAsync(chatHistory, settings);
-        chatHistory.AddAssistantMessage(replyMessage.Content);
+        chatHistory.AddAssistantMessage(replyMessage.Content!);
```
SergeyMenshykh, line 47: Why are we certain (using the `!` operator) that the content is not null here?

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/AI/TextToImage/ImageContent.cs
RogerBarreto: ## Points to consider 

How are we going to distinct referenced contents from non referenced ones?

### Following the abstraction path: 

`StreamingContentBase` has a `abstract byte[] ToByteArray()` method.

We might consider bringing this to a `ByteContent` abstract class or directly to the `ContentBase` to be similar.

### How to handle Referenced/Link/Anchor contents?

Any content could be referenced by a Uri, might worth having a `ILinkContent` that ImageContent/AudioContent, etc could inherit or implement

```csharp
interface ILinkContent {
    Uri Uri { get }
}

public OpenAIReferencedImageContent : ContentBase, IReferencedContent {
    Uri Uri { get; set; }
}
```

dotnet/src/SemanticKernel.Abstractions/AI/TextToImage/ImageContent.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.SemanticKernel.TextToImage;
```
RogerBarreto, line 7: Out of scope of this PR, but I think would be useful to move all the content abstractions to the root namespace.

- ContentBase
- ChatMessageContent
- TextContent
- ImageContent

Streaming also.

```suggestion
namespace Microsoft.SemanticKernel;
```

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -664,6 +665,27 @@ private static ChatRequestMessage GetRequestMessage(ChatRole chatRole, string co
...
+            ImageContent imageContent => new ChatMessageImageContentItem(imageContent.Uri),
+            _ => throw new NotSupportedException("Unsupported content type of chat message item.")
+        };
+    }
+
+    private static ChatRequestUserMessage GetChatRequestUserMessage(ChatMessageContent message, string? functionName)
+    {
+        if (message.Items is { Count: > 0 })
+        {
+            var contentItems = message.Items.Select(GetChatMessageContentItem).ToList();
```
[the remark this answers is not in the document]
dmytrostruk: Yes, thank you. Forgot to remove it after initial version of this implementation.

dotnet/samples/KernelSyntaxExamples/Example49_LogitBias.cs
dmytrostruk: That's a good question. Currently, we operate with `ChatMessageContent` as model for request and response at the same time. In case of request, property `Content` can be used for plain text or property `Items` for multi-modal request. While for response, only `Content` property is used. That's why `Content` property is nullable, because for multi-modal request it's not required.. I think we can either use different classes for chat message request and response scenario or remove `Content` property completely and use only `Items` in request and response. In this case, we will avoid null checks. Although, with `Items`, users will have to do `Items[0]` to get response every time, so we will have to provide additional methods for better usability.

In this specific example, we know that `Content` is not null, because we set a response from AI into this property, and in case of some issue we just throw an exception.

dotnet/src/SemanticKernel.Abstractions/AI/TextToImage/ImageContent.cs
dmytrostruk: > Sooner or later, we will need to support images represented as byte arrays. It would be good to consider whether this will be modeled either through a new `byte[] Bytes {get; set;}` property of the class or as an `ImageByteContent` subclass of `ContentBase`. This exercise will be useful in shaping the SK public API surface now to prevent potential breaking changes in the future.

Both options (new property or new separate class) look like addition, so I'm not sure there will be breaking changes when we want to extend it. In case of new property, we will have to convert `Uri` property to nullable, because it will be either image link or image byte array. 

> Any content could be referenced by a Uri, might worth having a `ILinkContent` that ImageContent/AudioContent, etc could inherit or implement

Hmm, that's interesting idea. At the moment, I can't find a scenario when `ILinkContent` abstraction will be useful for us or end-users, could you share any examples?

dotnet/src/SemanticKernel.Abstractions/AI/TextToImage/ImageContent.cs
dmytrostruk: Thanks, I think it's a good idea. In case we make a decision to do that, I will create a separate issue to avoid big number of changes in this PR.

dotnet/src/SemanticKernel.Abstractions/AI/TextToImage/ImageContent.cs
SergeyMenshykh: > In case of new property, we will have to convert Uri property to nullable.

Making a property nullable can be considered a breaking change, right? Particularly if the existing consumers are not expecting or handling a null value for the property. Should we make it nullable now to be prepared?

dmytrostruk: @SergeyMenshykh Yes, that makes sense. Converted `Uri` parameter to nullable.

SergeyMenshykh (approved):

stephentoub:
What's the relationship now between ChatMessageContentItemCollection and ChatHistory? They're both a collection of chat message contents, right?

dmytrostruk:
> They're both a collection of chat message contents, right?

No, `ChatMessageContentItemCollection` is a collection of `ContentBase` items, like `TextContent`, `ImageContent` or `ChatMessageContent` itself, while `ChatHistory` is a collection of concrete type `ChatMessageContent`. Yes, it looks like `ChatMessageContentItemCollection` can be re-used in the same way as `ChatHistory`, although both classes have different responsibilities. 

In our case, `ChatHistory` is a main container for all chat messages with methods like `AddUserMessage`, `AddSystemMessage` etc, while `ChatMessageContentItemCollection` is a collection of chat message contents, which can be text or image url for example.

stephentoub:
> No, ChatMessageContentItemCollection is a collection of ContentBase items, like TextContent, ImageContent or ChatMessageContent itself, while ChatHistory is a collection of concrete type ChatMessageContent. 

Right, that's my point: ChatHistory _is_ a collection of ContentBase items, just guaranteeing a more specialized ContentBase-derived type. So should that relationship be modeled in the API, so that a ChatHistory can be used anywhere a ChatMessageContentItemCollection is employed?

dmytrostruk:
> So should that relationship be modeled in the API, so that a ChatHistory can be used anywhere a ChatMessageContentItemCollection is employed?

I think the reason why it's confusing is because we use `TextContent` or `ChatMessageContent` models as response from AI and at the same time we use `TextContent` inside `ChatMessageContent` as part of chat message text content. Maybe it's not a problem, but probably that's the reason why hierarchy is not clear. (That's why I've added class diagram to this PR :) ).

From user perspective, I would like to clearly differentiate between `ChatHistory` from `ChatMessageContentItemCollection`. In initial `ChatMessageContentItemCollection` I've added a check that `ChatMessageContent` (as `ContentBase`) can't be added to the collection (unline in `ChatHistory`), because it doesn't make sense to include chat message as content of chat message. But I removed that check, because this restriction doesn't actually provide any benefits, and on top of that, checking `ContentBase` type every time on adding new item to the collection may be costly.

Answering your question, I'm not sure if using ChatHistory anywhere a ChatMessageContentItemCollection is employed will provide any benefits, but maybe I'm missing something. Some use-cases or scenarios will definitely help.

PR #4115, opened 2023-12-08
Python: Uses Python standard logging

### LINKED ISSUE #3711: Python: move to python standard logging approach
SK currently does not use the standard approach to logging compared to python best practices and many other packages, where each file that needs to log creates a logger and uses it in code, instead it is passing around a object or when nothing is passed creating a NullLogger, this should not be the way. 

Preferred is, at the top of every file that needs it:
```python
import logging
log: logging.Logger = logging.getLogger(__name__)
```
Potentially with something like this to control log levels based on env and for specific subpackages (some packages are very chatty): https://github.com/openai/openai-python/blob/9f828cb1ea7c22a6388fcd71b5de1a0dde3a9a9f/src/openai/_utils/_logs.py#L16

juliomenendez:
### Motivation and Context

This replaces passing Logger instances as arguments around with the preferred Python way of having each module define a module-wide `logger` instance at the top of the module. Fixes #3711 

### Description

Removes any `log` argument and argument documentation. Replaces usage of those `log` arguments with the module-wide `logger` instance.

### Contribution Checklist


eavanvalkenburg (changes requested):
Overall this is great stuff, thank you very much for making the effort!

One main thing to add is something like this: https://github.com/openai/openai-python/blob/main/src/openai/_utils/_logs.py to set the logging format, I don't know if we want to control logging level through a env variable, so that can be skipped for now, maybe in one of the samples add a example of how to set the loglevel.

moonbox3:
Thank you for your work on this, @juliomenendez. Let me know if you need any support with this.

python/semantic_kernel/connectors/ai/chat_completion_client_base.py
```diff
@@ -15,7 +14,6 @@ async def complete_chat_async(
         self,
         messages: List["ChatMessage"],
         settings: "ChatRequestSettings",
-        logger: Optional[Logger] = None,
```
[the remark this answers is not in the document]
juliomenendez: Thank you for the suggestion @eavanvalkenburg. I just pushed those changes.

juliomenendez:
Thank you @moonbox3! If you could help review the PR that would be a great help. Sorry is a rather big one :)
@eavanvalkenburg I made those changes, feel free to review it again.
Thanks!

python/semantic_kernel/utils/null_logger.py
eavanvalkenburg, line 1: Let's not delete this yet, people might still import it in their code, but add a warning with the new logger that it will be eliminated in a future version and that the same goes for any log/logger variables

eavanvalkenburg (changes requested):
Added one note on the null logger and I didn't see a util that sets the logger format and level once that I mentioned in my previous review!

python/semantic_kernel/utils/null_logger.py
juliomenendez: Added it back with a warning.

juliomenendez:
> Added one note on the null logger and I didn't see a util that sets the logger format and level once that I mentioned in my previous review!

Yes, missed that. Just added it on my last commit with a very simple sample. Let me know if you think it should have something else @eavanvalkenburg

python/samples/kernel-syntax-examples/setup_logging.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) Microsoft. All rights reserved.
+import logging
+
+import semantic_kernel as sk
+from semantic_kernel.connectors.ai.open_ai import OpenAIChatCompletion
+
+
+def setup_logging():
```
moonbox3, line 8: I understand this is simply an example to show how one could set up their logging; however, do we need to document that if someone wants to configure the logging, they'd need to essentially call a method like this from SK's `__int__.py` so that it does set up the logging properly?

juliomenendez: They wouldn't need to call it from SK's `__init__.py`, right? They can set up the logging from their own scripts before importing SK.

moonbox3 (approved):

python/samples/kernel-syntax-examples/setup_logging.py
moonbox3: I didn't ask my question properly, apologies. I think it would be useful if we configured logging automatically, via the __init__.py, and then users could toggle the verbosity (or off). But we can discuss that later.

juliomenendez: Oh! Makes sense but I agree we can look into that in a follow up PR.

python/semantic_kernel/__init__.py
```diff
@@ -30,7 +29,6 @@
 
 __all__ = [
     "Kernel",
-    "NullLogger",
```
eavanvalkenburg, line 33: let's not delete it here either, if people import direct or use sk.NullLogger.

python/semantic_kernel/utils/null_logger.py
```diff
@@ -35,6 +37,12 @@ class NullLogger(Logger, metaclass=_NullerMeta):
 
     def __init__(self):
         super().__init__(None)
+        logger.warning(
+            (
+                "NullLogger is deprecated and will be removed in a future release,",
+                "same as any 'log' and 'logger' arguments.",
```
eavanvalkenburg, line 43: nit: "the same goes for all 'log' and 'logger' arguments."

python/samples/kernel-syntax-examples/setup_logging.py
eavanvalkenburg, line 8: I would like to see this function, without the last line in a utils.log.py file, and this sample loading and calling it, then setting the default loglevel in code. The format we can be opiniated about while the loglevel is common enough that most python devs know how to set it or use regular structures for dealing with it.

eavanvalkenburg (changes requested):
awesome work, some small changes!

python/semantic_kernel/__init__.py
juliomenendez: Fixed

python/samples/kernel-syntax-examples/setup_logging.py
juliomenendez: Good call, fixed

python/semantic_kernel/utils/null_logger.py
juliomenendez: Fixed

eavanvalkenburg (approved):
This is great! thanks for doing the work here @juliomenendez!

alexchaomander (approved):
LGTM! Thank you for doing this!

PR #4122, opened 2023-12-09
.Net: [OpenApi][DataTypeSupport][Part1] Headers serialization

SergeyMenshykh:
### Motivation and Context  
Currently, OpenAPI functionality does not properly serialize arguments for headers, as described in [OpenAPI serialization](https://swagger.io/docs/specification/serialization/). This is not an issue when all arguments are of the string type, since it is the responsibility of the OpenAPI consumer code to provide properly serialized arguments. However, in a context where arguments can be represented by various types - both primitive and complex - this approach is insufficient. This is because the arguments must be serialized into headers in accordance with the metadata specified by the OpenAPI document. This PR is the first in a series of follow-up PRs to refactor header-building functionality to support proper header serialization methods and gradually prepare the functionality for the support of primitive and complex types.

### Description
- The unnecessary `Headers` collection has been removed from the `RestApiOperation` class, as all the required header metadata is already available through the `Parameters` property.  
- The `RenderHeaders` method has been refactored to work with the header metadata available through the `Parameters` collection. The method has also been simplified by not using the header default value, as the default is set on the service, and the client is allowed not to pass it.

### Contribution Checklist


dotnet/src/Functions/Functions.OpenAPI/Model/RestApiOperation.cs
```diff
@@ -127,43 +119,24 @@ public Uri BuildOperationUrl(IDictionary<string, string> arguments, Uri? serverU
...
+        var headersMetadata = this.Parameters.Where(p => p.Location == RestApiOperationParameterLocation.Header);
+
+        foreach (var headerMetadata in headersMetadata)
         {
-            var headerName = header.Key;
-            var headerValue = header.Value;
+            var headerName = headerMetadata.Name;
 
-            //A try to resolve header value in arguments.
+            // A try to resolve header value in arguments.
```
markwallace-microsoft, line 128: Just `Try to resolve ...`

markwallace-microsoft (approved):

dotnet/src/Functions/Functions.OpenApi/Model/RestApiOperation.cs
```diff
@@ -183,13 +156,13 @@ string ReplaceParameter(Match match)
...
+            // Try to find parameter value in arguments
             if (arguments.TryGetValue(parameterName, out string? value) && value is not null)
             {
                 return value;
             }
 
-            //A try to find default value for the parameter
+            // Try to find default value for the parameter
             var parameterMetadata = this.Parameters.First(p => p.Location == RestApiOperationParameterLocation.Path && p.Name == parameterName);
             if (parameterMetadata?.DefaultValue == null)
```
RogerBarreto, line 167: nit
```suggestion
            if (parameterMetadata?.DefaultValue is null)
```

RogerBarreto (approved):

RogerBarreto (approved):

PR #4124, opened 2023-12-09
.Net: Unnecessary helper method is removed

SergeyMenshykh:
Unnecessary helper method - `private static MethodInfo Method(Delegate method)` is removed from a few unit tests.


stephentoub (approved):
Nice. This has been bugging me, too :)

markwallace-microsoft (approved):

PR #4142, opened 2023-12-10
.Net: Streamline KernelPromptTemplate.AddMissingInputVariables

stephentoub:
Use loops instead of more complicate LINQ queries, avoid intermediate lists and sets, etc.


markwallace-microsoft (approved):
Thanks for updating this, changes make the logic much clearer

SergeyMenshykh (approved):

Issue #4102, opened 2023-12-08
.Net OpenAI Support for Tools / Tool Calling - ClientCore internal implementation


RogerBarreto:
Change internal implementation of `Connector.OpenAI.ClientCore` to use Tools abstraction instead of legacy Function Call.

Issue #4101, opened 2023-12-08
.Net OpenAI Support for Tools / Tool Calling


RogerBarreto:
Inner Issues needed accomplish this requirement.

```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/semantic-kernel/issues/4102
- [ ] .Net OpenAI Support for Tools - Adapts FunctionCallBehavior abstractions
- [ ] .Net OpenAI Support for Tools - Swithc OpenAI Planner to use Tools
- [ ] .Net OpenAI Support for Tools - Switch all samples to use Tools
- [ ] .Net OpenAI Support for Tools - Drop support for legacy function calling
```

Issue #3149, opened 2023-10-11
Epic: Function calling


matthewbolanos:
> [!NOTE]  
> This issue tracks all of the items necessary for ensuring samples and docs are ready for Ignite.

```[tasklist]
### .Net
- [ ] https://github.com/microsoft/semantic-kernel/issues/3017
- [ ] https://github.com/microsoft/semantic-kernel/issues/3126
```

```[tasklist]
### Python
```

```[tasklist]
### Java
```

matthewbolanos:
Marking as closed for .NET, will create new issues for Python and Java

PR #4153, opened 2023-12-11
.Net: Redo OpenAI function calling to be tool calling

stephentoub:
Function calling is being deprecated in favor of tool calling, which is a superset of function calling.  This change revamps both the public APIs and the implementation, to remove the function calling support from the public surface area and replace it with tool calling support. The implementation changes accordingly.  Along the way, I fixed some other bugs or design issues.

Please review not only for implementation, but also for public API impact. Thanks!

Fixes https://github.com/microsoft/semantic-kernel/issues/4101
Fixes https://github.com/microsoft/semantic-kernel/issues/4102
Contributes to https://github.com/microsoft/semantic-kernel/issues/3149


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunction.cs
```diff
@@ -10,62 +10,66 @@
 
 namespace Microsoft.SemanticKernel.Connectors.OpenAI;
 
+// NOTE: Since this space is evolving rapidly, in order to reduce the risk of needing to take breaking
+// changes as OpenAI's APIs evolve, these types are not externally construcible. In the future, once
```
markwallace-microsoft, line 14: Typo `construcible`

markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunction.cs
stephentoub, line 14:
```suggestion
// changes as OpenAI's APIs evolve, these types are not externally constructible. In the future, once
```

RogerBarreto (approved):

SergeyMenshykh (approved):

markwallace-microsoft (approved):

Issue #4147, opened 2023-12-10
.Net: Add extension method for InvokeStreamingAsync


matthewbolanos:
For non-streaming, you can invoke a function using the `pluginName` and `functionName`.

`public Task<FunctionResult> InvokeAsync(string? pluginName, string functionName, KernelArguments? arguments = null, CancellationToken cancellationToken = default(CancellationToken))`

There should be an equivalent for streaming.

PR #4155, opened 2023-12-11
.Net: Add Kernel.InvokeStreamingAsync overloads that take pluginName/functionName

stephentoub:
Fixes https://github.com/microsoft/semantic-kernel/issues/4147


markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -444,7 +444,35 @@ public CultureInfo Culture
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>An <see cref="IAsyncEnumerable{T}"/> for streaming the results of the function's invocation.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="functionName"/> is null.</exception>
+    /// <exception cref="ArgumentException"><paramref name="functionName"/> is composed entirely of whitespace.</exception>
+    /// <exception cref="KernelFunctionCanceledException">The <see cref="KernelFunction"/>'s invocation was canceled.</exception>
+    /// <remarks>
+    /// The function will not be invoked until an enumerator is retrieved from the returned <see cref="IAsyncEnumerable{T}"/>
+    /// and its iteration initiated via an initial call to <see cref="IAsyncEnumerator{T}.MoveNextAsync"/>.
+    /// </remarks>
+    public IAsyncEnumerable<StreamingContentBase> InvokeStreamingAsync(
```
dmytrostruk, line 465: nit: Since `InvokeStreamingAsync` is one of the main entry points in Kernel and users will interact with result type a lot, I'm wondering if it makes sense to rename `StreamingContentBase` to just `StreamingContent`. 

Instead of this:
```csharp
await foreach (StreamingContentBase chunk in kernel.InvokeStreamingAsync(function))
{
   // Process streaming...
}
```

Users will have this:
```csharp
await foreach (StreamingContent chunk in kernel.InvokeStreamingAsync(function))
{
   // Process streaming...
}
```

Another way would be to make it consistent with `Task<FunctionResult> InvokeAsync` and for streaming have `IAsyncEnumerable<FunctionStreamingResult> InvokeStreamingAsync`.

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
RogerBarreto: @dmytrostruk It was `StreamingContent` before and the analogous for non-streaming was `ModelContent`, since `ModelContent` was not a good name `ContentBase` was used, but I agree, when thinking about `StreamingContentBase` we need to also think about `ContentBase`, any change there should reflect in non-streaming as well.

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: > any change there should reflect in non-streaming as well

That's not always the case, because if we look from usability perspective, `StreamingContentBase` class will be used more than `ContentBase` and it will be involved in main scenarios (`kernel.InvokeStreamingAsync`), so the importance of having a good name for `StreamingContentBase` is bigger, comparing to `ContentBase`. While `ContentBase` could be used to create concrete content like `TextContent : ContentBase`, `ImageContent: ContentBase` etc, so maybe in this case `ContentBase` naming makes sense. So, my point is that it's possible to trade-off consistency in naming hierarchy for better usability. 

But ideal scenario would be to have both consistent and usable naming :) In this case, maybe we can consider having `KernelStreamingContent` and `KernelContent` models or something similar.

dmytrostruk: Resolving this comment, as I'm working on naming and namespaces in scope of separate PR.

PR #4156, opened 2023-12-11
.Net Removing special "input" argument

### LINKED ISSUE #3887: .Net Special 'input' argument
Today, SK has a special 'input' argument that is used by a few components/classes:

1. [CodeBlock](https://github.com/microsoft/semantic-kernel/blob/a511a9d9ec51e3ebcbda712430b48db7de84d751/dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/CodeBlock.cs#L159C5-L159C5) saves a value of the first positional parameter in the template to the 'input' argument:

   - Expression with value as a function parameter - {{plugin.f1 'v1'}} 
   - Expression with variable 'p1' having 'v1' value as a function parameter {{plugin.f1 $p1}}
   
   In both cases the 'v1' will be saved as a value for 'input' argument

2. [KernelFunctionFromMethod](https://github.com/microsoft/semantic-kernel/blob/a511a9d9ec51e3ebcbda712430b48db7de84d751/dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs#L359) has a heuristic that maps the value of the 'input' argument to the first parameter of a native method:
   - The value of the 'input' argument is mapped to the 'input' parameter - f1(string input) { Assert.Equal('v1', input); //Success }
   - The value of the 'input' argument is mapped to the any first parameter having no argument or default value - f1(Kernel kernel, string monkey) { Assert.Equal('v1', monkey); //Success }

3. [InvokeAsync(..., string input, ...)](https://github.com/microsoft/semantic-kernel/blob/a511a9d9ec51e3ebcbda712430b48db7de84d751/dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs#L97) adds the value of the input parameter as a value of the 'input' argument to the arguments.

We need to decide whether to keep it or drop it. If we decide to keep it, we should model it properly by adding an Input property to the [KernelArguments](https://github.com/microsoft/semantic-kernel/blob/a511a9d9ec51e3ebcbda712430b48db7de84d751/dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs#L14) class instead of using the [input](https://github.com/microsoft/semantic-kernel/blob/a511a9d9ec51e3ebcbda712430b48db7de84d751/dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs#L19C1-L19C1) constant to add or get the property from the arguments. If we decide to drop it, then we might consider how to do so in a backward-compatible way, ensuring that all existing prompts with expressions like this {{plugin.f1 'v1'}} continue working after the change and the overall solution is not worse than the exiting one for the 'input' argument.

RogerBarreto:
### Motivation and Context

Resolves #3887 

Before:
```csharp
var result = await kernel.InvokeAsync(excuseFunction, new("I missed the F1 final race"));
```

After:
```csharp
var result = await kernel.InvokeAsync(excuseFunction, new() { ["input"] = "I missed the F1 final race" });
```


dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/CodeBlock.cs
```diff
@@ -175,7 +175,7 @@ private KernelArguments EnrichFunctionArguments(KernelArguments arguments)
         {
             object? input = ((ITextRendering)this._tokens[1]).Render(arguments);
             // Keep previous trust information when updating the input
-            arguments[KernelArguments.InputParameterName] = input;
+            arguments["input"] = input;
```
markwallace-microsoft, line 178: Why do we still need this reference to  `input`?

RogerBarreto: Nope, this was resolved when the other PR merged in, will ajust when it conflicts.

stephentoub:
The mentions of "input" should be removed from https://github.com/microsoft/semantic-kernel/blob/ae64b16fdc1291b125858ff429a29341f0dcb75e/dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionAttribute.cs#L33-L43

dmytrostruk (approved):
We should probably update PR description to describe new syntax for better visibility.

Before:
```csharp
var result = await kernel.InvokeAsync(excuseFunction, new("I missed the F1 final race"));
```

After:
```csharp
var result = await kernel.InvokeAsync(excuseFunction, new() { ["input"] = "I missed the F1 final race" });
```

SergeyMenshykh (approved):

PR #4158, opened 2023-12-11
Java: Initial Update of ContextVariables to v1

johnoliver:
### Contribution Checklist


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/ContextVariables.java
```diff
@@ -61,35 +77,34 @@ interface Builder extends SemanticKernelBuilder<ContextVariables> {
          * @param content Entry to place in the "input" slot
          * @return an instantiation of ContextVariables
          */
-        Builder withInput(String content);
+        <T> Builder withInput(ContextVariable<T> content);
```
dsgrieve, line 80: Not sure how the generic type helps here. I think ContextVariable<?> would be fine.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/ContextVariables.java
```diff
@@ -61,35 +77,34 @@ interface Builder extends SemanticKernelBuilder<ContextVariables> {
...
         /**
          * Set variable
          *
-         * @param key variable name
+         * @param key   variable name
          * @param value variable value
          * @return builder for fluent chaining
-         * @apiNote Breaking change: s/String value/Object value/
          */
+        <T> Builder withVariable(String key, ContextVariable<T> value);
```
dsgrieve, line 105: ContextVariable<?> should be good

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/FunctionResult.java
```diff
@@ -50,13 +50,5 @@ public interface FunctionResult {
...
-    String getValue();
-
-    /**
-     * Get typed data from metadata.
-     * @param key The key of the metadata.
-     * @return The metadata, or {@code null} if the key is not present.
-     * @param <T> The type of the function result.
-     */
-    <T> T tryGetMetadataValue(String key);
+    <T> T getValue();
```
dsgrieve, line 53: TryGetMetadataValue is an API in C#. Maybe they've done away with it.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/SKContext.java
```diff
@@ -60,16 +60,16 @@ SKContext build(
...
-     * Appends data to the given key
+     * Sets the given variable
      *
-     * @param key key to set
+     * @param key     if null defaults to the "input" key
      * @param content value to set
      * @return Context for fluent calls
      */
-    SKContext appendToVariable(String key, String content);
+    SKContext setVariable(String key, Object content);
```
dsgrieve, line 72: I think I'd deprecate this one. The question this will raise is, "What's the difference between this and the other?" If we're going to retain this, then the javadoc should say that it is a convenience method for the other. But then, it can only deal with well-known types. Seems problematic.

dsgrieve (approved):

PR #4161, opened 2023-12-11
.Net: Some random minor cleanup

stephentoub:
- Found and fixed stray occurrences of "SK" => "Kernel" (e.g. "ISKFunction" in comments or traced messages.
- Cleaned up a little tracing code inside KernelFunctionFromMethod.
- Consistently handle case of CreateLogger returning null. While it's not supposed to, a custom logger could, and we should handle that gracefully.
- Disable CA1508 (avoid dead conditional code). There are too many false positives.


SergeyMenshykh (approved):

markwallace-microsoft (approved):

lukins-cz (commented):
O)

PR #4167, opened 2023-12-11
.Net ChatHistory Streaming Message Helper API

### LINKED ISSUE #3961: There should be helper methods to make printing streaming results less verbose
image

RogerBarreto:
- Added missing properties to OpenAIStreamingChatMessageContent
- Remove code duplication in Unit Tests
- Resolves #3961


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIStreamingChatMessageContent.cs
```diff
@@ -30,6 +30,16 @@ public sealed class OpenAIStreamingChatMessageContent : StreamingChatMessageCont
...
 
+    /// <summary>
+    /// The ID of the tool call update.
+    /// </summary>
+    public string ToolCallUpdateId { get; set; }
+
+    /// <summary>
+    /// The reason why the completion finished.
+    /// </summary>
+    public CompletionsFinishReason? FinishReason { get; set; }
```
stephentoub, line 41: This is going to conflict (logically) with https://github.com/microsoft/semantic-kernel/pull/4153. Can you handle any merging / updating that's necessary to whichever PR merges second?

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -124,6 +124,36 @@ public void AddRange(IEnumerable<ChatMessageContent> items)
...
+    /// <param name="streamingMessageContents"><see cref="IAsyncEnumerator{T}"/> list of streaming message contents</param>
+    /// <returns>Returns the original streaming results with some message processing</returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    public async IAsyncEnumerable<T> AddStreamingMessageAsync<T>(IAsyncEnumerable<T> streamingMessageContents) where T : StreamingChatMessageContent
+    {
+        List<T> messageContents = new();
+        AuthorRole? role = null;
+        StringBuilder messageContent = new();
+
+        await foreach (var chatMessage in streamingMessageContents)
```
stephentoub, line 140:
```suggestion
        await foreach (var chatMessage in streamingMessageContents.ConfigureAwait(false))
```

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -124,6 +124,36 @@ public void AddRange(IEnumerable<ChatMessageContent> items)
...
+            messageContent.Append(chatMessage.Content);
+            messageContents.Add(chatMessage);
+
+            // Is aways expected to have at least one chunk with the role provided from a streaming message
+            role ??= chatMessage.Role;
+
+            yield return chatMessage;
+        }
+
+        if (messageContents.Count != 0)
```
stephentoub, line 151: How likely is this?  If it's expected to not be super rare, we should consider lazily-initializing messageContents and messageContent so we don't pay for those allocations if there are no messages.  If it's expected to be very rare (which I assume is the case), then just proactively allocating those is fine.

dotnet/src/SemanticKernel.UnitTests/Utilities/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static async IAsyncEnumerable<T> ConvertToAsyncEnumerable<T>(this IEnumerable<T> source)
+    {
+        foreach (var item in source)
+        {
+            yield return item;
+        }
+    }
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+#pragma warning restore IDE1006 // Naming rule violation: Missing suffix: 'Async'
+}
```
stephentoub: This already exists here:
https://github.com/microsoft/semantic-kernel/blob/1f6018dd21a2e64749a803a97d5184318ab21101/dotnet/src/InternalUtilities/src/Linq/AsyncEnumerable.cs#L33-L43

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
RogerBarreto: I think is super rare getting streaming without contents, I'm adding this as a safety check.

dotnet/src/SemanticKernel.UnitTests/Utilities/IEnumerableExtensions.cs
RogerBarreto: The ToAsyncEnumerable in the unittest project is clashing with the referenced ones. I'm going to change the internal utilities to avoid that.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -124,6 +125,36 @@ public void AddRange(IEnumerable<ChatMessageContent> items)
         this._messages.AddRange(items);
     }
 
+    /// <summary>
+    /// Add a message to the chat history at the end of the streamed message
+    /// </summary>
+    /// <typeparam name="T">Specific chat message content type</typeparam>
+    /// <param name="streamingMessageContents"><see cref="IAsyncEnumerator{T}"/> list of streaming message contents</param>
+    /// <returns>Returns the original streaming results with some message processing</returns>
+    /// <exception cref="InvalidOperationException"></exception>
```
dmytrostruk, line 134: It doesn't look like this method throws `InvalidOperationException` but if it's happening somewhere down the stack, it's probably worth to mention it in what cases this exception is thrown.

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
RogerBarreto: Leftover, it has, but I changed and forgot to remove the description, thanks.

dmytrostruk (approved):

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -124,6 +125,35 @@ public void AddRange(IEnumerable<ChatMessageContent> items)
...
+
+            // Is always expected to have at least one chunk with the role provided from a streaming message
+            role ??= chatMessage.Role;
+
+            yield return chatMessage;
+        }
+
+        if (messageContents.Count != 0)
+        {
+            this.AddMessage(new ChatMessageContent(role ?? AuthorRole.Assistant, messageContent.ToString(), messageContents[0].ModelId, messageContents, messageContents[0].Encoding, messageContents[0].Metadata));
```
stephentoub, line 153: What's in a chunk's metadata? Do we need to somehow combine them? For example, is usage information included in any that we might want to sum?

RogerBarreto: > Is usage information included in any

No, streaming from OpenAI doesn't provide Usage data in the metadata, only, Id, Created and other fields.

Currently the OpenAI `ClientCore` and `WithData` implementations uses the same metadata for all the chunks, this implementation considers this knowledge. I can add a description to make it clear in the code, 

Any fancy implementation to merge multiple dictionaries in here to support other clients that may or may not add different metadata per chunk is something we want? @markwallace-microsoft 
I didn't add this now to not consume extra resources (Less is more).

RogerBarreto: > we might want to sum?

The `Sum` part here can be unpredictable 😊, as each client could provide the same metadata in all chunks, as the example above, or only in the first, and the Sum logic would have to be in theory held by the client.

From a better perspective what can be done here is each `Connector` adds an extension method that handles its special cases, but the method name would have to be different or enforce the `sk developer` to provide an `OpenAIStreamingChatMessageContent`

OpenAI Connector Extension example:
```
public static async IAsyncEnumerable<OpenAIStreamingChatMessageContent> AddMessageContent(this ChatHistory chatHistory, IAsyncEnumerable<OpenAIStreamingChatMessageContent> streaming) {
   // Specific Connector logic
}
```

RogerBarreto: @stephentoub @markwallace-microsoft Implementation moved to the connector, appreaciate a quick review.

stephentoub:
I don't understand the change that moves this into the connector. Now that the extension is specific to `IAsyncEnumerable<OpenAIStreamingChatMessageContent>`, doesn't that mean you only get to use the simplified helper if you're hardcoding to a specific connector / you know exactly the concrete type of the `IChatCompletion` you're working with?  Doesn't that largely defeat the purpose of the abstraction?

RogerBarreto:
@stephentoub The abstraction can't support all the possible scenarios of chat streaming, adding an abstraction that is incomplete or bound to fail for the next connector where the API behaves differently with the chunks will fail.

Today the "safest" (fail proof) approach is the Connector providing an extension for ChatHistory on how build its well know chat message specialized content.

This change introduces the non-generic thinking in other connectors that could reuse the same method name with their type in the signature

stephentoub:
> @stephentoub The abstraction can't support all the possible scenarios of chat streaming, adding an abstraction that is incomplete or bound to fail for the next connector where the API behaves differently with the chunks will fail.

If we can't write such a helper on top of the abstraction, how could anyone else reasonably consume the abstraction?

stephentoub:
What's the core problem that makes the general-purpose version not viable?

RogerBarreto:
Core problem, the metadata in open aí is the same for all chunks, so the abstraction logic will break if the metada is provided differently. 
Same if the Role is a combination of multiple chunks, theres no rule of tumb to sum or consider first hasvalue chunk for all scenarios

stephentoub:
Thanks. If we can't reason about the metadata abstractly, presumably no one else will be able to, either? Any time we're forced to do something connector-specific because the abstraction doesn't work is a red flag about the abstraction.

If there's no good / genetic way to reason about the metadata via the generic helper, maybe just don't include it in the aggregated contents, and document that limitation? We could _also_ have the connector provide one on the specific T if it could do better. And/or, we could augment the interface with a metadata aggregator function. Or via the interface say that the connector's implementation needs to make it so every update's metadata needs to be an aggregation of everything from previous updates (I don't know if that's viable given the kinds of things we expect to see in metadata). Or have a special aggregated metadata content update a connector can add at the end of the stream optionally, in order to support such scenarios. Etc.

I think there are a variety of options. But I would shy away from saying this functionality is important for the getting started scenario but then making it connector-specific.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ChatHistoryExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+using Azure.AI.OpenAI;
+using Microsoft.SemanticKernel.ChatCompletion;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Chat history extensions.
+/// </summary>
+public static class ChatHistoryExtensions
```
markwallace-microsoft, line 15: This should be experimental, right?

markwallace-microsoft (approved):

stephentoub:
This isn't core and doesn't provide anything folks can't do on their own. Since we know we're going to go in a different direction, why shio it at all?

Issue #4174, opened 2023-12-11
.Net: When using Dependency Injection (DI), I'm seeing planning/processing of native functions try to include function parameters that should have been resolved by DI.


jordanbean-msft:
**Describe the bug**
When using Dependency Injection (DI), I'm seeing planning/processing of native functions try to include function parameters that should have been resolved by DI.

**To Reproduce**
```csharp
 [KernelFunction, Description("...")]
 [return: Description("...")]
 public async Task<string> HistoricalWeatherLookupAsync(
     [Description("...")] double latitude,
     [Description("...")] double longitude,
     [Description("...")] int monthOfYear,
     [FromServices]DaprClient daprClient,
     ILogger logger,
     CancellationToken cancellationToken)
 {
```

Here is a print out of part of the plan.

```json
{
    "name": "HistoricalWeatherLookupPlugin_HistoricalWeatherLookup",
    "description": "...",
    "parameters": {
      "type": "object",
      "required": ["latitude", "longitude", "monthOfYear", "daprClient"],
      "properties": {
        ...
        "monthOfYear": {
          "type": "integer",
          "description": "..."
        },
        "daprClient": {
          "type": "object",
          "properties": {
            "JsonSerializerOptions": {
              "type": "object",
              "properties": {
                "Converters": { "$ref": "#/$defs/iListOfJsonConverter" },
                ...
```

This results in a huge number of extra tokens being used to try and deal with a parameter that OpenAI is not meant to resolve itself.

**Expected behavior**
In this example, I would have expected the `DaprClient` parameter to be resolved via DI (the `FromServices` attribute is how you indicate in a MVC controller that a parameter should be resolved via DI instead of passed in by the caller). The `ILogger` parameter does get resolved correctly. 

**Platform**
 - OS: Windows
 - IDE: Visual Studio
 - Language: C#
 - Source: rc3

**Additional context**
Of course, I can move this parameter to the constructor and it does get resolved via DI.

```csharp
public class HistoricalWeatherLookupPlugin(DaprClient daprClient)
{
```

Is the function parameter DI pattern supported or only via the constructor?

stephentoub:
> Is the function parameter DI pattern supported or only via the constructor?

[KernelFunction]s do not currently satisfy arguments from the IServiceProvider.  It's something that could be added fairly simply, with a bit of additional code right here:
[...]

PR #4185, opened 2023-12-11
.Net: Add support for DI service arguments to KernelFunctionFromMethod

stephentoub:
With this, you can add [FromKernelServices] to a parameter, e.g.
```C#
kernel.CreateFunctionFromMethod(([FromKernelServices] IExample example) =>
{
    ...
});
```
and `example` will be satisfied not from the `KernelArguments` but instead from `Kernel.Services`.

As part of this, I removed the special checking that was being done to ensure special types weren't used twice in the method signature. I'd originally added that as a precaution, but if someone wants to do it, fine, and it's not worth the extra complication, especially since we should ideally have similar behavior with this new FromKernelServices support, and we don't want to track each of these additional services explicitly (an arbitrarily large number of them).

Also, now that `GetRequiredService` maps down to Kernel.Services, I changed its `string? serviceId` to `object? serviceKey`, to align with what's underneath.

Fixes https://github.com/microsoft/semantic-kernel/issues/4174
cc: @SergeyMenshykh, @markwallace-microsoft, @matthewbolanos


dmytrostruk (approved):

dotnet/src/SemanticKernel.UnitTests/Functions/KernelFunctionFromMethodTests2.cs
```diff
@@ -172,19 +147,68 @@ public async Task ItFlowsSpecialArgumentsIntoFunctionsAsync()
...
+
+        Assert.True(invoked);
+
+        Assert.DoesNotContain(func.Metadata.Parameters, p => p.Name.Contains("service", StringComparison.Ordinal));
+    }
+
+    [Fact]
+    public async Task ItThrowsForMissingServicesWithoutDefaultsAsync()
+    {
+        var serviceA = new ExampleService();
```
RogerBarreto, line 189: nit. ServicA, B C aren't used in the test. Could be removed.

RogerBarreto (approved):

SergeyMenshykh (approved):

PR #4186, opened 2023-12-11
.Net: Fix InvokePromptAsync doesn't work for non-semantic-kernel templates

### LINKED ISSUE #4160: `InvokePromptAsync` doesn't work for non-semantic-kernel templates
The following code throws the error "Prompt template format semantic-kernel is not supported"

There should be a way to state that the template is a different type (i.e., "handlebars")

```csharp
var handleBarsResult = await kernel.InvokePromptAsync(
    prompt,
    new() {
        { "request", request },
        { "choices", new[] { "SendEmail", "SendMessage", "CompleteTask", "CreateDocument", "Unknown" } },
        { "history", history },
        { "fewShotExamples", fewShotExamples }
    },
    new HandlebarsPromptTemplateFactory()
);
```

markwallace-microsoft:
### Motivation and Context

Closes #4160 

### Contribution Checklist


RogerBarreto (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -41,13 +42,20 @@ internal sealed class KernelFunctionFromPrompt : KernelFunction
...
         string? description = null,
+        string? templateFormat = null,
         IPromptTemplateFactory? promptTemplateFactory = null,
         ILoggerFactory? loggerFactory = null)
     {
         Verify.NotNullOrWhiteSpace(promptTemplate);
 
+        if (promptTemplateFactory is not null)
+        {
+            Verify.NotNullOrWhiteSpace(templateFormat);
```
dmytrostruk, line 53: It may be confusing that we are throwing `ArgumentNullException/ArgumentException` for optional parameter, and explanation in XML docs may not be enough. 

One way how to improve it is to provide detailed error message: `Template format is required when prompt template factory is provided`. 

Another way would be to create separate model, where `TemplateFormat` and `PromptTemplateFactory` properties are non-nullable, but actual model is optional in parameters. So, when users want to provide `PromptTemplateFactory`, they will be forced to provide `TemplateFormat` as well (and vice versa, if this scenario is applicable). This solution should reduce the number of error scenarios.

```csharp
public sealed class PromptTemplateParameterers 
{
    public string TemplateFormat { get; set; }
    
    public IPromptTemplateFactory PromptTemplateFactory { get; set; }

    // Constructor...
}
```

```csharp
public static KernelFunction Create(
        string promptTemplate,
        string? functionName = null,
        string? description = null,
        PromptTemplateParameterers promptTemplateParams = null,
{
    // Implementation
}
```

dmytrostruk (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
markwallace-microsoft: I've added a better exception message

dmytrostruk (approved):

stephentoub:
Should we consider alternatives, like having PromptTemplateConfig's template format default to null rather than "semantic-kernel", such that the factory can then decide what to do if no format was specified? It could try to process without the format and only fail if it actually wasn't compatible.

stephentoub:
It's otherwise counterintuitive to me that I have to effectively tell it twice what format to use, once in the config and once by providing the factory; I'd expect providing a factory is sufficient unless disambiguation is needed, which the template format would provide (I could also proactively supply a format to help an incompatible factory avoid work).

PR #4192, opened 2023-12-12
.Net: [OpenApi][DataTypeSupport][Part3] support for string and non-string Types

SergeyMenshykh:
Closes - https://github.com/microsoft/semantic-kernel/issues/3692

### Motivation and Context
Currently, OpenAPI functionality can only work with arguments of the string type. This PR updates OpenAPI functionality to support other .NET primitive types such as int, double, bool, etc., in addition to string types

### Description
- The `OpenApiTypeConverter` class is introduced to perform .NET primitive to `JsonNode` conversions, taking into account the OpenAPI target type.  
- The OpenAPI serialization classes are modified to work with JsonNode arguments instead of string ones.  
- The payload building functionality is refactored to use the new `OpenApiTypeConverter` converter.  
- The path serialization logic is updated to align with the other serialization processes for headers and query string parameters.

### Contribution Checklist


dotnet/src/IntegrationTests/Plugins/PluginTests.cs
```diff
@@ -196,4 +196,35 @@ public class PluginTests
...
+            var arguments = new KernelArguments();
+            arguments["title"] = "Shopping List";
+            arguments["ingredients"] = new string[] { "Flour", "Sugar", "Eggs" };
+            arguments["instructions"] = new string[] { "Cream softened butter and granulated sugar", "Add eggs one at a time, mix well, and stir in vanilla extract", "Combine dry ingredients and mix" };
+            arguments["question"] = "what ingredients do I need to make chocolate cookies?";
+            arguments["partner_name"] = "OpenAI";
+
+            // Act
+            await plugin[functionName].InvokeAsync(kernel, arguments);
+        }
```
dmytrostruk, line 228: Do we need some assertion in this test?

dotnet/src/Functions/Functions.OpenApi/Serialization/FormStyleParameterSerializer.cs
```diff
@@ -44,18 +45,18 @@ public static string Serialize(RestApiOperationParameter parameter, string argum
...
     /// <param name="argument">The argument value.</param>
     /// <returns>The serialized parameter string.</returns>
-    private static string SerializeArrayParameter(RestApiOperationParameter parameter, string argument)
+    private static string SerializeArrayParameter(RestApiOperationParameter parameter, JsonNode argument)
     {
-        if (JsonNode.Parse(argument) is not JsonArray array)
+        if (argument is not JsonArray array)
         {
-            throw new KernelException($"Can't deserialize parameter name `{parameter.Name}` argument `{argument}` to JSON array");
+            throw new ArgumentOutOfRangeException(parameter.Name, argument, $"Unexpected argument type '{argument.GetType()}'.");
```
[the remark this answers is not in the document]
SergeyMenshykh: Fixed. I believe '_each check of type and exception depends on specific scenario.'_ is better approach than aligning them all.

dotnet/src/IntegrationTests/Plugins/PluginTests.cs
SergeyMenshykh: I don't think so, unless you have a case where there's a failure, but the test does not fail. I run them quite often, and they always show issues if there are any.

dotnet/src/Functions/Functions.OpenApi/Serialization/OpenApiTypeConverter.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Globalization;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+
+namespace Microsoft.SemanticKernel.Plugins.OpenApi.Serialization;
```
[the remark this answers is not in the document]
SergeyMenshykh: Fixed.

dmytrostruk (approved):

dotnet/src/Functions/Functions.OpenApi/Serialization/OpenApiTypeConverter.cs
```diff
@@ -0,0 +1,157 @@
...
+    /// <returns>A JsonNode representing the converted value.</returns>
+    public static JsonNode Convert(string name, string type, object argument)
+    {
+        Verify.NotNull(argument);
+
+        try
+        {
+            if (s_converters.TryGetValue(type, out var converter))
+            {
+                var result = converter(name, type, argument);
```
dmytrostruk, line 31: It doesn't look like `name` and `type` are used in any of the converters. Should we remove them from here?

dmytrostruk (approved):

dotnet/src/Functions/Functions.OpenApi/Serialization/OpenApiTypeConverter.cs
SergeyMenshykh: Right, they are not used there. They were originally used, but then I refactored the code to return null instead of throwing an exception and forgot to remove them. Thanks, removed.

dmytrostruk (approved):

dmytrostruk (approved):

markwallace-microsoft (approved):

PR #4211, opened 2023-12-12
Service ID in execution_settings should be promoted as the key of the list/dictionary

### LINKED ISSUE #3981: .Net: Service ID in execution_settings should be promoted as the key of the list/dictionary
Need to adjust execution settings to match Azure prompt syntax.

tl;dr, the service ID needs to be moved up to be the key of a dictionary. This likely needs to use an ordered dictionary to not break the existing AI selection service logic.

"Default" can just be a service ID unless we want to make it "special" (but I suspect that can be a non-breaking change in the future)

### Example
https://github.com/Azure/azureml-assets/blob/63a83f07443a2bcb9dcc7f711d457a3dd1846c6e/assets/large_language_models/prompts/completions_chatbot/prompt/prompt.yaml#L12

markwallace-microsoft:
### Motivation and Context

Resolves #3981 
### Contribution Checklist

### DIFF
--- a/dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
+++ b/dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
@@ -436,12 +436,12 @@ public sealed class OpenAICompletionTests : IDisposable
         var azurePromptModel = PromptTemplateConfig.FromJson(
             @"{
                 ""name"": ""FishMarket2"",
-                ""execution_settings"": [
-                    {
+                ""execution_settings"": {
+                    ""default"": {
                         ""max_tokens"": 256,
                         ""service_id"": ""azure-text-davinci-003""
                     }
-                ]
+                }
             }");
         azurePromptModel.Template = prompt;
 
--- a/dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
@@ -21,8 +21,9 @@ public sealed class PromptTemplateConfig
 
     /// <summary>Lazily-initialized input variables.</summary>
     private List<InputVariable>? _inputVariables;
-    /// <summary>Lazily-initialized execution settings.</summary>
-    private List<PromptExecutionSettings>? _executionSettings;
+
+    /// <summary>Lazily-initialized execution settings. The key is the service id or "default" for the default execution settings.</summary>
+    private Dictionary<string, PromptExecutionSettings>? _executionSettings;
 
     /// <summary>
     /// Name of the kernel function.
@@ -72,7 +73,7 @@ public sealed class PromptTemplateConfig
     /// Prompt execution settings.
     /// </summary>
     [JsonPropertyName("execution_settings")]
-    public List<PromptExecutionSettings> ExecutionSettings
+    public Dictionary<string, PromptExecutionSettings> ExecutionSettings
     {
         get => this._executionSettings ??= new();
         set
@@ -81,6 +82,11 @@ public sealed class PromptTemplateConfig
             this._executionSettings = value;
         }
     }
+
+    /// <summary>
+    /// Default execution settings.
+    /// </summary>
+    public PromptExecutionSettings? DefaultExecutionSettings => this._executionSettings is not null && this._executionSettings.TryGetValue(PromptExecutionSettings.DefaultServiceId, out PromptExecutionSettings? settings) ? settings : null;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="PromptTemplateConfig"/> class.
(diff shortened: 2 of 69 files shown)


dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
```diff
@@ -97,6 +103,19 @@ public PromptTemplateConfig(string template)
...
+    /// Adds the <see cref="PromptExecutionSettings"/> to the <see cref="ExecutionSettings"/> dictionary.
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="PromptExecutionSettings.ServiceId"/> is used as the key if provided. Otherwise, the key is "default".
+    /// </remarks>
+    /// <param name="settings"></param>
+    public void AddExecutionSettings(PromptExecutionSettings settings)
+    {
+        Verify.NotNull(settings);
+        this.ExecutionSettings[settings.ServiceId ?? PromptExecutionSettings.DefaultServiceId] = settings;
```
dmytrostruk, line 116: Do we need to allow to overwrite existing settings in this method? If yes, in this case maybe we should name the method as `SetExecutionSettings`.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -66,7 +66,7 @@ internal sealed class KernelFunctionFromPrompt : KernelFunction
 
         if (executionSettings is not null)
         {
-            promptConfig.ExecutionSettings.Add(executionSettings);
+            promptConfig.ExecutionSettings.Add(executionSettings.ServiceId ?? PromptExecutionSettings.DefaultServiceId, executionSettings);
```
dmytrostruk, line 69: There are a couple of places with this logic `executionSettings.ServiceId ?? PromptExecutionSettings.DefaultServiceId` and there is a chance to miss it somewhere. Should we convert `ServiceId` to non-nullable and set it to `default` on `PromptExecutionSettings` creation? In this case we will be sure that `ServiceId` is always configured, and it will be possible for user to set different value if needed.

```csharp
public class PromptExecutionSettings
{
    public const string DefaultServiceId = "default";

    [JsonPropertyName("service_id")]
    public string ServiceId { get; set; } = DefaultServiceId;

    // Other properties
```

dotnet/src/SemanticKernel.UnitTests/Functions/MultipleModelTests.cs
```diff
@@ -158,8 +156,8 @@ public async Task ItUsesServiceIdWithJsonPromptTemplateConfigAsync()
...
-        mockTextGeneration2.Verify(a => a.GetTextContentsAsync("template", It.Is<PromptExecutionSettings>(settings => settings.ServiceId == "service2"), It.IsAny<Kernel>(), It.IsAny<CancellationToken>()), Times.Once());
-        mockTextGeneration3.Verify(a => a.GetTextContentsAsync("template", It.Is<PromptExecutionSettings>(settings => settings.ServiceId == "service3"), It.IsAny<Kernel>(), It.IsAny<CancellationToken>()), Times.Never());
+        mockTextGeneration1.Verify(a => a.GetTextContentsAsync("template", It.IsAny<PromptExecutionSettings>(), It.IsAny<Kernel>(), It.IsAny<CancellationToken>()), Times.Never());
+        mockTextGeneration2.Verify(a => a.GetTextContentsAsync("template", It.IsAny<PromptExecutionSettings>(), It.IsAny<Kernel>(), It.IsAny<CancellationToken>()), Times.Once());
+        mockTextGeneration3.Verify(a => a.GetTextContentsAsync("template", It.IsAny<PromptExecutionSettings>(), It.IsAny<Kernel>(), It.IsAny<CancellationToken>()), Times.Never());
```
dmytrostruk, line 161: It looks like `ServiceId` property is no longer accessible from Connectors. Leaving this comment just to double-check if this is desired behavior. (e.g. Maybe `ServiceId` was used for telemetry).

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
```diff
@@ -437,12 +437,12 @@ public async Task MultipleServiceLoadPromptConfigTestAsync()
         var azurePromptModel = PromptTemplateConfig.FromJson(
             @"{
                 ""name"": ""FishMarket2"",
-                ""execution_settings"": [
-                    {
+                ""execution_settings"": {
+                    ""default"": {
                         ""max_tokens"": 256,
                         ""service_id"": ""azure-text-davinci-003""
                     }
```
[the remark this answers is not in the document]
markwallace-microsoft: I'm going to remove service_id in a follow up PR. Not including it now to keep the PR simple.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
markwallace-microsoft: `ServiceId` will be eliminated and doing that now results in a bunch of unit test changes that will need to be reverted.

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
markwallace-microsoft: Add code to throw an exception if a duplicate is detected

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
```diff
@@ -97,6 +103,28 @@ public PromptTemplateConfig(string template)
...
+
+        var key = settings.ServiceId ?? PromptExecutionSettings.DefaultServiceId;
+        /*
+        if (this.ExecutionSettings.ContainsKey(key))
+        {
+            throw new ArgumentException($"Execution settings for service id '{key}' already exists.");
+        }
+        */
+
+        this.ExecutionSettings[key] = settings;
```
SergeyMenshykh, line 125: Please revise the method.

SergeyMenshykh (approved):

dmytrostruk (approved):

PR #4220, opened 2023-12-13
.Net: PromptExecutionSettings in KernelArguments should be an ordered dictionary

### LINKED ISSUE #4212: .Net: PromptExecutionSettings in KernelArguments should be an ordered dictionary
tl;dr, it's not possible to pass multiple execution settings (one for each model ID, service ID, or another key) into an `IAIServiceSelector`

I have the following prompt

```csharp
var chat = kernel.CreateFunctionFromPrompt(
    new PromptTemplateConfig()
    {
        Name = "Chat",
        Description = "Chat with the assistant.",
        Template = @"{{ConversationSummaryPlugin.SummarizeConversation $history}}
        User: {{$request}}
        Assistant: ",
        TemplateFormat = "semantic-kernel",
        InputVariables = [
            new() { Name = "history", Description = "The history of the conversation.", IsRequired = false, Default = "" },
            new() { Name = "request", Description = "The user's request.", IsRequired = true }
        ],
        ExecutionSettings = [
            new() {
                // Default settings for all models
                ExtensionData = new() {
                    { "MaxTokens", 1000 },
                    { "Temperature", 0 }
                }
            },
            new() {
                // Settings for gpt-3.5-turbo
                ModelId = "gpt-3.5-turbo",
                ExtensionData = new() {
                    { "MaxTokens", 4000 },
                    { "Temperature", 0.1 }
                }
            },
            new() {
                // Settings for gpt-4
                ModelId = "gpt-4-1106-preview",
                ExtensionData = new() {
                    { "MaxTokens", 8000 },
                    { "Temperature", 0.3 }
                }
            }
        ]
    }
);
```

Notice that I have settings to support 3 different models. During invocation time, however, I want to override these settings using `KernelArguments`. Unfortunately, `KernelArguments` only lets me define a single `PromptExecutionSetting` and at this point, I don't know which model is best because that's the responsibility of the `IAIServiceSelector`

```csharp
var result = await kernel.InvokeAsync<string>(chat, new(new PromptExecutionSettings()
{
    ModelId = "gpt-3.5-turbo",
    ExtensionData = new() { { "MaxTokens", 1000 } }
}) { { "input", "Hello world!" } });
```

Once I'm in the `IAIServiceSelector`, I only get a single `PromptExecutionSettings`, so if the service selector chooses a different model than I provided in the `KernelArguments`, then I won't be able to override it corectly

```csharp
public bool TrySelectAIService<T>(
        Kernel kernel, KernelFunction function, KernelArguments arguments,
        [NotNullWhen(true)] out T? service, out PromptExecutionSettings? serviceSettings) where T : class, IAIService
```

markwallace-microsoft:
### Motivation and Context

Closes #4212 

### Contribution Checklist


SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -40,7 +44,7 @@ public KernelArguments(PromptExecutionSettings? executionSettings = null)
     /// If <paramref name="executionSettings"/> is non-null, it is used as the <see cref="ExecutionSettings"/> for this new instance.
     /// Otherwise, if the source is a <see cref="KernelArguments"/>, its <see cref="ExecutionSettings"/> are used.
     /// </remarks>
-    public KernelArguments(IDictionary<string, object?> source, PromptExecutionSettings? executionSettings = null)
+    public KernelArguments(IDictionary<string, object?> source, Dictionary<string, PromptExecutionSettings>? executionSettings = null)
```
dmytrostruk, line 47: Proposal: Maybe we should keep `KernelArguments` constructor that accepts `PromptExecutionSettings? executionSettings = null` for easier usability, if we want to pass just default settings. 

We can add new constructor that accepts `Dictionary<string, PromptExecutionSettings>? executionSettings` for more advanced scenarios.

dmytrostruk (approved):

dotnet/src/SemanticKernel.UnitTests/Functions/KernelArgumentsTests.cs
```diff
@@ -113,6 +113,6 @@ public void ItCanBeInitializedFromAnotherSettingsInstance()
         Assert.True(sut.ContainsName("fake-key"));
         Assert.Equal("fake-value", sut["fake-key"]);
 
-        Assert.Same(executionSettings, sut.ExecutionSettings);
+        Assert.Same(executionSettings, sut.ExecutionSettings?["default"]);
```
RogerBarreto, line 116:
```suggestion
        Assert.Same(executionSettings, sut.ExecutionSettings?[PromptExecutionSettings.DefaultServiceId]);
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
markwallace-microsoft: Yes, we have that on line 28

dotnet/src/SemanticKernel.UnitTests/Functions/KernelArgumentsTests.cs
markwallace-microsoft: Will make this change in the follow up PR

RogerBarreto (approved):

PR #4236, opened 2023-12-13
.Net: Fix default OpenAI execution settings values

### LINKED ISSUE #4235: OpenAIPromptExecutionSettings has incorrect default settings for Temperature and TopP
Both Temperature and TopP are being set to 0 as the default when their default should actually be 1 according to OpenAI docs: https://platform.openai.com/docs/api-reference/chat/create

matthewbolanos:
Correctly sets the default value for Temperature and TopP to 1 for OpenAI models

Closes #4235

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
@@ -19,16 +19,18 @@ public sealed class OpenAIPromptExecutionSettings : PromptExecutionSettings
     /// <summary>
     /// Temperature controls the randomness of the completion.
     /// The higher the temperature, the more random the completion.
+    /// Default is 1.0.
     /// </summary>
     [JsonPropertyName("temperature")]
-    public double Temperature { get; set; }
+    public double Temperature { get; set; } = 1;
 
     /// <summary>
     /// TopP controls the diversity of the completion.
     /// The higher the TopP, the more diverse the completion.
+    /// Default is 1.0.
     /// </summary>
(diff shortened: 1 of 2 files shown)


markwallace-microsoft (approved):

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -19,16 +19,18 @@ public sealed class OpenAIPromptExecutionSettings : PromptExecutionSettings
     /// <summary>
     /// Temperature controls the randomness of the completion.
     /// The higher the temperature, the more random the completion.
+    /// Default is 1.0.
```
RogerBarreto, line 22: Are we sure to put this as full 1.0 for both?

This is full creative mode 😊

> Using a GPT model, setting both Top P (nucleus sampling) and Temperature to 1 can be a good starting point. Top P at 1 means the model will consider a wider range of words for each choice, leading to more diverse responses. Similarly, a Temperature of 1 maintains a balance between randomness and predictability in the responses. You can adjust these settings based on the specific requirements of your application, like more creative versus more predictable responses.

RogerBarreto (approved):

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
RogerBarreto: Those are the defaults from OpenAI, answers the question.

image

PR #4237, opened 2023-12-13
Python: Resolve hanging issue in Jupyter notebooks when re-running async functions

### LINKED ISSUE #4137: Python:  semantic function hangs if I call more than once.
**Describe the bug**
Semantic function hangs if I call it more than once.

**To Reproduce**
From the 00-getting-started.ipynb, run the semantic function cell again, then it hangs.
For example
```
print(joke_function("time travel to dinosaur age"))  # the first call of this function returns output

print(joke_function("any input here...")) # this hangs
```

**Expected behavior**
The function call doesn't hang.

**Platform**
 - OS: Linux
 - IDE: VS Code
 - Language: Python

**Additional context**
chat service: Azure OpenAI, GPT4 model.

moonbox3:
### Motivation and Context

Resolves #4137. The issue was that if you try to re-run the sample joke in the 00-getting-started notebook, the subsequent call to the joke hangs.

### Description

There were issues when re-running asynchronous functions (`_invoke_semantic_async` and `_invoke_native_async`) due to improper handling of asyncio event loops in a multi-threaded env. Specifically, the code had an issue when an event loop was already running, leading to hanging or blocking behavior on subsequent runs.

The change replaced the old code Python's ThreadPoolExecutor, which simplifies thread management. It handles the creation, execution, and cleanup of threads more efficiently and safely.

The new `run_async_in_executor` method ensures that each async function runs in its own new event loop, avoiding conflicts with the main thread's event loop or other threads.

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/orchestration/sk_function.py
+++ b/python/semantic_kernel/orchestration/sk_function.py
@@ -657,7 +657,7 @@ class SKFunction(SKFunctionBase):
 
     def run_async_in_executor(self, coroutine_func: Callable[[], Any]) -> Any:
         """
-        A unified method for async execution for more efficent and safer thread management
+        A unified method for async execution for more efficient and safer thread management
 
         Arguments:
             coroutine_func {Callable[[], Any]} -- The coroutine to run
--- a/python/semantic_kernel/orchestration/sk_function.py
+++ b/python/semantic_kernel/orchestration/sk_function.py
@@ -3,7 +3,7 @@
 import asyncio
 import platform
 import sys
-import threading
+from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
 from logging import Logger
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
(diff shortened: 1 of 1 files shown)


eavanvalkenburg (approved):
Looks good, thanks for fixing this!

alliscode (approved):

moonbox3:
Hi, @BeeShall. Are you referring to [this code?](https://github.com/microsoft/semantic-kernel/blob/main/python/README.md#semantic-functions-are-prompts-with-input-parameters). Can you send us the exact code you're trying to run as a standalone python file? Thanks.

NatanMish:
> I think this fix has caused a new problem. When trying to run the semantic functions with input parameters from this notebook [https://github.com/microsoft/semantic-kernel/blob/main/python/README.md], my event loop hangs now:
> 
> In addition, when trying to run this code as a standalone python file, I get **RuntimeError: There is no current event loop in thread 'MainThread'.**
> 
> I'm trying to find the exact problem and fix associated with this but reporting here so the community can also look into it.

I am seeing the same issue as @BeeShall .
And yes, @moonbox3 , the code is very similar to that example. Let me write the code and I'll send it over

NatanMish:
`import semantic_kernel as sk
from semantic_kernel.connectors.ai.open_ai import AzureChatCompletion


kernel = sk.Kernel()

deployment = "gpt35turbo"
api_key = ""
endpoint = ""

azure_text_service = AzureChatCompletion(deployment_name=deployment, endpoint=endpoint, api_key=api_key)    # set the deployment name to the value of your text model
kernel.add_chat_service("dv", azure_text_service)

prompt = """{{$input}}\n\nEXPLAIN THESE SYSTEM LOGS IN A HUMAN READABLE FORMAT."""
# prompt = "{{$input}}/n/nSummarize this"
log_explainer = kernel.create_semantic_function(prompt_template=prompt, max_tokens=2000, temperature=0.2, top_p=0.5)

for i in range(10):
    print(log_explainer(f"ERROR: {i}"))`

BeeShall:
@moonbox3 Yes that's the one. keeps hanging in Jupyter notebook in a thread lock situation and I get that error when I try to run standalone.

@NatanMish yep that's exactly how I see it too.

moonbox3:
@BeeShall, thanks for confirming. And @NatanMish, thank you for the code. Appreciate you pointing this out. This is fixed in #4485.

Issue #3154, opened 2023-10-11
.Net: Why do AzureTextCompletion and OpenAITextCompletion exist?


stephentoub:
> [!IMPORTANT]  
> Labeled **Urgent** because it may require a breaking change if we should remove OpenAITextCompletion.

As far as I can tell, its functionality is 100% available via AzureChatCompletion.  Why do we have both?

Same question for OpenAITextCompletion and OpenAIChatCompletion

stephentoub:
> Now, since [OpenAI text completion API](https://platform.openai.com/docs/guides/gpt/completions-api) will be deprecated, we will need to do the same for text completion classes

Even if they weren't being deprecated, though, the chat completion classes are pure supersets of the text completion ones: as far as I can tell, they share the exact same implementation for the text completion functionality and then the chat types add more.  So there's zero need for the text completion ones, functionally.

Regardless, I'm glad to hear they're going away.

matthewbolanos:
Per conversation, consider deprecating the AzureTextCompletion and OpenAITextCompletion connectors in favor of just having everyone use the Chat Completion versions.

stephentoub:
> At the very least, we should add the experimental tag to these.

Why not just delete them? Sprcifically what benefit at all do they add? As far as I can tell, there's literally nothing you can do with them you can't do with the other types.

dmytrostruk:
> > At the very least, we should add the experimental tag to these.
> 
> Why not just delete them? Specifically what benefit at all do they add? As far as I can tell, there's literally nothing you can do with them you can't do with the other types.

@stephentoub A couple of weeks ago I tried to remove them and change all integration tests to use chat completion models and it appeared that there is some overload on Azure/OpenAI endpoints when all tests start to use just chat completion endpoint. Since our code changed a lot, I will try to do that again and see if our tests will pass successfully. In the future, we will improve our testing infrastructure and add some load balancing to avoid such issues. The final goal is still to remove text completion services, it's in progress now.

PR #4246, opened 2023-12-13
.Net: Removed Azure and OpenAI text generation services

dmytrostruk:
### Motivation and Context

Resolves: https://github.com/microsoft/semantic-kernel/issues/3154

Removed Azure and OpenAI text generation services in favor of chat completion services.

### Contribution Checklist


RogerBarreto (approved):

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
```diff
@@ -184,7 +184,7 @@ public async Task OpenAIHttpRetryPolicyTestAsync(string prompt, string expectedO
 
         this._kernelBuilder.Services.AddSingleton<ILoggerFactory>(this._testOutputHelper);
         this._kernelBuilder
-            .AddOpenAITextGeneration(
+            .AddOpenAIChatCompletion(
                 serviceId: openAIConfiguration.ServiceId,
                 modelId: openAIConfiguration.ModelId,
```
SergeyMenshykh, line 189: Should not the `ChatModelId`be used instead of `ModelId` ?

SergeyMenshykh (approved):

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
dmytrostruk: Thanks, resolved.

RogerBarreto (approved):

markwallace-microsoft (changes requested):
Blocking until I've had a chance to review

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAICompletionTests.cs
```diff
@@ -435,7 +438,7 @@ public async Task MultipleServiceLoadPromptConfigTestAsync()
             @"{
                 ""name"": ""FishMarket2"",
                 ""execution_settings"": {
-                    ""azure-open-ai"": {
+                    ""azure-text-davinci-003"": {
```
RogerBarreto: Do we need this?

dotnet/src/IntegrationTests/PromptTests.cs
```diff
@@ -101,6 +101,7 @@ private void ConfigureAzureOpenAI(IKernelBuilder kernelBuilder)
 
         kernelBuilder.AddAzureOpenAIChatCompletion(
             deploymentName: azureOpenAIConfiguration.DeploymentName,
+            modelId: azureOpenAIConfiguration.ModelId,
```
RogerBarreto, line 104:
```suggestion
            modelId: azureOpenAIConfiguration.ChatModelId,
```

dotnet/src/IntegrationTests/TestSettings/AzureOpenAIConfiguration.cs
```diff
@@ -12,16 +12,25 @@ internal sealed class AzureOpenAIConfiguration
 
     public string DeploymentName { get; set; }
 
+    public string ModelId { get; set; }
```
RogerBarreto, line 15:
```suggestion
```

dotnet/src/IntegrationTests/TestSettings/AzureOpenAIConfiguration.cs
```diff
@@ -12,16 +12,25 @@ internal sealed class AzureOpenAIConfiguration
...
     public string Endpoint { get; set; }
 
     public string ApiKey { get; set; }
 
-    public AzureOpenAIConfiguration(string serviceId, string endpoint, string apiKey, string deploymentName, string? embeddingModelId = null)
+    public AzureOpenAIConfiguration(string serviceId, string deploymentName, string endpoint, string apiKey, string chatDeploymentName, string? modelId = null, string? chatModelId = null, string? embeddingModelId = null)
     {
         this.ServiceId = serviceId;
         this.DeploymentName = deploymentName;
+        this.ModelId = modelId ?? deploymentName;
```
RogerBarreto, line 31:
```suggestion
```

dotnet/src/IntegrationTests/TestSettings/OpenAIConfiguration.cs
```diff
@@ -10,12 +10,14 @@ internal sealed class OpenAIConfiguration
 {
     public string ServiceId { get; set; }
     public string ModelId { get; set; }
```
RogerBarreto, line 12:
```suggestion
```

dotnet/src/IntegrationTests/TestSettings/OpenAIConfiguration.cs
```diff
@@ -10,12 +10,14 @@ internal sealed class OpenAIConfiguration
...
+    public string ChatModelId { get; set; }
     public string ApiKey { get; set; }
 
-    public OpenAIConfiguration(string serviceId, string modelId, string apiKey)
+    public OpenAIConfiguration(string serviceId, string modelId, string apiKey, string chatModelId)
     {
         this.ServiceId = serviceId;
         this.ModelId = modelId;
```
RogerBarreto, line 19:
```suggestion
```

dotnet/src/IntegrationTests/testsettings.json
```diff
@@ -1,12 +1,14 @@
 ﻿{
   "OpenAI": {
     "ServiceId": "open-ai",
-    "ModelId": "gpt-3.5-turbo",
+    "ModelId": "text-davinci-003",
```
RogerBarreto, line 4:
```suggestion
```

dotnet/src/IntegrationTests/testsettings.json
```diff
@@ -1,12 +1,14 @@
...
+    "ChatModelId": "gpt-3.5-turbo",
     "ApiKey": ""
   },
   "AzureOpenAI": {
     "ServiceId": "azure-open-ai",
-    "DeploymentName": "gpt-35-turbo",
+    "DeploymentName": "text-davinci-003",
```
RogerBarreto, line 10:
```suggestion
```

RogerBarreto (commented):
Suggestion

dotnet/src/IntegrationTests/TestSettings/OpenAIConfiguration.cs
dmytrostruk: @RogerBarreto this change won't work because there is another settings for embeddings which uses `ModelId` and we use the same C# class for all the settings:
```json
"OpenAIEmbeddings": {
  "ServiceId": "text-embedding-ada-002",
  "ModelId": "text-embedding-ada-002",
  "ApiKey": ""
},
```

That's why I removed `Chat` prefix initially, reverted it back and proposed to keep it as it is for now as alternative, because settings and testing infrastructure need to be revisited completely.

dmytrostruk:
Closing this PR as text generation services can work with `gpt-3.5-turbo-instruct` model which is not going to be deprecated.

stephentoub:
> Closing this PR as text generation services can work with `gpt-3.5-turbo-instruct` model which is not going to be deprecated.

The ITextGenerationService implementation in OpenAIChatCompletionService doesn't work with it? Isn't it _identical_ to the implementation used by OpenAITextGenerationService?

dmytrostruk:
> The ITextGenerationService implementation in OpenAIChatCompletionService doesn't work with it? Isn't it _identical_ to the implementation used by OpenAITextGenerationService?

When you use kernel with `AddOpenAIChatCompletion` extention method and use `gpt-3.5-turbo-instruct` model - it will throw following exception:
```
Microsoft.SemanticKernel.Http.HttpOperationException: 'This is not a chat model and thus not supported in the v1/chat/completions endpoint. Did you mean to use v1/completions?
Status: 404 (Not Found)
```

That's because `OpenAIChatCompletionService` implements both `IChatCompletionService` and `ITextGenerationService` and we check for chat interface first in our logic:
https://github.com/microsoft/semantic-kernel/blob/d3773b21a9a9d221f910444b17313c38f628e425/dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs#L145-L157

So, in order to use `gpt-3.5-turbo-instruct`, the service should implement just `ITextGenerationService`, as it is in case of `OpenAITextGenerationService` and usage of `AddOpenAITextGeneration` extension method.

But I think ideally `OpenAIChatCompletionService` should implement just `IChatCompletionService` and `OpenAITextGenerationService` should implement just `ITextGenerationService` for clear differentiation. Users shouldn't pass text generation AI model to chat completion service class or extension method.

dmytrostruk:
The reason why `OpenAIChatCompletionService` implements both `IChatCompletionService` and `ITextGenerationService` interfaces is because previously kernel and functions worked only with `ITextGenerationService` interface and it was a limitation. Now, since we have this differentiation between interfaces inside a core logic, I don't think that `OpenAIChatCompletionService` still needs to implement `ITextGenerationService`.

Issue #4129, opened 2023-12-09
.Net: Are there any Debug level logger in RC3?


alkampfergit:
I configured my kernel with this simple instruction and I choose to have Debug level. 

Configuration just taken from post https://devblogs.microsoft.com/semantic-kernel/release-candidate-1-for-the-semantic-kernel-net-sdk-is-now-live/

```csharp
    {
        var kernelBuilder = new KernelBuilder();
        kernelBuilder.Services.AddLogging(l => l
            .AddConsole()
            .SetMinimumLevel(LogLevel.Debug)
        );
```

But when I launch the program I have no debug level, just info. It is a simple c# console application with no ASP.NET or any other functionalities, just a plain console app to demonstrate the basic capabilities of SK. 

Here is the output where I have only info level, I'm expecting to have debug with the sample prompt sent to my language model.

```
info: Microsoft.SemanticKernel.Connectors.AI.OpenAI.ChatCompletion.AzureOpenAIChatCompletionService[0]
      Prompt tokens: 152. Completion tokens: 25. Total tokens: 177.
Extracting audio file from video C:\temp\ssh.mp4
info: ExtractAudio[4]
      Function completed. Duration: 0.1818454s
info: Microsoft.SemanticKernel.Connectors.AI.OpenAI.ChatCompletion.AzureOpenAIChatCompletionService[0]
      Prompt tokens: 196. Completion tokens: 53. Total tokens: 249.
info: Chat[4]
      Function completed. Duration: 9.4503992s
```

I'd like to have debug to output more information about prompt used, etc.

If I Set the level to trace I'm able to seee trace level, but no debug, it seems that in RC Semantic Kernel is doing less logging that before.

stephentoub:
> I'd like to have debug to output more information about prompt used, etc.

Anything that might contain sensitive information (like a prompt) is traced at Trace level, which is the intended level for such information:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel
"Logs that contain the most detailed messages. These messages may contain sensitive application data. These messages are disabled by default and should never be enabled in a production environment."

Is there something in particular that's not getting logged that you'd hope would be logged, regardless of the level at which it's being logged?  Or is there something getting logged at Trace level that you believe is more appropriate for a different level?

alkampfergit:
If you look at langsmith for LangChain you will see that you have a
complete log of all interaction with the language model.

I'd like to have log of what is happening, like

[...]

PR #4247, opened 2023-12-13
.Net: Log complex objects as Jsons

TaoChenOSU:
### Motivation and Context

Logging complex objects as Json allows developers to see the entire object and makes it easier to parse the logs.
Related: https://github.com/microsoft/semantic-kernel/issues/4129

### Description

Log complex objects as Jsons while keeping the high-performant logging pattern.
image

### Contribution Checklist


dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -127,11 +128,21 @@ internal static partial class PlannerInstrumentation
...
-        Message = "Plan:\n{Plan}")]
-    static partial void LogPlan(this ILogger logger, object plan);
+    private static readonly Action<ILogger, string, Exception?> s_logPlan =
+        LoggerMessage.Define<string>(
+            logLevel: LogLevel.Trace,   // Sensitive data, logging as trace, disabled by default
+            eventId: 3,
+            "Plan:\n{Plan}");
+    private static void LogPlan(this ILogger logger, object plan)
+    {
+        if (logger.IsEnabled(LogLevel.Trace))
```
glahaye, line 138: With this type of code, we have to check whether the given log level is enabled or not ourselves?  This isn't taken care of automagically?

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -127,11 +128,21 @@ internal static partial class PlannerInstrumentation
...
-    [LoggerMessage(
-        EventId = 3,
-        Level = LogLevel.Trace, // Sensitive data, logging as trace, disabled by default
-        Message = "Plan:\n{Plan}")]
-    static partial void LogPlan(this ILogger logger, object plan);
+    private static readonly Action<ILogger, string, Exception?> s_logPlan =
+        LoggerMessage.Define<string>(
+            logLevel: LogLevel.Trace,   // Sensitive data, logging as trace, disabled by default
+            eventId: 3,
+            "Plan:\n{Plan}");
```
teresaqhoang, line 135: Can we log the goal here too? Helps gauge expected behavior

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -122,7 +122,8 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
 
         // Ensure arguments are initialized.
         arguments ??= new KernelArguments();
-        logger.LogFunctionInvokingWithArguments(this.Name, arguments);
+        logger.LogFunctionInvoking(this.Name);
+        logger.LogFunctionArguments(arguments);
```
glahaye, line 126: Do we have to split the logging here and for the results?

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -160,11 +171,22 @@ internal static partial class PlannerInstrumentation
...
+            eventId: 8,
+            "Plan result: {Result}");
+
+    private static void LogPlanResult(this ILogger logger, object planResult)
+    {
+        if (logger.IsEnabled(LogLevel.Trace))
+        {
+            s_logPlanResult(
+                logger,
+                JsonSerializer.Serialize(planResult),
```
teresaqhoang, line 186: What if the plan result is a string? It'd be doubly serialized
```suggestion
                planResult.GetType() == typeof(string) ? planResult : JsonSerializer.Serialize(planResult),
```

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
stephentoub: > This isn't taken care of automagically?

Not before the arguments are evaluated.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
TaoChenOSU: Yes, because one is Information level, while the other is Trace level.

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
TaoChenOSU: The goal is logged before the planner is invoked.

teresaqhoang (approved):

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -159,7 +160,7 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
                 throw new OperationCanceledException($"A {nameof(Kernel)}.{nameof(Kernel.FunctionInvoked)} event handler requested cancellation after function invocation.");
             }
 
-            logger.LogFunctionInvokedSuccess(functionResult.Value);
+            logger.LogFunctionResult(functionResult.Value);
```
markwallace-microsoft, line 163: This just logs the `FunctionResult.Value` so the function name isn't accurate. 

Should we change the function name or really log the `FunctionResult`

markwallace-microsoft (commented):
One nit on naming

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -160,11 +171,22 @@ internal static partial class PlannerInstrumentation
...
 
-    [LoggerMessage(
-        EventId = 8,
-        Level = LogLevel.Trace, // Sensitive data, logging as trace, disabled by default
-        Message = "Plan result: {PlanResult}")]
-    static partial void LogPlanResult(this ILogger logger, object planResult);
+    private static readonly Action<ILogger, string, Exception?> s_logPlanResult =
+        LoggerMessage.Define<string>(
+            logLevel: LogLevel.Trace,   // Sensitive data, logging as trace, disabled by default
+            eventId: 8,
```
RogerBarreto, line 177: CAn we have an enumerable or a description of what `8` really is?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionLogMessages.cs
```diff
@@ -17,18 +18,38 @@ internal static partial class KernelFunctionLogMessages
...
+
+    /// <summary>
+    /// Logs arguments to a <see cref="KernelFunction"/>.
+    /// The action provides the benefit of caching the template parsing result for better performance.
+    /// And the public method is a helper to serialize the arguments.
+    /// </summary>
+    private static readonly Action<ILogger, string, Exception?> s_logFunctionArguments =
+        LoggerMessage.Define<string>(
+            logLevel: LogLevel.Trace,   // Sensitive data, logging as trace, disabled by default
+            eventId: 1,
```
RogerBarreto, line 35: Need more details on what is `1`, enumerable here may help

RogerBarreto (commented):
Add identifiers for the given event numbers so we can track this to where it's used.

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
TaoChenOSU: These numbers don't carry any meaning at the moment. I think we need to audit the entire SK project (all places where logs are generated) to determine the event code for each event if we want to identify the events. For now, I have change all of them to 0, which is the default if you are just using `ILogger.LogXXX`. And I have also suppressed the warning.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionLogMessages.cs
TaoChenOSU: Same as above

dotnet/src/InternalUtilities/planning/PlannerInstrumentation.cs
```diff
@@ -109,38 +110,49 @@ internal static partial class PlannerInstrumentation
...
+#pragma warning disable SYSLIB1006 // Multiple logging methods cannot use the same event id within a class
     [LoggerMessage(
         EventId = 0,
         Level = LogLevel.Information,
         Message = "Plan creation started.")]
     static partial void LogCreatePlanStarted(this ILogger logger);
 
     [LoggerMessage(
-        EventId = 1,
+        EventId = 0,
```
stephentoub, line 121: Why are all of these event ids being set to zero?

TaoChenOSU: Copied from another response:
These numbers don't carry any meaning at the moment. I think we need to audit the entire SK project (all places where logs are generated) to determine the event code for each event if we want to identify the events. For now, I have change all of them to 0, which is the default if you are just using ILogger.LogXXX. And I have also suppressed the warning.

It will become very trivial to maintain and identify each log event with a unique ID. For now, I think we can use the default. Please correct me if I am wrong.

alliscode: My only question would be if changing these to non-zero in the future constitutes a breaking change? My understanding is that it wouldn't be since they are all set to zero now, no one will be writing queries utilizing specific event IDs. @TaoChenOSU can you confirm this?

markwallace-microsoft (approved):

TaoChenOSU:
After this PR, all existing logs, whether it's using the `LoggerMessage` attribute or the `ILogger.LogXXX` APIs, will have an event ID of 0. In future versions, if we decide we need to identify log events using IDs, we can then evaluate a comprehensive ID schema to ensure proper ID range selection and maintainability of the SDK. It wouldn't be considered a breaking change to establish an ID schema since currently the event IDs don't carry any meaning (they are all 0).

alliscode (approved):

alliscode (approved):

Issue #4164, opened 2023-12-11
.Net: Function calling passes empty arguments to a prompt function


qmatteoq:
**Describe the bug**
I have a sample project which uses Semantic Kernel which includes 3 plugins:

- A plugin with two native functions, which are used to retrieve statistical data about the United States
- A plugin with a prompt function, which is used to generate a business mail

My goal is to process the following prompt:

```
Write a mail to share the population of the United States in 2015. Then add also, among the population, how many people identified themselves as male. Don't share approximations, please share the exact numbers.
```

In order to get a response, I need Semantic Kernel to use all the three plugins I've loaded. As such, I've enabled function calling in the following way:

```
OpenAIPromptExecutionSettings settings = new()
{
    FunctionCallBehavior = FunctionCallBehavior.AutoInvokeKernelFunctions
};

var chatHistory = new ChatHistory();
chatHistory.AddUserMessage("Write a paragraph to share the population of the United States in 2015. Then add also, among the population, how many people identified themselves as male. Don't share approximations, please share the exact numbers.");

var chatCompletionService = kernel.Services.GetService<IChatCompletionService>();
var result = await chatCompletionService.GetChatMessageContentAsync(chatHistory, settings, kernel);
```

The code works, but the outcome is partially correct. I can see that the two native functions are properly called, while the prompt function to write a business mail don't. The text of the mail has been clearly generated by the LLM without following any of the instructions I have in my prompt function.
I've switched to manual function calling to better debug what's going on with the following code:

```
OpenAIPromptExecutionSettings settings = new()
{
    FunctionCallBehavior = FunctionCallBehavior.EnableKernelFunctions
};

var chatHistory = new ChatHistory();
chatHistory.AddUserMessage("Write a paragraph to share the population of the United States in 2015. Then add also, among the population, how many people identified themselves as male. Don't share approximations, please share the exact numbers.");

var chatCompletionService = kernel.Services.GetService<IChatCompletionService>();
var result = await chatCompletionService.GetChatMessageContentAsync(chatHistory, settings, kernel);

//as long as the content is null, it means that the chat completion service is waiting for a function call to be processed
var functionCall = ((OpenAIChatMessageContent)result).GetOpenAIFunctionResponse();
while (functionCall != null)
{
    KernelFunction pluginFunction;
    KernelArguments arguments;
    kernel.Plugins.TryGetFunctionAndArguments(functionCall, out pluginFunction, out arguments);
[...]

PR #4256, opened 2023-12-14
.Net: Overhaul how JSON Schema is handled

stephentoub:
- KernelJsonSchema now validates that the schema provided is actually valid schema: previously it was only validating it was valid JSON. This should help to reduce invalid type information flowing through the system.
- Kernel{Return}ParameterMetdata now infers a schema from the provided Type if no schema was provided. This schema is cached in the instance.  That means we're no longer (re)creating the schema every single time we try to send this parameter to OpenAI, to the Handlebars planner, etc., and every connector that needs the schema doesn't need to recreate the same logic: if a schema can be created from the Type, it's done implicitly.
~~- To accomodate that, the Kernel{Return}Metadata types now use a constructor rather than init properties. With init props, there's no good place to initialize the schema (we'd need to do the work as part of the ParameterType's init, but at that point we don't know if it'll be throwaway because the schema may not have been set _yet_).~~
- KernelFunctionFromMethod was generating super complicated and not very helpful schema for return types of methods that returned Task and friends. KernelFunctionFromMethod now unwraps such types to get at the actual T and describe that instead, since that's what's actually marshaled out as the result; the intermediate Task is just plumbing.
- To keep the schema a bit smaller, it no longer include the description in the schema if the description is empty.
- If a parameter doesn't have a schema in a PromptTemplateConfig, it now defaults to string.
- If a parameter doesn't have any type or schema information at all, it's now described to OpenAI as a string rather than being left out entirely.
- The dependency on Newtonsoft was removed.
- Note that all this required moving down the dependency on the JsonSchema library as an implementation detail to the abstractions library.

Fixes https://github.com/microsoft/semantic-kernel/issues/4164


alliscode (approved):

stephentoub:
I just realized I can make the schema creation lazy, in which case I can also revert the public api changes to the metadata types. I'll do so later today.

stephentoub:
Ok, ready for review again. I undid the public API changes on Kernel{Return}ParameterMetadata.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #4265, opened 2023-12-14
OrderedAIServiceSelector logic should change as specified in issue 4234

### LINKED ISSUE #4234: OrderedAIServiceSelector logic should change
The new logic of the OrderedAIServiceSelector should be the following:

1. Loop over all services to see if there's a match on service ID
2. If not, loop over all the services again to see if there's a match on model ID
3. If not, use the default settings

markwallace-microsoft:
### Motivation and Context

Closes #4234 

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Services/OrderedAIServiceSelector.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Services/OrderedAIServiceSelector.cs
@@ -35,6 +35,7 @@ internal sealed class OrderedAIServiceSelector : IAIServiceSelector
         else
         {
             PromptExecutionSettings? defaultExecutionSettings = null;
+            // Search by service id first
             foreach (var keyValue in executionSettings)
             {
                 var settings = keyValue.Value;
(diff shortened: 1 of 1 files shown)


SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Services/OrderedAIServiceSelector.cs
```diff
@@ -52,6 +53,13 @@ internal sealed class OrderedAIServiceSelector : IAIServiceSelector
                         return true;
                     }
                 }
+            }
+
+            // Search by model id next
+            foreach (var keyValue in executionSettings)
```
RogerBarreto, line 59: Two loops in the same thing, see if we can use the same loop. 

Check this in the first
``` 
if (string.IsNullOrWhiteSpace(settings.ModelId))
```

RogerBarreto (approved):
The foreach can be improved

dotnet/src/SemanticKernel.Abstractions/Services/OrderedAIServiceSelector.cs
markwallace-microsoft: It was previously one loop but Matthew wants it split to be two so we search by service id first and only if that fails do we try by model id.

dotnet/src/SemanticKernel.Abstractions/Services/OrderedAIServiceSelector.cs
```diff
@@ -52,6 +53,13 @@ internal sealed class OrderedAIServiceSelector : IAIServiceSelector
...
+            {
+                var settings = keyValue.Value;
+                var serviceId = keyValue.Key;
```
RogerBarreto, line 62:
```suggestion
```

PR #4306, opened 2023-12-15
.Net: Public API improvements

dmytrostruk:
### Motivation and Context

Small public API improvements:
1. Removed unused classes from OpenAI Connector - `TextEmbeddingRequest`, `TextEmbeddingResponse`, `AzureOpenAIImageOperationStatus`, `AzureOpenAITextToImageResponse`.
2. Made following classes in OpenAI Connector as internal - `TextToImageRequest`, `TextToImageResponse`.
3. Moved `HttpOperationException` class to the root namespace - `Microsoft.SemanticKernel`.
4. Removed `ChatHistory.AddMessage` public method since there is `ChatHistory.Add` method with the same signature.
5. Moved `MemoryBuilder` from `Plugins.Memory` to `SemanticKernel.Core`, which allowed to decouple several connectors (including OpenAI) from `Plugins.Memory` project.

### Contribution Checklist


SergeyMenshykh (approved):

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/TextToImageResponse.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.SemanticKernel.Connectors.OpenAI;
 /// Text to image response
 /// </summary>
 [Experimental("SKEXP0012")]
```
markwallace-microsoft, line 12: Does it need to be experimental if it is internal?

dotnet/src/Connectors/Connectors.OpenAI/TextToImage/TextToImageRequest.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.SemanticKernel.Connectors.OpenAI;
 /// Text to image request
 /// </summary>
 [Experimental("SKEXP0012")]
```
markwallace-microsoft, line 11: Does it need to be experimental if it is internal?

markwallace-microsoft (approved):

PR #4319, opened 2023-12-15
new skill/plugin suggestion

rom212:
### Motivation and Context

I created this skill/plugin to illustrate some capabilities in a blog post. It turns out, this skill works well and I thought we could add to the sample library. 

### Contribution Checklist

### DIFF
--- a/samples/plugins/ClassificationPlugin/Spam/config.json
+++ b/samples/plugins/ClassificationPlugin/Spam/config.json
@@ -1,0 +1,12 @@
+{
+  "schema": 1,
+  "type": "completion",
+  "description": "Tells you whether a given text is spam",
+  "completion": {
+    "max_tokens": 256,
+    "temperature": 0.0,
+    "top_p": 0.0,
+    "presence_penalty": 0.0,
+    "frequency_penalty": 0.0
+  }
+}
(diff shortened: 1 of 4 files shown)


Krzysztof318:
How effective (%) is that prompt on many different examples? Did you do tests? I think this prompt is too general.

rom212:
> How effective (%) is that prompt on many different examples? Did you do tests? I think this prompt is too general and requires more specific to add that plugin to kernel.

It scores 100% on this test set: https://github.com/rom212/semantic_skill_evaluation/blob/master/batchdata.jsonl

By comparison, it's not less specific than the other two that already exist in the kernel for classification use cases (question and importance).

Did you find that it was not working well?

Krzysztof318:
I wasn't testing this yet. Makes me wonder how it would work as an extra layer for example when filtering an email just after the spam assasin layer.

markwallace-microsoft:
@rom212 Thanks for the PR but this has been inactive for a while and the code is out of sync with latest Semantic Kernel patterns. If you are interested in updating this please take a look at the current sample plugins https://github.com/microsoft/semantic-kernel/tree/main/samples/plugins

PR #4329, opened 2023-12-17
.Net: Update to Azure.AI.OpenAI beta 12

### LINKED ISSUE #4336: .Net: ChatHistory is not serializable RC4
**Describe the bug**
When serializing and deserializing a ChatHistory object, 2 problems occur: tool_calls have a null content instead of an empty string, which gets rejected by the OpenAI API the next time it sees those messages, and something within the internal structure of the ChatHistory changes, prompting the error "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'" even though the objects look the same before and after serialization

**To Reproduce**
Here is my demo:
```
using Microsoft.SemanticKernel.ChatCompletion;
using Microsoft.SemanticKernel.Connectors.OpenAI;
using System.ComponentModel;
using System.Text.Json;

namespace SKTest
{
    public class DemoPlugin
    {
        [KernelFunction, Description("creates a special poem")]
        public string CreateSpecialPoem()
        {          
            return "ABCDE";
        }
    }
    internal class Program
    {
        static async Task Main(string[] args)
        {
            Console.WriteLine("Hello, World!");

            builder.AddAzureOpenAIChatCompletion(
                     "gpt-4-turbo",                      // Azure OpenAI Deployment Name
                     "endpoint", // Azure OpenAI Endpoint
                     "key");      // Azure OpenAI Key
            
            builder.Plugins.AddFromType<DemoPlugin>();
           var kernel = builder.Build();
            AzureOpenAIChatCompletionService service = new AzureOpenAIChatCompletionService("gpt-4-turbo",                      // Azure OpenAI Deployment Name
                     "endpoint", // Azure OpenAI Endpoint
                     "key");      // Azure OpenAI Key
      OpenAIPromptExecutionSettings settings = new() { ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions };
      ChatHistory history = new ChatHistory();

      history.AddUserMessage("Make me a special poem");
var historyBeforeJson = JsonSerializer.Serialize(history.ToList(), new JsonSerializerOptions { WriteIndented = true});

      ChatMessageContent result = await service.GetChatMessageContentAsync(history, settings, kernel);          
      Console.WriteLine(result.ToString());
      history.AddUserMessage("ok thank you");

      ChatMessageContent resultOriginalWorking = await service.GetChatMessageContentAsync(history, settings, kernel);

      //problem 2: after deserialization, something within the message structure triggers the problem 
      //Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'
      //even though the JSONs are the same

      var historyJson = JsonSerializer.Serialize<ChatHistory>(history, new JsonSerializerOptions { WriteIndented = true });
      var historyAfterSerialization = JsonSerializer.Deserialize<ChatHistory>(historyJson);

      //problem 1: a call_tools message has a null content, gets serialized as null, and can not be used again without fixing that
      foreach (var damagedMessage in historyAfterSerialization.Where(msg => msg.Content == null))
      {
          damagedMessage.Content = string.Empty;
      }

      var historyAfterSerializationJson = JsonSerializer.Serialize(historyAfterSerialization, new JsonSerializerOptions { WriteIndented = true });


      ChatMessageContent resultError = await service.GetChatMessageContentAsync(historyAfterSerialization, settings, kernel);
      Console.WriteLine(resultError.ToString());
}
}
```

**Expected behavior**
ChatHistory to be serializable, an easier access to the underlying ChatMessages to recreate the History including toolcalls, or a hint on how the ChatHistory is supposed to be persisted instead

**Platform**
RC4, .net8

stephentoub:
Adapting to breaking changes in how streaming tooling updates surface.

Closes #4336


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunctionToolCall.cs
```diff
@@ -84,4 +85,90 @@ public override string ToString()
...
+                {
+                    functionArgmentBuildersByIndex[ftc.ToolCallIndex] = argBuilder = new();
+                }
+
+                argBuilder.Append(ftc.ArgumentsUpdate);
+            }
+        }
+    }
+
+    // TODO: Remove this temporary hack once the Azure SDK has been updated to set `base.Type = "function"` in the ChatCompletionsFunctionToolCall ctor
```
stephentoub, line 136: @trrwilson, am I correct that this is a bug in Beta 12? If not, how should I be changing this code based on what's exposed in the Azure SDK?

stephentoub:
This also fixes https://github.com/microsoft/semantic-kernel/issues/4336.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunctionToolCall.cs
```diff
@@ -84,4 +86,95 @@ public override string ToString()
...
+    /// </summary>
+    /// <param name="update">The tool call update to incorporate.</param>
+    /// <param name="toolCallIdsByIndex">Lazily-initialized dictionary mapping indices to IDs.</param>
+    /// <param name="functionNamesByIndex">Lazily-initialized dictionary mapping indices to names.</param>
+    /// <param name="functionArgumentBuildersByIndex">Lazily-initialized dictionary mapping indices to arguments.</param>
+    internal static void TrackStreamingToolingUpdate(
+        StreamingToolCallUpdate? update,
+        ref Dictionary<int, string>? toolCallIdsByIndex,
+        ref Dictionary<int, string>? functionNamesByIndex,
+        ref Dictionary<int, StringBuilder>? functionArgumentBuildersByIndex)
```
markwallace-microsoft, line 101: nit: Would it be better to have one Dictionary and a class to store the tool call id, function name and function arguments?

markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunctionToolCall.cs
stephentoub: If we expect only a small number of tool requests per interaction, your suggestion is better. If we expect a larger number, this is better.  It's probably the case that fewer is more likely.  If it's ok with you, though, let's hold off on changing it until we have more runway. Here I just copied the recommended approach from the Azure SDK readme.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ChatHistoryExtensions.cs
```diff
@@ -25,40 +24,24 @@ public static class ChatHistoryExtensions
...
-            if (metadata is null && chatMessage.Metadata is not null)
-            {
-                metadata = (Dictionary<string, object?>)chatMessage.Metadata;
-            }
+            metadata ??= (Dictionary<string, object?>?)chatMessage.Metadata;
 
             if (chatMessage.Content is { Length: > 0 } contentUpdate)
             {
-                contentBuilder.Append(contentUpdate);
+                (contentBuilder ??= new()).Append(contentUpdate);
```
RogerBarreto, line 41: I think majority 99.9% of our cases we will have a streaming with content, and checking if the contentBuilder is null might not be necessary.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -420,14 +420,21 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
         var chatOptions = CreateChatCompletionsOptions(chatExecutionSettings, chat, kernel, this.DeploymentOrModelName);
 
         StringBuilder? contentBuilder = null;
+        Dictionary<int, string>? toolCallIdsByIndex = null;
```
RogerBarreto, line 423: Can we have a List of Struct instead of 3 dictionaries? Might be faster.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ChatHistoryExtensions.cs
stephentoub: Majority, yes. When there are function calls, though, it's often empty when not using auto invoke.

RogerBarreto (approved):

PR #4332, opened 2023-12-17
Fix a few issues found via compiler warnings

stephentoub:
I temporarily added a net8.0 target to a few projects and fixed the resulting issues.

### DIFF
--- a/dotnet/src/SemanticKernel.Core/Contents/StreamingMethodContent.cs
+++ b/dotnet/src/SemanticKernel.Core/Contents/StreamingMethodContent.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+using System;
 using System.Text;
 
 namespace Microsoft.SemanticKernel;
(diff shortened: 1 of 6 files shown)


SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Core/Contents/StreamingMethodContent.cs
```diff
@@ -23,13 +24,15 @@ public override byte[] ToByteArray()
...
+        return this.Content?.ToString() is string s ?
+            Encoding.UTF8.GetBytes(s) :
+            Array.Empty<byte>();
     }
 
     /// <inheritdoc/>
     public override string ToString()
     {
-        return this.Content.ToString();
+        return this.Content.ToString() ?? string.Empty;
```
WeihanLi, line 35: Should we change the return type from `string` to `string?` since the base method defined in `Object` is using `string?`

https://github.com/dotnet/runtime/blob/4dffd80c4d77c27e772a0be26e8036af77fbb26e/src/libraries/System.Private.CoreLib/src/System/Object.cs#L39

stephentoub: No, we want this to be non-nullable.

WeihanLi: Thanks

dmytrostruk (approved):

PR #4344, opened 2023-12-17
.Net: Fixing issue that causes a stack overflow in handlebars planner.

alliscode:
The handlebars planner converts plugin parameter types to `HandlebarsParameterTypeMetadata` objects by recursively traversing the properties of the type. If the type is itself defined recursively, this causes stack overflow exception. This PR addresses this issue by checking if a given type has already been processed before processing it.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Planners/Planners.Handlebars/Handlebars/Extensions/KernelParameterMetadataExtensions.cs
+++ b/dotnet/src/Planners/Planners.Handlebars/Handlebars/Extensions/KernelParameterMetadataExtensions.cs
@@ -34,6 +34,11 @@ internal static class KernelParameterMetadataExtensions
     /// </summary>
     public static HashSet<HandlebarsParameterTypeMetadata> ToHandlebarsParameterTypeMetadata(this Type type)
     {
+        return type.ToHandlebarsParameterTypeMetadata(new HashSet<Type>());
+    }
+
+    private static HashSet<HandlebarsParameterTypeMetadata> ToHandlebarsParameterTypeMetadata(this Type type, HashSet<Type> processedTypes)
+    {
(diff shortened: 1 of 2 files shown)


teresaqhoang (approved):

dotnet/src/Planners/Planners.Handlebars/Handlebars/Extensions/KernelParameterMetadataExtensions.cs
```diff
@@ -73,7 +73,10 @@ private static void AddNestedComplexTypes(this HashSet<HandlebarsParameterTypeMe
         // Add nested complex types
         foreach (var property in properties)
         {
-            parameterTypes.UnionWith(property.PropertyType.ToHandlebarsParameterTypeMetadata());
+            if (!parameterTypes.Any(pt => pt.Name == property.PropertyType.Name))
```
RogerBarreto, line 76: Shouldn't we match the types directly instead of name?
pt == property.PropertyType ?
```suggestion
            if (!parameterTypes.Any(pt => pt == property.PropertyType))
```

alliscode: pt isn't a Type, it's a `HandlebarsParameterTypeMetadata` which unfortunately doesn't have a Type in it. I'll see if it 's reasonable to add it...

RogerBarreto: All good, if this will take long we can go with current state.

markwallace-microsoft (approved):

markwallace-microsoft (approved):

PR #4345, opened 2023-12-17
.Net InputVariable.Default type is changed from string to object?

SergeyMenshykh:
### Motivation and Context    
The SK is no longer string-centric and can now work with any .NET type, including strings. Functions can have parameters and return types of any type. Prompt variables can also be of any type. The last remaining element tied to the string type is the `InputVariable.Default` property. This PR addresses that.

### Description  
- The `InputVariable.Default` property type changed from string to object.  
- All instances where the property is used are refactored accordingly. Where the string value of the property is required, it is converted to a string; in other cases, it is used as is.  
- Safeguards that prevent supplying default values of any type other than string in JSON and YAML prompt configurations are temporarily added until a proper, permanent solution is designed and implemented. More details below.

### JsonSerializer & DeserializerBuilder handling primitives, objects, and arrays  
   
Both JsonSerializer and DeserializerBuilder (from YamlDotNet) can infer the types of properties of the data model class into which they deserialize a string when a concrete, non-object type is specified. However, when the target type of a property is `object`, each attempts to accomplish deserialization differently.  
   
The JsonSerializer, when deserializing to a property of `object` type, assigns a JsonElement (configurable, can be JsonNode) to the property, regardless of whether the type specified in the JSON is a primitive, array, or object.  
   
The DeserializerBuilder deserializes all primitives as strings unless the `!!type` tag is used - `!!int`, `!!bool`, etc. As for complex types and arrays, those are deserialized as `Dictionary<object, object>` and `List<object>`, respectively.  
   
Taking that into account, the solution should address the problem of having or using different types to represent the same data. The different types can be adapted to:  
  
  - One universal data type for JSON/YAML objects (perhaps JsonElement?)  
  - `List<object>` for JSON/YAML arrays  
  - .NET primitives for JSON and YAML primitives - number, boolean, string  
   
This way, function authors can write their functions regardless of whether the function is called from a JSON prompt or a YAML prompt.

### Default value and schema  
   
In addition to standard attributes/properties of function parameter metadata such as name, description, isRequired, and default, it's possible to specify a [JSON schema](https://json-schema.org/) to add extra information about the type of root and nested elements. 

Ideally, the solution should take the schema into account, if specified, and use the type indicated there to convert the default value to at deserialization. In cases where the schema is not provided, the functionality should rely solely on the data types provided in JSON/YAML documents.


dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
```diff
@@ -136,7 +136,7 @@ internal IReadOnlyList<KernelParameterMetadata> GetKernelParametersMetadata()
                 Description = p.Description,
                 DefaultValue = p.Default,
                 IsRequired = p.IsRequired,
-                ParameterType = !string.IsNullOrWhiteSpace(p.JsonSchema) ? null : typeof(string),
+                ParameterType = !string.IsNullOrWhiteSpace(p.JsonSchema) ? null : p.Default?.GetType(),
```
stephentoub, line 139: If there's a null default, should it default to string? Otherwise, what happens for inferred inputs?

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
```diff
@@ -169,7 +169,29 @@ internal IReadOnlyList<KernelParameterMetadata> GetKernelParametersMetadata()
...
     {
-        var result = JsonSerializer.Deserialize<PromptTemplateConfig>(json, JsonOptionsCache.ReadPermissive);
-        return result ?? throw new ArgumentException("Unable to deserialize prompt template config from argument. The deserialization returned null.", nameof(json));
+        var promptTemplateConfig = JsonSerializer.Deserialize<PromptTemplateConfig>(json, JsonOptionsCache.ReadPermissive);
+        if (promptTemplateConfig == null)
+        {
+            throw new ArgumentException("Unable to deserialize prompt template config from argument. The deserialization returned null.", nameof(json));
+        }
+
+        // Prevent the default value from being any type other than a string.
```
stephentoub, line 178: Why?

dotnet/src/Extensions/PromptTemplates.Handlebars/HandlebarsPromptTemplate.cs
```diff
@@ -96,10 +96,12 @@ private KernelArguments GetVariables(KernelArguments? arguments)
 
         foreach (var p in this._promptModel.InputVariables)
         {
-            if (!string.IsNullOrEmpty(p.Default))
+            if (p.Default == null || (p.Default is string stringDefault && string.IsNullOrEmpty(stringDefault)))
```
stephentoub, line 99: Nit: by the time we get to the IsNullOrEmpty check, we know it's not null, so it could just be a length check

dotnet/src/Functions/Functions.Yaml/KernelFunctionYaml.cs
```diff
@@ -30,6 +31,15 @@ public static class KernelFunctionYaml
 
         var promptTemplateConfig = deserializer.Deserialize<PromptTemplateConfig>(text);
 
+        // Prevent the default value from being any type other than a string.
+        foreach (var inputVariable in promptTemplateConfig.InputVariables)
+        {
+            if (inputVariable.Default != null && inputVariable.Default is not string)
```
stephentoub, line 37: Nit: is not null instead of != null for consistency?

stephentoub (approved):

dotnet/src/Functions/Functions.UnitTests/Yaml/Functions/KernelFunctionYamlTests.cs
```diff
@@ -72,78 +82,117 @@ public void ItShouldSupportCreatingOpenAIExecutionSettings()
...
+        template:        Say hello world to {{$name}} in {{$language}}
+        description:     Say hello to the specified person using the specified language
+        name:            SayHello
+        input_variables:
+          - name:          name
+            description:   The name of the person to greet
+            default:       John
+          - name:          language
+            description:   The language to generate the greeting in
+            default: English
```
markwallace-microsoft, line 180: nit: Formatting

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
```diff
@@ -169,7 +169,29 @@ internal IReadOnlyList<KernelParameterMetadata> GetKernelParametersMetadata()
...
+            // The value of the default property becomes a JsonElement after deserialization because that is how the JsonSerializer handles properties of the object type.
+            if (inputVariable.Default is JsonElement element)
+            {
+                if (element.ValueKind == JsonValueKind.String)
+                {
+                    inputVariable.Default = element.ToString();
+                }
+                else
+                {
+                    throw new ArgumentException($"Default value for input variable '{inputVariable.Name}' must be a string. Prompt function - '{promptTemplateConfig.Name ?? promptTemplateConfig.Description}'.");
```
markwallace-microsoft, line 190: This is a temporary limitation so can the error message reflect that.

dotnet/src/SemanticKernel.UnitTests/PromptTemplate/PromptTemplateConfigTests.cs
```diff
@@ -187,4 +188,81 @@ public void DeserializingExpectOutputVariable()
...
+        // string
+        var json = CreateJson((string)"123");
+        var config = PromptTemplateConfig.FromJson(json);
+
+        Assert.NotNull(config?.InputVariables);
+        Assert.Equal("123", config.InputVariables[0].Default?.ToString());
+    }
+
+    [Fact]
+    public void ItShouldThrowExceptionWhenDeserializingConfigWithDefaultValueOtherThanString()
```
markwallace-microsoft, line 225: nit: Would be good to have a comment here to state this is a temporary limitation.

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
SergeyMenshykh: It's a temporary limitation that allows us to shape the public API surface (changing the type of the `Default` property to object) now, before the release. This helps avoid a breaking change while we work on a proper solution for dealing with the different deserialization outputs of JSON/YAML prompt configurations.

dotnet/src/Extensions/PromptTemplates.Handlebars/HandlebarsPromptTemplate.cs
RogerBarreto: Should we also consider whitespaces?

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
RogerBarreto: NotSupportedException ?
```suggestion
                    throw new NotSupportedException($"Default value for input variable '{inputVariable.Name}' must be a string. Prompt function - '{promptTemplateConfig.Name ?? promptTemplateConfig.Description}'.");
```

dotnet/src/Extensions/PromptTemplates.Handlebars/HandlebarsPromptTemplate.cs
SergeyMenshykh: Thanks, fixed.

RogerBarreto (approved):

dotnet/src/SemanticKernel.UnitTests/PromptTemplate/PromptTemplateConfigTests.cs
SergeyMenshykh: Done

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
SergeyMenshykh: Done

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
SergeyMenshykh: To be honest, I'm not sure that it should default to a string. Currently, both the `KernelPromptTemplate` and `HandlebarsPromptTemplate` classes support string and non-string data types, such as primitives, arrays, and objects. Neither class infers the types of prompt variables - https://github.com/microsoft/semantic-kernel/issues/4350. Defaulting to a string could be misleading since it may not match the actual type of the prompt variables, the type of which is impossible to infer—see the 'idea' variable in the issue mentioned above. As such, null would be a more appropriate indicator of an unknown type. However, there is no harm in defaulting to a string since there's logic in place to convert string arguments to their respective function parameter types, which would be triggered if string arguments were provided for non-string function parameters. For now, I will likely add this fallback to a string to be on the safe side, and we might revise the logic later.

stephentoub: Ok. We very recently added the string default fallback, but we can revisit if there's a better answer. For that particular motivating case, I believe we also added a string fallback to openai, so it shouldn't actually matter for that.

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
stephentoub: Ok. It'd be good if the comment indicated the why.

RogerBarreto (approved):

PR #4358, opened 2023-12-18
.Net: BugFix Serializing and Deserializing ChatHistory with ToolCalling details Update OpenAI Connector to use FunctionToolCallsProperty and add ChatHistoryTests

### LINKED ISSUE #4336: .Net: ChatHistory is not serializable RC4

RogerBarreto:
Serializing the ToolCalls was losing the details about Name and Arguments, and serializing this lack of content back was rendering ToolCalls as an invalid history message and giving the Issue Error.

Resolves #4336 

Next PR, moving the Integration Tests as Unit Tests
## Summary
This pull request updates the OpenAI Connector to use the FunctionToolCallsProperty metadata key instead of the ToolCallsProperty metadata key. This change allows for the use of a list of ChatCompletionsFunctionToolCall objects instead of ChatCompletionsToolCall objects. Additionally, a new integration test, ChatHistoryTests.cs, is added to the IntegrationTests/Connectors/OpenAI directory. The test file contains tests for the ChatHistory feature of the OpenAI Connector, ensuring that the feature returns the expected results and handles errors correctly.

## Changes
- Update ClientCore.cs to use FunctionToolCallsProperty instead of ToolCallsProperty
- Update OpenAIChatMessageContent.cs to use FunctionToolCallsProperty instead of ToolCallsProperty
- Add ChatHistoryTests.cs to IntegrationTests/Connectors/OpenAI directory
- Implement tests for ChatHistory feature of OpenAI Connector
- Dispose of logger and test output helper in ChatHistoryTests.cs

---

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
@@ -21,12 +21,12 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
     /// <summary>
     /// Gets the metadata key for the list of <see cref="ChatCompletionsFunctionToolCall"/>.
     /// </summary>
-    public static string FunctionToolCallsProperty => $"{nameof(ChatResponseMessage)}.FunctionToolCalls";
+    internal static string FunctionToolCallsProperty => $"{nameof(ChatResponseMessage)}.FunctionToolCalls";
 
     /// <summary>
     /// Gets the metadata key for the list of <see cref="ChatCompletionsToolCall"/>.
     /// </summary>
-    public static string ToolCallsProperty => $"{nameof(ChatResponseMessage)}.ToolCalls";
+    internal static string ToolCallsProperty => $"{nameof(ChatResponseMessage)}.ToolCalls";
 
     /// <summary>
     /// Initializes a new instance of the <see cref="OpenAIChatMessageContent"/> class.
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
@@ -21,7 +21,12 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
     /// <summary>
     /// Gets the metadata key for the <see cref="ChatCompletionsToolCall.Id"/> name property.
     /// </summary>
-    public static string ToolCallsProperty => $"{nameof(ChatResponseMessage)}.{nameof(ChatResponseMessage.ToolCalls)}";
+    public static string FunctionToolCallsProperty => $"{nameof(ChatResponseMessage)}.FunctionToolCalls";
+
+    /// <summary>
+    /// Gets the metadata key for the <see cref="ChatCompletionsToolCall.Id"/> name property.
+    /// </summary>
+    public static string ToolCallsProperty => $"{nameof(ChatResponseMessage)}.ToolCalls";
 
     /// <summary>
     /// Initializes a new instance of the <see cref="OpenAIChatMessageContent"/> class.
@@ -32,13 +37,13 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
     internal OpenAIChatMessageContent(ChatResponseMessage chatMessage, string modelId, IReadOnlyDictionary<string, object?>? metadata = null)
         : base(new AuthorRole(chatMessage.Role.ToString()), chatMessage.Content, modelId, chatMessage, System.Text.Encoding.UTF8, CreateMetadataDictionary(chatMessage.ToolCalls.OfType<ChatCompletionsFunctionToolCall>().ToList(), metadata))
     {
-        this.ToolCalls = chatMessage.ToolCalls.OfType<ChatCompletionsFunctionToolCall>().ToList();
+        this.ToolCalls = chatMessage.ToolCalls.OfType<ChatCompletionsToolCall>().ToList();
     }
 
     /// <summary>
     /// Initializes a new instance of the <see cref="OpenAIChatMessageContent"/> class.
     /// </summary>
-    internal OpenAIChatMessageContent(ChatRole role, string? content, string modelId, IReadOnlyList<ChatCompletionsFunctionToolCall> toolCalls, IReadOnlyDictionary<string, object?>? metadata = null)
+    internal OpenAIChatMessageContent(ChatRole role, string? content, string modelId, IReadOnlyList<ChatCompletionsToolCall> toolCalls, IReadOnlyDictionary<string, object?>? metadata = null)
         : base(new AuthorRole(role.ToString()), content, modelId, content, System.Text.Encoding.UTF8, CreateMetadataDictionary(toolCalls, metadata))
     {
         this.ToolCalls = toolCalls;
@@ -47,7 +52,7 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
     /// <summary>
     /// Initializes a new instance of the <see cref="OpenAIChatMessageContent"/> class.
     /// </summary>
-    internal OpenAIChatMessageContent(AuthorRole role, string? content, string modelId, IReadOnlyList<ChatCompletionsFunctionToolCall> toolCalls, IReadOnlyDictionary<string, object?>? metadata = null)
+    internal OpenAIChatMessageContent(AuthorRole role, string? content, string modelId, IReadOnlyList<ChatCompletionsToolCall> toolCalls, IReadOnlyDictionary<string, object?>? metadata = null)
         : base(role, content, modelId, content, System.Text.Encoding.UTF8, CreateMetadataDictionary(toolCalls, metadata))
     {
         this.ToolCalls = toolCalls;
@@ -56,7 +61,7 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
     /// <summary>
     /// A list of the tools called by the model.
     /// </summary>
-    public IReadOnlyList<ChatCompletionsFunctionToolCall> ToolCalls { get; }
+    public IReadOnlyList<ChatCompletionsToolCall> ToolCalls { get; }
 
     /// <summary>
     /// Retrieve the resulting function from the chat result.
@@ -68,7 +73,10 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
 
         foreach (var toolCall in this.ToolCalls)
         {
-            (functionToolCallList ??= new List<OpenAIFunctionToolCall>()).Add(new OpenAIFunctionToolCall(toolCall));
+            if (toolCall is ChatCompletionsFunctionToolCall functionToolCall)
+            {
+                (functionToolCallList ??= new List<OpenAIFunctionToolCall>()).Add(new OpenAIFunctionToolCall(functionToolCall));
+            }
         }
 
         if (functionToolCallList is not null)
@@ -80,7 +88,7 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
     }
 
     private static IReadOnlyDictionary<string, object?>? CreateMetadataDictionary(
-        IReadOnlyList<ChatCompletionsFunctionToolCall> toolCalls,
+        IReadOnlyList<ChatCompletionsToolCall> toolCalls,
         IReadOnlyDictionary<string, object?>? original)
     {
         // We only need to augment the metadata if there are any tool calls.
@@ -108,6 +116,7 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
             }
 
             // Add the additional entry.
+            newDictionary.Add(FunctionToolCallsProperty, toolCalls.OfType<ChatCompletionsFunctionToolCall>().ToList());
             newDictionary.Add(ToolCallsProperty, toolCalls);
 
             return newDictionary;
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
@@ -19,12 +19,12 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
     public static string ToolIdProperty => $"{nameof(ChatCompletionsToolCall)}.{nameof(ChatCompletionsToolCall.Id)}";
 
     /// <summary>
-    /// Gets the metadata key for the <see cref="ChatCompletionsToolCall.Id"/> name property.
+    /// Gets the metadata key for the list of <see cref="ChatCompletionsFunctionToolCall"/>.
     /// </summary>
     public static string FunctionToolCallsProperty => $"{nameof(ChatResponseMessage)}.FunctionToolCalls";
 
     /// <summary>
-    /// Gets the metadata key for the <see cref="ChatCompletionsToolCall.Id"/> name property.
+    /// Gets the metadata key for the list of <see cref="ChatCompletionsToolCall"/>.
     /// </summary>
     public static string ToolCallsProperty => $"{nameof(ChatResponseMessage)}.ToolCalls";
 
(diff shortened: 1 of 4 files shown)


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
```diff
@@ -55,7 +56,7 @@ internal OpenAIChatMessageContent(AuthorRole role, string? content, string model
     /// <summary>
     /// A list of the tools called by the model.
     /// </summary>
-    public IReadOnlyList<ChatCompletionsToolCall> ToolCalls { get; }
+    public IReadOnlyList<ChatCompletionsFunctionToolCall> ToolCalls { get; }
```
stephentoub, line 59: Isn't this going to be a problem when the Azure SDK adds other kinds of tools?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
```diff
@@ -82,7 +80,7 @@ public IReadOnlyList<OpenAIFunctionToolCall> GetOpenAIFunctionToolCalls()
     }
 
     private static IReadOnlyDictionary<string, object?>? CreateMetadataDictionary(
-        IReadOnlyList<ChatCompletionsToolCall> toolCalls,
+        IReadOnlyList<ChatCompletionsFunctionToolCall> toolCalls,
```
stephentoub, line 83: I would have thought just changing the type stored into the metadata dictionary word be sufficient. Is it not?

dotnet/src/IntegrationTests/Connectors/OpenAI/ChatHistoryTests.cs
```diff
@@ -0,0 +1,133 @@
...
+        history.AddUserMessage("Make me a special poem");
+        var historyBeforeJson = JsonSerializer.Serialize(history.ToList(), s_jsonOptionsCache);
+        var service = kernel.GetRequiredService<IChatCompletionService>();
+        ChatMessageContent result = await service.GetChatMessageContentAsync(history, settings, kernel);
+        history.AddUserMessage("Ok thank you");
+
+        ChatMessageContent resultOriginalWorking = await service.GetChatMessageContentAsync(history, settings, kernel);
+        var historyJson = JsonSerializer.Serialize(history, s_jsonOptionsCache);
+        var historyAfterSerialization = JsonSerializer.Deserialize<ChatHistory>(historyJson);
+        ChatMessageContent shouldNotThrow = await service.GetChatMessageContentAsync(historyAfterSerialization!, settings, kernel);
```
dmytrostruk, line 68: nit:
```suggestion
        var exception = await Record.ExceptionAsync(() => service.GetChatMessageContentAsync(historyAfterSerialization!, settings, kernel));
        
        Assert.Null(exception);
```

dotnet/src/IntegrationTests/Connectors/OpenAI/ChatHistoryTests.cs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private void ConfigureAzureOpenAIChatAsText(IKernelBuilder kernelBuilder)
+    {
+        var azureOpenAIConfiguration = this._configuration.GetSection("Planners:AzureOpenAI").Get<AzureOpenAIConfiguration>();
```
dmytrostruk, line 73: Is this correct configuration for `ChatHistory` tests?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
RogerBarreto: I'm trying only that and see if it suffices.

dotnet/src/IntegrationTests/Connectors/OpenAI/ChatHistoryTests.cs
RogerBarreto: Yeah, I need the planners configuration because it uses the gpt-4-preview to work. Older models will break.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
```diff
@@ -110,6 +116,7 @@ public IReadOnlyList<OpenAIFunctionToolCall> GetOpenAIFunctionToolCalls()
             }
 
             // Add the additional entry.
+            newDictionary.Add(FunctionToolCallsProperty, toolCalls.OfType<ChatCompletionsFunctionToolCall>().ToList());
```
SergeyMenshykh, line 119: Are both necessary?

RogerBarreto: ONe don't serialize the function call specific details, the other does.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
stephentoub: If it doesn't, we'll need to do something else... I don't think we can change the public surface area like this.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
RogerBarreto: ONe way to resolve this later is to actually transform the ChatCompletionsToolCall object in a JsonElement and add it to the dictionary, or better approach.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
RogerBarreto: Yes, moved back to use the abstraction, but the serialization is lost. I added a specific metadata property for it but I still don't think this is the best approach, probably should just add a JsonElement to the dictionary object.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
```diff
@@ -18,9 +19,14 @@ public sealed class OpenAIChatMessageContent : ChatMessageContent
...
-    /// Gets the metadata key for the <see cref="ChatCompletionsToolCall.Id"/> name property.
+    /// Gets the metadata key for the list of <see cref="ChatCompletionsFunctionToolCall"/>.
+    /// </summary>
+    public static string FunctionToolCallsProperty => $"{nameof(ChatResponseMessage)}.FunctionToolCalls";
+
+    /// <summary>
+    /// Gets the metadata key for the list of <see cref="ChatCompletionsToolCall"/>.
     /// </summary>
-    public static string ToolCallsProperty => $"{nameof(ChatResponseMessage)}.{nameof(ChatResponseMessage.ToolCalls)}";
+    public static string ToolCallsProperty => $"{nameof(ChatResponseMessage)}.ToolCalls";
```
stephentoub, line 29: Curious why you changed this?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
stephentoub: Serializing both is going to duplicate information. If we have to serialize non-function tool calls, can we filter it down to just the ones that aren't functions? For now that should be empty, so most of the time we don't need to add anything.

stephentoub: Or just do the jsonelement thing now, as you suggest

RogerBarreto: Moving to internal

RogerBarreto: I moved the properties to internal and I will be changing this behavior internally.

Looking currently in a definitive solution, I tried the JsonElement approach but it didn't work initially, will look at this in more detail.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
RogerBarreto: Copy paste. Reverted.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
```diff
@@ -110,6 +116,7 @@ public IReadOnlyList<OpenAIFunctionToolCall> GetOpenAIFunctionToolCalls()
             }
 
             // Add the additional entry.
+            newDictionary.Add(FunctionToolCallsProperty, toolCalls.OfType<ChatCompletionsFunctionToolCall>().ToList());
             newDictionary.Add(ToolCallsProperty, toolCalls);
```
markwallace-microsoft, line 120: Should we remove this? 

This should only contain the same data as what is stored under `FunctionToolCallsProperty`. We could just serialise the one type of tools we know about and extend this later to serialise other tools.

stephentoub: Yeah, I think the only way this goes wrong is if the sdk sends us another kind of tool, which doesn't exist yet. Let's do the simple thing now that doesn't negatively impact the serialized state, and then subsequently follow up with a change to use something that gives us more control over what's serialized and how.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunctionToolCall.cs
```diff
@@ -176,5 +176,5 @@ internal static ChatCompletionsFunctionToolCall CreateChatCompletionsFunctionToo
         s_type?.SetValue(c, "function");
         return c;
     }
-    private static readonly PropertyInfo? s_type = typeof(ChatCompletionsToolCall).GetProperty("Type", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
+    private static readonly PropertyInfo? s_type = typeof(ChatCompletionsFunctionToolCall).GetProperty("Type", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
```
stephentoub, line 179: Was this change necessary? I wouldn't expect it to be.

dotnet/src/IntegrationTests/Connectors/OpenAI/ChatHistoryTests.cs
```diff
@@ -0,0 +1,137 @@
...
+    public class MyHttpMessageHandler : DelegatingHandler
+    {
+        public MyHttpMessageHandler()
+        {
+            this.InnerHandler = new HttpClientHandler();
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            string? requestContent = await request.Content!.ReadAsStringAsync(cancellationToken);
```
stephentoub, line 112: What is the purpose of this?

dotnet/src/IntegrationTests/Connectors/OpenAI/ChatHistoryTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    public void Dispose()
+    {
+        this.Dispose(true);
+        GC.SuppressFinalize(this);
+    }
+
+    ~ChatHistoryTests()
```
stephentoub, line 124: This doesn't need to be finalizable

dotnet/src/IntegrationTests/Connectors/OpenAI/ChatHistoryTests.cs
RogerBarreto: I used to understand how to request was being sent to the server, was important to spot the problem, I can remove it....

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs
RogerBarreto: Removed the property

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunctionToolCall.cs
RogerBarreto: Reverted

dmytrostruk (approved):

SergeyMenshykh (approved):

Issue #3237, opened 2023-10-18
Python: recall method in StepwisePlan not using context


tarockey:
**Describe the bug**
When using the StepwisePlanner, context is not being passed to skills. This is manifesting when using the `recall` ability of the TextMemorySkill in a skill's prompt, always using the default settings (e.g. collection name, limits, etc.). This means that skills cannot be context-dependent.

**To Reproduce**
1. Create a new Kernel, register the TextMemorySkill, and a memory store (currently tested with Azure Cognitive Search as memory store)
2. Create a skill which uses the {{recall}} method inside the prompt template, and register to the kernel.
3. Define a new context with the name of your ACS index as the COLLECTION_PARAM for TextMemorySkill (as shown below):
```python
context = kernel.create_new_context()
context[sk.core_skills.TextMemorySkill.COLLECTION_PARAM] = memory_collection_name
context[sk.core_skills.TextMemorySkill.RELEVANCE_PARAM] = 0.5
context[sk.core_skills.TextMemorySkill.LIMIT_PARAM] = 5
```
4. Create a new SequentialPlanner, passing in your kernel, and a question as the goal.
5. Invoke(`invoke_async`) your plan, passing in the context and memory as parameters:
```python
result = await plan.invoke_async(context=context, memory=kernel.memory)
```
6. Review results, noting that if your ACS index is not named "generic", it will fail to retrieve the relevant memories.
7. Enumerate through the plan._steps, and note the outputs. You will likely see the error
```
Error occurred: Function `recall` execution failed. ResourceNotFoundError: () The index 'generic' for service 'YOUR ACS INDEX' was not found.
```

**Expected behavior**
The recall function should have been passed the correct collection (index) name, and returned relevant memories for the prompt to use as additional context.

**Platform**
 - OS:Windows 11
 - IDE: VS Code
 - Language: Python
 - Source: pip package version 0.3.13.dev0

**Additional Info**

Full repro code below:

```python
import semantic_kernel as sk
from dotenv import load_dotenv
from semantic_kernel.connectors.ai.open_ai import AzureChatCompletion, AzureTextEmbedding
from requests import get
import json
import os
from Plugins.HAPlugin.Homeassistant import HomeAssistant
from semantic_kernel.planning import StepwisePlanner
from semantic_kernel.planning.stepwise_planner.stepwise_planner_config import (
    StepwisePlannerConfig,
)
[...]

PR #4366, opened 2023-12-18
Python: Prevents LLMs from stripping complex function names

juliomenendez:
### Motivation and Context

When using the `StepwisePlanner` planner along with core skills, for example `TextMemorySkill`, the LLMs tend to plans with function names stripped down to the name without the skill name. This makes the plan fail with an exception like:

`Error invoking action recall: (<ErrorCodes.UnknownError: -1>, "The function 'recall' was not found.", None)`

This error is because the completion returned by the planner prompt does not include the fully structured skill function name. Ex.: instead of returning `_GLOBAL_FUNCTIONS_.recall` it returns `recall` or `GLOBAL_FUNCTIONS.recall`.

This is somehow relates #3237

### Description

This change instructs the LLM to keep the skill function name unchanged in its response by adding an example that includes `_` in the name. This approach mimics how the `SequentialPlanner` solves this same issue.

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
+++ b/python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
@@ -24,6 +24,18 @@ Here is an example of a valid $JSON_BLOB:
   "action": "functionName",
   "action_variables": {"parameterName": "some value", ...}
 }
+
+Here is another example of a valid $JSON_BLOB:
+{
+  "action": "_Function_.Name",
+  "action_variables": {"parameterName": "some value", ...}
+}
+
+Here is another example of a valid $JSON_BLOB:
+{
+  "action": "FunctionName2",
+  "action_variables": {"parameterName": "some value", ...}
+}
 [END USAGE INSTRUCTIONS]
 [END INSTRUCTION]
 
--- a/python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
+++ b/python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
@@ -36,6 +36,8 @@ Here is another example of a valid $JSON_BLOB:
   "action": "FunctionName2",
   "action_variables": {"parameterName": "some value", ...}
 }
+
+The $JSON_BLOB must contain an "action_variables" key, with the {"parameterName": "some value", ...} value in the response.
 [END USAGE INSTRUCTIONS]
 [END INSTRUCTION]
 


python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
```diff
@@ -24,6 +24,18 @@ Here is an example of a valid $JSON_BLOB:
   "action": "functionName",
   "action_variables": {"parameterName": "some value", ...}
 }
+
+Here is another example of a valid $JSON_BLOB:
```
rewrlution, line 28: Good to know! Out of curiosity, is there way to unit test prompts?

rewrlution (commented):
LGTM, just have a high level question.

python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
rewrlution: I heard people develop the idea of testing LLM prompts with another LLM.

In this case, how do you test your change? By running the notebooks?

juliomenendez: We are not unit testing them at the moment AFAIK, but yes, we can write unit tests around prompts. If we do, we should consider testing them against all the LLMs SK supports to ensure consistency in the responses.

python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
```diff
@@ -24,6 +24,18 @@ Here is an example of a valid $JSON_BLOB:
...
+{
+  "action": "_Function_.Name",
+  "action_variables": {"parameterName": "some value", ...}
+}
+
+Here is another example of a valid $JSON_BLOB:
+{
+  "action": "FunctionName2",
```
rewrlution, line 36: another question is about this example.

What makes it different from the first example?

```txt
"action": "functionName"
```

Is it just the naming convention?

rewrlution (commented):
Left another comment on the prompt itself.

python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
juliomenendez: By running a script provided by the user in #3237 and comparing the results from `SequentialPlanner` which worked out of the box vs `StepwisePlanner` which wasn't working.

python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
juliomenendez: Yes, capitalization was also not being respected. The LLM was "camelCasing" the function name in some instances. We want it to leave it as-is.

rewrlution: That makes sense to me.

rewrlution (approved):

python/semantic_kernel/planning/stepwise_planner/Skills/StepwiseStep/skprompt.txt
eavanvalkenburg: This would have to be a integration test since the unit tests do not interact with outside services, but having testing for plans, both creation and execution would be quite helpful.

eavanvalkenburg (approved):
Looks good, adding testing would be good, but can be a separate PR.

moonbox3 (approved):
Thanks so much for working on this!

moonbox3 (approved):

PR #4383, opened 2023-12-20
.Net Added configuration validation for examples 05-20

Krzysztof318:
### Motivation and Context

Validation of configuration in dotnet examples is inconsistent.
In some examples the validation is there in others it is gone which causes an Exception instead of bypassing the example.

### Description

Added new class ConfigurationValidator. 
Added more verbose information about missing configuration parameters.
Added validation to examples 05-20.
From now missing configuration doesn't cause exception instead skip example or method in example.

I'm waiting for your feedback if all is ok I will add validations to the other examples.

### Contribution Checklist


Krzysztof318:
Hi @RogerBarreto. This PR hans already 1month since I've created that.
Now I'm wondering does it have any sense after samples has converted to integration tests.
Give me feedback, should I close this PR?

Krzysztof318:
Due to conversion samples to tests I don't think so this PR makes any sense now.

Issue #4295, opened 2023-12-15
.Net: Upgrade LangVersion to C# 12 across all projects after 1.0 is released


stephentoub:
This is best achieved by requiring the .NET 8 SDK for building. If we don't want to drop support for the .NET 6/7 SDKs, it can also be enabled by referencing the compilers toolset nuget package, which is already done in several test projects.

RogerBarreto:
Can't wait! 😊

PR #4387, opened 2023-12-20
.Net: Use C# 11/12 features throughout repo

stephentoub:
~~Fixes https://github.com/microsoft/semantic-kernel/issues/4295. For now this is done using a package reference to the latest C# toolset.  Once we drop use of the .NET 6/7 SDKs in CI, this reference can be removed.~~

The changes to upgrade projects to LangVersion 12 was done separately. This PR now rolls out use of C# 12 features.


RogerBarreto (approved):

dmytrostruk (approved):

dotnet/samples/KernelSyntaxExamples/Example21_OpenAIPlugins.cs
```diff
@@ -47,12 +47,14 @@ public static async Task CallKlarnaAsync()
...
-        arguments["countryCode"] = "US";// ISO 3166 country code with 2 characters based on the user location.
-                                        // Currently, only US, GB, DE, SE and DK are supported.
+        var arguments = new KernelArguments
+        {
+            ["q"] = "Laptop",      // Category or product that needs to be searched for.
+            ["size"] = "3",        // Number of products to return
+            ["budget"] = "200",    // Maximum price of the matching product in local currency
+            ["countryCode"] = "US" // ISO 3166 country code with 2 characters based on the user location.
+        };
+        // Currently, only US, GB, DE, SE and DK are supported.
```
markwallace-microsoft, line 57: One goal of the kernel syntax example sis developer sand copy/paste snippets of code into their applications. Will the majority of developers already be using C# 12?

stephentoub: I don't have data on that, unfortunately. The repo was already using C# 10, which is later than anyone using NET Framework targets by default (by default it uses C# 7.3 unless overridden). But anyone can opt-in to a newer language version, regardless of target framework, so it's generally not an imposition to move to a newer language version, and features in these newer releases are natural fits for the kinds of workloads SK supports, eg raw string literals for working with JSON, or collection expression syntax being very natural for someone familiar with python, etc. Your concern is totally fair, but I don't think it should stop us from making forward, simplifying progress here.

Krzysztof318:
It would be nice if we can also use c#11 "required" feature still targeting net standard2.0 like here [LINK](https://stackoverflow.com/questions/74447497/is-it-possible-to-use-the-c11-required-modifier-with-net-framework-4-8-and)
But I never tried solution from link, is that good way?

stephentoub:
@markwallace-microsoft, are you ok with this?

stephentoub:
> It would be nice if we can also use c#11 "required" feature still targeting net standard2.0 like here [LINK](https://stackoverflow.com/questions/74447497/is-it-possible-to-use-the-c11-required-modifier-with-net-framework-4-8-and) But I never tried solution from link, is that good way?

It's possible to do that, but it would force any consumer to be using C# 11+.

JohnYoungers:
@stephentoub - my apologies for the unrelated comment, but assuming these changes to the 223 files were not painstakingly done by hand, would you be able to elaborate on the steps involved in generating the changes for the lang version upgrade?

We'll be upgrading our applications to dotnet8 soon, but as of today we don't have any automated steps on improving code quality with the switch; if that's a simple step (e.g. `dotnet format ...`), I'd like to include it if possible

stephentoub:
> @stephentoub - my apologies for the unrelated comment, but assuming these changes to the 223 files were not painstakingly done by hand, would you be able to elaborate on the steps involved in generating the changes for the lang version upgrade?
> 
> We'll be upgrading our applications to dotnet8 soon, but as of today we don't have any automated steps on improving code quality with the switch; if that's a simple step (e.g. `dotnet format ...`), I'd like to include it if possible

The actual upgrade part was just this:
https://github.com/microsoft/semantic-kernel/pull/4387/commits/4e90039c50d2e6b76a83fe8eb7c507d4010a1467

Everything else (the vast majority of the change) was taking advantage of newer features in the language. And that was all (or almost all) done via the various Roslyn analyzers that are available in Visual Studio, e.g.
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0290
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0300
and just running the auto-fixer over the solution, as well as refactorings like "Convert to raw string":
image

markwallace-microsoft (approved):

Krzysztof318:
@stephentoub  When will it be merged? I see here are 3 approval, auto-merge is enabled, it only needs one small formatting fix to pass the pipeline.

stephentoub:
> it only needs one small formatting fix to pass the pipeline.

That's not the case. It's running into assembly loading issues when using the .NET 7 SDK. I haven't had time to investigate further.

Krzysztof318:
@stephentoub I didn't notice the build error, I thought it was just a formatter error. I corrected my comment actually immediately, but you responded so fast ;)

Krzysztof318:
@stephentoub I have made small research and that doesn't look good. I probably found what is problem described here  https://github.com/dotnet/format/issues/1546 This looks bundled dotnet format causes this issue, but I can be wrong.

dotnet/Directory.Build.props
```diff
@@ -6,11 +6,17 @@
...
-    <LangVersion>10</LangVersion>
+    <LangVersion>Latest</LangVersion>
     <Nullable>enable</Nullable>
     <ImplicitUsings>disable</ImplicitUsings>
   </PropertyGroup>
 
+  <ItemGroup>
+    <!-- Use newest available compiler to support C# 12. -->
+    <!-- This can be removed once we no longer target the .NET 6/7 SDKs in CI. -->
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="4.9.2" />
```
WeihanLi, line 17: Should we specify the `PrivateAssets="all"` to avoid this becoming the package dependency

stephentoub:
Depends on https://github.com/microsoft/semantic-kernel/pull/5802

stephentoub:
Parts refactored out into https://github.com/microsoft/semantic-kernel/pull/5819. The rest depends on that.

stephentoub:
This is ready for review.

dmytrostruk (approved):
Thanks for this PR!

Issue #3875, opened 2023-12-01
.Net: Need to audit for other expected breaking changes to come when Azure AI SDK goes v1.0 to match OpenAI


matthewbolanos:
Create tasks to support the following:

```[tasklist]
### Tasks
- [ ] OpenAI JSON mode. Possibly need to extend `ChatHistory` to include other attributes e.g. `response_format`. See https://platform.openai.com/docs/guides/text-generation/json-mode
- [ ] OpenAI System Fingerprint. Additional metadata that can be included in response from OpenAI chat completion. See https://platform.openai.com/docs/api-reference/chat/object#chat/object-system_fingerprint.
```

Issue #3540, opened 2023-11-17
.Net: Support for seed and response_format


askldjd:
Is there any plans to update the SK to support the recently announced OpenAI features - [JSON Mode](https://platform.openai.com/docs/guides/text-generation/json-mode) and [Reproducible Format](https://platform.openai.com/docs/guides/text-generation/reproducible-outputs)?

matthewbolanos:
@markwallace-microsoft, do you think these should be added as execution settings? Or should they be something else?

Alerinos:
Also needs this feature, SK wants to be competitive against other frameworks. OpenAI is owned by Microsoft just like SK. Such features should be introduced before OpenAI releases an update. 
Using SK I am still forced to write my own API or use other libraries.

RyanWebsterWSS:
I would also really benefit from an option to use json mode in responses - many of my prompts enforce a JSON output so it is easy for me to programmatically extract multiple bits of information. Without JsonMode, there is a less than desirable amount of parsing failures due to the model including extra information.

I hope this is planned for the future!

nsulikowski:
"To enable JSON mode, set the response_format parameter to { "type": "json_object" }. This configuration is crucial for receiving outputs in JSON format"

Having this feature as a first class would be great.

stephentoub:
Seed was added. The response mode is not there yet.

nsulikowski:
BTW, I do see this code in the Azure.AI.OpenAI.Tests project (i.e. JsonObject)... so I assume it is implemented in that nuget package:

```csharp
ChatCompletionsOptions chatCompletionsOptions = new()
{
    DeploymentName = deploymentOrModelName,
    Messages = { new ChatRequestUserMessage("give me a list of five fruits. JSON is a delightful wire format, don't you think?") },
    ResponseFormat = ChatCompletionsResponseFormat.JsonObject,
};
```

Alerinos:
I think we should withdraw ``[Experimental("SKEXP0010")] already.``

PR #4391, opened 2023-12-20
.Net: Add OpenAIPromptExecutionSettings.ResponseFormat

stephentoub:
Closes https://github.com/microsoft/semantic-kernel/issues/3540
Closes https://github.com/microsoft/semantic-kernel/issues/3875

I'm not sure what type to use here, so for now I've kept it `object?` and marked it experimental.  If we're ok tying OpenAIPromptExecutionSettings to the Azure SDK, then it'd probably be best to make the property of type `ChatCompletionsResponseFormat?`. Otherwise, if we don't think it'll ever evolve to have more properties, we could make it of type `string?`, or we could add a custom SK type for it. `object?` isn't as user-friendly, but it keeps the doors open to supporting various things.

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
@@ -755,6 +755,28 @@ internal abstract class ClientCore
             Seed = executionSettings.Seed,
         };
 
+        switch (executionSettings.ResponseFormat)
+        {
+            case ChatCompletionsResponseFormat formatObject:
+                // If the response format is an Azure SDK ChatCompletionsResponseFormat, just pass it along.
+                options.ResponseFormat = formatObject;
+                break;
+
+            case string formatString:
+                // If the response format is a string, map the ones we know about, and ignore the rest.
+                switch (formatString)
+                {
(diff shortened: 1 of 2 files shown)


RogerBarreto (approved):

SergeyMenshykh (approved):

Alerinos:
@stephentoub 
an update just came out, why is this feature marked as experimental? Why does it accept object possible null? 
This is a very important function for me in SK but because of this I can't use it....
```
Validity Code Description Project File Line Status Omitted Details
Error SKEXP0013 The element "Microsoft.SemanticKernel.Connectors.OpenAI.OpenAIPromptExecutionSettings.ResponseFormat" is for evaluation purposes only and may be changed or removed in future updates. Pause this diagnostic to continue.
```

stephentoub:
> an update just came out, why is this feature marked as experimental?

Because it was only recently added to the Azure.AI.OpenAI SDK (which is still beta) and thus is more likely to change, and because we're not sure if this should be the exact shape of the property due the long run.

> Why does it accept object possible null?

To use the default settings.

> but because of this I can't use it

Why not? If you're ok that it might change on the future, just suppress the warning, either with a pragma around it or a NoWarn in the project file.

PR #4408, opened 2023-12-23
.Net: Bug fixes for Handlebars Prompt Template and Planner

teresaqhoang:
### Motivation and Context

Squashing bugs related to variable access in the Handlebars Prompt Template Factory and enhancing the Handlebars Planner Create Plan prompt.

Since we allow the planner to set variables after initial render call, these variables that are added during plan execution are not bound to the template and will resolve as an `UndefinedBindingResult` as the template evaluates. This PR adds a fix to handle that scenario.

### Description

Resolving variables:
- If the argument is of type `UndefinedBindingResult`, it means that Handlebars attempted to retrieve the value for a binding but was unable to do so because the variable was not defined or not passed to the template context at the time of render. Thus, we try to get the value from the kernel arguments dictionary.

### Contribution Checklist


dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -152,20 +153,22 @@ private static bool IsExpectedParameterType(KernelParameterMetadata parameterMet
     private static void ProcessPositionalArguments(KernelFunctionMetadata functionMetadata, KernelArguments executionContext, Arguments handlebarsArguments)
     {
         var requiredParameters = functionMetadata.Parameters.Where(p => p.IsRequired).ToList();
+
         if (requiredParameters.Count <= handlebarsArguments.Length && handlebarsArguments.Length <= functionMetadata.Parameters.Count)
         {
             var argIndex = 0;
-            foreach (var arg in handlebarsArguments)
+            var arguments = KernelHelpersUtils.ProcessArguments(handlebarsArguments, executionContext);
+            foreach (var arg in arguments)
```
SergeyMenshykh, line 161: The code accesses the same argument using two different methods: directly via the 'arg' variable in line 164, and through an array indexer in lines 163, 166, and 171. To maintain consistency, consider using a `for` loop instead of `foreach`, which will allow you to access the argument solely through the array indexer.

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
```diff
@@ -25,7 +25,7 @@ public sealed class HandlebarsPlanner
     public static readonly HandlebarsPromptTemplateOptions PromptTemplateOptions = new()
     {
         // Options for built-in Handlebars helpers
-        Categories = new Category[] { Category.Math },
+        Categories = new Category[] { Category.DateTime },
```
SergeyMenshykh, line 28: Is this category for the default helpers included with the 'Handlebars.Net' NuGet package, or is it for the SK default helpers?

Issue #3643, opened 2023-11-23
Python: Chinese Windows Encoding Must Use GBK


Claireluyingling:
**Describe the bug**
UnicodeDecodeError: 'gbk' codec can't decode byte 0x8a in position 2: illegal multibyte sequence

**To Reproduce**
The general encoding uses utf8, but the Chinese Windows system seems to need to change the encoding to GBK to run properly.

**Expected behavior**
Adapt the Chinese Windows utf8 encoding format.

![Error](https://github.com/microsoft/semantic-kernel/assets/91041009/e1652a38-5630-49f1-8ecd-acc9f588d9ec)
![GBK](https://github.com/microsoft/semantic-kernel/assets/91041009/16106d34-e123-419f-a179-388ffda52a2f)

**Platform**
 - OS: Windows
 - IDE: VS Code
 - Language: Python

moonbox3:
Closing this issue as it is stale. Please recreate an issue if you continue to have a problem. Thanks.

sakesun:
On Windows, set environment variable "PYTHONUTF8" = "1" to workaround this problem.

PR #4409, opened 2023-12-23
Python: Fix issue 3643,  Python: Chinese Windows Encoding Must Use GBK

brianMacao:
### Motivation and Context

1. This is a issue 3643 and the bug is confirmed.
2. It will solve kernel.import_semantic_skill_from_directory method use encoding by locale.getencoding() problem.
3. When different language prompt gather in one project, and when this server default encoding is not utf-8, it may run the project fail.
4. open issue link https://github.com/microsoft/semantic-kernel/issues/3643

### Description

1. it just modify kernel.py file by the following chagning
 > a. import locale  # this is support by build-in, this change will not cause another import error
 > b. add a new parameter encoding into import_semantic_skill_from_directory method, and set a default value of locale.getencoding(), because the following reason:
 >> i. default setting of `open`, will invoke locale.getencoding() when not setting of encoding
 > > ii. initiate parameter in function paramter, will make a fix memory in this function init parameter, it can avoid a dense invoke system setting  

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/kernel.py
+++ b/python/semantic_kernel/kernel.py
@@ -63,6 +63,7 @@ logger: logging.Logger = logging.getLogger(__name__)
 
 _sys_encoding = locale.getpreferredencoding() if sys.version_info < (3, 11) else locale.getencoding()
 
+
 class Kernel:
     _skill_collection: SkillCollectionBase
     _prompt_template_engine: PromptTemplatingEngine
--- a/python/semantic_kernel/kernel.py
+++ b/python/semantic_kernel/kernel.py
@@ -4,6 +4,7 @@ import glob
 import importlib
 import inspect
 import locale
+import sys
 import logging
 import os
 from typing import Any, Callable, Dict, List, Optional, Type, TypeVar, Union
@@ -60,6 +61,7 @@ T = TypeVar("T")
 
 logger: logging.Logger = logging.getLogger(__name__)
 
+_sys_encoding = locale.getpreferredencoding() if sys.version_info < (3, 11) else locale.getencoding()
 
 class Kernel:
     _skill_collection: SkillCollectionBase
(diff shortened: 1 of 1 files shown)


python/semantic_kernel/kernel.py
```diff
@@ -828,7 +829,7 @@ def import_native_skill_from_directory(
         return {}
 
     def import_semantic_skill_from_directory(
-        self, parent_directory: str, skill_directory_name: str
+        self, parent_directory: str, skill_directory_name: str, encoding=locale.getencoding()
```
moonbox3, line 832: The `locale.getencoding()` will work with Python 3.11+. To support Python 3.8 - 3.10, we'd need to use something like:

```
import locale
import sys

if sys.version_info >= (3, 11):
    default_encoding = locale.getencoding()
else:
    default_encoding = locale.getpreferredencoding()
```

This could be represented in the method definition as

```
encoding=locale.getpreferredencoding() if sys.version_info < (3, 11) else locale.getencoding()
```

moonbox3 (changes requested):

yale8848:
Good  job, may merge soon ：）

moonbox3 (changes requested):
Hi @brianMacao, apologies for this delay on this - I didn't see a notification that the change was addressed. We've made some further updates to the naming of methods. Could you please pull in latest code so that your changes are integrated with the new method `import_semantic_plugin_from_directory`.

I also think we can name it `sys_encoding` instead of `_sys_encoding`. I understand you want it regarded as private, but I am not sure if the underscore is completely necessary.

moonbox3:
Hi @brianMacao, as I mentioned a few weeks ago, we've had a lot of underlying code changes. Please re-create the PR against the latest code, and we can review it. I am going to close this older, stale PR.

Issue #4308, opened 2023-12-15
Add net8.0 target framework


stephentoub:
Once 1.0 ships, we should consider adding multitargeting, so that the packages contain both a netstandard2.0 asset and a net8.0 asset. This will enable improved performance in a variety of ways, some implicit just as part of the build, and some with changes that can be made in the code under ifdef to take advantage of newer features.

PR #4421, opened 2023-12-25
Add multitargeting to .NET libraries

stephentoub:
Adds net8.0 targets and updates various code to take advantage of newer APIs and also fix analyzers.
Fixes https://github.com/microsoft/semantic-kernel/issues/4308


Krzysztof318:
What benefit does net8.0 targeting give us if we won't be able to use all the new APIs to maintain compatibility with net standard 2.0?

stephentoub:
> What benefit does net8.0 targeting give us if we won't be able to use all the new APIs to maintain compatibility with net standard 2.0?

When running on .NET 8+, apps will use the net8.0 binaries rather than the netstandard2.0 binaries. That means it _can_ use newer APIs. So, for example, it can use SocketsHttpHandler and its finer-grained control when on .NET 8+. It can use overloads that take a CancellationToken that don't exist in netstandard in order to improve cancellation and responsiveness. It can use built-in collections like PriorityQueue rather than being to ship its own version. And so on. It also implicitly benefits from new APIs in places the compiler targets them, eg interpolated strings will compile with the more efficient implementation introduced in C# 10 / .NET 6. It further has deployment benefits, eg System.Text.Json doesn't need to be deployed as part of the app because it's built-in to netcoreapp8. Etc.

stephentoub:
Depends on https://github.com/microsoft/semantic-kernel/pull/5802

stephentoub:
Depends on https://github.com/microsoft/semantic-kernel/pull/5819

stephentoub:
Still need to figure out why code coverage is unhappy, but other than that, this should be ready for review @markwallace-microsoft

stephentoub:
@dmytrostruk, I think you set up the code coverage checking? Can you make a recommendation for what to do here?
image

dmytrostruk:
> @dmytrostruk, I think you set up the code coverage checking? Can you make a recommendation for what to do here?

Yes, I can see some changes in OpenAI connector, so it worth to run `dotnet/code-coverage.ps1` locally to generate HTML report and see potential places that can be covered with unit-tests in that assembly. If you want, I can assist with that.

stephentoub:
There aren't any meaningful changes. I think the problem is that the library is already razor-thin above the code coverage line (main is sitting at 81% branch coverage for the OpenAI connector, and at < 80% CI fails), so anything that perturbs it even slightly could cause CI to totally fail. There's nothing related to my changes that wasn't covered, so I went and added unrelated tests to boost the coverage back above 80%, but that's not really a sustainable position. I suggest someone take some time to bump the coverage number up above 90% so there's much more wiggle room.

stephentoub:
Finally green. @markwallace-microsoft, please review when you get a chance.

dmytrostruk:
> I think the problem is that the library is already razor-thin above the code coverage line (main is sitting at 81% branch coverage for the OpenAI connector, and at < 80% CI fails), so anything that perturbs it even slightly could cause CI to totally fail.

@stephentoub That's true. When I was improving code coverage, I spent some time to hit 80% threshold by adding a lot of tests and back then I couldn't afford even more time to make it 90% or higher. On the other hand, I haven't noticed this problem recurring often, but maybe it's just me.

> There's nothing related to my changes that wasn't covered, so I went and added unrelated tests to boost the coverage back above 80%, but that's not really a sustainable position. I suggest someone take some time to bump the coverage number up above 90% so there's much more wiggle room.

Thank you! Yes, I agree, we should add more tests to give us some space before it breaks CI pipeline.

stephentoub:
> On the other hand, I haven't noticed this problem recurring often, but maybe it's just me.

I think it got worse for this particular library when all the various audio, image, etc. services were added, as they don't have as good coverage in the tests.

dotnet/src/Extensions/PromptTemplates.Liquid/LiquidPromptTemplate.cs
```diff
@@ -16,18 +16,24 @@ namespace Microsoft.SemanticKernel.PromptTemplates.Liquid;
 /// <summary>
 /// Represents a Liquid prompt template.
 /// </summary>
-internal sealed class LiquidPromptTemplate : IPromptTemplate
+internal sealed partial class LiquidPromptTemplate : IPromptTemplate
```
markwallace-microsoft, line 19: @stephentoub can you clarify why the change to `partial` is needed here?

WeihanLi: required by the regex source generator `GeneratedRegex`

dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -226,7 +226,7 @@ private static void ProcessPositionalArguments(KernelFunctionMetadata functionMe
             // Deserialize any JSON content or return the content as a string
             if (restApiOperationResponse.ContentType?.IndexOf("application/json", StringComparison.OrdinalIgnoreCase) >= 0)
             {
-                var parsedJson = JsonValue.Parse(restApiOperationResponse.Content.ToString());
+                var parsedJson = JsonValue.Parse(restApiOperationResponse.Content.ToString() ?? "");
```
RogerBarreto, line 229:
```suggestion
                var parsedJson = JsonValue.Parse(restApiOperationResponse.Content.ToString() ?? string.Empty);
```

dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelSystemHelpers.cs
```diff
@@ -81,7 +79,7 @@ internal static class KernelSystemHelpers
             else
             {
                 var args = ProcessArguments(arguments, variables);
-                name = args[0].ToString();
+                name = args[0].ToString() ?? "";
```
RogerBarreto, line 82:
```suggestion
                name = args[0].ToString() ?? string.Empty;
```

dotnet/src/Functions/Functions.OpenApi.Extensions/Extensions/ApiManifestKernelExtensions.cs
```diff
@@ -87,6 +87,11 @@ public static class ApiManifestKernelExtensions
             var apiDependencyDetails = apiDependency.Value;
 
             var apiDescriptionUrl = apiDependencyDetails.ApiDescriptionUrl;
+            if (apiDescriptionUrl is null)
+            {
+                logger.LogWarning("ApiDescriptionUrl is missing for API dependency: {ApiName}", apiName);
```
RogerBarreto, line 92: Apply for the other logger.LogXXX calls touched

```suggestion
                if (logger.IsEnabled(LogLevel.Warning)
                {
                    logger.LogWarning("ApiDescriptionUrl is missing for API dependency: {ApiName}", apiName);
                }
```

dotnet/src/Extensions/PromptTemplates.Liquid/LiquidPromptTemplate.cs
stephentoub: https://youtu.be/ptKjWPC7pqw?t=1428

dotnet/src/InternalUtilities/src/Diagnostics/Verify.cs
```diff
@@ -11,31 +11,50 @@
 namespace Microsoft.SemanticKernel;
 
 [ExcludeFromCodeCoverage]
-internal static class Verify
+internal static partial class Verify
 {
-    private static readonly Regex s_asciiLettersDigitsUnderscoresRegex = new("^[0-9A-Za-z_]*$");
-    private static readonly Regex s_filenameRegex = new("^[^.]+\\.[^.]+$");
+#if NET7_0_OR_GREATER
```
RogerBarreto, line 16: Some other places for `[GenerateRegex]` uses `#if NET` without versioning but this one is specific to 7. What is the only available above 7 in here?

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -83,12 +86,10 @@ public async Task<string> ExecuteCodeAsync([Description("The valid Python code t
             code = SanitizeCodeInput(code);
         }
 
-        if (this._logger.IsEnabled(LogLevel.Trace))
-        {
-            this._logger.LogTrace("Executing Python code: {Code}", code);
-        }
+        this._logger.LogTrace("Executing Python code: {Code}", code);
```
RogerBarreto, line 89: Does the removal of the `IsEnabled` check in here has a special reason ?

dotnet/src/Plugins/Plugins.MsGraph/CloudDrivePlugin.cs
```diff
@@ -47,9 +47,11 @@ public CloudDrivePlugin(ICloudDriveConnector connector, ILoggerFactory? loggerFa
...
 
         using StreamReader sr = new(fileContentStream);
-        string content = await sr.ReadToEndAsync().ConfigureAwait(false);
-
-        return content;
+#if NET
+        return await sr.ReadToEndAsync(cancellationToken).ConfigureAwait(false);
+#else
+        return await sr.ReadToEndAsync().ConfigureAwait(false);
+#endif
```
RogerBarreto, line 54: Following the WriteAsync change above, this sounds less verbose
```suggestion
return await sr.ReadToEndAsync(
#if NET
    cancellationToken
#endif
).ConfigureAwait(false);
```

dotnet/src/Plugins/Plugins.MsGraph/Connectors/Client/MsGraphClientLoggingHandler.cs
```diff
@@ -65,13 +65,26 @@ private void LogHttpMessage(HttpHeaders headers, Uri? uri, string prefix)
     {
         if (this._logger.IsEnabled(LogLevel.Debug))
         {
-            StringBuilder message = new();
-            message.AppendLine($"{prefix} {uri}");
+            var message = new StringBuilder().Append(prefix).Append(' ').Append(uri).AppendLine();
```
RogerBarreto, line 68:
```suggestion
            var message = new StringBuilder().Append(prefix).Append(' ').AppendLine(uri);
```

dotnet/src/SemanticKernel.Abstractions/Services/AIServiceExtensions.cs
```diff
@@ -91,19 +91,19 @@ public static class AIServiceExtensions
             return (service, settings);
         }
 
-        var message = new StringBuilder($"Required service of type {typeof(T)} not registered.");
+        var message = new StringBuilder().Append("Required service of type ").Append(typeof(T)).Append(" not registered.");
```
RogerBarreto, line 94: nit.

Wonder if in a future .NET release all interpolated strings could be compiled as string builders, if that makes sense.

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/TextToAudioRequest.cs
```diff
@@ -7,27 +7,20 @@ namespace Microsoft.SemanticKernel.Connectors.OpenAI;
 /// <summary>
 /// OpenAI text-to-audio request model, see <see href="https://platform.openai.com/docs/api-reference/audio/createSpeech"/>.
 /// </summary>
-internal sealed class TextToAudioRequest
+internal sealed class TextToAudioRequest(string model, string input, string voice)
```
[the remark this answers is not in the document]
stephentoub: It's not a field from a language perspective, but rather a parameter that has a larger scope. I don't want to do the suggested rename as that leads to poor consumption of the parameters at call sites. If you'd prefer, I can store them into actual fields and then use those fields. I typically do that for larger classes but do this for very short ones where the parameter is obvious.

dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
stephentoub: Do we not use "" in this repo? I can change it if we consistently use Empty (there's no perf difference, just style)

dotnet/src/Functions/Functions.OpenApi.Extensions/Extensions/ApiManifestKernelExtensions.cs
stephentoub: The called method itself does a check. Other than style, there's not a huge need for such a check. The check is important when there's meaningful work done at the call site which you only want to happen when logging is enabled, eg work done to compute an argument.

dotnet/src/InternalUtilities/src/Diagnostics/Verify.cs
stephentoub: If memory serves, this shared file is built into one or more projects that have a net6.0 tfm.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
stephentoub: https://github.com/microsoft/semantic-kernel/pull/4421#discussion_r1596702964

dotnet/src/Plugins/Plugins.MsGraph/Connectors/Client/MsGraphClientLoggingHandler.cs
stephentoub: There's no AppendLine overload that works with a Uri object.

dotnet/src/SemanticKernel.Abstractions/Services/AIServiceExtensions.cs
stephentoub: https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/

dotnet/src/InternalUtilities/src/Diagnostics/Verify.cs
stephentoub: It looks like I removed that usage, so I changed this to just `#if NET`.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatPromptParser.cs
```diff
@@ -30,7 +30,9 @@ public static bool TryParse(string prompt, [NotNullWhen(true)] out ChatHistory?
         // the text contains "<message", as that's required in any valid XML prompt.
         const string MessageTagStart = "<" + MessageTagName;
         if (prompt is not null &&
+#pragma warning disable CA2249 //'string.Contains' not available for netstandard2.0
```
RogerBarreto, line 33: can we use contains from #if NET here instead?

RogerBarreto (approved):
Some minor changes. LGTM

dotnet/samples/Concepts/Plugins/CreatePluginFromOpenAI_AzureKeyVault.cs
```diff
@@ -121,7 +121,7 @@ private async Task GetSecretFromAzureKeyVaultWithRetryAsync(Kernel kernel, Kerne
 internal sealed class OpenAIAuthenticationProvider(Dictionary<string, Dictionary<string, string>>? oAuthValues = null, Dictionary<string, string>? credentials = null)
 {
     private readonly Dictionary<string, Dictionary<string, string>> _oAuthValues = oAuthValues ?? [];
-#pragma warning disable CA1823 // TODO: Use credentials
+#pragma warning disable CA1823, RCS1213 // TODO: Use credentials
     private readonly Dictionary<string, string> _credentials = credentials ?? [];
 #pragma warning restore CA1823
```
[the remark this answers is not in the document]
stephentoub: I think that's fine in general. In this case, though, there's a TODO that's very specific to this field... there's these credentials being passed in but then not used. That feels bad.

dotnet/src/InternalUtilities/src/Http/HttpClientProvider.cs
```diff
@@ -66,7 +70,33 @@ protected override void Dispose(bool disposing)
...
+#else
+        private static HttpClientHandler CreateHandler()
+        {
+            return new HttpClientHandler()
+            {
+                // Check cert revocation
+                CheckCertificateRevocationList = true,
+            };
+        }
+#endif
```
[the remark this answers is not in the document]
stephentoub: It has a different return type in each case, and if it's changed to the base type, that will trigger an analyzer saying the return type can be made more derived.

markwallace-microsoft (approved):

PR #4428, opened 2023-12-26
Python: Supports better exceptions when Azure OpenAI content filtering is triggered

juliomenendez:
### Motivation and Context

When an Azure OpenAI content filter error is triggered the current `AIException` hides the details of that specific error. This makes it hard for the consumer to see a more detailed description of the error in their prompt or input.
Solves #4197

### Description

Detects when a content filter error is raised, then parses the meaningful information about it and raises a new `ContentFilterAIException` with that information ready for the consumer to use.

### Contribution Checklist


python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
```diff
@@ -0,0 +1,105 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Dict, Optional
+
+from semantic_kernel.connectors.ai.ai_exception import AIException
+
+
+class ContentFilterAIException(AIException):
```
eavanvalkenburg, line 10: is this specific to OpenAI, Azure OpenAI, or are their similar things in other LLM services? If not this file should be somewhere within the opena_ai folders.

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
```diff
@@ -0,0 +1,105 @@
...
+        """Initializes a new instance of the ContentFilterAIException class.
+
+        Arguments:
+            message {str} -- The error message.
+            param {str} -- The parameter that caused the error.
+            content_filter_code {ContentFilterCode} -- The error code specific to the content filter.
+            content_filter_result {Dict[str, ContentFilterResult]} -- The result of the content filter checks.
+            inner_exception {Exception} -- The inner exception.
+        """
+        super().__init__(AIException.ErrorCodes.ServiceError, message, inner_exception)
```
eavanvalkenburg, line 74: I would expect a new class of errors, because this is not the same as other ServiceErrors, that way the specifics of this implementation can be added, while other LLM's can also raise a error related to this, maybe we should create `BadContentError` or `IncompliantContentError`, something like that? Also the question is if dotnet already has something for this, and if so what terminology they use!

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
```diff
@@ -0,0 +1,105 @@
...
+from enum import Enum
+from typing import Any, Dict, Optional
+
+from semantic_kernel.connectors.ai.ai_exception import AIException
+
+
+class ContentFilterAIException(AIException):
+    """AI exception for an error from Azure OpenAI's content filter"""
+
+    class ContentFilterCodes(Enum):
```
eavanvalkenburg, line 13: I'm not the biggest fan of classes defined within other classes, is there a reason we need it here?

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
```diff
@@ -0,0 +1,105 @@
...
+    # The parameter that caused the error.
+    _param: str
+
+    # The error code specific to the content filter.
+    _content_filter_code: ContentFilterCodes
+
+    # The results of the different content filter checks.
+    _content_filter_result: Dict[str, ContentFilterResult]
+
+    def __init__(
```
eavanvalkenburg, line 57: could we keep the signature the same between AIException and ContentFilterAIException, maybe add `**kwargs` to AIException and then use that for this as well, the init of this is mostly called by internal code, so kwargs are fine I think.

eavanvalkenburg (changes requested):
Added some notes and thoughts on this, mostly related to keeping as much the same between AIException and this one.

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
juliomenendez: As of now, yes, this is specific to Azure OpenAI.

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
juliomenendez: Not fan of them neither :D I was following how `ErrorCodes` is defined inside `KernerException`. I'm going to pull it out.

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
juliomenendez: It doesn't look like dotnet has this type of error handled. Just like here in Python they just handle any error as `KernelException`

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
juliomenendez: Moved it to open_ai folder

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
juliomenendez: Added a new type of error code and using it on the new exception.

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
juliomenendez: Refactored those inner classes outside of the exception class.

python/semantic_kernel/connectors/ai/content_filter_ai_exception.py
juliomenendez: Good call. Moved all the logic to the exception and left the signature the same as AIException.

moonbox3 (approved):
LGTM.

python/semantic_kernel/connectors/ai/open_ai/exceptions/content_filter_ai_exception.py
```diff
@@ -0,0 +1,116 @@
...
+from enum import Enum
+from typing import Any, Dict
+
+from openai import BadRequestError
+
+from semantic_kernel.connectors.ai.ai_exception import AIException
+
+
+class ContentFilterResultSeverity(Enum):
+    High = "high"
```
eavanvalkenburg, line 13: Python considers these constants, so the names should be all caps: `HIGH = "high"`

python/semantic_kernel/connectors/ai/open_ai/exceptions/content_filter_ai_exception.py
```diff
@@ -0,0 +1,116 @@
...
+            severity=ContentFilterResultSeverity(
+                inner_error_results.get(
+                    "severity", ContentFilterResultSeverity.Safe.value
+                )
+            ),
+        )
+
+
+class ContentFilterCodes(Enum):
+    ResponsibleAIPolicyViolation = "ResponsibleAIPolicyViolation"
```
eavanvalkenburg, line 49: Same as before, format is ALL_CAPS

eavanvalkenburg (changes requested):
just some small changes to the Enum, otherwise looks good!

python/semantic_kernel/connectors/ai/open_ai/exceptions/content_filter_ai_exception.py
eavanvalkenburg: source: https://docs.python.org/3/howto/enum.html

juliomenendez: Fixed

python/semantic_kernel/connectors/ai/open_ai/exceptions/content_filter_ai_exception.py
juliomenendez: Fixed

eavanvalkenburg (approved):
LGTM

PR #4437, opened 2023-12-28
.Net: Function and Prompt Filters

dmytrostruk:
### Motivation and Context

This PR contains changes to introduce Function and Prompt Filters - the functionality, which is similar to hooks, but uses different registration model. Instead of attaching event handlers (e.g. `kernel.FunctionInvoking += MyHandler`), it is possible to implement middleware functionality in separate class by implementing `IFunctionFilter` interface for function execution middleware and `IPromptFilter` interface for prompt rendering middleware.

It is possible to register multiple function and prompt filters. The functionality is implemented near hooks, so both hooks and filters are supported.

In similar way as hooks, filters allow to:
1. Get and override function result value.
2. Get function information and metadata (including token usage).
3. Get and override kernel arguments.
4. Cancel function execution.
5. Get and override rendered prompt before sending it to AI.

Usage Example:
MyFunctionFilter.cs
```csharp
public class MyFunctionFilter : IFunctionFilter
{
    public void OnFunctionInvoked(FunctionInvokedContext context)
    {
        // Process data after function invocation
    }

    public void OnFunctionInvoking(FunctionInvokingContext context)
    {
        // Process data before function invocation
    }
}
```

Registration
```csharp
builder.Services.AddSingleton<IFunctionFilter, MyFunctionFilter>();
var kernel = builder.Build();
```

### Contribution Checklist


stephentoub (changes requested):

dotnet/docs/EXPERIMENTS.md
```diff
@@ -16,6 +16,7 @@ You can use the following diagnostic IDs to ignore warnings or errors for a part
 - SKEXP0002: Image services
 - SKEXP0003: Memory connectors
 - SKEXP0004: Kernel Events
+- SKEXP0005: Kernel Filters
```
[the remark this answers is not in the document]
dmytrostruk: cc: @markwallace-microsoft @matthewbolanos to clarify rollout strategy for events and filters.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptRenderedContext.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// If there are multiple filters registered, subsequent filters may
+    /// overwrite a value set by a previous filter. The final result is what will
+    /// be the prompt used by the system.
+    /// </remarks>
+    public string RenderedPrompt
+    {
+        get => this._renderedPrompt;
+        set
+        {
+            Verify.NotNull(value);
```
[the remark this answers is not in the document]
dmytrostruk: Thanks, fixed here and in `PromptRenderedEventArgs` as it contains the same logic now.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptRenderedContext.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Initializes a new instance of the <see cref="PromptRenderedContext"/> class.
+    /// </summary>
+    /// <param name="function">The <see cref="KernelFunction"/> with which this filter is associated.</param>
+    /// <param name="arguments">The arguments associated with the operation.</param>
+    /// <param name="renderedPrompt">The prompt that was rendered by the associated operation.</param>
+    public PromptRenderedContext(KernelFunction function, KernelArguments arguments, string renderedPrompt)
+        : base(function, arguments, metadata: null)
+    {
+        Verify.NotNull(renderedPrompt);
+        this._renderedPrompt = renderedPrompt;
```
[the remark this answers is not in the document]
dmytrostruk: Thanks, fixed here and in `PromptRenderedEventArgs` as it contains the same logic now.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -190,6 +201,24 @@ public CultureInfo Culture
     [Experimental("SKEXP0004")]
     public event EventHandler<PromptRenderedEventArgs>? PromptRendered;
 
+    /// <summary>
+    /// Contains collection of function filters.
+    /// </summary>
+    [Experimental("SKEXP0005")]
+    public IList<IFunctionFilter>? FunctionFilters =>
```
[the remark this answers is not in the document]
dmytrostruk: Thanks, yes, I agree. I don't think we need to expose filter properties in Kernel. I removed them (option b) from your comment).

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -54,6 +58,11 @@ public sealed class Kernel
 
         // Store the provided plugins. If there weren't any, look in DI to see if there's a plugin collection.
         this._plugins = plugins ?? this.Services.GetService<KernelPluginCollection>();
+
+        // Store the provided function and prompt filters if they were registered.
+        this._functionFilters = this.Services.GetServices<IFunctionFilter>().ToList();
+        this._promptFilters = this.Services.GetServices<IPromptFilter>().ToList();
+
```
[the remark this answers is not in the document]
dmytrostruk: Do you mean that `_functionFilters` and `_promptFilters` fields should be non-nullable in this case, or you mean that we need to store these fields as `IEnumerable<T>` instead of `IList<T>` and perform `ToList()` only when we need them? (e.g. access `Count` in `Clone` method).

dotnet/src/SemanticKernel.Abstractions/Filters/Function/IFunctionFilter.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Interface for filtering actions during function invocation.
+/// </summary>
+[Experimental("SKEXP0005")]
+public interface IFunctionFilter
```
RogerBarreto, line 11: Suggestion:

1. Have one interface for each event, so we comply with interface segregation and don't enforce all implementations to implement all the events (can be very common to just listen to one of them) (preferred)

2. Use a base class where we just need to override the functions that we actually want to implement

```csharp
public interface IFunctionInvokingFilter 
{
    void OnFunctionInvoking(FunctionInvokingContext context)
}

public interface IFunctionInvokedFilter 
{
    void OnFunctionInvoked(FunctionInvokedContext context)
}

class MyFunctionFilter : IFunctionInvokingFilter, IFunctionInvokedFilter
{
    ...
}
```

RogerBarreto: I think with this approach we also make those events asynchronous, which can be very useful for handling them on expensive I/O operations.

stephentoub: > I think with this approach we also make those events asynchronous, which can be very useful for handling them on expensive I/O operations.

If we do that (which I'm fine with), presumably we'd remove the events and just have this mechanism?  It's possible but non-idiomatic to have async events.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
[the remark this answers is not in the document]
stephentoub: I expect the most common case to optimize for here is no filters registered; if you use `ToArray` instead of `ToList`, it'll special-case empty if it can determine it's empty and return a singleton array in that case. ToList can't do that because it returns a mutable list.

dotnet/src/SemanticKernel.Abstractions/Filters/Function/IFunctionFilter.cs
dmytrostruk: > 1. Have one interface for each event, so we comply with interface segregation and don't enforce all implementations to implement all the events (can be very common to just listen to one of them) (preferred)

I agree with point on interface segregation, but on the other hand, we will force users to know about 4 interfaces (IFunctionInvokingFilter, IFunctionInvokedFilter, IPromptRenderingFilter, IPromptRenderedFilter) instead of just 2 interfaces (IFunctionFilter, IPromptFilter). If users don't need pre or post event, it's possible to leave it empty. But from usability perspective, having only 2 interfaces with simple naming should be more user-friendly. Same approach is used in ASP.NET filters:
image


> I think with this approach we also make those events asynchronous, which can be very useful for handling them on expensive I/O operations.

For asynchronous scenarios we can add separate interfaces - `IAsyncFunctionFilter`, `IAsyncPromptFilter`. User may implement specific filter based on scenario. Again, something that is implemented in similar way in ASP.NET: 
image

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -105,6 +114,8 @@ public sealed class Kernel
             PromptRendering = this.PromptRendering,
             PromptRendered = this.PromptRendered,
             _data = this._data is { Count: > 0 } ? new Dictionary<string, object?>(this._data) : null,
+            _functionFilters = this._functionFilters is { Count: > 0 } ? new List<IFunctionFilter>(this._functionFilters) : null,
+            _promptFilters = this._promptFilters is { Count: > 0 } ? new List<IPromptFilter>(this._promptFilters) : null,
```
SergeyMenshykh, line 118: Both `_functionFilters` and `_promptFilters` fields are initialized in the class constructor - line 110, which is executed before the execution control reaches these two lines - 117 and 118. Do we still need to overwrite them here?

SergeyMenshykh (approved):

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionFilterContext.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Base class with data related to function invocation.
+/// </summary>
+[Experimental("SKEXP0005")]
+public abstract class FunctionFilterContext
```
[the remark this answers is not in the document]
dmytrostruk: > These new XxContext classes mean we now effectively have two identical sets of these types, one for use with the interfaces and one for use with the event handlers. Can we just use the existing event args types for both?

If it's okay to use event args types in other functionality than event handlers, then I can re-use event args types, please let me know.

> (Or should we get rid of the events entirely and just have the interface-based approach?)

I think the goal is to have both approaches available, but I would double-check with @markwallace-microsoft regarding that.

dotnet/src/SemanticKernel.Core/Functions/PromptRenderingResult.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.SemanticKernel.Services;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Contains result after prompt rendering process.
+/// </summary>
+internal class PromptRenderingResult
```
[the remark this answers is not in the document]
dmytrostruk: Is it possible to introduce warning in all places where class can be `sealed` but it's not? Something like this:
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1852
image

In documentation it says that introduced version for this rule is .NET 7. I'm wondering if it's possible to enable it for .NET Standard.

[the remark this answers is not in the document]
stephentoub: > Is it possible to introduce warning in all places where class can be sealed but it's not?

Yes, you just need to use a .NET 7 or .NET 8 SDK... the target framework doesn't matter, only the SDK you're using.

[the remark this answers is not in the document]
Krzysztof318: Resharper/rider also has this code inspection in settings :)

dotnet/docs/EXPERIMENTS.md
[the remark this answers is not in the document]
markwallace-microsoft: @dmytrostruk Matthew suggested we should at least have one release with them as experimental so we can collect some feedback (e.g., from the person that did the hackathon project around events) before promoting them. During that time, we should have seperate experimental IDs. 

Additionally we will likely remove Events if we get confirmation that Filters addresses the requirement.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: Good catch, I removed these lines and also added check in unit-tests that filters are cloned correctly.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -133,27 +133,52 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
             cancellationToken.ThrowIfCancellationRequested();
 
             // Invoke pre-invocation event handler. If it requests cancellation, throw.
-            if (kernel.OnFunctionInvoking(this, arguments)?.Cancel is true)
+            var invokingEventArgs = kernel.OnFunctionInvoking(this, arguments);
+
+            // Invoke pre-invocation filter. If it requests cancellation, throw.
+            var invokingContext = kernel.OnFunctionInvokingFilter(this, arguments);
```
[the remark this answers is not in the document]
dmytrostruk: Actually, this is a correct place of `invokingContext`, because if I move it a little bit lower and event hook will trigger cancellation, the pre-invocation filter won't be triggered in this case, and I believe it's incorrect approach. In current implementation, we trigger both event hook and filter, and only after that we use event args and context objects to process result:
image

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionFilterContext.cs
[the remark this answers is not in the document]
stephentoub: If we have both, I think it's fine to use the EventArgs for both. But it sounds like the intent is to get rid of the events.

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvokedContext.cs
```diff
@@ -0,0 +1,45 @@
...
+[Experimental("SKEXP0005")]
+public sealed class FunctionInvokedContext : FunctionFilterContext
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="FunctionInvokedContext"/> class.
+    /// </summary>
+    /// <param name="function">The <see cref="KernelFunction"/> with which this filter is associated.</param>
+    /// <param name="arguments">The arguments associated with the operation.</param>
+    /// <param name="result">The result of the function's invocation.</param>
+    public FunctionInvokedContext(KernelFunction function, KernelArguments arguments, FunctionResult result)
```
stephentoub, line 20: Should we validate that the function result is for the specified function? Or if we can extract the function from the function result, does this even need to take the function separately?

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptFilterContext.cs
```diff
@@ -0,0 +1,44 @@
...
+[Experimental("SKEXP0005")]
+public abstract class PromptFilterContext
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PromptFilterContext"/> class.
+    /// </summary>
+    /// <param name="function">The <see cref="KernelFunction"/> with which this filter is associated.</param>
+    /// <param name="arguments">The arguments associated with the operation.</param>
+    /// <param name="metadata">A dictionary of metadata associated with the operation.</param>
+    internal PromptFilterContext(KernelFunction function, KernelArguments arguments, IDictionary<string, object?>? metadata)
```
stephentoub, line 20: Shouldn't these be readonly dictionaries? Or is the intent that the handler can modify/augment the metadata?

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptRenderedContext.cs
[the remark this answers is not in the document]
stephentoub: This doesn't appear to have been fixed

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -133,27 +133,52 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
...
+
+            if (invokingEventArgs?.Cancel is true)
             {
                 throw new OperationCanceledException($"A {nameof(Kernel)}.{nameof(Kernel.FunctionInvoking)} event handler requested cancellation before function invocation.");
             }
 
+            if (invokingContext?.Cancel is true)
```
stephentoub, line 146: I find it strange to have both systems and they don't interact, eg an event handler can cancel cancelation by setting the property back to false, but a filter can't.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -32,6 +32,10 @@ public sealed class Kernel
     private CultureInfo _culture = CultureInfo.InvariantCulture;
     /// <summary>The collection of plugins, initialized via the constructor or lazily-initialized on first access via <see cref="Plugins"/>.</summary>
     private KernelPluginCollection? _plugins;
+    /// <summary>Contains collection of function filters.</summary>
+    private readonly IList<IFunctionFilter>? _functionFilters;
+    /// <summary>Contains collection of prompt filters.</summary>
+    private readonly IList<IPromptFilter>? _promptFilters;
```
stephentoub, line 38: With how you've written the initialization code, aren't these always non-null arrays rather than nullable ilists?

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvokedContext.cs
dmytrostruk: All context classes were implemented in exactly the same way as event args. I think if we apply this type of modifications to Context classes, we should probably do the same for event args.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptRenderedContext.cs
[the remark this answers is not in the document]
dmytrostruk: My bad, I forgot to assign it to property. But now I changed the logic a little, I removed `_renderedPrompt` field completely, use just property and move validation logic to constructor.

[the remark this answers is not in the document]
dmytrostruk: My bad, I forgot to assign it to property. Resolved.

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvokedContext.cs
dmytrostruk: Taking into account that function can be accessed from `FunctionResult` object, I removed `function` parameter in `FunctionInvokedContext` constructor. I leave it as it is in EventArgs though, to avoid changing public API.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptFilterContext.cs
dmytrostruk: Good catch, updated!

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: Makes sense, updated them to non-null arrays.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: So, if we use only event handlers or only filters, it is possible to cancel cancellation. But if we cancel cancellation in event handler, but we want to proceed with cancellation in filter - it's not going to work, and I'm not sure if it's a goal to be honest. I think the primary usage pattern should be to use only event handlers or only filters, based on personal preference, but not combine them together.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -318,6 +327,82 @@ public CultureInfo Culture
     }
     #endregion
 
+    #region Internal Filtering
```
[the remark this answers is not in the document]
dmytrostruk: It's good question whether it should be possible to add/remove events after Kernel initialization. I think the goal is to configure Kernel (e.g. in `Startup.cs`), add necessary filters and receive a call on specific event. But if we want to add/remove filters in the middle of runtime, then it's definitely not supported at the moment. 

If this is a requirement, we can add a possibility to add filter with something like this `kernel.FunctionFilters.Add(filter)` and `kernel.PromptFilters.Add(filter)`. But I'm wondering if it's useful at all. In theory, filters can be complex, include logger factory and other services to track function execution, and usually you want to configure these things on application startup only once. But maybe I'm missing specific use-case, please let me know.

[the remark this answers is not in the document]
stephentoub: > It's good question whether it should be possible to add/remove events after Kernel initialization.

My point is that you _can_ currently add/remove events after Kernel initialization... in fact that's the only way you can do it. But you can't for filters.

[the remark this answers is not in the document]
dmytrostruk: In general, it would be interesting to learn which way of event initialization is better in terms of usability - before Kernel initialization or after. For example, if we have ASP.NET app, controller and action, where exactly we attach an event, in controller constructor or action? Should we detach it at the end of action execution? Those usage patterns will allow us to understand with which approach we should stay in long-term.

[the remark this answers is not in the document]
stephentoub: > Those usage patterns will allow us to understand with which approach we should stay in long-term.

Yes, but how will we determine that from this experiment? You can't register an event handler pre construction and you can't register a filter post construction. So if someone is using the events, are they doing so because it's more natural for their use case, or because they needed to register it post construction. And if someone is using the filters, are they doing so because it's more natural for their use case, or because the needed to register it pre construction?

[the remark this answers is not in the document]
matthewbolanos: If I understand @stephentoub's points, it would be that we should _completely_ remove events (or deprecate them) when we add filters to make sure folks yell at us now in case they want to add events after kernel construction. Is that correct?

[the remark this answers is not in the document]
stephentoub: Yes, _if_ that's how we intend to actually ship. Basically let's make it now how we believe it should be. If we believe long term we want both events and filters, then leave it as this PR has it. If we believe long term we only want filters, then I'd recommend removing the events now. Otherwise, it's harder for us to learn whether what we're shipping is sufficient.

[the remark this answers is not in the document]
dmytrostruk: @stephentoub As part of this PR, I'm going to add a possibility to add filters to Kernel post construction, in order to cover case from events. Instead of removing events now, can we mark them as Obsolete to do a smooth rollout and avoid breaking changes?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: Resolving this as events will be removed in the future, so no need to make both approaches interact with each other.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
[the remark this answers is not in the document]
dmytrostruk: @stephentoub In this commit https://github.com/microsoft/semantic-kernel/pull/4437/commits/8ca9896fc64d1543e088506fe8b50621d777ef73 I've added collections for function and prompt filters and the ability to mutate filter lists after kernel registration. You can also insert filter in the middle of filter list and the order of filter execution will be updated as well. I've added related unit tests for new cases. Please let me know what you think about it.

dotnet/src/SemanticKernel.Abstractions/Events/PromptRenderedEventArgs.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.SemanticKernel;
 [Experimental("SKEXP0004")]
 public sealed class PromptRenderedEventArgs : CancelKernelEventArgs
 {
-    private string _renderedPrompt;
+    private string? _renderedPrompt;
```
stephentoub, line 13: Why is this nullable?

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionFilterCollection.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Provides a collection of <see cref="IFunctionFilter"/> instances.
+/// </summary>
+[Experimental("SKEXP0005")]
+public sealed class FunctionFilterCollection : IList<IFunctionFilter>, IReadOnlyList<IFunctionFilter>
```
stephentoub, line 14: This shouldn't be public. It should only be surfaced as an `IList<IFunctionFilter>`.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptFilterCollection.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Provides a collection of <see cref="IPromptFilter"/> instances.
+/// </summary>
+[Experimental("SKEXP0005")]
+public sealed class PromptFilterCollection : IList<IPromptFilter>, IReadOnlyList<IPromptFilter>
```
stephentoub, line 14: Same here. This should be internal.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptRenderedContext.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Class with data related to prompt after rendering.
+/// </summary>
+[Experimental("SKEXP0005")]
+public sealed class PromptRenderedContext : PromptFilterContext
+{
+    private string? _renderedPrompt;
```
stephentoub, line 13: Why is this nullable?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -67,6 +72,22 @@ public sealed class Kernel
...
             }
         }
+
+        // Enumerate any function filters that may have been registered.
+        IEnumerable<IFunctionFilter> functionFilters = this.Services.GetServices<IFunctionFilter>();
+
+        if (functionFilters is not ICollection<IFunctionFilter> functionFilterCollection || functionFilterCollection.Count != 0)
+        {
+            this._functionFilters = new(functionFilters);
+        }
```
stephentoub, line 82: Nice

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -116,6 +137,24 @@ public sealed class Kernel
         Interlocked.CompareExchange(ref this._plugins, new KernelPluginCollection(), null) ??
         this._plugins;
 
+    /// <summary>
+    /// Gets the collection of function filters available through the kernel.
+    /// </summary>
+    [Experimental("SKEXP0005")]
+    public FunctionFilterCollection FunctionFilters =>
```
stephentoub, line 144: IList

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -116,6 +137,24 @@ public sealed class Kernel
...
+        this._functionFilters ??
+        Interlocked.CompareExchange(ref this._functionFilters, new FunctionFilterCollection(), null) ??
+        this._functionFilters;
+
+    /// <summary>
+    /// Gets the collection of function filters available through the kernel.
+    /// </summary>
+    [Experimental("SKEXP0005")]
+    public PromptFilterCollection PromptFilters =>
```
stephentoub, line 153: IList

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -318,6 +357,82 @@ public CultureInfo Culture
     }
     #endregion
 
+    #region Internal Filtering
+
+    [Experimental("SKEXP0005")]
```
stephentoub, line 362: I think we should continue using the same experimental ID as for events. It's the same feature, just evolved with a new shape. Anyone currently using events will need to switch to the new syntax and shouldn't need to change the ID to do so.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -318,6 +357,82 @@ public CultureInfo Culture
...
+    internal FunctionInvokingContext? OnFunctionInvokingFilter(KernelFunction function, KernelArguments arguments)
+    {
+        FunctionInvokingContext? context = null;
+
+        if (this._functionFilters is { Count: > 0 })
+        {
+            context = new(function, arguments);
+
+            foreach (var filter in this._functionFilters)
```
stephentoub, line 371: Nit: this week allocate an enumerator. You'll avoid that if you instead use a for loop.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -318,6 +357,82 @@ public CultureInfo Culture
...
+    [Experimental("SKEXP0005")]
+    internal FunctionInvokedContext? OnFunctionInvokedFilter(KernelArguments arguments, FunctionResult result)
+    {
+        FunctionInvokedContext? context = null;
+
+        if (this._functionFilters is { Count: > 0 })
+        {
+            context = new(arguments, result);
+
+            foreach (var filter in this._functionFilters)
```
stephentoub, line 389: Do we care about the order in which we process the filters?

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -296,11 +305,30 @@ private async Task<(IAIService, PromptExecutionSettings?, string, PromptRendered
...
             {
-                this._logger.LogTrace("Rendered prompt changed by handler: {Prompt}", renderedEventArgs.RenderedPrompt);
+                this._logger.LogTrace("Rendered prompt changed by event handler: {Prompt}", renderedEventArgs.RenderedPrompt);
+            }
+        }
+
+        var renderedContext = kernel.OnPromptRenderedFilter(this, arguments, renderedPrompt);
+
+        if (renderedContext is not null &&
+            renderedContext.Cancel is false &&
```
stephentoub, line 315:
```suggestion
           !renderedContext.Cancel &&
```
?

stephentoub (commented):
Almost there. :)

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptRenderedContext.cs
dmytrostruk: @stephentoub Constructor still shows a warning that `_renderedPrompt` is not initialized:
image

stephentoub: You need to say what member isn't null in the attribute. I'm surprised that even compiles without it.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: Makes sense, taking into account that in follow-up PR I will mark events as Obsolete, I think we can use same ID for filters. I updated the code.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptRenderedContext.cs
dmytrostruk: Fixed by placing attribute on property setter.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
dmytrostruk: Yeah, I put it in similar way as for event args, I think maybe previously this property was nullable, not sure. I simplified in both places now :)

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: I think by default it should be processed in the same order as filters were initially registered/added. But now there is possibility to access `kernel.FunctionFilters` and `kernel.PromptFilters`, so it's possible to change the order programmatically if needed.

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptFilterCollection.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Provides a collection of <see cref="IPromptFilter"/> instances.
+/// </summary>
+[Experimental("SKEXP0004")]
+internal sealed class PromptFilterCollection : IList<IPromptFilter>, IReadOnlyList<IPromptFilter>
```
stephentoub, line 14: We could just have a:
```C#
internal sealed class NonNullItemsCollection<T> : IList<T>, IReadOnlyList<T>
{
   ...
}
```
rather than having both FunctionFilterCollection and PromptFilterCollection.

stephentoub (approved):

dotnet/src/SemanticKernel.Abstractions/Filters/Prompt/PromptFilterCollection.cs
dmytrostruk: Makes sense, thank you. I added generic collection for non-null items.

SergeyMenshykh (approved):

PR #4469, opened 2024-01-03
Java: Add code evaluation blocks

johnoliver:
### Contribution Checklist


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
```diff
@@ -64,12 +69,18 @@ KernelFunction createFunctionFromPrompt(
...
+
+
     interface Builder extends SemanticKernelBuilder<Kernel> {
 
         <T extends AIService> Builder withDefaultAIService(Class<T> clazz, T aiService);
 
         Builder withPromptTemplateEngine(PromptTemplate promptTemplate);
 
-        Builder withFunction(KernelFunction kernelFunction);
+        Builder withPlugins(KernelPlugin searchPlugin);
```
dsgrieve, line 84: s/withPlugins/withPlugin/

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/plugin/KernelPluginCollection.java
```diff
@@ -1,21 +1,48 @@
...
     @Override
     public Iterator<KernelPlugin> iterator() {
-        // TODO Auto-generated method stub
-        throw new UnsupportedOperationException("Unimplemented method 'iterator'");
+        return plugins.values().iterator();
     }
 
+    public void add(KernelPlugin plugin) {
+        if (plugins.containsKey(plugin.getName())) {
+            throw new IllegalArgumentException("Plugin " + plugin.getName() + " already exists");
```
dsgrieve, line 43: Log at debug level and let it overwrite

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/semanticfunctions/KernelPromptTemplateFactory.java
```diff
@@ -1,9 +1,21 @@
...
+import com.microsoft.semantickernel.templateengine.handlebars.HandlebarsPromptTemplate;
+import com.microsoft.semantickernel.templateengine.semantickernel.DefaultPromptTemplate;
+import java.util.Locale;
+
+public class KernelPromptTemplateFactory implements PromptTemplateFactory {
 
     @Override
     public PromptTemplate tryCreate(PromptTemplateConfig templateConfig) {
-        return null;
+        switch (templateConfig.getTemplateFormat().toLowerCase(Locale.ROOT)) {
```
dsgrieve, line 11: Move this to PromptTemplateConfig as a static method and get rid of the factory. 

In dotnet, they have "boolean tryCreate" which doesn't work in java. Maybe this should throw an "UnknownTemplateException"

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/KernelFunction.java
```diff
@@ -61,6 +61,8 @@ public interface KernelFunction extends Buildable {
      */
     String toManualString(boolean includeOutputs);
 
+    KernelFunctionMetadata getMetadata();
+
     @Deprecated
     default Class<?> getType() {
```
dsgrieve, line 67: I think we should have a `ContextVariable<?> getReturnType()`, or some such. I could create a function and supply its converter instead of having to add a convert to the defaults. `Class<?> getType()` would be revived. getReturnType would have a default impl `ContextVariableTypes.getDefaultVariableTypeForClass(getType())` 
Something like that.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/plugin/KernelPluginCollection.java
```diff
@@ -1,21 +1,48 @@
...
-import com.microsoft.semantickernel.plugin.KernelPlugin;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 
 public class KernelPluginCollection implements Iterable<KernelPlugin> {
 
+    private final Map<String, KernelPlugin> plugins = new HashMap<>();
```
dsgrieve, line 12: Somewhere else we have a collection that implements Map. Why should this be different? Any collection class we have should have a common interface.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/semanticfunctions/KernelFunctionFromPrompt.java
```diff
@@ -286,7 +287,17 @@ public KernelFunction build() {
             } else if (promptTemplateFactory != null) {
                 temp = promptTemplateFactory.tryCreate(config);
             } else {
-                temp = new HandlebarsPromptTemplate(config);
+                switch (config.getTemplateFormat()) {
```
dsgrieve, line 290: If PromptTemplateConfig had a static tryCreate, you wouldn't need this.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/templateengine/semantickernel/CodeTokenizer.java
```diff
@@ -0,0 +1,232 @@
...
+import com.microsoft.semantickernel.templateengine.semantickernel.blocks.Block;
+import com.microsoft.semantickernel.templateengine.semantickernel.blocks.FunctionIdBlock;
+import com.microsoft.semantickernel.templateengine.semantickernel.blocks.Symbols;
+import com.microsoft.semantickernel.templateengine.semantickernel.blocks.ValBlock;
+import com.microsoft.semantickernel.templateengine.semantickernel.blocks.VarBlock;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/// <summary>
```
dsgrieve: Seems like we should just have the BNF and autogen the tokenizer.

dsgrieve (approved):

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/semanticfunctions/KernelPromptTemplateFactory.java
johnoliver: Unfortunately removing this factory would  have quite a few impacts that would move us away from .net

Issue #4418, opened 2023-12-24
.Net: Semantic Kernel Exception when running TextMemory Plugin Sample


AshD:
**Describe the bug**
Running TextMemory Plugin Sample
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example15_TextMemoryPlugin.cs

in PART 4: TextMemoryPlugin Recall in a Prompt Function throws 
`Microsoft.SemanticKernel.KernelException: 'A function named argument must contain a name and value separated by a '=' character.'`

In line  
 `var aboutMeOracle = kernel.CreateFunctionFromPrompt(RecallFunctionDefinition, new OpenAIPromptExecutionSettings() { MaxTokens = 100 });`

Has the syntax for defining the Recall function changed? The below is the one from the sample.
```
            const string RecallFunctionDefinition = @"
Consider only the facts below when answering questions:

BEGIN FACTS
About me: {{recall='where did I grow up?'}}
About me: {{recall='where do I live now?'}}
END FACTS

Question: {{$input}}

Answer:
";
```

I am on dotnet 1.01 release and Kernel memory 1.0.1-alpha

Thanks,
Ash

glahaye:
Looking at this...

glahaye:
It turns out the CodeTokenizer routinely causes first-chance (caught) exceptions to determine whether tokens belong to function names of named argument blocks. This is inefficient but also can be startling depending on how one has set one's configuration for breaking on exceptions.

See PR #4471 for fix

PR #4471, opened 2024-01-03
.Net: Avoid causing first-chance exceptions in CodeTokenizer

glahaye:
### Motivation and Context
The CodeTokenizer routinely causes first-chance (caught) exceptions to determine whether tokens belong to function names of named argument blocks. This is inefficient but also can be startling depending on how one has set one's configuration for breaking on exceptions.

See #4418 

### Description
CodeTokenizer now uses a new method from NamedArgBlock to determine whether a string could result in a NamedArgBlock instead of resorting to causing exceptions to gather the same information.

### Contribution Checklist


crickman (approved):

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -61,6 +56,22 @@ public NamedArgBlock(string? text, ILoggerFactory? logger = null)
...
     }
 
+    /// <summary>
+    /// Attempts to extract the name and value of a named argument block from a string
+    /// </summary>
+    /// <param name="text">String from which to extract a name and value</param>
+    /// <param name="argParts">This array has its first and second elements set to the extracted name and value when successful.
+    /// An empty array is returned otherwise.</param>
+    /// <returns>true when a name and value are successfully extracted from the given text</returns>
+    internal static bool TryGetNameAndValue(string? text, out string[] argParts)
```
markwallace-microsoft, line 66: Consider defining this as `bool TryGetNameAndValue(string? text, out string name, out string value)`

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/NamedArgBlock.cs
```diff
@@ -61,6 +56,22 @@ public NamedArgBlock(string? text, ILoggerFactory? logger = null)
...
+    {
+        text ??= string.Empty;
```
markwallace-microsoft, line 68: If text is null or empty should just return false

markwallace-microsoft (commented):
Few minor comments

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/NamedArgBlock.cs
glahaye: I like it

dotnet/src/SemanticKernel.Core/TemplateEngine/Blocks/NamedArgBlock.cs
glahaye: I like this too!

markwallace-microsoft (approved):

PR #4473, opened 2024-01-04
.Net: Stable Diffusion img2img implementation

glahaye:
### Motivation and Context
Stable Diffusion img2img implementation

### Description
Stable Diffusion img2img implementation

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.StableDiffusion/ImageToImage/StableDiffusionImageToImageService.cs
+++ b/dotnet/src/Connectors/Connectors.StableDiffusion/ImageToImage/StableDiffusionImageToImageService.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.AI.ImageToImage;
+
+namespace Microsoft.SemanticKernel.Connectors.StableDiffusion.ImageToImage;
+
+/// <summary>
+/// Stable Diffusion implementation of image transformation (image to image service)
+/// </summary>
+public class StableDiffusionImageToImageService : IImageToImageService
+{
+    private const string EngineId = "stable-diffusion-xl-1024-v1-0"; // TODO: make this configurable
+    private const string ImageToImageUri = $"https://api.stability.ai/v1/generation/{EngineId}/image-to-image";
+
+    private readonly string _apiKey;
+    private readonly HttpClient _httpClient;
+
+    /// <summary>
+    /// Constructor
+    /// </summary>
+    public StableDiffusionImageToImageService(string apiKey, HttpClient? httpClient = null)
+    {
+        this._apiKey = apiKey;
+        this._httpClient = httpClient ?? new HttpClient();
+        this._httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", this._apiKey);
+        this._httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("image/png"));
+    }
+
+    /// <inheritdoc />
+    public async Task<byte[]> GenerateModifiedImageAsync(byte[] inputFile, string prompt, int width, int height, Kernel? kernel = null, CancellationToken cancellationToken = default)
+    {
+        using var formData = new MultipartFormDataContent();
+        using var stringContent = new StringContent(prompt);
+        using var byteArrayContent = new ByteArrayContent(inputFile);
+
+        formData.Add(stringContent, "text_prompts[0][text]");
+        formData.Add(byteArrayContent, "init_image");
+
+        HttpResponseMessage response = await this._httpClient.PostAsync(ImageToImageUri, formData, cancellationToken).ConfigureAwait(false);
+        response.EnsureSuccessStatusCode();
+
+        return await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+    }
+}
--- a/dotnet/src/Connectors/Connectors.StableDiffusion/StableDiffusionServiceCollectionExtensions.cs
+++ b/dotnet/src/Connectors/Connectors.StableDiffusion/StableDiffusionServiceCollectionExtensions.cs
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+using System.Net.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.SemanticKernel.AI.ImageToImage;
 using Microsoft.SemanticKernel.Connectors.StableDiffusion.ImageToImage;
-using System.Net.Http;
 
 namespace Microsoft.SemanticKernel.Connectors.StableDiffusion;
 
--- a/dotnet/src/Connectors/Connectors.StableDiffusion/StableDiffusionServiceCollectionExtensions.cs
+++ b/dotnet/src/Connectors/Connectors.StableDiffusion/StableDiffusionServiceCollectionExtensions.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.SemanticKernel.AI.ImageToImage;
+using Microsoft.SemanticKernel.Connectors.StableDiffusion.ImageToImage;
+using System.Net.Http;
+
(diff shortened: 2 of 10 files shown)


dotnet/src/SemanticKernel.Abstractions/AI/ImageToImage/IImageToImageService.cs
```diff
@@ -0,0 +1,32 @@
...
+
+/// <summary>
+/// Interface for image transformation (image to image)
+/// </summary>
+public interface IImageToImageService
+{
+    /// <summary>
+    /// Modify an input PNG image using a given prompt.
+    /// </summary>
+    /// <param name="inputFile">Input PNG image's bytes</param>
```
stephentoub, line 16: Every implementation of this interface will work with PNGs only?

dotnet/src/Connectors/Connectors.StableDiffusion/ImageToImage/StableDiffusionImageToImageService.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.AI.ImageToImage;
+
+namespace Microsoft.SemanticKernel.Connectors.StableDiffusion.ImageToImage;
+
+/// <summary>
+/// Stable Diffusion implementation of image transformation (image to image service)
+/// </summary>
+public class StableDiffusionImageToImageService : IImageToImageService
```
stephentoub, line 14: sealed

dotnet/src/Connectors/Connectors.StableDiffusion/ImageToImage/StableDiffusionImageToImageService.cs
```diff
@@ -0,0 +1,48 @@
...
+    private readonly string _apiKey;
+    private readonly HttpClient _httpClient;
+
+    /// <summary>
+    /// Constructor
+    /// </summary>
+    public StableDiffusionImageToImageService(string apiKey, HttpClient? httpClient = null)
+    {
+        this._apiKey = apiKey;
+        this._httpClient = httpClient ?? new HttpClient();
```
stephentoub, line 28: This should use a singleton HttpClient if none is provided. There's a shared file that provides what's needed. See how the OpenAI one does it.

dotnet/src/Connectors/Connectors.StableDiffusion/ImageToImage/StableDiffusionImageToImageService.cs
```diff
@@ -0,0 +1,48 @@
...
+        this._httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", this._apiKey);
+        this._httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("image/png"));
```
stephentoub, line 30: This should not use DefaultRequestHeaders. If someone supplied their own HttpClient, this would be adding headers to their client, which is both unexpected and buggy.

dotnet/src/Connectors/Connectors.StableDiffusion/ImageToImage/StableDiffusionImageToImageService.cs
```diff
@@ -0,0 +1,48 @@
...
+    public async Task<byte[]> GenerateModifiedImageAsync(byte[] inputFile, string prompt, int width, int height, Kernel? kernel = null, CancellationToken cancellationToken = default)
+    {
+        using var formData = new MultipartFormDataContent();
+        using var stringContent = new StringContent(prompt);
+        using var byteArrayContent = new ByteArrayContent(inputFile);
+
+        formData.Add(stringContent, "text_prompts[0][text]");
+        formData.Add(byteArrayContent, "init_image");
+
+        HttpResponseMessage response = await this._httpClient.PostAsync(ImageToImageUri, formData, cancellationToken).ConfigureAwait(false);
```
stephentoub, line 43: using

dotnet/src/Connectors/Connectors.StableDiffusion/StableDiffusionServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        string apiKey,
+        string? serviceId = null,
+        HttpClient? httpClient = null)
+    {
+        Verify.NotNull(builder);
+        Verify.NotNullOrWhiteSpace(apiKey);
+
+        builder.Services.AddKeyedSingleton<IImageToImageService>(serviceId, (serviceProvider, _) =>
+        {
+            return new StableDiffusionImageToImageService(apiKey, httpClient);
```
stephentoub, line 29: This should use the httpClient if it's not null, and otherwise fall back to trying to get an HttpClient from the serviceProvider. See how the OpenAI implementation does it for an example.

dotnet/src/Connectors/Connectors.StableDiffusion/StableDiffusionServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        this IServiceCollection services,
+        string apiKey,
+        string? serviceId = null)
+    {
+        Verify.NotNull(services);
+        Verify.NotNullOrWhiteSpace(apiKey);
+
+        return services.AddKeyedSingleton<IImageToImageService>(serviceId, (serviceProvider, _) =>
+        {
+            return new StableDiffusionImageToImageService(apiKey);
```
stephentoub, line 48: Similarly this should pull the HttpClient to use from the serviceProvider.

dotnet/src/SemanticKernel.Abstractions/AI/ImageToImage/ImageToImageExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <param name="height">Output image height in pixels</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Generated image in PNG format</returns>
+    public static async Task<byte[]> GenerateModifiedImageAsync(this IImageToImageService service, string inputFilePath, string prompt, int width, int height, Kernel? kernel = null, CancellationToken cancellationToken = default)
+    {
+        byte[] inputFile = File.ReadAllBytes(inputFilePath);
+
+        return await service.GenerateModifiedImageAsync(inputFile, prompt, width, height, kernel, cancellationToken).ConfigureAwait(false);
+    }
```
stephentoub, line 52: A few concerns about these:
- It's creating a new HttpClient on every use.
- It's not providing the caller the ability to supply their own HttpClient.
- It's not using the HttpClient that may have already been configured in the IImageToImageService, yet it's an extension off of that IImageToImageService.
- The two overloads are identical except for one parameter that's either a Uri or a string, which usually means both the Uri and the string represent a uri, but here the string is a file path.

I suggest not having these methods at all.

dotnet/src/SemanticKernel.Abstractions/AI/ImageToImage/IImageToImageService.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <param name="prompt">Prompt describing how to modify the output image</param>
+    /// <param name="width">Output image width in pixels</param>
+    /// <param name="height">Output image height in pixels</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Generated image in PNG format</returns>
+    public Task<byte[]> GenerateModifiedImageAsync(
+        byte[] inputFile,
```
stephentoub, line 24: Every implementation takes the input as a byte[]? Are there implementations that might accept a url rather than the actual data?

dotnet/src/SemanticKernel.Abstractions/AI/ImageToImage/IImageToImageService.cs
```diff
@@ -0,0 +1,32 @@
...
+        string prompt,
+        int width,
+        int height,
+        Kernel? kernel = null, // TODO: Is kernel even needed?
```
stephentoub, line 28: Part of the design is that the kernel is passed to everything, in case that thing needs something from it, e.g. kernel.Data provides the ability to smuggle in/out arbitrary information. So, yes, I'd keep this.

Issue #4470, opened 2024-01-03
.Net: ToolCallBehavior = ToolCallBehavior.RequireFunction(wikiFunction, true) seems to be broken


markwallace-microsoft:
**Describe the bug**
Also
ToolCallBehavior = ToolCallBehavior.RequireFunction(wikiFunction, true)
seems to be broken (happens for every function). This makes the prompt result come back empty all the time (automatic function calling does not work). Without this (just enable auto calls) it works fine, calling the function automatically and returns proper results.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Platform**
 - OS: [e.g. Windows, Mac]
 - IDE: [e.g. Visual Studio, VS Code]
 - Language: [e.g. C#, Python]
 - Source: [e.g. NuGet package version 0.1.0, pip package version 0.1.0, main branch of repository]

**Additional context**

stephentoub:
Repro?

kboom:
1. Import a plugin with a single function in it into the kernel instance.
2. Extract this function so that it can be passed to the relevant require method. Not sure if this is a proper way, but I could not find any other way: `kernel.Plugins[metadata.Name][nameof(MyType.MyFunc)].Metadata.ToOpenAIFunction()` - this returns a function

Then
```C#
var executionSettings = new OpenAIPromptExecutionSettings()
 {
     MaxTokens = 1000,
     Temperature = 0.9,
     ToolCallBehavior = ToolCallBehavior.RequireFunction(wikiFunction, true) // when this is changed to simply `ToolCallBehavior.AutoInvokeKernelFunctions` it works FINE, and it DOES call the function required
 };

var result = await kernel.InvokePromptAsync(
    _promptTemplate,
    arguments,
    HandlebarsPromptTemplateFactory.HandlebarsTemplateFormat,
    _promptFactory,
    cancellationToken).ConfigureAwait(false);

var troubleshootingGuideString = result.GetValue<string>();
```

The outcome is that the prompt response ALWAYS comes back empty. When the tool call behavior is not using required function but auto invoke it works fine (and still calls the function).

stephentoub:
Thanks. That was enough to help me narrow down the problem.
https://github.com/microsoft/semantic-kernel/pull/4474

PR #4474, opened 2024-01-04
.Net: Fix handling of required calls with OpenAI

stephentoub:
The {Azure} OpenAI service allows a tool to be specified as required, in which case it'll always respond with an invocation of that tool; essentially it's a way to get the service to populate the arguments to a method. However, when doing so, the service also always immediately completes the chat with a stop finish reason; it apparently assumes its job is done because it sent back the required function information. That means we'd stop processing the function calls because we were looking for a finish reason of tool call.  This changes it to invoke functions if any were specified, regardless of the finish reason.

Fixes https://github.com/microsoft/semantic-kernel/issues/4470


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -572,7 +577,7 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
 
             if (iteration >= chatExecutionSettings.ToolCallBehavior!.MaximumUseAttempts)
             {
-                chatOptions.Tools.Clear();
+                // Set the tool choice to none. We'd also like to clear the tools, but doing so can make the service unhappy ("[] is too short - 'tools'").
                 chatOptions.ToolChoice = ChatCompletionsToolChoice.None;
```
RogerBarreto, line 581: I had this issue in an old implementation, the way around it was adding one fake function called (DontCallMe) with the minimum amount of tokens, to avoid sending unneeded functions back to the API.

stephentoub: I considered that. My concern was that the service might end up caring if a tool mentioned in the history doesn't actually exist in the tools collection sent.  If we know for certain that it doesn't, we could do the same thing.

SergeyMenshykh (approved):

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
stephentoub: Let's deal with optimizing that separately and get the fix merged for now.

markwallace-microsoft (approved):

PR #4485, opened 2024-01-04
Python: Handle errors if no event loop running in current thread.

moonbox3:
### Motivation and Context

If invoking multiple functions, there may not be an event loop running in the current thread. The check for the event loop can throw an error, which is seen when trying to run functions in a loop, as an example.

### Description

Adding a try catch to handle an errors when checking for this event loop.

### Contribution Checklist


juliomenendez (approved):

eavanvalkenburg (approved):
LGTM

PR #4486, opened 2024-01-05
.Net - Add support for instruction templating on Agents

crickman:
### Motivation and Context

1. Agent instructions support template parameters 
2. Agent yaml identical / interchangeable with prompt config yaml

### Description

Templating agents introduces much more dynamic capabilities, beyond the standard assistant API.

As part of this, the agent template and kernel-function template have converged.

### Contribution Checklist


dotnet/src/Experimental/Agents/Internal/Agent.cs
```diff
@@ -181,6 +196,12 @@ private AgentPluginImpl DefinePlugin()
         return new AgentPluginImpl(this, functionAsk);
     }
 
+    private IPromptTemplate DefinePromptTemplate(PromptTemplateConfig config)
+    {
+        var factory = new KernelPromptTemplateFactory(); // $$$ HANDLEBARS
```
glahaye, line 201: Should something happen with this comment?

dotnet/samples/KernelSyntaxExamples/Example70_Agents.cs
```diff
@@ -33,7 +33,7 @@ public static async Task RunAsync()
             return;
         }
 
-        // "Hello agent"
+        //// "Hello agent"
```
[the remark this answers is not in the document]
crickman: Funny...there is a convention in dotnet that requires double-dashes certain cases when a specific code-analysis rule is enabled...

...but yeah, no...fixing

dotnet/src/Experimental/Agents/Internal/Agent.cs
crickman: Yes...this seeped into my awareness and haunted me over the weekend...good call out. 

If I understand correctly, the appropriate template factory needs to be used depending on the template-type (defined withing the template).  This means its possible for the caller to use the wrong factory with a given template.

Ideally, I'd expect a single entry point that was able to fork on whether handlerbars or "kernel prompt"...zooming into thsi now for re-analysis.

crickman: Updated

glahaye (approved):

alliscode (approved):

PR #4501, opened 2024-01-05
Python: Invoke the joke in the Jupyter notebook as async to avoid event loop issues.

### LINKED ISSUE #4492: Python: Invoking Skill More than once hangs the Kernel
**Describe the bug**
When trying to invoke a skill again in a notebook, the kernel hangs and continues to run with no completion eventually timing out.

**To Reproduce**
1. Go to "02-running-prompts-from-file.ipynb"
2. Run through notebook, then change the input text and re-run the last cell where you generate a joke

**Expected behavior**
The output should update or display something

**Screenshots**

**Platform**
 - OS: Windows
 - IDE: VS Code
 - Language: Python

moonbox3:
### Motivation and Context

Fixes #4492

When working with Jupyter notebooks and asyncio, it's important to remember that Jupyter already runs its own event loop. This means that when we try to manage another event loop (e.g., with asyncio.run), it can lead to conflicts or unexpected behaviors like hanging issues.

### Description

Given that Jupyter already has a running event loop, using asyncio.run in the Jupyter notebook is not necessary and could be causing the hang. We can use await to call the `invoke_async` method. This integrates well with Jupyter's existing event loop and won't cause the hanging behavior.

Note that if one is running code from a `.py` file, the original method to call the joke (`result = jokeFunction("time travel to dinosaur age")`), even multiple times one after another, succeeds as there aren't multiple event loops battling each other.

### Contribution Checklist


juliomenendez (approved):
Nice catch! LGTM

juliomenendez (approved):

moonbox3:
This was already fixed in a recent PR. Closing.

PR #4514, opened 2024-01-08
.Net: Make KernelFunction.ExecutionSettings public to enable access from IAIServiceSelector implementations

### LINKED ISSUE #4487: .Net: Can't access KernelFunction.ExecutionSettings from IAIServiceSelector implementation
**Describe the bug**

markwallace-microsoft:
### Motivation and Context

Resolves #4487 

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -73,7 +73,12 @@ public abstract class KernelFunction
     /// <summary>
     /// Gets the prompt execution settings.
     /// </summary>
-    internal IReadOnlyDictionary<string, PromptExecutionSettings>? ExecutionSettings { get; }
+    /// <remarks>
+    /// The execution settings are used to configure the behavior of the function when calling a LLM.
+    /// The currently configured instance of <@see cref="IAIServiceSelector"/> from the <see cref="Kernel"/> is used
+    /// to determine the <see cref="IAIService"/> and <see cref="PromptExecutionSettings"/> to use for the function.
+    /// </remarks>
+    public IReadOnlyDictionary<string, PromptExecutionSettings>? ExecutionSettings { get; }
```
stephentoub, line 81: This now makes KernelFunction partially mutable because PromptExecutionSettings is mutable, right?

markwallace-microsoft: @stephentoub I also created a branch where I made a defensive copy of the execution settings so that `KernelFunction` remains immutable. Here's the draft PR https://github.com/microsoft/semantic-kernel/pull/4531.

Interested to hear your thoughts on this approach?

stephentoub: Thanks. Left a comment on that PR.

PR #4526, opened 2024-01-08
.Net Kernel Examples as tests

RogerBarreto:
### Motivation and Context

Convert Kernel Examples as Integration Tests.


dotnet/samples/KernelSyntaxExamples/Example01_MethodFunctions.cs
```diff
@@ -1,22 +1,30 @@
...
 
-public static class Example01_MethodFunctions
+namespace Examples;
+
+public class Example01_MethodFunctions : BaseTest
 {
-    public static Task RunAsync()
+    public Example01_MethodFunctions(ITestOutputHelper output) : base(output)
+    {
+    }
```
SergeyMenshykh, line 14: If we cannot avoid these boilerplate constructors, we can consider prettifying them using C# 12 primary constructor syntax: `public class Example01_MethodFunctions(ITestOutputHelper output) : BaseTest(output)`.

dotnet/samples/KernelSyntaxExamples/Example45_MultiStreamingChatCompletion.cs
```diff
@@ -62,24 +60,25 @@ private static async Task OpenAIMultiStreamingChatCompletionAsync()
...
         var prompt = "Hi, I'm looking for 5 random title names for sci-fi books";
 
         await ProcessStreamAsyncEnumerableAsync(chatCompletionService, prompt, executionSettings, consoleLinesPerResult);
 
-        Console.WriteLine();
+        this._output.WriteLine();
 
         // Set cursor position to after displayed results
-        Console.SetCursorPosition(0, executionSettings.ResultsPerPrompt * consoleLinesPerResult);
+        // Console.SetCursorPosition(0, executionSettings.ResultsPerPrompt * consoleLinesPerResult);
```
SergeyMenshykh, line 73: Is it still needed? There's also commented out code in lines 64, 92 and 108.

dotnet/samples/KernelSyntaxExamples/KernelSyntaxExamples.csproj
```diff
@@ -3,26 +3,31 @@
     <UserSecretsId>5ee045b0-aea3-4f08-8d31-32d1a6f8fed0</UserSecretsId>
   </PropertyGroup>
   <PropertyGroup>
+    <AssemblyName>KernelSyntaxExamples</AssemblyName>
     <RootNamespace></RootNamespace>
     <TargetFramework>net6.0</TargetFramework>
```
SergeyMenshykh, line 8: net8.0 ?

dotnet/samples/KernelSyntaxExamples/Example45_MultiStreamingChatCompletion.cs
RogerBarreto: I leave it there because if a sk developer wants to use this example in a console application, this line will make it better to show the streaming behavior.

dotnet/samples/KernelSyntaxExamples/Example09_FunctionTypes.cs
```diff
@@ -11,19 +11,26 @@
...
+namespace Examples;
+
+public class Example09_FunctionTypes : BaseTest
 {
-    public static async Task RunAsync()
+    [Fact]
+    public async Task RunAsync()
     {
-        Console.WriteLine("======== Method Function types ========");
+        this._output.WriteLine("======== Method Function types ========");
```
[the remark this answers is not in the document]
RogerBarreto: Do I didn't, this is a good suggestion, I will do this change.

dotnet/samples/KernelSyntaxExamples/Example05_InlineFunctionDefinition.cs
```diff
@@ -49,14 +54,18 @@ public static async Task RunAsync()
...
+        this._output.WriteLine(result.GetValue<string>());
 
         var fixedFunction = kernel.CreateFunctionFromPrompt($"Translate this date {DateTimeOffset.Now:f} to French format", new OpenAIPromptExecutionSettings() { MaxTokens = 100 });
 
         result = await kernel.InvokeAsync(fixedFunction);
-        Console.WriteLine(result.GetValue<string>());
+        this._output.WriteLine(result.GetValue<string>());
+    }
+
+    public Example05_InlineFunctionDefinition(ITestOutputHelper output) : base(output)
```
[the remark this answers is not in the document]
RogerBarreto: I would say the opposite for the samples, I may started with the constructor on top, but it don't add any value to the example, moving it down free up the boiler plate. I will move all to the end only for the samples, what you think?

Krzysztof318:
I'm not sure this is better way. For example rider force build projects always before run tests session. That is annoying slow down.

Krzysztof318:
But if you prefer examples in the form of tests, I would go one step further and break the tests into smaller ones, i.e. depending on the deployment type (for example openai, azure, google, bing), and instead of checking whether the configuration is not null turn the configuration check into an assert. You could even break the sub-examples into separate tests.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #4536, opened 2024-01-09
.Net: Proposal abstraction improvement: add VideoContent and passing complex object to TextGenerationService

Krzysztof318:
### Motivation and Context

While creating Gemini Connector I have found some limitation in SK abstraction layer.
Gemini doesn't support currently vision model with chat service only from text generation service. But text generation service interface lacks of passing complex object like image as param, only text. Also abstraction lacks video content.

### Description

This code doesn't build. It's only proposal for now.
Added VideoContent class.
Changed ChatMessageContent to inherit from new class MessageContent. Added new methods to ITextGenerationService.

### Contribution Checklist


RogerBarreto:
Hi @Krzysztof318 thanks for proposing this, actually I highly recomment splitting this PR in two.

First, the new VideoCOntent with a concrete use-case for it and an example using it.
Second, I encourage dropping redefining our core abstractions from Chat to non Chat, this adds a more API's that we are avoiding at most, for the approach you expect I suggest using the Standardized Prompt to convert a string input into the detailed payload. 

Today we already have a Parser that converts strings from TextGeneration into ChatHistories, I suggest improving the current parser to accomodate your use case.

Thanks.

RogerBarreto:
Hi @Krzysztof318 thanks for proposing this, actually I highly recomment splitting this PR in two.

First, the new VideoCOntent with a concrete use-case for it and an example using it.
Second, I encourage dropping redefining our core abstractions from Chat to non Chat, this adds a more API's that we are avoiding at most, for the approach you expect I suggest using the Standardized Prompt to convert a string input into the detailed payload. 

Today we already have a Parser that converts strings from TextGeneration into ChatHistories, I suggest improving the current parser to accomodate your use case.

Thanks.

Krzysztof318:
Hi @RogerBarreto I have rethought the matter and extended the parser operation here #4602

RogerBarreto:
Teh concept of `Message` is directly associated to ChatModels, introducing this concept as not a ChatModel abstraction is not extensible. We have in our backlog what we consider to be an `UberService<TInput, TOutput>` where the service will specify in the concrete implementation what is the expected type. 

The first question I have is, why Gemini understands what is a Video Link and cannot understand what is a ChatHistory, may worth accomodating the IChatCompletion instead of extending the ITextGeneration to a Gemini specific scenario or create a whole new interface for that matter, but like I mentioned in the first paragraph, we are avoiding add any interface that we already know will be obsolete.

Krzysztof318:
@RogerBarreto Thanks for clarification, for now I can close this PR. I holds off with gemini vision implementation until it will fully support chat.

PR #4542, opened 2024-01-09
.Net: [Fix] .Net: FunctionCallingStepwisePlannerTests.DoesNotThrowWhenPluginFunctionThrowsNonCriticalExceptionAsync is failing

### LINKED ISSUE #4534: .Net: FunctionCallingStepwisePlannerTests.DoesNotThrowWhenPluginFunctionThrowsNonCriticalExceptionAsync is failing
**Describe the bug**
The `FunctionCallingStepwisePlannerTests.DoesNotThrowWhenPluginFunctionThrowsNonCriticalExceptionAsync` integration tests is failing` with this error `System.InvalidProgramException : Common Language Runtime detected an invalid program.`.

I am using `Planners:OpenAI:ModelId = gpt-3.5-turbo-1106`

**To Reproduce**
1. Enable and run the integration test

**Expected behavior**
All tests must pass.

**Platform**
 - OS: Windows
 - IDE: Visual Studio
 - Language: C#
 - Source: Latest code from main

gitri-ms:
### Motivation and Context

Fix for failing integration test (resolves #4534)

When using certain OpenAI models such as `gpt-3.5-turbo-1106`, the model would sometimes opt to call `GetEmailAddressAsync` when it wasn't actually needed. This was causing a critical exception to be thrown (which was by design for a different test case, but in this case was being called unexpectedly and causing the test to fail.)

### Description

This change modifies the test scenarios to minimize the possibility of a critical exception being thrown by mistake. Now `WriteJokeAsync` throws the critical exception, so the other test cases (that involve writing/emailing poems) should not encounter it.

### Contribution Checklist

### DIFF
--- a/dotnet/src/IntegrationTests/Planners/Stepwise/FunctionCallingStepwisePlannerTests.cs
+++ b/dotnet/src/IntegrationTests/Planners/Stepwise/FunctionCallingStepwisePlannerTests.cs
@@ -148,7 +148,7 @@ public sealed class FunctionCallingStepwisePlannerTests : IDisposable
 
     private readonly RedirectOutput _testOutputHelper;
     private readonly IConfigurationRoot _configuration;
-    private readonly ILoggerFactory _logger;
+    private readonly XunitLogger<Kernel> _logger;
 
     public void Dispose()
     {
--- a/dotnet/src/IntegrationTests/Planners/Stepwise/FunctionCallingStepwisePlannerTests.cs
+++ b/dotnet/src/IntegrationTests/Planners/Stepwise/FunctionCallingStepwisePlannerTests.cs
@@ -72,7 +72,7 @@ public sealed class FunctionCallingStepwisePlannerTests : IDisposable
         }
     }
 
-    [Fact(Skip = "System.InvalidProgramException : Common Language Runtime detected an invalid program.")]
+    [Fact]
     public async Task DoesNotThrowWhenPluginFunctionThrowsNonCriticalExceptionAsync()
     {
         Kernel kernel = this.InitializeKernel();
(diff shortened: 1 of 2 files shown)


SergeyMenshykh (approved):

dotnet/src/IntegrationTests/Planners/Stepwise/FunctionCallingStepwisePlannerTests.cs
```diff
@@ -104,8 +104,8 @@ public async Task ThrowsWhenPluginFunctionThrowsCriticalExceptionAsync()
             new FunctionCallingStepwisePlannerConfig() { MaxIterations = 5 });
 
         // Act & Assert
-        // Planner should call ThrowingEmailPluginFake.GetEmailAddressAsync, which throws InvalidProgramException
```
alliscode, line 107: Have you considered only adding the WriteJokeAsync function as a plugin to ensure that it doesn't get confused call the wrong one? Something like:

```
kernel.Plugins.Add(
     KernelPluginFactory.CreateFromFunctions(
        "JokePlanner",
        new[] {
            KernelFunctionFactory.CreateFromMethod({ThrowingEmailPluginFake instance}.WriteJokeAsync)
        }));
```

markwallace-microsoft (approved):

RogerBarreto (approved):

alliscode (approved):

dotnet/src/IntegrationTests/Planners/Stepwise/FunctionCallingStepwisePlannerTests.cs
gitri-ms: I didn't realize this was a supported pattern, but I agree it makes sense here. Updated!

markwallace-microsoft (approved):

markwallace-microsoft:
@gitri-ms Could you take a look as the problem still seems to be there, see: https://github.com/microsoft/semantic-kernel/actions/runs/7487880249/job/20381194144#step:6:319

gitri-ms:
> @gitri-ms Could you take a look as the problem still seems to be there, see: https://github.com/microsoft/semantic-kernel/actions/runs/7487880249/job/20381194144#step:6:319

Hmm... this looks like a different exception being thrown. Looking...

glahaye (approved):

PR #4547, opened 2024-01-10
.Net: Make KernelFunction.ExecutionSettings public to enable access from IAIServiceSelector implementations

### LINKED ISSUE #4487: .Net: Can't access KernelFunction.ExecutionSettings from IAIServiceSelector implementation

markwallace-microsoft:
### Motivation and Context

Resolves #4487 

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -15,7 +16,7 @@ namespace Microsoft.SemanticKernel;
 /// if the service they are calling supports additional properties. For an example, please reference
 /// the Microsoft.SemanticKernel.Connectors.OpenAI.OpenAIPromptExecutionSettings implementation.
 /// </remarks>
-public class PromptExecutionSettings
+public class PromptExecutionSettings : ICloneable
```
stephentoub, line 19: Implementing this interface is generally discouraged. You can still have a Clone method, but we generally don't implement the interface on new things anymore. One of those examples where the interface shipped without enough implementers and consumers to understand its real worth.

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -39,5 +40,15 @@ public class PromptExecutionSettings
     /// Avoid using this property if possible. Instead, use one of the classes that extends <see cref="PromptExecutionSettings"/>.
     /// </remarks>
     [JsonExtensionData]
-    public Dictionary<string, object>? ExtensionData { get; set; }
+    public virtual Dictionary<string, object>? ExtensionData { get; set; }
```
stephentoub, line 43: _Technically_ making something virtual is a breaking change. However, at this point in SK's adoption, it's ok. The impact is existing callers in compiled binaries will be using call rather than callvirt, which means they won't see any overrides and will continue to use the base implementation.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -97,7 +95,31 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
...
+        if (this._executionSettings == null)
+        {
+            return null;
+        }
+
+        var dictionary = new Dictionary<string, PromptExecutionSettings>();
+        foreach (var keyValue in this._executionSettings)
+        {
+            dictionary.Add(keyValue.Key, (PromptExecutionSettings)keyValue.Value.Clone());
+        }
```
stephentoub, line 120: I'm concerned about the expense of this.  How common will it be for a) functions to have non-null _executionSettings and b) for this GetExecutionSettings() to be accessed.  Is it called every time a function is invoked? That makes me nervous.

Is there an alternative design where something truly readonly could be surfaced?

Or maybe we give up on the guaranteed immutability and just document "don't mutate this"?

stephentoub: What information from the PromptExecutionSettings is actually required by consumers?

markwallace-microsoft: Typically for a service selector implementation the model id and service id would be used. But I could also imagine a scenario where someone wants to switch based on the max tokens that are configured.

The obvious risk is some code mutating the execution settings during  a request and impacting all future uses of that function.

I'll look at lazily creating a read only version of the execution settings for each instance and using that.

stephentoub: What if we added a virtual Freeze method, where after calling it attempting to set any properties throw? And we'd freeze it as part of function construction.

PR #4600, opened 2024-01-13
.Net: OpenAI embeddings refactor

### LINKED ISSUE #3294: .Net: Batch embeddings request

anthonypuppo:
### Motivation and Context
Continuation of #3295

The OpenAI embeddings endpoint and `text-embedding-ada-002` supports either a single item or an array of items. Rather than send text inputs one by one, the request should send them all at once for improved performance.

Note that this will require callers to manage their own batching strategy (added an example to demonstrate). I view this as a worthwhile tradeoff for improved performance out of the box.

Fixes #3294.

### Description
- Text data is sent as an array to the embeddings endpoint instead of an individual request per text.
- Aligns embedding request functionality with how the HuggingFace connector works.
- Remove unnecessary `ToArray()` call on the response embeddings.

### Contribution Checklist

- [ ] I didn't break anyone :smile: (implementations passing more than 16 inputs will get an error from the OpenAI endpoint and need to batch appropriately)


dotnet/samples/KernelSyntaxExamples/Example76_TextEmbedding.cs
```diff
@@ -0,0 +1,170 @@
...
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+using Microsoft.SemanticKernel.Text;
+using RepoUtils;
+using SharpToken;
+
+public static class Example76_TextEmbedding
+{
+    private const string EmbeddingModelName = "text-embedding-ada-002";
+    private const string ChatTranscript =
```
dmytrostruk, line 14: nit: It would be nice to hide this chat transcript somewhere lower in this file under region or maybe put it in separate txt file, so when developers open an example, they see actual code. I think the content of chat transcript doesn't matter in this example, please let me know if I'm missing something.

dotnet/samples/KernelSyntaxExamples/RepoUtils/EnumerableExtensions.cs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+
+namespace RepoUtils;
+
+public static class EnumerableExtensions
+{
+    public static IEnumerable<List<TSource>> ChunkByAggregate<TSource, TAccumulate>(
```
dmytrostruk, line 10: Is there a chance that this method will be useful for developers as well? If yes, maybe we can ship it as part of `TextChunker`, with name something like `TextChunker.SplitParagraphsByAggregate` (not as extension, just static method). We can add some tests, mark it as experimental for the first period.

Otherwise, if this method is for example purposes only, maybe we can add it as private method in Example76_TextEmbedding to avoid creating separate files. I'm not sure this method will be re-used in other examples.

dmytrostruk (commented):
Thanks @anthonypuppo , looks good to me! 
Small comments on example and there is small encoding issue:
image

dotnet/samples/KernelSyntaxExamples/Example76_TextEmbedding.cs
anthonypuppo: Moved transcript to the bottom of the file and inlined the embedding model name for clarity.

dotnet/samples/KernelSyntaxExamples/RepoUtils/EnumerableExtensions.cs
anthonypuppo: It would probably be helpful, though the usage here is less related to text chunking and more so transforming for a unique HTTP request. Not sure it belongs in the TextChunker or in the scope of this PR. Thinking an example is sufficient for now?

Having a dedicated extension method allows for a nicer code flow via method chaining imo. I made it a separate file due to not being able to nest extension classes and the lack of the `file private` modifier in .NET 6 (C# 10). If it was a separate class in the file all examples could still call the extension and potentially cause some confusion over where it lives.

Thoughts?

dmytrostruk:
@anthonypuppo In recent PR I added set of unit tests for OpenAI embedding service, which now should be updated in order to respect the new logic. Could you please take a look when you have a chance? Let me know if any assistance is needed. Thank you!

anthonypuppo:
> @anthonypuppo In recent PR I added set of unit tests for OpenAI embedding service, which now should be updated in order to respect the new logic. Could you please take a look when you have a chance? Let me know if any assistance is needed. Thank you!

@dmytrostruk Fixed unit tests and added a fail fast if an empty data list is passed.

dmytrostruk (approved):
@anthonypuppo Thank you!

dotnet/samples/KernelSyntaxExamples/Example76_TextEmbedding.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+using Microsoft.SemanticKernel.Text;
+using RepoUtils;
+using SharpToken;
+
+public static class Example76_TextEmbedding
+{
+    public static async Task RunAsync()
```
RogerBarreto, line 13: Please use xUnit IntegrationTests approach.

anthonypuppo: Updated, let me know how that looks.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -210,17 +210,18 @@ await foreach (Completions completions in response)
...
-            Response<Azure.AI.OpenAI.Embeddings> response = await RunRequestAsync(() => this.Client.GetEmbeddingsAsync(options, cancellationToken)).ConfigureAwait(false);
-            if (response.Value.Data.Count == 0)
+            if (embeddings.Count != data.Count)
             {
-                throw new KernelException("Text embedding not found");
+                throw new KernelException($"Expected {data.Count} text embedding(s), but received {embeddings.Count}");
             }
 
-            result.Add(response.Value.Data[0].Embedding.ToArray());
+            result.AddRange(embeddings.OrderBy((x) => x.Index).Select((x) => x.Embedding));
```
RogerBarreto, line 224: Is there a special reason why we are applying the order by index in here? I suggest not mutating the original payload and get the embedding arrays in the provided order by the server if there's no real benefit to it.

anthonypuppo: It's probably not necessary but, considering the index is a part of the response and I see no explicit language in the API reference that guarantees the response will come back ordered, a max 16 element sort operation seems like a small price to pay. Happy to remove if there's a different opinion or some clarification around response ordering guarantees?

dmytrostruk: @anthonypuppo I think we can avoid ordering on SK side. If that's needed, users can always perform ordering on their side.

anthonypuppo: Updated to remove the ordering.

dmytrostruk: Thank you!

markwallace-microsoft (approved):

dmytrostruk:
@anthonypuppo Looks like Azure OpenAI integration test for embeddings is failing:
image

anthonypuppo:
> @anthonypuppo Looks like Azure OpenAI integration test for embeddings is failing: image

@dmytrostruk What model/version is the test using? This error will come up if `text-embedding-ada-002` is deployed as Version 1. This should be changed to Version 2 for consistency with the OpenAI model.

From the [docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#embeddings):

> We strongly recommend using text-embedding-ada-002 (Version 2). This model/version provides parity with OpenAI's text-embedding-ada-002. To learn more about the improvements offered by this model, please refer to [OpenAI's blog post](https://openai.com/blog/new-and-improved-embedding-model). Even if you are currently using Version 1 you should migrate to Version 2 to take advantage of the latest weights/updated token limit. Version 1 and Version 2 are not interchangeable, so document embedding and document search must be done using the same version of the model.

dmytrostruk:
> @dmytrostruk What model/version is the test using? This error will come up if `text-embedding-ada-002` is deployed as Version 1. This should be changed to Version 2 for consistency with the OpenAI model.

Looks like it's Version 1. It makes sense to migrate to Version 2, as suggested by the documentation, but now I'm wondering if we should be backward compatible with Version 1. Because some users may still use Version 1, and as soon as they will update OpenAI Connector version, it will be breaking change for them.

anthonypuppo:
> > @dmytrostruk What model/version is the test using? This error will come up if `text-embedding-ada-002` is deployed as Version 1. This should be changed to Version 2 for consistency with the OpenAI model.
> 
> Looks like it's Version 1. It makes sense to migrate to Version 2, as suggested by the documentation, but now I'm wondering if we should be backward compatible with Version 1. Because some users may still use Version 1, and as soon as they will update OpenAI Connector version, it will be breaking change for them.

@dmytrostruk This will be breaking regardless as the existing implementation allows an arbitrary amount of inputs passed in. Batching logic will need to be handled by the caller based on their use case. Even if on Version 2 there are still batch limits (~8k tokens and 16 inputs). If they are on Version 1 (which they should really be alerted to as it means their embeddings are not cross compatible with OpenAI's embedding API) they will just need to batch single items.

Long story short, OpenAI upgraded their model in place while Azure OpenAI incremented the version. It's not possible to call "Version 1" on OpenAI as it doesn't exist anymore. Outside of this PR, users who are unknowingly on Azure OpenAI Version 1 (which I would think might be common as the naming is identical) aren't producing the same embeddings as the equivalent OpenAI endpoint would be. That's one nasty bug as you can only figure it out by doing a direct embedding comparison. This breaking change at least forces the caller to investigate the specifics of what model version their using and adapt accordingly.

dmytrostruk:
@anthonypuppo Thanks for explanation, that makes sense. We will update `text-embedding-ada-002` to Version 2 in our deployment, which we are using for testing, and after that we will merge this PR.

markwallace-microsoft (approved):

PR #4612, opened 2024-01-15
.Net: DefaultValue for OpenAPI payload properties

SergeyMenshykh:
### Motivation and Context  
Currently, the default values of parameters in OpenAPI plugins have the 'string?' type, even though their original type specified in the OpenAPI document schema might be different. A dedicated type conversion functionality exists to convert from the original type to a string. However, both the string type for the parameters/arguments and the type conversion logic are no longer necessary, as SK has moved away from string-type arguments and recently added support for .NET primitive types and complex types.  
   
Furthermore, the `RestApiOperationPayloadProperty` class does not have a property for default values, so a default value specified in an OpenAPI document is lost and cannot be accessed from the code.  
   
### Description  
1. The type of the `DefaultValue` property in the `RestApiOperationParameter` class has been changed from string? to object? to better represent the actual type of the parameter's default value and eliminate unnecessary conversion to string.  
   
2. A new `DefaultValue` property has been added to the `RestApiOperationPayloadProperty`, making it consistent with the RestApiOperationParameter class.


dotnet/src/Functions/Functions.OpenApi/Model/RestApiOperationPayloadProperty.cs
```diff
@@ -39,28 +39,37 @@ public sealed class RestApiOperationPayloadProperty
...
+    /// <param name="properties">A list of properties for the payload property.</param>
+    /// <param name="description">A description of the property.</param>
     /// <param name="schema">The schema of the payload property.</param>
+    /// <param name="defaultValue">The default value of the property.</param>
+    /// <returns>Returns a new instance of the <see cref="RestApiOperationPayloadProperty"/> class.</returns>
     public RestApiOperationPayloadProperty(
         string name,
         string type,
         bool isRequired,
         IList<RestApiOperationPayloadProperty> properties,
```
RogerBarreto, line 62: Does this has a strong reason to be a list or could this be a ICollection or Enumerable?

RogerBarreto (approved):

dotnet/src/Functions/Functions.OpenApi/Model/RestApiOperationPayloadProperty.cs
SergeyMenshykh: No specific reason, looking at implementation either of the suggested types would work. I'll keep it as is until the moment we have strong reasons to change.

markwallace-microsoft (approved):

stephentoub (commented):
LGTM

PR #4655, opened 2024-01-17
Java: Implement Example61_MultipleLLMs 

johnoliver:
### Contribution Checklist


java/aiservices/openai/src/main/java/com/microsoft/semantickernel/aiservices/openai/textcompletion/OpenAITextGenerationService.java
```diff
@@ -116,24 +131,13 @@ private CompletionsOptions getCompletionsOptions(
...
-        }
-
-        public Builder withOpenAIAsyncClient(OpenAIAsyncClient client) {
-            this.client = client;
-            return this;
-        }
+    public static class Builder extends TextGenerationService.Builder {
 
-        public OpenAITextGenerationService build() {
+        public TextGenerationService build() {
```
milderhc, line 136: This one returns the interface, but the OpenAIChatCompletion builder returns the class, should both return the same kind of object? either class or interface.

johnoliver: both returning TextGenerationService I think is good

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/services/OrderedAIServiceSelector.java
```diff
@@ -0,0 +1,191 @@
...
+    public OrderedAIServiceSelector() {
+        super(new HashMap<>());
+    }
+
+    public OrderedAIServiceSelector(Map<Class<? extends AIService>, AIService> services) {
+        super(services);
+    }
+
+    @Nullable
+    public AIServiceSelection trySelectAIService(
```
dsgrieve, line 32: I think all of the fall through cases should be split out into their own methods so that the logic here could be  iterating over a stream of these select methods and doing a 'findFirst' kinda thing. This would make it easier to add services in the future.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/services/ParentAIServiceSelector.java
```diff
@@ -0,0 +1,41 @@
...
+import com.microsoft.semantickernel.AIService;
+import com.microsoft.semantickernel.orchestration.KernelFunction;
+import com.microsoft.semantickernel.orchestration.contextvariables.KernelArguments;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class ParentAIServiceSelector implements AIServiceSelector {
```
dsgrieve, line 12: Isn't the java convention "BaseAIServiceSelector"

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/services/ParentAIServiceSelector.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ParentAIServiceSelector.class);
+
+    protected final Map<Class<? extends AIService>, AIService> services;
+
+    public ParentAIServiceSelector() {
```
dsgrieve, line 18: Just a nit, but ctor's in an abstract class should be protected, which reinforces the fact that you can't construct the base class.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/templateengine/semantickernel/blocks/CodeBlock.java
```diff
@@ -120,7 +121,7 @@ private <T> Mono<ContextVariable<T>> renderFunctionCallAsync(
             //Cloning the original arguments to avoid side effects - arguments added to the original arguments collection as a result of rendering template variables.
             arguments = this.enrichFunctionArguments(kernel, fBlock,
                 arguments ==
-                    null ? new DefaultKernelArguments() : new DefaultKernelArguments(arguments));
+                    null ? new DefaultKernelArguments() : new DefaultKernelArguments(arguments, new HashMap<>()));
```
dsgrieve, line 124: Is the hash map mutated downstream? If not, maybe use Collections.emptyMap() here?

dsgrieve (approved):

PR #4673, opened 2024-01-19
Python: Fixes `Plan` using an empty function in describe and incorrectly setting context variables for next steps.

### LINKED ISSUE #2933: Python: Faulty context variable resolution in SequentialPlanner
**Describe the bug**

In certain plans, the mechanism for resolving variables is not being applied correctly with functions that take more than one parameter.

Consider a simple case using `semantic_kernel.core_skills.MathSkill`: "Calculate (10 + 2) - (8 + 7)". The resulting plan (from gpt-3.5-turbo-16k-0613) is:

```
Function: mathskill.Add
Input vars: {'input': '10', 'amount': '2'}
Output vars: ['SUM1']

Function: mathskill.Add
Input vars: {'input': '8', 'amount': '7'}
Output vars: ['SUM2']

Function: mathskill.Subtract
Input vars: {'input': '$SUM1', 'amount': '$SUM2'}
Output vars: ['RESULT__FINAL_ANSWER']
```

The plan is correct. But when it comes to evaluation, the last step fails because it doesn't resolve `amount` to the corresponding result, leaving the string value `'$SUM2'` unresolved.

**To Reproduce**

```
planner = SequentialPlanner(kernel=kernel)
kernel.import_skill(MathSkill(), "mathskill")
plan = asyncio.run(planner.create_plan_async("Calculate (10 + 2) - (8 + 7)"))
result = asyncio.run(kernel.run_async(plan))
for index, step in enumerate(plan._steps):
   print("\nFunction: " + step.skill_name + "." + step._function.name)
   print("Input vars: " + str(step.parameters.variables))
   print("Output vars: " + str(step._outputs))

print("Result: " + str(result))
```

The final output is:
```
Result: Error: (<ErrorCodes.FunctionInvokeError: 8>, 'Error occurred while running plan step: Context amount provided is not in numeric format: $SUM2', ValueError('Context amount provided is not in numeric format: $SUM2'))
```

**Expected behavior**

`$SUM2` should resolve `15`, leading to a final plan result of `-3`.

**Platform**
 - OS: Windows 10
 - IDE: VSCode
 - Language: Python
 - Source: 0.3.13.dev0

**Additional context**

I tried investigating this further by creating a plugin class with the following functions:

```
class MyMathPlugin:
    @sk_function(
        description="Adds two numbers together",
        name="add",
    )
    @sk_function_context_parameter(
        name="input",
        description="The first number to add",
    )
    @sk_function_context_parameter(
        name="number2",
        description="The second number to add",
    )
    def add(self, context: SKContext) -> str:
        return str(float(context["input"]) + float(context["number2"]))
    
    @sk_function(
        description="Subtract two numbers",
        name="subtract",
    )
    @sk_function_context_parameter(
        name="input",
        description="The first number to subtract from",
    )
    @sk_function_context_parameter(
        name="number2",
        description="The second number to subtract away",
    )
    def subtract(self, context: SKContext) -> str:
        print(str(context.dict()['variables']))
        input = float(context["input"])
        number2 = float(context["number2"])
        out = str(input - number2)
        print(f'{input} - {number2} => {out}')
        return out
```

The printed dictionary is as follows:
```
{'variables': {'input': '12.0', 'number2': '$SUM2', 'sum1': '12.0', 'sum2': '15.0'}}
```

Interestingly, `input` is resolved correctly and there is a `sum2` variable with the actual result of what `number2` should be. 
```[tasklist]
### Tasks
```

### LINKED ISSUE #4658: Python: `Plan.describe()` raises an error if `function` param in the __init__ is `None`
**Describe the bug**

`semantic_kernel.planning.plan.Plan.describe` function raises an error if protected field `_function` is `None`. It is due to the missed check in the function before calling `self._fucntion.describe()`. A `None` check is implemented for other functions where `_function` is used, but not here.

**To Reproduce**

```python
from semantic_kernel.planning import Plan


plan = Plan(name="test")
assert plan._function is None
description = plan.describe()  # raises AttributeError: 'NoneType' object has no attribute 'describe'
```

**Expected behavior**

There should be a check if `self._function` is None and an appropriate handler for this case.

As an additional thing, `__init__` can be changed as well

```python
class Plan(SKFunctionBase):
...
    def __init__(
        ...
        function: Optional[SKFunctionBase] = None,
    ):
        self._function = None if function is None else function  # => self._function = function
```

**Platform**
 - Language: Python
 - Source: pip, v0.4.5.dev0

juliomenendez:
### Motivation and Context
This PR fixes #4658 and fixes #2933.

### Description
The first one basically needed a None check before using `self._function` in the `Plan.describe` method.
The second one was incorrectly setting the values for variables for the next step in the plan. The solution for this was suggested by @amritpal-singh-98, thank you!

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/planning/plan.py
+++ b/python/semantic_kernel/planning/plan.py
@@ -240,7 +240,7 @@ class Plan(SKFunctionBase):
         if self._function is not None:
             self._function.set_default_skill_collection(skills)
 
-    def describe(self) -> FunctionView:
+    def describe(self) -> Optional[FunctionView]:
         if self._function is not None:
             return self._function.describe()
         return None
--- a/python/semantic_kernel/planning/plan.py
+++ b/python/semantic_kernel/planning/plan.py
@@ -241,7 +241,9 @@ class Plan(SKFunctionBase):
             self._function.set_default_skill_collection(skills)
 
     def describe(self) -> FunctionView:
-        return self._function.describe()
+        if self._function is not None:
+            return self._function.describe()
+        return None
 
     def set_available_functions(self, plan: "Plan", context: SKContext) -> "Plan":
         if len(plan.steps) == 0:
--- a/python/semantic_kernel/planning/plan.py
+++ b/python/semantic_kernel/planning/plan.py
@@ -402,19 +402,17 @@ class Plan(SKFunctionBase):
             elif param.name in self._state and (self._state[param.name] is not None and self._state[param.name] != ""):
                 step_variables.set(param.name, self._state[param.name])
 
-        for param_var in step.parameters.variables:
-            if param_var in step_variables:
+        for param_name, param_val in step.parameters.variables.items():
+            if param_name in step_variables:
                 continue
 
-            expanded_value = self.expand_from_variables(variables, param_var)
-            if expanded_value.lower() == param_var.lower():
-                step_variables.set(param_var, step.parameters.variables[param_var])
-            elif param_var in variables:
-                step_variables.set(param_var, variables[param_var])
-            elif param_var in self._state:
-                step_variables.set(param_var, self._state[param_var])
+            if param_name in variables:
+                step_variables.set(param_name, param_val)
+            elif param_name in self._state:
+                step_variables.set(param_name, self._state[param_name])
             else:
-                step_variables.set(param_var, expanded_value)
+                expanded_value = self.expand_from_variables(variables, param_val)
+                step_variables.set(param_name, expanded_value)
 
         for item in variables.variables:
             if item not in step_variables:


python/semantic_kernel/planning/plan.py
```diff
@@ -241,7 +241,9 @@ def set_default_skill_collection(
             self._function.set_default_skill_collection(skills)
 
     def describe(self) -> FunctionView:
```
eavanvalkenburg, line 243: return type should be adjusted to Optional[FunctionView]

eavanvalkenburg (changes requested):
one small typing note, otherwise looks good

python/semantic_kernel/planning/plan.py
juliomenendez: Good catch! Fixed. Do you think we should setup the CI to run `mypy` to catch this errors automatically?

eavanvalkenburg (changes requested):
LGTM

eavanvalkenburg (approved):
Lgtm

moonbox3 (approved):
This is awesome, great work. LGTM.

PR #4682, opened 2024-01-19
Java: Add Example30_ChatWithPrompts

johnoliver:
### Contribution Checklist


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/contextvariables/FunctionResult.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FunctionResult<T> {
+
+    private final ContextVariable<T> result;
+    private final Map<String, ContextVariable<?>> metadata;
+
+    public FunctionResult(ContextVariable<T> result, Map<String, ContextVariable<?>> metadata) {
```
dsgrieve, line 12: 1. Is metadata always going to be arbitrary or is there a set of well known keys? I ask because a `Map<String, ContextVariable<?>>` doesn't help much if you don't know what could potentially be in the metadata. Maybe there needs to be a higher abstraction where you could have OpenAIMetadata or something like that. You'd then have `FunctionResult<T,M>` and `M metadata` and `M getMetadata`
2. FunctionResult isn't something you ever expect the user of this API to create. Rather, it is returned by some concrete implementation of some KernelFunction related API. So maybe this should be just pure interface just to make it clear that the caller consumes this but doesn't ever have to create this. 
3. I think I'd put this up in the orchestration package alongside KernelFunction.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/DefaultKernel.java
```diff
@@ -40,15 +40,15 @@ public DefaultKernel(
     }
 
     @Override
-    public <T> Mono<ContextVariable<T>> invokeAsync(
+    public <T> Mono<FunctionResult<T>> invokeAsync(
```
dsgrieve, line 43: Maybe this would be a good time to move DefaultKernel implementation to Kernel and do away with DefaultKernel.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/KernelFunctionFromMethod.java
```diff
@@ -56,17 +57,16 @@ private KernelFunctionFromMethod(
     }
 
     @Override
-    public <T> Mono<ContextVariable<T>> invokeAsync(
+    public <T> Mono<FunctionResult<T>> invokeAsync(
```
dsgrieve, line 60: Why isn't this in semanticfunctions where KernelFunctionFromPrompt is?

dsgrieve (approved):

milderhc (approved):

Issue #4300, opened 2023-12-15
.Net: Use automatic function invocation in the function calling stepwise planner


stephentoub:
Currently the function calling stepwise planner is doing its own manual handling of tool responses. We should reimplement it in terms of automatic function calling, and if there are features missing in the automatic function calling support, add whatever features are necessary to make it work. Having two different implementations is far from ideal, and the stepwise planner's handling currently isn't as robust (as of https://github.com/microsoft/semantic-kernel/pull/4258).

stephentoub:
cc: @gitri-ms

matthewbolanos:
Closing as completed

Issue #4496, opened 2024-01-05
.Net: Allow responses from Native Functions to be returned directly when using Automatic Function Calling


sandeepvootoori:
Currently when using Auto Function Calling, SK always assumes response from a Native Function is a tool message. We do have some scenarios where its desired to be sent the response from a Native Function as an Assistant message. For example we have a Native Function which answers question in a very specific format because of legal reasons but when we use Automatic Function calling it passes Native Function response as a tool message and LLM changes the text in the subsequent LLM call.

sandeepvootoori:
@matthewbolanos

stephentoub:
I'm not quite following.

You're saying you're including a function in the kernel, asking for auto-function invocation, the LLM asks for that function to be invoked, it's invoked, but you don't want the result of that function sent back to the LLM?  What do you want to happen?

sandeepvootoori:
I want the result of the native function to be returned and Auto Function calling should end at that point because the goal has been reached for generating a response to the user. I have tried prompting it to end but it never stops execution at Native Function

stephentoub:
I see, thanks. What I've been thinking should be done, and would likely be necessary for https://github.com/microsoft/semantic-kernel/issues/4300 as well, is some kind of callback that can be provided that would allow the caller to have much more fine-grained control over the auto-invocation, e.g. it would be provided with information like the KernelFunction, the KernelArguments, the number of functions that have already been invoked, etc., and it would get to determine whether to invoke the function or end the processing, and potentially even substitute a different function to invoke, tweak the arguments, etc.  It needs some design work, though.

bjsmiley:
@sandeepvootoori I'm running into the same problem. My use case is to have the planner respond with a question when it doesnt have enough information from the user to invoke the function it actually wants to use. A hacky workaround I have is to use a `CancellationTokenSource` & `Kernel.Data` to exit early & reply with the question. I am very open to alternatives.

```c#
        var cts = new CancellationTokenSource();
        kernel.Data["stop"] = cts;
        kernel.Data["question"] = new Reference<string>(null); // Reference is just a wrapper object to persist the inner value across Kernel.Clone() calls

        FunctionCallingStepwisePlannerResult? result = null;
        try
        {
            result = await planner.ExecuteAsync(kernel, ask, cts.Token);
        }
        catch (OperationCanceledException) when (cts.IsCancellationRequested)
        {
[...]

PR #4686, opened 2024-01-19
.Net: ADR for Automatic Tool Calling Controls

gitri-ms:
### Motivation and Context

ADR proposing additional controls for automatic tool calling

[Task 4300](https://github.com/microsoft/semantic-kernel/issues/4300) is underway to use automatic tool calling in the `FunctionCallingStepwisePlanner`. This will allow for more robust and consistent handling of tool calls and minimize duplication of code.

Currently, as part of the planner configuration, the caller can set the `MaxIterations` field to limit the number of steps (and therefore, calls to the model) utilized by the planner. This enables the caller to control things such as cost, time, etc., spent by the planner.

The current implementation of automatic tool calling allows for multiple consecutive tool calls to be requested/invoked by the model for a single request. However, the limit on how many consecutive tool calls can be performed per request is not exposed to the caller. Therefore, the caller cannot control or even measure after the fact how many model roundtrips occur for a single request. This control is necessary for the planner to accurately enforce the `MaxIterations` limit.

Code snippets are provided in the ADR, but you can see the full changes in #4687 (options 1 & 2) and #4688 (option 3)


docs/decisions/0031-automatic-tool-calling-controls.md
```diff
@@ -0,0 +1,198 @@
...
+
+[Task 4300](https://github.com/microsoft/semantic-kernel/issues/4300) is underway to use automatic tool calling in the `FunctionCallingStepwisePlanner`. This will allow for more robust and consistent handling of tool calls and minimize duplication of code.
+
+Currently, as part of the planner configuration, the caller can set the `MaxIterations` field to limit the number of steps (and therefore, calls to the model) utilized by the planner. This enables the caller to control things such as cost, time, etc., spent by the planner.
+
+The current implementation of automatic tool calling allows for multiple consecutive tool calls to be requested/invoked by the model for a single request. However, the limit on how many consecutive tool calls can be performed per request is not exposed to the caller. Therefore, the caller cannot control or even measure after the fact how many model roundtrips occur for a single request. This control is necessary for the planner to accurately enforce the `MaxIterations` limit.
+
+## Decision Drivers
+
+1. Developers should be able to control how many tool calls (and subsequent model calls) are performed for a single request
```
stephentoub, line 23: We should think about the various knobs someone will need, beyond just what the planner needs right now. Even if we don't expose all the knobs, we should implement this in a way that makes it easy for us to expose additional stuff in the future. For example, https://github.com/microsoft/semantic-kernel/issues/4496#issuecomment-1879175887 effectively wants to be able to examine which function is being invoked, and decide to allow the invocation but then stop further processing. That's more than just MaxIterations, but likely some kind of callback scheme where the callback is presented with a bag of information, like how many iterations have occurred, what function is being called, what arguments are being passed to it, etc., and the callback can then decide to allow the function to invoke, or allow it to invoke but end the processing with its results, or not let it be invoked at all, etc.  We might even want to allow it to do the invocation, so that it could have more control over how it's invoked, or even what's invoked, or what happens if it fails.

docs/decisions/0031-automatic-tool-calling-controls.md
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    // Check for final answer
+    // ...
+}
+```
+
+### Option #3: Add a pre-invoke callback and track model iterations
+
+Add a PreInvokeCallback field to `ToolCallBehavior`. This callback would be invoked inside `ClientCore.cs` before a tool call is invoked. If `true`, would proceed with the tool call invocation. This approach would give the caller control over whether or not a tool call is invoked, while still leveraging the kernel's code for invocation.
```
stephentoub, line 150: I think some variation of option 3 is the most robust / future-proof route. I imagine we'd want something like (waving my hands):
```C#
class ToolCallBehavior
{
    public static ToolCallBehavior Create(Func<AutoInvokeKernelSettings, ValueTask<AutoInvokeKernelResult>> callback);
}
```
where `AutoInvokeKernelSettings` and `AutoInvokeKernelResult` are classes with various information on them, with the input providing details about the Kernel/KernelFunction/KernelArguments/NumberOfIterations/etc., and the output storing details that inform the processing loop of the FunctionResult, what to do next, etc.  Just a rough sketch.

docs/decisions/0031-automatic-tool-calling-controls.md
dmytrostruk: These requirements sound similar to what we already have in filtering functionality. Can we leverage function filters to cover this scenario? We can extend the context of filter with more information if needed, like iteration number, etc, so it would be possible to make a decision about function execution in the scope of automatic tool calling.

crickman:
@dmytrostruk  FYI - Historical

PR #4704, opened 2024-01-23
.Net: Move documentation examples from documentation repo into SK repo

glahaye:
### Motivation and Context
In order to keep documentation examples up to date, moving them to a repo where they will actually be built and run.

### Description
Move documentation examples from doc repo to this one.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example02_Plugin.cs
```diff
@@ -0,0 +1,116 @@
...
+using Xunit.Abstractions;
+
+namespace Examples;
+
+/// <summary>
+/// This example shows how to create a plugin class and interact with as described at
+/// https://learn.microsoft.com/en-us/semantic-kernel/overview/
+/// This sample uses function calling, so it only works on models newer than 0613.
+/// </summary>
+public class Example02_Plugin : BaseTest
```
markwallace-microsoft, line 19: We have started to move away from the `ExampleNN_<Description>` naming convention so we should not follow the pattern. The pattern used by the Getting Started samples is the one we want to follow.

That said, I wonder do we want to add these to the Kernel Syntax Examples. I thought we where just going to move the documentation sample projects as is to this repo and then make sure they were being built and tested.

Adding them to the Kernel Syntax Examples could make them harder to consume by someone following the documentation.

@matthewbolanos What are your thoughts here?

Krzysztof318 (changes requested):
I found in many files that you still using Console.WriteLine instead this.WriteLine

Krzysztof318:
@markwallace-microsoft btw. I have to create a few new samples. What is the new pattern of samples?

dotnet/samples/KernelSyntaxExamples/Example02_Plugin.cs
glahaye: Moving these examples to a standalone project

crickman (approved):

teresaqhoang (approved):

crickman (approved):

crickman (approved):

alliscode (approved):

markwallace-microsoft:
> @markwallace-microsoft btw. I have to create a few new examples. What is the new pattern of examples?

@Krzysztof318 You can follow the ExampleNN syntax if it's a one of new sample. We are going to do an excise to group the samples and provide better names so make everything more discoverable.

Issue #4567, opened 2024-01-10
.Net: Proposal: Sanitize input by stripping invalid characters instead of throwing exception


igochkov:
**Describe the bug**
The current implementation of `ValidPluginName` (line 61) and `ValidFunctionName` (line 75) methods in `Functions.OpenAPI/Diagnostics/Verify.cs` throws an argument invalid name exception when the input name contains invalid characters. This behavior prevents using certain naming conventions like those used in GitHub's REST API OpenAPI Description, which allows characters such as `/` and `-` in their `operationId` fields.

**To Reproduce**
The issue can be reproduced by passing a string containing invalid characters (according to the current validation rules) to the `ValidPluginName` or `ValidFunctionName` methods.

Example:
```csharp
// This will throw an exception due to the presence of '/' and '-'.
ValidPluginName("issues/create-milestone");
ValidFunctionName("issues/create-milestone");
```

**Expected behavior**
Instead of throwing an exception, it would be better if the method sanitizes the input by removing any characters that do not match the validation regex (`s_asciiLettersDigitsUnderscoresRegex`).

For example, if we provide `"issues/create-milestone"` as an input, the method should return `"issuescreatemilestone"` or other sanitized version using Pascal or Camel case, removing the `/` and `-` characters.

**Platform**
 - OS: -
 - IDE: -
 - Language: C#
 - Source: Microsoft.SemanticKernel.Plugins.OpenApi package, version 1.0.1-alpha

**Additional context**
This update is vital for maintaining compatibility with APIs that have different naming conventions. By sanitizing input, you can ensure the methods are more flexible and can handle a broader range of valid identifiers used by such external systems.

For reference, here is the [GitHub's REST API OpenAPI Description](https://github.com/github/rest-api-description)

Thank you for considering this improvement.

mehrandvd:
In the current implementation, the `ValidPluginName` function performs two distinct validations:
1. It checks if the plugin name is valid.
2. It ensures that the plugin name is not a duplicate.

```csharp
internal static void ValidPluginName(...)
{
    NotNullOrWhiteSpace(pluginName);
    if (!s_asciiLettersDigitsUnderscoresRegex.IsMatch(pluginName))
    {
        ThrowArgumentInvalidName("plugin name", pluginName, paramName);
    }

    if (plugins is not null && plugins.Contains(pluginName))
    {
        throw new ArgumentException($"A plugin with the name '{pluginName}' already exists.");
    }
}
```

[...]

PR #4708, opened 2024-01-23
.Net: OpenApi operation name sanitization

SergeyMenshykh:
Fixed the issue that prevented OpenAPI operation name sanitization.

Partially closes - https://github.com/microsoft/semantic-kernel/issues/4567


dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiKernelExtensions.cs
```diff
@@ -356,7 +356,7 @@ private static string ConvertOperationIdToValidFunctionName(string operationId,
             Verify.ValidFunctionName(operationId);
             return operationId;
         }
-        catch (KernelException)
+        catch (ArgumentException)
```
markwallace-microsoft, line 359: How about a comment explaining why this exception is being caught and ignored?

markwallace-microsoft (approved):

dmytrostruk (approved):

dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiKernelExtensions.cs
SergeyMenshykh: Make sense, done.

PR #4714, opened 2024-01-23
.Net: Marked Kernel events as deprecated in favor of filters

dmytrostruk:
### Description

1. Added `Obsolete` attribute to Kernel events.
2. Added ADR with information about Kernel filters.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Getting_Started/Step7_Observability.cs
```diff
@@ -15,10 +15,35 @@ namespace GettingStarted;
 
 public class Step7_Observability : BaseTest
 {
+    /// <summary>
+    /// Shows how to observe the execution of a <see cref="KernelPlugin"/> instance with filters.
+    /// </summary>
+    [Fact]
+    public async Task ObservabilityWithFiltersAsync()
```
RogerBarreto, line 22: Would be interesting to have also one example of using filters without DI approach.

dotnet/samples/KernelSyntaxExamples/Getting_Started/Step7_Observability.cs
```diff
@@ -15,10 +15,39 @@ namespace GettingStarted;
...
+        // Invoke the kernel with a prompt and allow the AI to automatically invoke functions
+        OpenAIPromptExecutionSettings settings = new() { ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions };
+        WriteLine(await kernel.InvokePromptAsync("How many days until Christmas? Explain your thinking.", new(settings)));
+    }
+
     /// <summary>
     /// Shows how to observe the execution of a <see cref="KernelPlugin"/> instance with hooks.
     /// </summary>
     [Fact]
+    [Obsolete("Events are deprecated in favor of filters.")]
```
markwallace-microsoft, line 50: Why don't we just remove this from the sample to avoid confusion

markwallace-microsoft (approved):

SergeyMenshykh (approved):

dotnet/samples/KernelSyntaxExamples/Getting_Started/Step7_Observability.cs
dmytrostruk: @markwallace-microsoft My idea was to keep sample with events at least for the short period of time after release so developers can see before/after approaches, compare both of them and observe the differences. But I can remove it if it's not going to be useful, please let me know.

PR #4715, opened 2024-01-23
Python: Sets a valid default content filter code to ResponsibleAIPolicyViolation

### LINKED ISSUE #4709: Python: Error in Content Filtering Exception init
**Describe the bug**
When I call `invoke_async` for a semantic function and pass an offensive message as an input, I expect to get be an SKContext with `error_occurred=True` and `BadContentError` code.
Instead, the following string is returned:  `Error: '' is not a valid ContentFilterCodes`

The error comes from `content_filter_ai_exception.py`, line 78:
image
If inner_error comes without a "code" attribute (like happened to me) there's no such value "" for the ContentFilterCodes.

**Platform**
 - Language: Python
 - Package version: semantic-kernel==0.4.5.dev0

juliomenendez:
### Motivation and Context

When the RAI response doesn't contain a `code` the class `ContentFilterAIException` would fail to initialize because it was setting `""` as content_filter_code. This sets `"ResponsibleAIPolicyViolation"` instead in that case.

Fixes #4709 

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/connectors/ai/open_ai/exceptions/content_filter_ai_exception.py
+++ b/python/semantic_kernel/connectors/ai/open_ai/exceptions/content_filter_ai_exception.py
@@ -75,7 +75,9 @@ class ContentFilterAIException(AIException):
         self._param = inner_exception.param
 
         inner_error = inner_exception.body.get("innererror", {})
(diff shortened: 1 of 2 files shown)


moonbox3 (approved):

eavanvalkenburg (approved):
LGTM

PR #4730, opened 2024-01-24
Java: Proposal for adding hooks

johnoliver:
### Contribution Checklist


dsgrieve:
What is the use-case for potentially having more than one pre/post hook? For example, why would I want to `addHook("foo", preFunctionHookA)` and `addHook("bar", preFunctionHookB)`? 

Unfortunately, the KernelHook API makes it difficult (impossible?) to make my hook function a lambda, e.g., `addHook("foo", e -> blah)`

Overall, I think it is more complicated that it needs to be. I think a simple `KernelArgs beforeInvoke(FunctionInvokingEvent event)` and similar afterInvoke/beforeRender/afterRender would be easier/more obvious. Or `void beforeInvoke(Function<FunctionInvokingEvent, KernelArgs> hook)` would be even better since I could do that as a lambda or as a method reference (e.g., Hooks::beforeInvoke). 

And I probably want to have different hooks depending on what's being executed. So I think putting hooks on Kernel is not the right place. I'd put them on the invoke APIs, but that would be major suckage since you'd have to blow out the API to account for the combinations. 

Maybe if the invoke APIs returned an 'invocation context' which had the hook API, then I could do something like `InvocationContext context = kernel.contextOf(function, args, String.class); String result = context.beforeInvoke(Hooks::beforeInvoke).invokeAsync().block();`

Or maybe invoke just needs to have a Hook... vararg and leave it to the code to figure out which to call under what situation. Or just a single hook that gets called pre/post and let the hook implementation figure it out.

milderhc:
While you can have all your pre and post logic in a single hook, one reason to be able to add multiple hooks is readability. preFunctionHookA can do a completely different thing than preFunctionHookB, you might want to have the logic of each separated for the sake of readability and maintainability. They could also be more easily tested. If it adds too much complexity then I guess a single hook would suffice.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/KernelHook.java
```diff
@@ -0,0 +1,57 @@
+package com.microsoft.semantickernel.hooks;
+
+public interface KernelHook {
+
+    default <T extends HookEvent> T dispatch(T event) {
```
milderhc, line 5: All these checks happen every time `executeHooks` is called, could we avoid it by having two different maps, one for pre and another for post? That way the user is also able to be explicit about what hook they're adding to the kernel by adding `Kernel.addInvokingHook` and `Kernel.addInvokedHook` methods to the API.

johnoliver: this style of dispatch has been removed in the latest pr

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
```diff
@@ -95,6 +105,10 @@ public KernelPluginCollection getPlugins() {
         return plugins;
     }
 
+    public HookService getHookService() {
```
dsgrieve, line 108: I still don't think that hooks belong in the Kernel because if forces a centralized model for managing hooks. What if you have different parts of the code adding hooks to the hook service? Do you want all the hooks called? In what order? I think this is going to be problematic. In my opinion, a user more likely to have some piece of business logic that is invoking some function/prompt where they want to supply a hook or hooks and other places where they do not want hooks. I'd rather see a variation of the invoke method(s) that take a Hooks object.

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/syntaxexamples/Example57_KernelHooks.java
```diff
@@ -365,4 +369,51 @@ private static void chatCompletionHook(Kernel kernel) {
...
+            .withName("Writer")
+            .withDefaultExecutionSettings(PromptExecutionSettings
+                .builder()
+                .withMaxTokens(1000)
+                .withTemperature(1)
+                .withTopP(0.5)
+                .build())
+            .build();
+
+        HookService hooks = new HookService();
```
dsgrieve, line 390: Consider renaming this from "service" since that is an overloaded term in SK. Maybe InvocationHooks or just Hooks.

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/syntaxexamples/Example57_KernelHooks.java
```diff
@@ -365,4 +369,51 @@ private static void chatCompletionHook(Kernel kernel) {
...
+        hooks.addPreChatCompletionHook(event -> {
```
dsgrieve, line 391: "add" begs the question, "Can I add more than one?" I'd opt for `on<name of event>`

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/syntaxexamples/Example57_KernelHooks.java
```diff
@@ -365,4 +369,51 @@ private static void chatCompletionHook(Kernel kernel) {
...
+            ChatCompletionsOptions options = event.getOptions();
+            List<ChatRequestMessage> messages = options.getMessages();
+
+            messages = new ArrayList<>(messages);
+            messages.add(
+                new ChatRequestSystemMessage("Use upper case text when responding to the prompt."));
+
+            return new PreChatCompletionHookEvent(
```
dsgrieve, line 399: I think you can remove "Hook" from these event names since hooks don't themselves have events.

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/syntaxexamples/Example57_KernelHooks.java
```diff
@@ -365,4 +369,51 @@ private static void chatCompletionHook(Kernel kernel) {
...
+                PreChatCompletionHook.cloneOptionsWithMessages(options, messages)
```
dsgrieve, line 400: This is an odd bit of ceremony. Is it really necessary? Why can't I just modify the options/messages/etc and return the event I got called with?

johnoliver: Unfortunately ChatCompletionOptions is not part of semantic kernel and is immutable, so cannot modify it

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/syntaxexamples/Example57_KernelHooks.java
johnoliver: It is possible to have more than one

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/FunctionInvokedEventArgs.java
```diff
@@ -0,0 +1,31 @@
+package com.microsoft.semantickernel.hooks;
+
+import com.microsoft.semantickernel.orchestration.FunctionResult;
+import com.microsoft.semantickernel.orchestration.KernelFunction;
+import com.microsoft.semantickernel.orchestration.contextvariables.KernelArguments;
+
+public class FunctionInvokedEventArgs<T> implements HookEvent {
```
dsgrieve, line 7: It seems odd to me that FunctionInvokedEventArgs would implement HookEvent. I'd expect FunctionInvokedEventArgs to be called FunctionInvokedEvent since it is-a HookEvent. It would be clear (to me, at least) if you renamed HookEvent to HookArgs and s/EventArgs/Args/. Then you'd have FunctionInvokedArgs implements HookArgs, which makes better sense (to me, at least) that these are arguments to/from a function invocation and in use you get `FunctionInvokingHook preHook = functionInvokingArgs -> blah`

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/Hooks.java
```diff
@@ -0,0 +1,94 @@
...
+import com.microsoft.semantickernel.hooks.KernelHook.PromptRenderedHook;
+import com.microsoft.semantickernel.hooks.KernelHook.PromptRenderingHook;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+public class Hooks {
```
dsgrieve, line 15: Should this be KernelHooks? Hook, HookEvent, KernelHook - I'd be consistent one way or the other. Everything KernelHook* or just Hook*

dsgrieve (approved):

milderhc (approved):

dsgrieve (approved):

PR #4740, opened 2024-01-25
.Net: SK Agents

SergeyMenshykh:
### Motivation and Context
This ADR outlines high-level ideas/design for SK agents.


crickman:
I am also curious if "user-proxy" agent might need to be addressed to fully support agent collaboration/chat patterns?

crickman:
Perhaps we can consider the upcoming Azure support: https://github.com/Azure/azure-sdk-for-net/pull/40817

SergeyMenshykh:
> I am also curious if "user-proxy" agent might need to be addressed to fully support agent collaboration/chat patterns?

Yes, it may be necessary. Once the agents' abstraction is in place, we will begin adding concrete implementations for agents, and the "user-proxy" agent could be one of those implementations. There will be a user story to track the "user-proxy" agent for SK.

SergeyMenshykh:
> FWIW - ADR "31" is already overloaded

Yes, thank you. I will change the number before submitting the PR to the main branch.

SergeyMenshykh:
> Perhaps we can consider the upcoming Azure support: [Azure/azure-sdk-for-net#40817](https://github.com/Azure/azure-sdk-for-net/pull/40817)

Yep, definitely, we should migrate our agent to the new API when it's available. I'll create an issue to track this.

SergeyMenshykh:
> Why not implementing all the Autogen agent types? as in https://github.com/microsoft/autogen/pull/924/files#diff-9860f01d79405521f2af05fe4040ed4f50498c9108956723f4bc5f89d88ab5ad :
> 
> * ConversableAgent
> * AssistantAgent
> * GroupChatManager
> * UserProxyAgent...
> 
Once the agent's abstraction is in place, it will be a matter of prioritizing the design and implementation for the agents in a way that aligns with SK's overall approach, including DI, filters, tools, and so on.

SergeyMenshykh:
> But let me get to the core, I mean Kernel ;) The other point I want to bring is a suggestion on the design which the ADR does not mention: kernel centricity. There is no central kernel in the agent story. A new one is created on each Agent, that's why we see the configuration and the model passed on the agent builder. And then it is hidden from the developer, it is internal. IMHO this is not good.
> 
> Why not pass the "central" kernel on its builder, and then use it? in case that we need parallelized execution - the main reason I believe that triggered this design decision - each agent has its own kernel (heart) - why not create a pool of agents and then each agent can take the next available, free, kernel? ;)
> 
> And that could be easily coordinated by the kernel, with a semaphore, thread-safe et voila... we have a better solution than creating all the kernels beforehand. It also would scale better and be more performant.
> 
> And also solve the multi-model compatability (sort of) - as the Kernel is supporting already different AI Models.
> 
> Thoughts?

Even though it is not stated in the ADR, and it should have been and will be, every feature we design or add to SK is supposed to be built around the kernel and leverage as much as possible of all the benefits SK provides or supports, such as DI, plugins, function calling, abstractions, etc. Therefore, agents are no exception in this regard.

Regarding passing the kernel to agents, it is a good idea and aligns with the way we plan to implement agents. SK agents should be flexible and accept an instance of a kernel provided by the SK consumer code. Whether it is one instance of the kernel for all agents, or each agent gets its own instance, it is up to the SK consumer code to decide, depending on the scenario it wants to implement.

Kernel/Agents Pool: It might make sense to arrange agents or kernels into a pool if the cost of creating or destroying them is high. For example, if the classes use expensive system resources that can be shared and reused, it would be logical to organize them in a pool to minimize system resource usage. Regardless, the SK consumer code has a few mechanisms to optimize the cost if it is high. A kernel/agent can be registered as a singleton and reused during the hosting app's lifetime to minimize the cost. If the singleton solution is not sufficient, nothing prevents it from pooling kernels or agents if that is what the hosting app requires.

AshD:
What is the rationale for the Agent class requiring the Assistant APIs?

AutoGen uses the Chat Completion API which is implemented by multiple providers like vLLM, Llama.cpp, etc.

Thanks,
Ash

joslat:
> What is the rationale for the Agent class requiring the Assistant APIs?
> 
> AutoGen uses the Chat Completion API which is implemented by multiple providers like vLLM, Llama.cpp, etc.
> 
> Thanks, Ash

Hi Ash, 
Initially The agents was developed on top of the OpenAI Assistants API, which offers some features  like thread management, tools, function calling and files (not supported just yet but in short will - there are two PRs ongoing)

But as the team said, they want to decouple from the OpenAI implementation, and after this "base work is done"  - which is what this ADR is about, to provide a base architecture/structure from where to build the agent story for all models (and also OpenAI).
@SergeyMenshykh correct me if I am wrong or not precise enough...

docs/decisions/0031-agents.md
```diff
@@ -0,0 +1,277 @@
...
+<img src="./diagrams/open-ai-assistant-api-objects.png" alt="OpenAI Assistant API Objects.png" width="700"/>
+
+[Source](https://platform.openai.com/docs/assistants/how-it-works/objects)
+
+[Playground](https://platform.openai.com/playground)
+
+## Unified Interface And Data Contract
+To enable collaboration and customization scenarios, further details of which will be provided below, agents should have a unified interface. They should inherit from the same abstract Agent class and have one data type - AgentMessage, for both input and output data contracts. This will enable a very powerful and extensible model, allowing the addition of any agent to any chat/collaboration, regardless of its implementation, and enabling a seamless chat conversation where the agent's output message is added to the chat without any conversion, and the chat is sent as input for the next agent without any conversion as well.
+
+For example, the scenario where two agents need to communicate together can configured like this:
```
dmytrostruk, line 53: nit: 
```suggestion
For example, the scenario where two agents need to communicate together can be configured like this:
```

docs/decisions/0031-agents.md
```diff
@@ -0,0 +1,277 @@
...
+- No changes are needed for existing agent chat/collaboration classes to accommodate new agents.  
+- No custom logic is required in the chat/collaboration classes to support new agents.
+   
+Cons:  
+- Some features of certain agents might not be fully utilized, as chat/collaboration classes are only aware of the unified interface supported by all agents.
+
+## Agent Collaboration
+There are three identified methods for agents to collaborate with each other so far. As new collaboration strategies are discovered, it should be relatively easy to support them. One prerequisite for this is to ensure that agents remain unaware of the conversations they participate in, maintaining a one-way dependency - the chat/collaboration blocks know about the agents, but the agents do not know about them.
+
+### Collaboration Chat
```
dmytrostruk, line 98: nit: Feel free to skip if the naming is not important here.
Taking into account, that the name of next approach is `Turn-Based Chat`, and this name briefly describes the approach, in case of `Collaboration Chat` name, it's not clear that it assumes the presence of admin. For clear terminology, I would probably call it something like `Moderated Chat`.

SergeyMenshykh: The names are just for example purposes and might change the implementation phase.

dmytrostruk (approved):

AshD:
> > What is the rationale for the Agent class requiring the Assistant APIs?
> 
> > 
> 
> > AutoGen uses the Chat Completion API which is implemented by multiple providers like vLLM, Llama.cpp, etc.
> 
> > 
> 
> > Thanks, Ash
> 
> 
> 
> Hi Ash, 
> 
> Initially The agents was developed on top of the OpenAI Assistants API, which offers some features  like thread management, tools, function calling and files (not supported just yet but in short will - there are two PRs ongoing)
> 
> 
> 
> But as the team said, they want to decouple from the OpenAI implementation, and after this "base work is done"  - which is what this ADR is about, to provide a base architecture/structure from where to build the agent story for all models (and also OpenAI).
> 
> @SergeyMenshykh correct me if I am wrong or not precise enough...

Thanks @joslat Looking forward to test this

markwallace-microsoft (approved):

PR #4753, opened 2024-01-26
.Net Enable Usage of Custom Compatible Chat Message API Endpoints with OpenAI Connector + Examples

RogerBarreto:
### Motivation and Context

- Allow usage of custom Message API (OpenAI ChatCompletion Standard) compliant endpoints with the OpenAI Connector.
- Refactoring of OpenAI Models and Classes Structure
- Adding Examples on using the current changes against `LMStudio`, `Ollama` and `LocalAI` Message APIs.


Krzysztof318:
I have never used LM-studio, but what this have common with openai implementation?

stephentoub:
I assume this is temporary and having someone implement their own handlers and extension methods and whatnot isn't actually our answer for this, right?

RogerBarreto:
As stated in the Connectors ADR, while the biggest benefit of LM Studio is mimicking OpenAI's Endpoints and Contracts no specific Connector for this platform is needed, unless we have a specific requirement that the OpenAI connector isn't enough for consuming LM Studio Http APIs, we will suggest using the OpenAI's Connector.

We will add an extra configuration option for the OpenAI connectors that won't require to manually define an Handler for it and investigate if the usage of the Endpoint configuration would alone suffice for this.

dotnet/samples/KernelSyntaxExamples/Example78_LMStudio.cs
```diff
@@ -0,0 +1,132 @@
...
+            }
+
+            return base.SendAsync(request, cancellationToken);
+        }
+    }
+}
+
+public static class KernelBuilderExtensions
+{
+    public static IKernelBuilder AddLMStudioChatCompletion(this IKernelBuilder builder)
```
markwallace-microsoft, line 124: Can we create an extension method in the OpenAI connector project with a signature something like this:

`public static IKernelBuilder AddCustomChatCompletion(this IKernelBuilder builder, Uri requestUri)`

And then have samples that use this with LM Studio and the new Ollama. This would give people an OOB option to use our OpenAI connector with local model that are OpenAI REST API compatible.

stephentoub: > AddCustomChatCompletion

I'd personally prefer either AddOpenAICompatibleChatCompletion or just adding overloads of the existing OpenAI methods to also take a uri. Do we know what Azure.AI.OpenAI's plans are for also making this extensible?

If we do add such methods, some care will need to be taken to make them work well with DI, given that the implementation is based on supplying a custom message handler.

rosieks: Please also create same extension for `IServiceCollection`

matthewbolanos:
@RogerBarreto, we should have a better 1st class story for connecting to popular local deployment channels like LLM Studio and Ollama, but in the meantime, I think your sample would be valuable for people who need to customize the OpenAI API endpoint.

Could you update this PR to make it more "generic" (i.e., title the Kernel Syntax Example "custom OpenAI endpoint") so we can get this merged in?

dotnet/src/Connectors/Connectors.OpenAI/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,221 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids -->
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Microsoft.SemanticKernel.Connectors.OpenAI.OpenAIChatCompletionService.#ctor(System.String,System.String,System.String,System.Net.Http.HttpClient,Microsoft.Extensions.Logging.ILoggerFactory)</Target>
```
stephentoub, line 6: I'm having trouble following the diff. Why are these breaking changes being made? Aren't some of these (like this one) not experimental?

dotnet/src/Connectors/Connectors.OpenAI/Core/AzureSdk/OpenAIClientCore.cs
```diff
@@ -51,7 +59,60 @@ internal sealed class OpenAIClientCore : ClientCore
...
+            // Otherwise the error bellow will happen:
+            // System.InvalidOperationException : Bearer token authentication is not permitted for non TLS protected (https) endpoints.
+
+            var fieldInfo = type.GetField("_pipeline", BindingFlags.NonPublic | BindingFlags.Instance);
+
+            // Change the value of the private field
+            fieldInfo?.SetValue(this.Client, HttpPipelineBuilder.Build(options,
+                Array.Empty<HttpPipelinePolicy>(),
+                Array.Empty<HttpPipelinePolicy>(),
+                new ResponseClassifier())
```
stephentoub, line 102: We should not use private reflection to muck with other library internals.

dotnet/src/InternalUtilities/src/Schema/JsonSchemaMapper.ReflectionHelpers.cs
```diff
@@ -177,16 +177,16 @@ private static FieldInfo GetPrivateFieldWithPotentiallyTrimmedMetadata(this Type
     {
         public ParameterLookupKey(string name, Type type)
         {
-            Name = name;
-            Type = type;
+            this.Name = name;
+            this.Type = type;
```
stephentoub, line 181: I was purposefully avoiding making changes to these files for style, so that we can minimize the diffs as we keep it in sync with @eiriktsarpalis's "official" version until it's part of STJ. I'd added an editorconfig that should have suppressed the warnings on these files; did that not work?

RogerBarreto: My IDE was complaining about them, I can revert.

dotnet/src/Connectors/Connectors.OpenAI/CompatibilitySuppressions.xml
markwallace-microsoft: This PR shouldn't have any breaking changes.

Why are the file moves required as part of this PR? If it's some proposed tidy up then can we discuss and agree first, then handle in a seperate PR.

dotnet/src/Connectors/Connectors.OpenAI/Core/AzureSdk/OpenAIClientCore.cs
RogerBarreto: Yes, I had a bad feeling about it, that's why I pinged you 😊

Seems that we can use a Policy to override the Url, used this approach.

dotnet/src/Connectors/Connectors.OpenAI/CompatibilitySuppressions.xml
RogerBarreto: Reverted all other additional changes, including the potential breaking changes as discussed.

dotnet/samples/KernelSyntaxExamples/Example88_CustomMessageAPIEndpoint.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        WriteLine($"Example using local {messageAPIPlatform}");
+        // Setup Llama2 as the model in LM Studio UI.
+
+        var kernel = Kernel.CreateBuilder()
+            .AddOpenAIChatCompletion(
+                modelId: modelId,
+                apiKey: null,
+                endpoint: new Uri(url))
+            .Build();
```
SergeyMenshykh, line 51: I wonder whether this solution was considered:
```csharp
using var httpClient = new HttpClient();
httpClient.BaseAddress = new Uri(url);

var kernel = Kernel.CreateBuilder()
    .AddOpenAIChatCompletion(
        modelId: modelId,
        apiKey: "an api key",
        httpClient: httpClient)
    .Build();
```

stephentoub: I don't think that's viable:
1) HttpClient.BaseAddress is only used when the supplied address is relative. If the Uri is absolute, the BaseAddress is ignored.
2) We can't modify the BaseAddress of an HttpClient that's provided to us, only if we completely own it, both because we could be changing the BaseAddress in a way someone didn't expect, and because setting BaseAddress will fail if the client has already been used to make requests.

markwallace-microsoft (approved):

SergeyMenshykh (approved):

Issue #4696, opened 2024-01-22
.Net: Kernel function arguments cannot be complex type (e.g., array)


ameier38:
**Describe the bug**
I am trying to create a kernel function to sum an array of numbers however if I use an array as the function argument I get the error `Function failed. Error: Object of type 'System.String' cannot be converted to type 'System.Double[]'.`

**To Reproduce**
From the logs it looks like openai function call is correct.
```json
{"function":{"name":"MathPlugin_Sum","description":"Sum an array of numbers.","parameters":{"type":"object","required":["numbers"],"properties":{"numbers":{"type":"array","items":{"type":"number"},"description":"The array of numbers to sum."}}}},"type":"function"}
```
```json
"tool_calls": [
          {
            "id": "call_wj6NbjHTo1C4ErXMO0XyBU9e",
            "type": "function",
            "function": {
              "name": "MathPlugin_Sum",
              "arguments": "{\"numbers\": [1.1, 2.2, 3.3]}"
            }
          }
        ]
```
```
LogRecord.Severity:                Trace
LogRecord.SeverityText:            Trace
LogRecord.FormattedMessage:        Function call requests: MathPlugin_Sum({"numbers":[1.1,2.2,3.3]})
LogRecord.Body:                    Function call requests: {Requests}
LogRecord.Attributes (Key:Value):
    Requests: MathPlugin_Sum({"numbers":[1.1,2.2,3.3]})
    OriginalFormat (a.k.a Body): Function call requests: {Requests}
```
However later on it looks like the method is being called with a string.
```
LogRecord.Timestamp:               2024-01-22T16:57:32.0827729Z
LogRecord.TraceId:                 2f4f4d4f22bfdc2eba3093ee96725edb
LogRecord.SpanId:                  fd6afb6fd9204111
LogRecord.TraceFlags:              Recorded
LogRecord.CategoryName:            Sum
LogRecord.Severity:                Trace
LogRecord.SeverityText:            Trace
LogRecord.FormattedMessage:        Function arguments: {"numbers":"[1.1,2.2,3.3]"}
LogRecord.Body:                    Function arguments: {Arguments}
LogRecord.Attributes (Key:Value):
    Arguments: {"numbers":"[1.1,2.2,3.3]"}
    OriginalFormat (a.k.a Body): Function arguments: {Arguments}
```
Here is my F# script. Run with `dotnet fsi script.fsx`.
```fsharp
#r "nuget: OpenTelemetry"
[...]

Issue #3126, opened 2023-10-11
.Net: Support complex function parameters when using OpenAIFunction calling


frederikrosenberg:
It is currently impossible to have complex objects in the function parameters using `OpenAIFunction` and `OpenAIFunctionParameter`.

The following JSON is a valid request but can not currently be represented using the above classes.

```json
{
  "temperature":0.7,
  "top_p":0.95,
  "frequency_penalty":0,
  "presence_penalty":0,
  "max_tokens":800,
  "messages":[
     {
        "role":"user",
        "content":"What is the weather like in Boston?"
     }
  ],
  "functions":[
     {
        "name":"get_current_weather",
        "description":"Get the current weather in a given location",
        "parameters":{
           "required":[
              "city"
           ],
           "type":"object",
           "properties":{
              "city":{
                 "required":[
                    "name",
                    "country"
                 ],
                 "type":"object",
                 "properties":{
                    "name":{
                       "type":"string"
                    },
                    "country":{
                       "type":"string"
                    }
                 }
              }
           }
        }
     }
  ],
  "function_call":"auto"
}
 ```

Arrays are also not supported.

cloudscool:
Hello frederikrosenberg, you can consider using Native Functions to handle your needs, if it is just a simple ingress, consider using Semantic Functions

SergeyMenshykh:
Hi @frederikrosenberg, thanks for reporting the issue. We are working on fix for the similar issue - https://github.com/microsoft/semantic-kernel/issues/4696 at the moment. The issue you have might have the same root cause and can be addressed by the fix.

In case you think this issue if different please send us code example that demonstrates the problem.

frederikrosenberg:
Hi @SergeyMenshykh, it looks like it is related. I have created a code example just to be sure:

Program.cs
```csharp
using AITest;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Logging;
using Microsoft.SemanticKernel;
using Microsoft.SemanticKernel.Connectors.OpenAI;

var kernelBuilder = Kernel.CreateBuilder();
kernelBuilder.AddAzureOpenAIChatCompletion(...); // using GPT-4-turbo

kernelBuilder.Services.AddLogging(builder => builder.AddConsole());
kernelBuilder.Plugins.AddFromType<WeatherPlugin>("weather");
var kernel = kernelBuilder.Build();

[...]

PR #4757, opened 2024-01-26
.Net: Implicit JSON to target type conversion

SergeyMenshykh:
### Motivation and Context
It would be very convenient and beneficial for both SK consumer code and SK itself if there was an implicit mechanism for deserializing function arguments provided as JSON into their actual types specified in the function signature, similar to the implicit data type conversion mechanism SK currently has.

Closes https://github.com/microsoft/semantic-kernel/issues/4696, https://github.com/microsoft/semantic-kernel/issues/3126

### Description
The change adds the 'TryToDeserializeValue' fallback, which attempts to deserialize the argument value into the target parameter type if all the methods to convert the argument to its actual type have been exhausted.


dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
```diff
@@ -399,6 +408,55 @@ private static bool IsAsyncMethod(MethodInfo method)
...
+            if (value is JsonElement element)
+            {
+                deserializedValue = JsonSerializer.Deserialize(element, targetType);
+                return true;
+            }
+
+            // The JSON can be represented by other data types from various libraries. For example, JObject, JToken, and JValue from the Newtonsoft.Json library.  
+            // Since we don't take dependencies on these libraries and don't have access to the types here,
+            // the only way to deserialize those types is to convert them to a string first by calling the 'ToString' method.
+            deserializedValue = JsonSerializer.Deserialize(value.ToString(), targetType);
```
stephentoub, line 444: Did you consider JsonSerializer.Serialize(value) instead of value.ToString()? That may or may not be better. Or maybe it could be yet another fallback.

dotnet/src/SemanticKernel.UnitTests/Functions/KernelFunctionFromMethodTests1.cs
```diff
@@ -1204,4 +1207,129 @@ public async Task ItPropagatesOriginalExceptionTypeAsync()
...
+        public ThirdPartyJsonPrimitive(string jsonToReturn)
+        {
+            this._jsonToReturn = jsonToReturn;
+        }
+
+        public override string ToString()
+        {
+            return this._jsonToReturn;
+        }
+    }
```
stephentoub, line 1334: Are we able to come up with a function calling example where the LLM will provide a complex argument that can now be parsed successfully? That'd help to validate the full roundtrip, from .NET type to JSON schema to describe the parameter to JSON response for the parameter to deserialized .NET instance.

stephentoub (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
SergeyMenshykh: Anyway, the Duck Typing capability has been reverted for now, until we have more information to make an educated decision.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
```diff
@@ -399,6 +408,58 @@ private static bool IsAsyncMethod(MethodInfo method)
...
+            }
+
+            // The JSON can be represented by other data types from various libraries. For example, JObject, JToken, and JValue from the Newtonsoft.Json library.  
+            // Since we don't take dependencies on these libraries and don't have access to the types here,
+            // the only way to deserialize those types is to convert them to a string first by calling the 'ToString' method.
+            // Attempting to use the 'JsonSerializer.Serialize' method, instead of calling the 'ToString' directly on those types, can lead to unpredictable outcomes.
+            // For instance, the JObject for { "id": 28 } JSON is serialized into the string  "{ "Id": [] }", and the deserialization fails with the
+            // following exception - "The JSON value could not be converted to System.Int32. Path: $.Id | LineNumber: 0 | BytePositionInLine: 7."
+            deserializedValue = JsonSerializer.Deserialize(value.ToString(), targetType);
+            return true;
```
dmytrostruk, line 448: nit: a little simplification for try block, feel free to skip:
```csharp
deserializedValue = value switch
{
    JsonDocument document => document.Deserialize(targetType),
    JsonNode node => node.Deserialize(targetType),
    JsonElement element => element.Deserialize(targetType),
    _ => JsonSerializer.Deserialize(value.ToString(), targetType)
};

return true;
```

dmytrostruk (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
SergeyMenshykh: Yep, thanks.

dmytrostruk (approved):

dmytrostruk (approved):

dmytrostruk (approved):

dmytrostruk (approved):

Issue #4667, opened 2024-01-18
Python: Refactor PluginCollection/Base and ReadOnlyPluginCollection/Base. Add KernelFunction/KernelFunctionMetadata.


moonbox3:
Refactor (and remove as needed) the PluginCollection/Base and ReadOnlyPluginCollction/Base to better match the structure in dotnet. There is a lot of extra code and potentially base classes that aren't needed.

This work should also introduce the following:

	- KernelFunction
	- KernelFunctionMetadata
		Name
		PluginName
		Description
		Parameters
		ReturnParameter

Issue #4560, opened 2024-01-10
Python: Add plugin class that can hold multiple types of functions (e.g., native and prompts)


matthewbolanos:
This should also be used as an opportunity to bring all "skill_collection" classes together

moonbox3:
Planning to introduce the KernelPlugin (base class) and the DefaultKernelPlugin, along with unit tests, in this first iteration. Once we move to support the KernelFunction then we can also introduce the KernelFunctionMetadata. I think at that point it makes sense updating the code to start using the DefaultKernelPlugin.

moonbox3:
Also waiting for further review on #4595 so I can commit to main and then push a PR for the KernelPlugin as it depends upon the rename.

moonbox3:
Refactoring PluginCollection classes will be done in #4667

PR #4764, opened 2024-01-26
Python: Add kernel plugin collection and remove old plugin collection classes

moonbox3:
### Motivation and Context

This PR solves two works items: #4560 and #4667. The Python SDK had different ways of handling plugins and collections. The work in the PR helps align the SDK with dotnet nomenclature and structure.

### Description

A `KernelPluginCollection` is introduced with an underlying `DefaultKernelPlugin` and a `KernelPlugin` base class. Other code is refactored to use this new `KernelPluginCollection` model which should simplify how plugins are handled. Unit tests are added to test the new functionality.  Existing unit tests and integration tests, along with kernel examples and notebooks were updated based on these changes. 

Unlike before, we're requiring plugins to be defined with a name, which removes the GLOBAL_PLUGIN constant name of "__GLOBAL_PLUGIN__". In some senses, while creating a plugin or a function, a name may not be provided and so a random name in the form of a plugin (p_<random_ascii_chars>) or function (f_<random_ascii_chars>) is used. 

### Contribution Checklist


python/semantic_kernel/kernel.py
```diff
@@ -102,24 +104,56 @@ def prompt_template_engine(self) -> PromptTemplatingEngine:
...
-    def plugins(self) -> ReadOnlyPluginCollectionBase:
-        return self._plugin_collection.read_only_plugin_collection
+    def plugins(self) -> KernelPluginCollection:
+        return self._plugins
+
+    @plugins.setter
+    def plugins(self, value: KernelPluginCollection) -> None:
+        self._plugins = value
+
+    def add_plugin_to_collection(self, plugin_name: str, functions: List[KernelFunctionBase]) -> None:
```
juliomenendez, line 114: Would it make sense to also allow to add an existing `KernelPlugin` instance?

python/semantic_kernel/kernel.py
```diff
@@ -763,12 +816,21 @@ def import_semantic_plugin_from_directory(
             # Prepare lambda wrapping AI logic
             function_config = SemanticFunctionConfig(config, template)
 
-            plugin[function_name] = self.register_semantic_function(
-                plugin_directory_name, function_name, function_config
-            )
+            # TODO: this is an example of where plugins are added to the collection in the kernel
+            # as part of the register_semantic_function, seems weird to have it hidden?
+            # should the register function simply register the function and then we can add to the
+            # plugin collecton later?
```
juliomenendez, line 822: typo: collecton vs collection

python/semantic_kernel/plugin_definition/__init__.py
```diff
@@ -3,8 +3,11 @@
     kernel_function_context_parameter,
 )
 from semantic_kernel.plugin_definition.kernel_function_decorator import kernel_function
+from semantic_kernel.plugin_definition.kernel_plugin import KernelPlugin
 
 __all__ = [
     "kernel_function",
     "kernel_function_context_parameter",
+    "KernelPlugin",
+    "KernelPluginCollection" "DefaultKernelPlugin",
```
juliomenendez, line 12: I think this is missing a `,` after `"KernelPluginCollection"`.

python/semantic_kernel/plugin_definition/default_kernel_plugin.py
```diff
@@ -0,0 +1,122 @@
...
+
+        Raises:
+            ValueError: If the functions list contains duplicate function names.
+        """
+        functions_list = values.get("functions", [])
+        seen = set()
+        functions_dict = {}
+
+        for function in functions_list:
+            if function.name in seen:
```
juliomenendez, line 45: You shouldn't need the `seen` set. The same can be done with the keys of the `functions_dict`:
```python
if function.name in functions_dict:
  raise ValueError(f"Duplicate function name detected: {function.name}")
```

python/semantic_kernel/plugin_definition/default_kernel_plugin.py
```diff
@@ -0,0 +1,122 @@
...
+            The function if it exists, None otherwise.
+
+        Raises:
+            KeyError: If the function does not exist.
+        """
+        if name not in self.functions:
+            raise KeyError(f"Function {name} not found.")
+        return self.functions[name]
+
+    def from_function(function: "KernelFunctionBase") -> "DefaultKernelPlugin":
```
juliomenendez, line 112: Shouldn't this be a class method?

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,210 @@
...
+            The values, with the plugins input converted to a dictionary.
+
+        Raises:
+            ValueError: If the plugins input is invalid.
+        """
+        plugins_input = values.get("plugins")
+        if isinstance(plugins_input, KernelPluginCollection):
+            # Extract plugins from another KernelPluginCollection instance
+            values["plugins"] = {plugin.name: plugin for plugin in plugins_input.plugins.values()}
+        elif isinstance(plugins_input, (list, set, tuple)):
```
juliomenendez, line 43: Could this be more general using the `Iterable` `ABC` type?
```python
from collections.abc import Iterable

if isinstance(plugins_input, Iterable):
  ...
```

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,210 @@
...
+            functions (List[KernelFunctionBase]): The function to add to the plugin.
+            plugin_name (str): The name of the plugin to add the function to.
+
+        Raises:
+            ValueError: If the function or plugin_name is None or invalid.
+            ValueError: if the function already exists in the plugin.
+        """
+        if not functions or not plugin_name:
+            raise ValueError("Functions and plugin_name must not be None or empty")
+        if plugin_name not in self.plugins:
```
juliomenendez, line 104: Would it make sense to create the plugin if it does not exist?

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,210 @@
...
+        Raises:
+            ValueError: If the plugins list is None.
+        """
+
+        if plugins is None:
+            raise ValueError("Plugins must not be None")
+        for plugin in plugins:
+            self.add(plugin)
+
+    def remove(self, plugin: KernelPlugin) -> bool:
```
juliomenendez, line 130: Would it make sense to also allow to remove a plugin by its name?

python/tests/unit/planning/action_planner/test_action_planner.py
```diff
@@ -84,7 +88,8 @@ async def test_plan_creation():
     plan = await planner.create_plan(goal)
 
     assert plan is not None
-    assert plan.description == mock_function.description
+    # TODO: figure out why the returned plan.description is of type Mock instead of a string
+    # assert plan.description == mock_function.description
```
juliomenendez, line 92: Because `plugins.get_plugin.get_function` is returning a `Mock` that doesn't define what the `description` should be so it just fills it with another `Mock`. Try removing the `Mock` of `plugins` and fully creating that `KernelPluginCollection` instance and adding a plugin with the `mock_function` as its function.

python/tests/unit/plugin_definition/test_kernel_plugin_collection.py
```diff
@@ -0,0 +1,82 @@
...
+from typing import TYPE_CHECKING
+
+import pytest
+
+from semantic_kernel.plugin_definition.default_kernel_plugin import DefaultKernelPlugin
+from semantic_kernel.plugin_definition.kernel_plugin_collection import KernelPluginCollection
+
+if TYPE_CHECKING:
+    pass
+
```
juliomenendez, line 13: Is this needed? Not sure what its doing

juliomenendez (changes requested):

python/samples/kernel-syntax-examples/memory.py
```diff
@@ -61,7 +61,7 @@ async def setup_chat_with_memory(
     context["fact5"] = "what do I do for work?"
 
     context[sk.core_plugins.TextMemoryPlugin.COLLECTION_PARAM] = "aboutMe"
-    context[sk.core_plugins.TextMemoryPlugin.RELEVANCE_PARAM] = 0.8
+    context[sk.core_plugins.TextMemoryPlugin.RELEVANCE_PARAM] = "0.8"
```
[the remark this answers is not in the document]
moonbox3: It does as the `[sk.core_plugins.TextMemoryPlugin.RELEVANCE_PARAM]` was updated to be a string in other parts of the code, and wasn't caught here.

python/semantic_kernel/plugin_definition/__init__.py
moonbox3: Good catch! Thanks.

python/semantic_kernel/plugin_definition/default_kernel_plugin.py
moonbox3: Good call.

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: I like it, thank you

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: That's a good question. I wanted this method to be used to add functions to an existing plugin. Should we allow the plugin to be created at this point if it doesn't exist?

python/tests/unit/planning/action_planner/test_action_planner.py
moonbox3: Ah good catch, as we covered that today. :) I forgot this was pending.

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: I was thinking that passing in the instance of the plugin should be fine, since we require it to have a name, and then name is used later in the method. Do you think we also need to have the ability to only pass in the name?

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
juliomenendez: I think it makes sense instead of raising an error. It's an added feature that would remove the need to always create the plugin ahead of time.

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
juliomenendez: I have 3 use cases in mind that could benefit of being able to remove plugins based on their name:

1. When adding a plugin from a function using `add_plugin_from_function`. That method takes a plugin name. How can the user remove such plugin later on if they only have the plugin name?
2. When adding functions to a plugin using `add_functions_to_plugin`. Same as before, that method takes a plugin name. If we allow that method to create the plugin on the fly as per a previous comment, the user won't have any way of removing the plugin as they only have a name.
3. When creating plugins in bulk, let's say you have a `dict` of plugin names and functions that you loop through to add to the plugin collection. Why force the user to keep track of the plugin instances in order to be able to remove them when the plugin collection already has them? Keeping the plugin names should be enough to manage them in the plugin collection.

I'm not saying we shouldn't allow to remove by plugin instance, I just think that removing by plugin name could be useful too.

moonbox3: I appreciate the detailed explanation! Make sense to me.

moonbox3: For type safety, to emphasize that it's a first-class method of the API, and because the KernelPluginCollection isn't a huge class (right now), I've decided to add a method `remove_by_name` method that takes in the plugin_name (str).

python/semantic_kernel/kernel.py
moonbox3: I have added this in.

juliomenendez (approved):

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    The Kernel Plugin Collection class. This class is used to store a collection of plugins.
+
+    Attributes:
+        plugins (Dict[str, KernelPlugin]): The plugins in the collection, indexed by their name.
+    """
+
+    plugins: Optional[Dict[str, KernelPlugin]] = Field(default_factory=dict)
+
+    @model_validator(mode="before")
```
eavanvalkenburg, line 25: This should maybe just be a init function, that is shown to a developer with a proper docstring etc, which makes it easier to work with I think, a validator should focus on validation of values, more then rewriting.

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,225 @@
...
+            self.plugins.add(DefaultKernelPlugin(name=plugin_name))
+
+        plugin = self.plugins[plugin_name]
+        for func in functions:
+            if func.name not in plugin.functions:
+                plugin.functions[func.name] = func
+            else:
+                raise ValueError(f"Function with name '{func.name}' already exists in plugin '{plugin_name}'")
+
+    def add_range(self, plugins: List[KernelPlugin]) -> None:
```
eavanvalkenburg, line 115: maybe `add_list_of_plugins`, not sure where range comes from...

python/semantic_kernel/plugin_definition/kernel_plugin.py
```diff
@@ -0,0 +1,46 @@
...
+    Attributes:
+
+        name (str): The name of the plugin.
+        description (str): The description of the plugin.
+    """
+
+    name: str
+    description: str = Field(default="")
+
+    @field_validator("name", mode="after")
```
eavanvalkenburg, line 24: field_validation with mode=after is usually not a classmethod, but a regular method with self, so why not annotate the name parameter?

eavanvalkenburg: using this: https://docs.pydantic.dev/latest/api/types/#pydantic.types.StringConstraints

python/semantic_kernel/plugin_definition/kernel_plugin.py
```diff
@@ -0,0 +1,46 @@
...
+
+import re
+from abc import ABC, abstractmethod
+
+from pydantic import Field, field_validator
+
+from semantic_kernel.kernel_pydantic import KernelBaseModel
+
+
+class KernelPlugin(KernelBaseModel, ABC):
```
eavanvalkenburg, line 11: maybe rename to KernelPluginBase?

python/semantic_kernel/plugin_definition/default_kernel_plugin.py
```diff
@@ -0,0 +1,121 @@
...
+        description (str): The description of the plugin.
+        functions (Dict[str, KernelFunctionBase]): The functions in the plugin,
+            indexed by their name. Although this is a dictionary, the user
+            should pass a list of KernelFunctionBase instances when initializing,
+            and they will be automatically converted to a dictionary.
+    """
+
+    functions: Dict[str, "KernelFunctionBase"] = Field(default_factory=dict)
+
+    @model_validator(mode="before")
```
eavanvalkenburg, line 26: similar to comment about kernel_plugin_collection, this makes more sense as a init func.

python/semantic_kernel/plugin_definition/kernel_plugin.py
eavanvalkenburg: but, do we even need this?

python/samples/kernel-syntax-examples/self-critique_rag.py
```diff
@@ -48,7 +48,9 @@ async def main() -> None:
     kernel.add_text_completion_service(
         "dv",
         AzureTextCompletion(
-            deployment_name="gpt-35-turbo-instruct",
```
eavanvalkenburg, line 51: why was this undone?

python/semantic_kernel/plugin_definition/default_kernel_plugin.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Raises:
+            KeyError: If the function does not exist.
+        """
+        if name not in self.functions:
+            raise KeyError(f"Function {name} not found.")
+        return self.functions[name]
+
+    @classmethod
+    def from_function(cls, function: "KernelFunctionBase") -> "DefaultKernelPlugin":
```
eavanvalkenburg, line 111: supply the option to add a PluginName as well?

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,225 @@
...
+            plugin_name (str): The name of the plugin to remove from the collection.
+
+        Returns:
+            True if the plugin was removed, False otherwise.
+        """
+        if plugin_name is None:
+            return False
+        return self.plugins.pop(plugin_name, None) is not None
+
+    def get_plugin(self, name: str) -> Optional[KernelPlugin]:
```
eavanvalkenburg, line 159: should we add a get_function(self, plugin_name: str, function_name: str) here? would make looking up functions for auto-executing tool_calls much easier and cleaner...

python/semantic_kernel/kernel.py
```diff
@@ -63,20 +64,21 @@
 
 
 class Kernel:
-    _plugin_collection: PluginCollectionBase
+    # TODO: pydantic-ify these fields
+    _plugins: KernelPluginCollection
```
eavanvalkenburg, line 68: no time like the present, `plugins: KernelPluginCollection = Field(default_factory=KernelPluginCollection)` and then remove the property

python/semantic_kernel/kernel.py
```diff
@@ -763,12 +823,21 @@ def import_semantic_plugin_from_directory(
...
+            # as part of the register_semantic_function, seems weird to have it hidden?
+            # should the register function simply register the function and then we can add to the
+            # plugin collection later?
+            functions += [self.register_semantic_function(plugin_directory_name, function_name, function_config)]
+
+        plugin = DefaultKernelPlugin(name=plugin_directory_name, functions=functions)
 
         return plugin
 
+    def generate_random_ascii_name(self, length=16):
```
eavanvalkenburg, line 836: let's make this a static method or move to a utils file!

eavanvalkenburg (changes requested):
Initial thoughts added, will go through more detail later!

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: I love the implicit constructor when using Pydantic, but I agree that it may be more useful to have an actual __init__ constructor.

python/semantic_kernel/plugin_definition/kernel_plugin.py
moonbox3: dotnet has an ABC of `KernelPlugin`, so am trying to keep it to maintain parity. There can be other plugins created that are based on the ABC `KernelPlugin`.

python/samples/kernel-syntax-examples/self-critique_rag.py
moonbox3: in the subscription that is used for integration tests, we don't have a `gpt-35-turbo-instruct` model defined. I was testing it, and can revert.

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: Is the method chaining like: `plugin_collection.get_plugin(plugin_name).get_function(function_name)` not adequate? You'd rather see `plugin_collection.get_function(plugin_name, function_name`?

python/semantic_kernel/plugin_definition/kernel_plugin.py
moonbox3: I guess we can just get rid of it, and move the name/description fields into the DefaultKernelPlugin class. Does this align with your thinking?

python/semantic_kernel/plugin_definition/kernel_plugin.py
moonbox3: Very clean, thanks. Forgot about Pydantic's Annotated.

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: I've added this.

python/semantic_kernel/plugin_definition/kernel_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+    from typing import Annotated
+else:
+    from typing_extensions import Annotated
+
+from pydantic import Field, StringConstraints
+
+from semantic_kernel.kernel_pydantic import KernelBaseModel
+
+
+class KernelPluginBase(KernelBaseModel, ABC):
```
eavanvalkenburg, line 17: the file name should be changed to kernel_plugin_base

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            raise ValueError("Invalid type for plugins")
+
+        super().__init__(plugins=plugins)
+
+    @staticmethod
+    def _process_plugins_iterable(plugins_input: Iterable[KernelPluginBase]) -> Dict[str, KernelPluginBase]:
+        plugins_dict = {}
+        for plugin in plugins_input:
+            if plugin is None or plugin.name is None:
```
eavanvalkenburg, line 57: according to KernelPluginBase name cannot be none or empty, so no need to double check here.

python/semantic_kernel/plugin_definition/kernel_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    The KernelPluginBase Base Class. All plugins must inherit from this class.
+
+    Attributes:
+
+        name (str): The name of the plugin.
```
eavanvalkenburg, line 23: maybe add the requirements for the name here

python/semantic_kernel/plugin_definition/default_kernel_plugin.py
```diff
@@ -0,0 +1,122 @@
...
+from semantic_kernel.orchestration.kernel_function_base import KernelFunctionBase
+from semantic_kernel.plugin_definition.kernel_plugin import KernelPluginBase
+
+
+class DefaultKernelPlugin(KernelPluginBase):
+    """
+    Represents a Kernel Plugin with functions.
+
+    Attributes:
+        name (str): The name of the plugin.
```
eavanvalkenburg, line 16: same as in base, add the requirements for the name here

eavanvalkenburg (commented):
didn't go through the samples, notebooks and tests but made some notes on the core of the PR

python/semantic_kernel/plugin_definition/kernel_plugin.py
moonbox3: Good catch!

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+        Add a single plugin to the collection
+
+        Args:
+            plugin (KernelPluginBase): The plugin to add to the collection.
+
+        Raises:
+            ValueError: If the plugin or plugin.name is None.
+        """
+        if plugin is None or plugin.name is None:
```
eavanvalkenburg, line 74: can plugin.name be none?

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
```diff
@@ -0,0 +1,249 @@
...
+            plugin_name (str): The name of the plugin to add the function to.
+
+        Raises:
+            ValueError: If the function or plugin_name is None or invalid.
+            ValueError: if the function already exists in the plugin.
+        """
+        if not functions or not plugin_name:
+            raise ValueError("Functions and plugin_name must not be None or empty")
+        if plugin_name not in self.plugins:
+            self.plugins.add(DefaultKernelPlugin(name=plugin_name))
```
eavanvalkenburg, line 114: can't you add the functions directly to this plugin creation?

eavanvalkenburg (changes requested):
some comments, I think we can make things much more pythonic by leveraging built-in methods for more things.

python/semantic_kernel/kernel.py
```diff
@@ -63,20 +65,21 @@
 
 
 class Kernel:
```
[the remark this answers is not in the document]
moonbox3: I think it could be, but it seems out of scope to modify the kernel -> pydantic class in this PR? I can file a work item to do that next.

python/semantic_kernel/kernel.py
```diff
@@ -392,12 +431,24 @@ def on_function_invoked(self, function_view: FunctionView, context: KernelContex
...
-            plugin_name = PluginCollection.GLOBAL_PLUGIN
-            logger.debug(f"Importing plugin {plugin_name} into the global namespace")
-        else:
-            logger.debug(f"Importing plugin {plugin_name}")
+    def import_plugin(self, plugin_instance: Any, plugin_name: str) -> KernelPluginBase:
+        """
+        Import a plugin into the kernel.
+
+        Args:
+            plugin_instance (Any): The plugin instance.
```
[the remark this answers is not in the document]
moonbox3: I think it'd be nice if we defined a base class that plugins inherited, so we could be more explicit on the type. Right now, it can be any class that contains methods with the @kernal_function decorator.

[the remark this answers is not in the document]
eavanvalkenburg: ok, then let's start with adding that sentence to the docstring!

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: Technically yes, but I want the code below this to process each function to check for duplicate function names (which is not allowed).

python/semantic_kernel/plugin_definition/default_kernel_plugin.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Raises:
+            KeyError: If the function does not exist.
+        """
+        if name not in self.functions:
+            raise KeyError(f"Function {name} not found.")
+        return self.functions[name]
+
+    @classmethod
+    def from_function(cls, function: "KernelFunctionBase", plugin_name: Optional[str] = None) -> "DefaultKernelPlugin":
```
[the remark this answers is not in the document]
moonbox3: Going to update this to allow for a List of functions instead of a single function.

python/semantic_kernel/plugin_definition/kernel_plugin_collection.py
moonbox3: Talked offline. Update incoming.

eavanvalkenburg (approved):
Some small questions mostly.

python/semantic_kernel/kernel.py
```diff
@@ -335,10 +374,7 @@ async def run(
         return context
 
     def func(self, plugin_name: str, function_name: str) -> KernelFunctionBase:
-        if self.plugins.has_native_function(plugin_name, function_name):
-            return self.plugins.get_native_function(plugin_name, function_name)
-
-        return self.plugins.get_semantic_function(plugin_name, function_name)
+        return self.plugins[plugin_name][function_name]
```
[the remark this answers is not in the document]
moonbox3: It'll return a KeyError. I wanted to be explicit about it.

python/semantic_kernel/kernel.py
```diff
@@ -392,19 +428,30 @@ def on_function_invoked(self, function_view: FunctionView, context: KernelContex
...
+            logger.warn("Unable to import plugin due to missing plugin_name")
+            raise KernelException(
+                KernelException.ErrorCodes.InvalidPluginName,
+                "Plugin name cannot be empty",
+            )
+        logger.debug(f"Importing plugin {plugin_name}")
 
         functions = []
 
-        if isinstance(plugin_instance, dict):
```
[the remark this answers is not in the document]
moonbox3: I didn't see any other code paths that ever passed in a Dict of plugin_instances. Do we need to keep that functionality?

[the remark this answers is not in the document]
moonbox3: I left it in...

python/semantic_kernel/kernel.py
```diff
@@ -769,9 +821,13 @@ def import_semantic_plugin_from_directory(
...
             function_config = SemanticFunctionConfig(config, template)
 
-            plugin[function_name] = self.register_semantic_function(
-                plugin_directory_name, function_name, function_config
-            )
+            # TODO: this is an example of where plugins are added to the collection in the kernel
+            # as part of the register_semantic_function, seems weird to have it hidden?
+            # should the register function simply register the function and then we can add to the
+            # plugin collection later?
+            functions += [self.register_semantic_function(plugin_directory_name, function_name, function_config)]
```
eavanvalkenburg, line 828: I would use `functions.extend(self.reg...` over +=

eavanvalkenburg (approved):
Some small notes, nothing breaking.

python/semantic_kernel/kernel.py
moonbox3: Let me address in a subsequent PR refactor if needed.

Issue #4680, opened 2024-01-19
.Net GoogleAI connector progress (Gemini)


Krzysztof318:
### Motivation and Context

https://github.com/microsoft/semantic-kernel/issues/4334

### Description

Progress of implementation of Google Gemini connector.
Current implementation [GEMINI BRANCH](https://github.com/microsoft/semantic-kernel/tree/feature-connectors-gemini)

### TODO

1. TextGenerationService

- [x] Get text (#4742)
    - [x] Create examples (#4873)
- [x] Stream text (#4742)
    - [x] Create examples (#4873)
- [x] Add DI extensions (#4766)
- [x] Provide metadata from response
- [x] Function calling (#5044)
    - [x] Create examples

2. ChatCompletionService

- [x] Get text (#4742)
    - [x] Create examples (#4873)
- [x] Stream text (#4742)
    - [x] Create examples (#4873)
- [x] Add DI extensions (#4766)
- [x] Provide metadata from response (#4677)
- [x] Function calling (#5044)
- [x] Pass images with prompt (#4957) (#5105)

3. Others
- [x] Embeddings (#4745, #4758, #4786)
- [x] Counting tokens (added without public service) (#4677)
- [x] Vertex AI implementation
- [x] Logging
- [x] Telemetry

Notes:
- Token count in a metadata is provided only for the VertexAI, for the GoogleAI is zero,
- System messages are not supported by gemini models. If a chat history contains only one system message, it will be converted to the user message. If there are more than one system message an exception will be thrown.
- GoogleAI isn't available in the GDPR countries. If you want to use gemini in these countries, you need vpn or use VertexAI which works.
- VertexAI supports gemini-vision via both an URI and a BinaryData, GoggleAI supports only BinaryData. When you pass an image as URI you need to pass a mime_type as an ImageContent.MimeType.

Tested models:
1. GoogleAI
    - gemini-pro (text/chat generation/stream)
    - gemini-pro-vision (chat generation/stream with vision capability) (binary data only)
    - embedding-001 (embeddings)
2. VertexAI
    - gemini-pro (text/chat generation/stream)
    - gemini-pro-vision (chat generation/stream with vision capability)
    - textembedding-gecko (embeddings)

SavoInfostream:
@Krzysztof318  Thanks for the info. Is an alpha nuget  going to be available before, maybe?

Krzysztof318:
[...]

PR #4765, opened 2024-01-27
.Net Gemini refactor chat system messages handling

Krzysztof318:
### Motivation and Context

~~Gemini chat client could handle only one system message. Exception was thrown if there were more than one system message.~~
Lacks of some tests, and ref to chathistory param.

### Description

~~Fixes handling multiple system messages. Added revelant unit tests.~~
Refactoring and new unit tests.
State of connector progress https://github.com/microsoft/semantic-kernel/issues/4680

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Common/GeminiChatCompletionClient.cs
+++ b/dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Common/GeminiChatCompletionClient.cs
@@ -105,13 +105,13 @@ internal class GeminiChatCompletionClient : GeminiClient, IGeminiChatCompletionC
                 throw new KernelException("Chat history can't contain only system messages.");
             }
 
-            chatHistory = PrepareChatHistoryWithSystemMessages(chatHistory, systemMessages);
+            MergeSystemMessagesToOneUserMessageInChatHistory(chatHistory, systemMessages);
         }
 
         ValidateChatHistoryMessagesOrder(chatHistory);
     }
 
-    private static ChatHistory PrepareChatHistoryWithSystemMessages(ChatHistory chatHistory, List<ChatMessageContent> systemMessages)
+    private static void MergeSystemMessagesToOneUserMessageInChatHistory(ChatHistory chatHistory, List<ChatMessageContent> systemMessages)
     {
         // TODO: This solution is needed due to the fact that Gemini API doesn't support system messages. Maybe in the future we will be able to remove it.
         var systemMessageBuilder = new StringBuilder();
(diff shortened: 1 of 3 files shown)


Krzysztof318:
@RogerBarreto

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Common/GeminiChatCompletionClient.cs
```diff
@@ -56,7 +57,7 @@ internal class GeminiChatCompletionClient : GeminiClient, IGeminiChatCompletionC
         PromptExecutionSettings? executionSettings = null,
         CancellationToken cancellationToken = default)
     {
-        ValidateAndPrepareChatHistory(chatHistory);
+        ValidateAndPrepareChatHistory(ref chatHistory);
```
RogerBarreto, line 60: I don't think `ref` is necessary as `ChatHistory` is already a class.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Common/GeminiChatCompletionClient.cs
```diff
@@ -91,35 +92,48 @@ internal class GeminiChatCompletionClient : GeminiClient, IGeminiChatCompletionC
...
     {
         Verify.NotNullOrEmpty(chatHistory);
 
-        if (chatHistory.SingleOrDefault(message => message.Role == AuthorRole.System) is { } systemMessage)
+        chatHistory = new ChatHistory(chatHistory);
+
+        if (chatHistory.Where(message => message.Role == AuthorRole.System).ToList() is { Count: > 0 } systemMessages)
         {
-            if (chatHistory.Count == 1)
+            if (chatHistory.Count == systemMessages.Count)
```
RogerBarreto, line 103: Confusing in this part, if multiple can become a user message, why having only one gives an exception, to be concise with the behavior, I would expect that single system to become an user message.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Common/GeminiChatCompletionClient.cs
Krzysztof318: It is necessary because I reassigned variable chat history in ValidateAndPrepareChatHistory method with copy of chathistory. This is deliberate in order to manipulate the chathistory object without altering the passed object.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Common/GeminiChatCompletionClient.cs
Krzysztof318: It doesn't matter is none/one/many system messages in chat history. Exception would only be thrown in case chat history contains only system messages.
System message one or more always is converted to one user message.

Issue #4578, opened 2024-01-11
.Net: ADR for Semantic Kernel Memory Strategy


markwallace-microsoft:
The purpose of this task is to create an ADR to describe our memory strategy for Semantic Kernel.

Below are a series of user stories that must be addressed by the ADR.

### Strategy
As a developer using Semantic Kernel there are well defined options to integrate with vector memory stores so that I can implement [Retrieval Augmented Generation](https://en.wikipedia.org/wiki/Prompt_engineering#Retrieval-augmented_generation) (RAG) in my application.

1. Our strategy should allow us deprecate the existing connectors in favour of integrating using the [ChatGPT Retrieval Plugin](https://github.com/openai/chatgpt-retrieval-plugin).
1. Our strategy to should allow for the use of any .Net SDK's that provide access to vector memory stores.

### Contributing an Memory Connector
As a contributor to Semantic Kernel I can read the documentation which describes how to contribute an memory connector so that I can understand the development process and requirements to have a new memory connector included in a Semantic Kernel release

1. Describe how we successfully co-develop memory connectors with the community. 
1. We want to avoid receiving huge PR's that are very difficult to review.
1. We want to ensure each AI connector includes a defined level of unit tests and integration tests.

Issue #3455, opened 2023-11-11
.Net: Integrate ChatGPT's retrieval plugin capabilities in SK


markwallace-microsoft:
```[tasklist]
### Tasks
- [ ] SK throws exceptions during the import of the ChatGPT retrieval plugin, fix is proposed here: #3452
- [ ] SK also throws exceptions when attempting to store some data in memory, specifically when executing the "memory.upsert_upsert_post" function, with response status code of 422 (Unprocessable Entity), caused by the input parameters passed to the plugin.
- [ ] ...
```

PR #4771, opened 2024-01-28
.Net: ADR for RAG pattern in SK

dmytrostruk:
Resolves: https://github.com/microsoft/semantic-kernel/issues/4578
Resolves: https://github.com/microsoft/semantic-kernel/issues/3455

1. Added ADR which describes different approaches how to use RAG pattern in current version of Semantic Kernel and a couple of new possible approaches, which could be added in the future.
2. Added examples how to use RAG pattern with ChatGPT Retrieval Plugin, TextMemoryPlugin and custom plugin.
3. Small fixes to TextMemoryPlugin.


docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+
+// User is responsible for searching the data in a way of their choice, this is an example how it could look like.
+var data = await this._vectorDB.SearchAsync("Company budget by year");
+
+var arguments = new KernelArguments { ["budgetByYear"] = data };
+
+var result = await kernel.InvokePromptAsync("{{budgetByYear}} What is my budget for 2024?", arguments);
+```
+
+### Option 2 [Supported] - Memory as Plugin
```
matthewbolanos, line 141: My preference would be to support Option 2 as the main/preferred route since it strengthens the plugin story and also leverages prompt templates the primary asset (which would allow RAG templates to be shared between Python/.NET/Java)

docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+var kernel = Kernel.CreateBuilder()
+    .AddOpenAIChatCompletion("model-id", "api-key")
+    .Build();
+
+kernel.ImportPluginFromType<MyDataPlugin>();
+
+var result = await kernel.InvokePromptAsync("{{search 'Company budget by year'}} What is my budget for 2024?");
+```
+
+The reason why custom user plugin is more flexible than `TextMemoryPlugin` is because `TextMemoryPlugin` requires all vector DBs to implement `IMemoryStore` interface with disadvantages described above, while custom user plugin can be implemented in a way of developer's choice. There won't be any restrictions on DB record schema or requirement to implement specific interface.
```
matthewbolanos, line 219: Is there a way to improve `TextMemoryPlugin` so that it doesn't have as many of the issues? I agree that a "custom" plugin is the best escape valve (that's what most/all of our customers do to get hybrid search support), but hopefully we can make this a more rare occurrence.

docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+
+// Import ChatGPT Retrieval Plugin using OpenAPI specification
+// https://github.com/openai/chatgpt-retrieval-plugin/blob/main/.well-known/openapi.yaml
+await kernel.ImportPluginFromOpenApiAsync("ChatGPTRetrievalPlugin", openApi!, executionParameters: new(authCallback: async (request, cancellationToken) =>
+{
+    request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", "chat-gpt-retrieval-plugin-token");
+}));
+
+const string Query = "What is my budget for 2024?";
+const string Prompt = "{{ChatGPTRetrievalPlugin.query_query_post queries=$queries}} {{$query}}";
```
matthewbolanos, line 164: The ChatGPT retrieval plugin has different functions (name and behavior) when compared with our `TextMemoryPlugin`. Would it be reasonable to align so that developers could easily swap? The ChatGPT retrieval plugin has the following functions:

- `upsert_file`
- `upsert`
- `query`
- `delete`

https://github.com/openai/chatgpt-retrieval-plugin/blob/main/server/main.py

docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+kernel.PromptFilters.Add(new MyPromptFilter());
+
+var result = await kernel.InvokePromptAsync("What is my budget for 2024?");
+```
+
+From the usage perspective, prompt will contain just user query without additional data. The data will be added to the prompt behind the scenes.
+
+The reason why this approach is **partially supported** is because a call to vector DB most probably will be an asynchronous, but current Kernel filters don't support asynchronous scenarios. So, in order to support asynchronous calls, new type of filters should be added to Kernel: `IAsyncFunctionFilter` and `IAsyncPromptFilter`. They will be the same as current `IFunctionFilter` and `IPromptFilter` but with async methods.
+
+### Option 4 [Proposal] - Memory as part of PromptExecutionSettings
```
matthewbolanos, line 265: Is this similar to what Azure Open AI "with your data" does? Or is that different?

docs/decisions/0033-rag-in-sk.md
dmytrostruk: That makes sense!

docs/decisions/0033-rag-in-sk.md
dmytrostruk: There is always space for improvement :) 
I think the main disadvantage of `TextMemoryPlugin` is the requirement for all integrations to implement the same interface and have the same data schema, which is not very flexible. I think it works well for demonstration/experimentation purposes, but for more flexibility and easier memory-related extension in the future, personally I would probably go with custom plugin path, configure and use it in a way that covers my scenario the best.

In short:
- TextMemoryPlugin - easier setup, less flexibility
- Custom plugin - harder setup, more flexibility

docs/decisions/0033-rag-in-sk.md
dmytrostruk: While ChatGPT retrieval plugin contains methods to upsert, query and delete data, the OpenAPI specification, which we are using to import it as plugin, contains only `query` method described:
https://github.com/openai/chatgpt-retrieval-plugin/blob/main/.well-known/openapi.yaml

So, in order to make upsert and delete methods availabile, we will have to modify this OpenAPI specification and add remaining methods to it.

docs/decisions/0033-rag-in-sk.md
dmytrostruk: If you are talking about our implementation of Azure OpenAI "with your data" functionality, then yes, it's kind of similar. While "with your data" functionality accepts parameters like `endpoint`, `api key` and `index name`, in this proposed approach we will accept an instance of memory connector with configured parameters. Again, the disadvantage would be a requirement for all memory connectors to implement the same interface and have the same data schema, which is not very flexible. 

But this is still another approach, which could be potentially added for diversity of RAG development experience.

docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+
+All functions can be called directly from prompt. Moreover, as soon as these functions are registered in Kernel and Function Calling is enabled, LLM may decide to call specific function to achieve provided goal.
+
+`Retrieve` and `Recall` functions are useful to provide some context to LLM and ask a question based on data, but functions `Save` and `Remove` perform some manipulations with data in vector DB, which could be unpredicted or sometimes even dangerous (there should be no situations when LLM decides to remove some records, which shouldn't be deleted).
+
+## Decision Drivers
+
+1. All manipulations with data in Semantic Kernel should be safe.
+2. There should be a clear way(s) how to use RAG pattern in Semantic Kernel.
+3. Abstractions should not block developers from using vector DB of their choice with functionality, that cannot be achieved with provided interfaces or data types.
```
roji, line 77: I think that's true, but it's also important to realize that any abstraction here is going to block *some* subset of vector database functionality; that is, even if we evolve the IMemoryStorage interface to support a more powerful/flexible data scheme, there are still many other aspects beyond that which won't be accessible when using the abstraction. In other words, to get the full power/flexibility/feature-set of your specific chosen vector db solution, you most likely will need to drop the abstraction and use its SDK directly.

Just saying this because it's important to not create hopes for a better abstraction that will fully "solve" this problem. For comparison, relational databases seem generally much more similar to each other than vector databases are (there's a SQL standard etc.), and even there the abstraction (ADO.NET in the .NET case) doesn't actually allow fully working with them without things leaking and certain features being blocked.

docs/decisions/0033-rag-in-sk.md
roji: I agree with the general analysis here, i.e. that IMemoryStore imposes certain limitations that may lead some users to switch to direct interaction with their vector DB, foregoing the abstraction. However, it would be good to understand how prevalent that actually is... In many cases, requirements from metadata stored in the vector db simply aren't that advanced/specific, and the somewhat simplistic IMemoryStore abstraction isn't an actual hindrance.

Otherwise, I think it's not a bad story to have the abstraction as a way to get up and running really quickly - without having to mess around with a custom plugin and understanding the specific vector DB's SDK; then, as new requirements come in, some users may "graduate" out of that story to interact directly with the vector DB, if they really need to.

docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+    .Build();
+
+kernel.PromptFilters.Add(new MyPromptFilter());
+
+var result = await kernel.InvokePromptAsync("What is my budget for 2024?");
+```
+
+From the usage perspective, prompt will contain just user query without additional data. The data will be added to the prompt behind the scenes.
+
+The reason why this approach is **partially supported** is because a call to vector DB most probably will be an asynchronous, but current Kernel filters don't support asynchronous scenarios. So, in order to support asynchronous calls, new type of filters should be added to Kernel: `IAsyncFunctionFilter` and `IAsyncPromptFilter`. They will be the same as current `IFunctionFilter` and `IPromptFilter` but with async methods.
```
roji, line 263: Good point, that sounds like something that may need improvement indeed. Manipulating/filtering the prompt based on information/decision-making that needs to be accessed asynchronously sounds like an important capability.

roji (commented):
Great write-up, posted some comments/thoughts below.

stephentoub:
> Otherwise, I think it's not a bad story to have the abstraction as a way to get up and running really quickly - without having to mess around with a custom plugin and understanding the specific vector DB's SDK; then, as new requirements come in, some users may "graduate" out of that story to interact directly with the vector DB, if they really need to.

I agree. There's significant value to having the abstraction, and we'd be doing folks a disservice by getting rid of it even if you can't do everything with it. You can do a lot with it (and we should ensure it's as good / full-featured as possible), including bootstrapping / getting up and running, and in a way that lets you postpone choice, i.e. you can write code without committing to a specific db, you can experiment, you can get your feet under you, and then if you later find you need to go beyond what the abstraction offers, you can break glass and use direct functionality on a specific db you commit to at that point.

docs/decisions/0033-rag-in-sk.md
dmytrostruk: That makes sense! I think if we add more advanced examples how to quickly use memory with our abstractions and how to interact directly with the vector DB for better flexibility, we should cover most of the cases for users.

docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+`IMemoryStore` interface uses `MemoryRecord` class as storage schema in vector DB. This means that `MemoryRecord` properties should be aligned to all possible connectors. As soon as developers will use this schema in their databases, any changes to schema may break the application, which is not flexible approach.
+
+`MemoryRecord` contains property `ReadOnlyMemory<float> Embedding` for embeddings and `MemoryRecordMetadata Metadata` for embeddings metadata. `MemoryRecordMetadata` contains properties like:
+
+- `string Id` - unique identifier.
+- `string Text` - data-related text.
+- `string Description` - optional title describing the content.
+- `string AdditionalMetadata` - field for saving custom metadata with a record.
+
+Since `MemoryRecord` and `MemoryRecordMetadata` are not sealed classes, it should be possible to extend them and add more properties as needed. Although, current approach still forces developers to have specific base schema in their vector DBs, which ideally should be avoided. Developers should have the ability to work with any schema of their choice, which will cover their business scenarios (similarly to Code First approach in Entity Framework).
```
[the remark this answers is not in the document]
dmytrostruk: > If the developer extends MemoryRecord/MemoryRecordMetadata, I'm assuming it isn't actually expected that SK IMemoryStore implementations actually pick those changes up (via reflection), is it? At least, that's not what I saw some/most implementations do - they just hard-code the several known metadata fields, nothing more.

That's correct, additional mechanism should be added to take extended classes into account. Even current implementations don't take into account all fields from MemoryRecord/MemoryRecordMetadata. Ideally, developer should have control over it, and connector should just construct a request based on all properties that were passed to it.

> That might mean that the SK abstraction should also go in a more weakly-typed, JSON-oriented direction; this would allow users to have whatever data shapes they want in their metadata.

I think the ideal scenario would be to support strongly-typed schema and weakly-typed JSON-based approach to cover both cases. I will probably investigate in more detail which approach is used in each connector and what is supported by vector DB out-of-the-box, to have a better understanding.

docs/decisions/0033-rag-in-sk.md
dmytrostruk: That's a good example. I agree, as discussed in other comments, it sounds like keeping an abstraction should allow developers to handle basic scenarios quickly, while for full flexibility it will be possible to use vector DB directly (e.g. as plugin).

docs/decisions/0033-rag-in-sk.md
dmytrostruk: Yes, I think we will add support for asynchronous filters independently from memory-related work, since it's important feature that should cover asynchronous scenarios for filtering.

docs/decisions/0033-rag-in-sk.md
```diff
@@ -0,0 +1,297 @@
...
+{
+    // Other properties...
+    "status": {
+        "ready": false,
+        "state": "Initializing"
+    }
+}
+```
+
+In this case, it's impossible to insert a record to database immediately, so HTTP polling or similar mechanism should be implemented to cover this scenario.
```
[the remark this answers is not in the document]
roji: > Especially, when there is no .NET SDK available for vector database.

This is specifically a point that should gradually go away - the major vector databases should have good .NET SDKs. But to your general point, yeah, in some cases the SK IMemoryStore won't correspond directly to lower-level operations and will need to do extra stuff (like implement retry polling etc.).

docs/decisions/0033-rag-in-sk.md
BLaZeKiLL: To expand over the points made and point 5 in the out-of-scope section, My app (which uses Neo4j) required the modification of the vector retrieval query at runtime. I ended up creating the [neo4j connector](https://www.nuget.org/packages/Codeblaze.SemanticKernel.Connectors.Memory.Neo4j), which takes in a query factory that can be overridden, giving control to the user over db operations done by the semantic kernel.

The biggest thing that felt odd to me was that I only required a `Read Only Memory Connector` since we have a separate dashboard and API for writing to the database and don't want our AI-enabled APIs to have any write capabilities over the database. Extending such a read-only connector would allow integration with traditional search API.

Consider a use case where I am building a discord bot that references existing messages sent by other users on the channel. In such a case, I don't require embedding search as I can use the existing discord search API and retrieve the required messages.

docs/decisions/0033-rag-in-sk.md
matthewbolanos: Maybe put another way... if it's called `query` in the ChatGPT plugin, is there an opportunity to rename our `recall` function to `query`? That way, a user could "swap" without having to change any of their prompts (as long as the maintained the same plugin name during import)?

docs/decisions/0033-rag-in-sk.md
matthewbolanos: I think at the very least, however, that we should provide an ability within our connectors to provide a custom schema. In customer conversations, I've run into quite a few scenarios where they just want to query an index they've already created previously within Semantic Kernel.

I think this is important because it allows devs coming from other frameworks (e.g., LangChain) to quickly test our "getting started" abstractions without having to rebuild their index.

docs/decisions/0033-rag-in-sk.md
matthewbolanos: Also, it looks like they do have an example of how to expose the other endpoints (e.g., upsert for memory), so modifying the OpenAPI specification doesn't seem outside the realm of possibility.

docs/decisions/0033-rag-in-sk.md
matthewbolanos: I think a good litmus test is that a developer should (at the very least) be able to take an index they've built for LangChain and/or the ChatGPT Retrieval Plugin and be able to quickly do "simple" queries using our connectors. I think this test would encourage the following principels:
1. Devs should be able to bring their own schema
2. We should try to avoid requiring "non-standard" fields (e.g., IsReferenceField) that make import difficult/impossible (if someone needs "advanced" features like this, they should build a custom connector)
3. "Easy" optional configuration should still be possible on a connector-by-connector basis (for example, we should have config for Azure AI Search that has similar behavior as the [ChatGPT Retrieval plugin env variables for Azure AI Search](https://github.com/openai/chatgpt-retrieval-plugin/blob/main/docs/providers/azuresearch/setup.md))

roji: > The biggest thing that felt odd to me was that I only required a Read Only Memory Connector since we have a separate dashboard and API for writing to the database and don't want our AI-enabled APIs to have any write capabilities over the database.

FWIW this is a very odd use-case for me... if you need actual access control (allow read but not write), then that's a feature you need from the database itself; just hiding the write APIs at the SDK level does very little, and can be easily circumvented if someone wants to do that. A read-only SDK/driver is only to give the illusion of providing access control without actually doing it, which is in fact more dangerous.

On the other hand, someone may want a read-only SDK not for security purposes, but just to prevent **accidental** invocation of write APIs, as a very defensive way to guard against programmer error. That's very easy to do without writing dropping the abstraction and writing an SDK/connector - you can simply have a trivial ReadOnlyMemoryStore which wraps an existing Semantic Kernel IMemoryStore, delegates the read APIs to the wrapped memory store and throws from the write APIs. If that's deemed to be actually useful to a lot of people (not too sure about that), that's something that can be added to SK itself.

So I think the read-only access is a rather weak case for foregoing the abstraction.

roji: > "Easy" optional configuration should still be possible on a connector-by-connector basis

Remember that the abstraction covers using the vector DB (reading/writing), but you're still the one configuring and instantiating the IMemoryStore implementation. In other words, you should be able to do up-front configuration of your specific vector DB in whichever way you want, and then the resulting IMemoryStore can be used with the standard SK APIs etc. So here again, I don't see where the abstraction limits configurability.

docs/decisions/0033-rag-in-sk.md
roji: Sure, that may indeed be a thing that's too limiting/restrictive with the current IMemoryStore design - I don't necessarily disagree.

docs/decisions/0033-rag-in-sk.md
dmytrostruk: > So I think the read-only access is a rather weak case for foregoing the abstraction.

@roji We discussed it with SK team, and it looks like having an abstraction for read-only operations (like get a record by key or get a record using vector search) should be enough at this point, because the main idea behind memory connectors is the ability to import it as plugin and then re-use it directly in prompts for RAG scenarios, using following prompt as example:

Prompt: `{{search 'company budget by year'}} What is my budget for 2024?` 
LLM Response: `Your budget for 2024 is ...`

As for write operations, we don't expect users to save data or perform any kind of data manipulation or DB architecture changes (like creating collections/indexes) from prompt. For data operations like `add`, `update`, `remove` - users should probably handle it separately from Kernel.

So, we are thinking about leaving current `IMemoryStore` as it is, with read and write methods for experimentation purposes and quick setup, while providing new lightweight abstraction for search operations, that will support custom schema for more flexibility, so it will allow to use RAG pattern and perform vector search for already existing indexes/collections in user databases.

What do you think about it?

docs/decisions/0033-rag-in-sk.md
dmytrostruk: > Maybe put another way... if it's called `query` in the ChatGPT plugin, is there an opportunity to rename our `recall` function to `query`? That way, a user could "swap" without having to change any of their prompts (as long as the maintained the same plugin name during import)?

Yes, sorry, I misunderstood your initial question. I think it makes sense to rename `recall` method in current `TextMemoryPlugin` or create new plugin that will use `query` method name or something similar - `retrieve`, `search` etc.

docs/decisions/0033-rag-in-sk.md
roji: I think I'd try to separate the two problems, i.e. the custom schema question, and the read/write vs. read-only question... I'd recommend trying to avoid introducing a **new** abstraction, since that would require reimplementing connectors for all the vector databases, and then maintaining the two implementations for each one etc.

Instead, if the goal is to support flexible schemas, I'd rather see about how IMemoryStore can be evolved to support that (IIRC the abstraction is still considered experimental and therefore can be changed in breaking ways, right?). This might indeed be tricky given that you have quite a few implementations which you ideally want to keep working, but I'd at least give this a solid try.

Re read-only, I agree that users are unlikely to need to do data updates from prompt. However, as I wrote above I'm not sure I see a specific problem with using an abstraction (IMemoryStore) in a read-only context (prompt), where that abstraction happens to also support write operations; there's again no actual security aspect here (since just not having the write APIs in the abstraction doesn't provide security, only hiding), so the existence of the write APIs shouldn't be a problem in itself.

In addition, even if you don't need to write from prompt, there is still is value over having write operations in the abstraction: users will be able to easily swap a vector DB (during the test phase) in their data update code, just like they'd be able to do it in their prompt integration logic. There may also be other uses inside SK - outside the prompt - where doing updates to the vector database is useful.

stephentoub: > This might indeed be tricky given that you have quite a few implementations which you ideally want to keep working, but I'd at least give this a solid try.

This is an amazing blessing, not a curse. It's rare when developing an abstraction that you start out with 10-15 ready-made implementations that can be used immediately to vet the design of that abstraction. Demonstrating that it works with all of these will help prove it out immensely. We should only be so lucky in other areas :)

dmytrostruk: > I'd recommend trying to avoid introducing a **new** abstraction, since that would require reimplementing connectors for all the vector databases, and then maintaining the two implementations for each one etc.

So, if we will try to refactor current `IMemoryStore`, this will definitely require refactoring of all currently existing connectors to support it, while new abstraction can be applied to only a couple of them to see how it goes. 

On the other hand, @stephentoub is right, having 12 already implemented connectors will allow us to find common denominator for all of them and create good abstraction.

> IIRC the abstraction is still considered experimental and therefore can be changed in breaking ways, right?

Yes, we should be able to introduce breaking changes for that, since it's experimental at the moment.

markwallace-microsoft (approved):

SergeyMenshykh (approved):

Issue #4762, opened 2024-01-26
Python: Semantic function .invoke() hangs on multiple calls


sarah-widder:
**Describe the bug**
In the Python library, the synchronous call `.invoke()` on semantic functions does not seem to be internally handling the async call correctly. It works for a single call, but calling multiple times in a loop, it seems to freeze when it actually gets to `await self._function` on the subsequent call and execution does not continue.

**To Reproduce**
1. Invoke a semantic function with `.invoke()` in a loop.
2. See execution hangs after several calls.

**Expected behavior**
Each successive call to invoke should complete properly.

**Screenshots**

**Platform**
 - OS: Windows
 - IDE: VSCode
 - Language: Python
 - Source: pip package version 0.4.5.dev0

**Additional context**
This test function reproduces the error for me:
```
def test():
    kernel = sk.Kernel()
    aoai_settings_dict = azure_openai_settings_from_dot_env_as_dict(include_deployment=True, include_api_version=True)
    kernel.add_chat_service("chat_completion", AzureChatCompletion(**aoai_settings_dict))
    prompt_config = sk.PromptTemplateConfig()
    prompt_template = sk.ChatPromptTemplate(
        "{{$user_input}}", kernel.prompt_template_engine, prompt_config
    )

    function_config = sk.SemanticFunctionConfig(prompt_config, prompt_template)
    chat_function = kernel.register_semantic_function(f"testfn", f"testfn", function_config)
    while True:
        text = input("Enter text: ")
        context = kernel.create_new_context()
        context['user_input'] = text
        result = chat_function.invoke(context=context)
        print(result.result)

moonbox3:
Hi @sarah-widder, we've decided to remove support related to running functions synchronously. The PR for that work is [here](https://github.com/microsoft/semantic-kernel/pull/4796). Closing this issue.

PR #4780, opened 2024-01-29
Python: address sync invoke hanging issue when running a kernel function

moonbox3:
### Motivation and Context

Addresses issue: #4762 

SK Python is mainly async, but we do have a sync invoke on the kernel function. The underlying code is managed by asyncio, and there appears to be an issue managing the event loop when making multiple calls to a kernel function's invoke.

### Description

The PR is attempting to better handling the sync function invoke by starting a new event loop if one doesn't exist. If an event loop does exist (by means of running in a Jupyter notebook which does have its own event loop), then a task is returned. The caller should await this to get the result. Only a few notebooks use the sync invoke function method, and those instances were updated and tested to be working.

A new integration test and a kernel syntax example were added to test this sync invoke functionality, to make sure it's working properly. 

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/orchestration/kernel_function.py
+++ b/python/semantic_kernel/orchestration/kernel_function.py
@@ -345,6 +345,28 @@ class KernelFunction(KernelFunctionBase):
         settings: Optional[AIRequestSettings] = None,
         log: Optional[Any] = None,
     ) -> "KernelContext":
+        """
+        Invoke the KernelFunction synchronously. As the underlying functions are async,
+        this method is managed by asyncio and will block the current thread until the
+        function completes.
+
+        If running from a Jupyter notebook, which already has an event loop running,
+        the caller will need to await this synchronous call to get the function result.
+
+        Arguments:
+            input {Optional[str]} -- Input text to the function
+            variables {ContextVariables} -- Context variables to pass to the function
+            context {Optional[KernelContext]} -- KernelContext to pass to the function
+            memory {Optional[SemanticTextMemoryBase]} -- Memory to pass to the function
+            settings {Optional[AIRequestSettings]} -- AIRequestSettings to pass to the function
+            log {Optional[Any]} -- Deprecated
+
+        Returns:
+            KernelContext -- The result of the function
+
+        Raises:
+            KernelException: If an error occurs during the function invocation
+        """
         from semantic_kernel.orchestration.kernel_context import KernelContext
 
         if log:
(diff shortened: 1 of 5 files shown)


juliomenendez (approved):
Shouldn't the `await ...invoke` in the notebooks be `await ...invoke_async` instead?

moonbox3:
> Shouldn't the `await ...invoke` in the notebooks be `await ...invoke_async` instead?

@juliomenendez updates notebooks to use `invoke_async`.

python/semantic_kernel/orchestration/kernel_function.py
```diff
@@ -367,24 +389,49 @@ def invoke(
...
+                self.invoke_async(
+                    input=input,
+                    variables=variables,
+                    context=context,
+                    memory=memory,
+                    settings=settings,
+                    log=log,
+                )
+            )
+            return task
```
eavanvalkenburg, line 414: this is not a finished task, shouldn't it just run asyncio.run() and then the actual result is returned?

eavanvalkenburg (changes requested):
the new implementation seems to combine asyncio concepts at different abstraction levels, might be needed, but let's be clear on why and how.

python/semantic_kernel/orchestration/kernel_function.py
```diff
@@ -417,7 +464,7 @@ async def invoke_async(
             if self.is_semantic:
                 return await self._invoke_semantic(context, settings, **kwargs)
             else:
-                return await self._invoke_native(context, **kwargs)
```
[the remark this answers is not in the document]
moonbox3: The `_invoke_native` method in `kernel_function` does not have a `**kwargs` param.

python/semantic_kernel/orchestration/kernel_function.py
moonbox3: Simply calling `asyncio.run()` is the reason the execution hangs. A Jupyter notebook already has a running event loop, if you try to fire off an `asyncio.run()`, it will create another event loop, and we get into trouble. I've already converted the Jupyter notebook code to use `invoke_async` where it was previously using `invoke`, aside from the methods that are hitting the overriden `__call__` method in the `kernel_function`. I think we should move away from supporting sync invocations, but we can discuss offline.

moonbox3:
We're removing sync invoke, thus closing the PR.

Issue #4775, opened 2024-01-29
.Net: Option to control temperature to stop Handlebars planner hallucinating functions.


rayjk21:
I'm new to SK in the last few weeks and have only used the later versions (e.g. 1.2.0 and 1.0.0.-beta8).  I'm trying the Handlebars planner and finding that it is hallucinating some basic functions (e.g. "Example-Random") and in this example "length":  
```{{!-- Step 1: Identify Key Values --}}
{{set "tables" (myCode-GetTablesInDatabase)}}

{{!-- Step 2: Choose the Right Helpers --}}
{{set "chosenTable" (range 0 (length tables) 1)}}
{{set "variables" (myCode-GetVariablesFromTable (lookup tables chosenTable))}}
```

It would be useful to have a way of turning down the temperature with the HandleBars planner.

This is the exception that I got:
```Exception has occurred: CLR/Microsoft.SemanticKernel.KernelException
An unhandled exception of type 'Microsoft.SemanticKernel.KernelException' occurred in System.Private.CoreLib.dll: '[HallucinatedHelpers] The plan references hallucinated helpers: Helper 'length''
 Inner exceptions found, see $exception in variables window for more details.
 Innermost exception      HandlebarsDotNet.HandlebarsRuntimeException : Template references a helper that cannot be resolved. Helper 'length'
   at HandlebarsDotNet.Helpers.MissingHelperDescriptor.Invoke(HelperOptions& options, Context& context, Arguments& arguments)
   at HandlebarsDotNet.Helpers.MissingHelperDescriptor.HandlebarsDotNet.Helpers.IHelperDescriptor<HandlebarsDotNet.HelperOptions>.Invoke(HelperOptions& options, Context& context, Arguments& arguments)```

joslat:
Yep, we have been mentioning this hallucination and handlebar planners imagining the syntax, so they cannot be trusted but as an intermediate step to generate a plan just to review, polish and fix later... but not to use it right away.

Issue #4731, opened 2024-01-24
.Net: Expose prompt execution settings in HandlebarsPlannerOptions


gitri-ms:
Expose `PromptExecutionSettings` in `HandlebarsPlannerOptions` so the caller can override model settings.

Issue #4729, opened 2024-01-24
.Net: Update usage of PlannerConfigBase


gitri-ms:
`PlannerConfigBase` was intended to house the configuration options shared between multiple planners. However, it is currently extended by `FunctionCallingStepwisePlannerConfig` but not `HandlebarsPlannerOptions`. The latter class has several of the same properties as the base class, creating redundancy.

We should consider either:
- Having both planner options classes extend the base
- If there is a good reason not to do that, then remove the base class

Issue #4728, opened 2024-01-24
.Net: Implement or remove unused planner options


gitri-ms:
Some options, such as `GetFunctionCallback`, are not currently used by either planner. These should either be implemented or removed.

PR #4793, opened 2024-01-30
.Net: Planner options improvements

gitri-ms:
### Motivation and Context

Changes to improve consistency across the .NET planners. Addresses the following:
- #4727 
- #4728 
- #4729 
- #4731 
- #4775 

### Description

- Clean up planner options. Shared options live in parent class (`PlannerOptions`). Other options reside in the planner-specific options class (`HandlebarsPlannerOptions` or `FunctionCallingStepwisePlannerOptions`).
- Change default name delimiter to dash (was underscore for function calling stepwise planner). Update all tests accordingly. (Will explore making this configurable in a separate issue/PR)
- Change how we get the list of available functions for Handlebars planner, to reduce code duplication and get overridable options for free.
- Make `MaxTokens` setting optional for function calling stepwise planner (not yet supported for Handlebars planner)

### Contribution Checklist


glahaye (approved):
LGTM

teresaqhoang (approved):

dotnet/src/InternalUtilities/planning/PlannerOptions.cs
```diff
@@ -53,4 +36,10 @@ public abstract class PlannerConfigBase
     /// This configuration will be ignored if GetAvailableFunctionsAsync is set.
     /// </remarks>
     public SemanticMemoryConfig SemanticMemoryConfig { get; set; } = new();
+
+    /// <summary>
+    /// Name delimiter used to separate plugin name from function name. This should be a character or string
+    /// that is not found in any plugin or function names used by the planner.
+    /// </summary>
+    public virtual string NameDelimiter { get; set; } = "-";
```
gitri-ms, line 44: This needs to get propagated through to `OpenAIFunction.NameDelimiter`

gitri-ms: Looking more at this, it adds a bit of complexity because name delimiter would be configurable both from here and OpenAIFunction. I think more thought neesd to go into how to keep these values consistent and where the right place to configure it should be. Going to remove the option to override this setting from this PR for now.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunction.cs
```diff
@@ -101,7 +101,7 @@ public sealed class OpenAIFunction
     }
 
     /// <summary>Gets the separator used between the plugin name and the function name, if a plugin name is present.</summary>
-    public static string NameSeparator => "_";
+    public static string NameSeparator { get; set; } = "-";
```
markwallace-microsoft, line 104: Can you provide some background on this change?

I think we would be useful to add a `<remarks>` section to the API documentation explaining why we are using `-` in preference to `_`.

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
```diff
@@ -66,7 +67,18 @@ public Task<HandlebarsPlan> CreatePlanAsync(Kernel kernel, string goal, Cancella
...
+
+    private static readonly PromptExecutionSettings s_defaultPromptExecutionSettings = new()
+    {
+        ExtensionData = new Dictionary<string, object>()
+        {
+            { "temperature", 0.0 },
+            { "top_p", 0.0 },
+            { "presence_penalty", 0.0 },
+            { "frequency_penalty", 0.0 },
+        }
```
markwallace-microsoft, line 80: Can we use `OpenAIPromptExecutionSettings` given this is OpenAI specific? Or if we want to be model agnostic move this elsewhere?

markwallace-microsoft (approved):

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
gitri-ms: Good question. We don't want the Handlebars planner to take a dependency on OpenAI, but when users *do* use OpenAI, we want to use these defaults. Is there a better way to achieve that without taking a dependency on an OpenAI-specific package?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunction.cs
gitri-ms: There had been some discussions in December about aligning the two planners to use the same separator, and agreement that dash would be the better option (we've seen some issues from people trying to use underscores/snake case in their function names). However, the stepwise planner was never updated after those discussions. I agree, adding to the API documentation would be a good idea - will update.

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
gitri-ms: I think the best course of action might be to remove this, and just recommend that developers override the model settings when using the planner (especially to turn down the temperature when using OpenAI models). Ultimately it is up to them to tune the settings to what they want, although generally it seems lower temp yields a more reliable planner. I'll make a note of this in the examples.

teresaqhoang (approved):

markwallace-microsoft (approved):

glahaye (approved):

Issue #4794, opened 2024-01-31
Python: Remove sync function invoke


moonbox3:
Similar to dotnet, remove the ability to run functions synchronously.

PR #4796, opened 2024-01-31
Python: Remove sync function invoke. Update other code to use async invoke.

moonbox3:
### Motivation and Context

Completes #4794. 

Most of SK Python is written with async code; however, there were some code paths that masked the underlying async code and made it look like a sync function invoke. In reality, asyncio spun up event loops and attempted to manage them properly. This becomes tricky when running a Jupyter notebook as they have their own event loop.

### Description

This PR removes the ability to call a kernel function synchronously. Since the majority of the code base is written with async methods, and to match dotnet, which is async, we're removing the sync function invoke and its dependent code. Some Planners were also run synchronously -- these were updated to run asynchronously. Kernel examples, notebooks, docs, and tests were updated accordingly. 

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/orchestration/kernel_function.py
+++ b/python/semantic_kernel/orchestration/kernel_function.py
@@ -325,7 +325,7 @@ class KernelFunction(KernelFunctionBase):
             variables {ContextVariables} -- The variables for the function
             context {Optional[KernelContext]} -- The context for the function
             memory {Optional[SemanticTextMemoryBase]} -- The memory for the function
-            settings {Optional[AIRequestSettings]} -- The settings for the function
+            settings {Optional[PromptExecutionSettings]} -- The settings for the function
             log {Optional[Any]} -- A logger to use for logging. (Optional)
 
         Returns:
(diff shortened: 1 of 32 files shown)


juliomenendez (approved):
LGTM

juliomenendez (approved):

python/semantic_kernel/orchestration/kernel_function.py
```diff
@@ -307,7 +307,7 @@ def describe(self) -> FunctionView:
             parameters=self._parameters,
         )
 
-    def __call__(
+    async def __call__(
```
eavanvalkenburg, line 310: I'm not sure if __call__ is allowed to be overwritten as async func...

eavanvalkenburg: alternative would be to just delete it (there is also no equivalent in dotnet)

eavanvalkenburg (changes requested):
just a question on whether async __call__ is allowed, otherwise LGTM

python/semantic_kernel/orchestration/kernel_function.py
moonbox3: It works well to override as an async call. One can then `await some_func()`. The code has been updated from the previous sync override to async and notebooks, examples, and tests pass.

alexchaomander (approved):
LGTM!

eavanvalkenburg (approved):
Lgtm

Issue #4558, opened 2024-01-10
.Net - Support Open AI file upload


crickman:
https://platform.openai.com/docs/api-reference/files

PR #4803, opened 2024-01-31
.Net - OpenAI File Service ADR

crickman:
### Motivation and Context

ADR in alignment with https://github.com/microsoft/semantic-kernel/issues/4558

Adding support for **OpenAI File Service** to be used in conjunction with Agent/Assistant API: https://platform.openai.com/docs/api-reference/files

### Description

OpenAI provides a file service for uploading files to be used for *assistant retrieval* or *model fine-tuning*: `https://api.openai.com/v1/files`

> Note: *Azure Open AI* does not currently support the OpenAI file service API.

### Contribution Checklist


docs/decisions/0026-file-service.md
```diff
@@ -0,0 +1,148 @@
...
+}
+
+public interface IFileUploadRequest
+{
+    string FileName { get; }
+ 
+    Stream GetContent();
+}
+
+public class FileReference
```
matthewbolanos, line 110: Should be make "files" and/or "documents" a first class content type by having it extend `KernelContent`?

docs/decisions/0026-file-service.md
```diff
@@ -0,0 +1,148 @@
...
+
+## More Information
+
+Signatures for *Option 2*, but can adapt to any option.
+
+#### `Microsoft.SemanticKernel.Abstractions`
+```csharp
+namespace Microsoft.SemanticKernel.Files;
+
+public interface IFileService : IAIService
```
matthewbolanos, line 65: Should this be an abstract class to protect against breaking changes in the future if/when we need to add additional functionality.

docs/decisions/0026-file-service.md
```diff
@@ -0,0 +1,148 @@
...
+### Option 1. Add OpenAI file service support to `Microsoft.SemanticKernel.Experimental.Agents`
+**Pro:**
+1. No impact to existing AI connectors.
+
+**Con:**
+1. No reuse via AI connectors.
+1. No common abstraction.
+1. Unnatural dependency binding for uses other than with OpenAI assistants.
+
+### Option 2. Add a file service abstraction and implement support for OpenAI
```
matthewbolanos, line 40: Are there other file management abstractions/services that we could use within .NET that we could leverage? Or do we *have* to create our own abstraction?

docs/decisions/0026-file-service.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Decision Outcome
+
+> Option 2. **Add a file service abstraction and implement support for OpenAI**
+
+Defining a generalized file service interface provides an extensibility point for other vendors, in addition to *OpenAI*.
+
+## Pros and Cons of the Options
+
+### Option 1. Add OpenAI file service support to `Microsoft.SemanticKernel.Experimental.Agents`
```
matthewbolanos, line 31: Is there a middle ground between Option 1 and Option 2 where we provide an abstraction for docs/files (e.g., `DocumentContent`) and leave it up to the AI connector to handle how the file is "managed" (i.e., not introduce `IFileService`)?

docs/decisions/0026-file-service.md
crickman: I don't think `IFileProvider` matches our use: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.ifileprovider?view=dotnet-plat-ext-8.0

The system `Path`, 'Directory' and `File` utilities are all static: https://learn.microsoft.com/en-us/dotnet/api/system.io?view=net-8.0

Definately open to insights others may have here...and I'll continue to search.

docs/decisions/0026-file-service.md
crickman: Doesn't a base class and interface respond the same for equivalent changes w.r.t. backward compat?

docs/decisions/0026-file-service.md
crickman: Thinking hard...ultimately, something needs to call the rest api...I mean, I guess we could mangle `OpenAIChatCompletionService` without changing `IChatCompletionService` or `ITextGenerationService`

docs/decisions/0026-file-service.md
crickman: Do you mean instead of `FileReference` or in relation to the return type of `GetContent()`?

docs/decisions/0026-file-service.md
crickman: Updated

docs/decisions/0026-file-service.md
crickman: I think that's Option #3.  I've added signatures for that.

crickman: Talked to Matthew offline - doing option #3

docs/decisions/0026-file-service.md
crickman: Following up with @dmytrostruk on overlap w/ `BinaryContent`

docs/decisions/0026-file-service.md
```diff
@@ -0,0 +1,144 @@
...
+# These are optional elements. Feel free to remove any of them.
+status: proposed
+contact: crickman, mabolan, semenshi
+date: 2024-01-16
+---
+
+# File Services
+
+## Context and Problem Statement
+OpenAI provides a file service for uploading files to be used for *assistant retrieval* or *model fine-tuning*: `https://api.openai.com/v1/files`
```
glahaye, line 11: Probably better to link to the doc on Files rather than the API endpoint itself:

https://platform.openai.com/docs/api-reference/files

crickman: I avoided since the API doc URL's have exhibited a pattern of shifting around over time and create link-check problems.

The provided endpoint _is_ the service, after all.

glahaye: OK. I'll leave it to your discretion. But just bear in my that putting the API URL is not very useful because when anyone clicks on it, they'll just get a weird "page" not actually meant for direct human consumption giving them an error.

glahaye (approved):

SergeyMenshykh (approved):

Issue #4732, opened 2024-01-24
Include output scheme in functions manual for HandlebarsPlanner


gitri-ms:
The function descriptions in the handlebars planner prompt should include the scheme of the output. This would make it possible for the planner to generate a plan that can parse or extract data from an output object.

Issue #4661, opened 2024-01-17
Get Method missing in Handlebars planner to parse JSON output


rithvikb23:
When using the Handlebars planner in semantic kernel, I noticed an issue when the planner was parsing the actual JSON output. For example: I have an API that calculates the price target for a given stock ticker. This API is documented in an OpenAPI spec. After loading the spec as a plugin using the ImportPluginFromOpenApiAsync method and asking the query "what is the price target of GOOGL" the planner produces this plan:

**Original plan:
{{!-- Step 1: Set the ticker symbol for which the price target is to be retrieved --}}
{{set "ticker" "GOOGL"}}

{{!-- Step 2: Use the custom helper to retrieve the price target for the specified ticker symbol --}}
{{set "priceTarget" (PriceTargetPlugin-getPriceTarget Ticker=ticker)}}

{{!-- Step 3: Output the result --}}
{{json (concat "The price target of " ticker " is: $" priceTarget)}}

Result:
The price target of GOOGL is: ${"value":180}**

The problem in the result is that the json helper is not getting the actual value, instead it is returning the actual json object. 

Ideally it should look like: The price target of GOOGL is 180. I am using Semantic Kernel 1.0.1 in .NET 6.

Is there a setting allow getting the actual value, I have attempted to tweak the response section of my OpenAPI spec but it is still not retrieving the exact value.

joslat:
Isn't this the same as this one: https://github.com/microsoft/semantic-kernel/issues/4596 ".Net: get helper needs to be added back to the handlebars planner #4596"?
(found this after being mentioned during yesterday's Office Hours)

gitri-ms:
This appears to be a different issue than #4596. I think the real issue is that we are not including the response schema for custom helpers in the handlebars prompt template, so the planner does not know how to parse data out of the response object. I am working on a fix for this.

PR #4804, opened 2024-01-31
Fixing logic in custom `or` HB helper

teresaqhoang:
### Motivation and Context

The custom `or` helper had a bug where it was checking for null values, rather than checking for non-null values. This PR changes the condition to `not null`.

Also, renamed a heading in the CreatePlan prompt to better guide the model in identifying Handlebars built-in _block_ helpers.

### Description

Additional changes to the `GetArgumentValue` method in `KernelHelperUtils.cs` to return null if variable isn't defined.

Before, if the template referenced a variable that wasn't defined in `KernelArguments`, the method would just return the key as a string. Now, it just returns null (as expected).
> I had it returning a string before because I thought Handlebars would default the argument to its string representation since it templates regular text, but I double checked [HB docs](https://handlebarsjs.com/guide/expressions.html#helpers-with-multiple-parameters), and generally, string literals are surrounded by quotes when a raw value is passed as an argument.

Should resolve these issues:
- https://github.com/microsoft/semantic-kernel/issues/4661
- https://github.com/microsoft/semantic-kernel/issues/4732

### Contribution Checklist


dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelSystemHelpers.cs
```diff
@@ -136,7 +136,7 @@ internal static class KernelSystemHelpers
                 return arg switch
                 {
                     bool booleanArg => booleanArg,
-                    _ => arg is null
+                    _ => arg is not null
```
gitri-ms, line 139: I think this change is correct but there is some issue in the tests for this helper that is causing two of the cases to fail. I am investigating those.

teresaqhoang: Aligned offline - solution was a change to the `GetArgumentValue` method in `KernelHelperUtils.cs`.
 
Before, if the template referenced a variable that wasn't defined in KernelArguments, the method would just return the  key as a string. Now, it just returns null (as expected). 
 
I had it returning a string before because I thought Handlebars would default the argument to its string representation since it templates regular text, but I double checked [HB docs](https://handlebarsjs.com/guide/expressions.html#helpers), and generally, string literals are surrounded by quotes when the raw value is used as an argument

dmytrostruk: Yes, I remember that I changed this code, because I was adding new tests to increase our code coverage to 80%, and there were no tests for `or` helper. When I added unit tests with expected outcome, it appeared that this helper was not working as expected, so I slightly changed it. 

The correct logic should be - if any of values in `or` operator is not null - then it should return true. So, this change is correct :)

dmytrostruk (approved):

dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelSystemHelpers.cs
teresaqhoang: Awesome, thanks for the detail Dmytro!

gitri-ms (approved):

Issue #4702, opened 2024-01-22
.Net: Pinecone connector PineconeClient will not deserialize pod_type due to converter priority


royaltie-roman:
**Describe the bug**

In PineconeClient there is a line that always breaks

```csharp
public async Task<PineconeIndex?> DescribeIndexAsync(string indexName, CancellationToken cancellationToken = default)
{
      // .... 
      PineconeIndex? indexDescription = JsonSerializer.Deserialize<PineconeIndex>(responseContent, this._jsonSerializerOptions);
      // .... 
}
```
with the following error:

`The JSON value could not be converted to Microsoft.SemanticKernel.Connectors.Memory.Pinecone.Model.PodType. Path: $.database.pod_type | LineNumber: 0 | BytePositionInLine: 144.`

My research shows that this is because in PineconeUtils.cs there is line
```csharp
    internal static JsonSerializerOptions DefaultSerializerOptions { get; } = new()
    {
        // ....
        Converters = { new JsonStringEnumConverter(JsonNamingPolicy.CamelCase) },
    };
```

So, custom converter `PodTypeJsonConverter` for `PodType` enum will be added **AFTER** the enum converter and, as result, will never be called because default converter will error out.

**To Reproduce**

Happens on every call to get Pinecone index info. 

**Expected behavior**

Custom converter for `PodType` should be called first.

**Workaround**

To whoever has the same issue, here is **ABSOLUTELY UGLY** workaround

```csharp
    // Call this somewhere from static constructor
    public static void FixPineconePodTypeConverter()
    {
        var pineconeUtilsType = typeof(PineconeUtils);
        var optionsProperty = pineconeUtilsType.GetProperty("DefaultSerializerOptions", BindingFlags.NonPublic | BindingFlags.Static)!;
        var jsonOptions = (JsonSerializerOptions)optionsProperty.GetValue(null)!;
        var converterType = pineconeUtilsType.Assembly.GetType("Microsoft.SemanticKernel.Connectors.Memory.Pinecone.Model.PodTypeJsonConverter", true)!;
        var converterInstance = (JsonConverter<PodType>)Activator.CreateInstance(converterType, true)!;
        if (jsonOptions.Converters.All(c => c.GetType() != converterType))
        {
            jsonOptions.Converters.Insert(0, converterInstance!);
        }
    }
```

royaltie-roman:
[...]

PR #4852, opened 2024-02-02
.Net: Adding EnumType PodType.Nano to Pinecone Connector

Imir:
### Motivation and Context

I came across the same [issue](https://github.com/microsoft/semantic-kernel/issues/4702#issuecomment-1905017366) mentioned in issue #4702 on GitHub. The problem is related to Pinecone connector’s inability to deserialize the “nano” pod type, which is the free tier of Pinecone. Previously the pod type for free tier was “starter”. 

It seems like a minor issue, but it’s currently blocking me from proceeding. I would appreciate it if this bug could be addressed soon, rather than being left on the backlog for an extended period of time.

### Contribution Checklist


Imir:

dotnet/src/Connectors/Connectors.Memory.Pinecone/Model/PodType.cs
```diff
@@ -96,7 +96,13 @@ public enum PodType
     /// Enum Starter for value: starter
     /// </summary>
     [EnumMember(Value = "starter")]
-    Starter = 13
+    Starter = 13,
```
dmytrostruk, line 99: @Imir From PR description:
```
The problem is related to Pinecone connector’s inability to deserialize the “nano” pod type, which is the free tier of Pinecone. Previously the pod type for free tier was “starter”.
```

Does it mean that `starter` pod type doesn't exist anymore and instead of adding new enum value `Nano`, we should rename `Starter` to `Nano`?

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.Memory.Pinecone/Model/PodType.cs
Imir: It’s a valid observation. A few weeks ago, my current `starter` pod began returning `nano` as its pod type. This change is also reflected in Pinecone’s dashboard. After your comment, I created new “free” indexes using both `starter` and `nano` pod types for testing purposes. Both variations function as expected, although the dashboard consistently displays the pod type as `starter`. It looks like the `starter` pod is still valid, and I believe it’s necessary to retain both.

dmytrostruk: That makes sense, thank you for explanation!

markwallace-microsoft (approved):

PR #4860, opened 2024-02-02
.Net: Allow access KernelFunction.ExecutionSettings from IAIServiceSelector

### LINKED ISSUE #4487: .Net: Can't access KernelFunction.ExecutionSettings from IAIServiceSelector implementation

markwallace-microsoft:
### Motivation and Context

Closes #4487 

### Contribution Checklist


markwallace-microsoft:
@stephentoub Here's a first draft of the changes to allow is to freeze `PromptExecutionSettings`.

I could make Freeze return a copy of the settings which are frozen. What do you think?

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -39,5 +52,43 @@ public class PromptExecutionSettings
...
+    /// </summary>
+    public bool IsFrozen
+    {
+        get => this._isFrozen;
+    }
+
+    /// <summary>
+    /// Makes the current <see cref="PromptExecutionSettings"/> unmodifiable and sets its IsFrozen property to true.
+    /// </summary>
+    public PromptExecutionSettings Freeze()
```
stephentoub, line 80: I think this should be virtual so that derived implementations can react to it.

For example, this base implementation should probably do:
```C#
if (_extensionData is not null)
{
    _extensionData = new ReadOnlyDictionary<string, object?>(_extensionData);
}
```
or something like that in the Freeze call.

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -39,5 +52,43 @@ public class PromptExecutionSettings
...
+    {
+        this._isFrozen = true;
+
+        return this;
```
stephentoub, line 84: Is returning `this` necessary? Having this method return a `PromptExecutionSettings` makes it seem to me like it's returning a new frozen copy rather than mutating itself.

That would be another approach that could be taken if desired, e.g. a Clone method and a Freeze method, or alternatively a method that creates a read-only clone, rather than mutating this.

markwallace-microsoft: I've added a `Clone()` method and kept `Freeze()` but happy to combine these.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -22,66 +24,174 @@ public sealed class OpenAIPromptExecutionSettings : PromptExecutionSettings
...
-    public double TopP { get; set; } = 1;
+    public double TopP
+    {
+        get => this._topP;
+
+        set
+        {
+            if (this.IsFrozen)
+            {
+                throw new InvalidOperationException("OpenAIPromptExecutionSettings are frozen and TopP cannot be modified.");
```
stephentoub, line 55: Nit: I don't think you need a different message per property. I suggest just defining a shared private ThrowIfFrozen helper.

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -30,7 +32,19 @@ public class PromptExecutionSettings
...
+    public string? ModelId
+    {
+        get => this._modelId;
+
+        set
+        {
+            if (this._isFrozen)
+            {
+                throw new InvalidOperationException("PromptExecutionSettings are frozen and ModelId cannot be modified.");
+            }
```
stephentoub, line 44: Same ThrowIfFrozen comment here

stephentoub (commented):
This approach now looks pretty reasonable I think

SergeyMenshykh (approved):

dmytrostruk (commented):
A couple of concerns: 

1. The approach with `Freeze` method will generate runtime error when trying to update class property, which makes sense, if we want to make `PromptExecutionSettings` as read-only at some point. Although, I'm wondering if it's possible to make read-only variant of `PromptExecutionSettings` (e.g. `ReadOnlyPromptExecutionSettings`) which will contain setters only, will be initialized once and then it won't be possible to update properties, since you will receive compile-time error. 

2. `Freeze` method doesn't guarantee immutability of `{Connector}PromptExecutionSettings` classes. In our case, yes, `OpenAIPromptExecutionSettings` will be immutable, but for each new connector developers will have to know that new child type of `PromptExecutionSettings` should be immutable as well. In fact, in current implementation of Gemini model, `GeminiPromptExecutionSettings` class does not contain `freeze` logic: [link](https://github.com/microsoft/semantic-kernel/blob/feature-connectors-gemini/dotnet/src/Connectors/Connectors.GoogleVertexAI/Settings/GeminiPromptExecutionSettings.cs). Taking into account, that it's possible to implement custom connector with custom settings, the read-only requirement can be easily avoided. Just want to double-check if this scenario is not a problem for us.

markwallace-microsoft:
@dmytrostruk I've tried the `ReadOnlyPromptExecutionSettings` and I've started to see issues. The `IAIServiceSelector` expects to provide a `PromptExecutionSettings` as output, which means I would need to store a read only copy in the function and create another copy when the `IAIServiceSelector` is called or check that interface also. I think the current implementation is the best option for now.

```
internal sealed class OrderedAIServiceSelector : IAIServiceSelector
{
    public static OrderedAIServiceSelector Instance { get; } = new();

    /// <inheritdoc/>
    public bool TrySelectAIService<T>(
        Kernel kernel, KernelFunction function, KernelArguments arguments,
        [NotNullWhen(true)] out T? service,
        out PromptExecutionSettings? serviceSettings) where T : class, IAIService
```

dmytrostruk (approved):

SergeyMenshykh (approved):

PR #4863, opened 2024-02-02
Python: Pydantic-ify the KernelFunction class. Fix code paths related to tool_calls.

moonbox3:
### Motivation and Context

Solves #4864

The KernelFunction was configured as a Python class without regards to using Pydantic fields and its validations. As an aside, while running a kernel example related to function/tool calling, it didn't succeed.

### Description

This PR cleans up the KernelFunction by creating Pydantic fields. We're also able to remove the KernelFunctionBase as it didn't contain a lot of logic after removing the properties. 

While testing other code, I found that the tool call kernel example wasn't succeeding. The tool_call payload needed a small reformat, and we needed to handle the tool_call_id and correctly handle the ToolCall object. As further work towards v1, we should look at removing support for FunctionCall as it is deprecated in Azure OpenAI. 

We now see:

```
User:> What is 3+3?
Mosscap:> Ah, the timeless question of arithmetic! The sum of 3 and 3 is undoubtedly a splendid number. 
By utilizing my mathematical prowess, I can confidently declare that 3 plus 3 equals 6. 
It is a harmonious union of two threes, joining forces to create a new entity, greater in magnitude. 
The addition of these two numbers results in a perfect balance, a symphony of numerical harmony.
```

### Contribution Checklist


python/samples/kernel-syntax-examples/chat_gpt_api_function_calling.py
```diff
@@ -61,10 +62,11 @@
         max_tokens=2000,
         temperature=0.7,
         top_p=0.8,
-        function_call="auto",
-        functions=get_function_calling_object(kernel, {"exclude_plugin": ["ChatBot"]}),
+        tool_choice="auto",
+        tools=get_function_calling_object(kernel, {"exclude_plugin": ["ChatBot"]}),
```
eavanvalkenburg, line 66: Should we then also change this function name?

python/semantic_kernel/connectors/ai/open_ai/models/chat/tool_calls.py
```diff
@@ -21,3 +23,26 @@ def __add__(self, other: Optional["ToolCall"]) -> "ToolCall":
             type=self.type or other.type,
             function=self.function + other.function if self.function else other.function,
         )
+
+    def parse_arguments(self) -> Dict[str, str]:
```
eavanvalkenburg, line 27: The function can already has this function, why not let that handle it? Function call can be kept for the inner function and translating it into a kernel_function rather then a openai function

python/semantic_kernel/connectors/ai/open_ai/semantic_functions/open_ai_chat_prompt_template.py
```diff
@@ -13,9 +13,11 @@
 
 
 class OpenAIChatPromptTemplate(ChatPromptTemplate[OpenAIChatMessage]):
-    def add_function_response_message(self, name: str, content: Any) -> None:
+    def add_function_response_message(self, name: str, content: Any, tool_call_id: Optional[str]) -> None:
```
eavanvalkenburg, line 16: Let's give toolcallid a default None for backward compatibility

eavanvalkenburg (changes requested):
Some comments on the tool call/ function call changes, let's discuss. The rest looks good

python/semantic_kernel/connectors/ai/open_ai/models/chat/tool_calls.py
moonbox3: I see. I was thinking that `FunctionCall` would go away, but I guess we'll keep it to handle the inner portion of a tool_call.

python/semantic_kernel/connectors/ai/open_ai/semantic_functions/open_ai_chat_prompt_template.py
moonbox3: I think we need to eventually move to having the `OpenAIChatMessage` support a List[ToolCall] because the model may return multiple tool calls. Right now we're only supporting the first, but if we get more than one, and don't send back all tool_call_ids, it'll fail.

python/semantic_kernel/connectors/ai/open_ai/utils.py
```diff
@@ -57,7 +118,7 @@ def get_function_calling_object(kernel: Kernel, filter: Dict[str, List[str]]) ->
                     "exclude_function": ["plugin1-function1", "plugin2-function2"],
                     }
                 will return all functions except plugin1-function1 and plugin2-function2.
-        caller_function_name: the name of the function that is calling the other functions.
+        form_tool_call: if True, the function will return a list of tool calls, otherwise a list of functions.
```
eavanvalkenburg, line 121: from_tool_call?

python/semantic_kernel/connectors/ai/open_ai/utils.py
```diff
@@ -103,14 +162,16 @@ async def execute_function_call(kernel: Kernel, function_call: FunctionCall, log
     return str(result)
 
 
+async def execute_tool_call(kernel: Kernel, tool_call: ToolCall) -> str:
+    return await execute_function_call(kernel, tool_call.function)
```
eavanvalkenburg, line 166: should we inverse this so that when we get rid of function_call this does not have to be changed?

python/semantic_kernel/planning/plan.py
```diff
@@ -26,10 +26,10 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 
-class Plan(KernelFunctionBase):
+class Plan:
```
eavanvalkenburg, line 29: will this not break type checking? when a plan get's invoked in the place of a function? so this can then inherit from KernelFunction?

eavanvalkenburg (changes requested):
couple of open questions and one seeming typo!

python/semantic_kernel/connectors/ai/open_ai/utils.py
moonbox3: "form the tool call" -> "form_tool_call" not "from"

python/semantic_kernel/connectors/ai/open_ai/utils.py
moonbox3: Makes sense, I think that's a good idea.

python/semantic_kernel/connectors/ai/open_ai/utils.py
moonbox3: I've updated this to `is_tool_call` to remove any confusion.

python/semantic_kernel/connectors/ai/open_ai/utils.py
moonbox3: I've made a change that decouples the execution logic from the specific types of calls, making it easier for us to remove `function_call` in the future.

python/semantic_kernel/planning/plan.py
moonbox3: We spoke offline about this. Type checking isn't breaking related to removing this inheritance.

eavanvalkenburg (approved):

juliomenendez (approved):
LGTM

PR #4868, opened 2024-02-02
.Net - Open AI File Service Support

crickman:
### Motivation and Context

Add support for Open AI File Service: 
- https://platform.openai.com/docs/api-reference/files
- https://api.openai.com/v1/files

### Description

ADR: https://github.com/microsoft/semantic-kernel/pull/4803

### Contribution Checklist


crickman:
Needs coverage

dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" />
+    <PackageReference Include="System.ClientModel" />
```
dmytrostruk, line 26: If the purpose of adding `System.ClientModel` package is to use `BinaryData` type only (I can't find other types used, please correct me if I'm wrong), then you should probably use `System.Memory.Data` which contains `BinaryData` type:
https://www.nuget.org/packages/System.Memory.Data
image

The reason why `BinaryData` is accessible with `System.ClientModel` is because `System.ClientModel` references `System.Memory.Data`.

But the purpose of `System.ClientModel` is different: `System.ClientModel contains building blocks for communicating with cloud services. It provides shared primitives, abstractions, and helpers for .NET service client libraries.`

crickman: Azure library also references System.ClientModel...so we already are.  I thought it might be a more more efficient maintenance point to keep in sync with our existing dependency stack (avoid diverging)

SergeyMenshykh (approved):

alliscode (approved):

PR #4883, opened 2024-02-05
.Net - Fix Agent Message Ordering for Multimessage Runs

crickman:
### Motivation and Context

Observed agent messages not being provided in the expected order.  In practice they appear reversed, but could be anything depending on JSON data.

### Description

Changed run processing to return message-ids as `IAsyncEnumerable`, in the order they are generated.

In addition to addressing ordering issue, this also introduces the ability for the first message to be displayed as soon as it is produced (for multi-message runs)...even if the last message is added 30 seconds later.  (In the current version, message-ids are return only after all have been produced.)

Also explored both flavors of `Example75_AgentTools` as separate entry-points.

### Contribution Checklist

### DIFF
--- a/dotnet/samples/KernelSyntaxExamples/Example75_AgentTools.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example75_AgentTools.cs
@@ -39,7 +39,6 @@ public sealed class Example75_AgentTools : BaseTest
             return;
         }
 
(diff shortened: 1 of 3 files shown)


glahaye (approved):
LGTM

alliscode (approved):

PR #4908, opened 2024-02-06
Java: Add fluent invocations

johnoliver:
### Contribution Checklist


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/FunctionInvocation.java
```diff
@@ -0,0 +1,191 @@
...
+            LOGGER.debug(
+                "No variable type explicitly specified by calling 'withResultType' for function invocation: "
+                    + function.getSkillName() + "." + function.getName() + "."
+                    + " This may cause a runtime error (probably a ClassCastException) if the result type is not compatible with the expected type.");
+        }
+
+        function.invokeAsync(
+                kernel,
+                arguments,
+                null,
```
dsgrieve, line 161: Why are you not passing variableType here? Won't KernelFunction#invokeAsync still need to handle this argument in the manner of convertToType?

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/FunctionInvocation.java
```diff
@@ -0,0 +1,191 @@
...
+        return this;
+    }
+
+    public FunctionInvocation<T> withToolCallBehavior(@Nullable ToolCallBehavior toolCallBehavior) {
+        this.toolCallBehavior = unmodifiableClone(toolCallBehavior);
+        return this;
+    }
+
+    @Nullable
+    private static UnmodifiableToolCallBehavior unmodifiableClone(
```
dsgrieve, line 100: Just a nit... can the private methods come after the public and protected methods?

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
```diff
@@ -59,25 +54,19 @@ public Kernel(
...
-        @Nullable KernelFunctionArguments arguments) {
-        return invokeAsync(function, arguments, null);
-    }
-
-    public <T> Mono<FunctionResult<T>> invokeAsync(
-        KernelFunction<T> function,
-        @Nullable KernelFunctionArguments arguments,
-        @Nullable ContextVariableType<T> variableType) {
-        return invokeAsync(function, arguments, variableType, null);
+    public <T> FunctionInvocation<T> invokeAsync(
```
dsgrieve, line 57: Are you sure about not making KernelFunction fluent as well? It seems a little odd to have parallel API that isn't parallel.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/InvocationContext.java
```diff
@@ -28,7 +29,13 @@ public InvocationContext(
         this.toolCallBehavior = unmodifiableClone(toolCallBehavior);
     }
 
-    @SuppressFBWarnings(value="EI_EXPOSE_REP", justification="returns UnmodifiableKernelHooks")
```
dsgrieve, line 31: What's the downside to removing InvocationContext altogether?

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/contextvariables/ContextVariable.java
```diff
@@ -46,6 +49,28 @@ public static ContextVariable<?> untypedOf(
...
+    public static <T, U> ContextVariable<T> convert(
+        @Nullable
+        U it,
+        ContextVariableType<T> requestedResultType) {
+        try {
+            return convert(
+                it,
+                requestedResultType.getClazz(),
+                new ContextVariableTypes(
+                    Arrays.asList(requestedResultType.getConverter()))
```
dsgrieve, line 61: Why not just append the default types to this array?

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/semanticfunctions/KernelFunctionFromMethod.java
```diff
@@ -167,89 +167,102 @@ private static MethodDetails getMethodDetails(
...
+                        } else {
+                            return (T) it;
+                        }
+                    });
+
+                return r
+                    .map(it -> {
+                        // If given a variable type, use it.
+                        // If it's wrong, then it's a programming error on the part of the caller.
+                        if (variableType != null) {
```
dsgrieve, line 215: For void functions, does the code even get to here if null is returned? If we map void return to Mono.empty(), then we for sure get here, but you end up returning a FunctionResult<Mono> which isn't the right thing. At least that's what the code did before these changes. So I guess the question is, what gets returned when the void methods are called in Example09?

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/FunctionInvocation.java
```diff
@@ -0,0 +1,191 @@
...
+        if (arguments == null) {
+            this.arguments = null;
+        } else {
+            this.arguments = new KernelFunctionArguments(arguments);
+        }
+        return this;
+    }
+
+    public <U> FunctionInvocation<U> withResultType(ContextVariableType<U> resultType) {
+        return new FunctionInvocation<>(
```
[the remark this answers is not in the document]
johnoliver: I have changed the resultType to being immutable. The reason a new FunctionInvocation is created is because providing a resultType can change the type, notice that this function converts from `FunctionInvocation<T>` to `FunctionInvocation <U>` if the user requests a different type

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/FunctionInvocation.java
johnoliver: invokeAsync no longer performs any type conversion to the required return type, by passing null here we effectively allow it to return whatever it naturally would

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/orchestration/InvocationContext.java
johnoliver: you will end up passing a lot of extra arguments around in many locations

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/semanticfunctions/KernelFunctionFromMethod.java
johnoliver: this code is not reached for a void function, you end up returning a function of `Mono<FunctionResult>`, which imo is correct as you end up with a mono that never returns a result

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/FunctionInvocation.java
johnoliver: also note that in the absence of an explicit return type the functions own metadata return type is used

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/semanticfunctions/KernelFunctionFromMethod.java
dsgrieve: If I have a chain of function calls in the reactive stream and one of the functions is a void function, the stream is broken.

dsgrieve (approved):

PR #4916, opened 2024-02-07
.Net - Update "InternalUtilities" solution folder to include current code files.

crickman:
### Motivation and Context

Confusing that `InternalUtilities` does not display existing code files.

### Description

Synchronize solution view of `InternalUtilities` code files.

### Contribution Checklist


dotnet/SK-dotnet.sln
```diff
@@ -99,6 +99,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Diagnostics", "Diagnostics"
 	ProjectSection(SolutionItems) = preProject
 		src\InternalUtilities\src\Diagnostics\CompilerServicesAttributes.cs = src\InternalUtilities\src\Diagnostics\CompilerServicesAttributes.cs
 		src\InternalUtilities\src\Diagnostics\ExceptionExtensions.cs = src\InternalUtilities\src\Diagnostics\ExceptionExtensions.cs
+		src\InternalUtilities\src\Diagnostics\ExperimentalAttribute.cs = src\InternalUtilities\src\Diagnostics\ExperimentalAttribute.cs
```
glahaye, line 102: Feel free to ignore but maybe the best thing to do would be to make InternalUtilities a csproj that creates a lib and that can be included in the dependencies of the projects that need it instead of relying on InternalUtilities.props and individual file addition in the solution file...

dmytrostruk: > make InternalUtilities a csproj that creates a lib and that can be included in the dependencies of the projects

I also thought about that, but if you want to rely on InternalUtilities assembly in our public packages, you will have to make InternalUtilities public as well, and distribute it in same way as other packages, which is not preferred option.

crickman: I agree but it isn't my intention to alter the existing structure at this point.

SergeyMenshykh (approved):

markwallace-microsoft (approved):

alliscode (approved):

Issue #4067, opened 2023-12-07
.Net: Configurable handlebars planner prompt


rnortje-sygnia:
The handlebars planner is definitely more powerful and useful.  However, we find that being able to add extra domain knowledge/content that explains terms or concepts allows it to better choose the appropriate functions. The current prompt is inaccessible, and one cannot inject extra content into it, thus limiting its usability in certain domains. I understand that the prompt is very specialized, and one would not necessarily like users to break it, so if you want to keep the prompt closed to modification, I suggest some way to inject additional content, maybe via the configuration settings. Or alternatively allow us to specify our own prompt to use as an alternative to the built in one. Not sure how you see it best to tackle this, but it will be much appreciated if this can be added.  Currently I am forced to build my own planner from source code, which is not sustainable in the longer term.

matthewbolanos:
@teresaqhoang, is this something that will be supported when you enable partials support in the Handlebars template engine?

teresaqhoang:
Yes, that's exactly what the partials will help do! 

@rnortje-sygnia The plan is to split each section of the planner into isolated partials that we'll structure together piece-by-piece, so users will have more control on the order / structure of the planner prompt (omit or add any custom sections of their choosing).

Still thinking the design of exactly how to do this, where certain sections may be gated from omission, but what kind of additional content are you thinking of adding and where do you see that fitting into the existing prompt?

rnortje-sygnia:
We have some unique domain knowledge that helps the LLM understand the domain, specifically around how we use terms and interpret relationships. Currently, you could possible send that in as part of the ask I suppose. But in order for it to understand the descriptions of the functions we provide, it needs to understand our domain language, we found by adding this to most of our prompts we improve the ability of the LLM to reason over our domain quite drastically.

Now the handlebars planner is quite unique, in that it requires very specific instruction on how to get efficient plans generated using handlebars and all the helpers etc. The helpers you add dynamically, the core prompt on efficient handlebars usage etc I would not want to change as you guys would be the experts in how to get it working best, but we would need to be able to add custom instructions and content around the domain usage. For now we have tried to add this to our function descriptions, but it leads to repetitive very length descriptions. Personally, since you can add context variable to the prompt easily, I would just provide some extra config options for the planner where you can set extra content and variable for it to use when generating the plan, if not supplied these area then just skipped by the planner, similarly to the if statements you have around the allow loops functions.

But as you say, your solution below will address this if the planner prompt sections can be controlled.

[...]

PR #4918, opened 2024-02-07
.Net: Configurable Handlebars Planner prompt

teresaqhoang:
### Motivation and Context

Addresses issues:
- #4067 

This PR adds more flexibility to the HandlebarsPlanner by:
- Splitting the `CreatePlanPrompt` into well-defined partials. Each partial maps to an essential section within the prompt.
- Opening options to provide more context when invoking the planner.

Not in scope of this PR:
- Option to override CreatePlan prompt completely

### Description

As the `CreatePlan` prompt has been broken down into Partials, users have the ability to override these partials (any or all) to customize the prompt to their needs or add new partials. This can be done by modifying the `CreatePlanPrompt.handlebars` file.

New config enables the ability to call the HandlebarsPlanner with:
- a set of KernelArguments, which will serve as the initial context for the Handlebars Prompt Template; available during plan creation or execution
- additional prompt context: any domain knowledge or specific content that might help the model better fulfill the goal when creating a plan
Additional content after the user's goal, with variables shared in the same message as the goal and additional context as its own separate user message.

### Contribution Checklist


TaoChenOSU (approved):

alliscode (approved):

SergeyMenshykh (approved):

Cotspheer:
@teresaqhoang Sorry to bother you but I'm really confused how I can adjust the CreatePlanPrompt.handlebars. I was excited to test the changes but I struggle after trying a couple things.

I saw that there is CreatePlanPromptHandler but that does not what its name and documentation suggests. It only appends a bit at the very end of the pre-generated prompt. [HandlebarsPlannerExtensions.cs#L65](https://github.com/microsoft/semantic-kernel/blob/d1b8f259ee62e7638b8dc51fdab51d0f50b81565/dotnet/src/Planners/Planners.Handlebars/Handlebars/Extensions/HandlebarsPlannerExtensions.cs#L65). Everything else is untouched.

I also tried to create a new File "CreatePlanPrompt.handlebars" and mark it as Embedded Resource but that does not do the trick as the ExecutingAssembly will always be "Microsoft.SemanticKernel.Planners.Handlebars". So I tried the convention thing that is used for Localization-Resources => creating a folder structure matching the namespace and place the file there but this does also not do it.

What I thought that should be now possible is that I can create a custom CreatePlanPrompt.handlebars and only include the bits I need. I first tried to use ConstructHandlebarsPrompt/ReadAllPromptPartials directly but due the fact that the Extension-Methods are internal I can't use them. Neither in the CreatePlanPromptHandler as I do not get any Context object for example. 

I've also looked at the example changes here [Example65_HandlebarsPlanner.cs](https://github.com/teresaqhoang/semantic-kernel/blob/cc127f85ab204810500a53a5c50e0b78f4c46a51/dotnet/samples/KernelSyntaxExamples/Example65_HandlebarsPlanner.cs) but I didn't manage to find something that could help me.

What I hoped I can do is that I now only can insert the Partials I really need, add my custom ones or overwrite existing ones. 

Is there any documentation or full example available?

Thank you for your time, thoughts and any help!

**This is the result using HandlebarsPlannerOptions.CreatePlanPromptHandler. My part only gets appended to the end:**
```handlebars
{{#*inline "ComplexTypes"}}
## Complex types
Some helpers require arguments that are complex objects. The JSON schemas for these complex objects are defined below:

{{! Complex Parameter Types}}
{{#each complexTypeDefinitions}}
### {{Name}}:
{
  "type": "Object",
  "properties": {
  {{#each Properties}}
    "{{Name}}": {
      "type": "{{ParameterType.Name}}",
    },
  {{/each}}
  }
}

{{/each~}}

{{! JSON Schema Definitions}}
{{#each complexSchemaDefinitions}}
### {{@key}}:
{{this}}

{{/each}}
{{/inline}}
{{! Kernel functions as helpers}}
{{#*inline "CustomHelpers"}}
{{#if (or complexTypeDefinitions complexSchemaDefinitions)}}
{{> ComplexTypes }}
{{/if}}
## Custom helpers
Lastly, you have the following custom helpers to use.

{{#each functions}}
### `{{doubleOpen}}{{PluginName}}{{../nameDelimiter}}{{Name}}{{doubleClose}}`
Description: {{Description}}
Inputs:
  {{#each Parameters}}
    - {{Name}}:
    {{~#if Schema}} {{getSchemaTypeName this}} -
    {{~else}}
        {{~#if ParameterType}} {{ParameterType.Name}} -{{/if}}
    {{~/if}}
    {{~#if Description}} {{Description}}{{/if}}
    {{~#if IsRequired}} (required){{else}} (optional){{/if}}
  {{/each}}
Output:
{{~#if ReturnParameter}}
  {{~#if ReturnParameter.ParameterType}} {{ReturnParameter.ParameterType.Name}}
  {{~else}}
    {{~#if ReturnParameter.Schema}} {{getSchemaReturnTypeName ReturnParameter Name}}
    {{else}} string{{/if}}
  {{~/if}}
  {{~#if ReturnParameter.Description}} - {{ReturnParameter.Description}}{{/if}}
{{/if}}

{{/each}}
{{/inline}}
{{#*inline "LoopsAllowedExample"}}
{{#message role="user"}}## Goal
I want you to generate 10 random numbers and send them to another helper.
{{~/message}}

{{#message role="assistant"}}Here's a Handlebars template that achieves the goal:
```handlebars
\{{!-- Step 0: Extract key values --}}
\{{set
  "count"
  10
}}
\{{!-- Step 1: Loop using the count --}}
\{{#each
  (range
    1
    count
  )
}}
  \{{!-- Step 2: Create random number --}}
  \{{set
    "randomNumber"
    (Example{{nameDelimiter}}Random
      seed=this
    )
  }}
  \{{!-- Step 3: Call example helper with random number and print the result to the screen --}}
  \{{set
    "result"
    (Example{{nameDelimiter}}Helper
      input=randomNumber
    )
  }}
  \{{json (concat "The result" " " "is:" " " result)}}
\{{/each}}
```{{/message}}
{{/inline}}
{{#*inline "NoLoopsExample"}}
{{#message role="user"}}## Goal
What's the sum of 5+10+15?
{{~/message}}

{{#message role="assistant"}}Here's a Handlebars template that achieves the goal:
```handlebars
\{{!-- Step 0: Extract key values --}}
\{{set
  "num1"
  5
}}
\{{set
  "num2"
  10
}}
\{{set
  "num3"
  15
}}
\{{!-- Step 1: Call the Example helper with the variables and store the result --}}
\{{set
  "sum"
  (Example{{nameDelimiter}}AddNums
    num1=num1
    num2=num2
    num3=num3
  )
}}
\{{!-- Step 2: Print the result using the json helper --}}
\{{json (concat "The sum of " num1 "+" num2 "+" num3 " " "is" " " sum)}}
```{{/message}}
{{/inline}}
{{! Built-in helpers from the Handlebars library}}
{{#*inline "BlockHelpers"}}
## Built-in block helpers
- `\{{#if}}\{{/if}}`
- `\{{#unless}}\{{/unless}}`{{#if allowLoops}}
- `\{{#each}}\{{/each}}` - inside this block, you can use:
    - `this` to reference the element being iterated over
    - `@index` to reference the current index
    - `@key` to reference the current key (for object iteration){{/if}}
- `\{{#with}}\{{/with}}`
{{/inline}}
{{! Built-in system helpers from Semantic Kernel}}
{{#*inline "ComparisonHelpers"}}
## Comparison helpers
If you need to compare two values, you can use the `\{{equals}}` helper.
To use the {{#if allowLoops}}math and {{/if}}comparison helpers, you must pass in two positional values. For example, to check if the variable `var` is equal to number `1`, you would use the following helper like so: `\{{#if (equals var 1)}}\{{/if}}`.
{{/inline}}
{{! Built-in system helpers from Semantic Kernel}}
{{#*inline "LoopHelpers"}}
## Loop helpers
If you need to loop through a list of values with `\{{#each}}`, you can use the following helpers:
- `\{{range}}` – Generates a list of integral numbers within a specified range, inclusive of the first and last value.
- `\{{array}}` – Generates an array of values from the given values (zero-indexed).

IMPORTANT: `range` and `array` are the only supported data structures. Others like `hash` are not supported. Also, you cannot use any methods or properties on the built-in data structures.
{{/inline}}
{{! Built-in system helpers from Semantic Kernel}}
{{#*inline "MathHelpers"}}
## Math helpers
If you need to do basic operations, you can use these two helpers with numerical values:
- `\{{add}}` – Adds two values together.
- `\{{subtract}}` – Subtracts the second value from the first.
{{/inline}}
{{! Built-in system helpers from Semantic Kernel}}
{{#*inline "VariableHelpers"}}
## Variable helpers
If you need to create or retrieve a variable, you can use the following helpers:
- `\{{set}}` – Creates a variable with the given name and value. It does not print anything to the template, so you must use `\{{json}}` to print the value.
- `\{{json}}` – Serializes the given value and prints result as JSON string.
- `\{{concat}}` – Concatenates the given values into one string.
{{/inline}}
{{#*inline "AdditionalContext"}}
{{#if additionalContext}}
{{#message role="user"}}
Here is additional context that can help you achieve the goal. While the helpers provided are tools for building your template, the context here will inform the content and logic that guide your use of these tools.

### Additional Context
{{!-- Any domain knowledge or specific content that might help the model better fulfill the goal. Remember to keep the context relevant to the goal and helpers available.
   -- This section should just hold static information that's the same for every request. It should not include variables, use KernelArguments when invoking the planner instead.--}}
{{additionalContext}}
{{/message}}
{{/if}}
{{/inline}}
{{#*inline "Helpers"}}
{{#message role="user"}}The following helpers are available to you:

{{! Built-in helpers from Handlebars and Semantic Kernel}}
{{> BlockHelpers }}

{{#if allowLoops}}
{{> LoopHelpers }}
{{/if}}

{{> MathHelpers }}

{{> ComparisonHelpers }}

{{> VariableHelpers }}

{{! Kernel functions as helpers}}
{{> CustomHelpers }}
IMPORTANT: You can only use the helpers that are listed above. Do not use any other helpers that are not explicitly listed here. For example, do not use `\{{log}}` or any `\{{Example}}` helpers, as they are not supported.
{{/message}}
{{/inline}}
{{#*inline "IntroductionWithExample"}}
{{#message role="system"}}## Instructions
Explain how to achieve the user's goal using the available helpers with a Handlebars .Net template.

## Example
If the user posed the goal below, you could answer with the following template.
{{~/message}}

{{#if allowLoops}}{{> LoopsAllowedExample }}
{{~else}}{{> NoLoopsExample }}
{{/if}}
{{#message role="system"}}Now let's try the real thing.{{/message}}
{{/inline}}
{{#*inline "RetryLogic"}}
{{~#if lastError}}
{{#message role="system"}}## Previous attempt
This previous plan failed to achieve the goal:
handlebars
{{lastPlan}}


The error was:
{{lastError}}

Try again to achieve the goal while fixing the error.
{{/message}}
{{/if}}
{{/inline}}
{{#*inline "TipsAndInstructions"}}
{{#message role="system"}}
## Tips and reminders
- Add a comment above each step to describe what the step does.
- Each variable should have a well-defined name.
- Be extremely careful about types. For example, if you pass an array to a helper that expects a number, the template will error out.
- Each step should contain only one helper call.

## Start
Follow these steps to create one Handlebars template to achieve the goal:
0. Extract Key Values:
  - Read the goal and any user-provided content carefully and identify any relevant strings, numbers, or conditions that you'll need. Do not modify any data.
  - When generating variables or helper inputs, only use content that the user has explicitly provided or confirmed. If the user did not explicitly provide specific information, you should not invent or assume this information.
  - Use the `\{{set}}` helper to create a variable for each key value.
  - Omit this step if no values are needed from the initial context.
1. Choose the Right Helpers:
  - Use the provided helpers to manipulate the variables you've created. Start with the basic helpers and only use custom helpers if necessary to accomplish the goal.
  - Be careful with syntax, i.e., Always reference a custom helper by its full name and remember to use a `#` for all block helpers.
2. Don't Create or Assume Unlisted Helpers:
  - Only use the helpers provided. Any helper not listed is considered hallucinated and must not be used.
  - Do not invent or assume the existence of any functions not explicitly defined above.
3. What if I Need More Helpers?
  - Stop here if the goal cannot be fully achieved with the provided helpers or you need a helper not defined, and just return a string with an appropriate error message.
4. Keep It Simple:{{#if allowLoops}}
  - Avoid using loops or block expressions. They are allowed but not always necessary, so try to find a solution that does not use them.{{/if}}
  - Your template should be intelligent and efficient, avoiding unnecessary complexity or redundant steps.
5. No Nested Helpers:
  - Do not nest helpers or conditionals inside other helpers. This can cause errors in the template.
6. Output the Result:
  - Once you have completed the necessary steps to reach the goal, use the `\{{json}}` helper and print only your final template.
  - Ensure your template and all steps are enclosed in a ``` handlebars block.

Remember, the objective is not to use all the helpers available, but to use the correct ones to achieve the desired outcome with a clear and concise template.
{{/message}}
{{/inline}}
{{#*inline "UserGoal"}}
{{#message role="user"}}## Goal
{{goal}}

{{~#if predefinedArguments}}

### Predefined Variables
You have these predefined variables that can be used within the template. You can access these variables using `@root` (e.g., `@root.variableName`).
Only use the `@root` object when referencing values from this initial context.

{{#each predefinedArguments}}
- "{{@key}}" ({{Type}}): {{Value}}
{{/each}}

Please note that these variables are not necessarily relevant to your goal. Before you use a predefined variable in the template, ensure it makes sense in the context of the goal and helpers available. Remember to extract key values from the goal as well.
{{/if}}
{{/message}}
{{/inline}}
Help me fulfill my goal! {{> UserGoal }}
```

**My Console Code:**
```c#
using Microsoft.SemanticKernel.Planning.Handlebars;

var mockPromptOverride = "Help me fulfill my goal!";

var plannerOptions = new HandlebarsPlannerOptions()
{
    AllowLoops = true,
    CreatePlanPromptHandler = () => $"{mockPromptOverride} {{{{> UserGoal }}}}"
};

var planner = new HandlebarsPlanner(plannerOptions);

// set a break point here and step into the code => this will throw an exception as IChatCompletionService is not configured
var plan = await planner.CreatePlanAsync(new Microsoft.SemanticKernel.Kernel(), "Hello, {{name}}!");

Console.WriteLine(plan.Prompt);
```
**My Project-File:**
```csproj
<Project Sdk="Microsoft.NET.Sdk">

	<PropertyGroup>
		<OutputType>Exe</OutputType>
		<TargetFramework>net8.0</TargetFramework>
		<ImplicitUsings>enable</ImplicitUsings>
		<Nullable>enable</Nullable>
		<NoWarn>SKEXP0001,SKEXP0010,SKEXP0020,SKEXP0050,SKEXP0060</NoWarn>
	</PropertyGroup>

	<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
	  <Optimize>False</Optimize>
	</PropertyGroup>

	<ItemGroup>
	  <EmbeddedResource Include="CreatePlanPrompt.handlebars">
	    <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
	  </EmbeddedResource>
	</ItemGroup>

	<ItemGroup>
		<PackageReference Include="Microsoft.SemanticKernel.Planners.Handlebars" Version="1.6.3-preview" />
	</ItemGroup>

	<ItemGroup>
	  <Folder Include="Microsoft\SemanticKernel\Planners\Handlebars\" />
	</ItemGroup>

</Project>
```
**My CreatePlanPrompt.handlebars file:**
```handlebars
{{> UserGoal }}
```

PR #4919, opened 2024-02-07
Added support for BinaryData for ImageContent.

### LINKED ISSUE #4781: .Net: Add overload ctor to ImageContent that takes an string "url"
When using the Vision models of AzureOpenAI and OpenAI, the API allows for an Image to be attached either as an URL or as a Data URI. In Semantic Kernel we do have `ImageContent` as a KernelContent-Component, but unfortunately, it takes only an `System.Uri` as the parameter for the object. 

I suggest to following addition to `ImageContent`:

### Proposed API
```c#
public sealed class ImageContent : KernelContent
{
    // existing, I think we should just change the type
    // this is a breaking change, although it don't think we should have both
    // but alternatively we could just add another Property 'DataUri' instead of changing the type
    public Uri? Uri { get; set; }
    // replaced by this:
    public string Uri { get; set; }

    // existing, though changed to call the new overload instead of base
    public ImageContent(
        Uri uri,
        string? modelId = null,
        object? innerContent = null,
        Encoding? encoding = null,
        : this(uri.ToString(), modelId, innerContent, encoding, metadata)
    {
        this.Uri = uri;
    }

    public ImageContent(
        string dataUri,
        string? modelId = null,
        object? innerContent = null,
        Encoding? encoding = null,
    {
        this.Uri = uri;
    }
}
``` 
This would allow us to use Data URIs with embedded images for the Vision APIs.

### Usage
```c#
var chatHistory = new ChatHistory();
var image = System.Convert.ToBase64String(imageBytes);
var imageDataLink = $"data:{imageContentType};base64,{image}";
chatHistory.Add(new ImageContent(imageDataLink));
var result = await chat.GetChatMessageContentAsync(chatHistory);
```

### Alternate Design
This would be a breaking Change, since we change the Type of the `Uri` property on `ImageContent`, so instead 
* we could add another property `DataUri` to `ImageContent`, but this would also mean that we have to check which of the two properties are used and prefer one other the other, if both are filled. 
* We could also just make a `private string field` that holds the DataUri and when the `Uri` ctor is used we just store the Uri as string in that field and only use the field on implementation site.  
* We could also make `ImageContent` take a ROS<byte>, byte[] or even just a Stream of the image and handle all the Base64Encoding in the `ImageContent` ctor (I think I would prefer a static Create method over ctor for this option, sonce we can than read the stream async), but this would mean, that we also need a second parameter with the image-ContentType and this would not allow for non base64 encoded data uris

I am happy to work on the implementation, since I have to do it anyway, because we need the solution fairly soon.

dersia:
Added Tests

### Motivation and Context

As Described in #4781 right now there is no possibility  in SK to add Images as DataUris to ChatCompletion APIs, although the Azure OpenAI API and the Open AI API both support this.

Fixes #4781

### Description

As per Discussion added overload to the ImageContent ctor that takes BinaryData.
For backward Compat we kept the ctor that takes an URI.

Also the new ctor throws, if the BinaryData is null, empty or if there is not MediaType provided.

I thought about allowing plain, non base64 encoded DataUris with BinaryData.
The Idea was to not encode to base64, if the MediaType is set to "text/plain", but then I decided, that this is not needed, since `Uri` in general allows for DataUris like `new Uri("data:text/plain;http://exmpaledomain.com")` just not for DataUris that are longer than 65520 bytes. I feel like that is ok, for plain DataUris.

We can still add this if needed.

Also as per discussion in the issue, I did not add additional overloads for direct Streams support.

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Net.Mime;
 using System.Text;
 
 namespace Microsoft.SemanticKernel;
@@ -15,6 +16,11 @@ public sealed class ImageContent : KernelContent
     /// The URI of image.
     /// </summary>
     public Uri? Uri { get; set; }
+
+    /// <summary>
+    /// The Data used as DataUri for the image.
+    /// </summary>
+    public BinaryData? Data { get; }
 
     /// <summary>
     /// Initializes a new instance of the <see cref="ImageContent"/> class.
@@ -35,9 +41,43 @@ public sealed class ImageContent : KernelContent
         this.Uri = uri;
     }
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ImageContent"/> class.
+    /// </summary>
+    /// <param name="data">The Data used as DataUri for the image.</param>
+    /// <param name="modelId">The model ID used to generate the content</param>
+    /// <param name="innerContent">Inner content</param>
+    /// <param name="encoding">Encoding of the text</param>
+    /// <param name="metadata">Additional metadata</param>
+    public ImageContent(
+        BinaryData data,
+        string? modelId = null,
+        object? innerContent = null,
+        Encoding? encoding = null,
+        IReadOnlyDictionary<string, object?>? metadata = null)
+        : base(innerContent, modelId, metadata)
+    {
+        if (data.IsEmpty || string.IsNullOrWhiteSpace(data?.MediaType))
+        {
+            throw new ArgumentNullException(nameof(data), "MediaType is needed for DataUri Images");
+        }
+
+        this.Data = data;
+    }
+
     /// <inheritdoc/>
     public override string ToString()
     {
-        return this.Uri?.ToString() ?? string.Empty;
+        return this.BuildDataUri() ?? this.Uri?.ToString() ?? string.Empty;
+    }
+
+    private string? BuildDataUri()
+    {
+        if (this.Data is null)
+        {
+            return null;
+        }
+
+        return $"data:{this.Data.MediaType};base64,{Convert.ToBase64String(this.Data.ToArray())}";
     }
 }
--- a/dotnet/src/SemanticKernel.UnitTests/Contents/ImageContentTests.cs
+++ b/dotnet/src/SemanticKernel.UnitTests/Contents/ImageContentTests.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
+
 using Microsoft.SemanticKernel;
+
 using Xunit;
 
 namespace SemanticKernel.UnitTests.Contents;
@@ -12,10 +14,10 @@ namespace SemanticKernel.UnitTests.Contents;
 public sealed class ImageContentTests
 {
     [Fact]
-    public void ToStringReturnsString()
+    public void ToStringForUriReturnsString()
     {
         // Arrange
-        var content1 = new ImageContent(null!);
+        var content1 = new ImageContent((Uri)null!);
         var content2 = new ImageContent(new Uri("https://endpoint/"));
 
         // Act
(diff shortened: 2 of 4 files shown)


dersia:
@microsoft-github-policy-service agree [company="Sia Consulting GmbH"]

dersia:
@microsoft-github-policy-service agree company="Sia Consulting GmbH"

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
```diff
@@ -16,6 +17,11 @@ public sealed class ImageContent : KernelContent
     /// </summary>
     public Uri? Uri { get; set; }
 
+    /// <summary>
+    /// The Data used as DataUri for the image.
+    /// </summary>
+    public BinaryData? Data { get; }
+
     /// <summary>
```
Krzysztof318, line 25: But which has higher priority uri or base64 and why?

dotnet/src/SemanticKernel.UnitTests/Contents/ImageContentTests.cs
```diff
@@ -26,4 +28,107 @@ public void ToStringReturnsString()
...
+        // Arrange
+        var data = BinaryData.FromString("this is a test", null);
+        var data2 = BinaryData.FromString("this is a test", string.Empty);
+        var data3 = BinaryData.FromString("this is a test", " ");
+
+        // Assert
+        Assert.Throws<ArgumentNullException>(() => new ImageContent((BinaryData)null!));
+        Assert.Throws<ArgumentNullException>(() => new ImageContent(data));
+        Assert.Throws<ArgumentNullException>(() => new ImageContent(data2));
+        Assert.Throws<ArgumentNullException>(() => new ImageContent(data3));
```
Krzysztof318, line 75: You should use [Theory] and inline data in such cases.

dotnet/src/SemanticKernel.UnitTests/Contents/ImageContentTests.cs
```diff
@@ -26,4 +28,107 @@ public void ToStringReturnsString()
...
+        var result1 = content1.ToString();
+        var dataUriToExpect = $"data:{data.MediaType};base64,{Convert.ToBase64String(data.ToArray())}";
+
+        // Assert
+        Assert.Equal(dataUriToExpect, result1);
+
+        // Assert throws if mediatype is null
+        Assert.Throws<ArgumentNullException>(() => new ImageContent(BinaryData.FromBytes(bytes, null)));
+        Assert.Throws<ArgumentNullException>(() => new ImageContent(BinaryData.FromBytes(bytes, string.Empty)));
+        Assert.Throws<ArgumentNullException>(() => new ImageContent(BinaryData.FromBytes(bytes, " ")));
```
Krzysztof318, line 106: Same [Theory] and [InlineData] and throws assertions should be in another test.

dotnet/src/SemanticKernel.UnitTests/Contents/ImageContentTests.cs
```diff
@@ -26,4 +28,107 @@ public void ToStringReturnsString()
...
+    }
+
+    [Fact]
+    public void ToStringForDataUriFromStreamReturnsDataUriString()
+    {
+        // Arrange
+        var ms = new System.IO.MemoryStream(System.Text.Encoding.UTF8.GetBytes("this is a test"));
+        try
```
Krzysztof318, line 114: tests projects support c#11 you can do it simpler.
instead `System.Text.Encoding.UTF8.GetBytes("this is a test")`
you can write `"this is a test"u8

Krzysztof318 (commented):
In general I don't think this is a good solution. If we were to introduce base64 support it would probably be better to create a new class like ImageBinaryContent. I doubt the SK team would agree to this change. By the way, if you make any changes create a new branch instead working on a copy of the main.
And to implement this, a fix in the implementation of the openai connector would be required....

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
dersia: As the `ToString` override suggests, Data has priority. This is also assured by Tests

dotnet/src/SemanticKernel.UnitTests/Contents/ImageContentTests.cs
dersia: Unfortunately the `MemoryStream` ctor does not take a `ROS<byte>` and `u8` gives back a `ROS<byte>`.

I could do `"this is a test"u8.ToArray()` not sure if is any better, though

dersia:
> In general I don't think this is a good solution. If we were to introduce base64 support it would probably be better to create a new class like ImageBinaryContent. I doubt the SK team would agree to this change. By the way, if you make any changes create a new branch instead working on a copy of the main. And to implement this, a fix in the implementation of the openai connector would be required....

I usually create branches for features, I am not sure, why this happend this time. My Fault, Sorry.
I will push the changes, once they're confirmed, I will close this PR and create a new one from the other branch.

Regarding the issue with the OpenAI Connector, you are totally correct, I thought that it would use the `ToString()` method and use that for the SDK.

I have commented on an open issue regarding this and also referenced this issue. Once this is fixed, I will adjust the implementation to the Connectors.

Krzysztof318:
@dersia You can leave this PR with main branch as is. It doesn't matter for PR, only for your convenience.

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
Krzysztof318: `ToString()` rather will not used in connectors. We should clearly explain to others that they should prioritize a BinaryData over an Uri. For example by adding additional information to the doc comment in ImageContent class.

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
```diff
@@ -16,6 +16,11 @@ public sealed class ImageContent : KernelContent
     /// </summary>
     public Uri? Uri { get; set; }
 
+    /// <summary>
+    /// The Data used as DataUri for the image.
```
RogerBarreto, line 20: @dersia Thanks for the changes.
I would remove the DataUri from the comment on this field and move it to the ToString.

```suggestion
    /// The image binary data.
```

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
```diff
@@ -35,9 +40,43 @@ public sealed class ImageContent : KernelContent
...
+    {
+        if (string.IsNullOrWhiteSpace(data?.MediaType) || data!.IsEmpty)
+        {
+            throw new ArgumentNullException(nameof(data), "MediaType is needed for DataUri Images");
+        }
+
+        this.Data = data;
+    }
+
     /// <inheritdoc/>
```
RogerBarreto, line 67:
```suggestion
    /// <summary>
    /// Returns the string representation of the image.
    /// BinaryData images will be represented as DataUri
    /// Remote Uri images will be represented as is
    /// </summary>
```

dotnet/src/SemanticKernel.UnitTests/Contents/ImageContentTests.cs
```diff
@@ -26,4 +28,107 @@ public void ToStringReturnsString()
...
+
+        // Assert
+        Assert.Equal(dataUriToExpect, result1);
+    }
+
+    [Fact]
+    public void ToStringForDataUriFromStreamReturnsDataUriString()
+    {
+        // Arrange
+        var ms = new System.IO.MemoryStream(System.Text.Encoding.UTF8.GetBytes("this is a test"));
```
RogerBarreto, line 113: Any special reason why are you manually handling the disposal of `MemoryStream` in the test?

Can be simplified with the `using var`

```suggestion
        using var ms = new System.IO.MemoryStream(System.Text.Encoding.UTF8.GetBytes("this is a test"));
```

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
dersia: I was also looking at an issue opened in runtime asking to remove the length limitation on Uri. I think this is in general a good idea, but this won't be GA before net9 and this is way too late. SK and AI development is moving so quickly right now, that most wouldn't want to wait for that change.

But even with that change, I think `BinaryData` gives developers a lot of comfortable possibilities and they don't have to do all the base64 encoding and serialization themselfs. So maybe once the limit on Uri is removed, we could in the implementation just switch back to `Uri` as an implementation detail. I really hope the issue on `Azure.AI.OpenAI` is moving forward sooner rather than later, so we can finally merge this change.

Krzysztof318:
@dersia btw. new class was introduced to kernel [BinaryContent](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs)

I know this news may not be relevant to this PR, but it is worth taking a view.

dersia:
I didn't mean to close this PR. I will open a new one and reference this PR

PR #4922, opened 2024-02-08
.Net: Tool filters

### LINKED ISSUE #4300: .Net: Use automatic function invocation in the function calling stepwise planner

gitri-ms:
### Motivation and Context

Resolves #4300 
Associated ADR: #4686 
### Description

- Implement tool filters for automatic tool calling (similar to kernel function filters). These filters give developers visibility and control over context such as the tool call details, chat history, and the ability to stop auto invoke or tool calling altogether.
- Update the function calling stepwise planner to leverage tool filters in order to respect the maximum iterations setting in the planner options.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example66_FunctionCallingStepwisePlanner.cs
```diff
@@ -24,7 +24,7 @@ public async Task RunAsync()
 
         var options = new FunctionCallingStepwisePlannerOptions
         {
-            MaxIterations = 15,
+            MaxIterations = 3,
```
gitri-ms, line 27: Change this back before checking in

dotnet/src/Connectors/Connectors.OpenAI/ToolFilterContext.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The filter may set <see cref="Cancel"/> to true to indicate that the operation should
+    /// be canceled. If there are multiple filters registered, subsequent filters
+    /// may see and change a value set by a previous filter. The final result is what will
+    /// be considered by the component that triggers filter.
+    /// </remarks>
+    public bool Cancel { get; set; }
+
+    //public int ToolInvocations { get; } // ?
```
gitri-ms, line 81: remove

dotnet/src/Connectors/Connectors.OpenAI/ToolFilterContext.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    /// <summary>
+    ///
```
gitri-ms, line 84: Fill out comment

dotnet/src/Connectors/Connectors.OpenAI/ToolFilterContext.cs
```diff
@@ -0,0 +1,87 @@
...
+    StopTools,
+
+    /// <summary>
+    /// Continue requesting tools, but turn off auto-invoke
+    /// </summary>
+    StopAutoInvoke
+};
+
+/// <summary>
+///
```
gitri-ms, line 35: Fill out comment

dotnet/src/Connectors/Connectors.OpenAI/ToolFilterContext.cs
```diff
@@ -0,0 +1,87 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using Microsoft.SemanticKernel.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Connectors.OpenAI;
+
+/// <summary>
+///
```
gitri-ms, line 9: Fill out comment

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -70,82 +66,56 @@ public sealed class FunctionCallingStepwisePlanner
...
             chatHistoryForSteps.AddUserMessage(StepwiseUserMessage);
-            var chatResult = await this.GetCompletionWithFunctionsAsync(chatHistoryForSteps, clonedKernel, chatCompletion, stepExecutionSettings, logger, cancellationToken).ConfigureAwait(false);
+            var chatResult = await this.GetCompletionWithFunctionsAsync(iteration, chatHistoryForSteps, clonedKernel, chatCompletion, stepExecutionSettings, logger, cancellationToken).ConfigureAwait(false);
             chatHistoryForSteps.Add(chatResult);
 
-            // Check for function response
-            if (!this.TryGetFunctionResponse(chatResult, out IReadOnlyList<OpenAIFunctionToolCall>? functionResponses, out string? functionResponseError))
+            // Increment iteration based on the number of model round trips that occurred as a result of the request
+            object? value = null;
+            chatResult.Metadata?.TryGetValue("ModelIterations", out value);
```
gitri-ms, line 98:
```suggestion
            chatResult.Metadata?.TryGetValue("Iterations", out value);
```

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -329,6 +331,27 @@ await foreach (Completions completions in response)
                     continue;
                 }
 
+                try
```
gitri-ms, line 334: Apply filters to streaming chat result method as well

dotnet/src/Connectors/Connectors.OpenAI/ToolFilterContext.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// <summary>
+    /// Continue requesting tools, but turn off auto-invoke
+    /// </summary>
+    StopAutoInvoke
+};
+
+/// <summary>
+/// 
+/// </summary>
+public abstract class ToolFilterContext // TODO: make experimental?
```
gitri-ms, line 37: add experimental attribute

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -519,6 +593,27 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
...
```
gitri-ms, line 596: Still need test coverage for streaming. I am having issues getting the tests running, probably due to not mocking the streaming model responses properly. Working on it...

dmytrostruk: @gitri-ms Here is unit test for streaming and function call scenario, maybe it will help:
https://github.com/microsoft/semantic-kernel/blob/6652ce61c20ef8081a5dd3cc211a53e933d53e15/dotnet/src/Connectors/Connectors.UnitTests/OpenAI/ChatCompletion/AzureOpenAIChatCompletionServiceTests.cs#L436

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -176,25 +176,27 @@ await foreach (Completions completions in response)
...
     {
-        return new Dictionary<string, object?>(5)
+        return new Dictionary<string, object?>(6)
         {
             { nameof(completions.Id), completions.Id },
             { nameof(completions.Created), completions.Created },
             { nameof(completions.PromptFilterResults), completions.PromptFilterResults },
             { nameof(completions.SystemFingerprint), completions.SystemFingerprint },
             { nameof(completions.Usage), completions.Usage },
+            { "Iterations", iterations },
```
markwallace-microsoft, line 188: Can you define a constant for "Iterations"

Also the name for this variable is a bit ambiguous e.g. is it the maximum iterations, or completed iterations or something  else.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -367,6 +369,27 @@ await foreach (Completions completions in response)
...
+                        this.UpdateChatHistory(chat, chatOptions, chatExecutionSettings);
+
+                        // Check if filter has requested a stop
+                        this.HandleStopBehavior(invokingContext, chatOptions, ref autoInvoke);
+                    }
+                }
+                catch (OperationCanceledException)
+                {
+                    // Add cancellation message to chat history, turn off tools, and bail out of any remaining tool calls
+                    AddResponseMessage(chatOptions, chat, null, "A tool filter requested cancellation before tool invocation.", toolCall.Id, this.Logger);
```
markwallace-microsoft, line 388: I'd like to be able to tell which iteration this occurred on.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -367,6 +369,27 @@ await foreach (Completions completions in response)
                     continue;
                 }
 
+                try
+                {
+                    // Invoke the pre-invocation filter.
+                    var invokingContext = chatExecutionSettings.ToolCallBehavior?.OnToolInvokingFilter(openAIFunctionToolCall, chat, iteration);
+                    if (invokingContext is not null)
+                    {
+                        // Need to update the chat options in case chat history has changed
```
markwallace-microsoft, line 378: Can you elaborate here or in the documentation for the method what happens to make this a required step

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -557,6 +631,27 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
...
+                        this.HandleStopBehavior(invokingContext, chatOptions, ref autoInvoke);
+                    }
+                }
+                catch (OperationCanceledException)
+                {
+                    // Add cancellation message to chat history, turn off tools, and bail out of any remaining tool calls
+                    AddResponseMessage(chatOptions, chat, streamedRole, toolCall, metadata, null, "A tool filter requested cancellation before tool invocation.", this.Logger);
+                    chatOptions.ToolChoice = ChatCompletionsToolChoice.None;
+                    break;
+                }
```
markwallace-microsoft, line 653: This block looks to be repeated , can it be extracted into a method? Same comment for invoked block below.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -447,6 +490,37 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
...
+                break;
+            case ToolFilterStopBehavior.StopTools:
+                chatOptions.ToolChoice = ChatCompletionsToolChoice.None;
+                break;
+            case ToolFilterStopBehavior.Cancel:
+                throw new OperationCanceledException();
+        }
+    }
+
+    private void UpdateChatHistory(ChatHistory chatHistory, ChatCompletionsOptions options, OpenAIPromptExecutionSettings executionSettings)
```
alliscode, line 508: This is being called every time a filter is called right? Is that to handle the case of the filter handler changes the history in some way and we need to sync it? Do we do any updates to the history to clean up if the tool call is canceled?

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
```diff
@@ -35,6 +37,12 @@ public abstract class ToolCallBehavior
     /// </remarks>
     private const int DefaultMaximumAutoInvokeAttempts = 5;
```
alliscode, line 38: Are we going to expose this configuration?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
gitri-ms: > This is being called every time a filter is called right? Is that to handle the case of the filter handler changes the history in some way and we need to sync it? 

Yes and yes. If the filter changes the chat history, we would want that updated chat history to be included in the next request to the model. This helper function updates the chat history stored in `chatOptions` (which is what is sent to the model) to match what is in the `chatHistory` object.

> Do we do any updates to the history to clean up if the tool call is canceled?

If the tool call is cancelled *before* invocation, instead of a response message, we add an error message to the chat history. If the tool call is cancelled *after* invocation, I just have it keeping the regular tool response message in the chat history, but bailing out of future tool calls. (I am not necessarily opposed to a different behavior here, but this made the most sense to me.) In both these cases, `AddResponseMessage` handles updating both the `chatHistory` and `chatOptions` objects.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
gitri-ms: Will change to "ModelIterationsCompleted". It's the number of round trips to the model that have occurred for the request.

gitri-ms:
> Could you extract common abstraction for ToolFilters and move that to the SemanticKernel.Abstraction? Then I could use this abstraction in the gemini connector.

@Krzysztof318 Tool filters take a dependency on the tool call construct which is specific to OpenAI. We don't currently have a generic abstraction for tool calls, because we haven't seen many (any?) other model providers that support it. When we start to see that, I think we can revisit moving this abstraction out of the connector.

For your Gemini work, have you looked at `IFunctionFilter`? It is similar to tool filters, but applies the filters right before/after invoking an SK function. It doesn't have the OpenAI specific stuff, so might work for your needs? Check out the documentation here: https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0033-kernel-filters.md

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
gitri-ms: I would argue that should be in a separate PR, since it's unrelated to the tool filters or the planner updates that this PR covers. Also, if we expose this field, do we want to expose `ToolCallBehavior.MaximumUseAttempts` as well?

stephentoub:
Does gemini support function calling?

Krzysztof318:
@gitri-ms 
@stephentoub 

Yes, gemini has support for Function calling. I will come back to filters topic when I will finish the implementation.

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
gitri-ms: Actually, since it's pretty minor, I've included it in this PR. Will also add a unit test for it.

gitri-ms:
Not sure why these unit tests are failing -- they are passing locally for me. Investigating....

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
```diff
@@ -35,6 +37,12 @@ public abstract class ToolCallBehavior
...
 
+    /// <summary>
+    /// Gets the collection of filters that will be applied to tool calls.
+    /// </summary>
+    [Experimental("SKEXP0016")]
+    public IList<IToolFilter> Filters { get; } = new List<IToolFilter>();
```
stephentoub, line 44: I don't think we want this here, at least not as a mutable list. It means any code can just do `ToolCallBehavior.AutoInvokeKernelFunctions.Filters.Add(myFilter)`, and it'll be added to the singleton that will apply to everyone, which also means you need to remember to remove filters after you're done with them, even in the case of exception.

I think instead we should add overloads to the existing factories below, e.g.
```C#
public static ToolCallBehavior EnableFunctions(
    IEnumerable<OpenAIFunction>? functions, // if null, functions are sourced from the Kernel ala AutoInvokeKernelFunctions,
    EnableFunctionsOptions options);
...
public sealed class EnableFunctionsOptions
{
    public bool AutoInvoke { get; set; }
    public IList<IToolFilter> Filters { get; }
    ... // any other customization desired
}
```
or something along those lines.  That's just a sketch; names and overall shape are debatable.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -557,6 +638,19 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
                     continue;
                 }
 
+                try
+                {
+                    // Invoke the pre-invocation filter.
+                    var invokingContext = chatExecutionSettings.ToolCallBehavior?.OnToolInvokingFilter(openAIFunctionToolCall, chat, iteration);
```
stephentoub, line 644: There's a lot that happens between here and the actual function invocation, including actually getting the corresponding KernelFunction object and creating the KernelArguments to pass to it. I see the value of a callback that's really early in the process, so that a callback can see the raw request from the model, but in that case, should it actually be even earlier and be passed just the raw string name and string arguments?  And then have a separate callback that's invoked with the Kernel, KernelFunction, KernelArguments, etc. just before function.InvokeAsync is actually called?

I think it'd be helpful to enumerate all the extensibility scenarios we're trying to enable here, i.e. the different things we expect folks will want to do with this, and then write out the example code for each, showing both that it's possible and what the code would look like. Those can all become samples, too.

For example:
- Want to limit the number of back and forths with the model, to avoid runaway costs or infinite loops, disabling additional function calling after some number of iterations
- Want to update what functions are available based on the interactions with the model
- Want to limit the number of recursive function invocations that can be made (e.g. agents talking back and forth to each other via function calling)
- Want to screen the arguments being passed to a function and replace the argument with something else
- Want to screen the results of a function and replace it with something else (it's possible this and the above would already be handled by normal function filters)
- Want to stop iterating if a particular function is requested, returning that function's result as the result of the operation (basically the eventual invocation of that function was the ultimate goal)
- ...

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -557,6 +638,19 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
...
+                    this.ApplyToolFilterContextChanges(invokingContext, chatOptions, chat, chatExecutionSettings, ref autoInvoke);
+                }
+                catch (OperationCanceledException)
```
stephentoub, line 647: This feels a bit icky to me. Does this mean that we're saying the way you early-exit non-exceptionally is to throw an OperationCanceledException?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -597,6 +691,18 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
...
 
+                try
+                {
+                    // Invoke the post-invocation filter.
+                    var invokedContext = chatExecutionSettings.ToolCallBehavior?.OnToolInvokedFilter(openAIFunctionToolCall, functionResult, chat, iteration);
+                    this.ApplyToolFilterContextChanges(invokedContext, chatOptions, chat, chatExecutionSettings, ref autoInvoke);
+                }
+                catch (OperationCanceledException)
+                {
+                    // This tool call already happened so we can't cancel it, but bail out of any remaining tool calls
```
stephentoub, line 702: We're adding a message in the case of early-exit before tool invocation but not after tool invocation?

dotnet/src/Connectors/Connectors.OpenAI/ToolInvokedContext.cs
```diff
@@ -0,0 +1,31 @@
...
+public sealed class ToolInvokedContext : ToolFilterContext
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ToolInvokedContext"/> class.
+    /// </summary>
+    /// <param name="toolCall">The <see cref="OpenAIFunctionToolCall"/> with which this filter is associated.</param>
+    /// <param name="result">The result of the tool's invocation.</param>
+    /// <param name="chatHistory">The chat history associated with the operation.</param>
+    /// <param name="modelIterations">The number of model iterations completed thus far for the request.</param>
+    public ToolInvokedContext(OpenAIFunctionToolCall toolCall, object? result, ChatHistory chatHistory, int modelIterations)
```
stephentoub, line 21: FunctionResult instead of object for the result? Also, for the function invoked filters, there's more information supplied, e.g. Kernel, KernelFunction, KernelArguments, etc.  Should this include all that, too? Can/should it derive from the same base classes as the other filters?

crickman:
@dmytrostruk  FYI - Historical

PR #4932, opened 2024-02-08
.Net: Audio-to-Text abstraction and OpenAI implementation

dmytrostruk:
### Motivation and Context

This PR contains new interface for audio-to-text services and implementation of OpenAI connector to use Azure/OpenAI Whisper model. The implementation relies on Azure .NET SDK.

### Description

1. Added new `IAudioToTextService` interface.
2. Added new `AzureOpenAIAudioToTextService` and `OpenAIAudioToTextService` implementations.
3. Added new `OpenAIAudioToTextExecutionSettings` to define audio-to-text configuration (e.g. filename, language etc.).
4. Added unit and integration tests.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/AudioToText/IAudioToTextService.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Get text content from audio binary data.
+    /// </summary>
+    /// <param name="audioData">Audio data.</param>
+    /// <param name="executionSettings">The AI execution settings (optional).</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Text content from audio binary data.</returns>
+    Task<TextContent> GetTextContentAsync(
+        BinaryData audioData,
```
Krzysztof318, line 26: Will we accept an audio data only as binary also in the other models, never as an URI?

dmytrostruk: That's a good question, thanks. Most probably I will change it from `BinaryData` to `AudioContent` model, which is available in other PR for text-to-audio feature. `AudioContent` currently has `BinaryData`, but most probably will also contain `Uri` which will keep remote link to audio. Something similar should be done to `ImageContent` (there is already existing PR for that) for the ability to keep image content in different formats.

markwallace-microsoft (approved):
LGTM

SergeyMenshykh (approved):

RogerBarreto (approved):

Swimburger:
Does the `BinaryData` class load all audio data into memory? If so, this doesn't seem like an efficient class for this.
If so, can we change the interface to have separate overloads for streams vs byte arrays vs memory? 🙇

dmytrostruk:
> Does the `BinaryData` class load all audio data into memory? If so, this doesn't seem like an efficient class for this. If so, can we change the interface to have separate overloads for streams vs byte arrays vs memory? 🙇

@Swimburger There is no need in adding overload to interface, we can add additional property to `AudioContent` type to hold not only `BinaryData` but `Stream` as well. The problem is that Azure .NET SDK, which we are using for this functionality, accepts `BinaryData` as audio input, so even if we add `Stream` on our side, we will have to convert it to `BinaryData` before passing it to Azure .NET SDK. 
https://github.com/Azure/azure-sdk-for-net/blob/2dc0ca8a6a16025011fca42c671f0052bab922a0/sdk/openai/Azure.AI.OpenAI/src/Custom/Audio/AudioTranscriptionOptions.cs#L38

As soon as `AudioTranscriptionOptions` class will accept other types than `BinaryData`, we will add it on our side as well.

Swimburger:
Got it. I am implementing the new `IAudioToTextService` and including it into our SK integration library, but our customers often transcribe large files, many gigabytes, so loading this into memory isn't a viable option. I'm surprised this isn't an issue for the Azure .NET SDK.

Krzysztof318:
@Swimburger your customer send one big file with one request? Without chunking?
With openai endpoint is not possible.
From OpenAI docs:
> File uploads are currently limited to 25 MB and the following input file types are supported: mp3, mp4, mpeg, mpga, m4a, wav, and webm.

Swimburger:
We support up to 2.2GB file uploads, and less relevant, but we also support 5GB files if they are uploaded already to a public URL.
We support these file types: .3ga, .webm, .8svx, .mts, .m2ts, .ts, .aac, .mov, .ac3, .mp2, .aif, .mp4, .m4p (with DRM), .m4v .aiff, .mxf, .alac, .amr, .ape, .au, .dss, .flac, .flv, .m4a, .m4b, .m4p, .m4r, .mp3, .mpga, .ogg, .oga, .mogg, .opus, .qcp, .tta, .voc, .wav, .wma, .wv.

dmytrostruk:
> We support up to 2.2GB file uploads, and less relevant, but we also support 5GB files if they are uploaded already to a public URL. We support these file types: .3ga, .webm, .8svx, .mts, .m2ts, .ts, .aac, .mov, .ac3, .mp2, .aif, .mp4, .m4p (with DRM), .m4v .aiff, .mxf, .alac, .amr, .ape, .au, .dss, .flac, .flv, .m4a, .m4b, .m4p, .m4r, .mp3, .mpga, .ogg, .oga, .mogg, .opus, .qcp, .tta, .voc, .wav, .wma, .wv.

@Swimburger I believe we are limited at the moment by API of AI provider with this file size. In order to resolve that issue, you will have to think about different strategy, for example data chunking. But even with data chunking, with 2.2GB file and 25MB limitation on OpenAI side, you will have to send 88 chunks just for one file. It depends on your business logic, if you have enough time to process large files, then maybe some background job will cover this scenario for you. 

To avoid API limitations, you can also try to use some AI audio-to-text model which can be hosted locally (you can check available models on Hugging Face for example).

Krzysztof318:
@dmytrostruk Maybe it will be better add stream overload however? This will give us better abstraction for work with local models and live streaming.

dmytrostruk:
> @dmytrostruk Maybe it will be better add stream overload however? This will give us better abstraction for work with local models and live streaming.

@Krzysztof318 https://github.com/microsoft/semantic-kernel/pull/5074

Swimburger:
To be clear, we are an ASR + many other models provider, not a consumer :)
We host our own model which are only accessible by through our API. We don't support running the models locally or through another host ATM.

SK has been doing a good job not limiting itself by OpenAI's limitations, so I think the same should apply here.
Thank you for working on this!

Krzysztof318:
@Swimburger  So you can add a new connector to yours models in Semantic Kernel via PR, then we can reconsider adding stream https://github.com/microsoft/semantic-kernel/pull/5074#discussion_r1494606743. But if your models are private and not be accessible for everyone, I don't think SK team will bend abstraction for closed implementation of connectors.

Swimburger:
We'll create a PR for it!

PR #4933, opened 2024-02-08
.Net: Text-to-Audio abstraction and OpenAI implementation

dmytrostruk:
### Motivation and Context

This PR contains new interface for text-to-audio services and implementation of OpenAI connector to use Azure/OpenAI TTS model. The implementation is custom (for now) and could be replaced with Azure .NET SDK when such functionality will be available.

### Description

1. Added new `ITextToAudioService` interface.
2. Added new `AzureOpenAITextToAudioService` and `OpenAITextToAudioService` implementations.
3. Added new `OpenAITextToAudioExecutionSettings` to define text-to-audio configuration (e.g. voice, speed, response_format etc.).
4. Added new `AudioContent` class which will store binary data of audio.
5. Added unit and integration tests.

### Contribution Checklist


dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/AzureOpenAITextToAudioService.cs
```diff
@@ -0,0 +1,157 @@
...
+        Kernel? kernel = null,
+        CancellationToken cancellationToken = default)
+        => this.InternalGetAudioContentAsync(text, executionSettings, cancellationToken);
+
+    #region private
+
+    private async Task<AudioContent> InternalGetAudioContentAsync(
+        string text,
+        PromptExecutionSettings? executionSettings = null,
+        CancellationToken cancellationToken = default)
```
SergeyMenshykh, line 90: For private methods, it's better not to set a default value for the cancellation token to force the method callers to provide it to make sure the cancellation token is propagated through method call chain within the class.

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/AzureOpenAITextToAudioService.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        this.AddAttribute(DeploymentNameKey, deploymentName);
+        this.AddAttribute(AIServiceExtensions.ModelIdKey, modelId);
+    }
+
+    /// <inheritdoc/>
+    public Task<AudioContent> GetAudioContentAsync(
+        string text,
+        PromptExecutionSettings? executionSettings = null,
+        Kernel? kernel = null,
```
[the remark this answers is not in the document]
dmytrostruk: As per our agreement, we pass kernel everywhere so all plugins and connectors will be able to work with it. In all our abstractions (chat completion, text generation etc) we accept nullable kernel instance, so specific connector can use it when required.

For this connector, kernel usage is not required, but maybe for `OtherTextToAudioService` in the future it will be required.

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/AzureOpenAITextToAudioService.cs
dmytrostruk: Good point, thanks! I removed default cancellation token for private/internal methods in both audio PRs.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
```diff
@@ -0,0 +1,37 @@
...
+/// <summary>
+/// Represents audio content return from a <see cref="ITextToAudioService" /> service.
+/// </summary>
+[Experimental("SKEXP0005")]
+public class AudioContent : KernelContent
+{
+    /// <summary>
+    /// The audio binary data.
+    /// </summary>
+    public BinaryData? AudioData { get; set; }
```
RogerBarreto, line 19: Suggest using only `Data` for the property name, as it is already an AudioContent.

Similar to [this](https://github.com/microsoft/semantic-kernel/pull/4919/files/859c75ea8fbfbec707ad8f5ad5c36d73efe767b3#diff-a122a929d0ddc07714a2a7dbe567ec7a063898c994aec34895e8c44de463b36fR23) `ImageContent` PR.

dmytrostruk: Thanks, I updated property name to `Data`.

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/AzureOpenAITextToAudioService.cs
```diff
@@ -0,0 +1,161 @@
...
+
+/// <summary>
+/// Azure OpenAI text-to-audio service.
+/// </summary>
+[Experimental("SKEXP0005")]
+public sealed class AzureOpenAITextToAudioService : ITextToAudioService
+{
+    private readonly Dictionary<string, object?> _attributes = new();
+    private readonly ILogger _logger;
+    private readonly HttpClient _httpClient;
```
markwallace-microsoft, line 26: The pattern we want to promote is having a seperate client to encapsulate all of the Http logic. Ideally we will get an official client for this service which we can use in the future.

I'm OK with this change happening in another PR, if that is the plan please link the issue and close this comment.

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/OpenAITextToAudioService.cs
```diff
@@ -0,0 +1,148 @@
...
+public sealed class OpenAITextToAudioService : ITextToAudioService
+{
+    /// <summary>
+    /// Gets the attribute name used to store the organization in the <see cref="IAIService.Attributes"/> dictionary.
+    /// </summary>
+    public static string OrganizationKey => "Organization";
+
+    private readonly Dictionary<string, object?> _attributes = new();
+    private readonly ILogger _logger;
+    private readonly HttpClient _httpClient;
```
markwallace-microsoft, line 31: Same comment about having a service client.

markwallace-microsoft (commented):
Couple of comments to address

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/AzureOpenAITextToAudioService.cs
dmytrostruk: > Ideally we will get an official client for this service which we can use in the future.

That's exactly the reason why I placed all the logic together, I'm expecting to use official client for this functionality in the future. But we don't know when it's going to happen, so I moved all HTTP-related logic to separate client in this PR. Thanks for feedback!

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/OpenAITextToAudioService.cs
dmytrostruk: Done

markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/OpenAITextToAudioService.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.SemanticKernel.Contents;
+using Microsoft.SemanticKernel.Services;
+using Microsoft.SemanticKernel.TextToAudio;
+
+namespace Microsoft.SemanticKernel.Connectors.OpenAI;
+
+/// <summary>
+/// OpenAI text-to-audio service.
+/// </summary>
+[Experimental("SKEXP0005")]
```
RogerBarreto, line 18: Missing the Experiements.md definition for this index `SKEXP0005`

RogerBarreto (approved):

dotnet/src/Connectors/Connectors.OpenAI/TextToAudio/OpenAITextToAudioService.cs
dmytrostruk: It exists in second PR which will be merged after this one:
image

PR #4940, opened 2024-02-08
.Net: Add home automation example to demonstrate how to use DI in a real app

glahaye:
### Motivation and Context
We want to give a template or starting point for people to get a solid idea of how to use SK in their own apps.

### Description
Add the Home Automation example

### Contribution Checklist


dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace HomeAutomation;
+
+internal static class Program
+{
+    internal static async Task Main(string[] args)
+    {
+        HostApplicationBuilder builder = Host.CreateApplicationBuilder(args);
+        builder.Services.AddHostedService<Worker>();
+        builder.Services.AddSingleton<AppConfiguration>();
+        builder.Services.AddSingleton<Kernel>(sp =>
```
matthewbolanos, line 30: Just realized I mispoke when I sent a message to you earlier... the kernel should probably a transient; I think the services _underneath_ it, should be singletons

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            AppConfiguration ac = sp.GetRequiredService<AppConfiguration>();
+            var builder = Kernel.CreateBuilder()
+                                // Add AI services here
+                                // Note that you can provide your custom http client when adding specific AI services
+                                .AddAzureOpenAIChatCompletion(ac.AzureOpenAiDeployment, ac.AzureOpenAiEndpoint, ac.AzureOpenAiApiKey);
```
matthewbolanos, line 36: This could be moved up as a singleton service in the host services.

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+    private readonly IHostApplicationLifetime _hostApplicationLifetime;
+    private readonly Kernel _kernel;
+
+    public Worker(IHostApplicationLifetime hostApplicationLifetime, Kernel kernel)
+    {
+        _hostApplicationLifetime = hostApplicationLifetime;
+        _kernel = kernel;
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
```
matthewbolanos, line 64: Could we make this a more interactive experience (e.g., Console.Read) so that the developer can test out different prompts without needing to rewrite the code?

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        var chatCompletionService = _kernel.GetRequiredService<IChatCompletionService>();
+
+        // Enable auto function calling
+        OpenAIPromptExecutionSettings openAIPromptExecutionSettings = new() { ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions };
+
+        // Get the response from the AI
+        ChatMessageContent chatResult = await chatCompletionService.GetChatMessageContentAsync(TimePrompt, openAIPromptExecutionSettings, _kernel, stoppingToken);
+        Console.WriteLine($"Result is: {chatResult}");
+
+        MyLightPlugin kitchenLight = new();
```
matthewbolanos, line 80: Should this be located when the kernel is built so everyone that gets the kernel also gets the lights?

dotnet/samples/HomeAutomation/Plugins/MyLightPlugin.cs
```diff
@@ -0,0 +1,38 @@
...
+
+/// <summary>
+/// Class that represents a controllable light.
+/// </summary>
+[Description("Represents a light")]
+public class MyLightPlugin
+{
+    private bool _turnedOn;
+
+    public MyLightPlugin(bool turnedOn = false)
```
matthewbolanos, line 16: Could we demonstrate DI in the light plugin constructor? E.g., logging or a room/house service?

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+
+namespace HomeAutomation;
+
+internal static class Program
+{
+    internal static async Task Main(string[] args)
+    {
+        HostApplicationBuilder builder = Host.CreateApplicationBuilder(args);
+        builder.Services.AddHostedService<Worker>();
```
matthewbolanos, line 28: Could we also demonstrate how to add logging, retry logic, and a custom http handler services? Those patterns have been confusing to folks as well.

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        // Enable auto function calling
+        OpenAIPromptExecutionSettings openAIPromptExecutionSettings = new() { ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions };
+
+        // Get the response from the AI
+        ChatMessageContent chatResult = await chatCompletionService.GetChatMessageContentAsync(TimePrompt, openAIPromptExecutionSettings, _kernel, stoppingToken);
+        Console.WriteLine($"Result is: {chatResult}");
+
+        MyLightPlugin kitchenLight = new();
+        MyLightPlugin bedroomLight = new();
+        _kernel.ImportPluginFromObject(kitchenLight, "Kitchen");
```
matthewbolanos, line 82: This should probably be "KitchenLight" instead of just "Kitchen"

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        // Enable auto function calling
+        OpenAIPromptExecutionSettings openAIPromptExecutionSettings = new() { ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions };
+
+        // Get the response from the AI
+        ChatMessageContent chatResult = await chatCompletionService.GetChatMessageContentAsync(TimePrompt, openAIPromptExecutionSettings, _kernel, stoppingToken);
+        Console.WriteLine($"Result is: {chatResult}");
+
+        MyLightPlugin kitchenLight = new();
+        MyLightPlugin bedroomLight = new();
```
matthewbolanos, line 81: Since we're using `ImportPluginFromObject`, it'd be good to show an example where it's more helpful than just `ImportPluginFromType`. In your current constructor, you can pass in the current state of the lightbulb, you could add that and/or other information (like the room the light bulb is in).

dotnet/samples/HomeAutomation/Program.cs
matthewbolanos: If this is done, we should also add some test cases to make sure this doesn't break in the future. For example,

| Question                  | Required Plugins    |
| ------------------- | ------------------- |
| "What time is it?"    | Time |
| "Can you turn off the ceiling light in the kitchen?" | Light |
| "Can you turn off all the lights in the kitchen?" | Lights |
| "When does the sunset?" | Weather |
| "When does the sunset tomorrow? | Weather + Time |
| "Will you let me know when its 5:00?" | Schedular + Time |
| "Will you turn on the lights at 7:00?" | Schedular + Time + Light |
| "Will you turn on the lights when the sunsets?" | Schedular + Time + Light + Weather |

matthewbolanos (changes requested):
Glad to see this come together! Left some comments based on some of the challenges I've seen customers have with DI.

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: Absolutely.  I'm just keeping it with static strings for now because I don't want to re-type stuff all the time until the app is finalized.  I'll switch it to read input when we're set.

Krzysztof318 (changes requested):
It would be nice have support openai endpoint out-of-box and optional azure. Many of customers doesn't have access the Azure OpenAI.

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace HomeAutomation;
+
+internal static class Program
+{
+    internal static async Task Main(string[] args)
+    {
+        HostApplicationBuilder builder = Host.CreateApplicationBuilder(args);
+        builder.Services.AddHostedService<Worker>();
+        builder.Services.AddSingleton<AppConfiguration>();
```
alliscode, line 29: I think the recommended approach is to use the options pattern for configuration.

One way would be something like:
```
builder.Services.Configure<AzureAiOptions>(builder.Configuration.GetSection(AzureAiOptions.Name));
```

Then things can use it by taking a dependency on IOptions<AzureAiOptions>

dotnet/samples/HomeAutomation/AppConfiguration.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Extensions.Configuration;
+
+namespace HomeAutomation;
+
+/// <summary>
+/// Loads configuration upon instantiation.
+/// </summary>
+internal sealed class AppConfiguration
```
alliscode, line 10: Can we rename this to something more specific like AzureAiOptions? If we need more config later we can create models for the sections we need.

alliscode: Also remove the constructor and make this a simple model that is bound to with the options patern.

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        builder.Services.AddSingleton<Kernel>(sp =>
+        {
+            AppConfiguration ac = sp.GetRequiredService<AppConfiguration>();
+            var builder = Kernel.CreateBuilder()
+                                // Add AI services here
+                                // Note that you can provide your custom http client when adding specific AI services
+                                .AddAzureOpenAIChatCompletion(ac.AzureOpenAiDeployment, ac.AzureOpenAiEndpoint, ac.AzureOpenAiApiKey);
+            // Add plugins to include in kernel here
+            builder.Plugins.AddFromType<MyTimePlugin>();
```
alliscode, line 38: If you register the kernelBuilder with `builder.Services.AddKernel();` then the plugins can be registered as singletons separately from the kernel and will be reused even if the kernel itself is transient.

dotnet/samples/HomeAutomation/Program.cs
alliscode, line 30: Should be transient

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        using IHost host = builder.Build();
+
+        await host.RunAsync();
+    }
+}
+
+/// <summary>
+/// Actual code to run.
+/// </summary>
+internal sealed class Worker : BackgroundService
```
alliscode, line 53: Can we break this out to is own file.

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        Console.WriteLine($"Worker running at: {DateTimeOffset.Now}");
+
+        // Get chat completion service
+        var chatCompletionService = _kernel.GetRequiredService<IChatCompletionService>();
+
+        // Enable auto function calling
+        OpenAIPromptExecutionSettings openAIPromptExecutionSettings = new() { ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions };
+
+        // Get the response from the AI
+        ChatMessageContent chatResult = await chatCompletionService.GetChatMessageContentAsync(TimePrompt, openAIPromptExecutionSettings, _kernel, stoppingToken);
```
alliscode, line 77: It looks like the result of the time plugin isn't used for anything. It would be good to connect the scenario a bit better... I bet the whole flow could be done with a single prompt that says something like "If it's after 7:00 pm then the kitchen light should be off and the bedroom light should be on. If it's after 10PM and before 8AM all lights should be off..." and given the 3 plugins the auto function calling would make it happen.

dotnet/samples/HomeAutomation/AppConfiguration.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: Changed

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: Did something equivalent by moving all registrations to the app DI container instead

dotnet/samples/HomeAutomation/Plugins/MyLightPlugin.cs
glahaye: Sure. But are you sure you want to increase the complexity of the light plugin (and of the example) to add more DI in a way that doesn't illustrate anything related to SK?

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
glahaye: This seems to confuse the LLM. I'm trying to find a simple way to phrase this and get it to perform the required function calls.

dotnet/samples/HomeAutomation/Program.cs
glahaye: Adding this.

dotnet/samples/HomeAutomation/Plugins/MyLightPlugin.cs
matthewbolanos: Yes, a lot of the questions I'm getting right now is how to use DI with plugins in Semantic Kernel. I think the value prop we have by deeply integrating with DI is a key part of why someone would want to use Semantic Kernel instead of another SDK like LangChain.

dotnet/samples/HomeAutomation/appsettings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "AzureOpenAiOptions": {
+    "Deployment": "",
+    "Endpoint": ""
+    // "ApiKey": "" /// Set this value in appsettings.Development.json, or using "dotnet user-secrets"
```
matthewbolanos, line 5: Could we update this so that the user can use the same user-secrets as the KernelSyntaxExamples?

matthewbolanos:
Could we add a .vscode folder with the necessary files so a developer can press the run button to test the sample

dotnet/samples/HomeAutomation/Options/AzureOpenAiOptions.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.ComponentModel.DataAnnotations;
+
+namespace HomeAutomation.Options;
+
+/// <summary>
+/// Loads configuration upon instantiation.
+/// </summary>
+public sealed class AzureOpenAiOptions
```
matthewbolanos, line 10: Could we also support OpenAI as part of the sample?

dotnet/samples/HomeAutomation/appsettings.Development.json
```diff
@@ -0,0 +1,5 @@
+{
+  "AzureOpenAiOptions": {
+    // "ApiKey": "" // Set value here, or using "dotnet user-secrets"
```
matthewbolanos, line 3: Adjusting this file didn't work for me for some reason; I had to update the appsettings.json before the application picked up my configuration.

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+        builder.Services.AddHostedService<Worker>();
+
+        // Get configuration
+        builder.Services.AddOptions<AzureOpenAiOptions>()
+                        .Bind(builder.Configuration.GetSection(nameof(AzureOpenAiOptions)))
+                        .ValidateDataAnnotations()
+                        .ValidateOnStart();
+
+        // Add plugins to include in all kernel instances here
+        builder.Services.AddSingleton<IEnumerable<KernelPlugin>>(sp =>
```
matthewbolanos, line 41: Ben and I synced with Stephen, and he recommended that we could use the following approach to avoid the "magic" feeling of injection a KernelPluginCollection (especially if you have multiple kernels).

```
// Add plugins that can be used by kernels
// The plugins are added as singletons so that they can be used by multiple kernels
builder.Services.AddSingleton<MyTimePlugin>();
builder.Services.AddKeyedSingleton<MyLightPlugin>("OfficeLight");
builder.Services.AddKeyedSingleton<MyLightPlugin>("PorchLight", (sp, key) =>
{
    return new MyLightPlugin(turnedOn: false);
});

// Add a home automation kernel to the dependency injection container
builder.Services.AddKeyedTransient<Kernel>("HomeAutomationAgent", (sp, key) =>
{
    // Create a collection of plugins that the kernel will use
    KernelPluginCollection pluginCollection = new();
    pluginCollection.AddFromObject(sp.GetRequiredService<MyTimePlugin>());
    pluginCollection.AddFromObject(sp.GetRequiredKeyedService<MyLightPlugin>("OfficeLight"), "OfficeLight");
    pluginCollection.AddFromObject(sp.GetRequiredKeyedService<MyLightPlugin>("PorchLight"), "PorchLight");

    return new Kernel(sp, pluginCollection);
});
```

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            return new KernelPlugin[]
```
rosieks, line 43: Please add some plugin the requires dependencies that have to be resolved (eg. database)

dotnet/samples/HomeAutomation/appsettings.json
glahaye: Done.

dotnet/samples/HomeAutomation/Options/AzureOpenAiOptions.cs
glahaye: Will look into this

dotnet/samples/HomeAutomation/appsettings.Development.json
glahaye: I have addressed to root cause to this.  The repo was missing a file required for this file to be taken into consideration because it was it the Git ignore list.

dotnet/samples/HomeAutomation/Program.cs
glahaye: With gpt-4-turbo this works fine now.

dotnet/samples/HomeAutomation/Program.cs
glahaye: Done

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            return new MyLightPlugin(turnedOn: true);
+        });
+
+        // Add a home automation kernel to the dependency injection container
+        builder.Services.AddKeyedTransient<Kernel>("HomeAutomationKernel", (sp, key) =>
+        {
+            // Create a collection of plugins that the kernel will use
+            KernelPluginCollection pluginCollection = new();
+            pluginCollection.AddFromObject(sp.GetRequiredService<MyTimePlugin>());
```
rosieks, line 75: If I want to add OpenAPI based plugin it requires me to call asynchronous method. Please cover that case too.

dotnet/samples/HomeAutomation/Plugins/MyLightPlugin.cs
glahaye: To keep things easy and more obvious, I've created a MyAlarmPlugin which itself takes in a MyTimePlugin as a dependency to illustrate using dependencies to create other dependencies.

dotnet/samples/HomeAutomation/Program.cs
glahaye: I've created a MyAlarmPlugin which itself takes in a MyTimePlugin as a dependency to illustrate this.

dotnet/samples/HomeAutomation/Program.cs
glahaye: Added some comments to that effect

dotnet/samples/HomeAutomation/Options/AzureOpenAiOptions.cs
glahaye: Added information to address this

dotnet/samples/HomeAutomation/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        // Actual code to execute is found in Worker class
+        builder.Services.AddHostedService<Worker>();
+
+        // Get configuration
+        builder.Services.AddOptions<AzureOpenAI>()
+                        .Bind(builder.Configuration.GetSection(nameof(AzureOpenAI)))
+                        .ValidateDataAnnotations()
+                        .ValidateOnStart();
+
```
matthewbolanos, line 40: Could we also show adding a logging service?

dotnet/samples/HomeAutomation/Worker.cs
```diff
@@ -0,0 +1,57 @@
...
+        // Get chat completion service
+        var chatCompletionService = _kernel.GetRequiredService<IChatCompletionService>();
+
+        // Enable auto function calling
+        OpenAIPromptExecutionSettings openAIPromptExecutionSettings = new()
+        {
+            ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions
+        };
+
+        ChatMessageContent chatResult = await chatCompletionService.GetChatMessageContentAsync(
```
matthewbolanos, line 37: Could we update this to be a turn-based chat with the user?

dotnet/samples/HomeAutomation/README.md
```diff
@@ -0,0 +1,40 @@
+# "House Automation" example illustrating how to use Semantic Kernel with dependency injection
+
+This example demonstrates a few dependency injection patterns that can be used with Semantic Kernel.
+
```
matthewbolanos, line 4: Once we turn the sample into a turn-based application, you can add the recommended questions to ask the AI here.

dotnet/samples/HomeAutomation/Program.cs
glahaye: That's the magic of DI: it's already added automagically!

dotnet/samples/HomeAutomation/Program.cs
glahaye: Questions have been suggested in app

dotnet/samples/HomeAutomation/README.md
glahaye: Questions have been suggested in app

matthewbolanos (approved):

alliscode (approved):

PR #4949, opened 2024-02-09
Python: Add CMK support to create Index

charotAmine:
### Motivation and Context

Thank you for your contribution to the semantic-kernel repo!
      When we try to create an index in AI Search where the encryption is enabled we have this issue : 
        Code: OperationNotAllowed
        Message: The creation of non-encrypted Indexes is not alowed when encryption policy is enforced.
      Which make the operation create_collection unusable
      Use create_collection function
      If the customer uses the CMK for the encryption

### Description

The Azure Search Documents supports the encryption key with the class SearchResourceEncryptionKey, adding it (Free to the customer to use it or not) resolve the issue. 
Maybe it may be use it in save_information function as well ?

### Contribution Checklist


python/semantic_kernel/connectors/memory/azure_cognitive_search/azure_cognitive_search_memory_store.py
```diff
@@ -136,6 +141,7 @@ async def create_collection(
                 name=collection_name.lower(),
                 fields=get_index_schema(self._vector_size),
                 vector_search=vector_search,
+                encryption_key=search_resource_encryption_key,
```
eavanvalkenburg, line 144: could you have a look to see if you can add a unit test for this, that tests both with and without key?

charotAmine: Hello @eavanvalkenburg , I have mocked and added the tests.

python/semantic_kernel/connectors/memory/azure_cognitive_search/azure_cognitive_search_memory_store.py
```diff
@@ -83,6 +84,7 @@ async def create_collection(
         self,
         collection_name: str,
         vector_config: Optional[HnswVectorSearchAlgorithmConfiguration] = None,
+        search_resource_encryption_key: SearchResourceEncryptionKey = None,
```
moonbox3, line 87: Tiny comment: I'd love to see the Optional type hint here for type checking, to keep it consistent with the optional type hint for the parameter before it.

charotAmine: hHello @moonbox3  I have added the optional type hint !

moonbox3 (approved):
LGTM

eavanvalkenburg (approved):

PR #4955, opened 2024-02-09
Java: Remove KernelFunctionCollection from public API

dsgrieve:
### Motivation and Context
Simplify the API

### Description
Use KernelFunctionCollection internally from Kernel. Add a couple of methods to Kernel to add/get a KernelPlugin

### Contribution Checklist

### DIFF
--- a/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
+++ b/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
@@ -1,26 +1,26 @@
 // Copyright (c) Microsoft. All rights reserved.
 package com.microsoft.semantickernel;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-
-import javax.annotation.Nullable;
-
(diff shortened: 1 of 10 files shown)


milderhc (approved):

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
```diff
@@ -1,82 +1,92 @@
...
-        KernelFunction function = plugins.getFunction(pluginName, functionName);
+        KernelFunction function = getFunction(pluginName, functionName);
         return invokeAsync(function);
     }
 
     public <T> FunctionInvocation<T> invokeAsync(KernelFunction<T> function) {
         return function.invokeAsync(this);
     }
 
+    public void addPlugin(KernelPlugin plugin) {
```
johnoliver, line 64: should this return `Kernel`?

johnoliver (approved):

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
dsgrieve: To be consistent with fluent style, I guess so. I'll make this adjustment along with the javadoc.

PR #4958, opened 2024-02-10
.Net - Add support for Azure Open AI Assistants API

crickman:
### Motivation and Context

Azure has added support for Open AI Assistants API - https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/assistant.  We now want to support both Open AI and Azure Open AI Assistant API endpoints.

### Description

Add `WithAzureOpenAIChatCompletion` to `AgentBuilder` and updated `OpenAIRestContext` to support `Endpoint` and `Version`.

> Turns out Azure is opinionated on the `object` property in the models and OpenAI doesn't care...so removed.

### Contribution Checklist


Krzysztof318:
I like this pattern with the UseOpenAI property I think we can push this property to the BaseTest class and use also in other examples.

dotnet/samples/KernelSyntaxExamples/Example72_AgentCollaboration.cs
```diff
@@ -19,7 +19,12 @@ public class Example72_AgentCollaboration : BaseTest
     /// Specific model is required that supports agents and function calling.
     /// Currently this is limited to Open AI hosted services.
     /// </summary>
-    private const string OpenAIFunctionEnabledModel = "gpt-4-0613";
+    private const string OpenAIFunctionEnabledModel = "gpt-4-1106-preview";
```
Krzysztof318, line 22: I think you can update the model to gpt-4-0125-preview

alliscode (approved):

crickman:
> I like this solution with the UseOpenAI property I think we can push this property to the BaseTest class and use also in other examples.

I agree with your vision.  I'd prefer not to churn the entire test-suite with this PR.  Happy to cycle-back and work on base-class semantics.

glahaye (approved):

Issue #4440, opened 2023-12-28
.Net: Exceptions thrown from the HandlebarsPlanner should be specific and should contain Exception-Data.


Cotspheer:
Currently when HandlebarsPlanner throws an exception I have no way to recover from that in a meaningful and graceful manner. I don't have the plan, I don't have the prompt template and I have no way to differentiate between Exceptions besides testing the message if it contains a specific text.

**Request:**

- Use of specific meaningful exceptions like InsufficientFunctionsForGoalException instead of KernelException (overall not specific to Planners!)
  - See [HandlebarsPlanner](https://github.com/microsoft/semantic-kernel/blob/240a2cbeaccc2e9562fa9aa82c856f74c5b5ace4/dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs#L91C84-L91C84) 
- Include Details in the Exception-Data-Bag. For the planner for example:
  - Include the created plan / attempt
  - Include the created prompt

**Reason why:**

The message "Additional helpers may be required" triggers the exception but the plan is perfectly fine. So I want to recover from that and therefore I need the created plan. I'm totally able to handle the "Additional helpers may be required" message myself. So my intent is to test for a specific exception then extract the plan form the Exception-Data-Bag and continue with executing the plan. A specific exception would prevent the abuse of the Exception-Data-Bag and the properties may could set explicite.

Maybe the HandlebarsPlannerOptions should be extended with an option: "ShouldThrow" or something similar. Then the HandlebarsPlan could be extended with HasError and an Error-Object. So that HandlebarsPlan is more treated like a "[Result Object](https://medium.com/@wgyxxbf/result-pattern-a01729f42f8c#:~:text=The%20%E2%80%9CResult%20Pattern%E2%80%9D%20is%20an,flow%20to%20run%20more%20smoothly.)".

**Current Code:**

```C#
        var planner = new HandlebarsPlanner(new HandlebarsPlannerOptions()
        {
            AllowLoops = true,
        });

        try
        {
            var plan = await planner.CreatePlanAsync(kernel, intent, cancellationToken);
            this.logger.LogInformation($"Plan: {plan}");
        }
        catch (Exception ex)
        {
            // try to extract the plan
            throw ex;
        }
```

**Example Code:**

```C#
        var plannerOptions = new HandlebarsPlannerOptions()
        {
            AllowLoops = true,
            ShouldThrow = false
        };

[...]

PR #4973, opened 2024-02-12
Adding HandlebarsPlanCreationException

teresaqhoang:
### Motivation and Context

Resolves https://github.com/microsoft/semantic-kernel/issues/4440

This PR adds a more detailed `HandlebarsPlanCreationException` type, which allows users to inspect the prompt and model results, if available.

Als includes a minor update to sample to add clarifying comments.

### Description

Users can catch `HandlebarsPlanCreationException` to inspect the prompt, proposed plan, and exception details on error. All exceptions will be bubbled up to the caller.

```
var planner = new HandlebarsPlanner();

try
{
     var plan = await planner.CreatePlanAsync(kernel, intent, cancellationToken);
}
catch (HandlebarsPlanCreationException ex)
{
     Console.WriteLine(ex.Message);
     Console.WriteLine(ex.InnerException?.Message);
     Console.WriteLine($"CreatePlan Prompt: {ex.CreatePlanPrompt}");
     Console.WriteLine($"Proposed plan (model output): {ex.ModelResults.Content}");
     throw ex;
}
```

### Contribution Checklist


Krzysztof318:
But isn't this a return to how kernel errors were once handled? ADR 0004 describes a departure from such solution in favor of throwing exceptions.

teresaqhoang:
> But isn't this a return to how kernel errors were once handled? ADR 0004 describes a departure from such solution in favor of throwing exceptions.

Great callout @Krzysztof318, that slipped my mind. Pivoted to creating a narrower exception type containing these details.

dotnet/src/Planners/Planners.Handlebars/Handlebars/Exceptions/HandlebarsPlanCreationException.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.Planning.Handlebars;
+
+/// <summary>
+/// Provides extension methods for rendering Handlebars templates in the context of a Semantic Kernel.
+/// </summary>
+internal sealed class HandlebarsPlanCreationException : KernelException
```
markwallace-microsoft, line 8: Should this be internal? Are we not expecting clients to catch this?

dotnet/src/Planners/Planners.Handlebars/Handlebars/Exceptions/HandlebarsPlanCreationException.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+    }
+
+    public HandlebarsPlanCreationException(string? message) : base(message)
+    {
+    }
+
+    public HandlebarsPlanCreationException(string? message, System.Exception? innerException) : base(message, innerException)
+    {
+    }
```
markwallace-microsoft, line 30: Why do we need these extra constructors?

crickman (approved):

dotnet/src/Planners/Planners.Handlebars/Handlebars/Exceptions/HandlebarsPlanCreationException.cs
teresaqhoang: Good catch - made the class public and added XML comments

dotnet/src/Planners/Planners.Handlebars/Handlebars/Exceptions/HandlebarsPlanCreationException.cs
teresaqhoang: To comply with linter rules since we're extending the `KernelException` class

image

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
```diff
@@ -79,33 +79,49 @@ public Task<HandlebarsPlan> CreatePlanAsync(Kernel kernel, string goal, KernelAr
...
+
+            var planTemplate = match.Groups[2].Value.Trim();
+            planTemplate = MinifyHandlebarsTemplate(planTemplate);
 
-        return new HandlebarsPlan(planTemplate, createPlanPrompt);
+            return new HandlebarsPlan(planTemplate, createPlanPrompt);
+        }
+        catch (KernelException ex)
+        {
+            throw new HandlebarsPlanCreationException(
```
SergeyMenshykh, line 118: I am just wondering how useful the "CreatePlan failed. See inner exception for details." message and exception are from the code consumer's point of view. The consumer will have to catch it and use if/switch logic inside to test the internal exception type/details to make a recovery decision.

Wouldn't it be easier for the consumer to catch a more specific exception like UnachievableGoalException with a message, goal, list of functions, and model results?

@stephentoub what do you think?

dotnet/src/Planners/Planners.Handlebars/Handlebars/Exceptions/HandlebarsPlanCreationException.cs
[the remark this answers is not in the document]
teresaqhoang: Done. Renamed to `PlanCreationException` and moved to `Microsoft.SemanticKernel.Planning` namespace

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
teresaqhoang: It's set up this way so that `createPlanPrompt` and `modelResults`, if available, would be populated into all exceptions, accounting for unhandled KernelExceptions. Otherwise, we could just override the KernelException and port the message over to this `PlanCreationException`:

```
throw new PlanCreationException(
                ex.Message,
                createPlanPrompt,
                modelResults
            );
```

The CreatePlanPrompt should contain all the necessary details like list of functions and goal and this field also captures the scenario where a user provides their own custom planner prompt to override the default

SergeyMenshykh (approved):

PR #5002, opened 2024-02-13
Python: .Net: [Agents] Agents abstraction + OpenAIChatCompletionAgent

SergeyMenshykh:
### Motivation and Context
Today, SK has two agents, [Agent](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Experimental/Agents/Internal/Agent.cs) and [ChatCompletionAgent](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Experimental/Agents/ChatCompletionAgent.cs), that do not have anything in common. Their public surfaces are shaped around the API they use to communicate with LLM. Organizing collaboration between them would require a significant amount of code since they have different data contracts and capabilities. Our goal is to build a set of abstractions that minimize the amount of code needed to organize collaboration and additionally allow us to use the agents interchangeably, regardless of the API/LLM utilized by them. 

The goal of the PR is to introduce the MVP of agent abstraction and collect feedback, which would enable us to enhance the design of the abstraction.

### Description
This PR adds the following classes:  
- KernelAgent is a base abstract class to be implemented by all SK agents.  
- AgentMessage is a data contract class for agent inputs and outputs.  
- OpenAIChatCompletionAgent is a real agent that utilizes the OpenAI chat completion API.

### Contribution Checklist


dotnet/src/Experimental/Agents/AgentMessage.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System.Threading;
+using Microsoft.SemanticKernel.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Experimental.Agents;
+
+/// <summary>
+/// Represents agent message used by agents inherited from <see cref="KernelAgent" /> class.
+/// </summary>
+public class AgentMessage
```
LittleLittleCloud, line 11: Any plan to support parallel function call for in this PR?

dotnet/src/Experimental/Agents/ChatCompletion/ChatCompletionAgent.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Experimental.Agents;
+
+/// <summary>
+/// Represent an agent that is built around the SK ChatCompletion API and leverages the API's capabilities.
+/// </summary>
+public sealed class ChatCompletionAgent : KernelAgent
```
crickman, line 14: No Description property?  I'd expect this to be part of the core KernelAgent...along with optional name?

dotnet/src/Experimental/Agents/AgentMessage.cs
matthewbolanos, line 11: Do we need agent message if we already have ChatMessageContent?

dotnet/src/Experimental/Agents/AgentMessage.cs
```diff
@@ -0,0 +1,72 @@
...
+
+    /// <summary>
+    /// Role of the author of the message.
+    /// </summary>
+    public AuthorRole Role { get; set; }
+
+    /// <summary>
+    /// The <see cref="KernelAgent"/> the message was created by.
+    /// </summary>
+    public KernelAgent? Agent { get; set; }
```
crickman, line 26: Will it be practical to marshal an actual agent reference to each message?  (which one?)

dotnet/src/Experimental/Agents/AgentMessage.cs
crickman, line 11: Any thought on the "cititation" requirement we discussed?

dotnet/src/Experimental/Agents/AgentMessage.cs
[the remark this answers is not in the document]
SergeyMenshykh: I considered a server-side scenario where a chat (list of agent messages) is being serialized and deserialized between user sessions. For this scenario, the setter might be useful for chat hydration functionality to inject an instance of the agent after or during the chat deserialization.  
   
I can remove the setter for now and add it later, if necessary, when we have more information or requirements for this.

dotnet/src/Experimental/Agents/KernelAgent.cs
```diff
@@ -0,0 +1,36 @@
...
+        this.Kernel = kernel;
+    }
+
+    /// <summary>
+    /// Invokes the agent to process the given messages and generate a response.
+    /// </summary>
+    /// <param name="messages">A list of the messages for the agent to process.</param>
+    /// <param name="cancellationToken">An optional <see cref="CancellationToken"/> to cancel the operation.</param>
+    /// <returns>List of messages representing the agent's response.</returns>
+    public abstract Task<IReadOnlyList<AgentMessage>> InvokeAsync(IReadOnlyList<AgentMessage> messages, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
SergeyMenshykh: We need to verify that the scenarios are supported. So far, I've identified three ways it can be done:  
1. If two or more assistant agents need to chat in the same thread, they can be configured, at creation time, to use the same thread. This way, all the messages passed to the InvokeAsync method of agent A or agent B will be added to the same shared thread. If no thread is configured for an agent, then every InvokeAsync on that agent will create a thread, add messages to it, execute run (OpenAI Assistants API run), get results, and remove the thread.  
2. We can add a special KernelContent type, let's say - AssistantThreadContent with only one property, ThreadId/Id, and agents can use it to pass information about the thread between each other.  
3. Each AgentMessage has the InternalMessage property, which will have the OpenAI Assistants API message with the ThreadId property. This property can be used by another assistant agent to identify a thread to work with.  
   
I like the first approach because it gives the consumer code more control over agent configuration and thread ownership and lifetime management. In contrast, with the other two approaches, it's not clear who owns the thread, who should create, and remove it.

dotnet/src/Experimental/Agents/ChatCompletion/ChatCompletionAgent.cs
SergeyMenshykh: Not yet, we can always add them later when we know what they are for and how they will be used.

matthewbolanos: +1 to including a description property; this will be necessary to make Agents as plugins feasible.

dotnet/src/Experimental/Agents/AgentMessage.cs
SergeyMenshykh: Agreed, offline, to drop the AgentMessage type and use ChatMessageContent instead.

dotnet/src/Experimental/Agents/AgentMessage.cs
SergeyMenshykh: The scenario I thought about is related to a pre-hook for a chat completion agent that would receive, let's say, an AssistantFileIdContent content type it is not familiar with from an assistant agent. To convert the content type to a BinaryContent, which the chat completion agent can understand, there should be a way to access the message owner (assistant agent) to download the file using the ID. Therefore, having the agent in the message will allow the chat completion agent's pre-hook to download the content by ID, wrap it in BinaryContent, and send it to the chat completion agent who knows how to work with BinaryContent.

dotnet/src/Experimental/Agents/AgentMessage.cs
SergeyMenshykh: I believe it can be easily modeled by another subclass of KernelType; however, we need to confirm that this is the case.

PR #5005, opened 2024-02-13
.Net: Create Plan Prompt Override

teresaqhoang:
### Motivation and Context

This PR adds the option to override the CreatePlan prompt.

### Description

Users can pass in a callback to `HandlebarsPlannerOptions.GetCreatePlanPrompt` that returns a valid Handlebars Template string to be used in place of the default CreatePlan prompt. 

Users have the options to select any predefined partials in their own template. These helpers are pre-registered and available for reference at the time of import override.

### Contribution Checklist


dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlannerOptions.cs
```diff
@@ -24,6 +24,16 @@ public sealed class HandlebarsPlannerOptions : PlannerOptions
...
     public Func<Task<string>>? GetAdditionalPromptContext { get; set; }
 
+    /// <summary>
+    /// Callback that returns an override for the CreatePlan prompt.
+    /// </summary>
+    /// <remarks>
+    /// The callback should return a valid Handlebars template string. If this is set, the planner will use this prompt instead of the default prompt.
+    /// Devs can select any partial defined in Planners.Handlebars.CreatePlanPromptPartials namespace when constructing their own prompt.
+    /// No partials are included by default; make sure to select partials such as "{{> UserGoal }}" or {{> AdditionalContext}} if needed.
+    /// </remarks>
```
markwallace-microsoft, line 34: Can you add a kernel syntax sample showing how to use this and add a link in the API documentation.

markwallace-microsoft (approved):

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlannerOptions.cs
```diff
@@ -24,6 +24,16 @@ public sealed class HandlebarsPlannerOptions : PlannerOptions
...
+    public Func<string>? GetCreatePlanPrompt { get; set; }
```
RogerBarreto, line 35: Suggest changing the name to represent its a handler that will be used as a callback.

```suggestion
    public Func<string>? CreatePlanPromptHandler { get; set; }
```

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlannerOptions.cs
teresaqhoang: Added `RunOverrideCreatePlanPromptSampleAsync` to Example65_HandlebarsPlanner.cs

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlannerOptions.cs
teresaqhoang: Done

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlannerOptions.cs
teresaqhoang: Seems like all of our API documentation is outdated, will work with @matthewbolanos to update docs for Handlebars Planner

glahaye (approved):

PR #5006, opened 2024-02-13
Python : Change from Cognitive Search to AI Search

charotAmine:
### Motivation and Context

  1. Why is this change required? This will change the name of cognitive search to ai search to allign with /Net
  2. What problem does it solve? Cognitive Search was renamed to AI Search
  3. What scenario does it contribute to? Allign with .Net SDK

### Description

**NOTE : FOR THE INTEGRATION TESTS, I HAVE CHANGED THE SECRETS AS WELL, SHOULD I REVERT ?**

### Contribution Checklist

- [X] I didn't break anyone :smile: **Breaking Change !**


.vscode/settings.json
```diff
@@ -9,7 +9,7 @@
   "[csharp]": {
     "editor.defaultFormatter": "ms-dotnettools.csharp",
     "editor.codeActionsOnSave": {
-      "source.fixAll": true
```
moonbox3, line 12: Is it necessary to edit this file?

python/samples/kernel-syntax-examples/azure_chat_gpt_with_data_api.py
```diff
@@ -35,7 +35,7 @@
 
 # Create the data source settings
 az_source = AzureAISearchDataSources(**azure_ai_search_settings)
-az_data = AzureDataSources(type="AzureCognitiveSearch", parameters=az_source)
+az_data = AzureDataSources(type="AzureAiSearch", parameters=az_source)
```
moonbox3, line 38: we have many other places in the codebase that format "AI" like "AzureOpenAI" or "OpenAI." I would prefer to keep the formatting like that, instead of the PascalCase as you are suggesting.

python/semantic_kernel/connectors/memory/azure_ai_search/azure_ai_search_memory_store.py
```diff
@@ -32,7 +32,7 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 
-class AzureCognitiveSearchMemoryStore(MemoryStoreBase):
+class AzureAiSearchMemoryStore(MemoryStoreBase):
```
moonbox3, line 35: Could we also please update the AzureAISearch package to the latest if we aren't using it already?

moonbox3 (changes requested):
Looks good overall, but a few small requests, please.

PR #5008, opened 2024-02-14
.Net: Added support for BinaryData for ImageContent.

### LINKED ISSUE #4781: .Net: Add overload ctor to ImageContent that takes an string "url"

dersia:
Reopend PR. See https://github.com/microsoft/semantic-kernel/pull/4919 for previous PR. 
Reopening this PR from another feature branch makes rebasing easier.

### Motivation and Context

As Described in #4781 right now there is no possibility  in SK to add Images as DataUris to ChatCompletion APIs, although the Azure OpenAI API and the Open AI API both support this.

Fixes #4781

### Description

As per Discussion added overload to the ImageContent ctor that takes BinaryData.
For backward Compat we kept the ctor that takes an URI.

Also the new ctor throws, if the BinaryData is null, empty or if there is not MediaType provided.

I thought about allowing plain, non base64 encoded DataUris with BinaryData.
The Idea was to not encode to base64, if the MediaType is set to "text/plain", but then I decided, that this is not needed, since `Uri` in general allows for DataUris like `new Uri("data:text/plain;http://exmpaledomain.com")` just not for DataUris that are longer than 65520 bytes. I feel like that is ok, for plain DataUris.

We can still add this if needed.

Also as per discussion in the issue, I did not add additional overloads for direct Streams support.

### Contribution Checklist


dersia:
@RogerBarreto I fixed the comments. I also checked for the `using` pattern and I couldn't replicate the behavior I had in the past with the memory stream, so I reverted it back to using the `using` pattern in the test.

dersia:
@Krzysztof318

> @dersia btw. new class was introduced to kernel [BinaryContent](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs) This class is currently used only by OpenAIFileService.
> 
> I know this news may not be relevant to this PR, but it is worth taking a view.

Interesting. Though I'm not sure I like the implementation, the Stream will advance if the `GetStreamAsync` or `GetContentAsync` is called and I also think that `Null`-Content/Stream would be fine and should not throw, it should just not be added to the request to the model.

Krzysztof318:
@dersia I'm not thrilled with it either but I sent it to you as an overview. I think the way it is now is ok but it's best for @RogerBarreto to comment.

RogerBarreto:
> I still think you should add information to the doc class comment that BinaryData takes precedence over URI.

@Krzysztof318 Thats a good point, @dersia you can add this as a `<remark></remark>` in the `ToString` `XmlDocs` comments.

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
```diff
@@ -35,9 +40,47 @@ public sealed class ImageContent : KernelContent
...
+        }
+
+        this.Data = data;
+    }
+
+    /// <summary>
+    /// Returns the string representation of the image.
+    /// BinaryData images will be represented as DataUri
+    /// Remote Uri images will be represented as is
+    /// </summary>
```
RogerBarreto, line 71:
```suggestion
    /// </summary>
    /// <remarks>
    /// When Data is provided it takes precedence over URI
    /// </remarks>

```

RogerBarreto (approved):

dmytrostruk (approved):

PR #5019, opened 2024-02-14
.Net Add Audio examples

Krzysztof318:
### Motivation and Context

Issue https://github.com/microsoft/semantic-kernel/issues/5016

### Description

Added samples which demonstrate using ITextToAudioService and IAudioToTextService.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example28_Audio.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        // Create a kernel with OpenAI text to audio service
+        var kernel = Kernel.CreateBuilder()
+            .AddOpenAITextToAudio(
+                modelId: TextToAudioModel,
+                apiKey: TestConfiguration.OpenAI.ApiKey)
+            .Build();
+
+        var textToAudioService = kernel.GetRequiredService<ITextToAudioService>();
+
```
joslat, line 36: An idea/suggestion would be to use an audio stream here, like capturing the system audio (for example, an ongoing skype/teams talk) and transcribing this live and processing it live - summarization, key topic extraction, etc..

Krzysztof318: I could add such sample but at the end audio file would be sent. This would have more sense if OpenAI support real time streaming for audio services but it doesn't.

markwallace-microsoft (approved):

RogerBarreto (approved):

dotnet/samples/KernelSyntaxExamples/Example28_Audio.cs
joslat: agree on this but for live-audio-streaming processing having this would enable creating real time call center ai support tools... maybe with a whisper local model... 
many companies would jump on SK only for this :)
such https://github.com/collabora/WhisperLive , https://github.com/sandrohanea/whisper.net (using whisper.cpp, https://github.com/ggerganov/whisper.cpp)

Krzysztof318: I created issue https://github.com/microsoft/semantic-kernel/issues/5032

PR #5021, opened 2024-02-14
Handle null content type for 204 responses in RestApiOperationRunner

### LINKED ISSUE #5018: RestApiOperationRunner handling of 204 | No Content responses
Today, RestApiOperationRunner.SerializeResponseContentAsync will throw if there is no `MediaType` on the `ContentType` response header. When the response is 204 or No Content, there will not be a ContentType|MediaType available. I'm curious the SK teams thoughts on whether the behavior should be updated in this scenario to not throw and support 204 responses or if this is the expected behavior and design with OpenAPI plugins.

Example yaml:
```yaml
paths:
  /resources/{id}:
    get:
      operationId: getResource
      summary: Get the details of a resource
      parameters:
      - in: path
        name: id
        schema:
            type: string
        required: true
        description: The ID of the resource.
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ResourceResponse'
components:
  schemas:
    ResourceResponse:
      type: object
      properties:
        resources:
          type: string
          items:
            type: string
          description: The list of resources.
```

### Function execution with API giving a 204 response
**Expected:**
Empty string is returned

**Actual:**
Exception thrown - `Microsoft.SemanticKernel.Kernelexception - No media type available.`

lemillermicrosoft:
## Summary
This pull request adds a check for null content type in the SerializeResponseContentAsync method of RestApiOperationRunner. If the content type is null, an empty response with no content type is returned for 204 responses. This change ensures that the method can handle 204 responses correctly.

## Changes
- Added check for null content type in SerializeResponseContentAsync method
- Return empty response with no content type for 204 responses

---
### Motivation and Context
Fixes #5018 

### Description

This modification checks if the contentType is null for 204 responses. If it is null, it returns an empty response with no content type. Otherwise, it proceeds with the serialization as before. This change allows the method to handle 204 responses gracefully without throwing an exception for missing content types.

### Contribution Checklist


dotnet/src/Functions/Functions.OpenApi/RestApiOperationRunner.cs
```diff
@@ -190,7 +190,14 @@ private static async Task<RestApiOperationResponse> SerializeResponseContentAsyn
     {
         var contentType = content.Headers.ContentType;
 
-        var mediaType = contentType?.MediaType ?? throw new KernelException("No media type available.");
+        // Check if content type is null for 204 response
+        if (contentType == null)
+        {
+            // Return an empty response with no content type for 204 responses
+            return new RestApiOperationResponse(string.Empty, string.Empty);
```
SergeyMenshykh, line 197: Consider making both the Content and ContentType properties of the RestApiOperationResponse class nullable, and pass null instead of an empty string to indicate the absence of content or value.

matthewbolanos:
@lemillermicrosoft, would you be able to make the requested changes? Or should we have someone on the Semantic Kernel team pick this up?

markwallace-microsoft:
@lemillermicrosoft This should be fixed as part of this PR: https://github.com/microsoft/semantic-kernel/pull/6632

PR #5026, opened 2024-02-14
.Net: Improved Audio API by adding default setting values

dmytrostruk:
### Motivation and Context

Current audio interfaces have optional execution settings parameters in order to be able to use audio API through Kernel in the future, where execution settings are optional. But when using audio services directly, this is not user friendly, because when optional execution settings are not configured, services will throw an error. 

In case of text-to-audio service, required field is voice. This PR sets default value to `alloy` with the ability to overwrite.
For audio-to-text service, required field is filename. This PR sets default value to `file.mp3` with the ability to overwrite. 

With these changes execution settings are no longer required (they are optional) for text-to-audio and audio-to-text conversion.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/AudioToText/OpenAIAudioToTextExecutionSettings.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AudioToText/OpenAIAudioToTextExecutionSettings.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using Microsoft.SemanticKernel.Text;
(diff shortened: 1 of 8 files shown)


dotnet/src/Connectors/Connectors.OpenAI/AudioToText/OpenAIAudioToTextExecutionSettings.cs
```diff
@@ -146,6 +156,8 @@ public override PromptExecutionSettings Clone()
 
     #region private ================================================================================
 
+    private const string DefaultFilename = "file.mp3";
```
markwallace-microsoft, line 159: This default file name is required because of an issue in the Azure AI SDK which Dmytro has raised here: https://github.com/Azure/azure-sdk-for-net/pull/40078#discussion_r1480392822

We should leave a note that we want to revisit this when the Azure SDK team have made their change.

markwallace-microsoft (approved):

RogerBarreto (approved):

dotnet/src/Connectors/Connectors.OpenAI/AudioToText/OpenAIAudioToTextExecutionSettings.cs
RogerBarreto: @markwallace-microsoft I think we can create an issue to keep track of this instead of leaving a note in the code. 👍🏻

PR #5061, opened 2024-02-17
.Net: TextClassification (Moderation) OpenAI implementation

### LINKED ISSUE #5062: .Net Implement OpenAI moderation endpoint support
PR #5061

Krzysztof318:
### Motivation and Context

Closes #5062

### Description

This PR introduces support for OpenAI moderation models.

**Most of the code is cloned from my gemini connector code and adjusted to openai.**

I have used the abstract class ClientBase in the code so that it will be possible to use it in the refactoring of  OpenAITextToImageClientCore. The IHttpRequestFaactory interface may have more implementations, such as for LMstudio or ollama (openai api) in the future.

### Contribution Checklist


Krzysztof318:
@markwallace-microsoft @RogerBarreto Could you review?

Krzysztof318:
@markwallace-microsoft I wonder should I rename OpenAIClassificationCategory, OpenAIClassificationEntry, OpenAIClassificationResult to OpenAIModerationCategory, OpenAIModerationEntry, OpenAIModerationResult. I think these classes are only related to moderation endpoint and they won't be applicable to feature classification (others than moderation) openai models.

dotnet/src/SemanticKernel.Abstractions/Contents/ClassificationContent.cs
```diff
@@ -0,0 +1,41 @@
...
+        : base(innerContent, modelId, metadata)
+    {
+        this.ClassifiedContent = classifiedContent;
+        this.Result = result;
+    }
+
+    /// <summary>
+    /// Represents the classified content.
+    /// </summary>
+    public string ClassifiedContent { get; }
```
RogerBarreto, line 35: I think this abstraction is still too abstract (Results Dictionary of `objects?`)

Is not very clear what to expect of the value in the `ClassifiedContent`, is that the `Prompt` sent or ?

This might need more investigation and probably an ADR where we can have a better traversable content for classification models.

Looking at `HuggingFace` inference API for `Text-Classification` we have this result:

Hugging Face
```
[
    [
        {
            "label": "disappointment",
            "score": 0.5044490694999695
        },
        {
            "label": "sadness",
            "score": 0.3469429612159729
        },
        ...
    ]
]
```

Open AI
```
{
    "id": "modr-XXXXX",
    "model": "text-moderation-007",
    "results": [
        {
            "flagged": true,
            "categories": {
                "sexual": false,
                "hate": false,
                "harassment": false,
                "self-harm": false,
                "sexual/minors": false,
                "hate/threatening": false,
                "violence/graphic": false,
                "self-harm/intent": false,
                "self-harm/instructions": false,
                "harassment/threatening": true,
                "violence": true
            },
            "category_scores": {
                "sexual": 1.2282071e-6,
                "hate": 0.010696256,
                "harassment": 0.29842457,
                "self-harm": 1.5236925e-8,
                "sexual/minors": 5.7246268e-8,
                "hate/threatening": 0.0060676364,
                "violence/graphic": 4.435014e-6,
                "self-harm/intent": 8.098441e-10,
                "self-harm/instructions": 2.8498655e-11,
                "harassment/threatening": 0.63055265,
                "violence": 0.99011886
            }
        }
    ]
}
```

Krzysztof318: I am also not convinced with this. But I don't see good way to abstract classification content because results from models are too different. So maybe we should remove abstraction and leave only implementation for classification service?

RogerBarreto:
Closing temporarily for the same reason as #5279

Issue #5070, opened 2024-02-19
Python: AttributeError: 'Choice' object has no attribute 'logprobs' （OpenAITextCompletionBase）


ymuichiro:
**Describe the bug**

This is the same issue as the following: Azure OpenAI API may not include logprobs.

https://github.com/microsoft/semantic-kernel/issues/4923

The recent change only modified OpenAIChatCompletionBase. However, it turns out that this issue also occurs with OpenAITextCompletionBase. Therefore, I propose to make a similar fix.

**To Reproduce**

The following users have tried it:
https://github.com/microsoft/semantic-kernel/pull/4924#issuecomment-1951639962

**Expected behavior**

Logprobs can be None.

**Platform**

 - OS: M2 Mac
 - IDE: VS Code
 - Language: Python
 - Source: pip package version 23.3.2, main branch of repository

**Additional context**

More detailed information is provided below.

- https://github.com/microsoft/semantic-kernel/discussions/5011
- https://github.com/microsoft/semantic-kernel/pull/4924#issuecomment-1951796494

※ I'm the same person as @yuichiromukaiyama

Issue #4923, opened 2024-02-08
Python: AttributeError: 'Choice' object has no attribute 'logprobs'


yuichiromukaiyama:
**Describe the bug**
I ran the following example with SemanticKernel 0.5.0.dev0.
https://github.com/microsoft/semantic-kernel/blob/main/python/notebooks/11-streaming-completions.ipynb

The code is as follows.

```python
service = AzureChatCompletion(
        deployment_name=AZURE_OPENAI_DEPLOYMENT_NAME,
        endpoint=AZURE_OPENAI_ENDPOINT,
        api_key=AZURE_OPENAI_API_KEY,
        api_version="2023-12-01-preview",
)

role = "system"
content = "You are an AI assistant that helps people find information."
message = {"role": role, "content": content}

az_oai_chat_prompt_execution_settings = AzureChatPromptExecutionSettings(
        max_tokens=150,
        temperature=0.7,
        top_p=1,
        frequency_penalty=0.5,
        presence_penalty=0.5,
        ai_model_id="gpt-4",
)
stream = service.complete_chat_stream(
        messages=[message], settings=az_oai_chat_prompt_execution_settings
)
async for text in stream:
         print(
            text[0].inner_content.choices[0].delta.content, end=""
        )
```

When I run it, I get the following error:

```shell
AttributeError: 'Choice' object has no attribute 'logprobs'
```

**Expected behavior**
The response from Azure OpenAI Chat Completion is output.

**Platform**
 - OS: Mac M2
 - IDE: VS Code
 - Language: Python
 - Source: pip package version 23.3.2 semantic kernel 0.5.0.dev0
 - API: Azure OpenAI API GPT-4

**Additional context**
I think the reason is that Azure OpenAI's Chat Completion does not include logprobs.
For example, the following curl will not return logprobs

```curl
curl "https://*******.openai.azure.com/openai/deployments/gpt-4/chat/completions?api-version=2023-12-01-preview" \
  -H "Content-Type: application/json" \
  -H "api-key: **************" \
  -d "{
  \"messages\": [{\"role\":\"system\",\"content\":\"You are an AI assistant that helps people find information.\"},{\"role\":\"user\",\"content\":\"hello\"},{\"role\":\"assistant\",\"content\":\"Hello! How can I help you today? If you have any questions or need assistance, feel free to ask.\"}],
  \"max_tokens\": 800,
  \"temperature\": 0.7,
  \"frequency_penalty\": 0,
[...]

PR #5071, opened 2024-02-19
Python: Support for cases where "logprobs" does not exist（OpenAITextCompletionBase）

ymuichiro:
### Motivation and Context

An error has occurred. This is the same issue as the following: Azure OpenAI API may not include logprobs.

The recent change only modified OpenAIChatCompletionBase. However, it turns out that this issue also occurs with OpenAITextCompletionBase. Therefore, I propose to make a similar fix.

https://github.com/microsoft/semantic-kernel/issues/4923

Non-existent property lobprobs is referenced when using Azure OpenAI Text Completion.

When using Azure OpenAI TextCompletion

https://github.com/microsoft/semantic-kernel/issues/5070

### Description

Azure OpenAI ChatCompletion does not return logprobs.(It seems that there are cases where it is returned, but I have not encountered that case yet.)
Thus, the following code would be "AttributeError: 'Choice' object has no attribute 'logprobs'".

`class OpenAITextCompletionBase`

```python
    def _get_metadata_from_text_choice(self, choice: CompletionChoice) -> Dict[str, Any]:
        """Get metadata from a completion choice."""
        return {
            "logprobs": choice.logprobs,
        }
```

do the following

```python
    def _get_metadata_from_text_choice(self, choice: CompletionChoice) -> Dict[str, Any]:
        """Get metadata from a completion choice."""
        return {
            "logprobs": getattr(choice, "logprobs", None),
        }
```

### Contribution Checklist

- [-] The code builds clean without any errors or warnings
- [-] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [-] All unit tests pass, and I have added new tests where possible
- [-] I didn't break anyone :smile:

### DIFF
--- a/python/semantic_kernel/connectors/ai/open_ai/services/open_ai_text_completion_base.py
+++ b/python/semantic_kernel/connectors/ai/open_ai/services/open_ai_text_completion_base.py
@@ -144,5 +144,5 @@ class OpenAITextCompletionBase(OpenAIHandler, TextCompletionClientBase):
     def _get_metadata_from_text_choice(self, choice: CompletionChoice) -> Dict[str, Any]:
         """Get metadata from a completion choice."""
         return {
-            "logprobs": choice.logprobs,
+            "logprobs": getattr(choice, "logprobs", None),
         }


eavanvalkenburg (approved):
Thanks for adding this!

juliomenendez (approved):

PR #5074, opened 2024-02-19
.Net: Added Stream property to AudioContent model

dmytrostruk:
### Motivation and Context

Added `Stream` property to `AudioContent` model to support not only `BinaryData` as input. This change will allow to use `Stream` for new audio connectors as more efficient way to pass big amount of data to connector.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
@@ -246,7 +246,9 @@ internal abstract class ClientCore
         PromptExecutionSettings? executionSettings,
         CancellationToken cancellationToken)
     {
-        Verify.NotNull(content.Data);
+        var audioData = await this.GetBinaryDataFromAudioContentAsync(content, cancellationToken).ConfigureAwait(false);
+
+        Verify.NotNull(audioData, nameof(content));
 
         OpenAIAudioToTextExecutionSettings? audioExecutionSettings = OpenAIAudioToTextExecutionSettings.FromExecutionSettings(executionSettings);
 
@@ -254,7 +256,7 @@ internal abstract class ClientCore
 
         var audioOptions = new AudioTranscriptionOptions
         {
-            AudioData = content.Data,
+            AudioData = audioData,
             DeploymentName = this.DeploymentOrModelName,
             Filename = audioExecutionSettings.Filename,
             Language = audioExecutionSettings.Language,
@@ -1058,5 +1060,19 @@ internal abstract class ClientCore
         s_promptTokensCounter.Add(usage.PromptTokens);
         s_completionTokensCounter.Add(usage.CompletionTokens);
         s_totalTokensCounter.Add(usage.TotalTokens);
+    }
+
+    private async Task<BinaryData?> GetBinaryDataFromAudioContentAsync(
+        AudioContent audioContent,
+        CancellationToken cancellationToken)
+    {
+        BinaryData? data = audioContent?.Data;
+
+        if (data is null && audioContent?.Stream is not null)
+        {
+            data = await BinaryData.FromStreamAsync(audioContent.Stream, cancellationToken).ConfigureAwait(false);
+        }
+
+        return data;
     }
 }
--- a/dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
 
 namespace Microsoft.SemanticKernel.Contents;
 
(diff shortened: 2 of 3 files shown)


dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
```diff
@@ -15,8 +16,21 @@ public class AudioContent : KernelContent
...
     public BinaryData? Data { get; set; }
 
+    /// <summary>
+    /// The audio stream.
+    /// </summary>
+    /// <remarks>
+    /// If both <see cref="Data"/> and <see cref="Stream"/> are provided,
+    /// <see cref="Data"/> takes precedence over <see cref="Stream"/>.
+    /// </remarks>
+    public Stream? Stream { get; set; }
```
RogerBarreto, line 32: We do have the Streaming classes to represent streaming content coming back from the Model, why do we need to expose the Stream as a new property, can't we get it from the BinaryData already or for those purposes points to the Streaming Endpoint?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1057,4 +1059,18 @@ private void CaptureUsageDetails(CompletionsUsage usage)
...
         s_totalTokensCounter.Add(usage.TotalTokens);
     }
+
+    private async Task<BinaryData?> GetBinaryDataFromAudioContentAsync(
+        AudioContent audioContent,
+        CancellationToken cancellationToken)
+    {
+        BinaryData? data = audioContent?.Data;
+
+        if (data is null && audioContent?.Stream is not null)
```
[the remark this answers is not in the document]
dmytrostruk: @SergeyMenshykh it's warning in this case.
image

Most probably it doesn't know whether `data` is null because `audioContent` is null or because `audioContent.Data` is null, so it requires additional check on compile-time.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
dmytrostruk: > We do have the Streaming classes to represent streaming content coming back from the Model,

@RogerBarreto As far as I understand, StreamingContent classes are used for `IAsyncEnumerable` scenarios and they represent content chunk, I guess it's a little bit different from audio scenario. 

> why do we need to expose the Stream as a new property, can't we get it from the BinaryData already

We can, but `Stream` through `BinaryData` will load data into memory, which is not very efficient in case of big files, while using `Stream` directly will allow to pass data to connector without loading data into memory. OpenAI connector accepts `BinaryData` as for now, but `Stream` should be helpful for other audio connectors (see comment: https://github.com/microsoft/semantic-kernel/pull/4932#issuecomment-1951728120)

> or for those purposes points to the Streaming Endpoint?

OpenAI doesn't have Streaming Endpoint for audio scenarios (at least for now). In case if it will be supported in the future, maybe we could re-use Streaming classes in this case, it will depend on audio streaming API shape. 

Let me know if that makes sense.

dmytrostruk: > Having URi, Data, and now Stream to this class seems that we are adding excessive responsibility to the abstraction...

We don't have URI at the moment, but we may possibly add it in the future. Same as in `ImageContent` scenario, it's just the wayof supporting different data types per content type. Another way would be to replace `AudioContent` as input parameter in `AudioToText` interface with multiple overloads that accept BinaryData, Stream and other possible types.

RogerBarreto: I see, thanks for the detail on this one. Not a super fan of adding multiple mutually exclusive properties in the abstraction level, without specialization.

Maybe having a `Experimental` would be advisable for now until we have a good strategy for those scenarios.

Krzysztof318: @dmytrostruk I wonder about Roger's words. To sending audio data using AudioContent.Stream is ok but let's take as an example a connector to wshiper live. Such a connector would return an AudioContent with a continuous stream. Maybe in such a case an additional class like StreamingAudioContent with property stream would actually be better?

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
dmytrostruk: > Maybe having a `Experimental` would be advisable for now until we have a good strategy for those scenarios.

@RogerBarreto Thanks! We already have `Experimental` flag on entire `AudioContent` class and all audio-related functionality, so we can change/improve it in the future.

> Maybe in such a case an additional class like StreamingAudioContent with property stream would actually be better?

@Krzysztof318 I didn't work with whisper live model yet and it's hard to predict what abstraction it will require - maybe another `Content` class, maybe new method in `AudioToText` interface to support `IAsyncEnumerable` or maybe completely separate interface. 

My recommendation would be to add/update the abstraction only when needed. When we implement whisper live connector, we will see which approach will cover both scenarios in the best possible way. Until then, with current functionality we should be fine. Same with this PR and `Stream` property - if we think it should be useful for other synchronous audio-to-text connectors - let's add it. If we are not sure it's going to be useful - we can close this PR for now and add it only when we see such use-case. Please let me know if that makes sense.

Krzysztof318: @dmystrostruk I am thinking about your opinion. And I think you are right, stream support is good idea but currently this will be unusless, until someone create whisper local and text to speech models local support. So if others are agree, we can close this now and back to this when we see new connector PR.

dmytrostruk: > @dmystrostruk I am thinking about your opinion. And I think you are right, stream support is good idea but currently this will be useless, until someone create whisper local and text to speech models local support. So if others are agree, we can close this now and back to this when we see new connector PR.

Even with whisper local and text to speech models local support, it's still possible to create such connectors with `BinaryData` right now, because if you have `Stream`, you can do `BinaryData.FromStreamAsync` on your side. I think having `Stream` as separate property may be useful for large audio data, but again, it's something that needs to be supported by AI model first, before modifying the abstraction. Having `Experimental` flag for some period of time will allow us to do modifications when needed. So, I'm going to close this PR for now and let's see how it goes. 

Thank you for feedback!

dmytrostruk:
Closing for now and we will update the abstraction when new audio connectors will be added, which require other data type than `BinaryData`.

Issue #4947, opened 2024-02-09
.Net: Update Hugging Face connector


markwallace-microsoft:
SK Hugging Face package should support the following:

1. TextGeneration
2. TextEmbedding
3. TextToImage

The connector implementation should include a generic client implementation and allow for the injection of custom HttpClients.

PR #5075, opened 2024-02-19
.Net Hugging Face HttpClient Connector Update

RogerBarreto:
### Motivation and Context

Resolves Partially #4947 

Update Hugging Face connector segregating the responsibilities of the TextGenerationService, TextEmbeddingService, HttpClient, JsonParsing, EndPointResolving.

### Description

This enables paving the way to support additional Hugging Face modalities


dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFacePromptExecutionSettings.cs
```diff
@@ -0,0 +1,203 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
Krzysztof318, line 1: Don't forget imlement freeze and clone override 😀

Krzysztof318 (changes requested):

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFacePromptExecutionSettings.cs
```diff
@@ -0,0 +1,223 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.SemanticKernel.Text;
+
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
```
stephentoub, line 9: Can we remove the ".Core" in the namespace?

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFacePromptExecutionSettings.cs
```diff
@@ -0,0 +1,223 @@
...
+        {
+            case null:
+                return new HuggingFacePromptExecutionSettings() { MaxTokens = DefaultTextMaxTokens };
+            case HuggingFacePromptExecutionSettings settings:
+                return settings;
+        }
+
+        var json = JsonSerializer.Serialize(executionSettings);
+        var huggingFacePromptExecutionSettings = JsonSerializer.Deserialize<HuggingFacePromptExecutionSettings>(json, JsonOptionsCache.ReadPermissive);
+        if (huggingFacePromptExecutionSettings is not null)
```
stephentoub, line 38: When would this ever be null?

dotnet/src/Connectors/Connectors.HuggingFace/Core/IStreamJsonParser.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
+
+/// <summary>
+/// Represents a JSON parser that can parse a Stream containing JSON data and yield the individual JSON objects.
+/// </summary>
+internal interface IStreamJsonParser
```
stephentoub, line 11: I guess I should wait until it's no longer WIP, but I'm curious as to the purpose of all these newly defined internal interfaces

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFacePromptExecutionSettings.cs
RogerBarreto: Moved, good catch.

dotnet/src/Connectors/Connectors.HuggingFace/Core/IStreamJsonParser.cs
RogerBarreto: This is an internal SSE streaming json parser for each object that contains a valid Json signature.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFacePromptExecutionSettings.cs
stephentoub: Deserialize should only ever return null if the input json were "null", and that won't ever be the case here, right? (Technically it could also return null if a misbehaving converter returned null eroneously, but we control all the converters here and none are doing that.)

RogerBarreto: From this perspective I removed the check and enforced `!`. Thanks.

dotnet/src/Connectors/Connectors.HuggingFace.UnitTests/MultipleHttpMessageHandlerStub.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
markwallace-microsoft, line 1: Do we need `HttpMessageHandlerStub` given we have this class?

dotnet/src/Connectors/Connectors.HuggingFace/Connectors.HuggingFace.csproj
```diff
@@ -18,6 +18,14 @@
     <Description>Semantic Kernel connectors for Hugging Face. Contains clients for text generation and text embedding generation.</Description>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="TextGeneration\HuggingFaceTextGenerationServiceOld.cs" />
+  </ItemGroup>
```
markwallace-microsoft, line 23: Is this needed?

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceEndpointProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// </remarks>
+    public HuggingFaceEndpointProvider() : this(null)
+    {
+    }
+
+    public Uri GetTextGenerationEndpoint(string modelId)
+        => new($"{this._endPoint}{this._separator}models/{modelId}");
+
+    public Uri GetEmbeddingGenerationEndpoint(string modelId)
+        => new($"{this._endPoint}{this._separator}pipeline/feature-extraction/{modelId}");
```
SergeyMenshykh, line 41: Can't we simply have the methods in the HuggingFaceClient class, use "https://api-inference.huggingface.co" as the default endpoint and override it with the value of BaseAddress, if it's set, in the provided HTTP client or the value of the endpoint parameter? This way, we can easily simplify the design by eliminating the need for the additional class.

dotnet/src/Connectors/Connectors.HuggingFace/Core/IEndpointProvider.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
+
+internal interface IEndpointProvider
```
markwallace-microsoft, line 7: Do we expect to have multiple implementations of this interface?

Given this is all internal maybe we can simplify things.

dotnet/src/Connectors/Connectors.HuggingFace/Core/IHttpRequestFactory.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using System.Net.Http;
+
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
+
+/// <summary>
+/// Represents a factory for creating HttpRequestMessage instances.
+/// </summary>
+internal interface IHttpRequestFactory
```
markwallace-microsoft, line 11: Same comment as above, are we expecting multiple implementations?

dotnet/src/Connectors/Connectors.HuggingFace/HuggingFaceKernelBuilderExtensions.cs
```diff
@@ -20,23 +21,23 @@ public static class HuggingFaceKernelBuilderExtensions
     /// <param name="builder">The <see cref="IKernelBuilder"/> instance to augment.</param>
     /// <param name="model">The name of the Hugging Face model.</param>
     /// <param name="apiKey">The API key required for accessing the Hugging Face service.</param>
-    /// <param name="endpoint">The endpoint URL for the text generation service.</param>
+    /// <param name="endPoint">The endpoint URL for the text generation service.</param>
```
markwallace-microsoft, line 24: `endpoint` is a single word so no need to change this

dotnet/src/Connectors/Connectors.HuggingFace/TextEmbedding/HuggingFaceTextEmbeddingGenerationService.cs
```diff
@@ -16,137 +16,46 @@ namespace Microsoft.SemanticKernel.Connectors.HuggingFace;
...
-        this._model = model;
-        this._endpoint = endpoint;
-        this._httpClient = httpClient;
-        this._attributes.Add(AIServiceExtensions.ModelIdKey, model);
-        this._attributes.Add(AIServiceExtensions.EndpointKey, endpoint ?? httpClient.BaseAddress!.ToString());
+        this.Client = new HuggingFaceClient(
+            modelId: model,
+            apiKey: apiKey,
+            httpRequestFactory: new HuggingFaceHttpRequestFactory(),
+#pragma warning disable CA2000
```
markwallace-microsoft, line 48: Useful to include the description of the warning

`#pragma warning disable CA2000 // Dispose objects before losing scope`

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceHttpRequestFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System;
+using System.Net.Http;
+using Microsoft.SemanticKernel.Http;
+
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
+
+internal sealed class HuggingFaceHttpRequestFactory : IHttpRequestFactory
+{
+    public HttpRequestMessage CreatePost(object requestData, Uri endpoint, string? apiKey)
```
SergeyMenshykh, line 11: 1. Do we foresee a necessity to swap the HTTP request factories? What would be the scenarios for that?  
2. Why are we externalizing the request creation methods into a separate class instead of having them in the HuggingFaceClient? Do we plan to reuse the HuggingFaceHttpRequestFactory with different classes?

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/HuggingFaceTextGenerationServiceOld.cs
```diff
@@ -0,0 +1,153 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
markwallace-microsoft, line 1: Can we remove this?

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        this.LogTextGenerationUsage(executionSettings);
+
+        return textContents;
+    }
+
+    public async IAsyncEnumerable<StreamingTextContent> StreamGenerateTextAsync(
+        string prompt,
+        PromptExecutionSettings? executionSettings = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
```
SergeyMenshykh, line 71: Considering that this class is an internal one, I would suggest always requiring the cancellation token argument to avoid forgetting to pass it from the calling method. Same for the GenerateEmbeddingsAsync method.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        foreach (var streamingTextContent in this.ProcessTextResponseStream(responseStream, executionSettings?.ModelId ?? this._modelId))
+        {
+            yield return streamingTextContent;
+        }
+    }
+
+    public async Task<IList<ReadOnlyMemory<float>>> GenerateEmbeddingsAsync(
+        IList<string> data,
+        Kernel? kernel = null,
```
SergeyMenshykh, line 93: 1. This method accepts kernel others don't. We need to be consistent.
2. Don't we need to be consistent and get modelId from execution settings as other public methods of the class do?

dotnet/src/Connectors/Connectors.HuggingFace.UnitTests/Connectors.HuggingFace.UnitTests.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
RogerBarreto: Yes, this is a new UnitTest project per Connector, is the new pattern I'm following, as this also avoid conflicts/clashes induced by the InternalUtilities being reused across different connectors.

dotnet/src/Connectors/Connectors.HuggingFace/Connectors.HuggingFace.csproj
RogerBarreto: No, left over, will remove.

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/TextGenerationResponse.cs
```diff
@@ -1,17 +1,21 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+using System.Collections.Generic;
 using System.Text.Json.Serialization;
+using static Microsoft.SemanticKernel.Connectors.HuggingFace.Core.TextGenerationResponse;
 
-namespace Microsoft.SemanticKernel.Connectors.HuggingFace;
+#pragma warning disable CA1812
```
markwallace-microsoft, line 7: Warning description?

dotnet/src/IntegrationTests/Connectors/HuggingFace/TextGeneration/HuggingFaceTextGenerationTests.cs
```diff
@@ -36,7 +36,7 @@ public async Task HuggingFaceLocalAndRemoteTextGenerationAsync()
         // Arrange
         const string Input = "This is test";
 
-        var huggingFaceLocal = new HuggingFaceTextGenerationService(Model, endpoint: Endpoint);
+        var huggingFaceLocal = new HuggingFaceTextGenerationService(Model, endPoint: new Uri(Endpoint));
```
markwallace-microsoft, line 39: Should be `endpoint`

dotnet/src/Connectors/Connectors.HuggingFace/Core/IHuggingFaceClient.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// Streams the generated text content asynchronously.
+    /// </summary>
+    /// <param name="prompt">The prompt for generating text content.</param>
+    /// <param name="executionSettings">The prompt execution settings (optional).</param>
+    /// <param name="cancellationToken">The cancellation token (optional).</param>
+    /// <returns>An asynchronous enumerable of <see cref="StreamingTextContent"/> streaming text contents.</returns>
+    IAsyncEnumerable<StreamingTextContent> StreamGenerateTextAsync(
+        string prompt,
+        PromptExecutionSettings? executionSettings = null,
+        CancellationToken cancellationToken = default);
```
SergeyMenshykh, line 34: similar comment about having default value for cancellation tokens for public methods of the internal implementation class.

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/TextGenerationStreamResponse.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    internal sealed class TextGenerationToken
+    {
+        [JsonPropertyName("id")]
+        public int Id { get; set; }
+
+        [JsonPropertyName("text")]
+        public string? Text { get; set; }
+
+        [JsonPropertyName("logprob")]
```
markwallace-microsoft, line 31: Odd this isn't `log_prob`, cannot find the reference for this could you add it as a link?

Would be useful to do this for various request and response models.

dotnet/src/Connectors/Connectors.HuggingFace.UnitTests/Connectors.HuggingFace.UnitTests.sln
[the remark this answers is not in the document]
RogerBarreto: I just found that this is a solution file, indeed no need.

dotnet/src/Connectors/Connectors.HuggingFace.UnitTests/Connectors.HuggingFace.UnitTests.sln
```diff
@@ -0,0 +1,25 @@
...
+		{A474D223-BC8A-475A-8F23-B2BDD60B47AE}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{A474D223-BC8A-475A-8F23-B2BDD60B47AE}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {35102C5F-500C-41C7-8836-5C44DD59B98B}
+	EndGlobalSection
+EndGlobal
```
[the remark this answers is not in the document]
RogerBarreto: Is not, removing

dotnet/src/Connectors/Connectors.HuggingFace.UnitTests/MultipleHttpMessageHandlerStub.cs
RogerBarreto: This class is duplicated for each UnitTest, I just moved it to InternalUtilities and removed it from the other projects.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -0,0 +1,209 @@
...
+{
+    private readonly IStreamJsonParser _streamJsonParser;
+    private readonly string _modelId;
+    private readonly string? _apiKey;
+
+    private IHttpRequestFactory HttpRequestFactory { get; }
+    private IEndpointProvider EndpointProvider { get; }
+    private HttpClient HttpClient { get; }
+    private ILogger Logger { get; }
+
```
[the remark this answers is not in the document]
RogerBarreto: Not really as this is a sealed class, we can have them all as fields.

dotnet/SK-dotnet.sln
```diff
@@ -87,6 +87,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{5C246969-D794-4EC3-8E8F-F90D4D166420}"
 	ProjectSection(SolutionItems) = preProject
 		src\InternalUtilities\test\AssertExtensions.cs = src\InternalUtilities\test\AssertExtensions.cs
+		HttpMessageHandlerStub.cs = HttpMessageHandlerStub.cs
 		src\InternalUtilities\test\TestInternalUtilities.props = src\InternalUtilities\test\TestInternalUtilities.props
```
Krzysztof318, line 91:
```suggestion
		src\InternalUtilities\test\AssertExtensions.cs = src\InternalUtilities\test\AssertExtensions.cs
		src\InternalUtilities\test\TestInternalUtilities.props = src\InternalUtilities\test\TestInternalUtilities.props
		src\InternalUtilities\test\HttpMessageHandlerStub.cs = src\InternalUtilities\test\HttpMessageHandlerStub.cs
```
From gemini, change this please to be same.

dotnet/src/InternalUtilities/test/HttpMessageHandlerStub.cs
```diff
@@ -8,7 +8,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace SemanticKernel.Connectors.UnitTests;
+namespace SemanticKernel.UnitTests;
```
Krzysztof318, line 11:
```suggestion
```
Proposal; remove namespace completely to not need using in tests classes like in gemini branch.

Krzysztof318 (changes requested):

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
RogerBarreto: I think the consistency is a good point here, this method is following what we have for the Memory Interface abstraction `IEmbeddingGenerationService`, any changes to this will be breaking changes moving forward or we can think about introducing a new Interface with `PromptExecutionSettings` as a parameter, removing the Kernel or thinking on a better strategy for this.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceEndpointProvider.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
+
+internal sealed class HuggingFaceEndpointProvider : IEndpointProvider
```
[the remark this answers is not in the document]
RogerBarreto: No initially, I can move those to be concretes as everything is internal.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceEndpointProvider.cs
RogerBarreto: As the endpoint is a `Url` now, I cant use `new` as the default for the constructor, had to do this way. Any suggestion?

RogerBarreto: EndPointProvider is a valid responsibility segregation, I can move this responsibility to the Client, sure, but I liked this approach from the Gemini impl which I used in Ollama's too.

dotnet/src/Connectors/Connectors.HuggingFace/Core/IHttpRequestFactory.cs
RogerBarreto: Those itens are valid for IOC into the Client, so for testing purposes we are able to override the behavior easily of the client, but I understand that this can be too much for it, was repeating a valid pattern I found in Gemini, will remove all of it....

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceEndpointProvider.cs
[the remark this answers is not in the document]
RogerBarreto: Will remove those segregations...

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/TextGenerationStreamResponse.cs
RogerBarreto: Unfortunately there's not documentation describing the Streaming Chunk object in Hugging Face documentation. 

This is the result I have using postman.

image

dotnet/src/InternalUtilities/test/HttpMessageHandlerStub.cs
RogerBarreto: I think is a valid point we may try this in a different PR as the current changes are already adding 20 unrelated files to the pull request.

SergeyMenshykh (approved):

Krzysztof318:
@RogerBarreto You need so big test data file embeddings_test_response_feature_extraction.json?

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -0,0 +1,236 @@
...
+    {
+        var endpoint = this.GetTextGenerationEndpoint(executionSettings?.ModelId ?? this._modelId);
+        var request = this.CreateTextRequest(prompt, executionSettings);
+        using var httpRequestMessage = this.CreatePost(request, endpoint, this._apiKey);
+
+        string body = await this.SendRequestAndGetStringBodyAsync(httpRequestMessage, cancellationToken)
+            .ConfigureAwait(false);
+
+        var response = DeserializeResponse<TextGenerationResponse>(body);
+        var textContents = GetTextContentFromResponse(response, executionSettings?.ModelId ?? this._modelId);
```
Krzysztof318, line 63: Small improvement;
```suggestion
        string modelId = executionSettings?.ModelId ?? this._modelId;
        var endpoint = this.GetTextGenerationEndpoint(modelId);
        var request = this.CreateTextRequest(prompt, executionSettings);
        using var httpRequestMessage = this.CreatePost(request, endpoint, this._apiKey);

        string body = await this.SendRequestAndGetStringBodyAsync(httpRequestMessage, cancellationToken)
            .ConfigureAwait(false);

        var response = DeserializeResponse<TextGenerationResponse>(body);
        var textContents = GetTextContentFromResponse(response, modelId);
```

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        using var httpRequestMessage = this.CreatePost(request, endpoint, this._apiKey);
+
+        using var response = await this.SendRequestAndGetResponseImmediatelyAfterHeadersReadAsync(httpRequestMessage, cancellationToken)
+            .ConfigureAwait(false);
+
+        using var responseStream = await response.Content.ReadAsStreamAndTranslateExceptionAsync()
+            .ConfigureAwait(false);
+
+        foreach (var streamingTextContent in this.ProcessTextResponseStream(responseStream, executionSettings?.ModelId ?? this._modelId))
```
Krzysztof318, line 87:
```suggestion
        string modelId = executionSettings?.ModelId ?? this._modelId;
        var endpoint = this.GetTextGenerationEndpoint(modelId);
        var request = this.CreateTextRequest(prompt, executionSettings);
        request.Stream = true;

        using var httpRequestMessage = this.CreatePost(request, endpoint, this._apiKey);

        using var response = await this.SendRequestAndGetResponseImmediatelyAfterHeadersReadAsync(httpRequestMessage, cancellationToken)
            .ConfigureAwait(false);

        using var responseStream = await response.Content.ReadAsStreamAndTranslateExceptionAsync()
            .ConfigureAwait(false);

        foreach (var streamingTextContent in this.ProcessTextResponseStream(responseStream, modelId))
```

Krzysztof318 (approved):
I added small suggest for passing modelId

RogerBarreto:
> You need so big test data file embeddings_test_response_feature_extraction.json?
Good point, didnt realize it was so big, updated to a smaller sample.

dotnet/samples/KernelSyntaxExamples/README.md
```diff
@@ -59,6 +59,7 @@ dotnet user-secrets set "KeyVault:TenantId" "..."
 
 dotnet user-secrets set "HuggingFace:ApiKey" "..."
 dotnet user-secrets set "HuggingFace:ModelId" "..."
+dotnet user-secrets set "HuggingFace:EmbeddingModelId" "..."
```
markwallace-microsoft, line 62: Given there are so many HF models and discovering the ones that support embeddings is not obvious can you provide a default here?

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/TextGenerationRequest.cs
```diff
@@ -1,23 +1,152 @@
...
 /// <summary>
-/// HTTP schema to perform completion request.
+/// HuggingFace text generation request object.
 /// </summary>
-public sealed class TextGenerationRequest
+internal sealed class TextGenerationRequest
 {
     /// <summary>
-    /// Prompt to complete.
+    /// A string to be generated from.
```
markwallace-microsoft, line 14: `The input string to generate embeddings for.`

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/TextGenerationStreamJsonParser.cs
```diff
@@ -0,0 +1,138 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Text.Json.Nodes;
+using Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
+
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.TextGeneration;
+
+internal sealed class TextGenerationStreamJsonParser : IStreamJsonParser
```
markwallace-microsoft, line 12: Can we do something like this if we are validating as json and otherwise just read as a string?

```
var serializer = new JsonSerializer();
using (var sr = new StreamReader(responseMessage))
using (var jsonTextReader = new JsonTextReader(sr))
{
    var jsObj= serializer.Deserialize(jsonTextReader);
}
```

What is the unit test coverage for this class? If we need this we want to ensure it is high and also we have negative test cases e.g. badly formed json with validation turned on.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -0,0 +1,238 @@
...
+
+    internal HuggingFaceClient(
+        string modelId,
+        HttpClient httpClient,
+        Uri? endpoint = null,
+        string? apiKey = null,
+        IStreamJsonParser? streamJsonParser = null,
+        ILogger? logger = null)
+    {
+        Verify.NotNullOrWhiteSpace(modelId);
```
markwallace-microsoft, line 38: Should we also verify the `httpClient`

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/TextGenerationStreamJsonParser.cs
RogerBarreto: Added the unit tests for this class, used the same tests from Gemini, quite useful class BTW, it reads the stream and split the different json blocks as they arrive {} {} {}

dotnet/src/Connectors/Connectors.HuggingFace/TextGeneration/TextGenerationRequest.cs
RogerBarreto: This is not an embedding request, the description I took from their documentation:

image

Will update to your suggestion.

markwallace-microsoft (approved):

markwallace-microsoft (approved):

PR #5088, opened 2024-02-20
.Net Single source of truth for chat message content

SergeyMenshykh:
Today, the ChatMessageContent class has two sources of truth for its content - the Content property and the Items property. This may be acceptable for now, as all SK and industry chat completion services follow the same protocol. They use the Content property for system, user, assistant, and tool messages and alternatively allow passing image and text content via the Items collection for user messages only.  
   
However, this might not be suitable when there's a chat completion service that doesn't follow the protocol mentioned above. For example, it could be a new advanced chat completion service with multimodal support for assistant messages. In this case, consumer code working through the IChatCompletionService interface won't be able to handle content for assistant messages polymorphically, and all consumers that need to work with both current and new chat completion services will have to use code like this:  
   
```C#  
var message = await chatCompletionService.GetChatMessageContentAsync(...);  
    
if (message.Content != null)  
{  
    // Handle content specified in the Content property of the assistance message  
}  
else if (message.Items is { Count: > 0 } items)  
{  
    // Handle content specified in the Items/items property of the assistance message  
}  
    
// or check the Items property first and then Content one?  
```

The problem becomes more apparent and manifests itself immediately in the agent's space. Each agent needs to have logic like the one above to identify the source of the content and map it to an internal API data model. For example, here's the code for a ChatCompletion agent that we would need to write  
   
```C#  
async Task<ChatMessageContent[]> InvokeAsync(ChatMessageContent[] messages)
{
    var chat = new ChatHistory();

    foreach (var message in messages)
    {
        if (message.Role == AuthorRole.User)
        {
            // User messages can have content in either 'Items' property or the 'Content' property.
            // Assuming one of the two properties has content, adding the message to the chat and continue.
            chat.Add(message);
            continue;
        }

        // The system, assistant and tool messages are expected to have content in the 'Content' property.
        // This expectation is specific to OpenAI chat completion services and may not be relevant to other chat   
        // completion service types, e.g., multimodality for assistant messages, where content would be expected   
        // to be provided via the 'Items' collection.  
        if (!string.IsNullOrEmpty(message.Content))
        {
            chat.Add(message);
            continue;
        }

        // Doing our best to identify content for the message and add it to the chat.
        if (message.Items is { Count: > 0 } items && items[0] is TextContent textContent)
        {
            // The problem with the clone code below is that we loose the original message type that could have been interpreted by
            // underlying API differently than the type of the clone - ChatMessageContent
            var clone = new ChatMessageContent(
                role: message.Role,
                content: textContent.Text,
                modelId: message.ModelId,
                innerContent: message.InnerContent,
                encoding: message.Encoding,
                metadata: message.Metadata
            );

            chat.Add(clone);
            continue;
        }

        // If we get here it means that all the above conditions failed, so we add the message 
        // to let underlying API handle it.
        chat.Add(message);
    }

    var chatMessageContent = await chatCompletionService.GetChatMessageContentsAsync(chat, ...);

    return chatMessageContent.Select(m => { m.Source = this; return m; }).ToArray();
}
```  

To avoid all the unnecessary 'if/else' mapping logic needed to identify the source of content depending on the message role, it would be beneficial to have only one source of content - Items. Ideally, the 'Content' property should be removed, but doing so would unnecessarily break a lot of consumer code.

As a middle-ground solution, this PR changes the purpose of the 'Content' property from being a separate source of content to a shortcut for the first item of text content type in the 'Items' collection. This way, the 'Content' property will be nothing more than just a convenient method to add, update, or return the text of the first item of text content type. The 'Items' collection, on the other hand, becomes the only source of content that can be used polymorphically by consumer code.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -18,20 +22,94 @@ public class ChatMessageContent : KernelContent
...
+    public string? Content
+    {
+        get
+        {
+            if (this.Items.Count == 0)
+            {
+                return null;
+            }
+
+            if (this.Items[0] is TextContent textContent)
```
Krzysztof318, line 37: I think this should be changed to FirstOrDefault or SingleOrDefault. What if first will be image and second TextContent? Then this property will fail.

SergeyMenshykh: Thanks @Krzysztof318. I received the same feedback during the design review, and it makes sense. I am more inclined to change the code to use SingleOrDefault, as it will prevent cases in which the getter returns different content from the one initially provided through the setter when there is more than one text content item in the Items collection.

stephentoub: If it's going to search through all the items for one that's text, it seems a bit arbitrary to then fail if there are multiple that are text, no?

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -18,20 +22,94 @@ public class ChatMessageContent : KernelContent
...
-    public ChatMessageContentItemCollection? Items { get; set; }
+    public ChatMessageContentItemCollection Items
+    {
+        get
+        {
+            return this._items ??
+                Interlocked.CompareExchange(ref this._items, new ChatMessageContentItemCollection(), null) ??
+                this._items;
+        }
+        set { this._items = value ?? new ChatMessageContentItemCollection(); }
```
stephentoub, line 85: Why is the `?? new ChatMessageContentItemCollection();` needed?

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -18,20 +22,94 @@ public class ChatMessageContent : KernelContent
...
     /// <summary>
     /// Chat message content items
     /// </summary>
-    public ChatMessageContentItemCollection? Items { get; set; }
+    public ChatMessageContentItemCollection Items
+    {
+        get
+        {
+            return this._items ??
+                Interlocked.CompareExchange(ref this._items, new ChatMessageContentItemCollection(), null) ??
```
stephentoub, line 82: Is this type intended to be read from concurrently? I would expect no, in which case this can just be `get => this._items ??= new ChatMessageContentItemCollection();`

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
SergeyMenshykh: fixed.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
SergeyMenshykh: fixed.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
SergeyMenshykh: Right, alternatively, we can always set and get the content of the first text item, regardless of its position and the number of text items in the Items collection. This way, our solution is more flexible, as the 'Content' property shortcut continues to work and does not explode if there are more than one item in the collection.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -76,13 +154,25 @@ public class ChatMessageContent : KernelContent
...
+#pragma warning disable CA1031 // Do not catch general exception types
+        try
+        {
+            return this.Content ?? string.Empty;
+        }
+        catch
+        {
+            return string.Empty;
+        }
+#pragma warning restore CA1031 // Do not catch general exception types
```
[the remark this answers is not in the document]
SergeyMenshykh: The change is no longer relevant and was necessary to prevent the ToString methods from throwing exceptions if the item with index 0 was not a text item. Since we moved away from the solution, I'm rolling back the change to the method.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -18,20 +21,84 @@ public class ChatMessageContent : KernelContent
     public AuthorRole Role { get; set; }
 
     /// <summary>
-    /// Content of the message
+    /// A convenience property to get or set the text of the first item in the <see cref="Items" /> collection of <see cref="TextContent"/> type.
     /// </summary>
-    public string? Content { get; set; }
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public string? Content
```
Krzysztof318, line 27: small suggest;
Mark Content property with Obsolete attribute.

SergeyMenshykh: I'm more inclined to keep the property around for now until we have strong reasons not to. Because on the one hand, it's logical to prevent unnecessary confusion and steer everyone to use the Items collection. On the other hand, there are several places that need updating either to use the Items instead or suppress the "obsolete" warning in SK, and it can be a painful breaking change for SK consumer code if we decide to remove it. Additionally, it's a convenient shortcut to work with text content item. What do you think?

Krzysztof318: I would agree with keeping the property if its operation was always predictable i.e. SingleOrDefault but due to the use of FirstOrDefault I am still in favor of marking it as obsolete.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -18,20 +21,84 @@ public class ChatMessageContent : KernelContent
...
+                textContent.Encoding = value;
+            }
+        }
+    }
+
+    /// <summary>
+    /// The source of the message that generated it.
+    /// </summary>
+    [Experimental("SKEXP0101")]
+    public object? Source { get; set; }
```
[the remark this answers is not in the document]
SergeyMenshykh: Excluded for now.

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
SergeyMenshykh: We plan to consider the change in scope of this item - https://github.com/microsoft/semantic-kernel/issues/5170

dotnet/samples/KernelSyntaxExamples/Example86_ChatHistorySerialization.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Examples;
 
-public class Example85_ChatHistorySerialization : BaseTest
```
greyaxie, line 14: can uou give me

greyaxie (commented):
can you give me the chat convertation

Issue #4913, opened 2024-02-07
.Net: IFunctionFilter does not provide the fully qualified name of the KernelFunction


Nurgo:
**Describe the bug**
`PluginName` is not set in `FunctionInvokingContext` and `FunctionInvokedContext`. Consequently, it is not possible to know the fully qualified name of the `KernelFunction`.

**Additional context**
The problem is that `IFunctionFilter` will often be used to tell the user when the chatbot is using a plugin, as it is done in ChatGPT.
For example, if the chatbot uses the `HttpPlugin` plugin, you need to be able to display "The agent is using the function HttpPlugin.GetAsync...", not "The agent is using the function GetAsync...".
It is therefore important that `IFunctionFilter.OnFunctionInvoking()` and `IFunctionFilter.OnFunctionInvoked()` provide the plugin name, not just the function name.

PR #5091, opened 2024-02-20
.Net The 'Source' property added to the ChatMessageContent class

SergeyMenshykh:
### Motivation and Context  
For agent customization story, when converting or translating messages from one agent to another, we need a way to access the agent that generated the messages.  
   
### Description  
This PR adds the 'Source' property to the ChatMessageContent class.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -33,6 +33,11 @@ public class ChatMessageContent : KernelContent
     [JsonIgnore]
     public Encoding Encoding { get; set; }
 
+    /// <summary>
+    /// The source of the message that generated it.
+    /// </summary>
+    public object? Source { get; set; }
```
SergeyMenshykh, line 39: need to be mark as experimental

Krzysztof318:
Does we can use this property to solve plugin name problem described here https://github.com/microsoft/semantic-kernel/issues/4913?

SergeyMenshykh:
> Does we can use this property to solve plugin name problem described here #4913?

I don't see, at least for now, how the property can solve the problem. Chat completion services that use the ChatMessageContent class and SK functions are two unrelated SK components/functionality. Even if it might be possible to solve the problem for a particular edge case scenario using the property, it should not be the way forward. Instead, a proper solution should be put in place that covers the majority, if not all, scenarios.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -33,6 +34,12 @@ public class ChatMessageContent : KernelContent
     [JsonIgnore]
     public Encoding Encoding { get; set; }
 
+    /// <summary>
+    /// The source of the message that generated it.
+    /// </summary>
+    [Experimental("SKEXP0101")]
+    public object? Source { get; set; }
```
dmytrostruk, line 41: Unit test for serialization is failing because this property is not included in expected result:
image

Since it's going to be source instance which can be anything, should we put `[JsonIgnore]` attribute on this property?

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
SergeyMenshykh: excluded for now.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -33,6 +34,13 @@ public class ChatMessageContent : KernelContent
     [JsonIgnore]
     public Encoding Encoding { get; set; }
 
+    /// <summary>
+    /// The source of the message that generated it.
```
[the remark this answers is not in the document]
SergeyMenshykh: Done

PR #5093, opened 2024-02-20
.Net: Added example of Azure AI Search as Plugin with custom schema and configurable search fields

dmytrostruk:
### Motivation and Context

Today, Azure AI Search connector has two limitations:
1. It's working only with predefined schema, which is not preferred approach, because index in Azure AI Search is customizable and may use different schema from the one that exists in SK today.
2. Search operation uses only `Embedding` index field to perform search, while it's possible to specify multiple vector fields with different names.

In order to resolve these limitations, major refactoring on abstraction level is required, which will impact not only Azure AI Search connector, but all other memory connectors as well, and it will take some time.

This PR contains example that shows how to register Azure AI Search functionality in Kernel as Plugin and use it to perform search operations and communicate with AI based on your data. In this example, index schema is custom and can be changed when needed, as well as `searchFields` parameter is configurable via `KernelArguments`. The usage of Azure AI Search plugin is very similar to already existing approach with `AzureAISearchMemoryStore` connector.

### Contribution Checklist


SergeyMenshykh (approved):

markwallace-microsoft (approved):
What is the getting started story to run this sample?

One suggestion (and this would be a new task) 

There is this sample https://github.com/Azure/azure-search-vector-samples which allows you to set up AI Search and vectorize some data e.g. you could use https://github.com/Azure-Samples/azure-search-sample-data

It would be great if there is a readme which suggests a path for someone to go from zero to having this sample up and running. This could be a Blog article or a page in our developer documentation.

dmytrostruk:
> What is the getting started story to run this sample?
> 
> One suggestion (and this would be a new task)
> 
> There is this sample https://github.com/Azure/azure-search-vector-samples which allows you to set up AI Search and vectorize some data e.g. you could use https://github.com/Azure-Samples/azure-search-sample-data
> 
> It would be great if there is a readme which suggests a path for someone to go from zero to having this sample up and running. This could be a Blog article or a page in our developer documentation.

@markwallace-microsoft Great suggestion, thanks! I will think about ways how to provide all required information to run this sample as part of separate task.

PR #5094, opened 2024-02-20
.Net: Add AssemblyAI connector for Audio-to-text

Swimburger:
### Motivation and Context

AssemblyAI is a speech AI company and SK provides a new `IAudioToTextService` for different connectors to implement.

### Description

I added a connector for AssemblyAI that implements `IAudioToTextService`.
The `AssemblyAIAudioToTextService` accepts `AudioContent` as mandated by the interface, but also a `Stream` as discussed [here](https://github.com/microsoft/semantic-kernel/pull/4932), a `FileInfo`, and a `Uri`.

Remarks:
- I couldn't run tests in Connectors.UnitTests because of compilation issues in referenced projects, but I could verify the functionality in the integration tests.
- We're working on our C# SDK, so the code will be updated once our C# SDK lands.

Questions:
- How should I set up the icon and README.md file in this connector project?
- Should the `IAudioToTextService` interface accept `PromptExecutionSettings`? It seems a little odd to me since this isn't prompting an LLM.
- Which of these overloads makes sense to pull into `IAudioToTextService`? I added `Stream`, `FileInfo`, and `Uri` to our implementation.

### Contribution Checklist


dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        var transcriptId = await this.CreateTranscriptAsync(uploadUrl, executionSettings, cancellationToken)
+            .ConfigureAwait(false);
+        var transcript = await this.WaitForTranscriptToProcessAsync(transcriptId, cancellationToken)
+            .ConfigureAwait(false);
+
+        return new TextContent(
+            text: transcript.RootElement.GetProperty("text").GetString(),
+            modelId: null,
```
Krzysztof318, line 74: modelId is always null?

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// <summary>
+    /// Transcribe audio file.
+    /// </summary>
+    /// <param name="content">Audio content.</param>
+    /// <param name="executionSettings">The AI execution settings (optional).</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Text content from audio content.</returns>
```
Krzysztof318, line 54:
```suggestion
    /// <inheritdoc />
```

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+
+    /// <summary>
+    /// Transcribe audio file.
+    /// </summary>
+    /// <param name="fileStream">Stream of the audio file</param>
+    /// <param name="executionSettings">The AI execution settings (optional).</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns></returns>
+    public async Task<TextContent> GetTextContentAsync(
```
Krzysztof318, line 90: If you want to introduce an additional method, you need to introduce it also in abstraction i.e. in the ITextToAudioService. Look how it is done in this closed PR https://github.com/microsoft/semantic-kernel/pull/5074.
But I think there is a better way, instead adding a stream to the AudioContent, create a new class StreamingAudioContent which will be descendant from the [StreamingKernelContent](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Contents/StreamingKernelContent.cs). And next create overload in the ITextToAudioService with the StreamingAudioContent as parameter.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+
+    /// <summary>
+    /// Transcribe audio file.
+    /// </summary>
+    /// <param name="fileUrl">Public URL of the audio file to transcribe</param>
+    /// <param name="executionSettings">The AI execution settings (optional).</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns></returns>
+    public async Task<TextContent> GetTextContentAsync(
```
Krzysztof318, line 121: I think we don't want method which accept URI as param. Add URI to the existing AudioContent class.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+    /// <summary>
+    /// Transcribe audio file.
+    /// </summary>
+    /// <param name="file">Audio file to transcribe</param>
+    /// <param name="executionSettings">The AI execution settings (optional).</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns></returns>
+    public async Task<TextContent> GetTextContentAsync(
+        FileInfo file,
```
Krzysztof318, line 157: Same like with I said about URI...

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+            return;
+        }
+
+        if (response.Content.Headers.ContentType.MediaType == "application/json")
+        {
+            var jsonStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
+            var json = await JsonDocument.ParseAsync(jsonStream, cancellationToken: ct).ConfigureAwait(false);
+            if (json.RootElement.TryGetProperty("error", out var property))
+            {
+                throw new AssemblyAIApiException($"{errorMessagePrefix} Reason: {property.GetString()!}");
```
Krzysztof318, line 286: If it is sufficient throws any common .net exception if not use KernelException not custom.

Krzysztof318 (changes requested):
Hi, glad to see your new connector! I will be happy to help you by answering your questions.
I left a few comments regarding your code. I haven't reviewed all of it yet, so it's possible I'll add something else. By the way, I recommend you go to [Gemini branch](https://github.com/microsoft/semantic-kernel/tree/feature-connectors-gemini) and then to dotnet/src/connectors/Gemini.Connector. I've built a nice abstraction for gemini that you can use at your place without writing the code from scratch.

Krzysztof318:
> Should the IAudioToTextService interface accept PromptExecutionSettings? It seems a little odd to me since this isn't prompting an LLM.

Yes, but you don't need create class derived from PromptExecutionSettings

> Which of these overloads makes sense to pull into IAudioToTextService? I added Stream, FileInfo, and Uri to our implementation.

I have answered in review.

> I couldn't run tests in Connectors.UnitTests because of compilation issues in referenced projects, but I could verify the functionality in the integration tests.

Could you send a log?

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+    {
+        const string URL = "https://api.assemblyai.com/v2/upload";
+        using var content = new StreamContent(audioStream);
+        content.Headers.ContentType = new MediaTypeHeaderValue("application/octet-stream");
+        using var request = new HttpRequestMessage(HttpMethod.Post, URL);
+        request.Headers.Authorization = new AuthenticationHeaderValue(this._apiKey);
+        request.Content = content;
+        using var response = await this._httpClient.SendAsync(request, ct).ConfigureAwait(false);
+        await ThrowIfNotSuccessStatusCodeAsync("Failed to upload file.", response, ct)
+            .ConfigureAwait(false);
```
Krzysztof318, line 193:
```suggestion
        using var response = await this._httpClient.SendWithSuccessCheckAsync(request, ct).ConfigureAwait(false);
```

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+        jsonRequest["audio_url"] = audioUrl;
+        if (executionSettings?.ExtensionData is not null)
+        {
+            foreach (var attribute in executionSettings.ExtensionData)
+            {
+                jsonRequest[attribute.Key] = JsonValue.Create(attribute.Value);
+            }
+        }
+
+        using var content = new StringContent(jsonRequest.ToJsonString(), Encoding.UTF8, "application/json");
```
Krzysztof318, line 217: Look for class HttpRequest, there are ready method to create http request with payload.
Example from my code
```cs
var request = HttpRequest.CreatePostRequest(endpoint, requestData);
        request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", this._apiKey);
        request.Headers.Add("User-Agent", HttpHeaderConstant.Values.UserAgent);
        request.Headers.Add(HttpHeaderConstant.Names.SemanticKernelVersion,
            HttpHeaderConstant.Values.GetAssemblyVersion(typeof(OpenAIHttpRequestFactory)));
```

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+        {
+            throw new AssemblyAIApiException($"Failed to create transcript. Reason: {property.GetString()!}");
+        }
+
+        return json.RootElement.GetProperty("id").GetString()!;
+    }
+
+    private async Task<JsonDocument> WaitForTranscriptToProcessAsync(string transcriptId, CancellationToken ct)
+    {
+        var url = $"https://api.assemblyai.com/v2/transcript/{transcriptId}";
```
Krzysztof318, line 236: Hardcoded url in methods is not good idea. Create at least a property BaseUri in class.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        while (!ct.IsCancellationRequested)
+        {
+            using var request = new HttpRequestMessage(HttpMethod.Get, url);
+            request.Headers.Authorization = new AuthenticationHeaderValue(this._apiKey);
+            using var response = await this._httpClient.SendAsync(request, ct).ConfigureAwait(false);
+            await ThrowIfNotSuccessStatusCodeAsync("Error waiting for transcript.", response, ct)
+                .ConfigureAwait(false);
+            var jsonStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
Krzysztof318, line 245: better way with exception mapping
```suggestion
            var json = await response.Content.ReadAsStringWithExceptionMappingAsync().ConfigureAwait(false);
```

dotnet/src/IntegrationTests/Connectors/AssemblyAI/AssemblyAIAudioToTextTests.cs
```diff
@@ -0,0 +1,220 @@
...
+                     throw new ArgumentException("'AssemblyAI:ApiKey' configuration is required.");
+
+        var service = new AssemblyAIAudioToTextService(apiKey, httpClient);
+
+        await using Stream audio = File.OpenRead($"./TestData/{Filename}");
+        var audioData = await BinaryData.FromStreamAsync(audio);
+
+        // Act
+        var result = await service.GetTextContentAsync(new AudioContent(audioData));
+
```
Krzysztof318, line 53: You can use ITestOutputHelper to write result from connector.

Krzysztof318 (changes requested):

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: Our customers are automatically on our latest and best model.
We do have a lightweight model coming in which case the `modelId` parameter may be useful for us, because the customer will be able to opt for this new model by its name.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,292 @@
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Text content from audio content.</returns>
+    public async Task<TextContent> GetTextContentAsync(
+        AudioContent content,
+        PromptExecutionSettings? executionSettings = null,
+        Kernel? kernel = null,
+        CancellationToken cancellationToken = default)
+    {
+        string uploadUrl;
+        using var stream = content.Data!.ToStream();
```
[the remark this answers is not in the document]
Swimburger: Good catch

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: I created `AudioStreamContent`. When looking at `StreamingKernelContent`, the class didn't feel relevant to this use case, but more relevant to streaming LLM results.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: I copied and modified the code of `SendWithSuccessCheckAsync`.
Our error response contains a JSON `error` property which is vital for users to debug the error, so I have to parse that differently.

Swimburger:
@Krzysztof318 I think I got to all the feedback.
I'm not convinced that adding `File` and `Url` to the `AudioContent` class is a good approach, but it gets the job done.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
dmytrostruk: nit: If url is used only in this method, we made a decision to keep constants closer to the place where it's used. If it's used in multiple methods, then we can place it on class level. But in general, both approaches are acceptable.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
```diff
@@ -17,6 +18,16 @@ public class AudioContent : KernelContent
...
+    /// <summary>
+    /// URL to the audio file.
+    /// </summary>
+    public Uri? AudioUrl { get; set; }
+
+    /// <summary>
+    /// FileInfo for the local audio file.
+    /// </summary>
+    public FileInfo? AudioFile { get; set; }
+
```
Krzysztof318, line 30: Remove FileInfo completely, it is the user's responsibility to upload the file to Binary Data. Copy all the code from [ImageContent](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs).

Sorry for not specifying this earlier.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextExecutionSettings.cs
```diff
@@ -1,14 +1,18 @@
...
 
 namespace Microsoft.SemanticKernel.Connectors.AssemblyAI;
 
 /// <summary>
-/// Execution settings for AssemblyAI speech-to-text request.
+/// Execution settings for AssemblyAI speech-to-text execution.
 /// </summary>
-[Experimental("SKEXP0005")]
+[Experimental("SKEXP0033")]
 public class AssemblyAIAudioToTextExecutionSettings : PromptExecutionSettings
```
Krzysztof318, line 12: Take a look at OpenAIPromptExecution settings and use same pattern...

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,282 @@
...
+    /// <param name="apiKey">OpenAI API Key</param>
+    /// <param name="httpClient"></param>
+    public AssemblyAIAudioToTextService(
+        string apiKey,
+        HttpClient httpClient
+    )
+    {
+        this._apiKey = apiKey;
+        this._httpClient = httpClient;
+    }
```
Krzysztof318, line 48:
```suggestion
    public AssemblyAIAudioToTextService(
        string apiKey,
        HttpClient? httpClient = null
    )
    {
        this._apiKey = apiKey;        
        this._httpClient = HttpClientProvider.GetHttpClient(httpClient);
    }
```

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,282 @@
...
+
+    /// <summary>
+    /// Create a URL string that includes the default BaseUrl if the BaseAddress on _httpClient isn't set.
+    /// </summary>
+    /// <param name="url">URL without base.</param>
+    /// <returns>URL with or without BaseUrl.</returns>
+    private string Url(string url)
+    {
+        return this._httpClient.BaseAddress is null ? $"{BaseUrl}{url}" : url;
```
Krzysztof318, line 223: I think you wanted to combine _httpClient.BaseAdress with the url and not return the url itself?

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,282 @@
...
+                case "error":
+                    var errorString = json.RootElement.GetProperty("error").GetString()!;
+                    throw new KernelException($"Failed to create transcript. Reason: {errorString}");
+                default:
+                    throw new KernelException("Unexpected transcript status. This code shouldn't be reachable.");
+            }
+        }
+
+        ct.ThrowIfCancellationRequested();
+        throw new KernelException("This code is unreachable.");
```
Krzysztof318, line 213: If the code is unreachable, what is it for?

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,282 @@
...
+            }
+
+            response.EnsureSuccessStatusCode(); // will always throw
+        }
+        catch (Exception e)
+        {
+            throw new HttpOperationException(response.StatusCode, responseContent, e.Message, e);
+        }
+
+        throw new KernelException("Unreachable code.");
```
Krzysztof318, line 280: If the code is unreachable, what is it for?

Krzysztof318 (changes requested):
I left some coments. Please also take a look at OpenAi, HuggingFace and Gemini connector (feature-connector-gemini branch), in these connectors you can see many used patterns.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContent.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+
+namespace Microsoft.SemanticKernel.Contents;
+
+/// <summary>
+/// Represents audio content.
+/// </summary>
+[Experimental("SKEXP0005")]
+public class AudioStreamContent : KernelContent
```
Krzysztof318, line 13: What do you think about this @dmytrostruk @RogerBarreto ? Should it derive from StreamingKernelContent like I said earlier?

RogerBarreto:
Hi @Swimburger thanks for your contribution! 

This PR was promoted to a dedicated Connector feature branch, `feature-connectors-assemblyai`.

Please any additional change feel free to create small incremental PRs againts your dedicated branch.

FOr more details about the process see: https://github.com/microsoft/semantic-kernel/blob/feature-connectors-assemblyai/docs/decisions/0031-feature-branch-strategy.md

I'm closing this and waiting for you great contributions on the new branch. Thanks.

RogerBarreto:
@Swimburger, suggestion for the changes on the connector, seggregate the HttpClient out of your service, make it independent and being able to be provided following similar patterns we have for `OpenAI`, `Ollama`, `Gemini` and `HuggingFace (WIP)`.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: If there's a BaseAddress configured on the HTTP client, then we don't need to add the const `BaseUrl`.
`Url()` returns `https://api.assemblyai.com/path` if `this._httpClient.BaseAddress is null`
otherwise it returns `path`.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: If `ct.ThrowIfCancellationRequested()` is reached, then the cancellation token is cancelled, so `ct.ThrowIfCancellationRequested()` will throw.
Thus the next line is unreachable, but for compiler reasons, the method needs to return something or throw.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger:
```csharp
            response.EnsureSuccessStatusCode(); // will always throw
        }
        catch (Exception e)
        {
            throw new HttpOperationException(response.StatusCode, responseContent, e.Message, e);
        }
```
Because the HTTP response at this point is not a success code, `EnsureSuccessStatusCode()` will always throw.
Inside of the catch, it will always throw.
So `throw new KernelException("Unreachable code.")` is unreachable, but for the compiler it needs to return or throw something.

Krzysztof318: If you know `ensure` method will always throws in your case, just throw new right exception. This will be more elegant :)

Krzysztof318:
btw. You can apply experimental attribute at assembly level like here https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Experimental/Agents/AssemblyInfo.cs#L6

Swimburger:
I incorporated the new feedback.

Some thoughts: 
I don't like how `AudioContent` requires you to specify the `MediaType` of the binary data as it's an extra thing that devs need to ensure to configure correctly.
The MediaType isn't something that needs to be specified for our API, not sure about other STT APIs.

I feel like both `ImageContent` and `AudioContent` need a separate class for passing the `Uri` vs the `BinaryData`.
It seems odd to combine them.

Swimburger:
FYI, I had a conversation with Matthew Bolanos at Orlando Code Camp 🔥 
For now, we're only adding our "async" transcript model, meaning upload audio, process audio, return transcript.
When SK adds support for real-time transcription, we'll be happy to integrate our real-time transcript service too.

We'll think about other models and features too, but we want to get the basic in first.

Krzysztof318:
@Swimburger feel free to extend SK abstraction with live streaming :)
but do it in next PRs and try making PRs small so it will be easy for us to review

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,276 @@
...
+                throw new ArgumentException("File URI is not allowed. Use `AudioContent.Stream` or `AudioContent.File` to transcribe a local file instead.");
+            }
+
+            uploadUrl = content.Uri.ToString();
+        }
+        else if (content.Data is not null)
+        {
+            using var stream = content.Data!.ToStream();
+            uploadUrl = await this.UploadFileAsync(stream, cancellationToken).ConfigureAwait(false);
+        }
```
Krzysztof318, line 70: It's a bit inconsistent with AudioContent where you gave priority to BinaryData and here you check the URI first.
`/// <summary>
    /// Returns the string representation of the audio.
    /// BinaryData audio will be represented as DataUri
    /// Remote Uri audio will be represented as is
    /// </summary>
    /// <remarks>
    /// When Data is provided it takes precedence over URI
    /// </remarks>
    {
        return this.BuildDataUri() ?? this.Uri?.ToString() ?? string.Empty;
    }`

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,276 @@
...
+        this._httpClient = HttpClientProvider.GetHttpClient(httpClient);
+    }
+
+    /// <inheritdoc />
+    public async Task<TextContent> GetTextContentAsync(
+        AudioContent content,
+        PromptExecutionSettings? executionSettings = null,
+        Kernel? kernel = null,
+        CancellationToken cancellationToken = default)
+    {
```
Krzysztof318, line 54:
```suggestion
    {
        Verify.NotNull(content);
```

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,276 @@
...
+    /// <param name="executionSettings">The AI execution settings (optional).</param>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Text content from audio content.</returns>
+    public async Task<TextContent> GetTextContentAsync(
+        AudioStreamContent content,
+        PromptExecutionSettings? executionSettings = null,
+        Kernel? kernel = null,
+        CancellationToken cancellationToken = default)
+    {
```
Krzysztof318, line 104:
```suggestion
    {
        Verify.NotNull(content);
```

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,276 @@
...
+    private async Task<string> CreateTranscriptAsync(
+        string audioUrl,
+        PromptExecutionSettings? executionSettings = null,
+        CancellationToken ct = default
+    )
+    {
+        var url = this.Url("v2/transcript");
+        var jsonRequest = new JsonObject();
+        jsonRequest["audio_url"] = audioUrl;
+        if (executionSettings?.ExtensionData is not null)
```
Krzysztof318, line 146: You check only extensions data but if it will be AssemblyAIPromoExecutionSettings you ignore polling property?
Take a look how executions settings are loaded in gemini https://github.com/microsoft/semantic-kernel/blob/642747890517a52f6f81a38d35c1ed68dbcf05c5/dotnet/src/Connectors/Connectors.GoogleVertexAI/Settings/GeminiPromptExecutionSettings.cs#L185

Krzysztof318 (changes requested):
Looks better :)
I left some nit comments

dotnet/src/IntegrationTests/Connectors/AssemblyAI/AssemblyAIAudioToTextTests.cs
```diff
@@ -0,0 +1,260 @@
...
+                     throw new ArgumentException("'AssemblyAI:ApiKey' configuration is required.");
+
+        var service = new AssemblyAIAudioToTextService(apiKey, httpClient);
+
+        await using Stream audio = File.OpenRead($"./TestData/{Filename}");
+        var textExecutionSettings = new AssemblyAIAudioToTextExecutionSettings
+        {
+            ExtensionData = new Dictionary<string, object>
+            {
+                ["language_code"] = "en_us"
```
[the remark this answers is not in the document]
Swimburger: We actually have a lot of options. We will have an SDK that will have a class `TranscriptParams` for all the options.
When that is ready, should I include a property of `TranscriptParams` in `AssemblyAIAudioToTextExecutionSettings`, or should I look for a solution that will generate all the options as properties of `AssemblyAIAudioToTextExecutionSettings`?

I was thinking to use `ExtensionData` for now.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: The polling property isn't being passed to the API, it is used by `AssemblyAIAudioToTextService.WaitForTranscriptToProcessAsync` where the polling of the transcript resource occurs.
This may be inappropriate use of execution settings. Maybe it should be a property of `AssemblyAIAudioToTextService` itself?

dotnet/src/IntegrationTests/Connectors/AssemblyAI/AssemblyAIAudioToTextTests.cs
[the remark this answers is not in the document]
Krzysztof318: Ok, so leave as is for now

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Krzysztof318: User will access your service by abstraction IAudioToTextService so all additional properties should be in PromptExecutionSettings.

dotnet/src/IntegrationTests/Connectors/AssemblyAI/AssemblyAIAudioToTextTests.cs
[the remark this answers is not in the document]
RogerBarreto: @Swimburger, this is a good question this change can be in the ExecutionSettings but add this in a subsequent PR, incremental PRs are preferable as you have a dedicated feature branch.

Krzysztof318:
@Swimburger merge main branch, there are changes in audio abstraction. Fix all building warnings/errors if any after merge and at the end run dotnet format.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: Got it. I ended up moving it to `AssemblyAIAudioToTextServiceOptions` and removed `AssemblyAIPromptExecutionSettings`.

Swimburger:
I added extensions to add the service to the kernel builder.

Krzysztof318:
@Swimburger I will take a look tomorrow your new changes. But please don't do this PR bigger. I had hope we can merge this,  but now I have to review new changes. 

Try making future PRs small.

dotnet/src/Connectors/Connectors.AssemblyAI/AssemblyAIServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,202 @@
...
+using Microsoft.SemanticKernel.Connectors.AssemblyAI;
+using Microsoft.SemanticKernel.Http;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Provides extension methods for <see cref="IServiceCollection"/> and related classes to configure AssemblyAI connectors.
+/// </summary>
+public static class AssemblyAIServiceCollectionExtensions
+{
```
Krzysztof318, line 17: These extensions are too complicated. I don't think using an options is needed. Leave only two methods, first for kernelBuilder and second for serviceCollection. Service extensions should also allow passing HttpClient by end developer. As example first two methods: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs Only pass apiKey, endpoint and httpclient. Polling interval pass as execution settings like before.

dotnet/Directory.Packages.props
```diff
@@ -22,6 +22,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.3.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.0" />
     <PackageVersion Include="Microsoft.Bcl.TimeProvider" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="8.0.0" />
```
Krzysztof318, line 25: Remove this after you remove options

Krzysztof318 (changes requested):

dotnet/src/Connectors/Connectors.AssemblyAI/AssemblyAIServiceCollectionExtensions.cs
Swimburger: This is usually how I set up my libraries, following [this guidance for library authors](https://learn.microsoft.com/en-us/dotnet/core/extensions/options-library-authors).
I'm not sure why it would be different here, but I'll simplify it for now.

Krzysztof318: We can consider add more extensions to DI but then we also should extend others connectors to keep things consistent. But for now I'd like to minimize less important code, we can back to that when the assemblyAI branch will be ready for merge to main.

Swimburger:
I simplified the service extensions and move the polling interval back to execution settings.

dotnet/src/Connectors/Connectors.AssemblyAI.UnitTests/AudioToText/AssemblyAIAudioToTextServiceExtensionsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.AudioToText;
+using Microsoft.SemanticKernel.Connectors.AssemblyAI;
+using Xunit;
+
+namespace SemanticKernel.Connectors.AssemblyAI.UnitTests.AudioToText;
+
+/// <summary>
+/// Unit tests for <see cref="AssemblyAIAudioToTextService"/> class.
```
Krzysztof318, line 11:
```suggestion
/// Unit tests for <see cref="AssemblyAIServiceCollectionExtensions"/> class.
```

dotnet/src/Connectors/Connectors.AssemblyAI.UnitTests/AudioToText/AssemblyAIAudioToTextServiceExtensionsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+/// </summary>
+public sealed class AssemblyAIAudioToTextServiceExtensionsTests
+{
+    private const string ApiKey = "Test123";
+    private const string Endpoint = "http://localhost:1234/";
+    private const string ServiceId = "AssemblyAI";
+
+    [Fact]
+    public void AddServiceWithParameters()
```
Krzysztof318, line 20: You can add one more test for IServiceProvider also

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+            foreach (var attribute in executionSettings.ExtensionData)
+            {
+                jsonRequest[attribute.Key] = JsonValue.Create(attribute.Value);
+            }
+        }
+
+        using var request = HttpRequest.CreatePostRequest(url, jsonRequest);
+        this.AddDefaultHeaders(request);
+        using var response = await this.SendWithSuccessCheckAsync(this.HttpClient, request, ct).ConfigureAwait(false);
+        var jsonStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
Krzysztof318, line 161:
```suggestion
        using var jsonStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
I know disposing response dispose also stream, but good practice is always call dispose if something is disposable.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+            pollingInterval = aaiSettings.PollingInterval;
+        }
+
+        while (true)
+        {
+            ct.ThrowIfCancellationRequested();
+            using var request = HttpRequest.CreateGetRequest(url);
+            this.AddDefaultHeaders(request);
+            using var response = await this.SendWithSuccessCheckAsync(this.HttpClient, request, ct).ConfigureAwait(false);
+            var jsonStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
Krzysztof318, line 191:
```suggestion
            using var jsonStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
I know disposing response dispose also stream, but good practice is always call dispose if something is disposable.

dotnet/src/Connectors/Connectors.AssemblyAI/NUGET.md
Krzysztof318, line 1: You can remove that we have common nuget file here https://github.com/microsoft/semantic-kernel/tree/feature-connectors-gemini/dotnet/nuget

dotnet/src/Connectors/Connectors.AssemblyAI/icon.png
Krzysztof318, line 1: Same like with nuget.md to remove because https://github.com/microsoft/semantic-kernel/tree/feature-connectors-gemini/dotnet/nuget

dotnet/src/Connectors/Connectors.UnitTests/Connectors.UnitTests.csproj
```diff
@@ -42,6 +42,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\SemanticKernel.Core\SemanticKernel.Core.csproj" />
+    <ProjectReference Include="..\Connectors.AssemblyAI\Connectors.AssemblyAI.csproj" />
```
Krzysztof318, line 45: You have separate AssemblyAI.UnitTests project, you still need this line? Same for `NoWarn` and `<None Update="AssemblyAI\TestData\*">`

Krzysztof318 (approved):
A few minor fixes. The rest of the code satisfies me :)

I would like to thank you at all for your contribution, I hope you and AssemblyAI will bring audio in SK to a new level 😄

dotnet/src/Connectors/Connectors.UnitTests/Connectors.UnitTests.csproj
Swimburger: Good catch

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+namespace Microsoft.SemanticKernel.Connectors.AssemblyAI;
+
+/// <summary>
+/// AssemblyAI speech-to-text service.
+/// </summary>
+public sealed class AssemblyAIAudioToTextService : IAudioToTextService
+{
+    private const string FallbackBaseUrl = "https://api.assemblyai.com/";
+    internal string ApiKey { get; }
+    internal HttpClient HttpClient { get; }
```
SergeyMenshykh, line 27: Shouldn't the two properties - ApiKey and HttpClient - be private?

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+    }
+
+    /// <summary>
+    /// Create a URL string that includes the default BaseUrl if the BaseAddress on HttpClient isn't set.
+    /// </summary>
+    /// <param name="url">URL without base.</param>
+    /// <returns>URL with or without BaseUrl.</returns>
+    private string Url(string url)
```
SergeyMenshykh, line 217: Please add a verb to the method name, for example: "CreateUrl" or "GetUrl."

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+            metadata: null
+        );
+    }
+
+    private async Task<string> UploadFileAsync(Stream audioStream, CancellationToken ct)
+    {
+        var url = this.Url("v2/upload");
+        using var content = new StreamContent(audioStream);
+        content.Headers.ContentType = new MediaTypeHeaderValue("application/octet-stream");
+        using var request = new HttpRequestMessage(HttpMethod.Post, url);
```
SergeyMenshykh, line 131: Please separate logical blocks - getting the URL, creating the stream, creating the request, etc - with empty lines for better readability.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+        this.AddDefaultHeaders(request);
+        request.Content = content;
+        using var response = await this.SendWithSuccessCheckAsync(this.HttpClient, request, ct).ConfigureAwait(false);
+        var jsonStream = await response.Content.ReadAsStreamAndTranslateExceptionAsync().ConfigureAwait(false);
+        var json = await JsonDocument.ParseAsync(jsonStream, cancellationToken: ct).ConfigureAwait(false);
```
SergeyMenshykh, line 136: Shouldn't the jsonStream be disposed??

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+        return json.RootElement.GetProperty("upload_url").GetString()
+               ?? throw new KernelException("Property 'upload_url' expected but not found.");
+    }
+
+    private async Task<string> CreateTranscriptAsync(
+        string audioUrl,
+        PromptExecutionSettings? executionSettings = null,
+        CancellationToken ct = default
```
SergeyMenshykh, line 144: We tend to avoid setting default or null values for cancellation tokens in non-public methods, in order to ensure that method callers always provide them. Same for the other non-public methods in this PR.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+                case "queued":
+                    await Task.Delay(pollingInterval, ct).ConfigureAwait(false);
+                    break;
+                case "completed":
+                    return json;
+                case "error":
+                    var errorString = json.RootElement.GetProperty("error").GetString()!;
+                    throw new KernelException($"Failed to create transcript. Reason: {errorString}");
+                default:
+                    throw new KernelException("Unexpected transcript status. This code shouldn't be reachable.");
```
SergeyMenshykh, line 207: Please include the unknown status into the exception details.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -0,0 +1,281 @@
...
+
+    private void AddDefaultHeaders(HttpRequestMessage request)
+    {
+        request.Headers.Authorization = new AuthenticationHeaderValue(this.ApiKey);
+        request.Headers.Add("User-Agent", HttpHeaderConstant.Values.UserAgent);
+        request.Headers.Add(HttpHeaderConstant.Names.SemanticKernelVersion,
+            HttpHeaderConstant.Values.GetAssemblyVersion(typeof(AssemblyAIAudioToTextService)));
+    }
+
+    private async Task<HttpResponseMessage> SendWithSuccessCheckAsync(
```
SergeyMenshykh, line 230: Please use the existing SendWithSuccessCheckAsync utility method and consider parsing the response content, if it really needed, at a higher level to avoid spreading the logic of the SendWithSuccessCheckAsync method throughout the code.

Krzysztof318:
@Swimburger Please familiarize yourself with this ADR, especially the part that the link leads to, it will help you avoid problems in future PRs. https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0031-feature-branch-strategy.md#merging-main-branch-to-feature-branch-before-finish-feature

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: I made them internal for unit testing purposes. 
The unit test verifies that the API key and http client configured through the extensions is the same as on the service.

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: I'll remove this for now. We can discuss this in a separate PR.

Swimburger:
I updated the code resolving the latest feedback.

dotnet/src/Connectors/Connectors.AssemblyAI/AssemblyAIServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// <param name="endpoint">The endpoint URL to the AssemblyAI API.</param>
+    /// <param name="serviceId">A local identifier for the given AI service.</param>
+    /// <param name="httpClient">The HttpClient to use with this service.</param>
+    /// <returns>The same instance as <paramref name="services"/>.</returns>
+    public static IServiceCollection AddAssemblyAIAudioToText(
+        this IServiceCollection services,
+        string apiKey,
+        string? endpoint = null,
+        string? serviceId = null,
+        HttpClient? httpClient = null
```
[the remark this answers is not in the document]
dmytrostruk: @Krzysztof318 It's only for `IServiceCollection` extension methods, since we expect HttpClient to be configured in service collection. While for `IKernelBuilder` extension methods we should accept `HttpClient` as a parameter.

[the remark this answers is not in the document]
Swimburger: I see, in that case, I added the parameter back to the kernel builder extension, but not the service collection extension.

Swimburger:
The OpenAI audio tests are failing because we made `BinaryData.MediaType` mandatory to match image content logic.
I think we should be laxer here as the user doesn't need to specify this to do audio-to-text.

Krzysztof318:
@Swimburger You are behind main branch. Meanwhile in main there already are many fixes and changes related to BinaryContent. And one important PR waits for merge https://github.com/microsoft/semantic-kernel/pull/5229

But keep in mind you will have complications if you merge main to this PR.
Ask for merge via git command main to your feature branch someone with write access to SK repo.

> @Swimburger Please familiarize yourself with this ADR, especially the part that the link leads to, it will help you avoid problems in future PRs. https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0031-feature-branch-strategy.md#merging-main-branch-to-feature-branch-before-finish-feature

Swimburger:
I'll wait for #5229 to merge and then request a merge from the main to the feature branch.

Swimburger:
> @Swimburger You are behind main branch. Meanwhile in main there already are many fixes and changes related to BinaryContent. And one important PR waits for merge #5229
> 
> But keep in mind you will have complications if you merge main to this PR. Ask for merge via git command main to your feature branch someone with write access to SK repo.

Now that #5229 has been merged, can someone merge `main` via git command into the `feature-connectors-assemblyai` branch?
```bash
git checkout feature-connectors-assemblyai && git merge main
```

RogerBarreto:
@Swimburger Yes, will look into it.

Krzysztof318:
Small suggestion;
Since your branch will likely live some time and you want to develop a comprehensive connector for AssemblyAI. I propose you to create similar issue to this one #4680 it allows us tracking changes and eventually help you with writing new features.

This is not necessary, but it will nice to have such issue :)

Swimburger:
I merged the changes from the feature branch into this PR branch, and implemented the new interface method for OpenAI's STT services.

Swimburger:
> @Swimburger Small suggestion; Since your branch will likely live some time and you want to develop a comprehensive connector for AssemblyAI. I suggest you to create similar issue to this one #4680 it allows us tracking changes and eventually help you with writing new features.
> 
> This is not necessary, but it will be nice to have such issue :)

Done: #5392

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContent.cs
```diff
@@ -29,4 +29,20 @@ public AudioStreamContent(Stream stream, string? modelId = null, IReadOnlyDictio
     {
         this.Stream = stream;
     }
+
+    /// <summary>
+    /// Converts an AudioStreamContent to AudioContent by loading the stream data into memory.
+    /// </summary>
+    /// <returns></returns>
+    public AudioContent ToAudioContent()
```
Krzysztof318, line 37: Suggestion;
Move this method to static AudioStreamContentExtensions class

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContent.cs
```diff
@@ -29,4 +29,20 @@ public AudioStreamContent(Stream stream, string? modelId = null, IReadOnlyDictio
...
+    {
+        if (this.Stream is MemoryStream stream)
+        {
+            return new AudioContent(stream.ToArray());
+        }
+
+        using var memoryStream = new MemoryStream();
+        this.Stream.CopyTo(memoryStream);
+        return new AudioContent(memoryStream.ToArray());
```
Krzysztof318, line 46: Maybe it's not big deal but using BinaryReader will be more efficient.
https://code-maze.com/create-byte-array-from-stream-in-csharp/

dotnet/src/SemanticKernel.Abstractions/AI/AudioToText/IAudioToTextService.cs
```diff
@@ -27,4 +27,18 @@ public interface IAudioToTextService : IAIService
         PromptExecutionSettings? executionSettings = null,
         Kernel? kernel = null,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Get text contents from audio content.
+    /// </summary>
+    /// <param name="content">Audio content.</param>
```
Krzysztof318, line 34: audio stream content or <see cref="AudioStreamContent"...

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
```diff
@@ -45,19 +44,19 @@ public sealed class AssemblyAIAudioToTextService : IAudioToTextService
...
         PromptExecutionSettings? executionSettings = null,
         Kernel? kernel = null,
-        CancellationToken cancellationToken = default)
+        CancellationToken cancellationToken = default
+    )
     {
         Verify.NotNull(content);
 
         string uploadUrl;
         if (content.Data is not null)
```
Krzysztof318, line 57: content.Data is { Empty: false }

Krzysztof318 (changes requested):
nit; 😄

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Krzysztof318: This is not good practise; bending code to unit testing. Change this properties to private.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContent.cs
Swimburger: I moved it to an extension class, is there a specific location these should go?

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Swimburger: At this level we should check whether `Data` is null, if it is null, it should mean the user created the `AudioContent` with `Uri`.
I can check whether the `content.Data.IsEmpty` inside the first null check. What would the expected behavior be in case it is empty? Return an empty array of text content, or throw an exception?

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContent.cs
Krzysztof318: Now leave as is. We will see later

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContentExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        if (content is null) { throw new ArgumentNullException(nameof(content)); }
+
+        byte[] bytes;
+        using (var binaryReader = new BinaryReader(content.Stream))
+        {
+            bytes = binaryReader.ReadBytes((int)content.Stream.Length);
+        }
+
+        return new AudioContent(bytes);
```
Krzysztof318, line 27: simplify
```suggestion
        using var binaryReader = new BinaryReader(content.Stream);
        return new AudioContent(binaryReader.ReadBytes((int)content.Stream.Length));
```

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Krzysztof318: I think current solution is good if data is null or non-null but empty it checks uri, if uri is also null it throws exception

dotnet/src/Connectors/Connectors.AssemblyAI/AudioToText/AssemblyAIAudioToTextService.cs
Krzysztof318: @Swimburger I have never used this before. But it may satisfy your requirement without making properties internal/public. https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-8.0

Swimburger: @Krzysztof318 net8 feature, but all projects target netstandard2.0 or net6.
I don't mind targeting net8, or omitting the verification of API key and http client.

Krzysztof318: Okay, I forgot we are still targeting net6.0 in tests projects.

Swimburger: I changed them to private fields and removed the verification in unit tests.

Swimburger:
What are the next steps to get this PR merged? 🙇

RogerBarreto:
Refreshed the feature branch with the latest `main`. Will add some comments on top of it. 
Looking forward to get this one merged next week in the feature branch.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContentExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <summary>
+    /// Converts an AudioStreamContent to AudioContent by loading the stream data into memory.
+    /// </summary>
+    /// <returns>An AudioContent object from AudioStreamContent's stream</returns>
+    public static AudioContent ToAudioContent(this AudioStreamContent content)
+    {
+        if (content is null) { throw new ArgumentNullException(nameof(content)); }
+
+        using var binaryReader = new BinaryReader(content.Stream, Encoding.Default, leaveOpen: true);
+        return new AudioContent(binaryReader.ReadBytes((int)content.Stream.Length));
```
Krzysztof318, line 23: nit;
with leaveOpen true you should also reset stream position after reading all data.

Swimburger: Updated. Does it make sense to reset it back to its original position like I did here?

Krzysztof318: I don't think so, reset position to 0. This method should has also locking mechanism to be thread-safe in my opinion.

Swimburger: I set it to 0, and used `Stream.Synchronized` to make it thread-safe.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContentExtensions.cs
```diff
@@ -19,15 +19,14 @@ public static AudioContent ToAudioContent(this AudioStreamContent content)
     {
         if (content is null) { throw new ArgumentNullException(nameof(content)); }
 
-        var originalPosition = content.Stream.Position;
+        var stream = Stream.Synchronized(content.Stream);
```
Krzysztof318, line 22: This doesn't guarantee thread safety. `Stream.Synchronized(content.Stream);` doesn't block an other caller from making the call again `Stream.Synchronized(content.Stream);`

This would make sense it content.stream would be instance of `Stream.Synchronized(content.Stream)`

RogerBarreto:
@Swimburger the PR is merged in the feature branch as it's getting bigger and long running.

New changes needs to be made regarding the Abstractions as suggested above, looking forward for the next PR's!

dotnet/src/SemanticKernel.Abstractions/Contents/AudioStreamContent.cs
Swimburger: An `IAsyncEnumerable<StreamingAudioContent>` may make sense for a different scenario, like listening for audio chunks from a microphone, maybe?
The point was to use the native `Stream` in .NET as that's the most efficient and expected way to load files in .NET, and we support file uploads up to 2.2GB.
There's an `OpenAIFileService` for OpenAI, maybe an `AssemblyAIFileService` makes more sense to upload the file with a stream and use the URL to pass to the `AssemblyAIAudioToTextService`.

PR #5112, opened 2024-02-21
.Net AudioContent class moved to the Microsoft.SemanticKernel namespace

SergeyMenshykh:
The AudioContent class lives in the Microsoft.SemanticKernel.Contents namespace, while the other content classes reside in the Microsoft.SemanticKernel namespace. This PR moves the AudioContent class to the Microsoft.SemanticKernel namespace to align it with the other content classes.


dmytrostruk (approved):
My mistake, thanks for fixing this. Pipelines are failing, since it's breaking change. I've added suppression files to proceed with merging.

SergeyMenshykh:
> My mistake, thanks for fixing this. Pipelines are failing, since it's breaking change. I've added suppression files to proceed with merging.

No problem at all. We should have done a better job reviewing the PR. Thanks for adding the suppression.

PR #5133, opened 2024-02-23
.Net: [HB Planner] Content type fix

### LINKED ISSUE #5127: .Net: Handlebars prompt template doesn't deserialize JSON response when `Content-Type` includes `charset`
**Describe the bug**
If the `Content-Type` header of a REST API JSON response includes a `charset` (e.g. `Content-Type: application/json; charset=utf-8`), the Handlebars prompt template handles the response as a string instead of as deserialized JSON.  

**To Reproduce**
The issue appears to be due to how the `Content-Type` is matched here:

https://github.com/microsoft/semantic-kernel/blob/fe29d1abba770914878b428f9b52c1f1f6ee7861/dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs#L211-L221

**Expected behavior**
The JSON response should be deserialized.

**Platform**
 - Language: C#
 - Source: `Microsoft.SemanticKernel.PromptTemplates.Handlebars` version `1.4.0`

teresaqhoang:
### Motivation and Context

Resolves #5127: If the Content-Type header of a REST API JSON response includes a charset (e.g. Content-Type: application/json; charset=utf-8), the Handlebars prompt template handles the response as a string instead of as deserialized JSON.

This PR changes the check from strict equality to a check that determine if the content type simply contains `application/json`

### Contribution Checklist


TaoChenOSU (approved):

dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
```diff
@@ -211,7 +211,7 @@ private static void ProcessPositionalArguments(KernelFunctionMetadata functionMe
         if (resultAsObject is RestApiOperationResponse restApiOperationResponse)
         {
             // Deserialize any JSON content or return the content as a string
-            if (string.Equals(restApiOperationResponse.ContentType, "application/json", StringComparison.OrdinalIgnoreCase))
+            if (restApiOperationResponse.ContentType.IndexOf("application/json", StringComparison.OrdinalIgnoreCase) > 0)
```
SergeyMenshykh, line 214: For some requests, such as 204, there might not be a content type, and the ContentType will return null. There is a PR - https://github.com/microsoft/semantic-kernel/pull/5021 to support these cases.

SergeyMenshykh (approved):

dotnet/src/Extensions/PromptTemplates.Handlebars/Helpers/KernelHelpers/KernelFunctionHelpers.cs
glahaye, line 214: Why not just use 
if (restApiOperationResponse.ContentType.Contains("application/json"))
???

teresaqhoang: Thanks for the call out! I added a null check

teresaqhoang: Something about this project doesn't recognize `public bool Contains(String value, StringComparison comparisonType);` and using Contains without the StringComparison parameter will cause a linter error

glahaye (approved):

Issue #4751, opened 2024-01-26
.Net: HandlebarsPlanner "hallucinates" helpers that actually *do* exist (and sometimes work) when using LLamaSharp


godefroi:
**Describe the bug**
I am attempting to use plugin function with a local language model using the LLamaSharp implementation of `IChatCompletionService` and am having trouble. I have had two different results across two runs; the first time I got an exception:
```
Unhandled exception. Microsoft.SemanticKernel.KernelException: [HallucinatedHelpers] The plan references hallucinated helpers: Helper 'unless'
```

The generated Handlebars template looked like this:
```
{{!-- Step 1: Initialize variables --}}
{{set "currentState" ""}}
{{set "newState" false}}

{{!-- Step 2: Get current state of the light --}}
{{#if (LightPlugin-GetState)}}
  {{set "currentState" (LightPlugin-GetState)}}
{{/if}}

{{!-- Step 3: Change the state of the light if it's currently off --}}
{{#if (unless currentState)}}
  {{set "newState" true}}
  {{set "result" (LightPlugin-ChangeState newState)}}
  {{json result}}
{{/if}}
```

The second time I ran the code, I got a different template generated:
```
{{!-- Step 1: Initialize variables --}}
{{set "currentState" ""}}
{{set "newState" false}}

{{!-- Step 2: Get current state of the light --}}
{{#if (LightPlugin-GetState)}}
  {{set "currentState" (LightPlugin-GetState)}}
{{/if}}

{{!-- Step 3: Change state of the light if it's currently off --}}
{{#unless (equals currentState true)}}
  {{set "newState" true}}
  {{set "result" (LightPlugin-ChangeState newState)}}
{{else}}
  {{json "The lights are already on."}}
{{/unless}}

{{!-- Step 4: Output the result --}}
{{json (concat "The lights have been " (if (equals currentState false) "turned on" "remain unchanged"))}}
```

Note that this plan, while different, still references the `unless` handler. Nevertheless, this plan was executed successfully (as shown in the full output below). After the function was executed, an exception was thrown indicating that `if` is a hallucinated handler:
```
Microsoft.SemanticKernel.KernelException: [HallucinatedHelpers] The plan references hallucinated helpers: Helper 'if'
```

**Platform**
 - OS: Windows
 - IDE: VS Code
 - Language: C#
 - Source: NuGet packages:
   - Microsoft.SemanticKernel version 1.0.1
[...]

Issue #4721, opened 2024-01-24
.Net: Handlebars Planner Failing to Parse JSON Array


rithvikb23:
I have an openapi plugin to get a list of people that hold a certain asset. When loading the openapi spec the handlebars planner is failing to parse the plan as the handlebars planner is incorrectly parsing the template. The openapi spec is rather complex with many fields and complex data types, but is still appropriately formatted and I have tested it in a swagger editor however the plan is failing. The planner is failing to parse the json array in step 2 below.

Here is the current step by step by the planner:
Goal: "What is my exposure to AAPL, id 123"

Handlebars Plan:
{{!-- Step 1: Set your necessary parameters --}}
{{set "id" "123"}}

{{!-- Step 2: Set your symbol(AAPL) in a fieldValues array --}}
{{set "fieldValues" (array "AAPL")}}

{{!-- Step 3: Set your applied filters --}}
**{{set "appliedFiltersArray" (array (json{ "op": "EQLT", "field": "exposure_ss", "id": "1", "fieldValues": fieldValues}))}}**

{{!-- Step 4: Invoke the PositionCrossReferencePlugin helper and store the result --}}
{{set "response" (PositionCrossReferencePlugin-getPositionCrossReference id=id faId=(array faId) appliedFilters=appliedFiltersArray)}}

**Expected behavior**
When loading this saved plan the issue is in step 2 with parsing the json array. I am given the error
**Plan creation failed. Error: Reached unparseable token in expression: )}}**
**HandlebarsDotNet.HandlebarsParserException: Reached unparseable token in expression: )}}**

 - OS: Windows
 - IDE: Visual Studio
 - Language: C#
 - Source: NuGet package version 1.1.0, pip package

Are json array handling in the handlebars planner currently not supported? Or is there a workaround?

joslat:
@rithvikb23 did you find any workaround to this?

teresaqhoang:
The error is a syntax error by the model here:  `json{ "op": "EQLT", "field": "exposure_ss", "id": "1", "fieldValues": fieldValues}`. I wondering if adding an example directly to the `json` helper definition in the prompt will help. 

We recently added more direction to better guide the model in creating initial variables. Can you pull the latest changes and see if your planner results have improved? 

If not, please reply to this thread with:
- openapi spec
- The CreatePlan prompt template
- Resulting plan 

And I'll help troubleshoot

teresaqhoang:
[...]

Issue #4250, opened 2023-12-14
.Net: [HandlebarsPlanner] Handle case when schema and ParameterType can co-exist


teresaqhoang:
Schema and ParameterType can co-exist i.e., when ParameterType = RestApiResponse 
- Default to schema? Needs investigation on other response types from SK

PR #5137, opened 2024-02-23
.Net: Tuning HB Planner - CreatePlan prompt, template extraction, and type handling

teresaqhoang:
### Motivation and Context

Resolves 
- #4751 
- #4250 
- #4721

This PR contains fixes for template extraction from model results; handling case where schema and parameter type can co-exist; + prompt tuning

### Contribution Checklist


SergeyMenshykh (approved):

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
```diff
@@ -153,7 +161,26 @@ private async Task<HandlebarsPlan> CreatePlanCoreAsync(Kernel kernel, string goa
...
+            else
+            {
+                // Parse the schema to extract any primitive types and set in ParameterType property instead
+                var parsedParameter = parameter.ParseJsonSchema();
+                if (parsedParameter.Schema is not null)
+                {
+                    complexParameterSchemas[parameter.GetSchemaTypeName()] = parameter.Schema.RootElement.ToJsonString();
+                }
+
+                return parsedParameter;
```
markwallace-microsoft, line 180: Single return would be better

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
```diff
@@ -89,13 +89,21 @@ private async Task<HandlebarsPlan> CreatePlanCoreAsync(Kernel kernel, string goa
             var chatCompletionService = kernel.GetRequiredService<IChatCompletionService>();
             modelResults = await chatCompletionService.GetChatMessageContentAsync(chatMessages, executionSettings: this._options.ExecutionSettings, cancellationToken: cancellationToken).ConfigureAwait(false);
 
-            Match match = Regex.Match(modelResults.Content, @"```\s*(handlebars)?\s*(.*)\s*```", RegexOptions.Singleline);
-            if (!match.Success)
+            // Regex breakdown:
+            // (```\s*handlebars){1}\s*: Opening backticks, starting boundary for HB template
+            // ((([^`]|`(?!``))+): Any non-backtick character or one backtick character not followed by 2 more consecutive backticks
+            // (\s*```){1}: Closing backticks, closing boundary for HB template
+            MatchCollection matches = Regex.Matches(modelResults.Content, @"(```\s*handlebars){1}\s*(([^`]|`(?!``))+)(\s*```){1}", RegexOptions.Multiline);
```
markwallace-microsoft, line 96: Should this regex be configurable?

markwallace-microsoft (approved):

dotnet/src/Planners/Planners.Handlebars/Handlebars/HandlebarsPlanner.cs
teresaqhoang: It's better to return here to avoid further unnecessary processing. Before the second conditional was encompassed in an `else if` so single return worked, but there can be scenarios with Schema and ParameterType can co-exist and we explicitly handle one scenario above, but for all other scenarios, we only want to prioritize and process the schema only and skip the ParameterType

markwallace-microsoft (approved):

PR #5149, opened 2024-02-26
.Net: Updated audio abstractions to return multiple values

dmytrostruk:
### Motivation and Context

This PR contains changes for audio abstractions to align with abstraction pattern in SK to return multiple values instead of just one to be more flexible in cases when AI connector returns multiple results (like in chat completion service with choices). 

For easier usage when only one result is provided by AI connector, it's possible to use extension method that is responsible for returning single result.

### Contribution Checklist


dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAIAudioToTextTests.cs
```diff
@@ -70,10 +70,10 @@ public async Task AzureOpenAIAudioToTextTestAsync()
         var audioData = await BinaryData.FromStreamAsync(audio);
 
         // Act
-        var result = await service.GetTextContentAsync(new AudioContent(audioData), new OpenAIAudioToTextExecutionSettings(Filename));
+        var result = await service.GetTextContentsAsync(new AudioContent(audioData), new OpenAIAudioToTextExecutionSettings(Filename));
```
RogerBarreto, line 73: For tests I believe we can use the extension method, and use the List when testing multiple.

dotnet/samples/KernelSyntaxExamples/Resources/test_audio.wav
RogerBarreto, line 1: Yeah a forgot about that, happy you added. 👍

RogerBarreto (approved):

dotnet/samples/KernelSyntaxExamples/Example82_Audio.cs
```diff
@@ -78,8 +78,9 @@ public async Task AudioToTextAsync()
         };
 
         // Read audio content from a file
-        ReadOnlyMemory<byte> audioData = await File.ReadAllBytesAsync(AudioFilePath);
-        AudioContent audioContent = new(new BinaryData(audioData));
+        var audioFileStream = EmbeddedResource.ReadStream(AudioFilename);
```
SergeyMenshykh, line 81: nit: use the 'using' operator to dispose the stream.

SergeyMenshykh (approved):

RogerBarreto (approved):

PR #5159, opened 2024-02-26
Python: split kernelfunction into fromprompt and frommethod classes

### LINKED ISSUE #4633: Python: Unify code paths for prompt (semantic function) and native functions
In .NET, prompts and native functions use the same underlying KernelFunction class to ensure they behave the same. This issue is to replicate this. If there aren't any functional changes because of this issue (i.e., the interface stays the same), then this issue should be ranked lower.

eavanvalkenburg:
### Motivation and Context

Closes #4633 and brings this part of the python sdk in sync with dotnet, also makes for a much more readable set of classes.

### Description

Added:
- KernelFunctionFromPrompt class
- KernelFunctionFromMethod class
- tests for both.

Changed:
- removed parts of KernelFunction that are specific to one of the two subtypes
- removed expose attribute of KernelParameterMetadata as that is no longer needed

### Contribution Checklist


python/semantic_kernel/functions/kernel_function_from_prompt.py
```diff
@@ -0,0 +1,223 @@
...
+                return data
+        if isinstance(prompt_execution_settings, PromptExecutionSettings):
+            data["prompt_execution_settings"] = {
+                prompt_execution_settings.service_id or "default": prompt_execution_settings
+            }
+        if isinstance(prompt_execution_settings, list):
+            data["prompt_execution_settings"] = {s.service_id or "default": s for s in prompt_execution_settings}
+        return data
+
+    async def _invoke_internal(
```
moonbox3, line 137: No need to call `gather_function_parameters()` as part of this method?

moonbox3 (approved):
Pure awesomeness. Great work!

python/semantic_kernel/functions/kernel_function_from_method.py
```diff
@@ -0,0 +1,150 @@
...
+            function=self.metadata,
+            value=result,
+            metadata={"arguments": arguments, "used_arguments": function_arguments},
+        )
+
+    async def _invoke_internal_stream(
+        self,
+        kernel: "Kernel",
+        arguments: KernelArguments,
+    ) -> AsyncIterable[List[StreamingKernelContent | Any]]:
```
moonbox3, line 117: Looks like Python 3.8/3.9 would be happier with a `Union` :)

python/semantic_kernel/functions/kernel_function_from_prompt.py
eavanvalkenburg: no, the internal function uses kernel and arguments, and with those they are complete and good to go.

moonbox3 (approved):

juliomenendez (approved):

alliscode (approved):

juliomenendez (approved):

PR #5201, opened 2024-02-27
Java: Implement a number of fixes from API review

johnoliver:
- Make constructors on classes that have builders protected
- Remove the addPlugin function from the kernel to make plugin set immutable, in favour of adding a copy method for creating a clone of a kernel that can have plugins added, i.e:
```
        kernel = kernel.copy()
                             .withPlugin(timePlugin)
                             .build()
```
- Add sync invocation methods in addition to async
- Add warning if `FunctionInvocation` is used in an unusual manner that might indicate a bug

### DIFF
--- a/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
+++ b/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
@@ -88,12 +88,26 @@ public class Kernel implements Buildable {
     }
 
     /**
+     * Creates a Builder that can create a copy of the {@code Kernel}. Use this method if you wish
+     * to modify the state of the kernel such as adding new plugins or services.
+     *
+     * @param kernel The kernel to copy.
+     * @return A Builder that can create a copy of the instance of {@code Kernel}.
+     */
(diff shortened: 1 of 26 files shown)


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/Kernel.java
```diff
@@ -65,10 +83,20 @@ public Kernel(
...
         return new Kernel.Builder();
     }
 
+    /**
+     * Creates a Builder that can create a copy of the current instance of {@code Kernel}. Use this
+     * method if you wish to modify the state of the kernel such as adding new plugins or services.
+     *
+     * @return A Builder that can create a copy of the current instance of {@code Kernel}.
+     */
+    public Builder copy() {
```
dsgrieve, line 96: I'd expect 'copy' to return a Kernel, not a Kernel.Builder. Maybe 'toBuilder' or a static 'Builder from(Kernel)' method in Kernel.Builder itself instead. I have a feeling this will be a pattern we repeat. Maybe a 'toBuilder' API in Buildable is in order?

dsgrieve (approved):

Issue #5226, opened 2024-02-27
.Net:  Reduce dependencies used by SK Abstractions project


markwallace-microsoft:
Goal is the dependency list will drop to just Microsoft.Extensions.DependencyInjection.Abstractions.dll and Microsoft.Extensions.Logging.Abstractions.dll.

- [x] Remove System.Memory.Data. BinaryData is just a convenience type introduced for the Azure SDK. It's not necessary, byte[] / Stream / ReadOnlyMemory<byte>... any or all of those is fine. BinaryData was introduced as a way to accept any of those three and then present any of those three out to a consumer. - https://github.com/microsoft/semantic-kernel/pull/5229
- [ ] Copy the minimum-needed source from the JSON schema lib into the abstractions lib so that dependency can be clipped.
- [ ] Add a net6.0 (and net8.0) asset that removes the netstandard dependency and the dependency's that are effectively polyfills for things in .NET Core itself

ihvo:
@SergeyMenshykh, are you ok if I take this issue?

ihvo:
> @SergeyMenshykh, are you ok if I take this issue?

A bit late I guess :)

SergeyMenshykh:
> > @SergeyMenshykh, are you ok if I take this issue?
> 
> I was a bit late I guess :)

Hi @ihvo, sorry, I already started :)

PR #5229, opened 2024-02-28
.Net: Removing dependency on BinaryData class from content classes

SergeyMenshykh:
### Motivation and Context
Today, SemanticKernel.Abstractions package has unnecessary dependency on System.Memory.Data package which according to this issue https://github.com/microsoft/semantic-kernel/issues/5226 should be removed to minimize number of dependencies and keep only absolutely necessary ones.

### Description
This PR replaces the usage of the System.Memory.Data.BinaryData type with ReadonlyMemory in the AudioContent, BinaryContent, and ImageContent classes. It also removes the dependency of SemanticKernel.Abstractions on the System.Memory.Data package.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
```diff
@@ -16,18 +16,29 @@ public class AudioContent : KernelContent
...
 
     /// <summary>
     /// Initializes a new instance of the <see cref="AudioContent"/> class.
     /// </summary>
-    /// <param name="data">The audio binary data.</param>
     /// <param name="modelId">The model ID used to generate the content.</param>
     /// <param name="innerContent">Inner content,</param>
     /// <param name="metadata">Additional metadata</param>
     [JsonConstructor]
+    public AudioContent(string? modelId = null, object? innerContent = null, IReadOnlyDictionary<string, object?>? metadata = null)
```
SergeyMenshykh, line 28: This constructor was introduced to support deserialization of AudioContent class. Otherwise, the deserialization process fails with the error - _"Each parameter in the deserialization constructor on type 'Microsoft.SemanticKernel.AudioContent' must bind to an object property or field on deserialization. Each parameter name must match with a property or field on the object. Fields are only considered when 'JsonSerializerOptions.IncludeFields' is enabled. The match can be case-insensitive."_

I believe it's related to a try to call the class constructor that has ReadOnlyMemory<byte>> data parameter with an argument having a different type - Nullable<ReadOnlyMemory<byte>> that the Data property has.

Krzysztof318:
@SergeyMenshykh So you need also remove BinaryData from imagecontent class.

SergeyMenshykh:
> @SergeyMenshykh So you need also remove BinaryData from imagecontent class.
 
Right, the plan is to have a PR per type. So, there will be at least two more PRs - one for ImageContent, and the other one for BinaryContent. I updated the PR description to make it clear.

markwallace-microsoft (approved):

SergeyMenshykh:
> > @SergeyMenshykh So you need also remove BinaryData from imagecontent class.
> 
> Right, the plan is to have a PR per type. So, there will be at least two more PRs - one for ImageContent, and the other one for BinaryContent. I updated the PR description to make it clear.

Actually, it appears that the PR per class will be too excessive, so I'll push all the changes as one PR.

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
```diff
@@ -17,9 +17,14 @@ public sealed class ImageContent : KernelContent
     public Uri? Uri { get; set; }
 
     /// <summary>
-    /// The image binary data.
+    /// The image media type.
     /// </summary>
-    public BinaryData? Data { get; set; }
+    public string? MediaType { get; set; }
```
Krzysztof318, line 22: You introduced MediaType for ImageContent but why not for AudioContent also?

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
Krzysztof318, line 1: ImageContent constructor with URI param should also accept optional MediaType param. Gemini requires always providing MediaType.

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
SergeyMenshykh: Make sense, added.

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
SergeyMenshykh: added

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
```diff
@@ -17,9 +17,14 @@ public sealed class ImageContent : KernelContent
...
+
+    /// <summary>
+    /// The image data.
+    /// </summary>
+    public ReadOnlyMemory<byte>? Data { get; set; }
```
SergeyMenshykh, line 27: @dersia, just a heads-up that we are changing the type of the Data property that you've added.

dotnet/samples/KernelSyntaxExamples/Example80_OpenAIFiles.cs
```diff
@@ -49,7 +50,7 @@ public async Task RunFileLifecycleAsync()
         WriteLine("SOURCE:");
         WriteLine($"# Name: {fileReference.FileName}");
         WriteLine("# Content:");
-        WriteLine(await fileContent.GetContentAsync());
+        WriteLine(Encoding.UTF8.GetString((await fileContent.GetContentAsync()).ToArray()));
```
stephentoub, line 53: This and others like it should be able to use .Span rather than .ToArray.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
```diff
@@ -88,17 +101,22 @@ public async Task<Stream> GetStreamAsync()
     /// <summary>
     /// The content stream
     /// </summary>
-    public async Task<BinaryData> GetContentAsync()
+    public async Task<ReadOnlyMemory<byte>> GetContentAsync()
     {
-        if (this._streamProvider != null)
+        if (this._streamProvider is not null)
```
stephentoub, line 106: I missed the PR that added all this BinaryContent stuff. What is the scenario where it's provided with a `Func<Task<Stream>>`?

dotnet/samples/KernelSyntaxExamples/Example80_OpenAIFiles.cs
SergeyMenshykh: Thanks, fixed.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
SergeyMenshykh: The callback/streamProvider is used in file download and upload scenarios - [GetFileContent](https://github.com/microsoft/semantic-kernel/blob/57ce2ad7bd8f2170b042b51f3421f7027613235a/dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs#L81), [UploadFileAsync](https://github.com/microsoft/semantic-kernel/blob/57ce2ad7bd8f2170b042b51f3421f7027613235a/dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs#L118). 

It looks like the provider allows to postpone stream creation (lazy stream initialization?) until the moment when it's actually requested via - [GetSrreamAsync](https://github.com/microsoft/semantic-kernel/blob/57ce2ad7bd8f2170b042b51f3421f7027613235a/dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs#L73) by a caller. 

However, taking into account that the GetStreamAsync method is called immediately after the GetFileContent one in [file download scenario](https://github.com/microsoft/semantic-kernel/blob/57ce2ad7bd8f2170b042b51f3421f7027613235a/dotnet/samples/KernelSyntaxExamples/Example85_AgentCharts.cs#L81) and just before sending POST http request in the [file upload scenario](https://github.com/microsoft/semantic-kernel/blob/57ce2ad7bd8f2170b042b51f3421f7027613235a/dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs#L129), I wonder if the lazy behavior is necessary? 

CC: @crickman, @dmytrostruk

stephentoub: From an API perspective, it's concerning that there's a `Content` property that won't return the content, and can't with such a delegate because it's synchronous and would require asynchronously blocking waiting for the content to be available.

It's also redoing the work on every access, so if someone uses GetStreamAsync/GetContentAsync multiple times, they're redoing all the work each time. And if the func that's provided is idempotent and returns the same stream each time, the consumer is both responsible for disposing the stream that's returned, but can't safely do so because a subsequent access to GetStreamAsync/GetContentAsync would then be using a disposed stream.

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
dersia: thanks for letting me know. interestingly enough I did implement it this way in my first implementation and it was suggested to switch to BinaryData during discussion. with all the dependencies that Abstraction had it didn't seem like a bad idea so I changed the implementation.

but good that we get rid of dependencies and I am happy that this stays in the way it was designed originally.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
SergeyMenshykh: The constructor is removed as it's no longer relevant.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
SergeyMenshykh: I've registered the task - https://github.com/microsoft/semantic-kernel/issues/5295 to revise the design of the class.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/AzureOpenAITextToAudioClient.cs
```diff
@@ -79,7 +79,7 @@ internal sealed class AzureOpenAITextToAudioClient
 
         var binaryData = await BinaryData.FromStreamAsync(stream, cancellationToken).ConfigureAwait(false);
 
-        return new List<AudioContent> { new(binaryData, modelId) };
+        return new List<AudioContent> { new(binaryData.ToMemory(), modelId) };
```
dmytrostruk, line 82: nit: There is small unnecessary conversion here, first we read HTTP response as `Stream`, then we convert it to `BinaryData` and then to `ReadOnlyMemory`. 

Instead, we should be able to read HTTP response as `byte[]` and convert it directly to `ReadOnlyMemory`:
```csharp
var bytes = await response.Content.ReadAsByteArrayAndTranslateExceptionAsync().ConfigureAwait(false);

return new List<AudioContent> { new(new(bytes), modelId) };
```

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAITextToAudioClient.cs
```diff
@@ -71,7 +71,7 @@ internal sealed class OpenAITextToAudioClient
 
         var binaryData = await BinaryData.FromStreamAsync(stream, cancellationToken).ConfigureAwait(false);
 
-        return new List<AudioContent> { new(binaryData, this._modelId) };
+        return new List<AudioContent> { new(binaryData.ToMemory(), this._modelId) };
```
dmytrostruk, line 74: Same comment about `Stream -> BinaryData -> ReadOnlyMemory` conversion.

dotnet/src/SemanticKernel.Abstractions/Contents/AudioContent.cs
```diff
@@ -16,7 +16,15 @@ public class AudioContent : KernelContent
     /// <summary>
     /// The audio binary data.
```
dmytrostruk, line 17: I noticed you did the same for `ImageContent`, should we update it here as well?
```suggestion
    /// The audio data.
```

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAITextToAudioClient.cs
SergeyMenshykh: Make sense, fixed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/AzureOpenAITextToAudioClient.cs
SergeyMenshykh: fixed

dmytrostruk (approved):

dmytrostruk (approved):

PR #5245, opened 2024-02-29
.Net - Feature: Agents Abstraction (DRAFT)

crickman:
### Contribution Checklist


kbeaugrand:
@SergeyMenshykh, 

Does agents support planners? I don't see this in the PR, but I think this is relevant to take care of that. In my mind, agents are interesting if they can have some behaviour and this could be achieved by those planners no?

Issue #4635, opened 2024-01-16
Python: add complex type support


matthewbolanos:

eavanvalkenburg:
@matthewbolanos @alliscode what is the scope of this one? native functions only?

eavanvalkenburg:
ping @alliscode I think is done with kernel arguments.

moonbox3:
Closing since #5260 is merged to main.

PR #5260, opened 2024-03-01
Python: Complex Type support for Method functions

eavanvalkenburg:
### Motivation and Context

Added support for complex types (classes) for KernelFunctionFromMethod.

Implements #4635 

### Description

This PR adds a field to the KernelParameterMetadata class for type_object
When that field is filled and supplied by a KernelFunctionFromMethod decorator, it is used during invoke to cast the type.
It checks if it is a pydantic type and calls `model_validate_json` in that case, otherwise invokes it directly as a init (like with a str), this then also casts strs to ints for example.

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/functions/kernel_function_decorator.py
+++ b/python/semantic_kernel/functions/kernel_function_decorator.py
@@ -3,9 +3,9 @@
 
 import logging
 from inspect import Parameter, Signature, isasyncgenfunction, isgeneratorfunction, signature
-from types import NoneType
 from typing import Any, Callable, Dict, Optional
 
+NoneType = type(None)
 logger = logging.getLogger(__name__)
 
 
--- a/python/semantic_kernel/functions/kernel_function_decorator.py
+++ b/python/semantic_kernel/functions/kernel_function_decorator.py
@@ -3,7 +3,7 @@
 
 import logging
 from inspect import Parameter, Signature, isasyncgenfunction, isgeneratorfunction, signature
-from typing import Callable, Optional, Tuple
+from typing import Any, Callable, Dict, Optional
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


markwallace-microsoft:
| Tests | Skipped | Failures | Errors | Time |
| ----- | ------- | -------- | -------- | ------------------ |
| 967 | 11 :zzz: | 0 :x: | 0 :fire: | 17.848s :stopwatch: |

python/semantic_kernel/functions/kernel_function_decorator.py
```diff
@@ -53,61 +54,66 @@ def decorator(func: Callable):
...
-    }
+        ret = _parse_annotation(param.annotation)
+    ret["name"] = param.name
+    if param.default != Parameter.empty:
+        ret["default_value"] = param.default
+    return ret
 
 
-def _parse_annotation(annotation: Parameter) -> Tuple[str, str, bool]:
+def _parse_annotation(annotation: Parameter) -> Dict[str, Any]:
```
juliomenendez, line 79: Would it make sense to create a class for this return value? Given that it has a defined structure maybe it could benefit from being a defined and typed object.

python/semantic_kernel/functions/kernel_function_decorator.py
```diff
@@ -53,61 +54,66 @@ def decorator(func: Callable):
...
+    ret = _parse_internal_annotation(annotation, True)
     if hasattr(annotation, "__metadata__") and annotation.__metadata__:
-        description = annotation.__metadata__[0]
-    return (description, *_parse_internal_annotation(annotation, True))
+        ret["description"] = annotation.__metadata__[0]
+    return ret
 
 
-def _parse_internal_annotation(annotation: Parameter, required: bool) -> Tuple[str, bool]:
+def _parse_internal_annotation(annotation: Parameter, required: bool) -> Dict[str, Any]:
```
juliomenendez, line 92: Same here, could a defined class help this returned structure?

python/semantic_kernel/functions/kernel_function_decorator.py
juliomenendez: Or is that what `KernelParameterMetadata` is for? Confusing because of the `Dict` being used across this function.

juliomenendez (approved):
Looking good, just one small comment.

python/semantic_kernel/functions/kernel_function_decorator.py
eavanvalkenburg: yeah this will be stored in the __kernel_function_parameters__ field of a function, so I don't want to make that more heavy then necessary, when creating a actual KernelFunctionFromMethod these get parsed into KernelParameterMetadata.

moonbox3 (approved):

PR #5266, opened 2024-03-01
.Net: minor change so that inputs are not skipped in Planner example

nkelly75:
### Description

While learning Semantic Kernel (and having already run various other tests successfully) I wanted to try out the Planner example (described [here](https://learn.microsoft.com/en-us/semantic-kernel/agents/planners/)). I added some sample inputs to `SimulatedInputText`. I then noticed that the chat loop was effectively skipping every second input due to a double call to `ReadLine()`. If there was an odd number of inputs we saw a failure due to the last `ReadLine()` returning null

The simple fix replaces the second `ReadLine()` call with the `userInput` variable from the first call.

Note: for illustration purposes, the PR includes my sample inputs commented out. These can be removed if necessary for approval.

### Contribution Checklist


dotnet/samples/DocumentationExamples/Planner.cs
```diff
@@ -93,5 +93,11 @@ await foreach (var content in result)
 
     public Planner(ITestOutputHelper output) : base(output)
     {
+        //SimulatedInputText = [
+        //    "What is 49 divided by 37?",
+        //    "What is the square root of 49?",
+        //    "What is the sum of all numbers from 1 to 10?",
+        //    "If my investment of 2130.23 dollars increased by 23%, how much would I have after I spent $5 on a latte?"
+        //];
```
markwallace-microsoft, line 101: @nkelly75 Thanks for the fix. Can you remove these commented out lines and we can get this merged.

markwallace-microsoft (approved):

dotnet/samples/DocumentationExamples/Planner.cs
nkelly75: Thanks @markwallace-microsoft . Removed those comments -- they were only used as sample inputs while illustrating the issue.

SergeyMenshykh (approved):

glahaye (approved):

PR #3419, opened 2023-11-08
Python: Added metadata specifying connection stems from MSK code

Jibola:
### Description

* MongoDB drivers are used in various flavors and languages. Making sure we exercise our due diligence in identifying the "origin" of the library calls makes it best to understand how our Atlas servers get accessed. 
* Additionally, bumped the motor version to 3.3.2 to support the DriverInfo tracking behavior.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #5232, opened 2024-02-28
Python: Added metadata specifying connection stems from MSK code

Jibola:
### Description

* MongoDB drivers are used in various flavors and languages. Making sure we exercise our due diligence in identifying the "origin" of the library calls makes it best to understand how our Atlas servers get accessed. 
* Additionally, bumped the motor version to 3.3.2 to support the DriverInfo tracking behavior.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #5269, opened 2024-03-01
.Net: Added metadata specifying connection stems from MSK code

BorisDog:
### Description

MongoDB drivers are used in various flavors and languages. Making sure we exercise our due diligence in identifying the "origin" of the library calls makes it best to understand how our Atlas servers get accessed.
Similar to [Python PR](https://github.com/microsoft/semantic-kernel/pull/3419).

### Contribution Checklist


markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.Memory.MongoDB/MongoDBMemoryStore.cs
```diff
@@ -219,6 +220,14 @@ protected virtual void Dispose(bool disposing)
     private static FilterDefinition<MongoDBMemoryEntry> GetFilterByIds(IEnumerable<string> ids) =>
         Builders<MongoDBMemoryEntry>.Filter.In(m => m.Id, ids);
 
+    private static MongoClientSettings GetMongoClientSettings(string connectionString)
+    {
+        var settings = MongoClientSettings.FromConnectionString(connectionString);
+        var skVersion = typeof(IMemoryStore).Assembly.GetName().Version.ToString();
+        settings.LibraryInfo = new LibraryInfo("Microsoft Semantic Kernel", skVersion);
```
dmytrostruk, line 227: nit: We can also use already existing code that provides library name and version, which we use for other integrations. It's used for HTTP requests, and I think it should be fine to use it here as well.
https://github.com/microsoft/semantic-kernel/blob/14a0881989e2bd23c75568fe196c705f9bb65f79/dotnet/src/InternalUtilities/src/Http/HttpHeaderConstant.cs#L18-L32

If in the future we will decide to change the logic to get library name or version, this place in MongoDB connector will be automatically resolved.

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.Memory.MongoDB/MongoDBMemoryStore.cs
BorisDog: Thank you for the suggestion @dmytrostruk,
Ideally it would be great to be consistent with [Python implementation](https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/connectors/memory/mongodb_atlas/mongodb_atlas_memory_store.py#L52).  So if that's the preferred way, we might want to make this change in both connectors.

prakul: @dmytrostruk we will really benefit from having consistency in the Python, C# versions of MSK (and  maybe Java & JS as well in the near future as well)

dmytrostruk: > Thank you for the suggestion @dmytrostruk,
> Ideally it would be great to be consistent with [Python implementation](https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/connectors/memory/mongodb_atlas/mongodb_atlas_memory_store.py#L52). So if that's the preferred way, we might want to make this change in both connectors.

@BorisDog In Python we also have the same value for user agent:
https://github.com/microsoft/semantic-kernel/blob/dfd866b9adf51397a1d5b12bb0a0545adad1002d/python/semantic_kernel/connectors/telemetry.py#L12-L14

But it looks like it was not used in this PR: https://github.com/microsoft/semantic-kernel/pull/5232. I agree about consistency, and it would be nice to keep user agent consistent across connectors and SK languages.

Tagging @moonbox3 for visibility.

prakul:
Thanks for reviewing the PR @dmytrostruk and @markwallace-microsoft!

PR #5275, opened 2024-03-02
.Net: Update classes so namespaces are consistent and match project location

markwallace-microsoft:
### Contribution Checklist


dotnet/src/Connectors/Connectors.HuggingFace/Core/TextEmbeddingRequest.cs
```diff
@@ -3,7 +3,7 @@
 using System.Collections.Generic;
 using System.Text.Json.Serialization;
 
-namespace Microsoft.SemanticKernel.Connectors.HuggingFace;
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
```
dmytrostruk, line 6: This class is public, and for public classes we want to keep namespaces easy and short. That's why for public classes the rule to "align namespace to the project location" doesn't apply. But in general, if this class is used as HTTP request schema, then I think it should be internal.

dotnet/src/Connectors/Connectors.HuggingFace/Core/TextEmbeddingResponse.cs
```diff
@@ -3,7 +3,7 @@
 using System;
 using System.Collections.Generic;
 
-namespace Microsoft.SemanticKernel.Connectors.HuggingFace;
+namespace Microsoft.SemanticKernel.Connectors.HuggingFace.Core;
```
dmytrostruk, line 6: Same comment as above.

dotnet/src/Connectors/Connectors.HuggingFace/Core/TextEmbeddingRequest.cs
markwallace-microsoft: The request/response classes for text generation are internal as is the `HuggingFaceClient` so I'll make these two internal also.

dmytrostruk (approved):

RogerBarreto:
Thanks for:
- Splitting the ServiceCollection in it's own class, the other connectors are doing in the same, this is a good move.
- Making internal the missing request/response `Models`, missed that.

REgarding the `namespace` change I don't fully agree, as this defeats the purpose of having specific modality `namespaces` where classes that could be in it are all mixed in `Core`.

RogerBarreto (approved):

PR #5276, opened 2024-03-02
.Net: Extend embeddings abstraction

Krzysztof318:
### Motivation and Context

ADR #5277

### Description

Extended abstraction to satisfy above criteria.

- [x] Update HuggingFace connector
- [ ] Update OpenAI connector
- [ ] Update GoogleVertexAI connector
- [x] Create ADR

### Contribution Checklist

- [ ] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [ ] All unit tests pass, and I have added new tests where possible
- [ ] I didn't break anyone :smile:

PR #5277, opened 2024-03-02
Embeddings abstraction ADR

Krzysztof318:
### Motivation and Context

Currently embeddings abstraction is limited. It doesn't allow passing metadata or additional execution settings.
For example: VertexAI embeddings model has optional parameters, returns metadata and generates embeddings from multi-modal content.

### Description

Related to #5276


docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+informed: ???
+---
+
+# Embeddings Generation in Semantic Kernel
+
+## Context and Problem Statement
+
+### General Information
+
+The current abstraction of generating embeddings only allows for generation from text data.\
```
dmytrostruk, line 17: That's a case for `ITextEmbeddingGenerationService` but `IEmbeddingGenerationService<TValue, TEmbedding>` is not coupled to particular data type. 

Today, if you create new connector that implements `IEmbeddingGenerationService<ImageContent, float>`, it should work as expected. But this interface doesn't support metadata in response and does not accept execution settings, but these are separate problems. So, maybe it will be good to re-phrase `Content and Problem Statement` section to avoid confusion.

docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+We want to enable the generation of embeddings from other types of data, such as images and others.\
+We also want to make it possible to parameterize the query.
+
+### Embeddings models
+
+OpenAI provides models that allow for generating embeddings only from textual data.\
+The models do not return metadata, and also do not allow for parameterizing the query.
```
dmytrostruk, line 24: It looks like OpenAI API for embeddings does return metadata (token usage) and does allow to pass additional parameters. From documentation:
image

docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+        IReadOnlyList<ReadOnlyMemory<TEmbedding>> data,
+        string? modelId = null,
+        object? innerContent = null,
+        IReadOnlyDictionary<string, object?>? metadata = null)
+        : base(innerContent, modelId, metadata)
+    {
+        this.Data = data;
+    }
+
+    public IReadOnlyList<ReadOnlyMemory<TEmbedding>> Data { get; set; }
```
dmytrostruk, line 90: In `EmbeddingContent` class `Data` property is `IReadOnlyList` but then in all provided options, interface method `GenerateEmbeddingsAsync` returns `IReadOnlyList<EmbeddingContent<TEmbedding>>`, so it's collection of collection of embedding. Not sure if it's needed. 

In `EmbeddingContent` class we can keep `Data` as `ReadOnlyMemory<TEmbedding>` and in interface return `IReadOnlyList<EmbeddingContent<TEmbedding>>`, which would be multiple responses with its own vector and metadata. I think it should be enough.

docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+}
+```
+Pros:
+- Allows for generating embeddings from different types of data.
+- Allows for parameterizing the query.
+- Allows for returning metadata.
+- Interface segregation.
+
+Cons:
+- No generic interface.
```
dmytrostruk, line 118: That's not always a disadvantage, but in this case, in theory, if we keep generic `IEmbeddingGenerationService<TValue, TEmbedding>` as it is today, but update `GenerateEmbeddingsAsync` method to return `IReadOnlyList<EmbeddingContent<TEmbedding>>` and add `PromptExecutionSettings` as optional parameter, we should resolve the disadvantage provided in this list and cover all required cases, is that correct? I think we should add this approach as another option in this ADR.

Then, if specific AI connector supports embeddings for texts and images, it could implement `IEmbeddingGenerationService` interface multiple times with different data types:
image

docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+```
+
+Pros:
+- Allows for generating embeddings from different types of data.
+- Allows for parameterizing the query.
+- Allows for returning metadata.
+
+Cons:
+- No generic interface.
+- One common interface for different types of data. Many models don't support all types of data.
```
dmytrostruk, line 163: That's correct, it would be nice to avoid this approach since connectors will be forced to implement functionality which they don't support. But it's still an option, so it's good that you mentioned it in ADR, thanks!

docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+
+Similar to option 2 or 3, but with a non-generic `EmbeddingContent` class.
+
+This option forces the conversion of embeddings data always to `double`.
+
+```csharp
+public class EmbeddingContent : KernelContent
+{
+    public EmbeddingContent(
+        IReadOnlyList<ReadOnlyMemory<float>> data,
```
dmytrostruk, line 175: nit: `Data` property uses `double`, but here it accepts `float`.

docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+
+Pros:
+- Simplicity.
+
+Cons:
+- We cannot return embeddings with different types.
+
+### Option 5 [Proposed] - One interface for embeddings generation with metadata and query parameters
+
+Similar to option 3 but with only one metod with KernelContent param.
```
dmytrostruk, line 196:
```suggestion
Similar to option 3 but with only one method with KernelContent param.
```

docs/decisions/0035-embedding-abstraction.md
```diff
@@ -0,0 +1,257 @@
...
+    {
+        this.Data = data;
+    }
+
+    public IReadOnlyList<ReadOnlyMemory<TEmbedding>> Data { get; set; }
+}
+
+public interface IEmbeddingGenerationService : IAIService
+{
+    Task<IReadOnlyList<EmbeddingContent<double>>> GenerateEmbeddingsAsync(
```
dmytrostruk, line 216: If we will have one interface for embeddings, that will return `EmbeddingContent<double>`, then we don't need to make `EmbeddingContent` as generic, because only one `double` type will be used. So, it looks like Option 4 and Option 5 can be merged together, please correct me if I'm wrong.

dmytrostruk (commented):
@Krzysztof318 Thanks for this ADR! I've added some comments.

docs/decisions/0035-embedding-abstraction.md
Krzysztof318: yeah, I did not notice my error 🙃

docs/decisions/0035-embedding-abstraction.md
Krzysztof318: @dmytrostruk 
Personally, I prefer the current option No 4 but I have also added a generic version as No 5.

Krzysztof318: @dmytrostruk And do we really need generic EmbeddingContent class? Do you see any cases where  `double` type can be insufficient?

Krzysztof318: https://github.com/microsoft/semantic-kernel/pull/5279#discussion_r1511330971 this has confused me a bit, maybe option 2 is best after all?

dmytrostruk (commented):
Hi @Krzysztof318 . After team discussion, we decided to proceed with combination of Option 1 and Option 2. Specifically, we would like to proceed with following:
1. Keep generic interface `IEmbeddingGenerationService<TValue>` where `TValue` should be inherited from `KernelContent` (e.g. `TextContent`, `ImageContent` etc). For usage simplification, we can add extension methods to accept `string` instead of `TextContent`, but base interface should accept type that is inherited from `KernelContent`.
2. Generic interface `IEmbeddingGenerationService<TValue>` should accept list of `TValue` as input, optional execution settings, optional kernel and optional cancellation token (in this specific order, which is the same as in `IChatCompletionService`).
3. No need in `TEmbedding` generic parameter to specify type of numerical representation (int, float, double). To represent embeddings we will use `float`, since `double` precision is not used at the moment and `double` will use more memory and have slower performance comparing to `float`. This is the decision that has been made across multiple .NET libraries including Azure SDK.
4. Since we will use `float` type to represent embeddings, we don't need to make `EmbeddingContent` as generic.
5. We still want to keep specific interfaces for text and image embedding generation: `ITextEmbeddingGenerationService` and `IImageEmbeddingGenerationService` that will be inherited from generic `IEmbeddingGenerationService<TValue>` interface. These interfaces should be primarily used by customers in their applications for better usability, while generic interface will serve the purpose of creating custom embedding generation service that works with other type of data when needed.

I hope that makes sense and let me know in case you have any questions. Thanks!

Krzysztof318:
@dmytrostruk sounds good, I will update the ADR later on 😄

Krzysztof318:
@dmytrostruk can we finish this adr? What else you thoughts have?

dmytrostruk:
> @dmytrostruk can we finish this adr? What else you thoughts have?

Hi @Krzysztof318 ! 
Based on latest discussions we agreed on the following:
- Keep generic parameter for Embedding type (e.g. int, float or double) for future extensibility. Although, currently it has `unmanaged` constraint, so we should remove it. That will mean to keep `EmbeddingContent<TEmbedding>` generic as well.
- Data input should be `IEnumerable<TContent>` rather than `IList<TContent>` to avoid forcing users to perform `ToList` in case they have enumerable.
- Generic type for content should have a constraint to be derived from `KernelContent` and it should be contravariant (e.g. `IEmbeddingGenerationService<in TContent, TEmbedding>`. It will allow users to cast `IEmbeddingGenerationService<KernelContent, float>` to more specific `IEmbeddingGenerationService<TextContent, float>`. 

So, the final abstraction should look like this:
```csharp
public interface IEmbeddingGenerationService<in TContent, TEmbedding> : IAIService
    where TContent : KernelContent
{
    Task<IReadOnlyList<EmbeddingContent<TEmbedding>>> GenerateEmbeddingsAsync(
        IEnumerable<TContent> data,
        Kernel? kernel = null,
}

public class EmbeddingContent<TEmbedding> : KernelContent
{
    public EmbeddingContent(
        ReadOnlyMemory<TEmbedding> data,
        string? modelId = null,
        object? innerContent = null,
    {
        this.Data = data;
    }

    public ReadOnlyMemory<TEmbedding> Data { get; set; }
}
```

Please let me know if you have any questions. Thank you!

Krzysztof318:
@dmytrostruk It's ok, I have updated ADR but added only sample under "decision outcome" section. Look please and update ADR  ; pros, cons, information because you are better oriented what should be there written.

Krzysztof318:
Hi @dmytrostruk when will you have time to finish that?

dmytrostruk:
> Hi @dmytrostruk when will you have time to finish that?

Hi @Krzysztof318 , this issue should be prioritized based on other memory-related work which is currently in progress.
cc: @markwallace-microsoft @matthewbolanos

markwallace-microsoft:
@Krzysztof318 The plan is to include this as part of the feature graduation work which is being tracked here #8094

PR #5279, opened 2024-03-03
Text classification ADR

Krzysztof318:
ADR for text classification models

Related to #5061


RogerBarreto:
@Krzysztof318 Thanks for putting this ADR, very helpful and give some good options.

I have some observations:

- The abstractions methods if used for Classification Services should follow our naming convention:

   ``` 
   Get<KernelContent>sAsync(KernelContent content, PromptExecutionSettings executionSettings?, Kernel?, CancellationToken)
   ``` 
- `ClassificationContent` seems like the content that will be classified, but in this example I would suggest `ClassifiedContent` for the abstraction.

For the abstractions, analyzing both models, an abstraction that supports HuggingFace might be sufficient to cover OpenAI.

Ex:
Open AI would look like:
```
[
   { 
        "label": "flagged",
        "score": 1|0 (true|false)
   },
   {
        "label": "sexual",
        "score": 1.2282071e-6
   },
   ...
}]
```

Abstraction proposal:
```
    public ClassifiedContent : KernelContent
    {
        Dictionary<string, double> ScoredCategories
    }
```

- It still not clear why you suggesting the property `KernelContent ClassifiedContent` as part of the abstraction.

Krzysztof318:
@RogerBarreto I addresses your comment and I've added one more option to consider.
About option three we can also create special class like Category (in ClassifiedContent class) with properties Flagged and Score instead of this `IReadOnlyDictionary<string, (bool, double)>`

docs/decisions/0036-text-classification.md
```diff
@@ -0,0 +1,270 @@
...
+```csharp
+public interface IClassificationService : IAIService
+{
+    Task<IReadOnlyList<ClassifiedContent>> GetClassifiedContentsAsync(
+        IEnumerable<KernelContent> contents,
+        PromptExecutionSettings? executionSettings = null,
+        Kernel? kernel = null,
+        CancellationToken cancellationToken = default);
+
+    IList<Type> SupportedContents { get; }
```
RogerBarreto, line 154: This might not be needed for now in the interface as we plan to add in the mid/long term a generics approach to `IService<TIn,TOut>`. 

For the moment we may stick `ITextClassificationService` with the method only to retrieve the contents for `string` as input to start with as this is the current example and we might have `IImageClassificationService` and so forth. 

We had an interesting conversation on the abstractions of Audio abstractions and is was very conclusive the benefits of having `Modal` specific interfaces.

Krzysztof318: Do I understand correctly that the `IService<TIn,TOut>` interfaces will only be wrappers to the current interfaces?

Krzysztof318:
@RogerBarreto ??

RogerBarreto:
Closing temporarily as we are going to refactor and touch on this feature in the future, after we gradutate current contents.

Issue #5264, opened 2024-03-01
.Net upon upgrading to 1.5 function calling fails


joslat:
**Describe the bug**

**To Reproduce**
With packages Microsoft.SemanticKernel & Microsoft.SemanticKernel.Connectors.OpenAI  1.0 up to 1.4 with the following code there is no issue.
Upon upgrading to 1.5 I get the exception: 
```
Unhandled exception. System.NotSupportedException: Runtime type 'Microsoft.SemanticKernel.Connectors.OpenAI.OpenAIChatMessageContent' is not supported by polymorphic type 'Microsoft.SemanticKernel.KernelContent'. Path: $.
 ---> System.NotSupportedException: Runtime type 'Microsoft.SemanticKernel.Connectors.OpenAI.OpenAIChatMessageContent' is not supported by polymorphic type 'Microsoft.SemanticKernel.KernelContent'.
   at System.Text.Json.ThrowHelper.ThrowNotSupportedException_RuntimeTypeNotSupported(Type baseType, Type runtimeType)
   at System.Text.Json.Serialization.Metadata.PolymorphicTypeResolver.TryGetDerivedJsonTypeInfo(Type runtimeType, JsonTypeInfo& jsonTypeInfo, Object& typeDiscriminator)
   at System.Text.Json.Serialization.JsonConverter.ResolvePolymorphicConverter(Object value, JsonTypeInfo jsonTypeInfo, JsonSerializerOptions options, WriteStack& state)
   at System.Text.Json.Serialization.JsonConverter`1.TryWrite(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
   at System.Text.Json.Serialization.JsonConverter`1.WriteCore(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
   --- End of inner exception stack trace ---
   at System.Text.Json.ThrowHelper.ThrowNotSupportedException(WriteStack& state, NotSupportedException ex)
   at System.Text.Json.Serialization.JsonConverter`1.WriteCore(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo`1.Serialize(Utf8JsonWriter writer, T& rootValue, Object rootValueBoxed)
[...]

PR #5301, opened 2024-03-04
.Net: Fix serialization of OpenAIChatMessageContent type

SergeyMenshykh:
### Motivation and Context
This PR fixes the https://github.com/microsoft/semantic-kernel/issues/5264 issue that prevents auto calling of kernel functions created from prompt. 

### Description  
1. The PR checks whether the function execution result type is `ChatMessageContent` or its subtype and returns the `ChatMessageContent.Content` property value if that's the case. Otherwise, if the check was not provided, there would be an attempt to serialize the class and because the `OpenAIChatMessageContent` is not and can't be registered for polymorphic serialization on the [KernelContent](https://github.com/microsoft/semantic-kernel/blob/d65d3b617b848d33cc1816b1756c3cdb65f4f337/dotnet/src/SemanticKernel.Abstractions/Contents/KernelContent.cs#L18) class, the serialization would fail with the exception - _"System.NotSupportedException: Runtime type 'Microsoft.SemanticKernel.Connectors.OpenAI.OpenAIChatMessageContent' is not supported by polymorphic type 'Microsoft.SemanticKernel.KernelContent'."_ Alternatively, instead of the type check above, the function result serializer could be configured with a custom JsonTypeInfoResolver to handle polymorphic serialization of the `OpenAIChatMessageContent` class. However, in this case, all the public properties - ModelId, Metadata, Content, Items, Encoding, etc. of the `OpenAIChatMessageContent` class would be unnecessarily serialized.  
2. This PR also introduces the new `ToolCallBehavior.FunctionResultSerializerOptions` to manage the serialization behavior of function execution results. It can be used to provide a JsonTypeInfoResolver for the polymorphic serialization of a custom subclass of the `KernelContent` class.

### Contribution Checklist


markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1058,4 +1058,31 @@ private void CaptureUsageDetails(CompletionsUsage usage)
...
+        if (functionResult is string stringResult)
+        {
+            return stringResult;
+        }
+
+        // This is an optimization to use ChatMessageContent content directly  
+        // without unnecessary serialization of the whole message content class.  
+        if (functionResult is ChatMessageContent chatMessageContent)
+        {
+            return chatMessageContent.Content;
```
dmytrostruk, line 1085: nit: I'm wondering if we should use `chatMessageContent.ToString()` instead of using `Content` property directly. `ToString` override already use `Content` property, and in case if we would like to change the string representation of this type in the future by adding some additional properties, this place will be covered automatically:
https://github.com/microsoft/semantic-kernel/blob/8deaaf4e8285695f78d34360e8c7433f7da1d862/dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs#L156-L159

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
```diff
@@ -81,6 +82,11 @@ public static ToolCallBehavior RequireFunction(OpenAIFunction function, bool aut
         return new RequiredFunction(function, autoInvoke);
     }
 
+    /// <summary>
+    /// Options to control function call result serialization behavior.
+    /// </summary>
+    public JsonSerializerOptions? FunctionResultSerializerOptions { get; set; }
```
dmytrostruk, line 88: nit: Just double-checking on naming here, because this property is public. Since this is `ToolCallBehaviour` class, should we consider `ToolCallResultSerializerOptions`? I'm wondering if it won't be confusing with our `FunctionResult` type, because it's actual value of `FunctionResult` that is serialized, but not `FunctionResult` class itself.

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
SergeyMenshykh: _Since this is ToolCallBehaviour class, should we consider ToolCallResultSerializerOptions?_ I initially thought about it, but then I saw that the ToolCallBehavior class has a function-related member - the AllowAnyRequestedKernelFunction property, and thought that it would be okay to add another function-related property. On the other hand, we probably should avoid adding child class specifics to parent classes. So, the property has been renamed to ToolResultSerializerOptions.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: fixed

dmytrostruk (approved):

dmytrostruk (approved):

PR #5324, opened 2024-03-05
.Net: Python: Sk autogen

jzagelbaum:
### Motivation and Context

 Why is this change required? 
Extended 12-sk-autogen-agents.ipynb to include example of integrating prompt flow as a plug in. So now we have sk + autogen + pf!
 
What problem does it solve?
  Bringing sk into promptflow limits alot of the planner capabilities. I think a better design decision is the other way around. Now we can use sk and autogen to orchestrate and manage extensibility while leveraging prompt flow for evaluations and inference logging

What scenario does it contribute to? 
Dogfood frameworks to show how they can be used composably and thoughtfully to accelerate Gen AI use cases while maintaining responsible AI practices

### Description

Hopefully updated workbook is self describing enough


moonbox3:
Hi @jzagelbaum, thank you for your contribution. I'm not familiar with this feature branch. It's currently showing 2,086 files were changed -- many dotnet related. Is there a reason you're working in this branch? Thanks for the context.

moonbox3 (changes requested):
Let's please get the branch/PR in a better state so we know we're reviewing the correct files.

markwallace-microsoft:
@jzagelbaum Something has gone wrong when you created this PR and it's resulted in changes to over 2K files and 110K lines of code. Please re-create the PR.

PR #5415, opened 2024-03-11
.Net: Gemini/functions update2

Krzysztof318:
Gemini/functions update2


dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiContent.cs
```diff
@@ -15,8 +15,7 @@ internal sealed class GeminiContent
     /// Ordered Parts that constitute a single message. Parts may have different MIME types.
     /// </summary>
     [JsonPropertyName("parts")]
-    [JsonRequired]
-    public IList<GeminiPart> Parts { get; set; } = null!;
+    public IList<GeminiPart>? Parts { get; set; }
```
RogerBarreto, line 18: Add some tests to ensure when the response comes without parts, it can serialize correctly.

image

Krzysztof318: done

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/ClientBase.cs
```diff
@@ -85,7 +85,7 @@ protected static T DeserializeResponse<T>(string body)
         }
         catch (JsonException exc)
         {
-            throw new KernelException("Unexpected response from model", exc)
+            throw new KernelException("Unexpected response from model")
```
RogerBarreto, line 88: Please, pass the inner exception, is a good practice 👍🏻

Krzysztof318: But you don't want to expose details of Json validation like we talked on discord. or I misunderstood that.

RogerBarreto: Was a misunderstand. I can clarify.

Before if we had an exception it would be on a 200 OK response, which turns out to be something expected. 

Now that the parts is optional, that exception will not happen and we will be compatible to that result.

With that is ok to have the JsonException passed as InnerException for things that are indeed unexpected.

RogerBarreto (approved):

PR #5422, opened 2024-03-11
.Net: Create clone of KernelFunction when it is added to a KernelPlugin

markwallace-microsoft:
### Contribution Checklist


dotnet/src/SemanticKernel.Core/Functions/DefaultKernelPlugin.cs
```diff
@@ -30,7 +30,9 @@ internal DefaultKernelPlugin(string name, string? description, IEnumerable<Kerne
             foreach (KernelFunction f in functions)
             {
                 Verify.NotNull(f, nameof(functions));
-                this._functions.Add(f.Name, f);
+
+                var cloned = f.CreateForPlugin(name);
+                this._functions.Add(cloned.Name, cloned);
```
RogerBarreto, line 35: If `cloned.Name` is always the same `name` you can avoid an extra allocation of `cloned`.

```suggestion
                this._functions.Add(name, f.CreateForPlugin(name));
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -352,6 +379,12 @@ await using (enumerator.ConfigureAwait(false))
         }
     }
 
+    /// <summary>
+    /// Creates a new <see cref="KernelFunction"/> object that is a copy of the current instance.
+    /// <param name="pluginName">The name of the plugin this function instance will be added to.</param>
+    /// </summary>
+    public abstract KernelFunction CreateForPlugin(string pluginName);
```
RogerBarreto, line 386: 1. `<param>` should be outside of `<summary>`
2. `Clone` seems good to me, if we add a `remarks` on the xmldoc...

```suggestion
    /// <summary>
    /// Creates a new <see cref="KernelFunction"/> object that is a copy of the current instance.
will be added to.</param>
    /// </summary>
    /// <remarks>
    /// This should be used only for specific scenarios where plugin mutability is intended.
    /// </remarks>
    /// <param name="pluginName">The name of the plugin this function instance will be added to.</param>    
    public abstract KernelFunction Clone(string pluginName);
```

dotnet/src/SemanticKernel.UnitTests/Functions/KernelPluginTests.cs
RogerBarreto, line 1: Suggest adding a test where the function you add from one plugin is not the same instance you added to another.

```
var function1 = ...
var plugin1 = KernelPluginFactory.CreateFromFunctions("name", "description", new[] { function1 });

var plugin2 = KernelPluginFactory.CreateFromFunctions("name", "description", new[] { function1 });

Assert.NotSame(plugin1[function1.Name],plugin2[function1.Name]);

**Review by RogerBarreto — APPROVED** [review]
Some comments, overall CLGTM

File: dotnet/src/SemanticKernel.Core/Functions/DefaultKernelPlugin.cs (hunk 1 above)
Comment (markwallace-microsoft) [reply dotnet/src/SemanticKernel.Core/Functions/DefaultKernelPlugin.cs:35]: `cloned.Name` is the name of the function and `name` is the name of the plugin

**Review by SergeyMenshykh — APPROVED** [review]

File: dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs [hunk 3]
```diff
@@ -352,6 +379,17 @@ await using (enumerator.ConfigureAwait(false))
...
+    /// <summary>
+    /// Creates a new <see cref="KernelFunction"/> object that is a copy of the current instance
+    /// but the <see cref="KernelFunctionMetadata"/> has the plugin name set.
+    /// </summary>
+    /// <param name="pluginName">The name of the plugin this function instance will be added to.</param>
+    /// <remarks>
+    /// This method should only be used to create a new instance of a <see cref="KernelFunction"/> when adding
+    /// a function to a <see cref="KernelPlugin"/>.
+    /// </remarks>
+    public abstract KernelFunction CreateForPlugin(string pluginName);
```
stephentoub, line 391: This is more of a Clone rather than Create method, right? Maybe `CloneWithPluginName` or something along those lines?

stephentoub, line 391: Should we follow the "core" pattern here as well, e.g. have a non-virtual public method and a protected abstract method? It would allow for things like:
- sharing any validation desired for the pluginName
- validating that the cloned KernelFunction is non-null and has the same name as input KernelFunction

etc., if we care

stephentoub, line 391: I assume we don't support using version N of Microsoft.SemanticKernel.Abstractions but version N - 1 of Microsoft.SemanticKernel.Core. If we did, adding an abstract method to KernelFunction would be a breaking change, since even though the only things that can derive from it are things with internal access, a previous version of the Core assembly wouldn't have the necessary overrides.

I don't think this is a real problem, just calling it out for completeness.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
```diff
@@ -217,6 +218,21 @@ await foreach (var content in asyncReference.ConfigureAwait(false))
...
+    public override KernelFunction CreateForPlugin(string pluginName)
+    {
+        return new KernelFunctionFromPrompt(
+            this._promptTemplate,
+            this.Name,
+            pluginName,
+            this.Description,
+            this.Metadata.Parameters,
+            this.Metadata.ReturnParameter,
+            this.ExecutionSettings.ToDictionary(kv => kv.Key, kv => kv.Value),
```
stephentoub, line 231: Is there a reason this needs to clone the dictionary?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
markwallace-microsoft: That was my original name and no-one liked it but I will revert this.

markwallace-microsoft: Yeah, we've decided to keep things simple in our versioning story and require that all versions of all packages are the same. This is the only combination we test and will officially support.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
markwallace-microsoft: Because the source is a ReadOnlyDictionary and we need to pass in a regular Dictionary.

stephentoub: Is it a `ReadOnlyDictionary` or an `IReadOnlyDictionary`? If the latter, it could be:
```C#
this.ExecutionSettings as Dictionary<> ?? this.ExecutionSettings.ToDictionary(kv => kv.Key, kv => kv.Value),
```
but it's not a big deal.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
markwallace-microsoft: I started to try out these changes and I don't think this is necessary. There is real value in the Invoke scenario because we do all of our telemetry and filter stuff.

dmytrostruk: In this case, how about just `Clone`?
In the future it will be easier to add similar overload that will accept something different than single string (if there will be such requirement).

For example, instead of this:
```csharp
CloneWithPluginName(string pluginName);
CloneWithParameters(string pluginName, IReadOnlyList<KernelParameterMetadata> parameters);
```

It will be like this:
```csharp
Clone(string pluginName);
Clone(string pluginName, IReadOnlyList<KernelParameterMetadata> parameters);
```

Feel free to skip if you think it doesn't make sense :)

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromPrompt.cs
markwallace-microsoft: Yes, it's `IReadOnlyDictionary`

SergeyMenshykh (approved):

PR #5443, opened 2024-03-12
.Net VertexAI - bearer provider

Krzysztof318:
Added bearer provider `Func<string>` to vertexAI.

@RogerBarreto


RogerBarreto:
Thanks for that, will be very useful in common context of services that need to rely on expiring bearerKeys.

dotnet/samples/KernelSyntaxExamples/Example97_GeminiVision.cs
```diff
@@ -81,9 +81,9 @@ public async Task VertexAIAsync()
         //
         //   gcloud auth print-access-token
         //
-        // Above code pass bearer key as string, it is not recommended to pass bearer key as string in production code,
+        // Above code pass bearer key as string, it is not recommended way in production code,
         // especially if IChatCompletionService will be long lived, tokens generated by google sdk lives for 1 hour.
-        // You should provide token bearer key factory, which will be used to generate token on demand:
+        // You should provide bearer key factory, which will be used to generate token on demand:
```
RogerBarreto, line 86:
```suggestion
        // You should use bearer key provider, which will be used to generate token on demand:
```

Krzysztof318: yeah my fault... I think "provider" better conveys the meaning, so I am going to fix comments.

RogerBarreto (approved):

PR #5347, opened 2024-03-06
.Net: Reduce number of experimental flags used

markwallace-microsoft:
### Motivation and Context

Reduce the burden on developers who want to use SK experimental features by reducing the number of `SKEXP` flags to the following list:

| SKEXP​ | Experimental Features Category​​ |
|-------|--------------------------------|
| SKEXP0001 | Semantic Kernel core |
| SKEXP0010 | OpenAI and Azure OpenAI services |
| SKEXP0020 | Memory abstractions |
| SKEXP0040 | Function types |
| SKEXP0050 | Out-of-the-box plugins |
| SKEXP0060 | Planners |
| SKEXP0070 | AI connectors |
| SKEXP0100 | Advanced Semantic Kernel features |

### Description

Closes #5242 

### Contribution Checklist

- [ ] The code builds clean without any errors or warnings
- [ ] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [ ] All unit tests pass, and I have added new tests where possible
- [ ] I didn't break anyone :smile:

PR #5463, opened 2024-03-13
.Net Adding Experimental Gemini Connector to Main.

RogerBarreto:
@Krzysztof318, thanks for the great work and dedication on this feature.

### Motivation and Context

Adding the Gemini Connector to `main` as a new experimental package.

### Contribution Checklist


dotnet/docs/EXPERIMENTS.md
```diff
@@ -42,6 +42,7 @@ You can use the following diagnostic IDs to ignore warnings or errors for a part
 - SKEXP0030: MongoDB memory connector
 - SKEXP0031: Pinecone memory connector
 - SKEXP0032: Postgres memory connector
+- SKEXP0033: GoogleVertexAI Connector
```
stephentoub, line 45: How will this relate to https://github.com/microsoft/semantic-kernel/pull/5347?

dotnet/docs/EXPERIMENTS.md
```diff
@@ -57,7 +58,7 @@ You can use the following diagnostic IDs to ignore warnings or errors for a part
 - SKEXP0052: Memory plugins
 - SKEXP0053: Microsoft 365 plugins
 - SKEXP0054: Web plugins
-- SKEXP0055: Text chunkcer plugin
+- SKEXP0055: Text chunker plugin
```
stephentoub, line 61: Pre-existing, but "Text chunker" isn't a plugin

dotnet/samples/KernelSyntaxExamples/Example94_GeminiGetModelResult.cs
```diff
@@ -0,0 +1,68 @@
...
+        //         },
+        //         location: TestConfiguration.VertexAI.Location,
+        //         projectId: TestConfiguration.VertexAI.ProjectId)
+
+        // Create function
+        const string FunctionDefinition = "Hi, give me 5 book suggestions about: {{$input}}";
+        KernelFunction myFunction = kernel.CreateFunctionFromPrompt(FunctionDefinition);
+
+        // Invoke function through kernel
+        FunctionResult result = await kernel.InvokeAsync(myFunction, new() { ["input"] = "travel" });
```
stephentoub, line 59: Rather than complicating the example with CreateFunctionFromPrompt, this could just use InvokePromptAsync, right?

dotnet/docs/EXPERIMENTS.md
RogerBarreto: AS soon the feature branch is updated with the latest main changes, this will be fixed.

dotnet/samples/KernelSyntaxExamples/Example95_GeminiTextGeneration.cs
```diff
@@ -0,0 +1,161 @@
...
+        string geminiProject = TestConfiguration.VertexAI.ProjectId;
+
+        if (geminiApiKey is null || geminiModelId is null || geminiLocation is null || geminiProject is null)
+        {
+            this.WriteLine("Gemini vertex ai credentials not found. Skipping example.");
+            return;
+        }
+
+        Kernel kernel = Kernel.CreateBuilder()
+            .AddVertexAIGeminiTextGeneration(
```
stephentoub, line 53: What's the difference between GoogleAIGemini and VertexAIGemini?

dotnet/samples/KernelSyntaxExamples/Example99_GeminiEmbeddingGeneration.cs
```diff
@@ -0,0 +1,301 @@
...
+    private async Task RunSimpleSampleAsync(Kernel kernel)
+    {
+        this.WriteLine("== Simple Sample: Generating Embeddings ==");
+
+        // Obtain an embedding generator.
+        var embeddingGenerator = kernel.GetRequiredService<ITextEmbeddingGenerationService>();
+
+        var generatedEmbeddings = await embeddingGenerator.GenerateEmbeddingAsync("My name is Andrea");
+        this.WriteLine($"Generated Embeddings count: {generatedEmbeddings.Length}, " +
+                       $"First five: {string.Join(", ", generatedEmbeddings[..5].ToArray())}...");
```
stephentoub, line 127: This ToArray should not be necessary

dotnet/samples/KernelSyntaxExamples/Example99_GeminiEmbeddingGeneration.cs
```diff
@@ -0,0 +1,301 @@
...
+END FACTS
+
+Question: {{$input}}
+
+Answer:
+";
+
+        var aboutMeOracle = kernel.CreateFunctionFromPrompt(RecallFunctionDefinition, new GeminiPromptExecutionSettings() { MaxTokens = 1000 });
+
+        result = await kernel.InvokeAsync(aboutMeOracle, new()
```
stephentoub, line 255: I imagine this was just copied from some existing sample, but it'd be nice if we strived to make these examples as simple as possible, e.g. if the sample isn't about creating a KernelFunction and InvokePromptAsync would otherwise suffice, just use InvokePromptAsync.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/ClientBase.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        try
+        {
+            T? deserializedResponse = JsonSerializer.Deserialize<T>(body);
+            if (deserializedResponse is null)
+            {
+                throw new JsonException("Response is null");
+            }
+
+            return deserializedResponse;
```
stephentoub, line 81:
```suggestion
            return JsonSerializer.Deserialize<T>(body) ?? throw new JsonException("Response is null");
```
?

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -0,0 +1,611 @@
...
+    /// we could end up in an infinite loop. This const is a backstop against that happening. We should never come close
+    /// to this limit, but if we do, auto-invoke will be disabled for the current flow in order to prevent runaway execution.
+    /// With the current setup, the way this could possibly happen is if a prompt function is configured with built-in
+    /// execution settings that opt-in to auto-invocation of everything in the kernel, in which case the invocation of that
+    /// prompt function could advertise itself as a candidate for auto-invocation. We don't want to outright block that,
+    /// if that's something a developer has asked to do (e.g. it might be invoked with different arguments than its parent
+    /// was invoked with), but we do want to limit it. This limit is arbitrary and can be tweaked in the future and/or made
+    /// configurable should need arise.
+    /// </remarks>
+    private const int MaxInflightAutoInvokes = 5;
```
stephentoub, line 46: Separately, we should really find a way to unify all of the function calling logic.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -0,0 +1,611 @@
...
+    }
+
+    /// <inheritdoc/>
+    public async Task<IReadOnlyList<ChatMessageContent>> GenerateChatMessageAsync(
+        ChatHistory chatHistory,
+        PromptExecutionSettings? executionSettings = null,
+        Kernel? kernel = null,
+        CancellationToken cancellationToken = default)
+    {
+        var state = ValidateInputAndCreateChatCompletionState(chatHistory, kernel, executionSettings);
```
stephentoub, line 146: Does this have the same issue addressed in https://github.com/microsoft/semantic-kernel/pull/5376?

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -0,0 +1,611 @@
...
+            // If we don't want to attempt to invoke any functions, just return the result.
+            yield return this.GetStreamingChatContentFromChatContent(messageContent);
+        }
+    }
+
+    private async Task ProcessFunctionsAsync(ChatCompletionState state, CancellationToken cancellationToken)
+    {
+        this.Logger.LogDebug("Tool requests: {Requests}", state.LastMessage!.ToolCalls!.Count);
+        this.Logger.LogTrace("Function call requests: {FunctionCall}",
+            string.Join(", ", state.LastMessage.ToolCalls.Select(ftc => ftc.ToString())));
```
stephentoub, line 269: This logging should be guarded by a check for IsEnabled. It's doing non-trivial work at the call site even if LogLevel.Trace isn't enabled. Same goes for any other similar call sites.

dotnet/samples/KernelSyntaxExamples/Example95_GeminiTextGeneration.cs
Krzysztof318: Both GoogleAI and Vertex supports gemini models.
These are different endpoints. GoogleAI is simple platform for customer. But VertexAI has more capabilities.
There are slight differences between them.

Instead, embeddings endpoints vary more.

More details are here https://cloud.google.com/vertex-ai/generative-ai/docs/migrate/migrate-google-ai#google-ai

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
Krzysztof318: Okay, I didn't know that not checking enabled LogLevel has performance impact.

Krzysztof318: done

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/ClientBase.cs
Krzysztof318: done

dotnet/samples/KernelSyntaxExamples/Example99_GeminiEmbeddingGeneration.cs
Krzysztof318: done

dotnet/samples/KernelSyntaxExamples/Example99_GeminiEmbeddingGeneration.cs
Krzysztof318: done

dotnet/samples/KernelSyntaxExamples/Example95_GeminiTextGeneration.cs
```diff
@@ -0,0 +1,161 @@
...
+        await StreamingFunctionFromPromptAsync(kernel);
+    }
+
+    private async Task StreamingFunctionFromPromptAsync(Kernel kernel)
+    {
+        this.WriteLine("======== Streaming Function From Prompt ========");
+
+        string prompt = "Describe what is GIT and why it is useful. Use simple words. Description should be long.";
+        var function = kernel.CreateFunctionFromPrompt(prompt);
+        await foreach (string text in kernel.InvokeStreamingAsync<string>(function,
```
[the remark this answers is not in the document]
Krzysztof318: done

dotnet/samples/KernelSyntaxExamples/Example94_GeminiGetModelResult.cs
Krzysztof318: done

dotnet/docs/EXPERIMENTS.md
Krzysztof318: not related after push

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
```diff
@@ -0,0 +1,238 @@
...
+using Microsoft.SemanticKernel.ChatCompletion;
+using Microsoft.SemanticKernel.Text;
+
+namespace Microsoft.SemanticKernel.Connectors.GoogleVertexAI;
+
+/// <summary>
+/// Represents the settings for executing a prompt with the Gemini model.
+/// </summary>
+[JsonNumberHandling(JsonNumberHandling.AllowReadingFromString)]
+public sealed class GeminiPromptExecutionSettings : PromptExecutionSettings
```
stephentoub, line 18: I was going to comment here that there's no support for generically specifying a function call behavior via the base PromptExecutionSettings, and then I realized that was removed as part of https://github.com/matthewbolanos/semantic-kernel/commit/d32802e416e898e6e82a60b0ab13a1a5ca5839ea#diff-2806827066f872cf10cc8ac3d6ff1989abc4971d08a953b341f0698b8429f65aL75-L82. @markwallace-microsoft, was that intentional? This ties in with the work item to come up with a generalized function calling abstraction / mechanism.

stephentoub (commented):
Thanks for working on this. Great to see the progress.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiTokenCounterClient.cs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /// <summary>
+    /// Counts the number of tokens asynchronously.
+    /// </summary>
+    /// <param name="prompt">The prompt to count tokens from.</param>
+    /// <param name="executionSettings">Optional settings for prompt execution.</param>
+    /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+    /// <returns>The number of tokens.</returns>
+    public async Task<int> CountTokensAsync(
```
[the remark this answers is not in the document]
Krzysztof318: This client is not used anywhere. We currently don't have appropriate service (interface) for that. But I left implementation for future and maybe internal usage.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFinishReason.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// <summary>
+    /// An enumerable collection of GeminiFinishReason.
+    /// </summary>
+    public static IEnumerable<GeminiFinishReason> GetAll() =>
+        typeof(GeminiFinishReason).GetFields(BindingFlags.Public |
+                                             BindingFlags.Static |
+                                             BindingFlags.DeclaredOnly)
+            .Select(f => f.GetValue(null))
+            .Cast<GeminiFinishReason>();
```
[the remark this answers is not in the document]
Krzysztof318: This method is not used by connector, but consumer can use it. I took it as a good pattern from https://learn.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/enumeration-classes-over-enum-types#implement-an-enumeration-base-class

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFunction.cs
```diff
@@ -0,0 +1,188 @@
...
+                { "properties", properties },
+            };
+        }
+
+        return new GeminiTool.FunctionDeclaration
+        {
+            Name = this.FullyQualifiedName,
+            Description = this.Description ?? throw new InvalidOperationException(
+                $"Function description is required. Please provide a description for the function {this.FullyQualifiedName}."),
+            Parameters = JsonSerializer.SerializeToNode(resultParameters),
```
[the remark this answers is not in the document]
Krzysztof318: I don't get it... could you elaborate?

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFunctionToolCall.cs
```diff
@@ -0,0 +1,83 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.Json;
+using Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;
```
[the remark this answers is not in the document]
Krzysztof318: yep all internals are in Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;
public in Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;
public Extensions in Microsoft.SemanticKernel;

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiMetadata.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.SemanticKernel.Connectors.GoogleVertexAI;
+
+/// <summary>
+/// Represents the metadata associated with a Gemini response.
+/// </summary>
+public sealed class GeminiMetadata : ReadOnlyDictionary<string, object?>
```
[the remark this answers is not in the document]
Krzysztof318: No, but HuggingFace has equivalent. This is convenient way to access metadata and fully compatible with actual metadata type.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFunctionToolCall.cs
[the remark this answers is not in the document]
stephentoub: > public in Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;

You mean "public in public in Microsoft.SemanticKernel.Connectors.GoogleVertexAI" (without the "Core"), right?

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiSafetyRating.cs
```diff
@@ -0,0 +1,134 @@
...
+    public GeminiSafetyProbability(string label)
+    {
+        Verify.NotNullOrWhiteSpace(label, nameof(label));
+        this.Label = label;
+    }
+
+    /// <summary>
+    /// An enumerable collection of GeminiSafetyProbability.
+    /// </summary>
+    public static IEnumerable<GeminiSafetyProbability> GetAll() =>
```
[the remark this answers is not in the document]
Krzysztof318: yeah my fault, should use properties, I explained purpose this method in upper comment.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -0,0 +1,612 @@
...
+
+        for (state.Iteration = 1; ; state.Iteration++)
+        {
+            using var httpRequestMessage = this.CreateHttpRequest(state.GeminiRequest, this._chatStreamingEndpoint);
+            using var response = await this.SendRequestAndGetResponseImmediatelyAfterHeadersReadAsync(httpRequestMessage, cancellationToken)
+                .ConfigureAwait(false);
+            using var responseStream = await response.Content.ReadAsStreamAndTranslateExceptionAsync()
+                .ConfigureAwait(false);
+
+            foreach (var messageContent in this.GetStreamingChatMessageContentsOrPopulateStateForToolCalling(state, responseStream))
```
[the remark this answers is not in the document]
Krzysztof318: GetStreamingChatMessageContentsOrPopulateStateForToolCalling this method is synchronous, if you follow entire way to GeminiStreamJsonParser you will know why I didn't make this async.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiSafetyRating.cs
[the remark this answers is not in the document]
stephentoub: What would the scenario be for having it? My thinking currently is that the fact that there are no built-in consumers of this and no tests that caught that this is currently producing an empty enumerable suggests we don't need it and that we should wait until there is a demonstrated need before providing it.

_If_ we keep it, we shouldn't use reflection to implement it.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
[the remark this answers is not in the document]
stephentoub: I actually started with GeminiStreamJsonParser, saw that it was synchronously reading character by character, which made me nervous, and I worked my way back here to say that it is indeed then doing synchronous I/O as part of an asynchronous method. We don't want to do that.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/StreamJsonParser.cs
```diff
@@ -0,0 +1,145 @@
...
+        while (ExtractNextJsonObject(reader, validateJson) is { } json)
+        {
+            yield return json;
+        }
+    }
+
+    private static string? ExtractNextJsonObject(TextReader reader, bool validateJson)
+    {
+        JsonParserState state = new();
+        while ((state.CharacterInt = reader.Read()) != -1)
```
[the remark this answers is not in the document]
Krzysztof318: It is quite universal parser. Gemini returns chunk as fully json objects. Each of chunks sometimes contain one or even few lines of text.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Extensions/VertexAIMemoryBuilderExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <param name="location">The location to process the request</param>
+    /// <param name="projectId">Your project ID</param>
+    /// <param name="httpClient">The optional custom HttpClient.</param>
+    /// <returns>The updated memory builder.</returns>
+    /// <remarks>
+    /// This <paramref name="bearerKeyProvider"/> will be called on every request,
+    /// when providing the token consider using caching strategy and refresh token logic
+    /// when it is expired or close to expiration.
+    /// </remarks>
+    public static MemoryBuilder WithVertexAITextEmbeddingGeneration(
```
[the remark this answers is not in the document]
Krzysztof318: @RogerBarreto told me I have to implement memory builder extensions.

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
```diff
@@ -0,0 +1,238 @@
...
+    /// </item>
+    /// </list>
+    /// For all options where an instance is provided, auto-invoke behavior may be selected. If the service
+    /// sends a request for a function call, if auto-invoke has been requested, the client will attempt to
+    /// resolve that function from the functions available in the <see cref="Kernel"/>, and if found, rather
+    /// than returning the response back to the caller, it will handle the request automatically, invoking
+    /// the function, and sending back the result. The intermediate messages will be retained in the
+    /// <see cref="ChatHistory"/> if an instance was provided.
+    /// </remarks>
+    public ToolCallBehavior? ToolCallBehavior
```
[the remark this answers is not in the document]
Krzysztof318: I don't think so. ToolCallBehavior is nested in GeminiPromptExecutionSettings class.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/StreamJsonParser.cs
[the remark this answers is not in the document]
stephentoub: > It is quite universal parser.

That's my concern. It's implemented doing synchronous I/O and processing one character at a time. That means it'll be blocking even for asynchronous consumers, and rather than doing vectorized searching (e.g. what IndexOfAny does), it's instead having to examine each character individually.

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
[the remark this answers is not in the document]
stephentoub: > I don't think so. ToolCallBehavior is nested in GeminiPromptExecutionSettings class.

The property, yes. But not the type:
```C#
namespace Microsoft.SemanticKernel.Connectors.GoogleVertexAI;

/// <summary>Represents a behavior for Gemini tool calls.</summary>
public abstract class ToolCallBehavior
```

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFunctionToolCall.cs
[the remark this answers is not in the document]
Krzysztof318: right :)

dotnet/src/Connectors/Connectors.GoogleVertexAI/ToolCallBehavior.cs
```diff
@@ -0,0 +1,189 @@
...
+    /// if this is 1, the first request will include the tools, but the subsequent response sending back the tool's result
+    /// will not include the tools for further use.
+    /// </remarks>
+    public int MaximumUseAttempts { get; set; } = int.MaxValue;
+
+    /// <summary>Gets how many tool call request/response roundtrips are supported with auto-invocation.</summary>
+    /// <remarks>
+    /// To disable auto invocation, this can be set to 0.
+    /// </remarks>
+    public int MaximumAutoInvokeAttempts { get; set; }
```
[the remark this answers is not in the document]
stephentoub: Because it's an implementation detail that still needs to be vetted, e.g. https://github.com/microsoft/semantic-kernel/pull/4686#discussion_r1464097022. Even as [Experimental], we should be surfacing only the surface area we think we're going to keep as stable.

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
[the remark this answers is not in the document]
Krzysztof318: hmm what we should do with that? In theory I could move this type to GeminiPromptExecutionSettings class but then I have to change name of this type due to conflict with property...

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/StreamJsonParser.cs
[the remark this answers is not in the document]
Krzysztof318: I will analyse this and let you what I can do with that

dotnet/src/Connectors/Connectors.GoogleVertexAI/ToolCallBehavior.cs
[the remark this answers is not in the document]
Krzysztof318: Okay, I will change these to private.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiMetadata.cs
```diff
@@ -0,0 +1,117 @@
...
+        GeminiMetadata metadata => metadata,
+        IDictionary<string, object?> metadata => new GeminiMetadata(metadata),
+        _ => new GeminiMetadata(dictionary.ToDictionary(pair => pair.Key, pair => pair.Value))
+    };
+
+    private void SetValueInDictionary(object? value, [CallerMemberName] string propertyName = "")
+        => this.Dictionary[propertyName] = value;
+
+    private object? GetValueFromDictionary([CallerMemberName] string propertyName = "")
+        => this.Dictionary.TryGetValue(propertyName, out var value) ? value : null;
```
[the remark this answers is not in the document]
Krzysztof318: It will be sufficient if I remove CallerMemberName and pass `nameof(<property>)` as method parameter?

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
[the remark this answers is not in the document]
Krzysztof318: I am going to take care for making this async tommorow.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiMetadata.cs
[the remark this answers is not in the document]
stephentoub: Yup! Thanks.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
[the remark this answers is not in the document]
stephentoub: Cool. We can also address it subsequently if we want to get the PR as it is merged. It's just important to address once this is "real".

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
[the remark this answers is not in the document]
stephentoub: I'm not sure what to do here yet... probably worth a discussion...

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFinishReason.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// </summary>
+    public static GeminiFinishReason Safety { get; } = new("SAFETY");
+
+    /// <summary>
+    /// The candidate content was flagged for recitation reasons.
+    /// </summary>
+    public static GeminiFinishReason Recitation { get; } = new("RECITATION");
+
+    /// <summary>
+    /// 	Unknown reason.
```
dmytrostruk, line 44:
```suggestion
    /// Unknown reason.
```

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFinishReason.cs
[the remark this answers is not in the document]
dmytrostruk: I would probably add this method only when there is such requirement for that. We have other strongly-typed enums in a codebase, like `AuthorRole`, but we don't have `GetAll` for it and so far we didn't see an interest in it.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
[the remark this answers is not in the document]
dmytrostruk: Also, we should probably simplify method `GetStreamingChatMessageContentsOrPopulateStateForToolCalling` :). The name of this method identifies that it is doing a lot of work, maybe it's possible to split the logic.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFinishReason.cs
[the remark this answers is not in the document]
Krzysztof318: yeah, removed

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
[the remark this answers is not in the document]
Krzysztof318: @dmytrostruk this is quite hard to refactor this method without performance impact. But method is not so complex indeed. This is version after update to async https://github.com/microsoft/semantic-kernel/blob/90b38059b129d936452242aa9af55b76d3623286/dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs#L233

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiFunction.cs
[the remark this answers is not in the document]
Krzysztof318: yeah, I have checked and you are right

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiMetadata.cs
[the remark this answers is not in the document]
Krzysztof318: @stephentoub Fixed but could you show me sample where CallerMemberName can fail? I am curious, should I avoid using CallerMemberName in code?

dotnet/src/Connectors/Connectors.GoogleVertexAI/ToolCallBehavior.cs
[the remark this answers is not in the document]
Krzysztof318: done

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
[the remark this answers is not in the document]
Krzysztof318: @stephentoub I modified parser to async version. I not used same pattern like azure there [LINK](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/src/Helpers/SseAsyncEnumerator.cs) because It requires more work, now I'd like to merge gemini and back to this later

dotnet/src/Connectors/Connectors.GoogleVertexAI/Extensions/VertexAIKernelBuilderExtensions.cs
```diff
@@ -0,0 +1,256 @@
...
+    /// <param name="bearerKey">The Bearer Key for authentication.</param>
+    /// <param name="location">The location to process the request</param>
+    /// <param name="projectId">Your project ID</param>
+    /// <param name="serviceId">The optional service ID.</param>
+    /// <param name="httpClient">The optional custom HttpClient.</param>
+    /// <returns>The updated kernel builder.</returns>
+    public static IKernelBuilder AddVertexAIGeminiTextGeneration(
+        this IKernelBuilder builder,
+        string modelId,
+        string bearerKey,
```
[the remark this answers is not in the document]
Krzysztof318: extensions  with const bearerKey param is intended only for short-lived chat completion services

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiMetadata.cs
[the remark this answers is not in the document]
stephentoub: For example, right now you have this:
```C#
    public string? PromptFeedbackBlockReason
    {
        get => this.GetValueFromDictionary() as string;
        internal init => this.SetValueInDictionary(value);
    }
```
That means if metadata was stored into this via PromptFeedbackBlockReason, a consumer could access it as:
```C#
var metadata = ...;
string? reason = metadata["PromptFeedbackBlockReason"];
```
Now imagine someone in the future innocently refactors to this:
```C#
    public string? PromptFeedbackBlockReason
    {
        get => GetStringValueFromDictionary();
        internal init => this.SetStringValueInDictionary(value);
    }

    private string? GetStringValueFromDictionary() => this.GetValueFromDictionary() as string;
    private void SetStringValueInDictionary(string value) => SetValueInDictionary(value);
```
That consumer will now be broken, because the string won't have been stored in metadata["PromptFeedbackBlockReason"] but rather in metadata["SetStringValueInDictionary"].

Or let's say it was refactored to just:
```C#
    public string? PromptFeedbackBlockReason
    {
        get => GetStringValueFromDictionary();
        internal init => this.SetValueInDictionary(value);
    }

    private string? GetStringValueFromDictionary() => this.GetValueFromDictionary() as string;
```
Now access to PromptFeedbackBlockReason will throw because it will be trying to read metadata["GetStringValueFromDictionary"] from the dictionary.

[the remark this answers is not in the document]
Krzysztof318: I didn't think about that case. Thanks for clarification. I will remember for the future :)

dotnet/src/Connectors/Connectors.GoogleVertexAI/Extensions/VertexAIMemoryBuilderExtensions.cs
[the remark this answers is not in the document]
RogerBarreto: Some of our examples does uses this, and is experimental, from this perspective we can drop this as well as the example as soon we mark `MemoryBuilder` as obsolete or don't invest on this extensibility at all.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Extensions/GeminiKernelFunctionMetadataExtensions.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using Microsoft.SemanticKernel.Connectors.GoogleVertexAI;
+
+namespace Microsoft.SemanticKernel;
```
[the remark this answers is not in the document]
RogerBarreto: @Krzysztof318 I had a conversation with @SergeyMenshykh and other folks from the team and for this both (PluginExtensions and KernelFUnctionMetadataExtrensions) following the OpenAI namespacing seems to be the ideal. 

@SergeyMenshykh, it is still inconclusive from that conversation what is the rational to use `Connector` namespace or `SemanticKernel`'s.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Extensions/VertexAIKernelBuilderExtensions.cs
```diff
@@ -0,0 +1,256 @@
...
+    /// <returns>The updated kernel builder.</returns>
+    /// <remarks>
+    /// This <paramref name="bearerKeyProvider"/> will be called on every request,
+    /// when providing the token consider using caching strategy and refresh token logic
+    /// when it is expired or close to expiration.
+    /// </remarks>
+    public static IKernelBuilder AddVertexAIGeminiTextGeneration(
+        this IKernelBuilder builder,
+        string modelId,
+        Func<string> bearerKeyProvider,
```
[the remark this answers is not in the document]
RogerBarreto: @Krzysztof318, I agree with the first point of Sergey on the BearerKey namning, we may consider renaming the parameter to `BearerToken` instead of key what you think @SergeyMenshykh?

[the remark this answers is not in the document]
Krzysztof318: @RogerBarreto I have already done that in PR which you merged :)

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -0,0 +1,642 @@
...
+            functionResponse: functionResult, errorMessage: null);
+    }
+
+    private async Task<GeminiResponse> SendRequestAndReturnValidGeminiResponseAsync(
+        Uri endpoint,
+        GeminiRequest geminiRequest,
+        CancellationToken cancellationToken)
+    {
+        using var httpRequestMessage = await this.CreateHttpRequestAsync(geminiRequest, endpoint).ConfigureAwait(false);
+        string body = await this.SendRequestAndGetStringBodyAsync(httpRequestMessage, cancellationToken)
```
dmytrostruk, line 373: nit: While all of these methods are internal/private, it's not a critical problem, but I noticed a couple of methods with this naming:
1. GetStreamingChatMessageContentsOrPopulateStateForToolCallingAsync
2. SendRequestAndReturnValidGeminiResponseAsync
3. SendRequestAndGetStringBodyAsync
4. SendRequestAndGetResponseImmediatelyAfterHeadersReadAsync
5. ValidateAndPrepareChatHistory

These names indicate one or a couple of problems:
1. Violation of Single Responsibility Principle.
2. The name itself has too much details.

Looking at these method implementations I don't think they violate Single Responsibility Principle, so most probably it's option 2.

For example, there is no need in `SendRequestAndGetStringBodyAsync` naming, because when you use this method, you already specify the type and nature of result - `string body`:
```csharp
string body = await this.SendRequestAndGetStringBodyAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
// or it could be just
string body = await this.SendRequestAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
// or to make it even more specific
string responseBody = await this.SendRequestAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
```

You shouldn't do the renaming at this point, but just a general recommendation for the future to keep it simple.
While it's not critical for internal/private methods, we should definitely keep this in mind for public methods. 

I hope that makes sense!

P.S. Yes, I know we already have `ReadAsByteArrayAndTranslateExceptionAsync` but this naming should be an exception, rather than common pattern :)

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiTextGenerationClient.cs
```diff
@@ -0,0 +1,97 @@
...
+/// </summary>
+internal sealed class GeminiTextGenerationClient
+{
+    private readonly IGeminiChatCompletionClient _chatCompletionClient;
+
+    /// <summary>
+    /// Represents a client for interacting with the text generation gemini model.
+    /// </summary>
+    /// <param name="chatCompletionClient">A Gemini chat completion client instance.</param>
+    internal GeminiTextGenerationClient(IGeminiChatCompletionClient chatCompletionClient)
```
dmytrostruk, line 23: Taking into account this dependency, do we need text generation client at all? Is there any difference from Google AI API point of view, or it's just a matter of usage in SK? Because if there is no functional difference between text generation and chat completion in Google AI API, I think it should be fine to release only chat completion client.

In OpenAI it's different case, because ITextGenerationService implementation supports different set of OpenAI models - text generation models.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiContent.cs
```diff
@@ -0,0 +1,28 @@
...
+internal sealed class GeminiContent
+{
+    /// <summary>
+    /// Ordered Parts that constitute a single message. Parts may have different MIME types.
+    /// </summary>
+    [JsonPropertyName("parts")]
+    public IList<GeminiPart>? Parts { get; set; }
+
+    /// <summary>
+    /// Optional. The producer of the content. Must be either 'user' or 'model'.
```
dmytrostruk, line 21: nit: Looking at `AuthorRoleConverter` class, it also handles `function` value, not sure if we need to update it here as well.

dotnet/src/Connectors/Connectors.GoogleVertexAI.UnitTests/Core/Gemini/Clients/GeminiChatGenerationFunctionCallingTests.cs
```diff
@@ -0,0 +1,403 @@
...
+    [Fact]
+    public async Task IfAutoInvokeShouldAddFunctionsCalledByModelToChatHistoryAsync()
+    {
+        // Arrange
+        using var handlerStub = new MultipleHttpMessageHandlerStub();
+        handlerStub.AddJsonResponse(this._responseContentWithFunction);
+        handlerStub.AddJsonResponse(this._responseContent);
+#pragma warning disable CA2000
+        var client = this.CreateChatCompletionClient(httpClient: handlerStub.CreateHttpClient());
+#pragma warning restore CA2000
```
dmytrostruk, line 192: nit: Noticed this warning suppression in a couple of places. Can we do the following?
```csharp
using var httpClient = handlerStub.CreateHttpClient();
var client = this.CreateChatCompletionClient(httpClient: httpClient);
```

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiChatCompletionClient.cs
Krzysztof318: Thanks, naming never be my strong side 😅 but I will take your comments for the future

dotnet/src/Connectors/Connectors.GoogleVertexAI.UnitTests/Core/Gemini/Clients/GeminiChatGenerationFunctionCallingTests.cs
Krzysztof318: If I remove pragram, pipieline will fail with formatter errors

Krzysztof318: ChatCompletion is common and refers to Generating at once and Streaming. ChatGeneration tests only generation , ChatStreaming only streaming.

dotnet/samples/KernelSyntaxExamples/Resources/sample_image.jpg
markwallace-microsoft, line 1: Need to verify came from a source which allows us to reuse it?

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiRequest.cs
```diff
@@ -0,0 +1,247 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.SemanticKernel.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;
+
+internal sealed class GeminiRequest
```
markwallace-microsoft, line 12: We should consider adding all of the internal model classes used by the client into a `Models` folder. Namespace can remain the same.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/IGeminiChatCompletionClient.cs
markwallace-microsoft, line 1: Why do we need this interface? If there will only ever be one implementation then this would be redundant.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/Clients/GeminiTokenCounterClient.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Net.Http;
+using System.Text.Json.Nodes;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;
+
+/// <summary>
+/// Represents a client for token counting gemini model.
```
markwallace-microsoft, line 13: nit: We have `gemini` in comments in multiple places and it should be `Gemini`

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/GoogleAI/GoogleAIEmbeddingClient.cs
```diff
@@ -0,0 +1,73 @@
...
+        ILogger? logger = null)
+        : base(
+            httpClient: httpClient,
+            logger: logger)
+    {
+        Verify.NotNullOrWhiteSpace(modelId);
+        Verify.NotNullOrWhiteSpace(apiKey);
+
+        this._embeddingModelId = modelId;
+        this._embeddingEndpoint = new Uri($"https://generativelanguage.googleapis.com/v1beta/models/{this._embeddingModelId}:batchEmbedContents?key={apiKey}");
```
markwallace-microsoft, line 41: The `v1beta` part of the URL is going to change and there may be times when multiple versions will be supported at the same time.

We should have:

1. A central place where the default API version is set
2. We should consider if we want to allow developers to change this. We do allow this in Azure OpenAI. It's a risky thing for a developer to do but that doesn't mean someone may have a good reason and we don't necessarily want to block them.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/StreamJsonParser.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Text.Json.Nodes;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;
+
+/// <summary>
+/// Internal class for parsing a stream of jsons into separate json objects.
```
markwallace-microsoft, line 15: nit: "stream of jsons into separate json objects" -> "stream of text which contains a series of discrete JSON strings into en enumerable containing each seperate JSON string"

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
```diff
@@ -0,0 +1,238 @@
...
+
+    /// <summary>
+    /// Temperature controls the randomness of the completion.
+    /// The higher the temperature, the more random the completion.
+    /// Range is 0.0 to 1.0.
+    /// </summary>
+    [JsonPropertyName("temperature")]
+    public double? Temperature
+    {
+        get => this._temperature;
```
markwallace-microsoft, line 42: Is there no default value defined for temperature? What happens when this is null?

stephentoub:
> Let's create a very simple and short ADR listing the implementation decisions that were made here so we have a reference for future connectors. Things like the decision not to use [.NET client library | Google Cloud](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.AIPlatform.V1/latest) should be included.

Why isn't that being used?

Krzysztof318:
@stephentoub @markwallace-microsoft 
Nuget from google targets net standard 2.1

stephentoub:
> @stephentoub @markwallace-microsoft Nuget from google targets net standard 2.1

Is it open source somewhere? Can we help them add a netstandard20 target?

@ericstj, @joperezr, @viktorhofer, this looks like another example where having a net462 asset but not a netstandard20 asset fails. I'm not sure why we've pushed for that or adopted it ourselves in some cases.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Connectors.GoogleVertexAI.csproj
```diff
@@ -0,0 +1,32 @@
...
+  </PropertyGroup>
+
+  <!-- IMPORT NUGET PACKAGE SHARED PROPERTIES -->
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props"/>
+  <Import Project="$(RepoRoot)/dotnet/src/InternalUtilities/src/InternalUtilities.props"/>
+
+  <PropertyGroup>
+    <!-- NuGet Package Settings -->
+    <Title>Semantic Kernel - Google AI/Vertex AI connectors</Title>
+    <Description>Semantic Kernel connectors for Google Gemini. Contains clients for text generation and text embedding generation.</Description>
```
RogerBarreto, line 19: @matthewbolanos Suggestion for the description and package name.

This package supports both GoogleAI and VertexAI platforms with current naming is not clear it also supports `GoogleAI`

### Package name suggestions
- keep `Microsoft.SemanticKernel.Connectors.GoogleVertexAI`
- change to `Microsoft.SemanticKernel.Connectors.Google`

### Description change

This change leaves the space open for adding more services without having to revisit the description every time a new modality is added.

```suggestion
    <Description>Semantic Kernel connectors for Google generation platforms (GoogleAI/VertexAI). Contains generation and embedding services.</Description>
```

Krzysztof318:
@stephentoub probably part of source is available, but for first look I saw only g.cs files... https://github.com/googleapis/google-cloud-dotnet

dotnet/src/Connectors/Connectors.GoogleVertexAI/Connectors.GoogleVertexAI.csproj
Krzysztof318: Are you sure? I also think current name is too long. But just "google" will be clear? So all new features related to google have to be in this package.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/IGeminiChatCompletionClient.cs
Krzysztof318: I can remove that, if we decide to get rid of Gemini Text generation service

ericstj:
> > @stephentoub @markwallace-microsoft Nuget from google targets net standard 2.1
> 
> Is it open source somewhere? Can we help them add a netstandard20 target?
> 
> @ericstj, @joperezr, @ViktorHofer, this looks like another example where having a net462 asset but not a netstandard20 asset fails. I'm not sure why we've pushed for that or adopted it ourselves in some cases.

We have found that developers get a better experience when consuming framework specific implementations (fewer dependencies) and library authors can access more API. We haven't recommended dropping `netstandard2.0` support in existing packages that are useful for the library ecosystem but we have reccomended adding more specific frameworks.  Some packages which focus more on application models have dropped support for netstandard2.0. https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/#what-you-should-target. 

@jskeet what was the motivation to drop `netstandard2.0` support from the Google-Cloud-Dotnet packages?  What's been the ecosystem reaction to that change?

jskeet:
> @jskeet what was the motivation to drop netstandard2.0 support from the Google-Cloud-Dotnet packages?

Looking back in my docs from 2021 when we were planning the change from GAX 3.0 to GAX 4.0, I have:

> [.NET Core 2.1 has already reached the end of support (August 2021).](https://devblogs.microsoft.com/dotnet/net-core-2-1-will-reach-end-of-support-on-august-21-2021/) We may find there are opportunities in moving to support via .NET Standard 2.1 (supported by .NET Core 3.1) instead of 2.0. This would enable work using `Span<T>`, default interface implementations etc more simply, for example - although .NET Framework support will still be a bit of a limiting factor here.

So it was somewhat speculative - it was hard to know in advance which features which are only in 2.1 would be useful in the course of the lifetime of the next major version.

As it happens, I've just tried building the GAX projects targeting netstandard2.0, and that *appears* to be feasible. (A couple of `#if`s needed fixing, but that's all.)

We could at least *consider* a new minor version changing the target from netstandard2.1 to netstandard2.0, and then change our generated libraries to target netstandard2.0 as well. I'd need to consult internally - this wouldn't be a *quick* change. It would be in line with what we want to do for Google.Protobuf though, which is (when I get time) to change the TFMs to netstandard2.0, net462 and net6.0 (from the rather haphazard collection that's currently targeted).

Could smarter minds than mine confirm that if everything builds and the tests pass, there should be no problem in changing the target from netstandard2.1 to netstandard2.0? (I'd *prefer* not to just add a new target, although I guess that's an option as well, and if there are concrete reasons that's safer, I'd be very happy to consider it carefully.)

> What's been the ecosystem reaction to that change?

I think this is the first time I've seen any pushback, to be honest. I think most of the consumers of the libraries are applications directly, where this doesn't cause a problem.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Connectors.GoogleVertexAI.csproj
Krzysztof318: Namespace also?

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiContent.cs
Krzysztof318: done

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/GoogleAI/GoogleAIEmbeddingClient.cs
Krzysztof318: So create static class or inject api version via DI extensions?

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
Krzysztof318: Any of params can be null then gemini takes default for used model. Defaults are described in google docs.

stephentoub:
Thanks, @jskeet. @ericstj, could you help with this part?

> Could smarter minds than mine confirm that if everything builds and the tests pass, there should be no problem in changing the target from netstandard2.1 to netstandard2.0?

ericstj:
> @ericstj, could you help with this part?
>> Could smarter minds than mine confirm that if everything builds and the tests pass, there should be no problem in changing the target from netstandard2.1 to netstandard2.0?

It should be fine - you might need a couple package references (`System.Memory`, for example).  Check for `#ifs` in your codebase -- `NETSTANDARD2_1_OR_GREATER` `NETSTANDARD2_1`.  Check your project files for any conditions that applied to `netstandard2.1` or similar.  To add some confidence that the APIs are the same you can run [APICompat](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/overview) to compare [before and after](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/baseline-version-validator).  This should help guarantee that no API was removed.

jskeet:
> It should be fine - you might need a couple package references (`System.Memory`, for example). Check for `#ifs` in your codebase -- `NETSTANDARD2_1_OR_GREATER` `NETSTANDARD2_1`. Check your project files for any conditions that applied to `netstandard2.1` or similar. To add some confidence that the APIs are the same you can run [APICompat](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/overview) to compare [before and after](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/baseline-version-validator). This should help guarantee that no API was removed.

Great, thanks. I did a quick test which highlighted a couple of `#ifs` that need changing, but that should be all. My colleague on the ".NET in Google" team is supportive of this as well, so we'll move forward with it. I can't guarantee any timelines, but I'll add a comment when we've made more progress.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Connectors.GoogleVertexAI.csproj
RogerBarreto: I spoke with @matthewbolanos, the idea is the Connector to have the name of the provider and any service within will be provided by the same package for the given provider. 

If we end up needing to have multiple projects for one provider we can always create a Meta Package that combines multiple platforms of a same provider.

jskeet:
> It should be fine - you might need a couple package references (`System.Memory`, for example). Check for `#ifs` in your codebase -- `NETSTANDARD2_1_OR_GREATER` `NETSTANDARD2_1`. Check your project files for any conditions that applied to `netstandard2.1` or similar. To add some confidence that the APIs are the same you can run [APICompat](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/overview) to compare [before and after](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/baseline-version-validator). This should help guarantee that no API was removed.

Hmm. Currently the compat checker seems to think I've removed *everything*. Will investigate in a tiny test project to see whether I'm just using the tooling badly, which is entirely likely.

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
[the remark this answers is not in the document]
RogerBarreto: @Krzysztof318, lets use a prefix for the `ToolCallBehavior` as `GeminiToolCallBehavior` to avoid the clash, we might have this sorted out once the Function Call abstraction is in place.

jskeet:
> Hmm. Currently the compat checker seems to think I've removed _everything_. Will investigate in a tiny test project to see whether I'm just using the tooling badly, which is entirely likely.

Correction: not everything, just a lot. All enum values, and members using nullable value types.

I've reproduced this in a tiny project. I *expect* this is a bug in the compat checker (which probably doesn't have many testcases for changing targets from netstandard2.1 to netstandard2.0!).

I'll file a bug in dotnet/sdk.

dotnet/src/Connectors/Connectors.GoogleVertexAI/GeminiPromptExecutionSettings.cs
markwallace-microsoft: @stephentoub Sergey is currently working on defining the core abstractions to support this. We will have three implementations to base the design on (OpenAI, Gemini and MistralAI).

dotnet/src/Connectors/Connectors.GoogleVertexAI/Extensions/GeminiKernelFunctionMetadataExtensions.cs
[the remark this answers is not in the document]
RogerBarreto: Just confirmed that rational is be to keep the namespace of the target extending class. This file can keep targeting `SemanticKernel` namespace instead of the connector.

jskeet:
> I'll file a bug in dotnet/sdk.

Filed https://github.com/dotnet/sdk/issues/39629 if anyone wants to take a look. I don't think that needs to block an alpha or beta release of GAX and Google.Cloud.AIPlatform.V1, but I'd be *slightly* nervous about releasing a GA version with the compat tester reporting errors unless we've at least confirmed that's a bug in the compat tester.

dotnet/src/Connectors/Connectors.Google/GeminiToolCallBehavior.cs
```diff
@@ -0,0 +1,189 @@
...
+        internal KernelFunctions(bool autoInvoke) : base(autoInvoke) { }
+
+        public override string ToString() => $"{nameof(KernelFunctions)}(autoInvoke:{this.MaximumAutoInvokeAttempts != 0})";
+
+        internal override void ConfigureGeminiRequest(Kernel? kernel, GeminiRequest request)
+        {
+            // If no kernel is provided, we don't have any tools to provide.
+            if (kernel is null)
+            {
+                return;
```
markwallace-microsoft, line 124: Below in `EnabledFunctions` if there is no kernel and auto invoke is set to true an exception is thrown. Why is the behaviour different in this case?

dotnet/src/Connectors/Connectors.Google/Core/Gemini/GeminiFunctionToolResult.cs
```diff
@@ -0,0 +1,32 @@
...
+        Verify.NotNull(functionResult);
+
+        this.FunctionResult = functionResult;
+        this.FullyQualifiedName = toolCall.FullyQualifiedName;
+    }
+
+    /// <summary>
+    /// Gets the result of the function.
+    /// </summary>
+    public FunctionResult FunctionResult { get; }
```
SergeyMenshykh, line 27: Having a dependency on the non-serializable FunctionResult class makes the GeminiFunctionToolResult non-serializable as well. As a result, it will be impossible to serialize chat history containing function call results.

Krzysztof318: So maybe we can do FunctionResult serializable?

dotnet/src/Connectors/Connectors.Google/GeminiToolCallBehavior.cs
Krzysztof318: OpenAI impl also throws exception here https://github.com/microsoft/semantic-kernel/blob/79620c727e61d056736674018156715cba690a92/dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs#L192

dotnet/src/Connectors/Connectors.Google/Core/Gemini/GeminiFunctionToolResult.cs
SergeyMenshykh: We can try, but after the exercise, most likely, we will see that the Function and ValueType properties are not serializable/deserializable and need to be ignored. This leaves us with not too many useful properties of the function result class.  
   
There could also be additional changes after those properties are ignored - they might need to be marked as nullable, and an additional constructor that does not have those properties in its signature needs to be added for deserialization purposes.  
   
We also need a way to pass exceptions to the connector in case the function invocation fails, and I don't see how function result class can help us with this.

Krzysztof318: So I suggest leave it as is for now and update this when we finish tool call abstraction.

dotnet/src/Connectors/Connectors.Google/GeminiToolCallBehavior.cs
RogerBarreto: @Krzysztof318, that was a good spot from @markwallace-microsoft, we should have consistency on those. 
By looking into this #5655, the pattern is to thrown an exception if no `kernel` is provided.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/GeminiFunctionToolResult.cs
RogerBarreto: Created #5658 for tracking in future.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/GoogleAI/GoogleAIEmbeddingClient.cs
RogerBarreto: Created #5659 for tracking of this.

dotnet/src/Connectors/Connectors.Google/GeminiToolCallBehavior.cs
Krzysztof318: @RogerBarreto @markwallace-microsoft  this PR #5637 only changes RequiredFunctions (gemini doesn't have this). Only has KernelFunction and EnabledFunction.
EnabledFunction throws exception if kernel is null, this is consistent with openAI ToolCallBehaviour.
KernelFunction doesn't throw exception if kernel is null, this is also consistent with openAI ToolCallBehaviour.

dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/Gemini/GeminiRequest.cs
RogerBarreto: Issue #5676 created to track this suggestion.

jskeet:
> I can't guarantee any timelines, but I'll add a comment when we've made more progress.

[Google.Cloud.AIPlatform.V1](https://www.nuget.org/packages/Google.Cloud.AIPlatform.V1) now supports netstandard2.0 :)

I can't guarantee that it'll be useful in this code - I don't know much about the context - but at least that's one blocker out of the way...

stephentoub:
Thank you, @jskeet!

RogerBarreto:
That's great news thanks @stephentoub @jskeet @ericstj!

SergeyMenshykh (approved):

markwallace-microsoft (approved):

NikolayTV:
Hi! Log waited update. When do you think gemini gets added to Python sdk?

PR #5475, opened 2024-03-14
.Net: Gemini - Async stream parser, read-only tool attempts and minor fixes

Krzysztof318:
Changed StreamJsonParser to async
Made read-only MaximumUseAttempts and MaximumAutoInvokeAttempts in ToolBehavior
Removed CallerMemberName from GeminiMetadata and used nameof instead.
Removed GetAll from FinishReason, SafetyRating and SafetySetting
Refactor bearer key logic to handle generating tokens async. Renamed from bearer key to bearer token.

@RogerBarreto

### DIFF
--- a/dotnet/src/Connectors/Connectors.GoogleVertexAI.UnitTests/Core/StreamJsonParserTests.cs
+++ b/dotnet/src/Connectors/Connectors.GoogleVertexAI.UnitTests/Core/StreamJsonParserTests.cs
@@ -4,6 +4,7 @@ using System;
 using System.IO;
 using System.Linq;
 using System.Text.Json;
+using System.Threading.Tasks;
 using Microsoft.SemanticKernel.Connectors.GoogleVertexAI.Core;
 using Xunit;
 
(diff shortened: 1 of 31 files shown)


dotnet/src/Connectors/Connectors.GoogleVertexAI.UnitTests/Core/StreamJsonParserTests.cs
```diff
@@ -121,7 +122,7 @@ public void ParseWhenStreamContainsOneObjectReturnsEnumerableWithOneObjectWithEs
         WriteToStream(stream, input);
 
         // Act
-        var result = parser.Parse(stream);
+        var result = parser.ParseAsync(stream);
```
RogerBarreto, line 125: To test async, make sure your tests area async and you await the task. You may not be testing it this way, as they will run and the test will end before.

```csharp
public async Task ParseWhen...
...
var result = await parser.ParseAsync(stream);

```

Krzysztof318: done

PR #5486, opened 2024-03-14
.Net - Expose Agent Thread Messages

crickman:
### Motivation and Context

Retrieving a thread based on ID is supported, but there's not a good way to retrieve messages.

### Description

Don't want to store messages in thread since this could impact memory pressure but they _do_ need to be exposed when re-approaching an existing thread.  Turns out there was some dead-code from the early POC to clean-up also.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Experimental/Agents.UnitTests/Integration/ThreadHarness.cs
+++ b/dotnet/src/Experimental/Agents.UnitTests/Integration/ThreadHarness.cs
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
-#define DISABLEHOST // Comment line to enable
+//#define DISABLEHOST // Comment line to enable
 using System.Threading.Tasks;
 using Microsoft.SemanticKernel;
 using Microsoft.SemanticKernel.Experimental.Agents;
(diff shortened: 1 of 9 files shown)


dotnet/src/Experimental/Agents.UnitTests/Integration/ThreadHarness.cs
```diff
@@ -68,4 +68,32 @@ public async Task VerifyThreadLifecycleAsync()
 
         await Assert.ThrowsAsync<HttpOperationException>(() => context.GetThreadModelAsync(thread.Id)).ConfigureAwait(true);
     }
+
+    /// <summary>
+    /// Verify retrieval of thread messages
+    /// </summary>
+    [Fact(Skip = SkipReason)]
```
alliscode, line 75: Should this new be skipped? If so, a more descriptive reason would be good.

alliscode (approved):

dotnet/src/Experimental/Agents.UnitTests/Integration/ThreadHarness.cs
crickman: Everything in this folder is disabled by default since they aren't unit-tests, they really dev-harnesses posing as integration tests.  Each file has a descriptor:

image

Github review tool hides this unless you dig!

TaoChenOSU (approved):

PR #5489, opened 2024-03-15
.Net: CJK support for text splitter

tonybaloney:
### Motivation and Context

1. The text splitting algorithm has hardcoded values for punctuation based on ASCII punctuation (`,.;` etc.) Chinese and Japanese use both full-width Unicode punctuation values https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block) or ideographic punctuation.
2. If you give the text splitter Japanese text it will incorrectly split in the middle of sentences
3. The current algorithm defaults to an approximation token counter, which is partly accurate for Indo-European/Latin Alphabet languages but totally out for CJK. This tests uses the BPE token length for cl100k_base (GPT3/4)

### Description

This change adds some basic full-width characters and ideographic punctuation to the builtin list.
It also adds a test which checks that it splits a sentence correctly. The first part of the test string (田中の猫はかわいいですね.) is 16 tokens for cl100k_base.

### Contribution Checklist


dotnet/src/SemanticKernel.Core/Text/TextChunker.cs
```diff
@@ -25,8 +25,8 @@ public static class TextChunker
     public delegate int TokenCounter(string input);
 
     private static readonly char[] s_spaceChar = new[] { ' ' };
-    private static readonly string?[] s_plaintextSplitOptions = new[] { "\n\r", ".", "?!", ";", ":", ",", ")]}", " ", "-", null };
-    private static readonly string?[] s_markdownSplitOptions = new[] { ".", "?!", ";", ":", ",", ")]}", " ", "-", "\n\r", null };
+    private static readonly string?[] s_plaintextSplitOptions = new[] { "\n\r", ".", "。", "．", "?!", ";", ":", ",", "，", "、", ")]}", " ", "-", null };
```
tonybaloney, line 28: I didn't add the full-width exclamation and question marks, or the like of ideographic brackets because it's pretty long https://www.w3.org/TR/jlreq/#cl-02

SergeyMenshykh (approved):

tonybaloney:
The build is broken because this uses the new Microsoft.ML.Tokenizers library which must have a breaking change because one of the sample files (unrelated to this PR) stops compiling. I'll take a look but it would have broken on the next release anyway

tonybaloney:
I made some changes to this PR to add the comma and fullstop variations to the string inside the list rather than individual items. This reduces the number of calls to TokenCounter. Implemented snapshot testing for some short stories in different languages to verify that those changes didn't impact how text was split and to ensure future regressions are captured.

markwallace-microsoft:
@tonybaloney could you take a look at the conflicts, thanks!

tonybaloney:
Sure, also the ML library API is going to change and there will be a sync version so the sample code in this PR will need updating. 

https://github.com/dotnet/machinelearning/issues/7077

markwallace-microsoft (approved):

Issue #5490, opened 2024-03-15
Python: Text splitters assume ASCII punctuation and don't support Chinese or Japanese


tonybaloney:
**Describe the bug**

The text splitters in Semantic Kernel assume ASCII punctuation. Languages that use full-width Unicode punctuation or ideographic punctuation (Chinese and Japanese being the most common) won't split correctly.

Test sentence: 
`田中の猫はかわいいですね。日本語上手。` doesn't split on the full-stop because it is the Unicode ideographic full-stop (the main one used in Japanese and Chinese text).

See [this article](https://tonybaloney.github.io/posts/cjk-chinese-japanese-korean-llm-ai-best-practices.html#text-splitting) for more details. 

I've tested this on both Python and .NET APIs.

Python: https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/text/text_chunker.py#L14-L38

In langchain you can override the separators and supply a list of [ideographic and full-width characters](https://www.w3.org/TR/jlreq/#character_classes)

```python
SEPARATORS = [
    "\n\n", ".",  # Paragraph boundary and ASCII full stop
    "。", "．",     # Logographic and full width full stop
    "!", "！",    # ASCII and full width exclamation mark
    "?", "？",     # ASCII and full width question mark
    ",", "、", "，", # ASCII, logographic, and full width comma
]
```

tonybaloney:
Submitted a PR for the .NET library #5489

evchaki:
@tonybaloney, thanks for entering the issue at the .net repo.  I will close this for now, until this is updated in .Net

PR #5494, opened 2024-03-15
.Net: Python: CJK support for text splitter

marlenezw:
### Motivation and Context
The text splitting algorithm has hardcoded values for punctuation based on ASCII punctuation (,.; etc.) Chinese and Japanese use both full-width Unicode punctuation values https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block) or ideographic punctuation.
If you give the text splitter Japanese text it will incorrectly split in the middle of sentences
The current algorithm defaults to an approximation token counter, which is partly accurate for Indo-European/Latin Alphabet languages but totally out for CJK. This tests uses the BPE token length for cl100k_base (GPT3/4)

This is the Python version of [this PR](https://github.com/microsoft/semantic-kernel/pull/5489)

### Description

This change adds some basic full-width characters and ideographic punctuation to the built-in list.
It also adds a test which checks that it splits a sentence correctly. The first part of the test string (田中の猫はかわいいですね.) is 16 tokens for cl100k_base.

### Contribution Checklist

The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
I didn't break anything!


python/pyproject.toml
```diff
@@ -32,6 +32,8 @@ pydantic = "^2"
 motor = "^3.3.2"
 defusedxml = "^0.7.1"
 pybars4 = "^0.9.13"
+jinja2 = "^3.1.3"
```
tonybaloney, line 35: Move these to the dev dependencies group below so they aren't dependencies of the whole project

python/semantic_kernel/text/text_chunker.py
```diff
@@ -14,10 +14,14 @@
 TEXT_SPLIT_OPTIONS = [
     ["\n", "\r"],
     ["."],
+    ["。"],
+    ["．"],
     ["?", "!"],
     [";"],
     [":"],
     [","],
```
tonybaloney, line 22: Put all the commas in a single list, like the line below with the brackets

python/semantic_kernel/text/text_chunker.py
```diff
@@ -14,10 +14,14 @@
 TEXT_SPLIT_OPTIONS = [
     ["\n", "\r"],
     ["."],
```
tonybaloney, line 16: Put all the fullstops in a single list, like the line below with the brackets

python/semantic_kernel/text/text_chunker.py
```diff
@@ -26,10 +30,14 @@
 
 MD_SPLIT_OPTIONS = [
     ["."],
+    ["。"],
```
tonybaloney, line 33: As above

python/semantic_kernel/text/text_chunker.py
```diff
@@ -26,10 +30,14 @@
...
+    ["．"],
     ["?", "!"],
     [";"],
     [":"],
     [","],
+    ["，"],
```
tonybaloney, line 39: As above

python/semantic_kernel/text/text_chunker.py
```diff
@@ -44,7 +52,7 @@ def _token_counter(text: str) -> int:
 
     TODO: chunking methods should be configurable to allow for different
```
tonybaloney, line 53: Note, this is configurable so I don't think this TODO is valid. We should raise a separate issue

tonybaloney:
Related #5490

python/pyproject.toml
marlenezw: Sounds good! I'm not the one that added Jinja, so hopefully tests still pass!

python/semantic_kernel/text/text_chunker.py
marlenezw: Sounds good, will do!

marlenezw:
Ooof, I tried to rebase to fix the merge conflicts with the poetry file and it turned into a bit of a mess. @tonybaloney I'm going to close this PR and re-do it so it doesn't take too much time 😅

Issue #3256, opened 2023-10-23
.Net [ChatRoles] 2. Completion service (ChatCompletion / TextCompletion) selection by SemanticFunction


SergeyMenshykh:
As an engineer working on SK,
I want to have a mechanism allowing semantic function to select either text or chat completion API/Connector for execution
So, text prompts are handled by text completion connectors
And chat prompts are handled by chat completion connectors
Or chat prompt handled by text completion connectors and the other way around for fallback scenarios

Scope:

1. Create ADR and present it to the team.
2. Collect feedback and implement the completion service selection functionality.
```[tasklist]
### Tasks
- [x] ADR
- [ ] Implement
```

stekiri:
@dmytrostruk, I read the ADR and it seems that the described logic has not been implemented. A different logic seems to have been introduced in https://github.com/microsoft/semantic-kernel/pull/4085. Is the ADR obsolete?

I stumbled upon it in the following scenario:
- Custom service that implements both `IChatCompletionService` and `ITextGenerationService`.
- Chat completion implementation is chosen without taking into account the logic of the ADR.

dmytrostruk:
@stekiri Thank you for this comment. You are correct, the ADR which we added previously is kind of obsolete, because later we decided to proceed with other approach, which wasn't described in the documentation. We created a new ADR that explains new approach and we updated status of previous ADR in this PR: https://github.com/microsoft/semantic-kernel/pull/5479

Do you have a specific scenario that is not covered by current approach?

stekiri:
Yes, my scenario is when a service implements both `IChatCompletionService` and `ITextGenerationService`.

My reasoning would be:
If a user implements `ITextGenerationService` for a service, then the user made a conscious choice that he considers it being the important (and maybe better as chat completion?) for a given use case. Therefore, it should take precedence over `IChatCompletionService` if there is no chat in his prompt.

Some code is worth more than a thousand words, so I made this small [draft PR](https://github.com/microsoft/semantic-kernel/pull/5495) to communicate what I had in mind. My logic might be flawed, so maybe I miss some use cases that would not be supported with such a change.

dmytrostruk:
> If a user implements `ITextGenerationService` for a service, then the user made a conscious choice that he considers it being important (and maybe better as chat completion?) for a given use case. Therefore, it should take precedence over `IChatCompletionService` if there is no chat in his prompt.

[...]

PR #5495, opened 2024-03-15
Modify service selection logic

stekiri:
Draft to facilitate discussion on https://github.com/microsoft/semantic-kernel/issues/3256.


stekiri:

PR #5502, opened 2024-03-15
.Net: Loosen TextChunker's lines input type

stephentoub:
It currently requires a `List<string>`. This both annoying and unnecessary.

I also removed duplicative experimental attributes.

(Note this is a binary breaking change, not source breaking, but the type is marked experimental.)


markwallace-microsoft (approved):

markwallace-microsoft (commented):
@stephentoub The suppressions file needs to be updated

RogerBarreto (approved):

PR #5503, opened 2024-03-15
Python: Update action_planner.py

scov8:
Some time with this replace json.loads goes in error Ad example if we have this str
…
“arguments”:””
…
It puts just one of “ and go in error

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/planners/action_planner/action_planner.py
+++ b/python/semantic_kernel/planners/action_planner/action_planner.py
@@ -114,7 +114,7 @@ class ActionPlanner:
             raise PlannerInvalidPlanError(f"No valid plan has been generated. Plan is:  {generated_plan_raw_str}")
 
         generated_plan_str = generated_plan_str.group()
-        generated_plan_str = generated_plan_str.replace('""', '"')
+        # generated_plan_str = generated_plan_str.replace('""', '"')
 
         try:
             generated_plan = json.loads(generated_plan_str)


python/semantic_kernel/planners/action_planner/action_planner.py
```diff
@@ -114,7 +114,7 @@ async def create_plan(self, goal: str) -> Plan:
             raise PlannerInvalidPlanError(f"No valid plan has been generated. Plan is:  {generated_plan_raw_str}")
 
         generated_plan_str = generated_plan_str.group()
-        generated_plan_str = generated_plan_str.replace('""', '"')
+        # generated_plan_str = generated_plan_str.replace('""', '"')
```
moonbox3, line 117: In most scenarios, this line of code is useful. Instead of simply commenting this out, could we maybe add some error handling around this action? What specific error is thrown if it fails to replace quotes? Would a try except help? You're saying with the `generated_plan_str` of the form `"arguments":""` it fails?

moonbox3 (changes requested):

python/semantic_kernel/planners/action_planner/action_planner.py
scov8: Thank you for your response and suggestions. I understand the importance of that line of code in many scenarios. However, the specific problem seems to arise when the replace mistakenly turns two consecutive quotes into a single quote, invalidating the JSON in specific cases such as
"arguments"'"". This could happen when the expected value is literally an empty string between double quotes, which is a valid case in JSON (as in my case).

One solution might be to check the string more carefully before performing the replace. We could try to avoid the replace when the double quotes are in specific contexts where they should not be modified. For example, we could check if the double quotes are within a valid JSON structure and avoid them in that case.
However, this could become complex and error-prone.

An alternative and simpler approach might be to introduce a post-replacement check to make sure that the JSON is still valid, and if not, attempt a recovery operation or undo the substitution made. Here is an exbochure of how this could be implemented:


original_plan_str = generated_plan_str generated_plan_str= generated_plan_str.replace('''' ''')
try:
    generated_plan = json.loads(generated_plan_str)
except json.decoder.JSONDecodeError:
    try:
        generated_plan = json.loads(original_plan_str)
    except json.decoder.JSONDecodeError as e:
        raise PlannerInvalidPlanError("'Encountered an error while parsing Plan JSON.") from e


In this way, we keep the attempt to clean double quotes, but have a fallback path that uses the original string if the first attempt results in an invalid JSON. This should mitigate the specific problem while maintaining the usefulness of the line of code for most use cases.
Please let me know if this solution seems reasonable or if you have any additional suggestions. Thanks again for your support.

moonbox3:
Closing this PR as we have deprecated the action planner. We appreciate your help and contribution.

PR #5520, opened 2024-03-18
Python: Enhance Chat Message Content creation and parsing

### LINKED ISSUE #5496: Python: ChatMessageContent Pythonic improvements
Pythonic improvements to streamline and improve how the ChatMessageContent classes are handled.

eavanvalkenburg:
### Motivation and Context

Introduces a polymorphic approach to creating and rendering Chat Message Contents and it's subclasses, leverages Pydantic Rootmodel for that.

Fixes #5496 

### Description

Adds ChatMessageContentBase that can create all types of CMC's, based on a type field (str)
Redoes some of the rendering and parsing
also retooled the hierarchy structure for Contents, including introducing a StreamingContentMixin instead of the StreamingKernelContent (which was never used), now all content classes (TextContent, CMC, etc) determine the fields used, while the Mixin adds one additional field and two additional methods for streaming messages, especially __add__ 
everything else runs the same way.

### Contribution Checklist


python/semantic_kernel/connectors/ai/open_ai/services/open_ai_chat_completion_base.py
```diff
@@ -241,7 +241,7 @@ def _create_streaming_chat_message_content(
             inner_content=chunk,
             ai_model_id=self.ai_model_id,
             metadata=metadata,
-            role=ChatRole(choice.delta.role) if choice.delta.role else None,
+            role=ChatRole(choice.delta.role) if choice.delta.role else ChatRole.ASSISTANT,
```
moonbox3, line 244: Same comment here about the default to Assistant.

python/semantic_kernel/contents/chat_message_content_base.py
```diff
@@ -0,0 +1,85 @@
...
+    after importing the ChatMessageContent and all it's subclasses. And you then have to use the root field.
+
+    The first two use dictionaries, directly or as kwargs to create the ChatMessageContent,
+    the last one uses an XML Element to create the ChatMessageContent.
+    All these methods then return the root field of the ChatMessageContentBase,
+      which is a instance of ChatMessageContent or the requested subclass.
+    """
+
+    root: Annotated[
+        Union["ChatMessageContent", "OpenAIChatMessageContent", "AzureChatMessageContent"],
```
moonbox3, line 35: Wouldn't it be better to make these "constants?"

python/tests/unit/contents/test_chat_history.py
```diff
@@ -184,14 +185,7 @@ def test_serialize():  # ignore: E501
...
         json_str
-        == '{\n    "messages": [\n        {\n            \
-"inner_content": null,\n            "ai_model_id": null,\n            \
-"metadata": {},\n            "role": "system",\n            \
-"content": "a test system prompt",\n            "encoding": null\n        },\
-\n        {\n            "inner_content": null,\n            \
-"ai_model_id": null,\n            "metadata": {},\n            \
-"role": "user",\n            "content": "Message",\n            \
-"encoding": null\n        }\n    ]\n}'
+        == '{\n    "messages": [\n        {\n            "metadata": {},\n            "type": "ChatMessageContent",\n            "role": "system",\n            "content": "a test system prompt"\n        },\n        {\n            "metadata": {},\n            "type": "ChatMessageContent",\n            "role": "user",\n            "content": "Message"\n        }\n    ],\n    "message_type": "ChatMessageContent"\n}'  # noqa: E501
```
moonbox3, line 188: Do you think it's necessary to include the metadata in the dump if there's nothing there?

moonbox3 (commented):
This is looking good. Some initial questions.

python/semantic_kernel/connectors/ai/open_ai/services/azure_chat_completion.py
```diff
@@ -280,7 +280,7 @@ def _create_streaming_chat_message_content(
             inner_content=chunk,
             ai_model_id=self.ai_model_id,
             metadata=metadata,
-            role=ChatRole(choice.delta.role) if choice.delta.role is not None else None,
+            role=ChatRole(choice.delta.role) if choice.delta.role else ChatRole.ASSISTANT,
```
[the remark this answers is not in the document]
eavanvalkenburg: That is correct, StreamingCMC is now a subclass of CMC, and role is mandatory there, and I preferred not to change that, rather set a default here.

python/tests/unit/contents/test_chat_history.py
eavanvalkenburg: this behaviour doesn't change, just formatted differently, a {} is not None, so it shows up with model_dump...

python/semantic_kernel/contents/chat_message_content_base.py
```diff
@@ -0,0 +1,84 @@
...
+
+from pydantic import Field, RootModel
+
+if TYPE_CHECKING:
+    from semantic_kernel.contents.chat_message_content import ChatMessageContent
+
+DISCRIMINATOR_FIELD: Final[str] = "type"
+
+
+class ChatMessageContentBase(RootModel):
```
moonbox3, line 22: I noticed in the three methods below, there's no error handling or checking that arguments are actually supplied. Is that intended?

python/tests/unit/contents/test_chat_history.py
```diff
@@ -281,6 +274,62 @@ async def test_template(chat_history: ChatHistory):
     assert chat_history_2.messages[2].role == ChatRole.USER
 
 
+@pytest.mark.asyncio
+async def test_chat_history_with_message_type():
+    chat_history = ChatHistory(message_type="OpenAIChatMessageContent")
```
moonbox3, line 279: As a dev, I am going to now need to supply the message type when I create a chat history?

moonbox3 (commented):
A few more questions.

python/semantic_kernel/contents/chat_history.py
```diff
@@ -188,26 +201,34 @@ def from_rendered_prompt(
...
         """
-        messages: List[chat_message_content_type] = []
+        messages: List[ChatMessageContent] = []
         prompt = rendered_prompt.strip()
         try:
-            xml_prompt = ET.fromstring(f"<prompt>{prompt}</prompt>")
-        except ET.ParseError as e:
+            xml_prompt = XML(text=f"<prompt>{prompt}</prompt>")
+        except ParseError as e:
             logger.error(f"Error parsing XML of prompt: {e}")
```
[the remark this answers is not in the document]
eavanvalkenburg: downgraded to info, and adapted the message.

python/semantic_kernel/contents/chat_message_content_base.py
eavanvalkenburg: yes, it relies on the "regular" pydantic validation errors if the right things are missing.

python/tests/unit/contents/test_chat_history.py
eavanvalkenburg: You can, but you don't have to, the ChatHistory does not care what kind of message is in there, as long as it is a CMCBase thing, during serialization (both to json and xml) the types of each message are passed along and when rebuilding the type of the whole history is used, if not specified, defaults to regular CMC. But if you want to have OpenAICMC's as every message (because you want to add stuff during hooks or something, you have the option!).

juliomenendez (approved):
LGTM

moonbox3 (approved):

PR #5548, opened 2024-03-19
.Net: Mistral Streaming Chat Completion

markwallace-microsoft:
### Motivation and Context

Mistral connector doesn't support streaming completion.

### Description

Support streaming completion for mistral connector.

### Contribution Checklist


dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
```diff
@@ -33,52 +34,89 @@ internal sealed class MistralClient
...
         using var response = await this.SendStreamingRequestAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
 
         using var responseStream = await response.Content.ReadAsStreamAndTranslateExceptionAsync().ConfigureAwait(false);
-        foreach (var streamingTextContent in this.ProcessResponseStream(responseStream, modelId))
+        using var reader = new StreamReader(responseStream);
+        string line;
+        string? rawChunk = null;
+        AuthorRole? currentRole = null;
+        while ((line = await reader.ReadLineAsync().ConfigureAwait(false)) != null)
         {
```
Krzysztof318, line 75: If you want, you can reuse my updated version of StreamJsonParser from gemini.
After update it is fully async and still fast.
https://github.com/microsoft/semantic-kernel/blob/feature-connectors-gemini/dotnet/src/Connectors/Connectors.GoogleVertexAI/Core/StreamJsonParser.cs

RogerBarreto: The way MistralAI outputs the chunks seems to be easier to Deserialize using a simpler approach for skipping the `line breaks`, @markwallace-microsoft may worth considering if the `IStreamJsonParser` interface can be applied to your streaming logic, this then can be another Parser utility we can use across different connectors.

stephentoub: Is it just using server-sent events?

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
```diff
@@ -33,52 +34,89 @@ internal sealed class MistralClient
...
-            yield return streamingTextContent;
+            if (!string.IsNullOrEmpty(line))
+            {
+                rawChunk = line.Substring("data:".Length).Trim();
```
RogerBarreto, line 78: Use a constant for "data:".Length 
i.e.: Avoid allocation of a new string and getting the same length.

```suggestion
                rawChunk = line.Substring(SSEDataLength).Trim();
```

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
```diff
@@ -33,52 +34,89 @@ internal sealed class MistralClient
...
+                {
+                    continue;
+                }
+                var chunk = await JsonSerializer.DeserializeAsync<MistralChatCompletionChunk>(
+                    new MemoryStream(Encoding.UTF8.GetBytes(rawChunk))).ConfigureAwait(false);
+                rawChunk = null;
+
+                if (chunk is null)
+                {
+                    throw new KernelException("Unexpected response from model")
```
RogerBarreto, line 92:
```suggestion
                    throw new KernelException("Unexpected chunk response from model")
```

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
```diff
@@ -33,52 +34,89 @@ internal sealed class MistralClient
...
+                rawChunk = line.Substring("data:".Length).Trim();
+            }
+            else
+            {
+                if (rawChunk is not null and "[DONE]")
+                {
+                    continue;
+                }
+                var chunk = await JsonSerializer.DeserializeAsync<MistralChatCompletionChunk>(
+                    new MemoryStream(Encoding.UTF8.GetBytes(rawChunk))).ConfigureAwait(false);
```
RogerBarreto, line 87: Any special reason why are you converting the string to a stream? 
AFAIK `DeserializeAsync<T>` accepts string as input.

```suggestion
                var chunk = await JsonSerializer.DeserializeAsync<MistralChatCompletionChunk>(rawChunk).ConfigureAwait(false);
```

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
```diff
@@ -33,52 +34,89 @@ internal sealed class MistralClient
...
+                    {
+                        Data = { { "ResponseData", rawChunk } },
+                    };
+                }
+
+                var role = chunk.GetRole();
+                if (role is not null)
+                {
+                    currentRole = role;
+                }
```
RogerBarreto, line 102:
```suggestion
                currentRole ??= chunk.GetRole();
```

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralChatCompletionChunk.cs
```diff
@@ -1,61 +1,73 @@
...
+            };
+        }
+
+        return this._metadata;
     }
 
     internal AuthorRole? GetRole()
     {
-        throw new NotImplementedException();
+        var role = this.Choices?.FirstOrDefault()?.Delta?.Role;
```
RogerBarreto, line 53: You should get the index of the current choice in the chunk, in a multiple result scenario this will break... 

i.e:
```csharp
    internal AuthorRole? GetRole(int index)
    {
        var role = this.Choices?[index]?.Delta?.Role;
        ...
    }
```

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralChatMessage.cs
```diff
@@ -10,22 +10,22 @@ namespace Microsoft.SemanticKernel.Connectors.MistralAI.Client;
...
     /// </summary>
-    /// <param name="role">Must be one of: system, user, assistant</param>
+    /// <param name="role">If provided must be one of: system, user, assistant</param>
     /// <param name="content">Content of the chat message</param>
     [JsonConstructor]
-    internal MistralChatMessage(string role, string content)
+    internal MistralChatMessage(string? role, string content)
     {
-        if (role != "system" && role != "user" && role != "assistant")
+        if (role is not null && role != "system" && role != "user" && role != "assistant")
```
RogerBarreto, line 26:
```suggestion
        if (role is not null && role is not "system" or "user" or "assistant")
```

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
markwallace-microsoft: @stephentoub  Yes, here's the extract from the documentation

_Whether to stream back partial progress. If set, tokens will be sent as data-only server-sent events as they become available, with the stream terminated by a data: [DONE] message. Otherwise, the server will hold the request open until the timeout or until completion, with the response containing the full result as JSON._

markwallace-microsoft: @Krzysztof318 @RogerBarreto My understanding of the stream parser is it's used to identify each chunk my checking for completed JSON. In the Mistral case I think it's simpler because each chunk is delimited with an empty line.

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
markwallace-microsoft: It only takes a stream

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralChatMessage.cs
markwallace-microsoft: That didn't work but I did clean this up

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
```diff
@@ -89,11 +127,9 @@ internal async Task<IList<ReadOnlyMemory<float>>> GenerateEmbeddingsAsync(IList<
...
     private readonly string _apiKey;
-    private readonly Uri _endpoint;
-    private readonly string _separator;
+    private readonly Uri? _endpoint;
     private readonly HttpClient _httpClient;
     private readonly ILogger _logger;
-    private readonly IStreamJsonParser _streamJsonParser;
 
     private ChatCompletionRequest CreateChatCompletionRequest(string modelId, bool stream, ChatHistory chatHistory, MistralAIPromptExecutionSettings? executionSettings)
     {
```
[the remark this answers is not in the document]
markwallace-microsoft: I've added validation based on what is documented.

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
RogerBarreto: Ah ok, it seems an overkill creating an stream just to use the `DeserializeAsync`, can't we just use the synchronous one and avoid allocating a stream?

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
RogerBarreto: @Krzysztof318 , might worth having a specialized JsonParser for SSE, this [DONE] is something similar to what OpenAI do with theirs... BTW, SSE is a pattern. 😊 (SSEStreamParser)

dotnet/src/IntegrationTests/Connectors/MistralAI/TextEmbedding/MistralAITextEmbeddingTests.cs
```diff
@@ -26,7 +26,7 @@ public MistralAITextEmbeddingTests()
             .Build();
     }
 
-    [Fact(Skip = "This test is for manual verification.")]
+    [Fact] // (Skip = "This test is for manual verification.")
```
RogerBarreto, line 29: Is this intended to be enabled ?

markwallace-microsoft: Yes, I'll keep them enabled in the feature branch otherwise I have to keep enabling and disabling with every PR. But there will be no way t merge to main with these enabled because they will fail so no worries of them ending up in main.

dotnet/src/Connectors/Connectors.MistralAI/Client/MistralClient.cs
markwallace-microsoft: Created this issue to track: https://github.com/microsoft/semantic-kernel/issues/5610

SergeyMenshykh (approved):

PR #5551, opened 2024-03-19
.Net: Remove default chat system prompt

### LINKED ISSUE #5544: By default, the OpenAI connector should not add any system messages.
Today, the OpenAI connector adds "Assistant is a large language model" as a system message if none is provided. This should be removed.

markwallace-microsoft:
### Motivation and Context

Closes #5544 

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
@@ -174,19 +174,13 @@ public sealed class OpenAIPromptExecutionSettings : PromptExecutionSettings
     /// Defaults to "Assistant is a large language model."
     /// </summary>
     [JsonPropertyName("chat_system_prompt")]
-    public string ChatSystemPrompt
+    public string? ChatSystemPrompt
     {
         get => this._chatSystemPrompt;
 
         set
         {
             this.ThrowIfFrozen();
-
-            if (string.IsNullOrWhiteSpace(value))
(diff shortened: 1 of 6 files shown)


dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -174,19 +174,13 @@ public int ResultsPerPrompt
     /// Defaults to "Assistant is a large language model."
     /// </summary>
     [JsonPropertyName("chat_system_prompt")]
-    public string ChatSystemPrompt
+    public string? ChatSystemPrompt
```
stephentoub, line 177: This change makes sense. Just calling out that it could introduce new warnings into someone's project if they're reading this property.

stephentoub (approved):

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
markwallace-microsoft: Yes, expecting the API check to complain about this.

stephentoub: I don't actually know if the API compat tooling understands nullable reference type annotations; it may not. Such changes aren't binary breaking.

RogerBarreto (approved):

SergeyMenshykh (approved):

Issue #5310, opened 2024-03-05
HttpClient 1m40s (100s) timeout 


arynaq:
**Describe the bug**
The builtin httpclient is not configurable, and it falls back to the default behavior of httpclient, which is 100s timeout. This can be unreasonable during load or long context windows.

There is probably some way of circumventing this but libraries such as this that aim to be a high level wrapper should have batteries included. Resiliency has no impact as this is a retry wrapper around the client. See

```
var builder = Kernel.CreateBuilder();
        builder.Services.AddSingleton(_loggerFactory);
        builder.Services.ConfigureHttpClientDefaults(c =>
        {
            c.ConfigureHttpClient((sp, c) =>
            {
                c.Timeout = TimeSpan.FromMinutes(5);
            });
            c.AddResilienceHandler("dumbretry", b =>
            {
                b.AddRetry(new HttpRetryStrategyOptions {
                    MaxRetryAttempts = 3,
                    Delay = TimeSpan.FromSeconds(3),
                    BackoffType = DelayBackoffType.Exponential,
                });
                b.AddTimeout(TimeSpan.FromSeconds(60 * 5));
                b.Build();
            });
        });
```

The client I am configuring here is not the client being used under the hood, the timeout value for HttpClient has no impact at all. 

**To Reproduce**
1. Try GPT4 with a very long context
2. You will eventually get a socket exception bubbling up to SK (socket timeout)

**Expected behavior**
The client I configure is the client being used or I can configure the timeout via SK or resiliency or the default is higher to allow for throttle/load windows in OpenAI.

arynaq:
Nothing? I am literally providing my own http client at this point, but the library is not using it or ends up down some call chain that does not use it, seeing the error below in production is not fun

```
Microsoft.SemanticKernel.KernelFunctionCanceledException: The invocation of function 'func5505e8747a6d4b96ab41ff9f245b88ae' was canceled.
 ---> System.Threading.Tasks.TaskCanceledException: The operation was cancelled because it exceeded the configured timeout of 0:01:40. Network timeout can be adjusted in ClientOptions.Retry.NetworkTimeout.
 ---> System.Threading.Tasks.TaskCanceledException: The operation was canceled.
 ---> System.Threading.Tasks.TaskCanceledException: The operation was canceled.
 ---> System.IO.IOException: Unable to read data from the transport connection: Operation canceled.
 ---> System.Net.Sockets.SocketException (125): Operation canceled
```

```
[...]

PR #5553, opened 2024-03-19
.Net: Disable Azure SDK network timeout when a custom HttpClient is supplied

stephentoub:
Fixes https://github.com/microsoft/semantic-kernel/issues/5310

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
@@ -749,6 +749,7 @@ internal abstract class ClientCore
         {
             options.Transport = new HttpClientTransport(httpClient);
             options.RetryPolicy = new RetryPolicy(maxRetries: 0); // Disable Azure SDK retry policy if and only if a custom HttpClient is provided.
+            options.Retry.NetworkTimeout = Timeout.InfiniteTimeSpan; // Disable Azure SDK default timeout
         }
 
         return options;


RogerBarreto (approved):

markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -749,6 +749,7 @@ internal static OpenAIClientOptions GetOpenAIClientOptions(HttpClient? httpClien
         {
             options.Transport = new HttpClientTransport(httpClient);
             options.RetryPolicy = new RetryPolicy(maxRetries: 0); // Disable Azure SDK retry policy if and only if a custom HttpClient is provided.
+            options.Retry.NetworkTimeout = Timeout.InfiniteTimeSpan; // Disable Azure SDK default timeout
```
SergeyMenshykh, line 752: I'm a little bit late with the review but I would like to point out that there's a similar code over here - [AzureOpenAITextToImageService GetClientOptions](https://github.com/microsoft/semantic-kernel/blob/51ee30fcbd438472ec45f90f68be77a591e04527/dotnet/src/Connectors/Connectors.OpenAI/TextToImage/AzureOpenAITextToImageService.cs#L148) that needs to be updated as well.

markwallace-microsoft: Created this https://github.com/microsoft/semantic-kernel/pull/5569

PR #5568, opened 2024-03-20
Python: .Net Gemini - internalize plugin extensions

Krzysztof318:
This reverts commit ad1175d6

cc: @RogerBarreto


RogerBarreto:
Use this PR later to internalize both those extension classes, and with the internalization they will be gone.

PR #5574, opened 2024-03-20
.Net Stream Json Parser as Utility for Connectors.

RogerBarreto:
### Motivation and Context

As a proven useful component created by @Krzysztof318 in the Gemini Connector implementation, I'm bringing this as a Utility that can be used internally across different connectors for streaming deserialization of SSE and Non-SSE responses.

- Hugging Face updates using the utility
- Unit tests moved to SemanticKernel.UnitTests for the tool
- StreamJsonParser tests removed from HuggingFace UnitTests
- Added extra Example using streaming with HuggingFace Zephyr model.


dotnet/src/InternalUtilities/src/Text/StreamJsonParser.cs
```diff
@@ -0,0 +1,222 @@
...
+/// <summary>
+/// Internal class for parsing a stream of text which contains a series of discrete JSON strings into en enumerable containing each separate JSON string.
+/// </summary>
+/// <remarks>
+/// This class is thread-safe.
+/// </remarks>
+internal sealed class StreamJsonParser
+{
+    /// <summary>
+    /// Parses a Stream containing JSON data and yields the individual JSON objects.
```
stephentoub, line 26: @eiriktsarpalis, FYI

dotnet/src/InternalUtilities/src/Text/StreamJsonParser.cs
```diff
@@ -0,0 +1,222 @@
...
+    /// </summary>
+    /// <param name="stream">The Stream containing the JSON data.</param>
+    /// <param name="validateJson">Set to true to enable checking json chunks are well-formed. Default is false.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>An enumerable collection of string representing the individual JSON objects.</returns>
+    /// <remarks>Stream will be disposed after parsing.</remarks>
+    public async IAsyncEnumerable<string> ParseAsync(
```
stephentoub, line 33: Maybe simplify the implementation to something more along the lines of this?
https://github.com/dotnet/runtime/issues/33030#issuecomment-1524227075

Krzysztof318: @stephentoub I checked this one and performance is similar, your solution even use less memory, so I eat my words, but it works only with cleaned input, so we can't use this with sse.
https://github.com/Krzysztof318/StreamParserBenchmark

dotnet/src/InternalUtilities/src/Text/StreamJsonParser.cs
```diff
@@ -0,0 +1,222 @@
...
+/// <remarks>
+/// This class is thread-safe.
+/// </remarks>
+internal sealed class StreamJsonParser
+{
+    /// <summary>
+    /// Parses a Stream containing JSON data and yields the individual JSON objects.
+    /// </summary>
+    /// <param name="stream">The Stream containing the JSON data.</param>
+    /// <param name="validateJson">Set to true to enable checking json chunks are well-formed. Default is false.</param>
```
eiriktsarpalis, line 31: What type of inputs does this method accept when `validateJson` is set to false? As far as I can tell it is implementing its own JSON parser, so clearly some structure is necessary for this to identify the end of the next element.

eiriktsarpalis: Put differently, would exposing some lax parsing mode in `Utf8JsonReader` remove the need for you to implement a bespoke parser?

Krzysztof318: StreamJsonparser only counts { and } and returns string when counter will be 0. validateJson = true additionally invoke JsonNode.Parse after completing json string.

Krzysztof318: I tried use Utf8JsonReader in early version of parser but often getting parser exceptions.

eiriktsarpalis: > I tried use Utf8JsonReader in early version of parser but often getting parser exceptions.

Can you share examples of JSON data that is causing `Utf8JsonReader` to fail?

eiriktsarpalis: > StreamJsonparser only counts { and } and returns string when counter will be 0. validateJson = true additionally invoke JsonNode.Parse after completing json string.

That might be a quick way to split documents, but given that you expect the input data to be invalid JSON, one misplaced curly brace could result in erroneous groupings spilling over to the entire data stream. cc @grabyourpitchforks

dotnet/src/InternalUtilities/src/Text/StreamJsonParser.cs
stephentoub: Thanks.

> but it works only with cleaned input

Can you elaborate? What can't that approach work with?

Krzysztof318: @stephentoub if input contains sse data like  `event: any-context\ndata: { "person": { "name": "example" } }` or any other not json data like `data: { "person": { "name": "example" } } data done` then `JsonDocument.TryParseValue` will fail

dotnet/src/InternalUtilities/src/Text/StreamJsonParser.cs
Krzysztof318: @eiriktsarpalis I don't remember exactly, I can't find first version of parser in my git history. But probably exception from Utf8JsonReader occurred if current slice contained small amount of data like `,{ "d` then was throwing exception not recognized TokenType, but I am not sure now how it was...

Not exactly invalid input, this Parser is designed to ignore SSE tags and return only json data. And for example if input will contain `data: } { "person": { "name": "example" } }` exception will be thrown

Krzysztof318:
Azure SDK also reads SSE line by line https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/src/Helpers/SseReader.cs

cc: @RogerBarreto @stephentoub 

We could use same implementation but then parser will support only SSE.

dotnet/src/InternalUtilities/src/Text/StreamJsonParser.cs
eiriktsarpalis: > But probably exception from Utf8JsonReader occurred if current slice contained small amount of data like ,{ "d then was throwing exception not recognized TokenType, but I am not sure now how it was...

That particular case suggests that `Utf8JsonReader` wasn't configured to support partial data (this can be done via the [`isFinalBlock` constructor parameter](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.-ctor?view=net-8.0#system-text-json-utf8jsonreader-ctor(system-buffers-readonlysequence((system-byte))-system-boolean-system-text-json-jsonreaderstate))). The second example is blatantly invalid JSON and `Utf8JsonReader` won't accept it.

Krzysztof318: I had isFinalBlock set to false, for sure, but I will try to reproduce problematic code. I'll get back to you later.

Krzysztof318: @eiriktsarpalis @stephentoub okay I can't reproduce same code... So it's probably that was any my fault exception was throwing when parsing but JsonDocument parser still doesn't solve from with SSE input where data contains other data than json. That's why my parser is more complicated, implementation from docs also is not so simple https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-utf8jsonreader#read-from-a-stream-using-utf8jsonreader

stephentoub: What's the situation where non-json support is needed? And isn't this still based on looking for bonds beginning/ending with braces?

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #5579, opened 2024-03-20
Java: Fix function invocation argument name assigning. Fix templating to be…

johnoliver:
… more customisable.

### Contribution Checklist

### DIFF
--- a/java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/documentationexamples/Prompts.java
+++ b/java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/documentationexamples/Prompts.java
@@ -1,3 +1,4 @@
+// Copyright (c) Microsoft. All rights reserved.
 package com.microsoft.semantickernel.samples.documentationexamples;
 
 import com.azure.ai.openai.OpenAIAsyncClient;
@@ -13,7 +14,8 @@ public class Prompts {
 
     // Only required if AZURE_CLIENT_KEY is set
     private static final String CLIENT_ENDPOINT = System.getenv("CLIENT_ENDPOINT");
-    private static final String MODEL_ID = System.getenv().getOrDefault("MODEL_ID", "gpt-35-turbo-2");
+    private static final String MODEL_ID = System.getenv().getOrDefault("MODEL_ID",
+        "gpt-35-turbo-2");
 
     public static void main(String[] args) {
         System.out.println("======== Prompts ========");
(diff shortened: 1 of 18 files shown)


java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/documentationexamples/Prompts.java
```diff
@@ -13,30 +14,31 @@ public class Prompts {
 
     // Only required if AZURE_CLIENT_KEY is set
     private static final String CLIENT_ENDPOINT = System.getenv("CLIENT_ENDPOINT");
-    private static final String MODEL_ID = System.getenv().getOrDefault("MODEL_ID", "gpt-35-turbo-2");
+    private static final String MODEL_ID = System.getenv().getOrDefault("MODEL_ID",
+        "gpt-35-turbo-2");
```
dsgrieve, line 18: If we're going to go with Azure OpenAI model ids as the default, then we need to be consistent with all the examples. And we need to state in the readme that these are Azure OpenAI model ids and if you use OpenAI model ids, then you need to set MODEL_ID. I can take this on in a separate PR if you'd like. 

There is also the issue of different models. One example uses both chat completion and text embedding. Some of the examples are chat others text. Maybe we should have CHAT_MODEL_ID, EMBEDDING_MODEL_ID, IMAGE_MODEL_ID, etc.

java/samples/sample-code/src/main/java/com/microsoft/semantickernel/samples/documentationexamples/SerializingPrompts.java
```diff
@@ -17,29 +18,30 @@ public class SerializingPrompts {
 
     // Only required if AZURE_CLIENT_KEY is set
     private static final String CLIENT_ENDPOINT = System.getenv("CLIENT_ENDPOINT");
-    private static final String MODEL_ID = System.getenv().getOrDefault("MODEL_ID", "gpt-35-turbo-2");
+    private static final String MODEL_ID = System.getenv().getOrDefault("MODEL_ID",
+        "gpt-35-turbo-2");
```
dsgrieve, line 22: gpt-35-turbo-2 is our deployment name, not a model name. I know this is the correct value since the deployment picks the model. But it's confusing because there is no "gpt-35-turbo-2" in the Azure OpenAI service models (at least not that I can see). 

"For Azure OpenAI, the model name must match the deployment name in Azure OpenAI Studio > Deployments." is another thing to add to the readme.

dsgrieve (approved):

Issue #5583, opened 2024-03-20
Python: Action Planner breaking when parameters are not explicitly given in the ask


ArfiArfath21:
**Describe the bug**
When we change the ask in the action planner example to something that doesnt explicitly mention the parameters, it fails to generate any response and breaks down.

**To Reproduce**
1. Go to 'semantic-kernel/python/samples/kernel-syntax-examples/action_planner.py'
2. change the "ask" in line 29 to something generic. Ex: ask="What is the date today?"
3. Run the script.

**Expected behavior**
It should have given a proper response like 

**Screenshots**
Error:
![image](https://github.com/microsoft/semantic-kernel/assets/76872284/0c3247ff-d069-427f-ab7e-9a017f4e8dd1)
Expected:
![image](https://github.com/microsoft/semantic-kernel/assets/76872284/a1fe357c-6cac-4d86-9f58-ada03d62c359)

**Platform**
 - OS: MacOS Sonoma 14.3.1
 - IDE: VS Code
 - Language: Python

moonbox3:
Closing as the fix is in main. Will be included in the next release.

PR #5584, opened 2024-03-20
Python: Fix - Action Planner breaking when parameters are not explicitly given in the ask

ArfiArfath21:
Solution to the issue: Python: Action Planner breaking when parameters are not explicitly given in the ask
https://github.com/microsoft/semantic-kernel/issues/5583

### Motivation and Context

  1. What problem does it solve? 
A. The action planner breaks when the user does not provide specific parameters in the ask/goal
  2. What scenario does it contribute to? 
A. In cases when user may not explicitly mention any parameters in the ask/goal
  3. If it fixes an open issue, please link to the issue here. 
A. It fixes the issue https://github.com/microsoft/semantic-kernel/issues/5583

### Description

Just adding an IF condition before iterating through the generated_plan variable to check if parameters are present or not does the job

### DIFF
--- a/python/semantic_kernel/planners/action_planner/action_planner.py
+++ b/python/semantic_kernel/planners/action_planner/action_planner.py
@@ -148,7 +148,6 @@ class ActionPlanner:
             )
             plan = Plan(description=goal, function=function_ref)
 
-
         if "parameters" in generated_plan['plan']:
             for key, val in generated_plan["plan"]["parameters"].items():
                 logger.info(f"Parameter {key}: {val}")
--- a/python/tests/unit/planners/action_planner/test_action_planner.py
+++ b/python/tests/unit/planners/action_planner/test_action_planner.py
@@ -5,20 +5,24 @@ from unittest.mock import MagicMock, Mock
 
 import pytest
 
+import semantic_kernel as sk
 from semantic_kernel import Kernel
-from semantic_kernel.connectors.ai.prompt_execution_settings import PromptExecutionSettings
-from semantic_kernel.exceptions import (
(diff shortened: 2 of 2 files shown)


ArfiArfath21:

moonbox3:
Hi @ArfiArfath21, thank you for your contribution. We're actually retiring the action planner in the following PR that is open. https://github.com/microsoft/semantic-kernel/pull/5586 FYI.

ArfiArfath21:
Hi @moonbox3, I get that the legacy Stepwise planner is being retired because of a newer version. Why is the action planner being retired though? It actually is an useful feature to have right?

moonbox3:
Hi @ArfiArfath21, could you please explain your use-case for the action planner and why you chose is versus something like the function calling stepwise planner? Using OpenAI function calling is more accurate, uses less tokens, and is faster.

ArfiArfath21:
> Hi @ArfiArfath21, could you please explain your use-case for the action planner and why you chose is versus something like the function calling stepwise planner? Using OpenAI function calling is more accurate, uses less tokens, and is faster.

I have a chatbot application which basically offers multiple different services in it. Depending on the user query/message it has to identify one best course of action (which can be done with action planner) and take the subsequent steps as necessary. Sort of like a classification problem.

moonbox3:
@ArfiArfath21 thanks for your response. Have you tried the new function calling stepwise planner? It's performing the same operations as the action planner, but wrapping the plan create as part of the invoke. It uses the available plugins/functions in the kernel and will choose the best one depending upon the user query/question.

Issue #5278, opened 2024-03-03
.Net: Support DI for memory connectors


SebastianStehle:
In .NET Core some patterns have evolved how to use services. It is widely used in ASP.NET Core, Orleans and many third party services. In my opinion Semantic Kernel should also follow this pattern.

There are just a wide range of small mistakes, that should be avoided:

1. Inject loggers as concrete classes, aka `ILogger<T>`
2. Inject http client using `IHttpClientFactory`, do not use singleton clients.
3. Don't use custom factories, e.g. `MemoryBuilder`
4. Inject configuration using the `IOptions<T>` interface.
5. Use named registrations if you allow multiple registration.
6. Allow easy chaining, do not use `builder.Plugins.AddFromType<T>()` but `builders.AddPluginFromType<T>()`

Community PR which added DI to old Memory connectors: https://github.com/microsoft/semantic-kernel/pull/5589

stephentoub:
> Inject loggers as concrete classes, aka ILogger<T>

Where do you see this being a mistake?

> Inject http client using IHttpClientFactory, do not use singleton clients.

By default it does use registered services, but for convenience for various scenarios it allows a singleton, which is fine, eg
https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use

> Don't use custom factories, e.g. MemoryBuilder

KernelBuilder is there to simplify situations where there isn't a higher-level DI container with which you're integrating. MemoryBuilder hasn't been revised yet to follow the desired scheme and is still marked [Experimental].

> Use named registrations if you allow multiple registration.

It supports keyed services. What specifically is your concern?

> Allow easy chaining

Fluency is enabled in some cases. Cases where it isn't was specifically on recommendation from key ASP.NET devs who suggested multiple API studies had found aspects of it to be confusing.

SebastianStehle:
I think my main feedback is around the memory builder. I know that it is experimental, but it does not feel integrated yet. But perhaps I was just too impatient ;)

>> Use named registrations if you allow multiple registration.
>
> It supports keyed services. What specifically is your concern?

It is also related to the MemoryBuilder. I thought it could e made obsolete or implemented similar to the HttpClientFactory.

stephentoub:
> I think my main feedback is around the memory builder. 

Got it. Yes, that needs to be revised.
cc: @dmytrostruk

dmytrostruk:
[...]

PR #5589, opened 2024-03-20
.Net: Making memory stores more configurable via dependency injection

aaronpowell:
### Motivation and Context

My goal here is to improve the interoperability with .NET Aspire on how we can work with memory using DI, addressing #5591.

### Description

This PR adapts the `MemoryBuilder` to be more like `KernelBuilder` in that is uses an `IServiceCollection` for resolving the dependencies, rather than capturing them as private fields.

I've then followed the similar pattern that is used by the `Kernel`/`IKernel` and associated plugins for registering the types with an existing `IServicesCollection`.

For example, you can now write the following:

```csharp
// Register the `SemanticTextMemory` instance in a container
builder.AddSemanticTextMemory();

// Register a Npgsql data source using Aspire integration
builder.AddNpgsqlDataSource(builder => builder.UseVector());

// Add Postgres as the memory store
builder.Services.AddPostgresMemoryStore();

// Add the AOAI Text Embedding service
builder.AddAzureOpenAIClient("aoai-from-aspire");
builder.Services.AddAzureOpenAITextEmbeddingGeneration("text-embedding-ada-002");
```

Elsewhere in the codebase we can just resolve `ISemanticTextMemory` and everything is resolved via DI and Aspire integration.

For most of the memory stores I have added extension methods, the only ones I haven't done are `Sqlite` and `DuckDB` as they don't have a public constructor, instead you need to asynchronously load them.

### Contribution Checklist


markwallace-microsoft:
@aaronpowell Thanks for creating this PR. We are going to take this as a requirement for the new Memory connector work we have in progress. I'm going to close this PR as we won't take the changes in their current form.

If you want to learn more about our plans for Memory connectors please take a look here: https://github.com/orgs/microsoft/projects/866/views/59?sliceBy%5Bvalue%5D=.net and specifically here: https://github.com/microsoft/semantic-kernel/issues/5278

PR #5605, opened 2024-03-21
.Net: introduce ApiManifestPluginParameters to support multiple API dependencies

### LINKED ISSUE #5603: .Net: [API Manifest] Introduce ApiManifestPluginParameters to cover multiple API dependencies scenario
**Describe the bug**
Current plugin generation from API Manifest relies on a single `OpenApiFunctionExecutionParameters` object, which contains authCallback, serverUrlOverride etc. An API Manifest file can contain multiple API dependencies with different `OpenApiFunctionExecutionParameters`. We need a wrapper that contains multiple `OpenApiFunctionExecutionParameters`.

zengin:
### Motivation and Context

- fixes #5603 

### Description
- introduces ApiManifestPluginParameters with three parameters:
  - `HttpClient` and `UserAgent` to be used in plugin initialization phase (to fetch OpenAPI documents)
  - `FunctionExecutionParameters`: A mapping of api dependency name to `OpenApiFunctionExecutionParameters`, so that we can introduce multiple API dependencies in a single API Manifest plugin.

- adds a test case where we connect to two different APIs using a single plugin:
  - Microsoft Graph authenticating with MSAL
  - NASA API authenticating with API key

### Contribution Checklist

  - well technically this is a breaking change to the public method signature, however, we haven't published the package yet and it was behind an experimental flag :)


zengin:
@markwallace-microsoft @RogerBarreto could you please review when you get a chance? Thank you!

dotnet/src/Functions/Functions.OpenApi.Extensions/Extensions/ApiManifestKernelExtensions.cs
```diff
@@ -28,17 +28,17 @@ public static class ApiManifestKernelExtensions
...
-    /// <param name="executionParameters">Optional execution parameters for the plugin.</param>
+    /// <param name="pluginParameters">Optional parameters for the plugin setup.</param>
     /// <param name="cancellationToken">Optional cancellation token.</param>
     /// <returns>The imported plugin.</returns>
     public static async Task<KernelPlugin> ImportPluginFromApiManifestAsync(
         this Kernel kernel,
         string pluginName,
         string filePath,
-        OpenApiFunctionExecutionParameters? executionParameters = null,
+        ApiManifestPluginParameters? pluginParameters = null,
```
markwallace-microsoft, line 38: This (and the change below) look like a breaking change in the public API. Can we keep the original methods and add new methods with the new behaviour and have the old method call the new one?

zengin: @markwallace-microsoft, I'd be happy to apply suggested changes, but I would like to double check before doing so: this extensions package hasn't been published yet. Do we have the same policy around breaking changes for unpublished packages? We would like to deprecate the old signature as the new signature better fits the spec.

markwallace-microsoft (approved):

PR #5627, opened 2024-03-25
.Net: Remove use of Console.SetOut as it impacts other tests

markwallace-microsoft:
### Motivation and Context

Calls to `Console.SetOut(this._testOutputHelper);` impact tests that are using `Console` during tear down. 
Removing these calls as the console output is already being captured in `RedirectOutput`.

### Contribution Checklist


stephentoub (approved):

dotnet/src/IntegrationTests/Connectors/OpenAI/ChatHistoryTests.cs
```diff
@@ -28,7 +28,6 @@ public ChatHistoryTests(ITestOutputHelper output)
     {
         this._logger = new XunitLogger<Kernel>(output);
         this._testOutputHelper = new RedirectOutput(output);
-        Console.SetOut(this._testOutputHelper);
```
stephentoub, line 31: If this isn't done, is _testOutputHelper used for anything?

SergeyMenshykh (approved):

stephentoub:
> Removing these calls as the console output is already being captured in RedirectOutput.

Maybe I'm misunderstanding, but isn't RedirectOutput just a tee that forwards to both stdout and the ITestOutputHelper? If nothing writes to that RedirectOutput, then it's not actually doing anything, and it seems like the purpose of the Console.SetOut was to ensure that Console.Write calls would end up in both places.

RogerBarreto (approved):

stephentoub:
> but during stand-up we discussed this and decided to remove the unneeded the Console.SetOut calls because this pattern was likely being repeated when not strictly needed. There may be more code I can delete if there are places where the RedirectOutput isn't needed at all, I'll create a follow up PR to this one to fix that.

Ok

PR #5635, opened 2024-03-25
.Net: Remove JsonSchema.Net dependency from Microsoft.SemanticKernel.Abstractions/Core

stephentoub:
This removes the dependency, substituting in a lighter-weight schema generator that better aligns with System.Text.Json and which is likely to be built-in to STJ in a future release.  All of these get removed:
image

The JsonSchema.Net dependency still exists from Microsoft.SemanticKernel.Plugins.OpenApi in support of the validation used in https://github.com/microsoft/semantic-kernel/blob/9ab95132b5f460f1bf9a1d1e387fb18453a037f4/dotnet/src/Functions/Functions.OpenApi/Extensions/RestApiOperationResponseExtensions.cs#L48-L51
and we can subsequently decide what to do about that.

Note that the replacement highlighted some important issues with the schema previously being generated. In particular, it didn't align with how state would actually be serialized/deserialized, e.g. the schema included public fields even though the JsonSerializerOptions being used would ignore such fields.

This does incur one notable reduction in functionality: today KernelJsonSchema validates that supplied text is indeed a valid JSON schema... with this change, it only validates that it's valid JSON. I think that's a reasonable tradeoff for now, and we should be able to add back the stricter validation in the future when STJ provides support for it. I don't have a good alternative right now other than keeping the significant dependency.


stephentoub:
cc: @eiriktsarpalis

dotnet/src/Experimental/Agents/Extensions/AssistantsKernelFunctionExtensions.cs
```diff
@@ -86,4 +85,17 @@ private static string ConvertType(Type? type)
 
         return type.Name;
     }
+
+    private static bool IsNumber(this Type type) =>
```
eiriktsarpalis, line 89: Is this being used anywhere?

stephentoub: 13 lines above

eiriktsarpalis: Oh, so is these replacing some helper from JsonSchema.NET?

eiriktsarpalis (approved):

dotnet/src/Experimental/Agents/Extensions/AssistantsKernelFunctionExtensions.cs
stephentoub: Yup.

eiriktsarpalis: If this is being used to generate schemas, it should at some point be delegated to STJ. It's leaving out the new numeric types and custom converters can change what is regarded as a number and what isn't.

stephentoub: > it should at some point be delegated to STJ

What's the recommended way to do that?

eiriktsarpalis: There are no public APIs available right now for this, but if you wanted you could call the schema mapper and look up the `type` property of the resultant document.

dotnet/src/SemanticKernel.UnitTests/Functions/KernelParameterMetadataTests.cs
```diff
@@ -49,9 +49,9 @@ public void ItInfersSchemaFromType()
         Assert.Equal(JsonSerializer.Serialize(KernelJsonSchema.Parse("{ \"type\":\"number\" }")), JsonSerializer.Serialize(new KernelParameterMetadata("p") { ParameterType = typeof(double) }.Schema));
         Assert.Equal(JsonSerializer.Serialize(KernelJsonSchema.Parse("{ \"type\":\"string\" }")), JsonSerializer.Serialize(new KernelParameterMetadata("p") { ParameterType = typeof(string) }.Schema));
         Assert.Equal(JsonSerializer.Serialize(KernelJsonSchema.Parse("{ \"type\":\"boolean\" }")), JsonSerializer.Serialize(new KernelParameterMetadata("p") { ParameterType = typeof(bool) }.Schema));
-        Assert.Equal(JsonSerializer.Serialize(KernelJsonSchema.Parse("{ \"type\":\"object\" }")), JsonSerializer.Serialize(new KernelParameterMetadata("p") { ParameterType = typeof(object) }.Schema));
+        Assert.Equal(JsonSerializer.Serialize(KernelJsonSchema.Parse("{ }")), JsonSerializer.Serialize(new KernelParameterMetadata("p") { ParameterType = typeof(object) }.Schema));
```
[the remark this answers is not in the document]
eiriktsarpalis: The old schema reflects JsonSchema.NET generation semantics, which maps `object` to schemas of type `"object"`. However this is incorrect as arbitrary .NET objects can serialize to any JSON type.

dotnet/src/InternalUtilities/src/Schema/JsonSchemaMapper.cs
```diff
@@ -0,0 +1,831 @@
...
+        [typeof(Int128)] = (JsonSchemaType.Integer, null),
+#endif
+        [typeof(char)] = (JsonSchemaType.String, null),
+        [typeof(string)] = (JsonSchemaType.String, null),
+        [typeof(byte[])] = (JsonSchemaType.String, null),
+        [typeof(Memory<byte>)] = (JsonSchemaType.String, null),
+        [typeof(ReadOnlyMemory<byte>)] = (JsonSchemaType.String, null),
+        [typeof(DateTime)] = (JsonSchemaType.String, Format: "date-time"),
+        [typeof(DateTimeOffset)] = (JsonSchemaType.String, Format: "date-time"),
+        [typeof(TimeSpan)] = (JsonSchemaType.String, Format: "time"),
```
eiriktsarpalis, line 788: Would it be possible to update the implementation to the most recent version? I recently pushed some changes that fix the schema for `TimeSpan`.

stephentoub:
@markwallace-microsoft, @SergeyMenshykh, ready for another look

dotnet/src/SemanticKernel.UnitTests/Functions/KernelParameterMetadataTests.cs
```diff
@@ -140,7 +141,9 @@ public void ItInvalidatesSchemaForNewDefaultValue()
     internal sealed class Example
```
eiriktsarpalis, line 141: Type-level annotations should be supported as well, in case you wanted to test for that.

stephentoub: Thanks

SergeyMenshykh (approved):

markwallace-microsoft:
@stephentoub Thanks, once conflict is resolved we'll re-approve

stephentoub:
> @stephentoub Thanks, once conflict is resolved we'll re-approve

Resolved again

markwallace-microsoft (approved):

PR #5639, opened 2024-03-25
.Net: Adding to SerializerOptions property to Kernel

SergeyMenshykh:
### Motivation and Context
Today, SK has several locations where serialization/deserialization takes place. In almost all of these places, it is not possible to control the serialization/deserialization process via JsonSerializerOptions, as there is no way to pass the options to these locations. This PR addresses the issue by adding a SerializerOptions property to the Kernel class, allowing it to be used to control serialization/deserialization wherever the kernel is available.

### Description
1. Adds the SerializerOptions property to the Kernel class.
2. Obsoletes the ToolCallBehavior.ToolCallResultSerializerOptions that was originally introduced to serialize function results.

### Contribution Checklist


markwallace-microsoft (approved):

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.UnitTests/Functions/KernelFunctionFromMethodTests1.cs
```diff
@@ -1276,38 +1225,68 @@ public async Task ItShouldNotDeserializeIfParameterTypeAndArgumentTypeAreSameAsy
...
+    [InlineData(typeof(JsonNode))]
+    [InlineData(typeof(JsonDocument))]
+    [InlineData(typeof(JsonElement))]
+    public async Task ItCanUseKernelSerializerOptionsForArgumentsDeserializationAsync(Type paramType)
     {
         // Arrange
-        var thirdPartyJsonPrimitive = new ThirdPartyJsonPrimitive(@"{""id"":28}");
+        object? paramValue = paramType switch
+        {
+            Type t when t == typeof(string) => @"{""id"":28}",
```
RogerBarreto, line 1267: Can we use a constant for `"{\"id\":28}"` ?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -176,6 +177,12 @@ public CultureInfo Culture
         set => this._culture = value ?? CultureInfo.InvariantCulture;
     }
 
+    /// <summary>
+    /// Gets the Json serializer options currently associated with this <see cref="Kernel"/>.
+    /// </summary>
+    [Experimental("SKEXP0001")]
+    public JsonSerializerOptions? SerializerOptions { get; set; }
```
RogerBarreto, line 184: Should we have a collection of serializeroptions for different scenarios where its used?

```csharp
public KernelSerializerOptions? SerializerOptions { get; set; }

public KernelSerializerOptions {
    public JsonSerializerOptions Scenario1 { get; set; }
    public JsonSerializerOptions Scenario2 { get; set; }
    ...
}
```

dotnet/src/SemanticKernel.UnitTests/Functions/KernelFunctionFromMethodTests1.cs
SergeyMenshykh: Done

dotnet/src/SemanticKernel.UnitTests/Functions/KernelFunctionFromMethodTests1.cs
```diff
@@ -1318,7 +1297,29 @@ private sealed class CustomTypeForJsonTests
...
+            return this._expectedId;
+        }
+
+        public override void Write(Utf8JsonWriter writer, int value, JsonSerializerOptions options)
+        {
+            writer.WriteNumberValue(value);
+        }
+    }
+
+#pragma warning disable CA1812 // Avoid uninstantiated internal classes
```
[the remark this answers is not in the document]
SergeyMenshykh: I reintroduced the test that6 used it.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
stephentoub, line 184: @eiriktsarpalis, can you share insights here on what you'd expect to be passed throughout a system like this?

SergeyMenshykh: Ideally, we should use only one that would cover all scenarios. Over the time, when using the property in different SK components we may identify certain scenarios that need separate {de}serialization options.

RogerBarreto (approved):

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
eiriktsarpalis: JsonSerializerOptions controls every aspect of serialization, including what types can be serialized (in the context of source generators).

@SergeyMenshykh assuming that this options instance is used to serialize types that are internal to SK, this could create problems if and when you'd be looking to add Native AOT support: user-supplied source generated JSO instances cannot support SK internals.

ASP.NET Core addresses this issue by making their options properties get-only (ensuring that they are created and owned by them) and then apply fix-up logic to make sure that the instance works with their own internal models.

stephentoub: @eiriktsarpalis, we'll need to enable consumers to supply their own and then presumably merge with internal information in order to serialize both user and internal types. What is the recommended way to do that? I don't understand how making this property get-only would enable that. What shoukdbe exposed here to allow a consumer to contribute type information?

eiriktsarpalis: > What is the recommended way to do that?

The `JsonSerializerOptions.TypeInfoResolverChain` property is an accelerator API that lets you combine multiple source generated contexts in one instance. ASP.NET uses that API to append their own stuff after users have finished configuring the object.

> I don't understand how making this property get-only would enable that.

The property can be set to shared, read-only instances such as `JsonSerializerOptions.Default`. Arguably this can be worked around using the `JsonSerializerOptions` copy constructor and applying post processing steps directly inside the setter body, but a getter-only approach better communicates that the particular options (and its associated caches) are owned by the `Kernel` instance.

> What should be exposed here to allow a consumer to contribute converter and type information?

I don't think we need to expose anything other than the APIs already present in the `JsonSerializerOptions` itself.

eiriktsarpalis: I'm saying that this should be designed in way where users can configure a JSO instance that is owned by the kernel, but which can then be followed by a post-processing phase where source generated contexts defined internally in SK can be appended to user configuration. I don't think this could be achieved effectively with a property that can be mutated at any point in the lifetime of the kernel instance. ASP.NET Core does it using their configuration builder APIs.

Of course, I could be completely wrong in my original assumption that internal SK types are being serialized using this configuration in which case it doesn't matter a lot.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -88,6 +93,12 @@ public sealed class Kernel
         {
             this._promptFilters = new(promptFilters);
         }
+
+        // Copy the original options to prevent them from being mutated.
```
SergeyMenshykh, line 97: @eiriktsarpalis thanks for helping to identify the right solution for JSO in SK. Please review it when you have time.

SergeyMenshykh:
@stephentoub , we have the [JsonOptionsCache](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/InternalUtilities/src/Text/JsonOptionsCache.cs) cache, which is used by many SK components. I think it should be removed in favor of the `Kernel.SerializerOptions` property. The configuration of the cache can be moved to the kernel class and applied as the default one, and all components can be updated to use the `Kernel.SerializerOptions` instead. What do you think? 

@eiriktsarpalis what would be your recommendations in case some SK components need a slight deviation from the default options represented by `Kernel.SerializerOptions` property?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -46,12 +49,14 @@ public sealed class Kernel
...
     /// </param>
+    /// <param name="serializerOptions">The <see cref="JsonSerializerOptions"/> options.</param>
     /// <remarks>
     /// The KernelBuilder class provides a fluent API for constructing a <see cref="Kernel"/> instance.
     /// </remarks>
     public Kernel(
         IServiceProvider? services = null,
-        KernelPluginCollection? plugins = null)
+        KernelPluginCollection? plugins = null,
+        JsonSerializerOptions? serializerOptions = null)
```
SergeyMenshykh, line 59: This is a breaking change that can be worked around by adding the property setter instead:
```csharp
/// <summary>
/// Gets the Json serializer options currently associated with this <see cref="Kernel"/>.
/// </summary>
[Experimental("SKEXP0001")]
public JsonSerializerOptions SerializerOptions
{
    get
    {
        if (this._serializerOptions is null)
        {
            this._serializerOptions = new JsonSerializerOptions(this.Services.GetService<JsonSerializerOptions>() ?? JsonSerializerOptions.Default);
            this._serializerOptions.MakeReadOnly(true);
        }

        return this._serializerOptions;
    }
    set
    {
        if (this._serializerOptions is not null)
        {
            throw new InvalidOperationException("Serializer options can be initialized once either explicitly via the SerializerOptions property setter, or implicitly via the property getter.");
        }

        this._serializerOptions = new JsonSerializerOptions(value);
        this._serializerOptions.MakeReadOnly(true);
    }
}
```
@eiriktsarpalis , @stephentoub thoughts?

eiriktsarpalis: > This is a breaking change

You could try adding a second constructor overload that accepts `JsonSerializerOptions` as a nullable, non-optional parameter.

Regarding the suggested workaround in the setter:

* Having the getter implicitly populate a default JSO and additionally disabling subsequent sets on that basis seems a bit error prone to me. My personal preference would be to have the instance resolved by the constructor.
* I would recommend using the copy constructor sparingly and only cases where you would want to make modifications to a read-only instance provided by the user. The reason is that each `JsonSerializerOptions` instance defines its own sets of serialization caches, so creating a copy will eventually force an expensive recomputation. 

I would probably write it as follows, roughly:
```C#
[Experimental("SKEXP0001")]
public JsonSerializerOptions SerializerOptions
{
    get
    {
        // instance resolved by the constructor
        return this._serializerOptions;
    }
    set
    {
        if (!value.TypeInfoResolverChain.Any(resolver => resolver is not DefaultJsonTypeInfoResolver))
        {
             // User gave us an instance that doesn't use reflection.
             // Need to register resolvers for our internal types.
             if (value.IsReadOnly)
             {
                   // Use gave us a read-only instance.
                   // Need to use a copy that we own and modify.
                   value = new(value);
             }

             value.TypeInfoResolverChain.Add(SemanticKernelContext.Default);
        }

        value.MakeReadOnly(); // prevent future modifications
        this._serializerOptions = value;
    }
}
```

SergeyMenshykh: I wonder, should we not always use the copy constructor to prevent the mutation of options provided by users? For example, calling the MakeReadOnly() function on options supplied by the user will make them read-only, which might not be what the hosting app or user expects.

eiriktsarpalis: > calling the MakeReadOnly() function on options supplied by the user will make them read-only, which might not be what the hosting app or user expects.

It's something that happens implicitly whenever you pass the JSO instance to one of the `JsonSerializer` methods anyway, so using it at the setter is just a way to make its freezing predictable.

stephentoub: > This is a breaking change

The typical way to evolve such things is to change the existing member to not have any optional arguments, and then add a new one where everything is optional, e.g.
```C#
public Kernel(
        IServiceProvider? services,
        KernelPluginCollection? plugins);

public Kernel(
        IServiceProvider? services,
        KernelPluginCollection? plugins = null,
        JsonSerializerOptions? serializerOptions = null);
```

MithrilMan:
Is there any ETA about this PR?

PR #5644, opened 2024-03-25
.Net - Remove empty properties for content serialization.

crickman:
### Motivation and Context

A lot of extra bytes for empty content.

```
 [
  {
    "Role": {
      "Label": "user"
    },
    "Content": "Discuss the potential long-term consequences for the Earth\u0027s ecosystem as well.",
    "Items": [
      {
        "$type": "TextContent",
        "Text": "Discuss the potential long-term consequences for the Earth\u0027s ecosystem as well.",
        "ModelId": null,
        "Metadata": null,
        "MimeType": null
      },
      {
        "$type": "ImageContent",
        "Uri": "https://fake-random-test-host:123",
        "Data": null,
        "ModelId": null,
        "Metadata": null,
        "MimeType": null
      },
      {
        "$type": "BinaryContent",
        "Content": "WzEsMiwzXQ==",
        "ModelId": null,
        "Metadata": null,
        "MimeType": null
      },
      {
        "$type": "AudioContent",
        "Data": "WzEsMiwzXQ==",
        "ModelId": null,
        "Metadata": null,
        "MimeType": null
      }
    ],
    "ModelId": null,
    "Metadata": null,
    "MimeType": null
  }
]
```
### Description

Fixed - no data loss:

```
    [
  {
    "Role": {
      "Label": "user"
    },
    "Items": [
      {
        "$type": "TextContent",
        "Text": "Discuss the potential long-term consequences for the Earth\u0027s ecosystem as well."
      },
      {
        "$type": "ImageContent",
        "Uri": "https://fake-random-test-host:123",
        "Data": null
      },
      {
        "$type": "BinaryContent",
        "Content": "WzEsMiwzXQ=="
      },
      {
        "$type": "AudioContent",
        "Data": "WzEsMiwzXQ=="
      }
    ]
  }
]
```

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
@@ -24,6 +24,7 @@ public class ChatMessageContent : KernelContent
     /// A convenience property to get or set the text of the first item in the <see cref="Items" /> collection of <see cref="TextContent"/> type.
     /// </summary>
     [EditorBrowsable(EditorBrowsableState.Never)]
+    [JsonIgnore]
     public string? Content
     {
         get
--- a/dotnet/samples/KernelSyntaxExamples/Example87_ChatHistorySerialization.cs
+++ b/dotnet/samples/KernelSyntaxExamples/Example87_ChatHistorySerialization.cs
@@ -14,6 +14,7 @@ namespace Examples;
 
 public class Example87_ChatHistorySerialization : BaseTest
 {
+    private static readonly JsonSerializerOptions s_options = new() { WriteIndented = true };
     /// <summary>
     /// Demonstrates how to serialize and deserialize <see cref="ChatHistory"/> class
     /// with <see cref="ChatMessageContent"/> having SK various content types as items.
@@ -36,7 +37,7 @@ public class Example87_ChatHistorySerialization : BaseTest
 
         var chatHistory = new ChatHistory(new[] { message });
 
-        var chatHistoryJson = JsonSerializer.Serialize(chatHistory);
+        var chatHistoryJson = JsonSerializer.Serialize(chatHistory, s_options);
 
         var deserializedHistory = JsonSerializer.Deserialize<ChatHistory>(chatHistoryJson);
 
@@ -52,6 +53,8 @@ public class Example87_ChatHistorySerialization : BaseTest
         WriteLine($"Binary content: {Encoding.UTF8.GetString((deserializedMessage.Items![2]! as BinaryContent)!.Content!.Value.Span)}");
 
         WriteLine($"Audio content: {Encoding.UTF8.GetString((deserializedMessage.Items![3]! as AudioContent)!.Data!.Value.Span)}");
+
+        WriteLine($"JSON:\n{chatHistoryJson}");
     }
 
     /// <summary>
@@ -72,7 +75,8 @@ public class Example87_ChatHistorySerialization : BaseTest
         // The custom resolver should be used to serialize and deserialize the chat history with custom .
         var options = new JsonSerializerOptions
         {
-            TypeInfoResolver = new CustomResolver()
+            TypeInfoResolver = new CustomResolver(),
+            WriteIndented = true,
         };
 
         var chatHistoryJson = JsonSerializer.Serialize(chatHistory, options);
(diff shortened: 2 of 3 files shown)


glahaye (approved):

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -24,6 +24,7 @@ public class ChatMessageContent : KernelContent
     /// A convenience property to get or set the text of the first item in the <see cref="Items" /> collection of <see cref="TextContent"/> type.
     /// </summary>
     [EditorBrowsable(EditorBrowsableState.Never)]
+    [JsonIgnore]
```
SergeyMenshykh, line 27: I wonder what would happen here if someone:  
1. Has chat history serialized and stored.  
2. Updates their code to the SK with the change.  
3. Deserializes the chat history.  
   
Would the property be assigned since it exists in the serialized JSON, despite the property being marked as ignored? Or would there be data loss?

crickman: There is no data loss.  If you are referring to the `Content` property, this is a derived/computed property and doesn't need to particiapte in the serialization contract.  This `Content` property will be ignored, but the history will not have any data-loss since the state of this derived/computed property is informed by the items collection.

It is common for properties in JSON not have a reciever property in the corresponding object definition.

SergeyMenshykh: It began being computed quite recently - just one month ago. Before that, it was merely a getter and setter. My concern is that there might be a serialized version of the chat history out there, however, the chances may be very low, with a value for the Content but without the value in the Items collection.

markwallace-microsoft: Can we ignore for serialisation but use the value for deserialisation?

crickman: I'm address the current implementation along with other optimizations.  I commonly create console projects to demonstrate to myself certain behaviors.  In this case, this is a non-breaking & non-data loss change.  (I also agree that this is a very low likelihood at this point as well)

crickman: Previously chaninging the nature of the `Content` property sematatics is an unfortunate evolution of the data-contract that might be best addressed with a release note given the low likelyhood that this case even exists in the wild.

crickman: > Can we ignore for serialisation but use the value for deserialisation?

I'll review what possible.

crickman: FYI - Follow-up / validation on my deserialization behavior description: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/missing-members

crickman: I don't believe so.  I'm not seeing an ingore on write-only option...perhaps with a customer serializer, that that pushes requirements to the caller.

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/ignore-properties
https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonignorecondition

crickman: Sergey has already identified this is a likely non-issue.  I think we just need blessing on release note.

matthewbolanos: Since both you and Sergey have identified this as a rare / non-issue, I'm ok with proceeding.

crickman: Let's def call out in release note.  Is there a way to track next release notes?

matthewbolanos: @markwallace-microsoft, are you able to provide guidance on how to add items to the release notes? Is it possible to also add some sample code demonstrating how to massage the previous serialization if someone is using the old version and has them saved in something like COSMOS DB?

crickman: I'm happy to gin up a translatin sample.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
markwallace-microsoft: @crickman The information in the PR description and details are what get's used in the release notes but we can also add additional content. So let's start with making sure the PR details highlights the potential issue.

PR #5664, opened 2024-03-26
.Net - Fixed creation of Kernel for Azure Open AI case.

crickman:
### Motivation and Context

Azure Open AI version of semantic kernel always creating OpenAIChatCompletionService internally.  This only affects the use of prompt-functions as plug-ins, not broader agent behavior.

### Description

- Create AzureOpenAIChatCompletion service based on context definition.
- Updated prompt-funciton agent test to run in AOAI mode
- Verified test execution

### Contribution Checklist

### DIFF
--- a/dotnet/src/Experimental/Agents/Internal/Agent.cs
+++ b/dotnet/src/Experimental/Agents/Internal/Agent.cs
@@ -119,10 +119,9 @@ internal sealed class Agent : IAgent
         IKernelBuilder builder = Kernel.CreateBuilder();
 
         this.Kernel =
-            Kernel
-                .CreateBuilder()
-                .AddOpenAIChatCompletion(this._model.Model, this._restContext.ApiKey)
-                .Build();
+            this._restContext.HasVersion ?
+                builder.AddAzureOpenAIChatCompletion(this._model.Model, this.GetAzureRootEndpoint(), this._restContext.ApiKey).Build() :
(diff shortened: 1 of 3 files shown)


alliscode (approved):

dotnet/src/Experimental/Agents/Internal/Agent.cs
```diff
@@ -119,10 +119,9 @@ internal sealed class Agent : IAgent
         IKernelBuilder builder = Kernel.CreateBuilder();
 
         this.Kernel =
-            Kernel
-                .CreateBuilder()
-                .AddOpenAIChatCompletion(this._model.Model, this._restContext.ApiKey)
-                .Build();
+            this._restContext.HasVersion ?
```
glahaye, line 122: So whether we use OpenAI or AOAI depends on the presence of a version in the rest context?

Shouldn't it be the other way around: the presence of a version in the rest context would depend on whether we can OpenAI or AOAI?

I'm not criticizing; I'm actually wondering!

crickman: This wasn't a principled design decision.  Its just an outgrowth of each service implementation:

- Open AI doesn't require version parameter in endpoint url.
- Azure does.  

We track endpoint state in the rest context...its not explosed publically and we're basically freezing this code for future removal.

glahaye (approved):

PR #5666, opened 2024-03-26
.Net - Add support for Name property to ChatMessageContent

crickman:
### Motivation and Context

Tracking identity is critical for multi-agent conversations.  It is also supported as part of the core chat-completion api: https://platform.openai.com/docs/api-reference/chat/create

```
 {
    "messages": [
        {
            "content": "Write one paragraph in response to the user that rhymes",
            "name": "Echo",
            "role": "system"
        },
        {
            "content": "Why is AI awesome",
            "name": "Ralph",
            "role": "user"
        }
    ],
    "temperature": 1,
    "top_p": 0.5,
    "n": 3,
    "presence_penalty": 0,
    "frequency_penalty": 0,
    "model": "gpt-4"
}
```

### Description

Add support for `ChatMessageContent.Name` property with optional, non-breaking patterns.

### Contribution Checklist


crickman:
@markwallace-microsoft  These signature changes won't break any consumer's build. How do we update our baseline?

crickman:
> > @markwallace-microsoft These signature changes won't break any consumer's build. How do we update our baseline?
> > I usually use the command - dotnet build --configuration Release /p:GenerateCompatibilitySuppressionFile=true
> > More details - https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/diagnostic-ids#suppression-file

Thanks, I must've been missing the `Release` parameter.

crickman:
FYI - Engaged with Matthew and re-evaluating `ChatHistory.SystemName` and how/if assistant response name is captured.

crickman:
FYI - Name _not_ returned in response:

**REQUEST**
```
 {
    "messages": [
        {
            "content": "Write one paragraph in response to the user that rhymes",
            "name": "Echo",
            "role": "system"
        },
        {
            "content": "Why is AI awesome",
            "name": "Ralph",
            "role": "user"
        }
    ],
    "temperature": 1,
    "top_p": 0.5,
    "n": 3,
    "presence_penalty": 0,
    "frequency_penalty": 0,
    "model": "gpt-4"
}
```

**RESPONSE**
```
{
  "id": "chatcmpl-977K8yKPbvVxqaC6rrT6EBc5S4E6s",
  "object": "chat.completion",
  "created": 1711483476,
  "model": "gpt-4-0613",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "AI is a dream, a gleaming beam, with a stream of knowledge that seems supreme. It's like a team in the extreme, creating schemes and decoding themes. It's awesome because it's keen, it can glean unseen scenes, and it's never mean or obscene. It's the best machine you've ever seen."
      },
      "logprobs": null,
      "finish_reason": "stop"
    },
    {
      "index": 1,
      "message": {
        "role": "assistant",
        "content": "AI is a gem, it's true, it shines with a gleam so bright and new. It learns and grows, it's never the same, constantly evolving, it's not just a game. It's awesome because it never tires, always ready, it never retires. It's the future, it's the now, to its brilliance, we must bow."
      },
      "logprobs": null,
      "finish_reason": "stop"
    },
    {
      "index": 2,
      "message": {
        "role": "assistant",
        "content": "AI is a cosmic blossom, it's awesome because it's a possum playing possum. It's a system that's fathomless, solving problems that seem bottomless. It's a key to a lock, a tick to a tock, a shepherd to a flock. AI is awesome, it's not just talk, it's a journey around the clock."
      },
      "logprobs": null,
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 31,
    "completion_tokens": 216,
    "total_tokens": 247
  },
  "system_fingerprint": null
}
```

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -201,6 +207,17 @@ public void RemoveRange(int index, int count)
         this._messages.RemoveRange(index, count);
     }
 
+    /// <summary>
+    /// An in-place enumeration of the chat history starting with the most recent message first.
+    /// </summary>
+    public IEnumerable<ChatMessageContent> ToDescending()
```
markwallace-microsoft, line 213: We don't need to add utility methods to our public API surface, developers can implement things like this themselves and we can include implementations in our samples if needed

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -56,46 +56,52 @@ public ChatHistory(IEnumerable<ChatMessageContent> messages)
     /// <param name="content">Message content</param>
     /// <param name="encoding">Encoding of the message content</param>
     /// <param name="metadata">Dictionary for any additional metadata</param>
+    /// <param name="name">Name of the author of the message</param>
     /// </summary>
-    public void AddMessage(AuthorRole authorRole, string content, Encoding? encoding = null, IReadOnlyDictionary<string, object?>? metadata = null) =>
-        this.Add(new ChatMessageContent(authorRole, content, null, null, encoding, metadata));
+    public void AddMessage(AuthorRole authorRole, string content, Encoding? encoding = null, IReadOnlyDictionary<string, object?>? metadata = null, string? name = null) =>
+        this.Add(new ChatMessageContent(authorRole, content, null, null, encoding, metadata, name));
```
markwallace-microsoft, line 62: This can be problematic if a developer has a dependency on SK and also on a library that uses SK. If the direct and indirect dependencies are for different SK versions and the older version get's loaded first then there is an issue. We've seen this problem in our polyglot notebooks.

What this and the other changes indicates it that all of these helper methods need updating as new properties get added to the `ChatMessageContent`.

We should consider marking the helpers as obsolete and just relying on `ChatHistory.AddMessage(ChatMessageContent chatMessage)`. Developers can create their own extension methods for the additional helpers easily enough and this simplifies our API surface.

@matthewbolanos @stephentoub Thoughts?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ChatHistoryExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static class ChatHistoryExtensions
     /// <param name="streamingMessageContents"><see cref="IAsyncEnumerator{T}"/> list of streaming message contents</param>
     /// <returns>Returns the original streaming results with some message processing</returns>
     [Experimental("SKEXP0010")]
-    public static async IAsyncEnumerable<StreamingChatMessageContent> AddStreamingMessageAsync(this ChatHistory chatHistory, IAsyncEnumerable<OpenAIStreamingChatMessageContent> streamingMessageContents)
+    public static async IAsyncEnumerable<OpenAIStreamingChatMessageContent> AddStreamingMessageAsync(this ChatHistory chatHistory, IAsyncEnumerable<OpenAIStreamingChatMessageContent> streamingMessageContents)
```
markwallace-microsoft, line 24: Why do we need this in our public API surface?

The only use I see in the current codebase is in the associated unit test.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
crickman: Obsolete, ok...I like that.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ChatHistoryExtensions.cs
crickman: I don't have the context.  Nor have I changed visibility.  The only change here is noticing that the down-cast is degrading the caller context (anti-pattern) as the caller already must be passing in a set of `OpenAIStreamingChatMessageContent` objects.

By inspection, it appears that raison d'etra for this is to append the result of streaming messages to an existing chat-history object.  A common operation that doesn't need to be federated to every single caller.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
crickman: This is a good change to zoom into, as I haven't provided any context.  This is an outgrowth of a new requirement I've discovered around chat history.  I can roll-it back for now and re-approach in an isolated PR.

Here is the core case to consider:

- Chat history is designed to always be in acending order (earliest message first).
- Common retrieval pattern for stored history it to retrieve in descending order (latest first).

As such, chat history should be able to support bi-directional enumeration without creating copies of arrays like the `Linq` extensions do (e.g. Reverse()).

I think this will be easier to grok in isolation...including now was opportunistic.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
crickman: Ok...here's the deal.  Adding an overload with a nullable string parameter at the end creates this error for every single caller:

CS0121	The call is ambiguous between the following methods or properties.

I really struggled with adding `name` as the final parameter as it didn't seem like a natural ordering.  We can create an ordering with `name` in a different position.  This will remove the ambiguity, but it also implies we leave the old methods non-[Obsolete] to avoid breaking existing patterns.

If these's are the only methods we're addressing...the scope is contrained enough, extra overloads seem fine.  I'll move to this posture for next review.

**IF** this pattern needs to propagate through a bunch of internal method signatures (which I'm not hearing), I'd be curious to examine those as well.

*If* these methods were returning the message being created (instead of void), the caller would have the option to explicitly add the name or any other state adjustment.  Does changing the return type break the transient dependency case?

crickman: Pushed update.  

Answered my own question: (as expected) changing return type is a breaking signature change.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -69,34 +79,76 @@ public ChatHistory(IEnumerable<ChatMessageContent> messages)
...
     /// <param name="content">Message content</param>
     public void AddUserMessage(string content) =>
         this.AddMessage(AuthorRole.User, content);
 
+    /// <summary>
+    /// Add a user message to the chat history
+    /// </summary>
+    /// <param name="name">Name of the author of the message</param>
+    /// <param name="content">Message content</param>
+    public void AddUserMessageWithName(string? name, string content) =>
```
markwallace-microsoft, line 104: This pattern doesn't scale i.e. what happens the next time a property is added.

Let's leave this out as it will be very difficult (if not impossible) to remove this in future.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -60,6 +60,16 @@ public ChatHistory(IEnumerable<ChatMessageContent> messages)
...
         this.Add(new ChatMessageContent(authorRole, content, null, null, encoding, metadata));
 
+    /// <summary>
+    /// <param name="name">Name of the author of the message</param>
+    /// <param name="authorRole">Role of the message author</param>
+    /// <param name="content">Message content</param>
+    /// <param name="encoding">Encoding of the message content</param>
+    /// <param name="metadata">Dictionary for any additional metadata</param>
+    /// </summary>
+    public void AddMessage(string? name, AuthorRole authorRole, string content, Encoding? encoding = null, IReadOnlyDictionary<string, object?>? metadata = null) =>
```
markwallace-microsoft, line 70: `ChatHistory` is out generic abstraction that is AI model agnostic.

Is the `Name` property support widely among chat completion models?

If not, we should consider adding this to the metadata as an optional property. Developers can create their own extension methods to make this more convenient in the future.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -142,13 +148,64 @@ public ChatMessageContent()
...
     /// </summary>
+    /// <param name="name">Name of the author of the message</param>
+    /// <param name="role">Role of the author of the message</param>
+    /// <param name="content">Content of the message</param>
+    /// <param name="modelId">The model ID used to generate the content</param>
+    /// <param name="innerContent">Inner content object reference</param>
+    /// <param name="encoding">Encoding of the text</param>
+    /// <param name="metadata">Dictionary for any additional metadata</param>
+    public ChatMessageContent(
+        string? name,
```
markwallace-microsoft, line 159: See earlier comment about how applicable this property is across multiple different LLM's.

The `metadata` property as added to allow us to add additional properties without requiring changes to the API surface.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
MediVik333: Try to found where is it

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
MediVik333: Ok

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -15,6 +15,12 @@ namespace Microsoft.SemanticKernel;
 /// </summary>
 public class ChatMessageContent : KernelContent
 {
+    /// <summary>
+    /// Name of the author of the message
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Name { get; set; }
```
markwallace-microsoft, line 22: Changes to the public API should not be immediately GA. We should mark the changes as experimental to allow us to get some feedback.

crickman: SKEXP0011

dotnet/docs/EXPERIMENTS.md
```diff
@@ -39,6 +39,7 @@ You can use the following diagnostic IDs to ignore warnings or errors for a part
 | SKEXP0010 | OpenAI parameters | | | | | |
 | SKEXP0010 | OpenAI chat history extension | | | | | |
 | SKEXP0010 | OpenAI file service | | | | | |
+| SKEXP0011 | OpenAI message identity | | | | | |
```
markwallace-microsoft, line 42: We are not doing fine grained experimental flags anymore as it's to inconvenient for developers. This can be included under `SKEXP0001`.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ChatHistoryExtensions.cs
markwallace-microsoft: Let's leave the API as is rather than introduce the noise of a breaking change. I've created this task to see can we remove this: https://github.com/microsoft/semantic-kernel/issues/5740

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -623,9 +627,7 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
                     // as the called function could in turn telling the model about itself as a possible candidate for invocation.
                     functionResult = (await function.InvokeAsync(kernel, functionArgs, cancellationToken: cancellationToken).ConfigureAwait(false)).GetValue<object>() ?? string.Empty;
                 }
-#pragma warning disable CA1031 // Do not catch general exception types
-                catch (Exception e)
-#pragma warning restore CA1031
+                catch (Exception e) when (!e.IsCriticalException())
```
dmytrostruk, line 630: Not sure about this change, do we still need to use `IsCriticalException` check across the code?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -571,8 +573,10 @@ await foreach (StreamingChatCompletionsUpdate update in response.ConfigureAwait(
 
             // Add the original assistant message to the chatOptions; this is required for the service
             // to understand the tool call responses.
-            chatOptions.Messages.Add(GetRequestMessage(streamedRole ?? default, content, toolCalls));
-            chat.Add(new OpenAIChatMessageContent(streamedRole ?? default, content, this.DeploymentOrModelName, toolCalls, metadata));
+            chatOptions.Messages.Add(GetRequestMessage(streamedRole ?? default, content, streamedName, toolCalls));
+#pragma warning disable SKEXP0001 // Assigning nullable property self-consistent for internal/private code.
```
dmytrostruk, line 577: I tried to remove this suppression and code builds without warnings. Do we still need it? Same in other places.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ChatHistoryExtensions.cs
dmytrostruk: I agree, if the change is not related to the purpose of this PR and it's forcing us to create suppression file, it's probably better to avoid it here and track it as part of separate task.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
crickman: This is just a way to do general exception handling w/o the pragma.  If its a critical exception, the process is coming down.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
crickman: We will...I think I dropped a property assignment.  Good eye.

crickman: Ok, yeah...its in the project file is all.

crickman: Cleaned

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
dmytrostruk: I remember there were some conversations about which exceptions should be considered as critical etc. Also, I'm not sure if this method is still up-to-date. Can we leave it as it was before (with `pragma`) and track the purpose of `IsCriticalException` as separate issue? Thanks!

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -20,6 +20,11 @@ public class StreamingChatMessageContent : StreamingKernelContent
     /// </summary>
     public string? Content { get; set; }
 
+    /// <summary>
+    /// Name of the author of the message
+    /// </summary>
+    public string? Name { get; set; }
```
dmytrostruk, line 26: Should this be marked as experimental as well?

crickman: Absolutely!

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -15,6 +15,13 @@ namespace Microsoft.SemanticKernel;
 /// </summary>
 public class ChatMessageContent : KernelContent
 {
+    /// <summary>
+    /// Name of the author of the message
+    /// </summary>
+    [Experimental("SKEXP0001")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Name { get; set; }
```
stephentoub, line 23: Is "Name" the right name here? At first glance it would seem to be the name of the message rather than the name of the author of the message. Should it be "AuthorName" to correspond with "AuthorRole"?

dmytrostruk (approved):

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -20,6 +21,13 @@ public class StreamingChatMessageContent : StreamingKernelContent
     /// </summary>
     public string? Content { get; set; }
 
+    /// <summary>
+    /// Name of the author of the message
+    /// </summary>
+    [Experimental("SKEXP0001")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Name { get; set; }
```
stephentoub, line 29: Ditto

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -42,7 +50,8 @@ public class StreamingChatMessageContent : StreamingKernelContent
     /// <param name="encoding">Encoding of the chat</param>
     /// <param name="metadata">Additional metadata</param>
     [JsonConstructor]
-    public StreamingChatMessageContent(AuthorRole? role, string? content, object? innerContent = null, int choiceIndex = 0, string? modelId = null, Encoding? encoding = null, IReadOnlyDictionary<string, object?>? metadata = null) : base(innerContent, choiceIndex, modelId, metadata)
+    public StreamingChatMessageContent(AuthorRole? role, string? content, object? innerContent = null, int choiceIndex = 0, string? modelId = null, Encoding? encoding = null, IReadOnlyDictionary<string, object?>? metadata = null)
+        : base(innerContent, choiceIndex, modelId, metadata)
```
stephentoub, line 54: what changed here? just formatting?

crickman: Yes, for readability

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
crickman: In the case of `Role` the type is `AuthorRole` and the property is `Role`.

I think the naming common usage and content and is consistent with the current type naming:  `Role`, `Content`, `Name` (sans prefixing).

I do understand and agree that overly simple naming can be problematic.  This often applies to types, such as `Node`, `Item`, `Kernel`, etc...  Because properties are type qualified, the concerns aren't identical (symbol collision).

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
crickman: Ditto

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
stephentoub: > In the case of Role the type is AuthorRole and the property is Role

It's unfortunate the author role property is just "Role", but even the it's arguable that's the role the message plays. It's a system message, a user message, etc, it's a property of the message.

This Name property really sounds to my ears like it's the name of the message. If there were an Id property, I'd again assume it was the id of the message.

> I think the naming common usage and content and is consistent with the current type naming: Role, Content, Name

Which highlights the problem. It's the content of the message but the name _of the author_ of the message.

crickman: If this must deviate from the underlying model / spec...perhaps just `Author`.  

I'd have to add json mapping and reset functional testing.

crickman: Consensus: `AuthorName` it is.  I'll ping when ready.

stephentoub (approved):

dmytrostruk (approved):

PR #3357, opened 2023-11-01
.Net: Added better formatting for responses from Bing Searches & Ability to use custom Bing Search endpoint.

SOE-YoungS:
### Motivation and Context
Returning just the page snippet from a Bing search results in a lack of context or ability to follow up on a response.

This PR addresses that by adding better formatting for Bing Search responses via the "Plugins.Web" package. 
It also adds a method to use a custom endpoint for Bing Search.

#### Formatting Changes
Previous response format example:
image

New response format example:
image

As can be seen by the examples provided, the new response format contains much more context & provides the user with the URL of the search result so that they can click through and read further.

This can be obtained programmatically later on using regex matching, then could be fed into the "[Search Url Plugin](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Plugins/Plugins.Web/SearchUrlPlugin.cs)", to scrape the page directly, before finally summarizing and returning a complete summary of the page to the user.

#### Endpoint Changes
The ability to use a custom endpoint for Bing Search also enables a path to use an API Management instance as a front end for the Bing Search API. This is required by some end users for scenarios such as enterprise logging and usage counting for cross charge.

The default behaviour for the BingConnector is to target the Bing Search endpoint directly. It is entirely optional to use a custom endpoint. Doing so, does not introduce any other code change requirements to achieve use of Bing Search.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#dev-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #5673, opened 2024-03-27
.Net: Added better formatting for responses from Bing Searches & Ability to use custom Bing Search endpoint. 

SOE-YoungS:
Motivation and Context
Returning just the page snippet from a Bing search results in a lack of context or ability to follow up on a response.

This PR addresses that by adding better formatting for Bing Search responses via the "Plugins.Web" package.
It also adds a method to use a custom endpoint for Bing Search.

Description
Formatting Changes
Previous response format example:
image

New response format example:
image

As can be seen by the examples provided, the new response format contains much more context & provides the user with the URL of the search result so that they can click through and read further.

This can be obtained programmatically later on using regex matching, then could be fed into the "[Search Url Plugin](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Plugins/Plugins.Web/SearchUrlPlugin.cs)", to scrape the page directly, before finally summarizing and returning a complete summary of the page to the user.

Endpoint Changes
The ability to use a custom endpoint for Bing Search also enables a path to use an API Management instance as a front end for the Bing Search API. This is required by some end users for scenarios such as enterprise logging and usage counting for cross charge.

The default behaviour for the BingConnector is to target the Bing Search endpoint directly. It is entirely optional to use a custom endpoint. Doing so, does not introduce any other code change requirements to achieve use of Bing Search.

Contribution Checklist
 The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#dev-scripts) raises no violations
 I didn't break anyone 😄


SOE-YoungS:
This PR was previously approved, but closed due to inactivity (I now cannot reopen the PR). #3357

SOE-YoungS:
@lemillermicrosoft in order to move this along, would you mind approving this PR please?

It's pretty much the same as the last PR, but with all prior comments & issues addressed.

Keen to get this one merged this time before it goes stale again.

Thanks.

SOE-YoungS:
@dmytrostruk thank you 👍

Issue #5153, opened 2024-02-26
.Net: [Agents] Support FunctionCallContent


SergeyMenshykh:
There are scenarios that require function calls across multiple models and agents. Today, our function calls are tied closely with the OpenAI connector, so it's difficult to reuse the same code for the Gemini connector and to allow agents to pass function calls to one another (AutoGen scenario)

This issue is to capture the work necessary to make the current implementation generic. This should include introducing a new 'FunctionCallContent' type that includes (at a minimum) the a reference to the 'KernelFunction' to be called and the arguments to call. This content type should then appear in the item collection for 'ChatMessageContent'.

As part of this change, all existing function calling behavior should still "work".

## Scenarios
As part of this issue, the following scenarios should be supported:
- **Core abstraction support** – A developer should be able to use function calls and tool results in `ChatHistory` and `ChatMessageContent` objects without relying on the OpenAI package.
- **Serialization/Deserializaion** – A developer should be able to serialize a `ChatHistory` object with function calls and tool results so they can rehydrate it in the future (and potentially run the `ChatHistory` with a different AI model).
- **Works with any model** – Any AI service (that supports tools) can consume tool calls and tool results without having service specific classes. This includes Gemini, Mistral, Anthropic, and Llama.
- **Passing function alls** – In multi-agent scenarios, one agent can create a tool call for another agent to complete it.
- **Simulated function calls** – A function call can be simulated by a developer; in other words, a developer should be able to add a `ChatMessageContent` with a tool call they created to a `ChatHistory` object and then run it with any AI model (this may require simulating tool call IDs in the case of OpenAI). This is helpful when a developer wants to templatize context and system prompts aren't working.
- ~~**YAML support for function calls** – A function call and tool result can be written in a YAML prompt template using the intermediate prompt template language. This will allow a developer to implement few shot examples for function calling within prompts. A new tag (e.g., `<FunctionCall>` is likely necessary to achieve this).~~
- ~~**YAML support for function results** – A function result can be written in a YAML prompt template as a tool message.~~
- ~~**YAML support for defining available functions** – as a prompt engineer, you should be able to state which functions are available for the LLM to call.~~

~~If possible, we should try to align the syntax with [.promptfile](https://promptfile.org/blocks). They use `<functions>` and `<function>` tags.~~

The YAML scenarios/requirements are moved to the issue -https://github.com/microsoft/semantic-kernel/issues/5632 to be prioritized separately.

crickman:
[...]

PR #5696, opened 2024-03-29
.Net: ADR for function call content model classes

SergeyMenshykh:
Today, in SK, LLM function calling is supported exclusively by the OpenAI connector, and the function calling model is specific to that connector. The new AI connectors being added to SK, which support function calling, introduce their specific models for function calling. The design, where each new connector introduces its own specific model class for function calling, does not scale well from the connector development perspective and does not allow for polymorphic use of connectors by the SK consumer code.

This ADR describes the high-level details of the service-agnostic function-calling model classes, while leaving the low-level details to the implementation phase. Additionally, this ADR outlines the identified options for various aspects of the design.

Requirements - https://github.com/microsoft/semantic-kernel/issues/5153


docs/decisions/0039-function-call-content.md
```diff
@@ -0,0 +1,435 @@
...
+    chatHistory.Add(new ChatMessageContent(AuthorRole.Tool, new ChatMessageContentItemCollection { new FunctionResultContent(functionCall, result) }));
+    
+    // of cause, the syntax can be simplified by having additional instance/extension methods
+    chatHistory.AddFunctionResultMessage(result); // Using the new AddFunctionResultMessage extension method of ChatHistory class
+}
+```
+
+Questions:
+- How to pass the original `FunctionCallContent` to connectors along with the function result. It's actually not clear atm whether it's needed or not. The current rationale is that some models might expect properties of the original function call, such as arguments, to be passed back to the LLM along with the function result. An argument can be made that the original function call can be found in the chat history by the connector if needed. However, a counterargument is that it may not always be possible because the chat history might be truncated to save tokens, reduce hallucination, etc.
+- How to pass function id to connector?
```
matthewbolanos, line 348: Couple of options from my side...
- Introduce a `public class FunctionCallResultContent : FunctionResult` that includes the ID (and perhaps the original `FunctionCallContent` too). This would be returned by the `functionCall.InvokeAsync` method.
- Introduce an optional `ID` property on `FunctionResult` (not the biggest fan of this...)

docs/decisions/0039-function-call-content.md
```diff
@@ -0,0 +1,435 @@
...
+- How to communicate exception to the connectors? It was proposed to add the `Exception` property the the `FunctionResult` class that will always be assigned by the `KernelFunction.InvokeAsync` method. However, this change will break C# function calling semantic, where the function should be executed if the contract is satisfied, or an exception should be thrown if the contract is not fulfilled.
```
matthewbolanos, line 349: Based on the filter work that @dmytrostruk is driving, I _think_ we want to have the following principle...
- Always throw an error (to match standard C# behavior)
- If invoked within the context of a kernel (e.g., prompt template or function calling), have a filter that exposes that the error occurred.
- At this point, the developer should have two options:
  1. Do nothing, in this case, the error should bubble up so it can be caught elsewhere
  2. Handle the error so that the error reason can be passed back to the AI

If we did this, would we even _need_ the `Exception` property since the exception reason would just be put in the main result field?

I'm assuming the main challenge would be maintaining backwards compatibility where we automatically handle exceptions on behalf of the user. Let me think on this some more...

docs/decisions/0039-function-call-content.md
```diff
@@ -0,0 +1,435 @@
...
+- If `FunctionResult` becomes a non-steaming content by inheriting `KernelContent` class, how the `FunctionResult` can represent streaming content capabilities represented by the `StreamingKernelContent` class when/if it needed later? C# does not support multiple inheritance.
```
matthewbolanos, line 350: This is a good one... Let me identify scenarios where this would be relevant.

docs/decisions/0039-function-call-content.md
SergeyMenshykh: At the moment `FunctionResult` is not {de}serializable. Let me prototype to see what it would take to make it {de}serialization friendly.

docs/decisions/0039-function-call-content.md
SergeyMenshykh: > If we did this, would we even _need_ the `Exception` property since the exception reason would just be put in the main result field?

if we do the `exception -> error message` translation inside a function filter and return the error message from the filter in a main result field, we won't need the `Exception` property.

> 
> I'm assuming the main challenge would be maintaining backwards compatibility where we automatically handle exceptions on behalf of the user. Let me think on this some more...

Do/will we handle exceptions automatically on behalf of users? Users will have to opt in by registering function filters to swallow/translate exceptions into error messages, won't they?

docs/decisions/0039-function-call-content.md
SergeyMenshykh: Here are the results https://github.com/microsoft/semantic-kernel/pull/5703 of the exercise to make `FunctionResult` {de}serialization friendly:  
   
- The `FunctionResult.Function` property is lost (the property value is null) during the serialization/deserialization process, as it is not possible to serialize executable code. We can try to preserve the function parameters/return type schema, but it may require custom serialization, and it's not clear if it's worth it.  
   
- The `FunctionResult.Function` property had to be changed to nullable; otherwise, `FunctionResult` deserialization fails because the property was originally serialized as null and serializer can't create function instance from null. As a result, the change propagated to the `FunctionFilterContext.Function` property, which also became nullable. This may be considered a breaking change since filter code did not expect the `Function` property to be null, and suddenly after the SK update, it's null. However, I don't see it having a null value, since the filters are executed on a valid function before the function invocation returns a result that might get serialized if added to chat history, and only after deserialization will the `FunctionResult.Function` property become null.  
   
- The `FunctionResult.Value` loses its original type during the deserialization process. So, in [the unit test](https://github.com/microsoft/semantic-kernel/pull/5703/files#r1544994422), the Guid type turned into a JsonElement because the type of the value property is an object, and the best the JsonDeserializer can do in the absence of a more specific target type is to deserialize the JSON to `JsonElement`.  
   
In summary, yes, `FunctionResult` can be serializable/deserializable and, correspondingly, can become a derivative of the `KernelContent` class and can be used as a base class - `public class FunctionCallResultContent : FunctionResult`. 

However, the cost of the change would be the loss of the `Function` property (which might not be needed by connectors, and if needed it can be found via `kernel.Plugins` connectors have access to) and unnecessary propagation of the nullable nature of the `FunctionResult.Function` property to the `FunctionFilterContext.Function` property.

Next steps to prototype:
- Inherit the `FunctionResult` from `KernelContent`.
- Inherit the `Function{Call}ResultContent` from `FunctionResult`.

docs/decisions/0039-function-call-content.md
```diff
@@ -0,0 +1,447 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: proposed
```
markwallace-microsoft, line 3: Update status and date

docs/decisions/0039-function-call-content.md
```diff
@@ -0,0 +1,447 @@
...
+chatHistory.Add(new ChatMessageContent(AuthorRole.Tool, items);
+
+// Sending chat history containing function calls and function results to the LLM to get the final response
+messageContent = await completionService.GetChatMessageContentAsync(chatHistory, settings, kernel);
+```
+
+### Decision Outcome
+Option 1.2 was chosen due to its explicit nature.
+
+## 2. Function calling protocol for chat completion connectors
```
markwallace-microsoft, line 208: Should this be `2.`

docs/decisions/0039-function-call-content.md
SergeyMenshykh: Done

dmytrostruk (approved):

markwallace-microsoft (approved):

PR #5700, opened 2024-03-29
.Net: Marking the ToolCallResultSerializerOptions as obsolete.

SergeyMenshykh:
The `ToolCallBehavior.ToolCallResultSerializerOptions` property will be replaced by an alternative solution soon.


stephentoub (approved):

stephentoub:
Thanks.

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
```diff
@@ -91,6 +93,8 @@ private ToolCallBehavior(bool autoInvoke)
     /// <summary>
     /// Options to control tool call result serialization behavior.
     /// </summary>
+    [Obsolete("This property is deprecated in favor of Kernel.SerializerOptions that will be introduced in one of the following releases.")]
```
stephentoub, line 96: You'll probably need to suppress the warnings you get on the various use sites.

SergeyMenshykh: Yep, suppressed.

markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs
MediVik333: Ok

MediVik333:
Ok

PR #5703, opened 2024-03-30
.Net: Making the FunctionResult class serialization friendly

SergeyMenshykh:
POC, for demonstration purposes only.

### DIFF
--- a/dotnet/src/SemanticKernel.UnitTests/Functions/FunctionResultTests.cs
+++ b/dotnet/src/SemanticKernel.UnitTests/Functions/FunctionResultTests.cs
@@ -3,6 +3,9 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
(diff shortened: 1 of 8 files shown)


dotnet/src/SemanticKernel.UnitTests/Functions/FunctionResultTests.cs
```diff
@@ -134,4 +136,44 @@ public void GetValueWhenValueIsKernelContentGenericTypeMatchShouldReturn()
         Assert.Equal(valueType, target.GetValue<TextContent>());
         Assert.Equal(valueType, target.GetValue<KernelContent>());
     }
+
+    [Fact]
+    public async Task ItCanBeSerializedAndDeserializedAsync()
```
SergeyMenshykh, line 141: bookmark

PR #5705, opened 2024-03-31
.Net - Voilà: Agent Framework

crickman:
### Motivation and Context

Initial Agent Framework.  Incudes:

- All critical abstractions
- ChatCompletionAgent
- All new projects / all marked with experimental attributes
- New `AgentSyntaxExamples`

### NOTE - NOT PUBLISHING NUGET 

**Test Coverage (blocks / lines):**
image

### Description

Shouldn't be any surprises.  Let's take personal notes on naming preference as there hasn't been finality in decision process.  A final refactor-rename and comment update at the end should be rapid (I'd rather do renaming/organziation once than multiple times).  

I'm going to stack PR's behind this, so in the interest of progress...

### Outstanding Tasks - In Order (each a future PR)

- [ ] AgentChat (our "GroupChat")
- [ ] Assistant-as-a-Plugin
- [ ] OpenAIAssistantAgent
- [ ] OpenAIAssistantAgent Citiation Content
- [ ] Port AutoGen examples
- [ ] Streaming
- [ ] YAML Templates

### Contribution Checklist


dotnet/docs/EXPERIMENTS.md
```diff
@@ -76,3 +77,6 @@ You can use the following diagnostic IDs to ignore warnings or errors for a part
 | | | | | | | |
 | SKEXP0101 | Experiment with Assistants | | | | | |
 | SKEXP0101 | Experiment with Flow Orchestration | | | | | |
+| | | | | | | |
+| SKEXP0110 | Agent Framework | | | | | |
+| SKEXP0111 | ChatCompletionAgent | | | | | |
```
SergeyMenshykh, line 82: We recently moved away from the fine-grained experimental IDs. SKEXP0110 for all agents should be aligned with the new convention.

crickman: I just learned that from Mark as well.  Keeps thing more simple, great!

dotnet/src/Agents/Framework/AgentException.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+using System;
+
+namespace Microsoft.SemanticKernel.Agents;
+
+/// <summary>
+/// Agent specific <see cref="KernelException"/>.
+/// </summary>
+public class AgentException : KernelException
```
SergeyMenshykh, line 9: We moved away from having an exception for each SK component and decided to go with a minimal number of custom exceptions, following the principle of "less is more." and add them later if they are really necessary. I wonder in which scenarios the agent exception can be useful to agent callers, considering that it does not add anything on top of the inherited kernel exception.

dotnet/samples/AgentSyntaxExamples/AgentInventory.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+namespace AgentSyntaxExamples;
+
+public static partial class AgentInventory
+{
+    public const string ParrotName = "Parrot";
```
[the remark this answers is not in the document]
crickman: This is a concept in motion...I'm happy to revisit and refine example concepts as we proceed.  As examples develop, I do expect to be re-using definitions.

dotnet/samples/AgentSyntaxExamples/BaseTest.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace Examples;
+
+public abstract class BaseTest
+{
+    /// <summary>
+    /// Flag to force usage of OpenAI configuration if both <see cref="TestConfiguration.OpenAI"/>
+    /// and <see cref="TestConfiguration.AzureOpenAI"/> are defined.
+    /// If 'false', Azure takes precedence.
+    /// </summary>
+    protected virtual bool ForceOpenAI { get; } = false;
```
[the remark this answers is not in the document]
crickman: It used by the user to control test execution according to their preference.   This pattern already exists in KernelSyntaxExamples.  The issue is if a user has defined both config sections, Azure will always take precedence.  This allows that behavior to be overrided, as commented.

dotnet/src/Agents/Framework/Extensions/ChatMessageContentExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+/// <summary>
+/// Extension methods for <see cref="ChatMessageContent"/>
+/// </summary>
+public static class ChatMessageContentExtensions
+{
+    /// <summary>
+    /// Determines if <see cref="ChatMessageContent"/> has content.
+    /// </summary>
+    public static bool HasContent(this ChatMessageContent? message)
+        => !string.IsNullOrWhiteSpace(message?.Content);
```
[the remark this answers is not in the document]
crickman: Desired, on-radar.  I was hoping to drill into that...maybe during this week's sync?

https://github.com/microsoft/semantic-kernel/issues/5753

dotnet/samples/AgentSyntaxExamples/RepoUtils/ObjectExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json;
+
+namespace RepoUtils;
+
+public static class ObjectExtensions
```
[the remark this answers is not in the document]
crickman: I pruned out a bunch of stuff agents will be unlikely to use, but I'd like to reserve the ability to use this extension in the future.  (Was wondering if this might spark discussion on a code sharing pattern for *SyntaxExamples.)

dotnet/src/Agents/Core/ChatCompletionAgent.cs
```diff
@@ -0,0 +1,95 @@
...
+            yield return message;
+        }
+
+        async Task FormatInstructionsAsync(string? instructions, CancellationToken cancellationToken)
+        {
+            if (!string.IsNullOrWhiteSpace(instructions))
+            {
+                instructions = (await this.FormatInstructionsAsync(instructions, cancellationToken).ConfigureAwait(false))!;
+
+                chat.Add(new ChatMessageContent(AuthorRole.System, instructions) { Name = $"{this.Name} Instructions" });
```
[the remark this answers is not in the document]
crickman: Was playing not critical.  Even without the prefix, models that don't support experience the same behavior inside and outside of agents (as we discussed yesterday).

dotnet/samples/AgentSyntaxExamples/AgentSyntaxExamples.csproj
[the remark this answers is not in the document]
crickman: I'm happy to merge projects if we decide that.  For now, I perfer the separation.

For one, `KernelSyntaxExamples` are documents in the readme as a "ollection of semi-random examples". 

The `AgentSyntaxExampels` are not intended to be random and will benefit from a targeted readme.

Yes, I was expecting to discuss duplication...but didn't want to invest cycles without consensus.

https://github.com/microsoft/semantic-kernel/issues/5755

dotnet/src/Agents/Core/ChatCompletionAgent.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// <inheritdoc/>
+    public override string Id { get; }
+
+    /// <inheritdoc/>
+    public override string? Name { get; }
+
+    /// <summary>
+    /// Additional instructions to always append to end of history (optional)
+    /// </summary>
+    public string? ExtraInstructions { get; set; }
```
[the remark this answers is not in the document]
crickman: Related to Ben's ask but will be superceded by a templating feature upon which Matthew is working to define.

dotnet/src/Agents/Framework/AgentException.cs
crickman: We already have `AgentException` in the experimental framework.  imho - Only throwing a single leaf exception type is a .Net & Java anti-pattern.  Since it inherits from KernelException, that's good...but it also allows for finer discrimintation in mixed app code.

https://github.com/microsoft/semantic-kernel/issues/5756

dotnet/samples/AgentSyntaxExamples/BaseTest.cs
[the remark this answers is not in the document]
crickman: `GetApiKey` is used by file service which is 100% going to be utilized in these examples.

dotnet/samples/AgentSyntaxExamples/AgentSyntaxExamples.csproj
```diff
@@ -0,0 +1,60 @@
...
+    <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
+    <PackageReference Include="Microsoft.Extensions.Http" />
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" />
+    <PackageReference Include="Microsoft.Extensions.Logging" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" />
+    <PackageReference Include="SharpToken" />
+    <PackageReference Include="Microsoft.ML.Tokenizers" />
+    <PackageReference Include="Microsoft.DeepDev.TokenizerLib" />
```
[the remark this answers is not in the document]
crickman: Removed except for logging.  That will 100% be utilized.

dotnet/samples/AgentSyntaxExamples/AgentSyntaxExamples.csproj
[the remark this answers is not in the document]
alliscode: +1 to having these in an Agents folder inside the existing KernelSyntaxExamples project, potentially with its own README.  I think SK users are already trained to go to the existing project to look for examples, even ones that show Agent based scenarios. I'd recommend we start that way and then promote them to a new project if we really need to.

dotnet/samples/AgentSyntaxExamples/BaseTest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    protected ITestOutputHelper Output { get; }
+
+    protected ILoggerFactory LoggerFactory { get; }
+
+    protected string GetApiKey()
+    {
+        if (string.IsNullOrEmpty(TestConfiguration.AzureOpenAI.Endpoint))
```
alliscode, line 28: Does this need to account for the `ForceOpenAI ` flag as well or is that handled somewhere else?
```suggestion
        if (string.IsNullOrEmpty(TestConfiguration.AzureOpenAI.Endpoint) || ForceOpenAI)
```

dotnet/samples/AgentSyntaxExamples/README.md
```diff
@@ -0,0 +1,36 @@
+#Semantic Kernel: Agent syntax examples
+
+This project contains a collection examples on how to use SK Agents.
```
alliscode, line 3:
```suggestion
This project contains a collection of examples on how to use SK Agents.
```

dotnet/src/Agents/Framework/AgentException.cs
alliscode: I don't love that we only have a single exception in SK, I find that it makes programmatically handling details of the issue challenging.  However, I also think it's weird to have one place that breaks the pattern and has its own exception, especially since it's not adding functionality to KernelException. I recommend that we stick with the existing pattern unless we have a really compelling reason to break it.

As a separate issue and PR, I would love if we could holistically improve the granularity of our exceptions in a non-breaking way.

dotnet/samples/AgentSyntaxExamples/BaseTest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    protected BaseTest(ITestOutputHelper output)
+    {
+        this.Output = output;
+        this.LoggerFactory = new XunitLogger(output);
+
+        LoadUserSecrets();
+    }
+
+    private static void LoadUserSecrets()
```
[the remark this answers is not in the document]
crickman: Agree (with what I think you've indicated).  This was copied, but now updated.

dotnet/samples/AgentSyntaxExamples/BaseTest.cs
crickman: Doh!  Indeed...both spots.

dotnet/src/Agents/Framework/Internal/BroadcastQueue.cs
```diff
@@ -0,0 +1,140 @@
...
+        Exception? failure = null;
+
+        while (!queue.IsEmpty)
+        {
+            if (queue.TryPeek(out var messages)) // Leave payload on queue for retry on failure
+            {
+                try
+                {
+                    await channel.Channel.ReceiveAsync(messages).ConfigureAwait(false);
+                    queue.TryDequeue(out _); // Queue has already been peeked.  Remove head on success.
```
alliscode, line 109: If I'm reading this right then this equates to "at least once" semantics for channels receiving a given message. I'm curious if the channel itself has any de-duping in it or if there will be issues with a given message being received more than once.

crickman: Good question.  Appreciate eyes on this code.  I'm going to first describe what's in my head and then try to answer your question.

At this point, whatever messages have been queued are taken as proper and must be recieved by the channel to remain in sync.  `Enqueue` is only exposed to the `AgentNexus` (base class for all nexus...so the usage is a closed system.  The nexus ensures that only the stream of individual messages it encounters are broadcast.  (FYI - Broadcast is a literal no-op for single channel case.)  

If the caller happens to add two of the same user message to the history...they won't be de-duped.  We don't have the visibilty to second guess caller intent.  But any single instance passed to the nexus (user or assistant/model generated) will not be duplicated due to agent processing or concurrency.

crickman: One loose end to call-out...I'm currently working on exposing a "Flush" semantic to the `AgentChat`/`AgentNexus` that will be in next PR that takes advantage of `EnsureSynchronizedAsync`.

crickman: Follow-up: Create sequence diagram https://github.com/microsoft/semantic-kernel/issues/5766

alliscode (approved):

dotnet/src/Agents/Framework/Extensions/ChatMessageContentExtensions.cs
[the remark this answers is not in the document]
crickman: I've actually developed some concepts around this in the active PR.

dotnet/src/Agents/Framework/AgentException.cs
crickman: Totally agree.  They can still all be `KernelException`...I actually reached out to Sergey last year when I saw this happening.  Text parsing exception messages just doesn't scale.

dotnet/samples/AgentSyntaxExamples/AgentSyntaxExamples.csproj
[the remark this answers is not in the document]
crickman: We've already been expanding focused examples and have received feedback from users that they want more targeted content.  I'm not sure overloading KSE is the direction things are actually trending...

Also...**AGENTS**

dotnet/src/Agents/Framework/AgentException.cs
SergeyMenshykh: > I don't love that we only have a single exception in SK, I find that it makes programmatically handling details of the issue challenging. 

Nothing prevents us from adding new exceptions to SK. Having only one exception was never the goal. The objective was to have one exception for the majority of cases and add more exceptions if additional and actionable details are truly necessary. So, if we find that SKException is missing some crucial information that needs to be expressed through exception properties, let's consider introducing another exception that would include those. As always, we should follow the "less is more" principle, because it's easy to add exceptions but difficult to modify or remove them without causing breaking changes in the open-source world.

> Totally agree. They can still all be KernelException...I actually reached out to Sergey last year when I saw this happening. Text parsing exception messages just doesn't scale.

The parsing of messages was never a goal. If we find ourselves doing it, we should stop. I thought I conveyed this during our conversation at that time. - _"The idea is simple - minimize number of custom SK exceptions by replacing them with SKException, use standard .NET exceptions where possible, and add derivatives of SKException with typed properties having additional details(**no string parsing**) only and only in case SK or SK consumer code needs to make a decision(resiliency, fallback, etc) based on exception details."_

dotnet/src/Agents/Framework/Internal/BroadcastQueue.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                return;
+            }
+        }
+
+        while (!queue.IsEmpty) // ChannelQueue is ConcurrentQueue, no need for _queueLock
+        {
+            lock (this._queueLock)
+            {
+                // Activate non-empty queue
```
SergeyMenshykh, line 78: Could you add the reason for the non-empty queue activation to the comment?

dotnet/src/Agents/Framework/Internal/BroadcastQueue.cs
```diff
@@ -0,0 +1,140 @@
...
+    public TimeSpan BlockDuration { get; set; } = TimeSpan.FromSeconds(1);
+
+    /// <summary>
+    /// Enqueue a set of messages for a given channel.
+    /// </summary>
+    /// <param name="channels">The target channels for which to broadcast.</param>
+    /// <param name="messages">The messages being broadcast.</param>
+    public void Enqueue(IEnumerable<ChannelReference> channels, IList<ChatMessageContent> messages)
+    {
+        lock (this._queueLock)
```
SergeyMenshykh, line 39: Has any alternative no-lock design (e.g., using concurrent dictionaries) been considered? Asking because of the potential problems with locks that are usually very difficult to debug, such as deadlocks and race conditions.

SergeyMenshykh (approved):

dotnet/src/Agents/Framework/Internal/BroadcastQueue.cs
crickman: Really glad you've gotten to this.  I'd really like to spend some cycles with you on evaluating.  At this point, I've significantly invested in functional testing and improvements in the next PR...as introducing AgentChat provided additional insight / experience.  

TLDR - A lot of what is here is already invalidated / evolved / fixed.  On merging this PR we can pivot to that one to continue engagement.

The updated version is really worth digging into.  The version in this PR is fine enough for a framework placeholder with tests.

dotnet/src/Agents/Framework/Internal/BroadcastQueue.cs
crickman: Yes...in fact in the latest PR I've backed away from using ConcurrentQueue.  The reason is that the locking semantics are at a higher granularity that a single container or single invocation.  (Race management.)  

I'd love to spend some time with your paired up going through the latest version.

dotnet/src/Agents/Core/ChatCompletionAgent.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// <summary>
+    /// Optional execution settings for the agent.
+    /// </summary>
+    public PromptExecutionSettings? ExecutionSettings { get; set; }
+
+    /// <inheritdoc/>
+    public override async IAsyncEnumerable<ChatMessageContent> InvokeAsync(
```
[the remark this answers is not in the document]
crickman: I really appreciate this observation.  This is certainly a valid concern.

I have been working to identify where a list makes more sense than enumerable (see strategy classes in draft PR).

In this specific case, I'm wondering if the same analysis holds true for non-local agents (OpenAI Assistant) in which there can be a significant lag between messages being generated.  That is, message generation is truly asynchronous in certain cases. 

I'm going to file an issue to track this.  Whan I make the OpenAI Assistant Agent available next week we'll be able to triangulate across different impementations:

https://github.com/microsoft/semantic-kernel/issues/5777

If you see any extraneous use of IEnumerable (esp method parameter), please let me know so we can address.

dotnet/src/Agents/Framework/AgentNexus.cs
```diff
@@ -0,0 +1,176 @@
...
+using Microsoft.SemanticKernel.Agents.Extensions;
+using Microsoft.SemanticKernel.Agents.Internal;
+using Microsoft.SemanticKernel.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Agents;
+
+/// <summary>
+/// Point of interaction for one or more agents.
+/// </summary>
+public abstract class AgentNexus
```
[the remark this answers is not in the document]
crickman: https://github.com/microsoft/semantic-kernel/issues/5780

dotnet/samples/AgentSyntaxExamples/Example01_Agent.cs
```diff
@@ -0,0 +1,70 @@
...
+        // Nothing to do...
+    }
+
+    /// <summary>
+    /// A basic nexus for the agent example.
+    /// </summary>
+    /// <remarks>
+    /// For further exploration of AgentNexus, see: Example03_Chat.
+    /// </remarks>
+    private sealed class TestChat : AgentNexus
```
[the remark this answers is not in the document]
crickman: https://github.com/microsoft/semantic-kernel/issues/5780

dotnet/src/Agents/Framework/Internal/BroadcastQueue.cs
crickman: Scheduled an hour on April 8th for us to dig into this.

dotnet/src/Agents/Framework/Internal/BroadcastQueue.cs
crickman: Scheduled an hour on April 8th for us to dig into this.

dotnet/src/Agents/Framework/ILocalAgent.cs
```diff
@@ -0,0 +1,21 @@
...
+public interface ILocalAgent
+{
+    /// <summary>
+    /// Entry point for calling into an agent with locally managed chat-history.
+    /// </summary>
+    /// <param name="history">The nexus history at the point the channel is created.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>Asynchronous enumeration of messages.</returns>
+    IAsyncEnumerable<ChatMessageContent> InvokeAsync(
+        IEnumerable<ChatMessageContent> history,
```
markwallace-microsoft, line 19: Should this be `ChatHistory`?

dotnet/src/Agents/Framework/ILocalAgent.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.SemanticKernel.Agents;
+
+/// <summary>
+/// Contract for an agent that acts upon a local message history.
+/// </summary>
+public interface ILocalAgent
```
markwallace-microsoft, line 10: The name suggests this defines a Local Agent but this is an interface that defines the contract to invoke an Agent with a chant history.

dotnet/samples/AgentSyntaxExamples/AgentSyntaxExamples.csproj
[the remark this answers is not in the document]
joslat: +1 too into grouping the agents samples, maybe as @alliscode mentions, in a folder under KSE  :)
Also, this will grow considerably as Agents features evolve so I stand with Mr. @crickman 's opinion in decoupling this from the main, mixed, examples. But it is also good to have them there as tests too... if the tests do work in a nested folder... the better :)

dotnet/src/Agents/Framework/ILocalAgent.cs
joslat: IAgentWithChatHistory? just thinking loud (now catching up with the ADR but will dig in deeper with the code...)
Question(s) to the masters:
- My understanding is that this PR implements https://github.com/microsoft/semantic-kernel/blob/adr_agent_architecture/docs/decisions/0039-agent-interactions.md and is aligned with https://github.com/microsoft/semantic-kernel/blob/adr_agent_architecture/docs/decisions/0032-agents.md - yes?
- What would be the best way to get the bits aside from getting this branch and building it locally? do you have a sample configuration file to fulfill? (that would help lots)

I will dig "quite a bit" in as I have a Session on the 19th April on Global Azure - on Agents ;)
So I plan to show some "what can we do now" and "in the short future" - again thanks for building this - it is amazing!!

joslat:
Question(s) to the masters @crickman & @markwallace-microsoft :
- My understanding is that this PR implements https://github.com/microsoft/semantic-kernel/blob/adr_agent_architecture/docs/decisions/0039-agent-interactions.md and is aligned with https://github.com/microsoft/semantic-kernel/blob/adr_agent_architecture/docs/decisions/0032-agents.md - yes?
- What would be the best way to get the bits aside from getting this branch and building it locally? do you have a sample configuration file to fulfill? (that would help lots)

I will dig "quite a bit" in as I have a Session on the 19th April on Global Azure - on Agents ;)
So I plan to show some "what can we do now" and "in the short future" - again thanks for building this - it is amazing!!

dotnet/src/Agents/Framework/ILocalAgent.cs
crickman: https://github.com/microsoft/semantic-kernel/issues/5780

crickman:
> Question(s) to the masters @crickman & @markwallace-microsoft :
> 
> * My understanding is that this PR implements https://github.com/microsoft/semantic-kernel/blob/adr_agent_architecture/docs/decisions/0039-agent-interactions.md and is aligned with https://github.com/microsoft/semantic-kernel/blob/adr_agent_architecture/docs/decisions/0032-agents.md - yes?
> * What would be the best way to get the bits aside from getting this branch and building it locally? do you have a sample configuration file to fulfill? (that would help lots)
> 
> I will dig "quite a bit" in as I have a Session on the 19th April on Global Azure - on Agents ;) So I plan to show some "what can we do now" and "in the short future" - again thanks for building this - it is amazing!!

Hi @joslat, the ADR intended to capture the high-level concepts and calling patterns from the POC branch.  As we shift to build this POC, some specifics have evolved (naming, signatures).  The goal is to return to the ADR and capture the final state to assist the Python and Java teams.  We're aiming for mid(-ish) month to achieve critical mass here (code-complete) and allow for ample time for feedback and stabilization.  I'd expect the ADR to be updated before the end of April.

I'm aiming to make information available early next week on building and consuming a local package...although, sometimes in "guerrilla-mode", I create an experimentation project with in local SK repo.

> Note: This change is primarily focused on the abstractions.  Even though `ChatCompletionAgent` is defined...the "chat" class is in the next PR.  Once added, we'll have MVP on this pattern for you to use.  Really looking forward to feedback.

dotnet/src/Agents/Framework/Agent.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    /// <summary>
+    /// The identifier of the agent (optional)
+    /// </summary>
+    public abstract string Id { get; }
+
+    /// <summary>
+    /// The name of the agent (optional)
+    /// </summary>
+    public abstract string? Name { get; }
```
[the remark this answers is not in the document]
crickman: I struggled with this as well.

OpenAIAssistantAgent sources this data from an underlying model as opposed to sourcing each property individually.

Ultimately, it just didn't feel right making implementation assumptions on the root abstraction.  I played around with shifting common patterns to `LocalKernelAgent`...maybe the `init` option gives us a path here...i'll play around.

dotnet/src/Agents/Abstractions/Agent.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <summary>
+    /// Produce the an <see cref="AgentChannel"/> appropriate for the agent type.
+    /// </summary>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>An <see cref="AgentChannel"/> appropriate for the agent type.</returns>
+    /// <remarks>
+    /// Every agent conversation, or <see cref="AgentChat"/>, will establish one or more <see cref="AgentChannel"/>
+    /// objects according to the specific <see cref="Agent"/> type.
+    /// </remarks>
+    protected internal abstract Task<AgentChannel> CreateChannelAsync(CancellationToken cancellationToken);
```
markwallace-microsoft, line 56: The abstraction defines multiple channel keys but only a single channel. Can you explain why this is?

dotnet/src/Agents/Abstractions/AgentChannel.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Agents;
+
+/// <summary>
+/// Defines the communication protocol for a particular <see cref="Agent"/> type.
+/// An agent provides it own <see cref="AgentChannel"/> via <see cref="Agent.CreateChannelAsync(CancellationToken)"/>.
+/// </summary>
+public abstract class AgentChannel
```
markwallace-microsoft, line 12: Should an `AgentChannel` have a channel key property?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /// <summary>
+    /// Append messages to the conversation.
+    /// </summary>
+    /// <param name="messages">Set of non-system messages with which to seed the conversation.</param>
+    /// <remarks>
+    /// Will throw KernelException if a system message is present, without taking any other action.
+    /// </remarks>
+    public void AppendHistory(IEnumerable<ChatMessageContent> messages)
```
markwallace-microsoft, line 65: I think `AddChatMessages` would be a better name. Using `Add` seems more consistent with .Net naming and `ChatMessages` is more accurate than just `History`.

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        return channel.GetHistoryAsync(cancellationToken);
+    }
+
+    /// <summary>
+    /// Append messages to the conversation.
+    /// </summary>
+    /// <param name="message">Set of non-system messages with which to seed the conversation.</param>
+    public void AppendHistory(ChatMessageContent message)
```
markwallace-microsoft, line 53: `AddChatMessage`

dotnet/src/Agents/Abstractions/ChatHistoryChannel.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.SemanticKernel.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Agents;
+
+/// <summary>
+/// A <see cref="AgentChannel"/> specialization for that acts upon a <see cref="IChatHistoryHandler"/>.
+/// </summary>
+public class ChatHistoryChannel : AgentChannel
+{
+    private readonly ChatHistory _chat;
```
markwallace-microsoft, line 16: nit: Elsewhere the variable is called `_history` would be good to be consistent

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        bool hasSystemMessage = false;
+        var cleanMessages =
+            messages.Where(
+                m =>
+                {
+                    bool isSystemMessage = m.Role == AuthorRole.System;
+                    hasSystemMessage |= isSystemMessage;
+                    return !isSystemMessage;
+                }).ToArray();
```
markwallace-microsoft, line 75: This creates another copy of the messages, is this needed? 
Can we instead check for the existence of non system messages and if none are found proceed with the original collection?

dotnet/src/Agents/Abstractions/Extensions/ChatMessageContentExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+/// <summary>
+/// Extension methods for <see cref="ChatMessageContent"/>
+/// </summary>
+internal static class ChatMessageContentExtensions
+{
+    /// <summary>
+    /// Determines if <see cref="ChatMessageContent"/> has content.
+    /// </summary>
+    public static bool HasContent(this ChatMessageContent? message)
+        => !string.IsNullOrWhiteSpace(message?.Content);
```
SergeyMenshykh, line 13: This logic will check only string/text content and does not take into account the other types of content like audio, image, etc.

SergeyMenshykh:
Could you please update the class diagram in this ADR - https://github.com/microsoft/semantic-kernel/pull/5528 to reflect recent changes/renaming? BTW: What are the plans for the ADR? Will it be superseded by another one, and is that why it was closed?

crickman:
> Could you please update the class diagram in this ADR - #5528 to reflect recent changes/renaming? BTW: What are the plans for the ADR? Will it be superseded by another one, and is that why it was closed?

I thought Mark had suggested (with agreement) that we sync-up the ADR post-code complete?

@markwallace-microsoft ?

markwallace-microsoft:
> > Could you please update the class diagram in this ADR - #5528 to reflect recent changes/renaming? BTW: What are the plans for the ADR? Will it be superseded by another one, and is that why it was closed?
> 
> I thought Mark had suggested (with agreement) that we sync-up the ADR post-code complete?
> 
> @markwallace-microsoft ?

Agreed, we can update the ADR once the changes are complete. We must do this as the Python and Java teams depend on this,

dotnet/samples/AgentSyntaxExamples/AgentInventory.ChatAgents.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Agents;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+
+namespace AgentSyntaxExamples;
+
+public static partial class AgentInventory
```
markwallace-microsoft, line 8: Why use a partial class?

The classes are both very small and to read the code I need to see content from both together to fully understand how the agents will behave.

dotnet/samples/AgentSyntaxExamples/AgentSyntaxExamples.csproj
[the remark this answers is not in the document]
markwallace-microsoft: Talk to @matthewbolanos about the new pattern we are using from samples. We are moving away from the pattern used in Kernel Syntax Examples

dotnet/samples/AgentSyntaxExamples/Example01_Agent.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        // Define the agent
+        ChatCompletionAgent agent =
+            new(
+                kernel: this.CreateKernelWithChatCompletion(),
+                instructions: AgentInventory.ParrotInstructions,
+                name: AgentInventory.ParrotName)
+            {
+                InstructionArguments = new() { { "count", 3 } },
+            };
```
markwallace-microsoft, line 32: One goal of samples is that the code can be copy/pasted

dotnet/samples/AgentSyntaxExamples/Example02_Plugins.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                this.WriteLine($"# {content.Role} - {content.AuthorName ?? "*"}: '{content.Content}'");
+            }
+        }
+    }
+
+    public Example02_Plugins(ITestOutputHelper output)
+        : base(output)
+    {
+        // Nothing to do...
```
markwallace-microsoft, line 65: I've seen this comment is multiple places, it doesn't add value so please remove it as others will copy this pattern

dotnet/samples/AgentSyntaxExamples/RepoUtils/ObjectExtensions.cs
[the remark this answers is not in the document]
markwallace-microsoft: Add the code when you need it.

As I mentioned in another comment the current pattern for syntax examples is being replaced so all of this code will need to be redone. I'd recommend keeping the amount of code in this project minimal to avoid building up technical debt.

dotnet/src/Agents/Abstractions/ChatHistoryKernelAgent.cs
```diff
@@ -0,0 +1,40 @@
...
+
+/// <summary>
+/// A <see cref="KernelAgent"/> specialization bound to a <see cref="ChatHistoryChannel"/>.
+/// </summary>
+public abstract class ChatHistoryKernelAgent : KernelAgent, IChatHistoryHandler
+{
+    /// <inheritdoc/>
+    protected internal sealed override IEnumerable<string> GetChannelKeys()
+    {
+        yield return typeof(ChatHistoryChannel).FullName;
```
markwallace-microsoft, line 16: Will the channel key always come from the channel instance?

If not, what is the use case for the channel not knowing it's own key?

Have you considered making the channel key a property of the `AgentChannel` class?

dotnet/src/Agents/Abstractions/ChatHistoryKernelAgent.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ChatHistoryKernelAgent"/> class.
+    /// </summary>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="instructions">The agent instructions</param>
+    protected ChatHistoryKernelAgent(Kernel kernel, string? instructions = null)
+        : base(kernel, instructions)
+    {
+        // Nothing to do...
```
markwallace-microsoft, line 38: Remove this comment

dotnet/src/Agents/Abstractions/Extensions/AgentChatExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+/// Extension methods for <see cref="KernelAgent"/>
+/// </summary>
+public static class AgentChatExtensions
+{
+    /// <summary>
+    /// Add user message to chat history
+    /// </summary>
+    /// <param name="chat">The target chat.</param>
+    /// <param name="input">Optional user input.</param>
+    public static ChatMessageContent? AppendUserMessageToHistory(this AgentChat chat, string? input)
```
markwallace-microsoft, line 16: Rename to `AddUserMessage` and add directly to the `AgentChat` class. Adding extension methods in the same package doesn't make sense.

dotnet/src/Agents/Abstractions/Extensions/ChatMessageContentExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Retrieves <see cref="ChatMessageContent.Content"/>, if defined.
+    /// </summary>
+    public static bool TryGetContent(this ChatMessageContent? message, out string content)
```
markwallace-microsoft, line 18: Would `TryGetTextContent` be a better name?

dotnet/src/Agents/Abstractions/Extensions/KernelAgentExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.SemanticKernel.Agents.Extensions;
+
+/// <summary>
+/// Extension methods for <see cref="KernelAgent"/>
+/// </summary>
+public static class KernelAgentExtensions
+{
+    /// <summary>
+    /// Render the provided instructions using the specified arguments.
```
markwallace-microsoft, line 14: How are the arguments specified?

dotnet/src/Agents/Abstractions/Internal/ChannelReference.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.SemanticKernel.Agents.Internal;
+
+/// <summary>
+/// Tracks channel along with its hashed key.
+/// </summary>
+internal readonly struct ChannelReference
```
markwallace-microsoft, line 7: Why can't the `AgentChannel` instance be used directly?

dotnet/src/Agents/Abstractions/Internal/PromptRenderer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Microsoft.SemanticKernel.Agents.Internal;
+
+/// <summary>
+/// Utility class for rendering an agent's system instructions.
+/// </summary>
+internal static class PromptRenderer
+{
+    private static readonly KernelPromptTemplateFactory s_factory = new();
+    private static readonly ConcurrentDictionary<string, TemplateReference> s_templates = new();
```
markwallace-microsoft, line 14: Why are we doing this caching?

dotnet/src/Agents/Abstractions/Internal/PromptRenderer.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Agents.Internal;
+
+/// <summary>
+/// Utility class for rendering an agent's system instructions.
+/// </summary>
+internal static class PromptRenderer
+{
+    private static readonly KernelPromptTemplateFactory s_factory = new();
```
markwallace-microsoft, line 13: Where does this get set?

dotnet/src/Agents/Abstractions/Internal/PromptRenderer.cs
```diff
@@ -0,0 +1,80 @@
...
+            if (!s_templates.TryGetValue(agent.Id, out TemplateReference templateReference) ||
+                !templateReference.IsConsistent(instructions!))
+            {
+                // Generate and cache prompt template if does not exist or if instructions have changed.
+                IPromptTemplate template =
+                    s_factory.Create(
+                        new PromptTemplateConfig
+                        {
+                            Template = instructions!
+                        });
```
markwallace-microsoft, line 41: This approach won't work if a developer wants to use Handlers or their own custom template format

dotnet/src/Agents/Abstractions/Internal/PromptRenderer.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(instructions))
+        {
+            return null;
+        }
+
+        if (agent.InstructionArguments != null)
+        {
+            if (!s_templates.TryGetValue(agent.Id, out TemplateReference templateReference) ||
+                !templateReference.IsConsistent(instructions!))
```
markwallace-microsoft, line 33: What happens if the arguments change?

dotnet/src/Agents/Abstractions/KernelAgent.cs
```diff
@@ -0,0 +1,34 @@
...
+
+/// <summary>
+/// Base class for agents utilizing <see cref="Microsoft.SemanticKernel.Kernel"/> plugins or services.
+/// </summary>
+public abstract class KernelAgent : Agent
+{
+    /// <summary>
+    /// The arguments used to optionally format <see cref="Instructions"/>.
+    /// </summary>
+    public KernelArguments? InstructionArguments { get; set; }
```
markwallace-microsoft, line 12: Do we need this in the initial implementation?

Do we expect the instruction arguments to change over the lifetime of the agent?

I would favour keeping this simple to start with and just passing in the instructions string and if a developer want to generate this from a prompt template they do this once up front.

dotnet/src/Agents/Core/ChatCompletionAgent.cs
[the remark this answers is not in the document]
markwallace-microsoft: Then let's remove this now if we know it's going to be removed

dotnet/samples/AgentSyntaxExamples/Example01_Agent.cs
```diff
@@ -0,0 +1,72 @@
...
+        // Create a chat for agent interaction. For more, see: Example03_Chat.
+        var chat = new TestChat();
+
+        // Respond to user input
+        await WriteAgentResponseAsync("Fortune favors the bold.");
+        await WriteAgentResponseAsync("I came, I saw, I conquered.");
+        await WriteAgentResponseAsync("Practice makes perfect.");
+
+        // Local function to invoke agent and display the conversation messages.
+        async Task WriteAgentResponseAsync(string input)
```
markwallace-microsoft, line 43: Rename to `InvokeAgentWithInputAsync` or just `InvokeAgentAsync`

dotnet/samples/AgentSyntaxExamples/Example01_Agent.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// <remarks>
+    /// For further exploration of <see cref="AgentChat"/>, see: Example03_Chat.
+    /// </remarks>
+    private sealed class TestChat : AgentChat
+    {
+        public IAsyncEnumerable<ChatMessageContent> InvokeAsync(
+            Agent agent,
+            CancellationToken cancellationToken = default) =>
+                base.InvokeAgentAsync(agent, cancellationToken);
+    }
```
markwallace-microsoft, line 71: If this is what we expect the default behaviour to be we should provide it. Developers should only need to provide a custom implementation if they are doing something custom.

dotnet/src/Agents/Abstractions/Extensions/ChatHistoryExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+/// <summary>
+/// Extension methods for <see cref="ChatMessageContent"/>
+/// </summary>
+internal static class ChatHistoryExtensions
+{
+    /// <summary>
+    /// Enumeration of chat-history in descending order.
+    /// </summary>
+    /// <param name="history">The chat-history</param>
+    public static IEnumerable<ChatMessageContent> ToDescending(this ChatHistory history)
```
markwallace-microsoft, line 17: Why are we doing this?

If the order of messages is important we should look at the design for this e.g. Roger suggested we should add a timestamp to chat messages. I'd like to understand do we have a good use case to justify this change.

dotnet/src/Agents/Abstractions/KernelAgent.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.SemanticKernel.Agents;
+
+/// <summary>
+/// Base class for agents utilizing <see cref="Microsoft.SemanticKernel.Kernel"/> plugins or services.
+/// </summary>
+public abstract class KernelAgent : Agent
```
markwallace-microsoft, line 7: Have you considered using an interface instead?

There is no behaviour provided here so if we had an interface it would provide more flexibility when designing class hierarchies.

dotnet/src/Agents/Core/ChatCompletionAgent.cs
```diff
@@ -0,0 +1,79 @@
...
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        var chatCompletionService = this.Kernel.GetRequiredService<IChatCompletionService>();
+
+        ChatHistory chat = new();
+        if (!string.IsNullOrWhiteSpace(this.Instructions))
+        {
+            chat.Add(new ChatMessageContent(AuthorRole.System, this.Instructions) { AuthorName = this.Name });
+        }
+        chat.AddRange(history);
```
markwallace-microsoft, line 41: Are we allowing someone to pass in a `history` that contains a new system message?

Elsewhere I think we disallow this and throw an exception.

dotnet/src/Agents/Core/ChatCompletionAgent.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// NOTE: Enable OpenAIPromptExecutionSettings.ToolCallBehavior for agent plugins.
+    /// (<see cref="ChatCompletionAgent.ExecutionSettings"/>)
+    /// </remarks>
+    public ChatCompletionAgent(
+        Kernel kernel,
+        string? description = null,
+        string? name = null)
+       : base(kernel)
+    {
+        this.Id = Guid.NewGuid().ToString();
```
markwallace-microsoft, line 75: Should we allow developers to provide their own ids?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -0,0 +1,200 @@
...
+        // Broadcast message to other channels (in parallel)
+        var channelRefs = this._agentChannels.Select(kvp => new ChannelReference(kvp.Value, kvp.Key));
+        this._broadcastQueue.Enqueue(channelRefs, messages);
+    }
+
+    /// <summary>
+    /// Add user message to chat history
+    /// </summary>
+    /// <param name="input">The user input, might be empty.</param>
+    public ChatMessageContent? AddUserMessage(string? input)
```
markwallace-microsoft, line 94: Let's remove this, see the comments here regarding this pattern: https://github.com/microsoft/semantic-kernel/pull/5800#discussion_r1555816250

dotnet/src/Agents/Abstractions/AgentChannel.cs
markwallace-microsoft: From the channel key documentation

    /// Any specific agent type may need to manage multiple channels.  For example, an
    /// agents targeting two different Azure OpenAI endpoints each require their own channel.
    /// In this case, the endpoint would be expressed as an additional key.

For the above example what would the additional endpoint specific keys look like?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        if (agent == null)
+        {
+            return this._history.ToDescendingAsync();
+        }
+
+        var channelKey = this.GetAgentHash(agent);
+        if (!this._agentChannels.TryGetValue(channelKey, out var channel))
+        {
+            return Array.Empty<ChatMessageContent>().ToAsyncEnumerable();
```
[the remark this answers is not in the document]
markwallace-microsoft: OK got it

dotnet/src/Agents/Abstractions/AgentChannel.cs
crickman: Ok, that comment needs fixing.  Will do.

Here's from the assistant work:

```c#
    protected override IEnumerable<string> GetChannelKeys()
    {
        // Distinguish from other channel types.
        yield return typeof(AgentChannel<OpenAIAssistantAgent>).FullName;

        // Distinguish between different Azure OpenAI endpoints or OpenAI services.
        yield return this._config.Endpoint ?? "openai"; 
    }
```

crickman: Updated:

```c#
    /// Two specific agents of the same type may each require their own channel.  This is
    /// why the channel type alone is insufficient.
    /// For example, two OpenAI Assistant agents each targeting a different Azure OpenAI endpoint
    /// would require their own channel. In this case, the endpoint could be expressed as an additional key.
```

markwallace-microsoft: So if I understand correctly if I have two different channels one which should use OpenAI and the other MistralAI I would do something like this

```csharp
    // OpenAI AgentChannel would do this
    protected override IEnumerable<string> GetChannelKeys()
    {
        // Distinguish from other channel types.
        yield return typeof(AgentChannel<OpenAIAssistantAgent>).FullName;

        // Distinguish between different Azure OpenAI endpoints or OpenAI services.
        yield return this._config.Endpoint ?? "OpenAI"; 
    }

    // MistralAI AgentChannel would do this
    protected override IEnumerable<string> GetChannelKeys()
    {
        // Distinguish from other channel types.
        yield return typeof(AgentChannel<OpenAIAssistantAgent>).FullName;

        // Distinguish between different Azure OpenAI endpoints or OpenAI services.
        yield return this._config.Endpoint ?? "MistralAI"; 
    }
```

Is this correct?

Using `Keys` in the name maybe what is confusing me. I typically associated keys with use in a key/value lookup. Whereas here I think it's the combination of keys the uniquely identifies the channel (is that right?).

Would `ChannelAttributes` be a more suitable name?

markwallace-microsoft (approved):

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -0,0 +1,184 @@
...
+using Microsoft.SemanticKernel.Agents.Extensions;
+using Microsoft.SemanticKernel.Agents.Internal;
+using Microsoft.SemanticKernel.ChatCompletion;
+
+namespace Microsoft.SemanticKernel.Agents;
+
+/// <summary>
+/// Point of interaction for one or more agents.
+/// </summary>
+public abstract class AgentChat
```
SergeyMenshykh, line 17: The {de}serialization aspect of the agent chat/chat history needs to be captured and designed if not already.

dotnet/src/Agents/Abstractions/ChatHistoryKernelAgent.cs
```diff
@@ -0,0 +1,29 @@
...
+
+/// <summary>
+/// A <see cref="KernelAgent"/> specialization bound to a <see cref="ChatHistoryChannel"/>.
+/// </summary>
+public abstract class ChatHistoryKernelAgent : KernelAgent, IChatHistoryHandler
+{
+    /// <inheritdoc/>
+    protected internal sealed override IEnumerable<string> GetChannelKeys()
+    {
+        yield return typeof(ChatHistoryChannel).FullName;
```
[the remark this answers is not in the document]
crickman: This is a good question and something I've spent a lot of time exploring.

By definition, this might imply two different `AgentChat` instances that present in an `AggregatorAgent` that participate in a parent `AgentChat`.  I'm exploring support for these concepts in the `Agent as a Plug-in` task (which also includes `AggregatorAgent`.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Totally agree.  I know we've discussed this on several occasions...and I'm not seeing an existing tracking task, so I've captured it here:

https://github.com/microsoft/semantic-kernel/issues/5828

Please feel free to add detail.

dotnet/samples/AgentSyntaxExamples/AgentSyntaxExamples.csproj
```diff
@@ -0,0 +1,52 @@
...
+    <ProjectReference Include="..\..\src\Agents\Abstractions\Agents.Abstractions.csproj" />
+    <ProjectReference Include="..\..\src\Connectors\Connectors.OpenAI\Connectors.OpenAI.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="Resources\*">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </EmbeddedResource>
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Resources\" />
```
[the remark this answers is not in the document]
crickman: Updated

dotnet/src/Agents/Abstractions/AgentChannel.cs
crickman: Wouldn't the Mistral AI case be within the scope of the `ChatCompletionAgent` (and use the `ChatHistoryChannel`)?  The implementation you are citing is specific to OpenAI Assistant API, which only has OpenAI and Azure support to my knowledge.  We specify `endpoint` for Azure, but it is omitted when targeting Open AI services directly.

Even if a 3rd party were to stand up their own Open AI Assistants API, the `endpoint` value would be unique and channel managment would proceed as expected.

If MistralAI were to publish their own assistants API, this case might be addressed by a dedicated agent/channel...not the Open AI Assistant Agent.

matthewbolanos (approved):

PR #5718, opened 2024-04-01
.Net: Update stepwise planner to accept optional chatHistory to resume execution as needed.

lifeng225:
### Contribution Checklist


HadiKotaich (approved):

markwallace-microsoft:
@lifeng225 Thanks for the PR. Can you provide some background on the motivation for this change i.e. describe the issue you are having. Thanks.

lifeng225:
> @lifeng225 Thanks for the PR. Can you provide some background on the motivation for this change i.e. describe the issue you are having. Thanks.

Hi @markwallace-microsoft , thanks for looking the PR. The intention of the PR is to make stepwise planner to be able to resume a task run with history provided. In FunctionCallingStepwisePlanner's ExecuteCoreAsync function, it does the step iterations with a MaxIterations limit from options. This makes sense so the run could be deterministic. But we also want to resume the step from wherever it stops later as needed. From the FunctionCallingStepwisePlannerResult object, we have the ChatHistory object, so we want to pass the object back into the planner for the next run. That's the idea of this PR, to be able to accept a chatHistory object for future pickup job of the same task run. Let me know whether I made myself clear or any further questions, thanks.

BTW, I don't have any edit permissions when I logged in with my "<alias>-microsoft" account so I have to use my personal account somehow for the PR. Not sure what I missed on microsoft account.

Leo Hu (lifhu@microsoft.com)

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -75,7 +80,11 @@ public sealed class FunctionCallingStepwisePlanner
         // Create and invoke a kernel function to generate the initial plan
         var initialPlan = await this.GeneratePlanAsync(question, clonedKernel, logger, cancellationToken).ConfigureAwait(false);
```
alliscode, line 81: The `initialPlan` appears to be unused in the case of a `chatHistoryForSteps` being provided. Since that step is making a call to the LLM it would be good to not waste the token/cycles.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
```diff
@@ -75,7 +80,11 @@ public sealed class FunctionCallingStepwisePlanner
...
 
-        var chatHistoryForSteps = await this.BuildChatHistoryForStepAsync(question, initialPlan, clonedKernel, promptTemplateFactory, cancellationToken).ConfigureAwait(false);
+        if (chatHistoryForSteps is null)
+        {
+            // Build chat history for the first step
+            chatHistoryForSteps = await this.BuildChatHistoryForStepAsync(question, initialPlan, clonedKernel, promptTemplateFactory, cancellationToken).ConfigureAwait(false);
+        }
```
alliscode, line 87: I like the idea of providing a `ChatHistory` to be used inside the planner but I'm a bit worried that the effect of providing it in `InvokePlanAsync` will not be clear to people. As it is now, providing a `ChatHistory` causes the initial planning step to be ignored/skipped which is totally valid for the scenario you describe when you're continuing on from a previous planner run. However, I could see this also being used to provide more context to the planner for a first run, in which case you would likely still want the initial planning step to happen.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
lifeng225: Thanks, updated.

dotnet/src/Planners/Planners.OpenAI/Stepwise/FunctionCallingStepwisePlanner.cs
lifeng225: Thanks, I got what you meant, so we might want to append the context in addition to the initial plan steps in some cases. It seems like we can have another flag to indicate this. But do you think it makes sense to keep it for future PR when needed?

alliscode (approved):

markwallace-microsoft (approved):

PR #5750, opened 2024-04-03
.Net: Google connector API version selection

### LINKED ISSUE #5659: .Net Gemini Connector - Allow Version Selection and Base Url overriding.
> The `v1beta` part of the URL is going to change and there may be times when multiple versions will be supported at the same time.

> We should have:

> A central place where the default API version is set
> We should consider if we want to allow developers to change this. We do allow this in Azure OpenAI. It's a risky thing for a developer to do but that doesn't mean someone may have a good reason and we don't necessarily want to block them.

Basically when using `AddGeminiChatCompletion` we should be able to specify the target version in the service `Settings`, this will be adjusting the correct Uri of the client `v1beta` or `v1`, etc.

Not very sure about be able to override the `endpoint` as I don't see a potential mock or virtualization of `GCP` to be targetted (edge case).

Krzysztof318:
### Motivation and Context

Closes #5659

### Description

The update introduces support for different versions of the Google API in various services and clients. A new enum, 'GoogleApiVersion', has been added to represent stable and beta versions of the Google API. Affected classes have been updated to accept this new parameter, and use its value when constructing API endpoints.

GoogleAI endpoints currently support only BETA.

cc: @RogerBarreto 

### Contribution Checklist


Krzysztof318:
Done. I think this way is ok:
image

dotnet/src/Connectors/Connectors.Google/Core/ClientBase.cs
```diff
@@ -109,4 +109,12 @@ protected void Log(LogLevel logLevel, string? message, params object[] args)
...
         }
     }
+
+    protected static string GetApiVersionSubLink(GoogleApiVersion apiVersion)
+        => apiVersion switch
+        {
+            GoogleApiVersion.Stable => "v1",
+            GoogleApiVersion.Beta => "v1beta",
+            _ => throw new NotSupportedException($"Google API version {apiVersion} is not supported.")
+        };
```
markwallace-microsoft, line 119: The advantage of this approach is we are being strict about the versions we support. 

There are some disadvantages e.g.

- Stable may be v1 now but this will change over time but what happens if there are multiple stable versions supported. Developers should be able to select the exact version they want.
- If `v1.1beta` is released developers cannot switch to it to test if their existing functionality still works es expected with the newer release, they will need to wait for a new release of SK.

Elsewhere we are allowing the API version to be passed in as a string value which we insert into the URL. This provides control to the developer to try out new versions as soon as they are available. @RogerBarreto Thoughts?

Krzysztof318: I can change this to string but... if we decide to move implementation to use google nuget, there is no place to specify api version. There are just two different nugets, package stable https://cloud.google.com/vertex-ai/docs/reference/rpc/google.cloud.aiplatform.v1 and package beta https://cloud.google.com/vertex-ai/docs/reference/rpc/google.cloud.aiplatform.v1beta1 (not available nuget for vertex currently)

RogerBarreto: Follow the approach from Azure SDK, enum for each version:

[Check here](https://github.com/Azure/azure-sdk-for-net/blob/d12489ab27079466308dbd9dcc89879be7f02fe7/sdk/openai/Azure.AI.OpenAI/src/Generated/OpenAIClientOptions.cs#L14)

RogerBarreto: @Krzysztof318 except for the approach how to enumerate them, which avoids @markwallace-microsoft concerns as well, we work with Latest and Specific enums.

Krzysztof318: I changed to V1 and V1Beta but I didn't add Latest property because it doesn't make sense there. VertexAI gemini works now only with V1, not beta, but GoogleAI works only with Beta now.

RogerBarreto: Should be different enumerators for each Platform, so you can have a latest for GoogleAI and a latest for VertexAI side by side.

Latest is not super necessary ATM, but thats the main idea, since those versions are related to different API's.

dotnet/src/Connectors/Connectors.Google/GoogleApiVersion.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticKernel.Connectors.Google;
+
+#pragma warning disable CA1707 // Identifiers should not contain underscores
+
+/// <summary>
+/// Represents the version of the Google API.
+/// </summary>
+public enum GoogleApiVersion
```
RogerBarreto, line 10:
```suggestion
public enum GoogleAIVersion
{
    V1_Beta
}

public enum VertexAIVersion
```

dotnet/src/Connectors/Connectors.Google/Core/ClientBase.cs
Krzysztof318: yeah, I changed that, I don't want to introduce latest due to vertex didn't work correctly with Beta now in my tests.

dotnet/src/Connectors/Connectors.Google/Extensions/GoogleAIServiceCollectionExtensions.cs
```diff
@@ -47,12 +50,14 @@ public static class GoogleAIServiceCollectionExtensions
...
     /// <param name="modelId">The model for embeddings generation.</param>
     /// <param name="apiKey">The API key for authentication Gemini API.</param>
+    /// <param name="apiVersion">The version of the Google API.</param>
     /// <param name="serviceId">Optional service ID.</param>
     /// <returns>The updated service collection.</returns>
     public static IServiceCollection AddGoogleAIEmbeddingGeneration(
         this IServiceCollection services,
         string modelId,
         string apiKey,
+        GoogleAIVersion apiVersion = GoogleAIVersion.V1_Beta, // todo: change beta to stable when stable version will be available
```
markwallace-microsoft, line 60: @RogerBarreto can we create an issue for this and put in the future sprint so we don't forget to do this?

markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.Google/Extensions/GoogleAIServiceCollectionExtensions.cs
RogerBarreto: We should be using the `Latest` as the default instead of providing an specific in here, this way this would be used as default instead of a specific enumerable, so every new version there's just one place for changes.

Similar as I pointed out with Azure SDK before in [here](https://github.com/Azure/azure-sdk-for-net/blob/d12489ab27079466308dbd9dcc89879be7f02fe7/sdk/openai/Azure.AI.OpenAI/src/Generated/OpenAIClientOptions.cs#L16C9-L16C89).

Will create an issue to get this updated as well as to keep track of versions

westey-m: In this case we should clearly call out in the XML doc for the parameter that it will always default to the latest, so that consumers know that they need to specify the version if they do not wish to be upgraded silently with a package version update.

RogerBarreto: @westey-m Good point, added to the issue your observation.

PR #5767, opened 2024-04-03
.Net: Include type fields and properties descriptions into JSON schema

SergeyMenshykh:


dotnet/src/SemanticKernel.Abstractions/Functions/KernelParameterMetadata.cs
```diff
@@ -140,7 +140,13 @@ internal static InitializedSchema InferSchema(Type? parameterType, object? defau
                         description += $"{(needsSpace ? " " : "")}(default value: {stringDefault})";
                     }
 
+                    // Register the handler for the 'System.ComponentModel.DescriptionAttribute' attribute to provide descriptions for .NET type members, such as fields and properties.
+                    // See the Attributes section for more details - https://docs.json-everything.net/schema/schemagen/schema-generation/.
+                    // This line must precede the JsonSchemaBuilder creation; otherwise, the registration will not take place.
+                    AttributeHandler.AddHandler<DescriptionAttributeHandler>();
```
stephentoub, line 146: Could we focus instead on https://github.com/microsoft/semantic-kernel/pull/5635 rather than increasing a reliance on this?

SergeyMenshykh: This was a try to quickly support for description in JSON schema. It was decided not to proceed with the fix taking into account it's static nature.

stephentoub: The other one already does.

PR #5771, opened 2024-04-04
.Net: Update MistralAI kernel syntax samples

markwallace-microsoft:
### Motivation and Context

Adding a new function calling sample

### Contribution Checklist


SergeyMenshykh (approved):

Krzysztof318:
I'm not sure this is good way, in my opinion better will be unify samples due to many models has almost identical samples.

markwallace-microsoft:
> I'm not sure this is good way, in my opinion better would be unify samples due to many models has almost identical samples.

There is a separate workstream underway to do this. Depending on when that work completes and when these changes get merged to main these samples may become obsolete and will be removed.

Krzysztof318:
@markwallace-microsoft okay, thanks for clarification

PR #5788, opened 2024-04-04
.Net: [Feature Branch][Option 4] Change IFunctionFilter signature by adding delegate to next filter or function

dmytrostruk:
### Motivation and Context

ADR: https://github.com/microsoft/semantic-kernel/pull/5787

New API for function filter - instead of `Invoking/Invoked` filter methods, it has one `Invocation` method that receives context and delegate to the next filter in pipeline or function itself.

Abstraction:

```csharp
public interface IFunctionFilter
{
    Task OnFunctionInvocationAsync(FunctionInvocationContext context, Func<FunctionInvocationContext, Task> next);
}
```

Usage:

```csharp
public class MyFilter : IFunctionFilter
{
    public async Task OnFunctionInvocationAsync(FunctionInvocationContext context, Func<FunctionInvocationContext, Task> next)
    {
        // Perform some actions before function invocation
        await next(context);
        // Perform some actions after function invocation
    }
}
```

Exception handling:

```csharp
public async Task OnFunctionInvocationAsync(FunctionInvocationContext context, Func<FunctionInvocationContext, Task> next)
{
    try
    {
        await next(context);
    }
    catch (Exception exception)
    {
        this._logger.LogError(exception, "Something went wrong during function invocation");

        // Example: override function result value
        context.Result = new FunctionResult(context.Function, "Friendly message instead of exception");

        // Example: Rethrow another type of exception if needed
        throw new InvalidOperationException("New exception");
    }
}
```

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Filters/Function/IFunctionFilter.cs
```diff
@@ -1,24 +1,22 @@
...
-
-    /// <summary>
-    /// Method which is executed after function invocation.
+    /// Method which is called asynchronously before function invocation.
     /// </summary>
-    /// <param name="context">Data related to function after invocation.</param>
-    void OnFunctionInvoked(FunctionInvokedContext context);
+    /// <param name="context">Instance of <see cref="FunctionInvocationContext"/> with function invocation details.</param>
+    /// <param name="next">Delegate to the next filter in pipeline or function itself. If it's not invoked, next filter or function won't be invoked.</param>
+    Task OnFunctionInvocationAsync(FunctionInvocationContext context, FunctionInvocationCallback next);
```
stephentoub, line 21: Why is FunctionInvocationCallback needed, rather than just having that be an IFunctionFilter?

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    internal object? ResultValue { get; private set; }
+
+    /// <summary>
+    /// Sets an object to use as the overridden new result for the function's invocation.
+    /// </summary>
+    /// <param name="value">The value to use as the new result of the function's invocation.</param>
+    public void SetResultValue(object? value)
+    {
+        this.ResultValue = value;
```
stephentoub, line 59: It's strange that calling SetResultValue won't impact Result.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -163,35 +162,29 @@ internal KernelFunction(string name, string? pluginName, string description, IRe
 
         TagList tags = new() { { MeasurementFunctionTagName, this.Name } };
         long startingTimestamp = Stopwatch.GetTimestamp();
-        FunctionResult? functionResult = null;
+        FunctionResult? functionResult = new(this);
```
stephentoub, line 165: Why is this initialized to something other than null? In what situation do we get to the return without this having been written to?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -279,36 +279,32 @@ public CultureInfo Culture
     #region Internal Filtering
 
     [Experimental("SKEXP0001")]
-    internal FunctionInvokingContext? OnFunctionInvokingFilter(KernelFunction function, KernelArguments arguments)
+    internal async Task<FunctionInvocationContext?> OnFunctionInvocationAsync(
```
stephentoub, line 282: Why does this return a nullable context? In what situation does it ever return null?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -163,35 +162,29 @@ internal KernelFunction(string name, string? pluginName, string description, IRe
...
-            functionResult = await this.InvokeCoreAsync(kernel, arguments, cancellationToken).ConfigureAwait(false);
+                // Update context with result.
+                context.Result = functionResult;
+            }).ConfigureAwait(false);
 
-            // Invoke the post-invocation filter. If it requests cancellation, throw.
-            var invokedContext = kernel.OnFunctionInvokedFilter(arguments, functionResult);
-
-            if (invokedContext is not null)
+            if (invocationContext is not null)
```
stephentoub, line 180: Why is this null check needed?

dotnet/src/SemanticKernel.Abstractions/Filters/Function/IFunctionFilter.cs
dmytrostruk: > If it's to be able to handle the final step where the actual function is invoked

That is an answer, we also need to be able to wrap function invocation in `try/catch` block to handle exceptions.

> that can be done by just implementing the interface with a private type

Not sure I understand this approach, could you please provide more details?

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvocationContext.cs
dmytrostruk: Yeah, this is how result overriding is designed in today's version of SK, I think we should redesign it as well. Maybe we should allow to set entire `Result` object.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: Currently we initialize it as null, but in fact, `InvokeAsync` in `KernelFunction` never returns null, the result type is also non-nullable, it's either some result or exception. With exception handling scenarios, I thought it would be better to initialize `FunctionResult` with a link to `KernelFunction`, so user will be able to understand which function failed, with with `null` result value. Otherwise, we need to think how we can tell user an info about failed function, maybe put it in specific exception type?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: > Why does this return a nullable context? In what situation does it ever return null?

In situation where there are no filters registered in kernel. No filters - no context, just function invocation and result.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: Same answer as above, `null` context means that there were no filters triggered during function execution. This allows us to understand whether we should override final function result or just return it as it is.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -279,24 +271,22 @@ internal KernelFunction(string name, string? pluginName, string description, IRe
...
-                // Invoke pre-invocation filter. If it requests cancellation, throw.
-                var invokingContext = kernel.OnFunctionInvokingFilter(this, arguments);
-
-                if (invokingContext?.Cancel is true)
+                var invocationContext = await kernel.OnFunctionInvocationAsync(this, arguments, (context) =>
                 {
-                    throw new OperationCanceledException("A function filter requested cancellation before function invocation.");
-                }
+                    // Invoke the function and get its streaming enumerator.
+                    enumerator = this.InvokeStreamingCoreAsync<TResult>(kernel, context.Arguments, cancellationToken).GetAsyncEnumerator(cancellationToken);
```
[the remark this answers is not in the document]
dmytrostruk: That's one of the limitations I described in ADR for streaming scenarios. I'm not sure how to override streamed response and handle exceptions during streaming yet. Will think about it.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -308,31 +298,32 @@ internal KernelFunction(string name, string? pluginName, string description, IRe
                 throw;
             }
 
-            // Ensure we clean up after the enumerator.
-            await using (enumerator.ConfigureAwait(false))
+            if (enumerator is not null)
```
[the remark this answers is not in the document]
dmytrostruk: It is not expected to have enumerator as null, but in this implementation enumerator is assigned in filter callback, so when I define an enumerator I need to assign some initial value to it, otherwise there will be a warning that I'm accessing unassigned variable.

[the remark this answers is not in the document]
stephentoub: > otherwise there will be a warning that I'm accessing unassigned variable.

If there's no way it could be null, you can just use `!` to suppress the warning. (Ideally instead you'd use Debug.Assert to suppress it, but that won't do so in a netstandard2.0 build.)

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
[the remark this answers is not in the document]
stephentoub: >  I'm not sure how to override streamed response and handle exceptions during streaming yet.

We can't just let the filter supply its own IAsyncEnumerable? It could just wrap the original, and unless it eats exceptions (which would be up to it as a filter), then any exceptions that emerge from the underlying enumerable would also surface from the wrapper. Basically the filter creates a decorator.

dotnet/src/SemanticKernel.Abstractions/Filters/Function/IFunctionFilter.cs
stephentoub: > Not sure I understand this approach, could you please provide more details?

```C#
private sealed class TerminalFilter(Kernel kernel, KernelFunction function) : IFunctionFilter
{
    public async Task OnFunctionInvocationAsync(FunctionInvocationContext context, FunctionInvocationCallback next)
    {
        Debug.Assert(next is null);
        context.Result = await function.InvokeAsync(kernel, context.Arguments, cancellationToken);
    }
}
...
IFunctionFilter last = new TerminalFilter(kernel, this);

File: dotnet/src/SemanticKernel.Abstractions/Kernel.cs (hunk 4 above)
Comment (stephentoub) [reply dotnet/src/SemanticKernel.Abstractions/Kernel.cs:282]: Even in that case, it's returning non-null:
https://github.com/microsoft/semantic-kernel/pull/5788/files/00175eba09591d8a9bac78d6a3428f8e4d549195#diff-d3103bc841c57ed1f9f7cbd3b3f3be955eca5f85913ae67041a13bd8b66b3daeR287

File: dotnet/src/SemanticKernel.Abstractions/Filters/Function/IFunctionFilter.cs (hunk 1 above)
Comment (dmytrostruk) [reply dotnet/src/SemanticKernel.Abstractions/Filters/Function/IFunctionFilter.cs:21]: @stephentoub So, in case of having `IFunctionFilter` as `next` parameter instead of `FunctionInvocationCallback`, the usage will look like this:
```csharp
public async Task OnFunctionInvocationAsync(FunctionInvocationContext context, IFunctionFilter next)
{
    // Perform some actions before function invocation
    await next.OnFunctionInvocationAsync(context, ???);
    // Perform some actions after function invocation
}
```

First of all, users will have to call `OnFunctionInvocationAsync` method manually, which is more complicated rather than do just `next(context)`. The other thing is that users will have to pass something as second parameter (question marks in code example) to this method (which should be reference to the next filter in pipeline or function itself. Maybe it will be possible to create extension method for IFunctionFilter interface that accepts only context, but I will need to double check if that will work. 

The other thing is that with private function filter that performs function invocation (e.g. `TerminalFilter` from your sample) - it will be visible in stack trace of exception, which may be confusing, especially when filters weren't registered at all. It shouldn't be a big problem though.

I guess my main concern is that `IFunctionFilter` as second parameter looks more complicated than callback. What do you think?

stephentoub: That's fair. Do we need the dedicated delegate type, or can it just use a Func?

dmytrostruk: > Do we need the dedicated delegate type, or can it just use a Func?

Just `Func` should work. I decided to choose dedicated delegate type to follow the same pattern as in ASP.NET filters just to be consistent, but `Func` should be fine. I will experiment with it.

dmytrostruk: Name of dedicated delegate type looks a little bit shorter than `Func<FunctionInvocationContext, Task>`, but I guess it's the only advantage over `Func` in this case. So, I updated `IFunctionFilter` interface to accept `Func` instead of dedicated delegate type.

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvocationContext.cs
dmytrostruk: `SetResultValue` was removed and `Result` property is now available with public setter. Usage example:
```csharp
context.Result = new FunctionResult(context.Function, "Friendly message instead of exception");
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: Based on latest changes, this was reverted to be initialized as null.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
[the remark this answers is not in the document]
dmytrostruk: That's a great idea. I implemented it and it allows to modify streaming response or catch exceptions during streaming. I also provided samples with these scenarios. Thanks a lot!

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvocationContext.cs
markwallace-microsoft: Using the code above has the side effect of removing the current `culture` and `metadata`. Developers will need to take care to avoid this, the old pattern didn't have this downside.
One other thing to consider is what happens if we add a new property to `FunctionResult`. Existing filter implementations would need to be updated to prevent data loss.

Would `context.Result = new FunctionResult(context.Result, "Friendly message instead of exception");` be a better pattern?

dotnet/samples/KernelSyntaxExamples/Example76_Filters.cs
```diff
@@ -99,30 +197,132 @@ public FirstPromptFilter(ITestOutputHelper output)
...
+            await next(context);
+
+            // Example: get function result value
+            var value = context.Result!.GetValue<object>();
 
             // Example: get token usage from metadata
             var usage = context.Result.Metadata?["Usage"];
+
+            // Example: override function result value
+            context.Result = new FunctionResult(context.Function, "Result from filter");
```
markwallace-microsoft, line 219: If we allows filters to modify the metadata then this will need to change.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -169,30 +168,21 @@ internal KernelFunction(string name, string? pluginName, string description, IRe
...
-            if (invokedContext?.Cancel is true)
-            {
-                throw new OperationCanceledException("A function filter requested cancellation after function invocation.");
-            }
+            // Apply any changes from the function filters to final result.
+            functionResult = new FunctionResult(
+                this,
+                invocationContext.Result?.Value ?? functionResult?.Value,
+                functionResult?.Culture,
+                functionResult?.Metadata);
```
markwallace-microsoft, line 185: Are we not allowing filters to modify the metadata and culture?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -279,36 +279,32 @@ public CultureInfo Culture
...
             {
-                this._functionFilters[i].OnFunctionInvoking(context);
+                if (index < this._functionFilters.Count)
+                {
+                    await this._functionFilters[index].OnFunctionInvocationAsync(context, (context) => InvokeFilterOrFunctionAsync(context, index + 1)).ConfigureAwait(false);
+                }
+                else
+                {
+                    await functionCallback(context).ConfigureAwait(false);
+                }
```
markwallace-microsoft, line 300: This code is non trivial so comments would help

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -287,16 +277,20 @@ internal KernelFunction(string name, string? pluginName, string description, IRe
...
                 {
-                    throw new OperationCanceledException("A function filter requested cancellation before function invocation.");
-                }
+                    // Invoke the function and get its streaming enumerable.
+                    var enumerable = this.InvokeStreamingCoreAsync<TResult>(kernel, context.Arguments, cancellationToken);
 
-                // Invoke the function and get its streaming enumerator.
-                enumerator = this.InvokeStreamingCoreAsync<TResult>(kernel, arguments, cancellationToken).GetAsyncEnumerator(cancellationToken);
+                    // Update context with enumerable as result value.
+                    context.Result = new FunctionResult(this, enumerable);
```
SergeyMenshykh, line 286: Kernel culture can be added to the function result as well.

dotnet/samples/KernelSyntaxExamples/Example76_Filters.cs
RogerBarreto: @markwallace-microsoft, @dmytrostruk If we want to change metadata wouldn't this be the way?

```suggestion
            var modifiedMetadata = new Dictionary<string, object?>(context.Result.Metadata)) 
            {
                ["new-data-index"] = "new-data-value"
            }
            context.Result = new FunctionResult(context.Function, "Result from filter", modifiedMetadata);
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -287,16 +277,20 @@ internal KernelFunction(string name, string? pluginName, string description, IRe
...
-                // Invoke pre-invocation filter. If it requests cancellation, throw.
-                var invokingContext = kernel.OnFunctionInvokingFilter(this, arguments);
-
-                if (invokingContext?.Cancel is true)
+                var invocationContext = await kernel.OnFunctionInvocationAsync(this, arguments, (context) =>
                 {
-                    throw new OperationCanceledException("A function filter requested cancellation before function invocation.");
-                }
+                    // Invoke the function and get its streaming enumerable.
+                    var enumerable = this.InvokeStreamingCoreAsync<TResult>(kernel, context.Arguments, cancellationToken);
```
RogerBarreto, line 283: When triggering the filter from an AsyncEnumerable we should have a flag on the context so the caller can know when to handle it properly, since we are using the same method.

stephentoub: > so the caller can know when to handle it properly instead of need to check Value is a IAsyncEnumerable impl.

I'm not seeing the concern. What is the code someone would write with or without such an additional property? Are you concerned about a non-streaming function returning an IAsyncEnumerable as its result, and the filter confusing that with streaming? Does the distinction matter at that point?

dmytrostruk: @stephentoub We just discussed this internally. So, I believe there are 2 different cases:
- You can invoke a function with `InvokeAsync` and `InvokeStreamingAsync`. `InvokeAsync` will return some result `T`, while `InvokeStreamingAsync` will return `IAsyncEnumerable<T>`.
- `KernelFunctionFromMethod` can return `IAsyncEnumerable<T>` itself.

When I write my filter, I don't know in what way my function was invoked, whether it's `InvokeAsync` or `InvokeStreamingAsync`, which means that in filter every time I will have to check whether I need to override result by simply setting a new value or by using `await foreach` construction. Since same function can be called in both streaming and non-streaming way, the filter will always need to have this check in place.

We were thinking about two different approaches:
- Define a separate type of filter for streaming scenarios (e.g. `IStreamingFunctionFilter`). 
- Define an attribute which you can apply to filter to specify whether its purpose is to work with streaming or non-streaming. No attribute will mean that it will be executed in both cases. Later, this attribute can be extended to specify list of function/plugin names the filter should be triggered for. The attribute should help to avoid `if` conditions to verify whether it's safe to override `IAsyncEnumerable` etc.

What do you think about it?

dotnet/samples/KernelSyntaxExamples/Example76_Filters.cs
dmytrostruk: @RogerBarreto Yes, this should be the pattern. I will add more tests and examples to make sure it is working as expected.

dotnet/src/SemanticKernel.Abstractions/Filters/Function/FunctionInvocationContext.cs
dmytrostruk: > Would `context.Result = new FunctionResult(context.Result, "Friendly message instead of exception");` be a better pattern?

I applied this pattern. `FunctionResult` has new constructor that accepts instance of `FunctionResult` to copy data from and additional `value` parameter to override result value.

> We should also consider obsoleting the current `FunctionResult` constructor in favour of init properties.

We can keep existing constructor without obsoleting it, but make every new property with `init` capability. I updated `Function`, `Metadata` and `Culture` properties with `init`. I also changed the visibility of `Culture` property from `internal` to `public`, since it's possible to set it via constructor of `FunctionResult`, but after that users don't have access to it, which is a little bit confusing. 

With latest updates it's possible to modify `Culture` and `Metadata` inside filter by creating new instance of `FunctionResult` and I've added more tests and examples to verify that it works correctly.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: FYI: With latest changes, it's again initialized as non-nullable. This is to make `FunctionResult` property inside filter context as non-nullable. In this case, users won't be forced to check whether `context.Result` is null every time, which simplifies the usage. Before filter invocation, `context.Result` will contain information about Function and kernel Culture, with `null` Value. After invocation, it will be the same, but Value will hold result of function, and Metadata dictionary will be populated.

dotnet/samples/KernelSyntaxExamples/Example76_Filters.cs
dmytrostruk: I've added more tests and examples to verify that updating Metadata works as expected.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: Fixed.

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: I simplified the code, moved it to separate method and added the comments.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
dmytrostruk: Good point, updated.

PR #5800, opened 2024-04-08
.Net: Function call content types

SergeyMenshykh:
Today, in SK, LLM function calling is supported exclusively by the OpenAI connector, and the function calling model is specific to that connector. The new AI connectors being added to SK, which support function calling, introduce their specific models for function calling. The design, where each new connector introduces its own specific model class for function calling, does not scale well from the connector development perspective and does not allow for polymorphic use of connectors by the SK consumer code.

This ADR describes the high-level details of the service-agnostic function-calling model classes, while leaving the low-level details to the implementation phase. Additionally, this ADR outlines the identified options for various aspects of the design.

Requirements - https://github.com/microsoft/semantic-kernel/issues/5153

### Description
ADR PR:  https://github.com/microsoft/semantic-kernel/pull/5696

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContent.cs
SergeyMenshykh, line 1: @markwallace-microsoft @rogerio-barreto, please provide your suggestions for alternative names for the  FunctionCallContent and FunctionResultContent classes.

dotnet/src/InternalUtilities/src/System/IListExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        if (source is null)
+        {
+            throw new ArgumentNullException(nameof(source));
+        }
+
+        foreach (var item in source)
+        {
+            target.Add(item);
+        }
```
stephentoub, line 33:
```suggestion
        if (target is List<T> list)
        {
            list.AddRange(source);
        }
        else
        {
            foreach (var item in source)
            {
                target.Add(item);
            }
        }
```

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionResultContent.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Id { get; }
+
+    /// <summary>
+    /// The plugin name.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? PluginName { get; }
+    /// <summary>
```
stephentoub, line 25:
```suggestion

    /// <summary>
```

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1026,15 +1058,30 @@ private static ChatRequestMessage GetRequestMessage(ChatMessageContent message)
...
+            // Handling function calls supplied via ChatMessageContent.Items collection elements of the FunctionCallContent type.
+            if (message.Items.OfType<FunctionCallContent>() is { } functionCallContents && functionCallContents.Any())
             {
-                foreach (ChatCompletionsToolCall tool in tools)
+                var ftcs = new List<ChatCompletionsToolCall>(tools ?? Enumerable.Empty<ChatCompletionsToolCall>());
+
+                foreach (var fcContent in functionCallContents)
                 {
-                    asstMessage.ToolCalls.Add(tool);
+                    if (!ftcs.Any(ftc => ftc.Id == fcContent.Id))
```
stephentoub, line 1068: This is O(N^2).

Separate from complexity, why is this check even needed? In what situation do we end up with two twoo calls with the same ID and we don't want to invoke it that many times?

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContent.cs
```diff
@@ -0,0 +1,113 @@
...
+            arguments: arguments);
+    }
+
+    /// <summary>
+    /// Invokes the function represented by the function call content type.
+    /// </summary>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The result of the function's execution.</returns>
+    public async Task<FunctionResultContent> InvokeAsync(Kernel kernel, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
stephentoub: Where is this even being used, outside of tests?

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,49 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
-                        JsonSerializer.Serialize((await function.InvokeAsync(kernel, arguments)).GetValue<object>()) :
-                        "Unable to find function. Please try again!";
-
-                    chatHistory.Add(new ChatMessageContent(
-                        AuthorRole.Tool,
-                        content,
-                        metadata: new Dictionary<string, object?>(1) { { OpenAIChatMessageContent.ToolIdProperty, toolCall.Id } }));
+                    try
+                    {
+                        FunctionResultContent functionResult = await functionCall.InvokeAsync(kernel); // Executing each function. Can be done in parallel.
```
stephentoub, line 98: "Can be done in parallel"... but it's not here. The comment is a bit confusing.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,49 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
+
+                        chatHistory.AddMessage(AuthorRole.Tool, functionResult); // Adding function result to chat history.
+                    }
+                    catch (Exception ex)
+                    {
+                        chatHistory.AddMessage(AuthorRole.Tool, new FunctionResultContent(functionCall, ex));
+                        // Adding exception to chat history.
+                        // or
+                        //string message = $"Error details that LLM can reason about.";
```
stephentoub, line 107:
```suggestion
                        //string message = "Error details that LLM can reason about.";
```

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -968,39 +968,71 @@ private static ChatRequestMessage GetRequestMessage(ChatRole chatRole, string co
...
             if (message.Metadata?.TryGetValue(OpenAIChatMessageContent.ToolIdProperty, out object? toolId) is true &&
                 toolId?.ToString() is string toolIdString)
             {
-                return new ChatRequestToolMessage(message.Content, toolIdString);
+                return new[] { new ChatRequestToolMessage(message.Content, toolIdString) };
             }
 
+            // Handling function call results represented by the FunctionResultContent type.
+            // Example: new ChatMessageContent(AuthorRole.Tool, items: new ChatMessageContentItemCollection { new FunctionResultContent(functionCall, result) })
+            if (message.Items.OfType<FunctionResultContent>() is { } resultContents && resultContents.Any())
```
stephentoub, line 990: OfType will never return null

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1026,15 +1058,30 @@ private static ChatRequestMessage GetRequestMessage(ChatMessageContent message)
                 }
             }
 
-            if (tools is not null)
+            // Handling function calls supplied via ChatMessageContent.Items collection elements of the FunctionCallContent type.
+            if (message.Items.OfType<FunctionCallContent>() is { } functionCallContents && functionCallContents.Any())
```
stephentoub, line 1062: OfType will never return null

dotnet/src/InternalUtilities/src/System/IListExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        if (target is null)
+        {
+            throw new ArgumentNullException(nameof(target));
+        }
+
+        if (source is null)
+        {
+            throw new ArgumentNullException(nameof(source));
+        }
```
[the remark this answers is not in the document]
SergeyMenshykh: The `Debug.Assert` will only work in debug builds, correct? In release builds, a NullReferenceException will be thrown if one or both of the parameters are null, without providing the parameter name information.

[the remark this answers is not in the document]
stephentoub: Yes, but this is internal. If this is called with null, doesn't that mean _we_ have a bug?

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
```diff
@@ -69,6 +70,27 @@ public ChatHistory(IEnumerable<ChatMessageContent> messages)
     public void AddMessage(AuthorRole authorRole, ChatMessageContentItemCollection contentItems, Encoding? encoding = null, IReadOnlyDictionary<string, object?>? metadata = null) =>
         this.Add(new ChatMessageContent(authorRole, contentItems, null, null, encoding, metadata));
 
+    /// <summary>
+    /// Adds <see cref="ChatMessageContent"/> with content items./>
+    /// </summary>
+    /// <param name="authorRole">Role of the message author</param>
+    /// <param name="items">The content items.</param>
+    [Experimental("SKEXP0001")]
+    public void AddMessage(AuthorRole authorRole, params KernelContent[] items)
```
RogerBarreto, line 79: Arent we going with `Add` and not use property or arguments anymore ? 

This might not be needed at all.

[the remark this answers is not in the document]
SergeyMenshykh: It could be slightly more convenient for callers to use a shorter syntax instead of the long one (commented out):
image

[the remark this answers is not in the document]
stephentoub: > It could be slightly more convenient for callers

Until they need to do something this doesn't support, like add AuthorName, and then they fall off a cliff. The params also means you can call this with no args beyond author role, e.g.
```C#
chatHistory.AddMessage(AuthorRole.User);
```
which is a pit of failure.

I don't think we should be adding any more of these "optimize for a very specific use case" helpers onto ChatHistory.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
```diff
@@ -169,6 +169,16 @@ public ChatMessageContent()
...
 
+    /// <summary>
+    /// Returns a list of function calls.
+    /// </summary>
+    /// <returns>The list of function calls.</returns>
+    [Experimental("SKEXP0001")]
+    public IEnumerable<FunctionCallContent> GetFunctionCalls()
+    {
+        return this.Items.OfType<FunctionCallContent>();
+    }
```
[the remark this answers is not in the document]
SergeyMenshykh: Yes, right, the function calls should not be more special than any other kernel content type. I wonder what our strategy for methods like this is. Should we keep the public API as slim as possible and demonstrate the possibility of having the methods as extension ones in examples?

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContent.cs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /// <summary>
+    /// Creates a new instance of the <see cref="FunctionCallContent"/> class.
+    /// </summary>
+    /// <param name="fullyQualifiedName">Fully-qualified name of the function.</param>
+    /// <param name="id">The function call ID.</param>
+    /// <param name="arguments">The function original arguments.</param>
+    /// <param name="functionNameSeparator">The function name separator.</param>
+    public static FunctionCallContent Create(string fullyQualifiedName, string? id, KernelArguments? arguments = null, string functionNameSeparator = "-")
```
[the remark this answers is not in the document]
SergeyMenshykh: It's a try to solve the ambiguity problem - `The call is ambiguous between the following methods or properties: 'FunctionCallContent.FunctionCallContent(string, string?, string?, KernelArguments?)' and 'FunctionCallContent.FunctionCallContent(string, string?, KernelArguments?, string)'`.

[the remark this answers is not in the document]
stephentoub: Having two different ways to construct one of these with two different places a developer has to look is not desirable. Why is this constructor even necessary? It's just splitting the fullyQualifiedName and then passing the two parts to one of the existing constructors? If so, either a) the caller can do that themselves, or b) if we really believe it's onerous to use Split or IndexOf, you could expose a Parse method somewhere that takes a fullyQualifiedName and a name separator and returns both parts.

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContent.cs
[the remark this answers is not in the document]
SergeyMenshykh: > Why is this desirable? Won't the caller also need to handle exceptional cases, and thus they'd need to also be capable of creating FunctionResultContent?

The method resolves the function in `Kernel.Plugins` collection and invokes it. It also preserves semantic of .NET function calling by propagating exceptions to callers.
image


> Where is this even being used, outside of tests?

The idea is that the callers will use it.

dotnet/src/InternalUtilities/src/System/IListExtensions.cs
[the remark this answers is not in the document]
stephentoub: The nice thing about Debug.Assert is it'll cause it to crash, whereas an exception could be caught and eaten.

If you want to include the ANEs, ok, it's just extra code that really should be dead.

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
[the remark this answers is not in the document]
SergeyMenshykh: Removed.

dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs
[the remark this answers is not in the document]
SergeyMenshykh: Right, the difference is not significant. I just considered function calling to be more fundamental than the other kernel content, which we agreed is not. So, the method has been removed.

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContent.cs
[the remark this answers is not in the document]
SergeyMenshykh: I have moved the parsing logic to the FullyQualifiedFunctionName internal utility class. Currently, the same logic exists in both OpenAI and Google connectors. Having it in a separate class is beneficial for all new connectors, as they can easily reuse it.

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionResultContent.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// The function name.
+    /// </summary>
+    public string FunctionName { get; }
+
+    /// <summary>
+    /// The result of the function call.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public object? Result { get; set; }
```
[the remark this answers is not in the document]
SergeyMenshykh: It's an oversight from my side. Fixed.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -21,6 +22,15 @@ public sealed class KernelArguments : IDictionary<string, object?>, IReadOnlyDic
     /// <summary>Dictionary of name/values for all the arguments in the instance.</summary>
     private readonly Dictionary<string, object?> _arguments;
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="KernelArguments"/> class with the specified AI execution settings.
+    /// </summary>
+    [JsonConstructor]
+    public KernelArguments()
```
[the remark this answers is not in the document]
SergeyMenshykh: Yep, fixed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: @stephentoub , I wonder if there's a more optimal way to avoid multiple enumeration of the `resultContents` variable of type IEnumerable<FunctionResultContent>? Or, is using ToArray acceptable/optimal enough in this case?
image

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: fixed

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -327,7 +327,7 @@ await foreach (Completions completions in response)
             // Or if we are auto-invoking but we somehow end up with other than 1 choice even though only 1 was requested, similarly bail.
             if (!autoInvoke || responseData.Choices.Count != 1)
             {
-                return responseData.Choices.Select(chatChoice => new OpenAIChatMessageContent(chatChoice.Message, this.DeploymentOrModelName, GetChatChoiceMetadata(responseData, chatChoice))).ToList();
+                return responseData.Choices.Select(chatChoice => this.GetChatMessage(chatChoice, responseData)).ToList();
```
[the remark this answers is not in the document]
SergeyMenshykh: These issues will be handled separately within the scope of different issues, which have not been created yet but will be soon.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
stephentoub: It'd be more efficient to do something like:
```C#
List<ChatRequestToolMessage>? toolMessages = null;
foreach (var resultContent in message.Items)
{
    if (resultContent is not FunctionResultContent frc) continue;

    toolMessages ??= new();
    ... // process resultContent
}

if (toolMessages is not null)
{
    return tooMessages;
}
...
```

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/ChatHistory.cs
SergeyMenshykh: removed

dotnet/src/InternalUtilities/src/System/IListExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>
+    /// Adds a range of elements from the specified <see cref="IEnumerable{T}"/> source to the target <see cref="IList{T}"/>.
+    /// </summary>
+    /// <typeparam name="T">The type of elements in the list.</typeparam>
+    /// <param name="target">The target <see cref="IList{T}"/> to add elements to.</param>
+    /// <param name="source">The source <see cref="IEnumerable{T}"/> containing elements to add to the target <see cref="IList{T}"/>.</param>
+    internal static void AddRange<T>(this IList<T> target, IEnumerable<T> source)
+    {
+        if (target is null)
```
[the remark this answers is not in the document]
SergeyMenshykh: The Debug.Assert(target is not null) is used instead. It'll be kept that way until there's agreement on the way we check the argument values.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,48 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
-
+                ChatMessageContent result = await chat.GetChatMessageContentAsync(chatHistory, settings, kernel);
                 if (result.Content is not null)
                 {
                     Write(result.Content);
                 }
 
-                List<ChatCompletionsFunctionToolCall> toolCalls = result.ToolCalls.OfType<ChatCompletionsFunctionToolCall>().ToList();
-                if (toolCalls.Count == 0)
+                IEnumerable<FunctionCallContent> functionCalls = result.Items.OfType<FunctionCallContent>(); // Getting list of function calls.
```
markwallace-microsoft, line 85: My suggestions for naming are:

1. `FunctionCallInvokeContent` & `FunctionCallResultContent`
1. `FunctionCallContent` & `FunctionCallResultContent`
1. `FunctionCallRequestContent` & `FunctionCallResultContent`

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,48 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
-                        JsonSerializer.Serialize((await function.InvokeAsync(kernel, arguments)).GetValue<object>()) :
-                        "Unable to find function. Please try again!";
-
-                    chatHistory.Add(new ChatMessageContent(
-                        AuthorRole.Tool,
-                        content,
-                        metadata: new Dictionary<string, object?>(1) { { OpenAIChatMessageContent.ToolIdProperty, toolCall.Id } }));
+                    try
+                    {
+                        FunctionResultContent functionResult = await functionCall.InvokeAsync(kernel); // Executing each function.
```
markwallace-microsoft, line 97: Using `FunctionCall` at the start of this class name feels confusing to me, if we use `FunctionCall` instead it makes things clearer.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,48 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
+
+                        chatHistory.AddMessage(AuthorRole.Tool, new ChatMessageContentItemCollection() { functionResult }); // Adding function result to chat history.
```
markwallace-microsoft, line 99: What do you think about this convenience method

`chatHistory.AddMessage(functionCallResult.GetChatMessage())`

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,48 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
+                    }
+                    catch (Exception ex)
+                    {
+                        chatHistory.AddMessage(AuthorRole.Tool, new ChatMessageContentItemCollection() { new FunctionResultContent(functionCall, ex) });
```
markwallace-microsoft, line 103: What do you think about this convenience method

`chatHistory.AddMessage(FunctionCallResult.CreateChatMessage(functionCallRequest, ex))`

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1069,6 +1116,41 @@ private static ChatRequestMessage GetRequestMessage(ChatResponseMessage message)
...
+            if (toolCall is ChatCompletionsFunctionToolCall functionToolCall)
+            {
+                KernelArguments? arguments = null;
+                try
+                {
+                    arguments = JsonSerializer.Deserialize<KernelArguments>(functionToolCall.Arguments);
+                }
+                catch (JsonException)
+                {
+                    // If the arguments are not valid JSON, we'll just leave them as null.
```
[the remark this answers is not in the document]
SergeyMenshykh: Yep, done

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContent.cs
markwallace-microsoft: My suggestions for naming are:

1. `FunctionCallInvokeContent` & `FunctionCallResultContent`
1. `FunctionCallContent` & `FunctionCallResultContent`
1. `FunctionCallRequestContent` & `FunctionCallResultContent`

My thinking is that the class names should start with `FunctionCall` to make it clear that are related to function calling and end with `Content` as other content classes do.

We "invoke" a function to get a "result" so using those words to identify the two content types feels clear and consistent to me. 

I've added in some other suggestions also.

markwallace-microsoft (commented):
Still have to review `ClientCore.cs` but here's my initial feedback

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,49 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
-
-                    chatHistory.Add(new ChatMessageContent(
-                        AuthorRole.Tool,
-                        content,
-                        metadata: new Dictionary<string, object?>(1) { { OpenAIChatMessageContent.ToolIdProperty, toolCall.Id } }));
+                    try
+                    {
+                        FunctionResultContent functionResult = await functionCall.InvokeAsync(kernel); // Executing each function. Can be done in parallel.
+
+                        chatHistory.AddMessage(AuthorRole.Tool, functionResult); // Adding function result to chat history.
```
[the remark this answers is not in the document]
SergeyMenshykh: done

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -426,21 +426,33 @@ await foreach (Completions completions in response)
...
                         Debug.Assert(result is null);
-                        logger.LogDebug("Failed to handle tool request ({ToolId}). {Error}", toolId, errorMessage);
+                        logger.LogDebug("Failed to handle tool request ({ToolId}). {Error}", toolCall.Id, errorMessage);
                     }
 
-                    // Add the tool response message to both the chat options and to the chat history.
+                    // Add the tool response message to the chat options
                     result ??= errorMessage ?? string.Empty;
-                    chatOptions.Messages.Add(new ChatRequestToolMessage(result, toolId));
-                    chat.AddMessage(AuthorRole.Tool, result, metadata: new Dictionary<string, object?> { { OpenAIChatMessageContent.ToolIdProperty, toolId } });
```
markwallace-microsoft, line 443: Is this change intentional and if yes what the the thinking behind this?

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
SergeyMenshykh: I'll choose option 3 - it's longer, but clearer.

SergeyMenshykh: Done

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
```diff
@@ -70,40 +68,49 @@ await foreach (var update in kernel.InvokePromptStreamingAsync("Given the curren
...
-
+                ChatMessageContent result = await chat.GetChatMessageContentAsync(chatHistory, settings, kernel);
                 if (result.Content is not null)
                 {
                     Write(result.Content);
                 }
 
-                List<ChatCompletionsFunctionToolCall> toolCalls = result.ToolCalls.OfType<ChatCompletionsFunctionToolCall>().ToList();
-                if (toolCalls.Count == 0)
+                IEnumerable<FunctionCallContent> functionCalls = result.GetFunctionCalls(); // Getting list of function calls.
```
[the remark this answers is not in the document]
SergeyMenshykh: Yep, done. New names - FunctionCallRequestContent & FunctionCallResultContent.

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContent.cs
SergeyMenshykh: Thanks, the classes have been renamed to FunctionCallRequestContent and FunctionCallResultContent, respectively.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
SergeyMenshykh: Fixed - the FunctionResultContent class has been renamed to FunctionCallResultContent.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: We need both function-calling models (not LLM models) - the current one that uses connector-specific classes and the new one that uses a connector-agnostic approach - to work alongside each other so that existing code continues to function when support for the new model is released. Eventually we may consider deprecating the connector-specific one.

Because of the two models exists alongside each other, there are cases when chat history has assistant messages with two duplicating pieces of information about the same call. 

The first one belongs to the current function calling model and represented either by the OpenAIChatMessageContent.ToolCall property or OpenAIChatMessageContent.FunctionToolCallsProperty metadata. 

The other one belongs to the new function calling model and represented by an item of the FunctionCallRequestContent type.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: Thanks, fixed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: Agreed offline to attempt removing code that reads tool calls from `OpenAIChatMessageContent.ToolCalls` and from `OpenAIChatMessageContent.Metadata` collections in favor of the new model. 

Additionally, we agreed to examine the impact of this change on consumer code that may deserialize a persisted chat history serialized using the current function-call model (not the one added by the PR) that relies on tool call information saved in chat message metadata with the `OpenAIChatMessageContent.FunctionToolCallsProperty` key.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1069,6 +1116,41 @@ private static ChatRequestMessage GetRequestMessage(ChatResponseMessage message)
...
+            {
+                KernelArguments? arguments = null;
+                try
+                {
+                    arguments = JsonSerializer.Deserialize<KernelArguments>(functionToolCall.Arguments);
+                }
+                catch (JsonException)
+                {
+                    // If the arguments are not valid JSON, we'll just leave them as null.
+                    // The original arguments and function tool call will be available via the 'InnerContent' property for the connector caller to access.
```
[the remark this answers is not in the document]
SergeyMenshykh: We need to return the exception to the caller, whether it's an LLM or a connector caller. The code is not there yet; I'm considering its design.  
   
This method iterates over LLM tool calls and converts each to a `FunctionCallRequestContent` item that contains all information about the function - plugin/function name, id, and arguments. As part of the conversion, function arguments are deserialized to kernel arguments. If the arguments' deserialization fails, the exception needs to be saved (to be replayed later when a caller invokes the function) in the `FunctionCallRequestContent` item, and the conversion should move to the next tool rather than stopping the whole operation just because one has invalid arguments.

An argument can be made in favor of stopping the entire operation if LLM provided invalid arguments for one or a few tools. As a counterargument, the caller may have code in place to fix those arguments, so the whole operation should not be stopped.  
   
If it's decided to proceed with the operation despite invalid arguments for a few functions, then it's up to the caller what to do with the exception. It can be converted to "Error: Failed to deserialize function arguments" and sent to LLM, or it can be re-thrown as any other .Net exception if the caller was not aware of it and does not have code for handling it.  
   
At the moment, there are a few ways the exception can be saved in the `FunctionCallRequestContent` item:  
1. In the `FunctionCallRequestContent.Exception` property, which is not there and needs to be added.  
2. In the `FunctionCallRequestContent.Metadata` with a well-known key.  
3. Passed to the `FunctionCallRequestContent` item constructor instead of `KernelArguments` and thrown when the function arguments are accessed by a caller.  
   
Additionally, the serialization and deserialization aspect of chat history, which the `FunctionCallRequestContent` item may be part of, should be taken into account. It's not easy/possible to deserialize a serialized exception back to the original type without knowing original exception type so an alternative solution needs to be found. InvalidKernelFunctionArgumentsException?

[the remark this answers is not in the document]
SergeyMenshykh: Decided to go with option 1 because it's a simple solution for the edge-case scenario.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: Results:
1. So, if the code that reads tool calls from `OpenAIChatMessageContent.ToolCalls` and from `OpenAIChatMessageContent.Metadata` collections is removed, no tool calls (tools LLM requested to invoke) are returned back to the LLM, and it responds with the error: _"400 (Bad Request) - Invalid parameter: messages with the role 'tool' must be a response to a preceding message with 'tool_calls'."_
2. If, in addition to not returning tool calls to LLM, the tool invocation results are not provided either, the LLM re-invokes the same tools/functions even though the chat history contains the final answer that was initially generated by LLM by calling the tools.

SergeyMenshykh: > This is O(N^2).
> Optimized and simplified.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
SergeyMenshykh: Yep, good idea - added.

dotnet/samples/KernelSyntaxExamples/Example59_OpenAIFunctionCalling.cs
SergeyMenshykh: Yep, good idea - added.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -968,39 +980,77 @@ private static ChatRequestMessage GetRequestMessage(ChatRole chatRole, string co
...
+            // Example: new ChatMessageContent(AuthorRole.Tool, items: new ChatMessageContentItemCollection { new FunctionCallResultContent(functionCall, result) })
+            List<ChatRequestToolMessage>? toolMessages = null;
+            foreach (var item in message.Items)
+            {
+                if (item is not FunctionCallResultContent resultContent)
+                {
+                    continue;
+                }
+
+                toolMessages ??= new();
```
dmytrostruk, line 1010:
```suggestion
                toolMessages ??= [];
```

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1069,6 +1137,51 @@ private static ChatRequestMessage GetRequestMessage(ChatResponseMessage message)
         throw new NotSupportedException($"Role {message.Role} is not supported.");
     }
 
+    private OpenAIChatMessageContent GetChatMessage(ChatChoice chatChoice, ChatCompletions responseData)
+    {
+        var message = new OpenAIChatMessageContent(chatChoice.Message, this.DeploymentOrModelName, GetChatChoiceMetadata(responseData, chatChoice));
+
+        foreach (var toolCall in chatChoice.Message.ToolCalls)
```
dmytrostruk, line 1144: Just double-checking if `Message` or `ToolCalls` can be null here. They are defined as non-nullable, but sometimes such properties can still hold `null` values. We had similar case with `ChatCompletions.Usage` property, which was fixed in this PR: https://github.com/microsoft/semantic-kernel/pull/4956

dotnet/src/Connectors/Connectors.UnitTests/OpenAI/ChatCompletion/AzureOpenAIChatCompletionServiceTests.cs
```diff
@@ -700,6 +700,212 @@ public async Task GetChatMessageContentsWithChatMessageContentItemCollectionAndS
         Assert.Equal("image_url", contentItems[1].GetProperty("type").GetString());
     }
 
+    [Fact]
+    public async Task FunctionCallsShouldBePropagatedToCallersViaChatMessageItemsOfTypeFunctionCallContentAsync()
```
dmytrostruk, line 704: `FunctionCallRequestContent`?
```suggestion
    public async Task FunctionCallsShouldBePropagatedToCallersViaChatMessageItemsOfTypeFunctionCallRequestContentAsync()
```

dotnet/src/Connectors/Connectors.UnitTests/OpenAI/ChatCompletion/OpenAIChatCompletionServiceTests.cs
```diff
@@ -323,6 +323,212 @@ public async Task GetChatMessageContentsWithChatMessageContentItemCollectionAndS
         Assert.Equal("image_url", contentItems[1].GetProperty("type").GetString());
     }
 
+    [Fact]
+    public async Task FunctionCallsShouldBePropagatedToCallersViaChatMessageItemsOfTypeFunctionCallContentAsync()
```
dmytrostruk, line 327: `FunctionCallRequestContent`? 
```suggestion
    public async Task FunctionCallsShouldBePropagatedToCallersViaChatMessageItemsOfTypeFunctionCallRequestContentAsync()
```

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallRequestContent.cs
```diff
@@ -0,0 +1,100 @@
...
+        this.Arguments = arguments;
+    }
+
+    /// <summary>
+    /// Invokes the function represented by the function call content type.
+    /// </summary>
+    /// <param name="kernel">The <see cref="Kernel"/> containing services, plugins, and other state for use throughout the operation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The result of the function's execution.</returns>
+    public async Task<FunctionCallResultContent> InvokeAsync(Kernel kernel, CancellationToken cancellationToken = default)
```
dmytrostruk, line 72: Having function invocation logic inside `Content` class feels a little bit complicated to me. It would be better if we keep `Content` classes for data representation only. 

Is this something that is going to be used in Connectors in the future or there will be another abstraction/implementation for it? If I want to add new type of filter for function calling (i.e. `IFunctionCallFilter`) should I invoke it here? Just want to understand future plans for this method.

dotnet/src/SemanticKernel.UnitTests/Contents/FunctionCallResultContentTests.cs
```diff
@@ -0,0 +1,104 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json;
+using Microsoft.SemanticKernel;
+using Xunit;
+
+namespace SemanticKernel.UnitTests.Contents;
+public class FunctionCallResultContentTests
```
dmytrostruk, line 8:
```suggestion

public class FunctionCallResultContentTests
```

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallResultContent.cs
```diff
@@ -0,0 +1,87 @@
...
+        this(functionCallContent, result.Value)
+    {
+        this.InnerContent = result;
+    }
+
+    /// <summary>
+    /// Creates <see cref="ChatMessageContent"/> and adds the current instance of the class to the <see cref="ChatMessageContent.Items"/> collection.
+    /// </summary>
+    /// <returns>The <see cref="ChatMessageContent"/> instance.</returns>
+    public ChatMessageContent ToChatMessage()
```
dmytrostruk, line 83: `ToChatMessageContent`?

dotnet/src/SemanticKernel.UnitTests/Utilities/FunctionNameTests.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.SemanticKernel;
+using Xunit;
+
+namespace SemanticKernel.UnitTests.Utilities;
+public class FunctionNameTests
```
dmytrostruk, line 7:
```suggestion

public class FunctionNameTests
```

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
SergeyMenshykh: Good point @dmytrostruk. I quickly scanned the code and found a few places where the collection is used as is, relying solely on the fact that it is declared as non-nullable. The code has been there since the December SK release. Example - https://github.com/microsoft/semantic-kernel/blob/a27a46dc4af47b33df933126e1d5938e919af04c/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIChatMessageContent.cs#L29C156-L29C177

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallRequestContent.cs
SergeyMenshykh: Discussed offline, and agreed to find a proper place for the method when there's clear understanding of function-calling filters for manual function calling scenario.

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallResultContent.cs
SergeyMenshykh: I wanted to keep it short and to omit the "unnecessary" content suffix. Otherwise, I would have to rename the convenience method as well - `FunctionCallRequestContent.GetFunctionCalls` to `FunctionCallRequestContent.GetFunctionCallContents` which would be made it unnecessarily verbose. In any case, I don't mind renaming if really necessary.

dmytrostruk: > `FunctionCallRequestContent.GetFunctionCalls` to `FunctionCallRequestContent.GetFunctionCallContents` which would be made it unnecessarily verbose

I think it's a pattern in other abstractions as well, `IChatCompletionService` as example:
https://github.com/microsoft/semantic-kernel/blob/65275737d66d5c5431656093af541b8667a59720/dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/IChatCompletionService.cs#L27

But since it's `Content` class, maybe we can follow different pattern here. I don't think it's really necessary, so feel free to proceed.

markwallace-microsoft (approved):

dmytrostruk (approved):

markwallace-microsoft (approved):

PR #5829, opened 2024-04-10
Java: Add OpenApi plugin importer and upgrade tool calling to cover more usecases

johnoliver:
### Contribution Checklist

### DIFF
--- a/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/KernelHooks.java
+++ b/java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/KernelHooks.java
@@ -5,6 +5,7 @@ import com.microsoft.semantickernel.hooks.KernelHook.FunctionInvokedHook;
 import com.microsoft.semantickernel.hooks.KernelHook.FunctionInvokingHook;
 import com.microsoft.semantickernel.hooks.KernelHook.PostChatCompletionHook;
 import com.microsoft.semantickernel.hooks.KernelHook.PreChatCompletionHook;
+import com.microsoft.semantickernel.hooks.KernelHook.PreToolCallHook;
 import com.microsoft.semantickernel.hooks.KernelHook.PromptRenderedHook;
 import com.microsoft.semantickernel.hooks.KernelHook.PromptRenderingHook;
 import java.util.Collections;
@@ -100,6 +101,17 @@ public class KernelHooks {
     public String addPreChatCompletionHook(
         Function<PreChatCompletionEvent, PreChatCompletionEvent> function) {
         return addHook((PreChatCompletionHook) function::apply);
+    }
+
+    /**
+     * Add a {@link PreToolCallHook} to the collection of hooks.
+     *
+     * @param function the function to add
+     * @return the key of the hook in the collection
+     */
+    public String addPreToolCallHook(
+        Function<PreToolCallEvent, PreToolCallEvent> function) {
+        return addHook((PreToolCallHook) function::apply);
     }
 
     /**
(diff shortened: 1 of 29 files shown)


java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/KernelHooks.java
```diff
@@ -102,6 +103,17 @@ public String addPreChatCompletionHook(
...
+     * Add a {@link PreToolCallHook} to the collection of hooks.
+     *
+     * @param function the function to add
+     * @return the key of the hook in the collection
+     */
+    public String addPreToolCallHook(
+        Function<PreToolCallEvent, PreToolCallEvent> function) {
+        return addHook((PreToolCallHook) function::apply);
+    }
+
```
dsgrieve, line 116: No PostToolCallHook? Maybe I have a function that computes degrees Celsius only and I want to convert to Farenheit.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/plugin/KernelPluginFactory.java
```diff
@@ -89,7 +89,34 @@ public class KernelPluginFactory {
...
+    public static KernelPlugin createFromObject(Class<?> clazz, Object target, String pluginName) {
+        if (!clazz.isInstance(target)) {
+            throw new SKException("Target object is not an instance of the provided class");
+        }
+
+        if (clazz.getName().endsWith("_ClientProxy")) {
+            LOGGER.warn(
+                "The class " + clazz.getName() + " is likely a proxy from a DI framework, "
+                    + "this likely will prevent Semantic Kernel from detecting methods. "
+                    + "Try using: KernelPluginFactory.createFromObject(Class<?> clazz, Object target, String pluginName).");
```
dsgrieve, line 116: Aren't you telling me to use the same function I just called?

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/plugin/KernelPluginFactory.java
```diff
@@ -89,7 +89,34 @@ public class KernelPluginFactory {
...
+     *                   value is {@code null}, a plugin name is derived from the type of the
+     *                   target.
+     * @return The new plugin.
+     */
+    public static KernelPlugin createFromObject(Class<?> clazz, Object target, String pluginName) {
+        if (!clazz.isInstance(target)) {
+            throw new SKException("Target object is not an instance of the provided class");
+        }
+
+        if (clazz.getName().endsWith("_ClientProxy")) {
```
dsgrieve, line 112: Isn't this dependent on the DI framework? Could this work as, if reflection fails then warn that a probable cause is proxy? Another possible reason would be a module that doesn't export or open.

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/plugin/KernelPluginFactory.java
johnoliver: Not necessarily, they may have got to this position through the invocation without the Class argument

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/KernelHooks.java
johnoliver: At the moment I did not have a usecase for PostToolCall hook, @milderhc is looking at the filtering requirement, I would leave it to him to see if it is needed for the filtering work

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/plugin/KernelPluginFactory.java
johnoliver: It is framework specific, I will also add a warning in the case of no methods found

java/semantickernel-api/src/main/java/com/microsoft/semantickernel/hooks/KernelHooks.java
dsgrieve: Odd to have pre/post for other events and not this one.

PR #5830, opened 2024-04-10
Python: Allow args to pass to func calling stepwise planner

### LINKED ISSUE #5824: Python - Ability to inject chat-history for plan execution
We've had a customer independently identify this. _ and _ as it turns out, we had a community contribution in the dotnet space this week that addresses this:

https://github.com/microsoft/semantic-kernel/pull/5718

By adding a `ChatHistory` argument to the `ExecuteAsync` method:

image

For python, I wonder if this would be accomodated here:

https://github.com/microsoft/semantic-kernel/blob/main/python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py#L88

moonbox3:
### Motivation and Context

A recent feature was added to dotnet to allow for a chat_history object to be passed into invoke. This only allows the user to affect the `chatHistoryForSteps`, but doesn't account for generating a plan with provided context (via the chat history). 

### Description

This PR allows the user to pass in optional kernel arguments or kwargs (that get turned into kernel arguments). If the user then overrides the prompt template/step prompt.txt, they can utilize any new arguments that are supplied. 
- Closes #5824 

### Contribution Checklist

### DIFF
--- a/python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py
+++ b/python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py
@@ -1,10 +1,12 @@
 # Copyright (c) Microsoft. All rights reserved.
 
+from __future__ import annotations
+
 import asyncio
 import logging
 import os
 from copy import copy
-from typing import Optional
+from typing import Any, Optional
 
 import yaml
 
@@ -89,6 +91,8 @@ class FunctionCallingStepwisePlanner(KernelBaseModel):
         self,
         kernel: Kernel,
         question: str,
+        arguments: KernelArguments | None = None,
+        **kwargs: Any,
     ) -> FunctionCallingStepwisePlannerResult:
         """
         Execute the function calling stepwise planner
@@ -96,6 +100,8 @@ class FunctionCallingStepwisePlanner(KernelBaseModel):
         Args:
             kernel: The kernel instance
             question: The input question
+            arguments: (optional) The kernel arguments
+            kwargs: (optional) Additional keyword arguments
 
         Returns:
             FunctionCallingStepwisePlannerResult: The result of the function calling stepwise planner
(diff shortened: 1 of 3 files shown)


juliomenendez (approved):

python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py
```diff
@@ -89,13 +91,17 @@ async def invoke(
...
+        arguments: KernelArguments | None = None,
+        **kwargs: Any,
     ) -> FunctionCallingStepwisePlannerResult:
         """
         Execute the function calling stepwise planner
 
         Args:
             kernel: The kernel instance
             question: The input question
+            arguments: (optional) The kernel arguments
```
eavanvalkenburg, line 103: Nit: No types in the docstring?

python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py
```diff
@@ -106,6 +112,9 @@ async def invoke(
         if not question:
             raise PlannerInvalidConfigurationError("Input question cannot be empty")
 
+        if not arguments:
```
eavanvalkenburg, line 115: Check `if arguments is not None` instead, this will return false for a Kernel args object without values but with execution settings, I've been changing that in some other places need to check all...

eavanvalkenburg (approved):
One small thing!

python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py
moonbox3: I actually want the check as `if argument is None`, then I want to create the KernelArguments with any **kwargs. By using `if arguments is not None` then we will skip the creation of the Kernel args -- other methods except them, even if they're just empty `{}`.

python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py
moonbox3: Will fix in the next iteration.

python/semantic_kernel/planners/function_calling_stepwise_planner/function_calling_stepwise_planner.py
moonbox3: I can make any adjustment in a next iteration, as needed.

PR #5831, opened 2024-04-10
.Net - Agents Fix: Chat Concurrency (Step #ANY)

crickman:
### Motivation and Context

Minimize locking / state management for `AgentChat` and `BroadcastQueue`.

### Description

On re-evaluation w/ Sergey, we wanted to consolidate `BroadcastQueue` state and determine if this had an impact on locking semantics.  While this did show some improvement, I traced the concurrency contract back into `AgentChat` and was able to establish a flow that eliminated the need for `BroadcastQueue._stateLock`.  There exists zero contention for `BroadcastQueue._queues.`

I've ran ran all tests, including my local concurrency performance harness that creates 3 agents with 3 separate channels and proceeds with over 100 turns while validating turn order.  For this test, I ran the harness using Open AI Assistant Agents for maximum latency / queueing.

Test execution quantitatively demonstrates lighter locking patterns.

### Contribution Checklist


dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -14,14 +13,31 @@ namespace Microsoft.SemanticKernel.Agents;
...
+    private readonly Dictionary<Agent, string> _channelMap; // Map agent to its channel-hash: one entry per agent.
 
     private int _isActive;
 
+    /// <summary>
+    /// Indicates if a chat operation is active.  This includes:
+    /// - <see cref="AddChatMessage"/>
+    /// - <see cref="AddChatMessages"/>
+    /// - <see cref="GetChatMessagesAsync"/>
+    /// - <see cref="InvokeAgentAsync"/>
```
markwallace-microsoft, line 37: Not getting the significance of all the class references, can you expand the comment a bit?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -34,26 +50,51 @@ public abstract class AgentChat
     /// <param name="agent">An optional agent, if requesting an agent history.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The message history</returns>
-    public IAsyncEnumerable<ChatMessageContent> GetChatMessagesAsync(Agent? agent = null, CancellationToken cancellationToken = default)
+    public async IAsyncEnumerable<ChatMessageContent> GetChatMessagesAsync(
+        Agent? agent = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
-        if (agent == null)
+        this.ThrowIfActive();
```
markwallace-microsoft, line 57: This behaviour should be explained in the comments. I'm curious why we are doing this?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -34,26 +50,51 @@ public abstract class AgentChat
...
-        return channel.GetHistoryAsync(cancellationToken);
     }
 
     /// <summary>
-    /// Append messages to the conversation.
+    /// Append a message to the conversation.  Adding a message while an agent
+    /// is active is not allowed.
     /// </summary>
-    /// <param name="message">Set of non-system messages with which to seed the conversation.</param>
+    /// <param name="message">A of non-system messages with which to append to the conversation.</param>
```
markwallace-microsoft, line 97: Please update the comment, looks like a bad edit

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -134,47 +186,73 @@ await foreach (var message in channel.InvokeAsync(agent, cancellationToken).Conf
...
+    /// </summary>
+    private void ClearActivitySignal()
+    {
+        Interlocked.Exchange(ref this._isActive, 0);
     }
 
-    private string GetAgentHash(Agent agent)
+    /// <summary>
+    /// Test to ensure chat is not concurrently active and throw exception if it is.
+    /// If not, activity is signaled.
```
markwallace-microsoft, line 221: This is a non obvious concept, can you expand the comment?

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Intent was to call out the public methods on this class.  I agree (with sync dialog) that being this explicit is just a point of maintenance.  Updated as suggested.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Was supposed to be the _singular_ (vs _plural_) version...fixed.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Done

crickman:
> Aside: None of these comments went to `pending`...github, I'm watchin' you!

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Updated

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -134,47 +193,81 @@ await foreach (var message in channel.InvokeAsync(agent, cancellationToken).Conf
...
+    /// </summary>
+    /// <remarks>
+    /// Rather than allowing concurrent invocation to result in undefined behavior / failure,
+    /// it is preferred to fail-fast in order to avoid side-effects / state mutation.
+    /// The activity signal is used to manage ability and visibility for taking actions based
+    /// on conversation history.
+    /// </remarks>
+    private void ThrowIfActive()
```
SergeyMenshykh, line 237: The name implies checking the state rather than setting it. I would propose to name it in such a way that it's clear it not only checks and throws, but also sets the signal - SetActivityOrThrow?

dotnet/src/Agents/Abstractions/Internal/BroadcastQueue.cs
```diff
@@ -34,53 +34,48 @@ internal sealed class BroadcastQueue
...
-            foreach (var channel in channels)
+            if (!this._queues.TryGetValue(channelRef.Hash, out var queueRef))
             {
-                if (!this._queues.TryGetValue(channel.Hash, out var queueRef))
-                {
-                    queueRef = new();
-                    this._queues.Add(channel.Hash, queueRef);
-                }
+                queueRef = new();
+                this._queues.Add(channelRef.Hash, queueRef);
```
SergeyMenshykh, line 47: Potentially, there could be a concurrent access to the `this._queues` dictionary.

crickman: Yes...but not in the established and closed invocation relationship with `AgentChat`...it is due to this special relationship that allowed `_stateLock` to be removed.  If this were not the case, we'd need to lock on the outside of the for-loop...concurrent dictionary alone wouldn't be sufficient.

This understanding should align with our sync discussion yesterday when we were walking through this PR.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Updated

SergeyMenshykh (approved):

markwallace-microsoft (approved):

PR #5863, opened 2024-04-14
.Net: Utbose/open api/fix/default value for params

utbose:
### Motivation and Context

  This change will help assign default values that is specified in openAPI schema to parameter objects if no arguments are provided from the caller. Without this change we wont be able to support urls which has default query parameters.
Currently if the caller does not set any arguments for the query parameters, the RestApiOperation.BuildQueryString() does not assign any value to the query parameter objects even though they have default value specified in the openAPI spec.
  4. What scenario does it contribute to?
  Executing REST API using openAPI spec.
  5. If it fixes an open issue, please link to the issue here.

### Description

Making changes in code to ensure if the query parameters cannot be populated from the caller provided arguments, then just use default values (if specified in openAPI spec) to populate them instead of dropping of the query parameter all together.

Created a new issue: https://github.com/microsoft/semantic-kernel/issues/5864

### Contribution Checklist

- [*] The code builds clean without any errors or warnings
- [*] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [*] All unit tests pass, and I have added new tests where possible
- [*] I didn't break anyone :smile:

### DIFF
--- a/dotnet/src/Functions/Functions.OpenApi/Model/RestApiOperation.cs
+++ b/dotnet/src/Functions/Functions.OpenApi/Model/RestApiOperation.cs
@@ -172,7 +172,7 @@ public sealed class RestApiOperation
                     throw new KernelException($"No argument or value is provided for the '{parameter.Name}' required parameter of the operation - '{this.Id}'.");
                 }
 
-                // Skipping not required parameter if no argument provided for it.
+                // Skipping not required parameter if no argument provided for it unless it has a default value.
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


utbose:
@microsoft-github-policy-service agree [company="{your company}"]

utbose:

utbose:
@microsoft-github-policy-service agree company="Microsoft"

SergeyMenshykh:
Closing the PR because a workable solution for using and accessing default values of parameters has been provided. See the https://github.com/microsoft/semantic-kernel/issues/5864 issue for more details.

Issue #5470, opened 2024-03-14
.Net: For Automatic Function Calling, there should be filters to control the loop behavior


matthewbolanos:
Today, we have filters for prompt rendering and function invoking. What's missing is the ability to control the behavior while inside of the automatic function call loop. While in the loop, the user should be able to break out of the loop in the following scenarios:

- **End before the LLM generates a final message** – Many customers are wanting the LLM to return back the JSON of a function call for an adapative card or verbatims from a RAG search result. Unfortunately, this is not possible today because the last step is to always have the LLM generate a message that is returned back to the user. With filters, the developer would be able to break out of the loop before this final step and return the JSON payload.
- **End after N-number of loops** – To avoid an infinite loop, the developer may want to implement a hard limit to the number of function calls that are invoked while in the automatic function call loop. With filters, the developer should be able to count the number of invocations and cancel after an arbitrary limit.
- **Telemetry** – The developer may want to log metadata about the loop (number of turns, the functions called via parallel function calling, etc.)

Below is a flow diagram of where the existing filters are and where the new filters need to exist for function calling

<img width="1364" alt="image" src="https://github.com/microsoft/semantic-kernel/assets/1409190/86354539-bc26-491d-a80d-9dee9e84e93e">

moonbox3:
@matthewbolanos, for the second point: in both dotnet and Python (not sure about Java) there is a `max auto invoke attempts` option that is configured by the user. It defaults to 10 or 15, I believe, but this will make sure we're not stuck in an infinite loop. Isn't this achieving the same behavior as the proposed filters that you mention? If not, what is the delta? Thanks!

lavinir:
@matthewbolanos , today if a Function throws an exception, it is caught and swallowed, the exception message is extracted and added to the chat history as if it a 'regular' function response.
It would make sense if an exception was thrown, the function filters would receive the exception as well as the function response so that they are aware that an exception was thrown.

What do you think ?

dmytrostruk:
> @matthewbolanos , today if a Function throws an exception, it is caught and swallowed, the exception message is extracted and added to the chat history as if it a 'regular' function response. It would make sense if an exception was thrown, the function filters would receive the exception as well as the function response so that they are aware that an exception was thrown.
> 
> What do you think ?

@lavinir Thanks for feedback! I'm working on it at the moment in scope of this task: https://github.com/microsoft/semantic-kernel/issues/4896.
I expect to create a PR for it in the next couple of days.

[...]

PR #5900, opened 2024-04-17
.Net: [Feature Branch] Function Call Filters

dmytrostruk:
### Motivation and Context

Related: https://github.com/microsoft/semantic-kernel/issues/5470

This PR contains changes for new type of filter - function call filters. They work similarly to function filters, but they have more information in a context (specific to function calling, like iteration number), and more control:
- The ability to cancel function call loop (inner loop) - when single request to LLM returns multiple function calls, and we want to stop further execution.
- The ability to cancel request loop (outer loop) - when all function are called, the last step is always a request to LLM to generate final message. This feature allows to avoid last request to LLM and return result of last executed function instead.
- The ability to skip function execution - in this case function call process will proceed to the next function in pipeline.

Usage:
```csharp
public class AutoFunctionInvocationFilter(ILogger logger) : IAutoFunctionInvocationFilter
{
    private readonly ILogger _logger = logger;

    public async Task OnAutoFunctionInvocationAsync(AutoFunctionInvocationContext context, Func<AutoFunctionInvocationContext, Task> next)
    {
        // Example: get function information
        var functionName = context.Function.Name;
    
        // Example: get chat history
        var chatHistory = context.ChatHistory;
    
        // Example: get request sequence number
        this._logger.LogDebug($"Request sequence number: {context.RequestSequenceNumber}");
    
        // Example: get function sequence number
        this._logger.LogDebug($"Function sequence number: {context.FunctionSequenceNumber}");
    
        // Calling next filter in pipeline or function itself.
        // By skipping this call, next filters and function won't be invoked, and function call loop will proceed to the next function.
        await next(context);
    
        // Example: get function result
        var result = context.Result;
    
        // Example: override function result value
        context.Result = new FunctionResult(context.Result, "Result from auto function invocation filter");
    
        // Example: cancel function invocation
        context.Cancel = true;
    }
}
```

### Contribution Checklist


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -403,14 +405,32 @@ await foreach (Completions completions in response.ConfigureAwait(false))
                 }
 
                 // Now, invoke the function, and add the resulting tool call message to the chat options.
+                FunctionResult functionResult = new(function) { Culture = kernel.Culture };
```
SergeyMenshykh, line 408: It might be convenient to have an instance of Kernel in the filters.

dotnet/samples/KernelSyntaxExamples/Example76_Filters.cs
```diff
@@ -212,6 +204,58 @@ public async Task OnFunctionInvocationAsync(FunctionInvocationContext context, F
...
+            context.Result = new FunctionResult(context.Result, "Result from function call filter");
+
+            // Example: stop further function calling. but proceed with request iteration
+            context.Action = FunctionCallAction.StopFunctionCallIteration;
+
+            // Example: call remaining functions, but stop request iteration
+            context.Action = FunctionCallAction.StopRequestIteration;
+
+            // Example: stop function calling and request iterations, return immediately
+            context.Action = FunctionCallAction.StopRequestIteration | FunctionCallAction.StopFunctionCallIteration;
```
markwallace-microsoft, line 255: Should we define something like `FunctionCallAction.ReturnResult` to aid discoverability?

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/IFunctionCallFilter.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.SemanticKernel;
+
+#pragma warning disable CA1716 // Identifiers should not match keywords (Func<FunctionCallInvocationContext, Task> next)
+
+/// <summary>
+/// Interface for filtering actions during function calling invocation.
+/// </summary>
+[Experimental("SKEXP0001")]
+public interface IFunctionCallFilter
```
markwallace-microsoft, line 15: Wondering if the interface name and description are clear enough.

Based on naming it's not obvious what the difference between `IFunctionFilter` and `IFunctionCallFilter` is.

We are also using the term `ToolCallBehavior`.

Some thoughts:

- `IToolCallFilter`
- `IToolFunctionFilter`
- `IToolFunctionCallFilter`

Let's discuss with @matthewbolanos

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/FunctionCallAction.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Request iteration will send function results achieved so far, so LLM can decide what to do next.
+    /// </summary>
+    StopFunctionCallIteration = 1,
+
+    /// <summary>
+    /// This action will stop request iteration.
+    /// No more requests will be performed to LLM after last function call.
+    /// </summary>
+    StopRequestIteration = 1 << 1
+}
```
markwallace-microsoft, line 29: Can you expand all of the comments in this enum so it's very clear when each option would be used, maybe include an example

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/FunctionCallInvocationContext.cs
```diff
@@ -0,0 +1,65 @@
...
+    public FunctionCallInvocationContext(
+        KernelFunction function,
+        FunctionResult result)
+    {
+        Verify.NotNull(function);
+        Verify.NotNull(result);
+
+        this.Function = function;
+        this.Result = result;
+    }
```
markwallace-microsoft, line 27: Include the `Kernel` as a property

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -424,10 +444,25 @@ await foreach (Completions completions in response.ConfigureAwait(false))
...
+                // Apply any changes from the function call filters context to final result.
+                functionResult = new FunctionResult(invocationContext.Result);
+
+                object functionResultValue = functionResult.GetValue<object>() ?? string.Empty;
+                var stringResult = ProcessFunctionResult(functionResultValue, chatExecutionSettings.ToolCallBehavior);
 
                 AddResponseMessage(chatOptions, chat, stringResult, errorMessage: null, toolCall.Id, this.Logger);
 
+                // If filter requested to stop function call iteration, breaking function calling loop.
+                if (invocationContext.Action.HasFlag(FunctionCallAction.StopFunctionCallIteration))
```
markwallace-microsoft, line 456: The iterator variable is called `toolCallIteration` but the flag we are using is called `StopFunctionCallIteration`. Should we align our terminology?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -475,14 +510,25 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
...
             {
                 autoInvoke = false;
                 if (this.Logger.IsEnabled(LogLevel.Debug))
                 {
                     this.Logger.LogDebug("Maximum auto-invoke ({MaximumAutoInvoke}) reached.", chatExecutionSettings.ToolCallBehavior!.MaximumAutoInvokeAttempts);
                 }
             }
+
+            // If filter requested to stop request iteration, returning latest result from LLM.
+            if (invocationContext is not null && invocationContext.Action.HasFlag(FunctionCallAction.StopRequestIteration))
```
markwallace-microsoft, line 523: For a readability perspective I think it would be easier to understand if the flag indicated what we wanted to happen rather than what we want to stop e.g. `if (...HasFlag(ReturnToolCallResult))` is easier to understand than `if (...HasFlag(StopRequestIteration))`

Krzysztof318:
It would be nice if you add also support for this in other connectors.

dmytrostruk:
> It would be nice if you add also support for this in other connectors.

@Krzysztof318 Thanks for feedback. The plan is to add this functionality to OpenAI connector first. In parallel, we are working on new function calling abstractions, and most probably all function calling logic will be extracted from connectors and moved to common place. Then, we will move filtering logic from OpenAI connector to this common place, which will be used by other connectors that support function calling. In this way, we will enable filtering for all connectors that support function calling feature.

Krzysztof318:
@dmytrostruk oh okay, I totally forgot work on function calling abstraction is in progress 😶‍🌫️

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
dmytrostruk: Thanks, added.

dotnet/samples/KernelSyntaxExamples/Example76_Filters.cs
dmytrostruk: I simplified the implementation and removed Action enumeration completely in favour of simple boolean `Cancel` property.

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/IFunctionCallFilter.cs
dmytrostruk: Agreed on following renaming:
- `IPromptFilter` -> `IPromptRenderFilter`
- `IFunctionFilter` -> `IFunctionInvocationFilter`
- `IFunctionCallFilter` -> `IAutoFunctionInvocationFilter`

Renaming for function and prompt filters will be completed in following PR.

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/FunctionCallAction.cs
dmytrostruk: I simplified the implementation and removed Action enumeration completely in favour of simple boolean `Cancel` property.

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/FunctionCallInvocationContext.cs
dmytrostruk: Thanks, added.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
dmytrostruk: There are no more flags in latest version, so we should be good. Regarding iterators in context, I renamed them to `RequestSequenceNumber` and `FunctionSequenceNumber`, happy to discuss this naming if needed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
dmytrostruk: I simplified the implementation and removed Action enumeration completely in favour of simple boolean `Cancel` property.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -451,7 +474,7 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
             chatOptions.ToolChoice = ChatCompletionsToolChoice.None;
             chatOptions.Tools.Clear();
 
-            if (iteration >= chatExecutionSettings.ToolCallBehavior!.MaximumUseAttempts)
+            if (requestIteration >= chatExecutionSettings.ToolCallBehavior!.MaximumUseAttempts)
```
RogerBarreto, line 477: nit.

Reading at first this variable didn't look like a `int`

```suggestion
            if (requestIterationCount >= chatExecutionSettings.ToolCallBehavior!.MaximumUseAttempts)
```

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -152,6 +140,15 @@ public sealed class Kernel
         Interlocked.CompareExchange(ref this._promptFilters, [], null) ??
         this._promptFilters;
 
+    /// <summary>
+    /// Gets the collection of auto function invocation filters available through the kernel.
+    /// </summary>
+    [Experimental("SKEXP0001")]
+    public IList<IAutoFunctionInvocationFilter> AutoFunctionInvocationFilters =>
```
RogerBarreto, line 147: Although `Auto` can be understood between us, this can be misunderstood between what is a Kernel Function Invocation and a `Function Calling Request Invocation` from the LLM, adding calling gives a bit more clarity to this.

```suggestion
    public IList<IAutoFunctionInvocationFilter> AutoFunctionCallFilters =>
```

dotnet/src/SemanticKernel.Abstractions/Filters/AutoFunctionInvocation/IAutoFunctionInvocationFilter.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel;
+
+#pragma warning disable CA1716 // Identifiers should not match keywords (Func<FunctionCallInvocationContext, Task> next)
```
RogerBarreto, line 9: Not sure why are we moving `Calling` away in favor of `Invocation` as this clashes with the already know concept `KernelFunctions.InvokeAsync`.

```suggestion
#pragma warning disable CA1716 // Identifiers should not match keywords (Func<AutoFunctionInvocationContext, Task> next)
```

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/IFunctionCallFilter.cs
RogerBarreto: @dmytrostruk I may have not aware of the agreement, but if thats the case, ignore my comments. I don't like the `Invocation` mixed with the other. Those are different things and can be very confusing as Auto Function Invocation is as concept that only how already knows Semantic Kernel understand, `Function Calling` although is something that community has an understanding.

`AutoFunctionCallRequestFilter` would be my suggestion for this.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
dmytrostruk: This variable is used in loop, not sure if `Count` would be a good name here:
https://github.com/microsoft/semantic-kernel/blob/40b2bba4ce9fc240aa5819a3a556282362ef5eea/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs#L316

How about `requestIndex`?

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: I would like to keep name of interface and kernel property aligned. So this change will require to rename interface as well. Interface name is based on `ToolCallBehavior.AutoInvokeKernelFunctions`. `Auto` in this case should point the difference between kernel function invocation and automatic function invocation. But I'm happy to discuss it further. @matthewbolanos could provide input here as well.

dotnet/src/SemanticKernel.Abstractions/Filters/AutoFunctionInvocation/IAutoFunctionInvocationFilter.cs
dmytrostruk: Based on our discussion with @matthewbolanos , there are 3 phases: Call -> Invoke -> Result. `Call` in this case means request to call something, but it's just a request, without actual action. `Invoke` it's when we are performing some action (invoking a function), and this is exactly when filter will be triggered, so that's why filter and context model naming was updated.

dotnet/src/SemanticKernel.Abstractions/Filters/FunctionCall/IFunctionCallFilter.cs
dmytrostruk: @RogerBarreto based on discussion with @matthewbolanos we decided to proceed with this naming. I think it will work fine with `ToolCallBehavior.AutoInvokeKernelFunctions` setting.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
dmytrostruk: Renamed to `requestIndex`.

Issue #5724, opened 2024-04-01
.Net: Support for consequential attribute and HTTP method attributes for KernelFunction


matthewbolanos:
For #5506 and #5694

## Context

Today, `KernelFunction` has a metadata property that includes the following information:

```csharp
public sealed class KernelFunctionMetadata {	
        public KernelFunctionMetadata(string name);
        public KernelFunctionMetadata(KernelFunctionMetadata metadata);
        public string Description { get; init; }
        public string Name { get; init; }
        public IReadOnlyList<KernelParameterMetadata> Parameters { get; init; }
        public string? PluginName { get; init; }
        public KernelReturnParameterMetadata ReturnParameter { get; init; }
}
```

## User story

This metadata doesn't have a way to add additional information that can be used elsewhere in the system to determine if a function is "consequential" (i.e., requires customer approval). This metadata would be most useful within the `FunctionInvokingEventArgs` so a developer can request human approval for functions.

For example, knowing the HTTP method of an OpenAPI imported function could be helpful to provide approval functionality (e.g., GET requests wouldn't require approval, but PUT and POST could).

## Options

There are a couple options for handling this. Several of these options can be used together to address short comings.

### Option 1: add a boolean value for `IsConsequential` in `KernelFunctionMetadata`

With a `bool IsConsequential` property on the metadata, an app developer could provide a boolean on whether a function requires approval or not.

*Pro:* This is very simple and maps to the is-consequential flag. Could be easily set with a decorator.
*Con:* This is not scalable to other metadata (like HTTP methods).

### Option 2: add a metadata property to `KernelFunctionMetadata`

With a `Dictionary<string, object> Metadata` property on the metadata, an app developer could provide any arbitrary data that could be accessed elsewhere in the system. There could be a reserved key for `IsConsequential` so that helper methods and decorators can be provided. HTTP methods could live under a key called `Method`, for example...

```json
{
    "method": "POST"
}
```

*Pro:* This can scale to any future attributes that we need to store.
*Con:* The logic to determine if something "is consequential" is forced to be included in each of the filters, which makes it difficult to reuse filters across functions with different attributes.

### Option 3: add an `IsConsequential()` method

Add a `bool IsConsequential()` method on `KernelFunction`. This would allow a developer to customize the business logic for each function. What's less clear is _how_ this method would be defined. There should likely be default behavior (e.g., if the is-consequential flag is set) and a way to set the method during import time.

[...]

Issue #5506, opened 2024-03-15
.Net: Add Support for x-openai-isConsequential Property in Semantic Kernel


lemillermicrosoft:
### Description    
We would like to request the addition of support for the `x-openai-isConsequential` property in Semantic Kernel. This property is used in the OpenAPI specification to mark certain endpoints as "consequential" and affects the way user confirmation and "always allow" buttons are handled. By implementing this feature, we can better control user interactions with actions and ensure a more consistent user experience.  
https://platform.openai.com/docs/actions/getting-started/consequential-flag
  
### Proposed Changes  
  
1. Update the parsing process in Semantic Kernel to recognize the `x-openai-isConsequential` property in the OpenAPI specification.  
2. Update the generated function models to include a boolean property indicating whether the operation is consequential or not.  
3. Adjust the confirmation behavior based on the value of the `x-openai-isConsequential` property:  
   - If `true`, always prompt the user for confirmation before running and do not show the "always allow" button.  
   - If `false`, show the "always allow" button.  
   - If the property is not present, default all GET operations to `false` and all other operations to `true`.  
  
### Benefits  
  
Adding support for the `x-openai-isConsequential` property will:  
  
- Improve user control over actions.  
- Ensure a more consistent user experience.  
- Enhance compatibility with the OpenAI platform.  
  
### Additional Considerations  
  
Please ensure that the proposed implementation is compatible with the existing features in Semantic Kernel and does not introduce breaking changes.  
  
Thank you for considering this feature request. We believe that adding support for the `x-openai-isConsequential` property will enhance the overall capabilities of Semantic Kernel and provide a more consistent experience for users interacting with actions.

matthewbolanos:
Assigning this to myself to answer the following questions...
- What is the attribute name for `KernelFunction`s to hold this info
- Should it be binary, enum, collection of attributes?
- How is this flag set for native functions
- Where should this be set? By the API author or the API consumer?
- What other types of these things do we need?
- What is the E2E scenario with hooks/filters?

westey-m:
Merged the changes.  See an example of usage here:
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs

PR #5911, opened 2024-04-17
.Net: Add support for making open api operation metadata and extension metadata available at function invocation time

westey-m:
### Motivation and Context

This change is to support scenarios where a consumer needs to use metadata about an operation to change behavior at function invocation time.

E.g. OpenAI has defined a [consequential flag](https://platform.openai.com/docs/actions/getting-started/consequential-flag) that can be added to open api operation schemas in order to indicate that the operation will have consequences if invoked. It may create a transaction in a database. Compare that to an operation that just retrieves data and has no additional consequences.

In some cases a consumer may want to use this flag as an input when deciding on whether to confirm an function invocations with a user before invocation. The LLM may have instructed the consumer to invoke the function, but since there is risk involved, the consumer may want to get user confirmation first.

Related Issues:
https://github.com/microsoft/semantic-kernel/issues/5724
https://github.com/microsoft/semantic-kernel/issues/5506

### Description

1. Added a readonly dictionary base class to KernelFunctionMetadata to allow adding any unstructured metadata to this dictionary in addition to the named properties already defined on KernelFunctionMetadata.
2. The base class approach was chosen because many other dictionary classes already exist for metadata and if we wanted to extend them with named properties a similar approach can be used without causing breaking changes.
3. When creating a plugin from an OpenAPI schema, properties for id, path and method are added to the metadata dictionary as well as an operation-extensions property, who's value is a dictionary and contains any x-* properties defined in the OpenAPI schema.

### Contribution Checklist


dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        // This HTTP client is optional. SK will fallback to a default internal one if omitted.
+        using HttpClient httpClient = new();
+
+        // Create a sample OpenAPI schema that calls the github versions api, and has an operation extension property.
+        // The x-openai-isConsequential property is the operation extension property.
+        var schema =
+            "{" +
+            "  \"openapi\": \"3.0.1\"," +
+            "  \"info\": {" +
```
dmytrostruk, line 38: With recent upgrades of C# language version in repo, you should be able to use raw string literals, which should simplify usage and representation of JSON in code. For example:
https://github.com/microsoft/semantic-kernel/blob/c72080d7da6baa539eab515fc065730d87d191e2/dotnet/src/Connectors/Connectors.UnitTests/OpenAI/ChatCompletion/OpenAIChatCompletionServiceTests.cs#L352-L356

dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs
```diff
@@ -0,0 +1,101 @@
...
+            "          }" +
+            "        }" +
+            "      }" +
+            "    }" +
+            "  }" +
+            "}";
+        var schemaStream = new MemoryStream();
+        WriteStringToStream(schemaStream, schema);
+
+        // Import an Open AI plugin from a stream.
```
dmytrostruk, line 64:
```suggestion
        // Import an OpenAPI plugin from a stream.
```

dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs
```diff
@@ -0,0 +1,101 @@
...
+        else
+        {
+            // Invoke the function and output the result.
+            var functionResult = await kernel.InvokeAsync(function, new KernelArguments());
+            var result = functionResult.GetValue<RestApiOperationResponse>();
+            WriteLine($"Function execution result: {result?.Content}");
+        }
+    }
+
+    public Example23_OpenAPIPlugin(ITestOutputHelper output) : base(output)
```
dmytrostruk, line 90: nit: This constructor can be converted to primary constructor.

dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiKernelExtensions.cs
```diff
@@ -200,6 +201,18 @@ public static class OpenApiKernelExtensions
...
+    private const string OperationExtensionsIdKey = "id";
+
+    /// <summary>The metadata property bag key to use when storing the path of an operation.</summary>
+    private const string OperationExtensionsPathKey = "path";
+
+    /// <summary>The metadata property bag key to use when storing the method of an operation.</summary>
+    private const string OperationExtensionsMethodKey = "method";
+
+    /// <summary>The metadata property bag key to use for the list of extension values provided in the swagger file at the operation level.</summary>
+    private const string OperationExtensionsMetadataKey = "operation-extensions";
```
dmytrostruk, line 214: Looks like `operation-extensions` key is used in example as well. Should we consider making this const public, so users will be able to re-use it? If we decide to convert them to public, we can think if we should keep them in `Extensions` class or more generic place.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionMetadata.cs
```diff
@@ -19,12 +20,16 @@ public sealed class KernelFunctionMetadata
...
+    /// <summary>A static empty dictionary to default to when none is provided.</summary>
+    private static readonly Dictionary<string, object?> s_emptyDictionary = new();
 
     /// <summary>Initializes the <see cref="KernelFunctionMetadata"/> for a function with the specified name.</summary>
     /// <param name="name">The name of the function.</param>
+    /// <param name="propertyBag">Optional metadata in addition to the named properties already available on this class.</param>
     /// <exception cref="ArgumentNullException">The <paramref name="name"/> was null.</exception>
     /// <exception cref="ArgumentException">An invalid name was supplied.</exception>
-    public KernelFunctionMetadata(string name)
+    public KernelFunctionMetadata(string name, ReadOnlyDictionary<string, object?>? propertyBag = null)
```
dmytrostruk, line 31: Is there an option to add new constructor overload with `propertyBag` parameter instead of modifying existing constructor? In this case there will be no need in compatibility suppression file.

Also, should be name new parameter as `metadata` instead of `propertyBag`? I think `metadata` should work better, based on description of this parameter.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFactory.cs
```diff
@@ -25,15 +26,17 @@ public static class KernelFunctionFactory
...
     public static KernelFunction CreateFromMethod(
         Delegate method,
         string? functionName = null,
         string? description = null,
         IEnumerable<KernelParameterMetadata>? parameters = null,
         KernelReturnParameterMetadata? returnParameter = null,
-        ILoggerFactory? loggerFactory = null) =>
-        CreateFromMethod(method.Method, method.Target, functionName, description, parameters, returnParameter, loggerFactory);
+        ILoggerFactory? loggerFactory = null,
+        ReadOnlyDictionary<string, object?>? metadataPropertyBag = null) =>
```
dmytrostruk, line 38: Same comment as above - I would probably keep the naming simple and use just `metadata` or just `propertyBag`. 
Most probably it won't be possible to add method overload here, taking into account amount of optional parameters.

dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs
westey-m: Oh yeah, that's much better. Will fix, thanks!

dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs
westey-m: Makes sense.  Will fix.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionMetadata.cs
dmytrostruk: Makes sense!

dotnet/src/SemanticKernel.Abstractions/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,11 @@
...
+<!-- https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids -->
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Microsoft.SemanticKernel.KernelFunctionMetadata.#ctor(System.String)</Target>
+    <Left>lib/netstandard2.0/Microsoft.SemanticKernel.Abstractions.dll</Left>
+    <Right>lib/netstandard2.0/Microsoft.SemanticKernel.Abstractions.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+</Suppressions>
```
markwallace-microsoft, line 11: As a general rule we never want to introduce breaking changes.

We've done this early on in the 1.0 cycle and for changes related to experimental features but let's try to avoid this.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionMetadata.cs
markwallace-microsoft: How about `additionalMetadata` or `extensionMetadata`

dotnet/src/SemanticKernel.Core/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,18 @@
...
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Microsoft.SemanticKernel.KernelFunctionFactory.CreateFromMethod(System.Reflection.MethodInfo,System.Object,System.String,System.String,System.Collections.Generic.IEnumerable{Microsoft.SemanticKernel.KernelParameterMetadata},Microsoft.SemanticKernel.KernelReturnParameterMetadata,Microsoft.Extensions.Logging.ILoggerFactory)</Target>
+    <Left>lib/netstandard2.0/Microsoft.SemanticKernel.Core.dll</Left>
+    <Right>lib/netstandard2.0/Microsoft.SemanticKernel.Core.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+</Suppressions>
```
markwallace-microsoft, line 18: Let's try to eliminate this also

dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs
```diff
@@ -0,0 +1,101 @@
...
+            "  }," +
+            "  \"servers\": [" +
+            "    {" +
+            "      \"url\": \"https://api.github.com\"" +
+            "    }" +
+            "  ]," +
+            "  \"paths\": {" +
+            "    \"/versions\": {" +
+            "      \"get\": {" +
+            "        \"x-openai-isConsequential\": false," +
```
markwallace-microsoft, line 50: Make this true for the sake of the sample

dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiKernelExtensions.cs
```diff
@@ -333,13 +346,24 @@ async Task<RestApiOperationResponse> ExecuteAsync(KernelArguments variables, Can
 
         var returnParameter = operation.GetDefaultReturnParameter();
 
+        // Add unstructured metadata, specific to Open API, to the metadata property bag.
+        var metadataPropertyBag = new Dictionary<string, object?>();
+        metadataPropertyBag.Add(OpenApiKernelExtensions.OperationExtensionsIdKey, operation.Id);
+        metadataPropertyBag.Add(OpenApiKernelExtensions.OperationExtensionsPathKey, operation.Path);
+        metadataPropertyBag.Add(OpenApiKernelExtensions.OperationExtensionsMethodKey, operation.Method.ToString());
+        if (operation.Extensions is not null && operation.Extensions.Any())
```
SergeyMenshykh, line 354:
```suggestion
        if (operation.Extensions is { Count: > 0 })
```
nit: the new pattern matching syntaxis that checks that the collection is initialized and contains elements.

dotnet/src/Functions/Functions.OpenApi/OpenApi/OpenApiDocumentParser.cs
```diff
@@ -183,14 +186,63 @@ internal static List<RestApiOperation> CreateRestApiOperations(string? serverUrl
...
+                {
+                    var schemaBuilder = new StringBuilder();
+                    var jsonWriter = new OpenApiJsonWriter(new StringWriter(schemaBuilder, CultureInfo.InvariantCulture), new OpenApiJsonWriterSettings() { Terse = true });
+                    extension.Value.Write(jsonWriter, Microsoft.OpenApi.OpenApiSpecVersion.OpenApi3_0);
+                    extensionValueObj = schemaBuilder.ToString();
+                }
+            }
+            else
+            {
+                throw new KernelException($"The type of extension property '{extension.Key}' is not supported while trying to consume the OpenApi schema.");
```
SergeyMenshykh, line 237: I'm not sure if we need to throw an exception here. I'm considering cases when someone imports an OpenAPI specification that might have extensions like `OpenApiDeprecationExtension`, `OpenApiPagingExtension`, and other derivatives of `IOpenApiExtension`, but the scenario doesn't require those for execution. In those cases, we should not stop the import operation because we don't support the particular type of extension.  
   
On the other hand, there could be scenarios that require the extensions. To support them, they need to be reported, and corresponding functionality added.

dotnet/samples/KernelSyntaxExamples/Example23_OpenAPIPlugin.cs
westey-m: I did consider it, but version is not a consequential action, so I felt making it true would be misleading.  Instead I suggested in the xml comments that people should play around with the example by changing the value and seeing the outcome.

dotnet/src/Functions/Functions.OpenApi/OpenApi/OpenApiDocumentParser.cs
westey-m: Maybe the best is just to avoid adding them to the collection if we do not recognize them. That way we can always add them later, once we have an appropriate way of representing them.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFactory.cs
westey-m: Adopted `additionalMetadata` as suggested by @markwallace-microsoft

dotnet/src/SemanticKernel.Abstractions/CompatibilitySuppressions.xml
westey-m: I used @dmytrostruk's suggestion to eliminate this one.  Will see if something similar works with the other one.

dotnet/src/SemanticKernel.Core/CompatibilitySuppressions.xml
westey-m: Found a way to remove this breaking change, but it required me to introduce a new options parameter. I think this is better anyway, since there are too many optional parameters already, and this is a better pattern.

dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiKernelExtensions.cs
westey-m: I'm looking into a few options to make these more easily accessible, and will get back on it later.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethodOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.ComponentModel;
+using System.Reflection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Optional options that can be provided when creating a <see cref="KernelFunction"/> from a method.
+/// </summary>
+public sealed class KernelFunctionFromMethodOptions
```
dmytrostruk, line 15: Since it's `SemanticKernel.Core`, I'm wondering if we should mark this class as `Experimental` for the first period of time. Same for public places where this class is used. This will give us more space if we want to update it later, rename it or fix some issues. In this case we should remember to remove `Experimental` flag after some time.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethodOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+    public string? FunctionName { get; init; }
+
+    /// <summary>
+    /// The description to use for the function. If null, it will default to one derived from the passed <see cref="Delegate"/> or <see cref="MethodInfo"/>, if possible
+    /// (e.g. via a <see cref="DescriptionAttribute"/> on the method).
+    /// </summary>
+    public string? Description { get; init; }
+
+    /// <summary>
+    /// >Optional parameter descriptions. If null, it will default to one derived from the passed <see cref="Delegate"/> or <see cref="MethodInfo"/>.
```
dmytrostruk, line 29:
```suggestion
    /// Optional parameter descriptions. If null, it will default to one derived from the passed <see cref="Delegate"/> or <see cref="MethodInfo"/>.
```

dmytrostruk (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethodOptions.cs
westey-m: Added, thanks.

SergeyMenshykh (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
```diff
@@ -49,22 +50,50 @@ internal sealed class KernelFunctionFromMethod : KernelFunction
...
+
+    /// <summary>
+    /// Creates a <see cref="KernelFunction"/> instance for a method, specified via an <see cref="MethodInfo"/> instance
+    /// and an optional target object if the method is an instance method.
+    /// </summary>
+    /// <param name="method">The method to be represented via the created <see cref="KernelFunction"/>.</param>
+    /// <param name="target">The target object for the <paramref name="method"/> if it represents an instance method. This should be null if and only if <paramref name="method"/> is a static method.</param>
+    /// <param name="options">Optional function creation options.</param>
+    /// <returns>The created <see cref="KernelFunction"/> wrapper for <paramref name="method"/>.</returns>
+    [Experimental("SKEXP0001")]
```
markwallace-microsoft, line 75: This is an internal class so I don't think we need this

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethodOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Reflection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Optional options that can be provided when creating a <see cref="KernelFunction"/> from a method.
+/// </summary>
+[Experimental("SKEXP0001")]
+public sealed class KernelFunctionFromMethodOptions
```
markwallace-microsoft, line 17: Does this need to be public?

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
```diff
@@ -49,22 +50,50 @@ internal sealed class KernelFunctionFromMethod : KernelFunction
...
+            method,
+            target,
+            new KernelFunctionFromMethodOptions
+            {
+                FunctionName = functionName,
+                Description = description,
+                Parameters = parameters,
+                ReturnParameter = returnParameter,
+                LoggerFactory = loggerFactory
+            });
```
markwallace-microsoft, line 64: Why is this change needed as part of this PR?

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFactory.cs
```diff
@@ -35,6 +35,18 @@ public static class KernelFunctionFactory
         ILoggerFactory? loggerFactory = null) =>
```
markwallace-microsoft, line 35: Not for this PR, but do we want to mark this old method as obsolete (at some time in the future) to encourage developers to use the new method?

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
westey-m: Oops, good catch.  Removed.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFactory.cs
westey-m: Agreed, I also think we should obsolete the existing ones, and makes sense to do it in a separate PR.

markwallace-microsoft (approved):

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs
```diff
@@ -117,18 +117,19 @@ internal KernelFunction(string name, string description, IReadOnlyList<KernelPar
...
         Verify.ParametersUniqueness(parameters);
 
-        this.Metadata = new KernelFunctionMetadata(name, properties)
+        this.Metadata = new KernelFunctionMetadata(name)
         {
             PluginName = pluginName,
             Description = description,
             Parameters = parameters,
             ReturnParameter = returnParameter ?? KernelReturnParameterMetadata.Empty,
+            AdditionalProperties = additionalMetadata ?? KernelFunctionMetadata.s_emptyDictionary,
```
markwallace-microsoft, line 132: Should this be named `AdditionalMetadata`?

westey-m: I think outside of the metadata class AdditionalMetadata makes sense, since it's clarifying where the data will end up.  However inside the metadata class the word metadata is unnecessary and properties make more sense.

dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethod.cs
```diff
@@ -206,7 +206,7 @@ private record struct MethodDetails(string Name, string Description, Implementat
         IReadOnlyList<KernelParameterMetadata> parameters,
         KernelReturnParameterMetadata returnParameter,
         ReadOnlyDictionary<string, object?>? additionalMetadata = null) :
-        base(functionName, pluginName, description, parameters, returnParameter, properties: additionalMetadata)
+        base(functionName, pluginName, description, parameters, returnParameter, additionalMetadata: additionalMetadata)
```
markwallace-microsoft, line 209: I'd keep the naming consistent either `additionalMetadata` or `additionalProperties` everywhere

westey-m: I think context matters.  Outside of the metadata class AdditionalMetadata makes sense, since it's clarifying where the data will end up. However inside the metadata class the word metadata is unnecessary and properties make more sense.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunctionMetadata.cs
```diff
@@ -91,4 +97,15 @@ public KernelReturnParameterMetadata ReturnParameter
             this._returnParameter = value;
         }
     }
+
+    /// <summary>Gets optional metadata in addition to the named properties already available on this class.</summary>
+    public ReadOnlyDictionary<string, object?> AdditionalProperties
```
markwallace-microsoft, line 102: Add the experimental marker

westey-m: Makes sense.  Done.

SergeyMenshykh (approved):

PR #5963, opened 2024-04-22
.Net: ADR: OTel LLM requests

TaoChenOSU:
### Motivation and Context

Observing LLM applications has been a huge ask from customers and the community. This work aims to ensure that SK provides the best developer experience while complying with the industry standards for observability in generative-AI-based applications.

### Description

This ADR outlines options which we can use to trace LLM requests from applications built with SK.

### Contribution Checklist


docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,460 @@
...
+|---|---|
+|`gen_ai.content.prompt`|`gen_ai.prompt`|
+|`gen_ai.content.completion`|`gen_ai.completion`|
+
+> The kernel must provide configuration options to disable these events because they may contain PII.
+> See the [Semantic conventions for generative AI](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/gen-ai) for requirement level for these attributes.
+
+## Where do we create the activities
+
+It is crucial to establish a clear line of responsibilities, particularly since certain service providers, such as the Azure OpenAI SDK, have pre-existing instrumentation (in which case SK won't need to track LLM requests). Our objective is to position our activities as close to the model level as possible to promote a more cohesive and consistent developer experience.
```
TaoChenOSU, line 54: The Azure SDK team is planning to adapt to the OTel semantic conventions for LLMs.

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,460 @@
...
+
+#### Static class `ModelDiagnostics`
+
+This class will live under `SemanticKernel.Abstractions/AI`.
+
+`ModelDiagnostics`
+
+```C#
+// Example
+public static class ModelDiagnostics
```
stephentoub, line 113: Why does this need to be public? We'd have a lot more flexibility, including for how it evolves, if it's just an implementation detail built into whatever needs it.

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,460 @@
...
+public interface IModelInstrumentation
+{
+    Activity? StartCompletionActivity(string name, string modelName, string modelProvider, ChatHistory chatHistory, PromptExecutionSettings? executionSettings);
+
+    Activity? StartCompletionActivity(string name, string modelName, string modelProvider, string prompt, PromptExecutionSettings? executionSettings);
+
+    void SetCompletionResponses(Activity? activity, IReadOnlyList<TextContent> completions, CompletionUsage? usage);
+
+    void SetCompletionResponses(Activity activity, IReadOnlyList<ChatMessageContent> completions, CompletionUsage? usage);
+}
```
stephentoub, line 209: Why does any of this need to be exposed right now?

What would another implementation of IModelInstrumentation do differently from the default that it needs to be configurable?

Are you viewing `builder.Services.AddSingleton<IModelInstrumentation, DefaultModelInstrumentation>();` as the way of opt-ing in to this information being included?

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: All connectors are individual projects, so they need to be able to reference this class.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: A developer can attach more tags/attributes depending on his or her use cases in a custom implementation. 

And yes, I see `builder.Services.AddSingleton<IModelInstrumentation, DefaultModelInstrumentation>();` as the way to include LLM request activities in the traces.

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,460 @@
...
+
+|      | Connectors/Services                                                                                           | Clients                                                                                                          |
+| ---- | ------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------- |
+| Pros | 1. Uniform implementation across all services.                                                                | 1. Enhanced intuitiveness regarding the overall data flow.<br>2. Complete model specs and LLM requests/response. |
+| Cons | 1. Inconsistencies in the origins of telemetry data.<br>2. Incomplete model specs and LLM requests/responses. | 1. Separate implementations may lead to increased maintenance overhead.                                          |
```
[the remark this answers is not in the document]
TaoChenOSU: Yes, the Azure SDK will offer observability features that comply with the new LLM semantic conventions in the near future. When those features become available, the OpenAI connector doesn't need to track LLM requests.

I don't think we will double count tokens or corrupt trace data when both the OpenAI connector and Azure SDK produce a similar data set, as the data collection is independent. The most pressing concern, however, is that users may observe two analogous activities as part of their trace data. I think it is essential that we eliminate the instrumentation code within the OpenAI connector upon updating to the version of Azure SDK that facilitates tracing LLM requests.

docs/decisions/0040-OTel-semantic-convention.md
stephentoub: It needn't be public. It could be a shared file built into each project, for example.

docs/decisions/0040-OTel-semantic-convention.md
stephentoub: > A developer can attach more tags/attributes depending on his or her use cases in a custom implementation.

And to do that they then need to provide a full implementation of this interface? How does it compose, e.g. if I want to use both the default implementation but then I also want to add this or that tag?

docs/decisions/0040-OTel-semantic-convention.md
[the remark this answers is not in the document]
stephentoub: > I don't think we will double count tokens or corrupt trace data when both the OpenAI connector and Azure SDK produce a similar data set, as the data collection is independent. 

Isn't some of it based on metrics and with fixed / known names? What happens when the same data is counted twice in the same instrument?

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: Sure, that is also an option. If we link this file in all connector projects, this file probably can't live under `SemanticKernel.Abstractions/AI`. `InternalUtilities` is a more suitable place.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: Yeah, that's a good use case. Right now, they'd need to rewrite everything. I will need to think about if this flexibility adds value, and if so how to go about it.

docs/decisions/0040-OTel-semantic-convention.md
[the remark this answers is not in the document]
TaoChenOSU: The conventions currently don't include metrics yet. But once it does and libraries adapt to it, we will need to remove existing metrics as well.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: I actually think we should not let users opt in to have this information included in the traces. It should be included by default. And the way it works in .Net is that if [ActivitySource.StartActivity](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitysource.startactivity) internally determines if there are any listeners recording the Activity. If there are no registered listeners or there are listeners that are not interested, no activities will be created.

Another option is to create static ModelDiagnostics member in each service/client. This approach brings the following benefits:
1. It still follows the [best practice](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-instrumentation-walkthroughs#best-practices-1) to create an activity source once within a component.
2. It makes it possible to have multiple activity sources with different names, e.g. `Microsoft.SemanticKernel.Connectors.HuggingFace` and `Microsoft.SemanticKernel.Connectors.Google`.

I will also include this in the ADR.

stephentoub: Yes, we should be doing all the main tracing/events/metrics by default. We will want all of the sensitive stuff to be opt-in, though, e.g. including the actual prompt as a tag on an activity.  There are a variety of ways that could be exposed.

As for adding additional information beyond that, I think it'd be better to model it similar to how the other filters are being modeled, basically a callback that gives you the activity and other relevant information (e.g. the Kernel, KernelFunction, KernelArguments, the prompt, whatever), and you can tag the activity however you like. If you have multiple filters, they'd all have an opportunity to get in on it. As with other filters, you could register these in DI, or just add them to the Kernel directly. Opting in to having the sensitive data traced could just be an extra built-in implementation of that interface, such that it'll be invoked like any others.  Or something like that.

TaoChenOSU: Updated. 

I agree. According to the semantic conventions, all the sensitive stuff (currently only includes prompts and responses) will be events, which are optional and configurable. I have listed two options in the ADR as to how to configure it (through `IsEventCollectionEnable`). 

For additional information, if we go the filter route and would like to trace information at the function level, I think it's probably worth use separate activities, as it is no longer relevant to the LLM requests. We actually have that implemented already: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Functions/KernelFunction.cs#L159.

stephentoub: > if we go the filter route

Why would that need separate activities?

TaoChenOSU: 1. Users who want to view LLM request details may not be interested in seeing kernel function details.
2. Activities that trace LLM requests contain a special set of attributes standardized by the LLM semantic conventions.
3. Even if the above two aren't enough to justify the separation, kernel functions and connectors/models are decoupled entities. Of course, we can attach attributes in connectors to activities created in kernel functions by using `Activity.Current?`. I personally think that makes the code harder to maintain.

stephentoub: But that's not filters requiring separate activities. We can choose to use multiple activities, and doing so makes sense, if we want to have one that covers all function invocation and another per service call, but that's orthogonal to augmenting them with additional tags. Filters can be invoked for each activity, and the filter can decide what to add to what.

TaoChenOSU: Agree. The interface approach appears to offer minimal added value for users while generating additional confusion, so I have removed it.

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,628 @@
...
+
+```C#
+// In client
+public async Task<IReadOnlyList<TextContent>> GenerateTextAsync(
+    string prompt,
+    PromptExecutionSettings? executionSettings,
+    CancellationToken cancellationToken)
+{
+    using var activity = ModelDiagnostics.StartCompletionActivity(
+        $"HuggingFace {this._modelId}",
```
lmolkova, line 145: is it the span name? 

the semconv suggests `{operation} {model}`, e.g. `chat.completions gpt-4`, but we should definitely get more feedback on it.

I believe that operation name/type is very useful in the name, but having both a model and a vendor might be redundant. WDYT? Any strong feelings on the naming?

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,628 @@
...
+- (Recommended)`gen_ai.response.id`
+- (Recommended)`gen_ai.response.model`
+- (Recommended)`gen_ai.response.finish_reasons`
+- (Recommended)`gen_ai.response.prompt_tokens`
+- (Recommended)`gen_ai.response.completion_tokens`
+
+The following events will be optionally attached to an activity:
+| Event name| Attribute(s)|
+|---|---|
+|`gen_ai.content.prompt`|`gen_ai.prompt`|
```
lmolkova, line 46: I want to change it to something like https://github.com/traceloop/semantic-conventions/pull/3, actively working on it.

(event per message with more-or-less standardizes structure)

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,628 @@
...
+    CancellationToken cancellationToken)
+{
+    using var activity = ModelDiagnostics.StartCompletionActivity(
+        $"HuggingFace {this._modelId}",
+        this._modelId,
+        "HuggingFace",
+        prompt,
+        executionSettings);
+
+    var completions = ...;
```
lmolkova, line 151: nit: it's probably implied, but it's crucial to catch (and rethrow) exceptions and set error details on the activity

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,628 @@
...
+Example usage
+
+```C#
+// In client
+public async Task<IReadOnlyList<TextContent>> GenerateTextAsync(
+    string prompt,
+    PromptExecutionSettings? executionSettings,
+    CancellationToken cancellationToken)
+{
+    using var activity = ModelDiagnostics.StartCompletionActivity(
```
lmolkova, line 144: are those `internal` or client `activities`?

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,628 @@
...
+- (Required)`gen_ai.system`
+- (Required)`gen_ai.request.model`
+- (Recommended)`gen_ai.request.max_token`
+- (Recommended)`gen_ai.request.temperature`
+- (Recommended)`gen_ai.request.top_p`
+- (Recommended)`gen_ai.response.id`
+- (Recommended)`gen_ai.response.model`
+- (Recommended)`gen_ai.response.finish_reasons`
+- (Recommended)`gen_ai.response.prompt_tokens`
+- (Recommended)`gen_ai.response.completion_tokens`
```
lmolkova, line 41: we'll add a few more attributes into the semconv soon: [`server.address` and `server.port`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/server.md) to identify the endpoint, [`error.type`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/error.md) to record errors.

docs/decisions/0040-OTel-semantic-convention.md
```diff
@@ -0,0 +1,628 @@
...
+## Appendix
+
+### `PromptExecutionSettings`
+
+```C#
+public class PromptExecutionSettings
+{
+    ...
+    [JsonPropertyName("is_event_collection_enable")]
+    public bool IsEventCollectionEnabled
```
lmolkova, line 486: this is a great start, but I think we'll need to have more options:

- don't collect events at all
- collect events, but without sensitive parts. E.g. we might collect `{"role":"assistant","tool_calls":[{"id":"call_hHM72v9f1JprJBStycQC4Svz","function":{"name":"get_weather","arguments":"{\"location\": \"REFDACTED\"}"},"type":"function"}]}` or `{"role":"user", "content":"REDACTED"}`

so maybe enum would work better, like

```csharp
enum EventCollectionLevel 
{
   None,
   RedactSensitiveData,
   All,
}
```

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: Good catch! It's meant to be `{operation} {model}`. I must have changed during one of the revisions.

TaoChenOSU:
The primary focus of our current deliberation centers on the necessity for uniformity across all AI connectors, specifically regarding the production of identical telemetry with a consistent data structure, irrespective of the variations in underlying libraries or implementations. One potential drawback of implementing such uniformity is the associated cost, as users might incur additional expenses due to the storage of redundant data. This is particularly relevant since libraries could generate identical telemetry for Large Language Model (LLM) requests. Conversely, a significant advantage of this approach is the convenience it offers users, who would not be required to configure telemetry each time they replace connectors, thus ensuring consistency.

TaoChenOSU:
An alternative strategy we might consider involves establishing uniformity at the connector level, wherein by default, all connectors generate telemetry for LLM requests. However, we would provide developers with the flexibility to disable this feature if they choose. This approach allows for a balance between maintaining consistency across our connectors and offering customization options to accommodate individual developer preferences.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: These will be internal.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: I think we can explore this level of granularity in the future. For now, I have provided the option to opt in to sensitive data collection instead of event collection.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: Will keep an eye on that.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: Once the ADR is approved. I will either submit a separate PR containing the full implementation.

docs/decisions/0040-OTel-semantic-convention.md
TaoChenOSU: Looking forward to it.

stephentoub (approved):
High level looks good. Will provide more detailed feedback on implementation.

docs/decisions/0044-OTel-semantic-convention.md
```diff
@@ -0,0 +1,326 @@
...
+Observing LLM applications has been a huge ask from customers and the community. This work aims to ensure that SK provides the best developer experience while complying with the industry standards for observability in generative-AI-based applications.
+
+For more information, please refer to this issue: https://github.com/open-telemetry/semantic-conventions/issues/327
+
+### Semantic conventions
+
+The semantic conventions for generative AI are currently in their nascent stage, and as a result, many of the requirements outlined here may undergo changes in the future. Consequently, several features derived from this Architectural Decision Record (ADR) may be considered experimental. It is essential to remain adaptable and responsive to evolving industry standards to ensure the continuous improvement of our system's performance and reliability.
+
+- [Semantic conventions for generative AI](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/gen-ai)
+- [Generic LLM attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/llm.md)
```
lmolkova, line 24:
```suggestion
- [Generic LLM attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/gen-ai.md)
```

the file was recently renamed. Once we have semconv release it could make sense to add a link to the tag, but it's not  possible now. Permalink would be out of date too soon :)

docs/decisions/0044-OTel-semantic-convention.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Telemetry requirements (Experimental)
+
+Based on the initial version, Semantic Kernel should provide the following attributes in activities that represent individual LLM requests:
```
lmolkova, line 28:
```suggestion
Based on the [initial version](https://github.com/open-telemetry/semantic-conventions/blob/651d779183ecc7c2f8cfa90bf94e105f7b9d3f5a/docs/attributes-registry/gen-ai.md?plain=1#L1), Semantic Kernel should provide the following attributes in activities that represent individual LLM requests:
```

here permalink seems to be more relevant. 

As you know some updates/changes are coming soon.

docs/decisions/0044-OTel-semantic-convention.md
```diff
@@ -0,0 +1,326 @@
...
+    AppContext.SetSwitch("Microsoft.SemanticKernel.Experimental.EnableModelDiagnosticsWithSensitiveData", true);
+}
+
+// Or in the project file
+<ItemGroup Condition="'$(Configuration)' == 'Release'">
+    <RuntimeHostConfigurationOption Include="Microsoft.SemanticKernel.Experimental.EnableModelDiagnostics" Value="true" />
+</ItemGroup>
+
+<ItemGroup Condition="'$(Configuration)' == 'Debug'">
+    <RuntimeHostConfigurationOption Include="Microsoft.SemanticKernel.Experimental.EnableModelDiagnosticsWithSensitiveData" Value="true" />
```
lmolkova, line 220: nice!

lmolkova (approved):

docs/decisions/0044-OTel-semantic-convention.md
```diff
@@ -0,0 +1,332 @@
...
+internal static class ModelDiagnostics
+{
+    // Consistent namespace for all connectors
+    private static readonly string s_namespace = typeof(ModelDiagnostics).Namespace;
+    private static readonly ActivitySource s_activitySource = new(s_namespace);
+
+    private const string EnableModelDiagnosticsSettingName = "Microsoft.SemanticKernel.Experimental.EnableModelDiagnostics";
+    private const string EnableSensitiveEventsSettingName = "Microsoft.SemanticKernel.Experimental.EnableModelDiagnosticsWithSensitiveData";
+
+    private static readonly bool s_enableSensitiveEvents = AppContextSwitchHelper.GetConfigValue(EnableSensitiveEventsSettingName);
```
alliscode, line 265: Should we not respect changes to this AppContext switches that are made at runtime with a call to `AppContext.SetSwitch`?

stephentoub: FWIW, we generally don't in dotnet/runtime. For example, we use a switch to determine whether to stick with ICU as the globalization library or to instead fall back to using NLS on Windows, and that's a choice we make once on first use and cache. I think it's fine to do that here, as well. Especially since we're talking about opting in to sharing sensitive information: this is a process-global flag, so if you were trying to do it around just an individual operation, you'd end up also enabling it for any other work in the process happening concurrently. When we want to enable this for real with public API, I think it'll make sense to do so with more fine-grained opt-in mechanisms on Kernel, which will let you do it for individual operations without concurrency concerns, but for now a global experimental switch that you just configure once should be good enough.

alliscode (approved):

PR #5982, opened 2024-04-23
.Net - Add Coverage for Agent Projects

crickman:
### Contribution Checklist

### DIFF
--- a/.github/workflows/dotnet-build-and-test.yml
+++ b/.github/workflows/dotnet-build-and-test.yml
@@ -127,7 +127,7 @@ jobs:
           targetdir: "./TestResults/Reports"
           reporttypes: "JsonSummary"
           # Report for production packages only
-          assemblyfilters: "+Microsoft.SemanticKernel.Abstractions;+Microsoft.SemanticKernel.Core;+Microsoft.SemanticKernel.PromptTemplates.Handlebars;+Microsoft.SemanticKernel.Connectors.OpenAI;+Microsoft.SemanticKernel.Yaml;"
+          assemblyfilters: "+Microsoft.SemanticKernel.Abstractions;+Microsoft.SemanticKernel.Core;+Microsoft.SemanticKernel.PromptTemplates.Handlebars;+Microsoft.SemanticKernel.Connectors.OpenAI;+Microsoft.SemanticKernel.Yaml;+Microsoft.SemanticKernel.Agents.Abstractions;+Microsoft.SemanticKernel.Agents.Core;+Microsoft.SemanticKernel.Agents.OpenAI"
 
       - name: Check coverage
         shell: pwsh


.github/workflows/dotnet-build-and-test.yml
```diff
@@ -127,7 +127,7 @@ jobs:
           targetdir: "./TestResults/Reports"
           reporttypes: "JsonSummary"
           # Report for production packages only
-          assemblyfilters: "+Microsoft.SemanticKernel.Abstractions;+Microsoft.SemanticKernel.Core;+Microsoft.SemanticKernel.PromptTemplates.Handlebars;+Microsoft.SemanticKernel.Connectors.OpenAI;+Microsoft.SemanticKernel.Yaml;"
+          assemblyfilters: "+Microsoft.SemanticKernel.Abstractions;+Microsoft.SemanticKernel.Core;+Microsoft.SemanticKernel.PromptTemplates.Handlebars;+Microsoft.SemanticKernel.Connectors.OpenAI;+Microsoft.SemanticKernel.Yaml;+Microsoft.SemanticKernel.Agents.Abstractions;+Microsoft.SemanticKernel.Agents.Core;+Microsoft.SemanticKernel.Agents.OpenAI"
```
dmytrostruk, line 130: As comment says, we are including code coverage check for production packages only. If we don't expect huge changes in these assemblies in the future, then we can include them and update this comment. If you expect big changes and more experimentation in the future, this step will slow the delivery process. It's not a bad thing, it's just for your information.

Another thing, since this PR does not include changes in .NET, the tests and coverage check weren't invoked. It would be nice to see test code coverage run to verify that all packages satisfy 80% coverage threshold.

dmytrostruk (approved):

.github/workflows/dotnet-build-and-test.yml
crickman: Runtime doesn't look affected much, but I need to increase coverage (I guess I was measuring the `AgentSyntaxExamples` coverage along with UT locally):

image

dmytrostruk: > I do feel maintaining normal dev-metrics on these is important as we do plan on graduating these and they are intended to become an important component in the SK eco-system.

Yes, totally agree!

> Runtime doesn't look affected much, but I need to increase coverage

It's good that we caught that here, because if we would merge this PR as it is, it would break other PR pipelines with low coverage error.

crickman: Agree - This is what I was seeing locally.  So glad we got these numbers.  Thanks for suggestion.

image

crickman: Unit-Tests Only:

image

SergeyMenshykh (approved):

Issue #5796, opened 2024-04-07
.Net: Gemini function calling throws error when a kernel function with `Enum` parameter is available


HillPhelmuth:
**Describe the bug**
When using function calling with Gemini `ToolCallBehavior` set to `GeminiToolCallBehavior.AutoInvokeKernelFunctions`, and the Kernel instance has a plugin function with an `Enum` parameter, the Gemini request will return a 400 (bad request) error.

**To Reproduce**
1. Create `Kernel` instance with a plugin containing a function with enum parameter (`TimePlugin` is simple example to test).
2. Create `GeminiPromptExecutionSettings` with `ToolCallBehavior = GeminiToolCallBehavior.AutoInvokeKernelFunctions`
3. See error when running `InvokePromptAsync` that includes the execution settings.

**Expected behavior**
Similar to current OpenAI function calling, the Gemini connector should handle necessary type changes when converting kernel functions to gemini function definitions to avoid these errors

**Platform**
 - OS: Windows
 - IDE: Visual Studio 2022
 - Language: C#
 - Source: Version 1.7.1 and 1.7.1-alpha

**Additional context**
[Created example that demonstrates the issue](https://github.com/HillPhelmuth/GeminiFunctionCallErrorExample/blob/main/GeminiFunctionCallErrorExample/Program.cs) - uses `TimePlugin` which causes error and `TimePlugin` with `DateMatchingLastDayName` function removed, which does not cause the error.

cisionmarkwalls:
I get errors even when no enum parameters are used with AutoInvokeKernelFunctions.

As an example:
```
...
kernel.ImportPluginFromFunctions("HelperFunctions",
[
    kernel.CreateFunctionFromMethod(() => DateTime.UtcNow.ToString("R"), "GetCurrentUtcTime", "Retrieves the current time in UTC."),            
]);
GeminiPromptExecutionSettings settings = new() {
    ToolCallBehavior = GeminiToolCallBehavior.AutoInvokeKernelFunctions
};
try
{
    FunctionResult result = await kernel.InvokePromptAsync(
       "What is the current time?", new(settings));

    // Display results
    var geminiMetadata = result.Metadata as GeminiMetadata;
...
```
Gives me a 500 error from Gemini for gemini-1.0-pro-002 and gemini-1.5-pro-preview-0409

cisionmarkwalls:
> You should check at first in google docs do selected versions of models support function calling.

[...]

PR #5998, opened 2024-04-24
.Net: Fix 5796 function calling enum params

Krzysztof318:
### Motivation and Context

Fixes https://github.com/microsoft/semantic-kernel/issues/5796

### Description

Type for enum wasn't correctly set in JsonSchemaMapper, it didn't matter for OpenAI but gemini was throwing exception if type isn't specified. Fixed that with `string` type.
Added new unit tests for Gemini and OpenAI. Both passed.

@RogerBarreto 
@SergeyMenshykh

DataHelper and BertOnyx was updated automatically by formatter.

### Contribution Checklist


dotnet/src/InternalUtilities/src/Schema/JsonSchemaMapper.cs
```diff
@@ -321,12 +321,8 @@ public static string ToJsonString(this JsonNode? node, bool writeIndented = fals
...
                     if (TryGetStringEnumConverterValues(typeInfo, effectiveConverter, out JsonArray? values))
                     {
-                        if (values is null)
-                        {
-                            // enum declared with the flags attribute -- do not surface enum values in the JSON schema.
-                            schemaType = JsonSchemaType.String;
-                        }
-                        else
+                        schemaType = JsonSchemaType.String;
+                        if (values != null)
```
markwallace-microsoft, line 325: I'm looking here: https://json-schema.org/understanding-json-schema/reference/enum

Which says _You can use enum even without a type, to accept values of different types._

Could the bug be in the Google service? Can you check to see has this been reported as a issue.

You mention things work fine with OpenAI. I just tried MistralAI and it works with and without the type.

Krzysztof318: Maybe this is limitation Google rest api. It always requires specify type.

Krzysztof318: @markwallace-microsoft I think adding type is good solution, it doesn't break other connectors and work with gemini, JsonSchemaMapper uses `TryGetStringEnumConverterValues` which internally adds values as strings
```cs
var namingPolicy = (JsonNamingPolicy?)namingPolicyField!.GetValue(converter)!;
                string[] names = Enum.GetNames(typeInfo.Type);
                values = [];
                foreach (string name in names)
                {
                    string effectiveName = namingPolicy?.ConvertName(name) ?? name;
                    values.Add((JsonNode)effectiveName);
                }
```

SergeyMenshykh: @Krzysztof318 I wonder if there is any other option on the table that we can consider for fixing the issue?
   
My concern with the fix is that it will get lost when/if the temporary [JSON schema generator classes](https://github.com/microsoft/semantic-kernel/tree/main/dotnet/src/InternalUtilities/src/Schema) are replaced by ones built into System.Text.Json:
image

CC: @eiriktsarpalis

eiriktsarpalis: Regarding this change in particular, it seems reasonable to also include the type keyword for enums. It might be redundant from the perspective of how `enum` works, but it also doesn't change evaluation semantics fundamentally. If it helps speed things up, you could also try filing a PR to the prototype repo directly.

eiriktsarpalis: I've created a PR in the prototype repo: https://github.com/eiriktsarpalis/stj-schema-mapper/pull/1

Feel free to update this PR with the diff used there.

Krzysztof318: @eiriktsarpalis @markwallace-microsoft I have updated code to be same like in schema mapper repo.

SergeyMenshykh (approved):

markwallace-microsoft:
@Krzysztof318 could you take a look at the build error? Thanks.

Krzysztof318:
@markwallace-microsoft until 19.05 I am out of office

RogerBarreto (approved):

markwallace-microsoft (approved):

PR #6009, opened 2024-04-25
Java: Assistants 

dsgrieve:
Initial source. Work in progress.
Code will wrap Azure OpenAI assistants models. 

- Need to annotate with Nullable.
- Need a AssistantService  
- Need unit and/or end-to-end tests
- No vector store API yet. Azure OpenAI does not have vector store at the moment.


dsgrieve:
Moved to [semantic-kernel-java](https://github.com/microsoft/semantic-kernel-java)

PR #6012, opened 2024-04-26
.Net: Text Search Service ADR (Work in Progress)

markwallace-microsoft:
### Motivation and Context

This is a POC for the new text search service abstraction. Sharing very early for team collaboration.

### Contribution Checklist


docs/decisions/00NN-text-search-service.md
```diff
@@ -0,0 +1,332 @@
...
+consulted: 
+informed: stephentoub, matthewbolanos
+---
+
+# Text Search Service
+
+## Context and Problem Statement
+
+Semantic Kernel has support for searching using popular Vector databases e.g. Azure AI Search, Chroma, Milvus and also Web search engines e.g. Bing, Google.
+There are two sets of abstractions and plugins depending on whether the developer wants to perform search against a Vector database of a Web search engine.
```
SergeyMenshykh, line 16: Should it be "a Vector database **or** a Web search engine."?

dotnet/samples/Concepts/Search/TextSearchPluginExample.cs
```diff
@@ -0,0 +1,44 @@
...
+        // Invoke the plugin to perform a text search and return string values
+        var question = "What is the Semantic Kernel?";
+        var function = kernel.Plugins["StringSearch"]["Search"];
+        var result = await kernel.InvokeAsync(function, new() { ["query"] = question });
+
+        Console.WriteLine(result);
+
+        // Invoke the plugin to perform a text search and return BingWebPage values
+        function = kernel.Plugins["PageSearch"]["Search"];
+        result = await kernel.InvokeAsync(function, new() { ["query"] = question, ["count"] = 2 });
```
dmytrostruk, line 40: `SearchAsync` method is generic, and each implementation has these checks: `typeof(T) == typeof(string)` or `typeof(T) == typeof(TextSearchResult)`. `TextSearchPlugin` class is also generic, and for each specific type you need to create separate instance (e.g. `TextSearchPlugin<string>` and `TextSearchPlugin<BingWebPage>`). 

My point during initial discussion was - with generics and type checking it's already a sacrifice, but at the end of the day we still can get only `string` from `kernel.InvokeAsync`, and if it's object, it will be serialized. I think it will be really useful to get the actual search result type like `result.GetValue<BingWebPage>` after `InvokeAsync` method. Otherwise, we can probably avoid using generics, work just with `object`, and the result of this sample will still be the same. 

Just some ideas what we can do in theory:
1. Make `ITextSearchService<T>` generic.
2. For specific service (e.g. `BingTextSearchService`), implement interface with all possible types that can be produced by this service: `BingTextSearchService: ITextSearchService<string>, ITextSearchService<BingWebPage>` - this should allow to avoid `typeof(T) == typeof(string)` check in `SearchAsync` implementation.
3. Based on that, we don't need `TextSearchPlugin` class, because we can mark all `SearchAsync` methods in `BingTextSearchService` as `KernelFunction` and name them as you did in line 27 and 29 - `StringSearch` and `PageSearch`. We will end up with `BingPlugin-StringSearch` and `BingPlugin-PageSearch` in kernel. If we share it with LLM, it will be able to decide which function to use based on user prompt. After specific function is invoked, it will return specific type, which will be possible to get with `result.GetValue<T>`.

Another way for point 3, we can keep `TextSearchPlugin` as non-generic, which will accept `ITextSearchService<string>` and/or `ITextSearchService<WebPage>` in constructor. Then, it will be possible to inject any service that implements `ITextSearchService<string>` and/or `ITextSearchService<WebPage>` (e.g. Bing or Google). In this case, you won't need to initialize `TextSearchPlugin` multiple times, DI can resolve all dependencies for you, there will be no need in type checking in search methods and complex type should be available as result of `InvokeAsync` method.

dotnet/src/Plugins/Plugins.Web/TextSearchPlugin.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System.ComponentModel;
+using System.Linq;
+using System.Text.Encodings.Web;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SemanticKernel.Search;
+
+namespace Microsoft.SemanticKernel.Plugins.Web;
```
rosieks, line 11: Does it make sense that it is in `Web` namespace? What about using non webpage related search services (eg. Azure AI Search?)

dotnet/samples/Concepts/TextSearch/TextSearchPluginExample.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Plugins;
+using Microsoft.SemanticKernel.Plugins.Web.Bing;
+
+namespace TextSearch;
+
+/// <summary>
+/// This example shows how to create and use a <see cref="TextSearchPlugin{T}"/>.
+/// </summary>
+public sealed class TextSearchPluginExample(ITestOutputHelper output) : BaseTest(output)
```
rosieks, line 11: Could you show how to use it with AzureAISearch - especially how to pass `IndexName`

dotnet/src/Plugins/Plugins.Memory/TextSearchPlugin.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <returns>A task that represents the asynchronous operation. The value of the TResult parameter contains the search results as a string.</returns>
+    [KernelFunction, Description("Perform a search for content related to the specified query.")]
+    public async Task<string> SearchAsync(
+        [Description("Search query")] string query,
+        [Description("Number of results")] int count = 1,
+        [Description("Number of results to skip")] int offset = 0,
+        CancellationToken cancellationToken = default)
+    {
+        var results = await this._service.SearchAsync<T>(query, new() { Count = count, Offset = offset }, null, cancellationToken).ConfigureAwait(false);
+        var resultList = await results.Results.ToListAsync(cancellationToken).ConfigureAwait(false);
```
rosieks, line 49: I think it will crash if used with `AzureAITextSearchService` and `T` is `string`. There is no way to provide `ValueField`

PR #6046, opened 2024-04-29
.Net - Agents Fix: Capture Function-Calling Content (ChatCompletionAgent)

crickman:
### Motivation and Context

Result from function-calling not being captured in agent-chat history for ChatCompletionAgent.

### Description

Updated logic to:

1. Capture mutated chat-history, in addition to the iterated messages.
2. Suppress exposing function-content as `AgentChat` interaction (while retaining in history).
3. "Tool" role transformed to "Assistant" for OpenAI Assistant API

### Contribution Checklist


dotnet/src/Agents/Core/ChatCompletionAgent.cs
```diff
@@ -34,14 +34,25 @@ public sealed class ChatCompletionAgent : ChatHistoryKernelAgent
...
+
+        IReadOnlyList<ChatMessageContent> messages =
             await chatCompletionService.GetChatMessageContentsAsync(
                 chat,
                 this.ExecutionSettings,
                 this.Kernel,
                 cancellationToken).ConfigureAwait(false);
 
-        foreach (var message in messages ?? [])
+        for (int messageIndex = messageCount; messageIndex < chat.Count; messageIndex++)
```
SergeyMenshykh, line 46: nit: messageIndex < messageCount

dotnet/src/Agents/Core/ChatCompletionAgent.cs
```diff
@@ -34,14 +34,25 @@ public sealed class ChatCompletionAgent : ChatHistoryKernelAgent
...
+        {
+            ChatMessageContent message = chat[messageIndex];
+
+            message.AuthorName = this.Name;
+
+            yield return message;
+        }
```
SergeyMenshykh, line 53: Won't returning the entire chat history by this method lead to duplicates in the chat history stored in the ChatHistoryChannel class?
image

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -179,6 +179,12 @@ await foreach (var message in channel.InvokeAsync(agent, cancellationToken).Conf
                 this.History.Add(message);
                 messages.Add(message);
 
+                if (message.Role == AuthorRole.Tool || message.Items.All(i => i is FunctionCallContent))
```
SergeyMenshykh, line 182: Is there a reason to filter out function call content and results? What I've noticed is that if both are excluded from a chat history and the chat history is sent to LLM again, the scenario that is possible when the whole chat is modeled as an agent, the LLM will call the functions again.

dotnet/src/Agents/Core/ChatCompletionAgent.cs
crickman: I believe only the incremental messages are added.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: This is only filtering out what is returned to the user.  The content is included in the history.

This is intended to align with core SK behavior

dotnet/src/Agents/Core/ChatCompletionAgent.cs
crickman: Updated

SergeyMenshykh: That's a wrong advise. Please rollback it if it's applied.

dotnet/src/Agents/Core/ChatCompletionAgent.cs
SergeyMenshykh: Right

dotnet/src/Agents/Core/ChatCompletionAgent.cs
crickman: Oh right...good catch.

crickman: Rolled-back.

SergeyMenshykh (approved):

alliscode (approved):

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -179,6 +179,12 @@ await foreach (var message in channel.InvokeAsync(agent, cancellationToken).Conf
...
+                {
+                    // Don't explore internal messages to caller.
```
markwallace-microsoft: Should it be 'Don't expose' ?

PR #6075, opened 2024-05-01
.Net - Agent Logging

crickman:
### Motivation and Context

Introduce logging for _Agent Framework_

### Description

Basic diagnostics are critical.
Added a single logging sample in _Getting Started_.  (Enabling logging for all samples obscures the sample focus)

Logging Levels:

- Pre-Action ([verb]-ing): DEBUG
- Post-Action ([verb]-ed): INFORMATION (This often contains the result of the action)
- Exception Thrown: ERROR (intuitive)
- Sensitive Data: TRACE  (Tao was explicit on this / I think from working with Toub)

Intent is that INFORMATION is for the messages we want enabled by default for a deployed service (with agents).  Whereas DEBUG/TRACE messages are more opt-in detail for developers, as you've suggested.

### Contribution Checklist


markwallace-microsoft (approved):

dotnet/src/Agents/Abstractions/Agent.cs
```diff
@@ -52,11 +53,12 @@ public abstract class Agent
...
     /// </summary>
+    /// <param name="logger">An agent specific logger.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>An <see cref="AgentChannel"/> appropriate for the agent type.</returns>
     /// <remarks>
     /// Every agent conversation, or <see cref="AgentChat"/>, will establish one or more <see cref="AgentChannel"/>
     /// objects according to the specific <see cref="Agent"/> type.
     /// </remarks>
-    protected internal abstract Task<AgentChannel> CreateChannelAsync(CancellationToken cancellationToken);
+    protected internal abstract Task<AgentChannel> CreateChannelAsync(ILogger logger, CancellationToken cancellationToken);
```
dmytrostruk, line 63: What is the reason of adding `ILogger` on method level instead of class level? In SK abstractions the pattern is to inject `ILogger` on class level, it would be nice to keep consistency here.

Also, if at some point you will decide to add some methods with default implementation to this abstract class or any derived classes (it looks like there is 4 level of inheritance), you will be able to easily re-use `ILogger` if it's on class level. With this implementation, you will be forced to add `ILogger` parameter to each method that requires logging. Since it's instrumentation abstraction, it will be required in most of the methods.

dotnet/src/Agents/Abstractions/AgentChannel.cs
```diff
@@ -2,15 +2,22 @@
...
 /// Defines the communication protocol for a particular <see cref="Agent"/> type.
-/// An agent provides it own <see cref="AgentChannel"/> via <see cref="Agent.CreateChannelAsync(CancellationToken)"/>.
+/// An agent provides it own <see cref="AgentChannel"/> via <see cref="Agent.CreateChannelAsync"/>.
 /// </summary>
 public abstract class AgentChannel
 {
+    /// <summary>
+    /// The <see cref="ILogger"/> associated with the <see cref="AgentChannel"/>.
+    /// </summary>
+    public ILogger Logger { get; set; } = NullLogger.Instance;
```
dmytrostruk, line 19: I like that it's on class level here, but wondering why not in `Agent.cs`.

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -24,13 +26,27 @@ public abstract class AgentChat
...
+    /// The <see cref="ILoggerFactory"/> associated with the <see cref="AgentChat"/>.
+    /// </summary>
+    public ILoggerFactory LoggerFactory { get; init; } = NullLoggerFactory.Instance;
+
+    /// <summary>
+    /// The <see cref="ILogger"/> associated with this chat.
+    /// </summary>
+    protected ILogger Logger =>
+        this._logger ??
+            Interlocked.CompareExchange(ref this._logger, this.LoggerFactory.CreateLogger(this.GetType()), null) ??
```
dmytrostruk, line 47: Is this required for `ILogger` instance? What problem it resolves?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -157,6 +178,8 @@ public void AddChatMessages(IReadOnlyList<ChatMessageContent> messages)
             // Note: Able to queue messages without synchronizing channels.
             var channelRefs = this._agentChannels.Select(kvp => new ChannelReference(kvp.Value, kvp.Key));
             this._broadcastQueue.Enqueue(channelRefs, messages);
+
+            this.Logger.LogInformation("[{MethodName}] Added Messages: {MessageCount}", nameof(AddChatMessages), messages.Count);
```
dmytrostruk, line 182: Similar log in line 170 has level `Debug` while this log has level `Information` but messages are almost the same. Should they have the same log level in this case?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -211,6 +238,8 @@ await foreach (ChatMessageContent message in channel.InvokeAsync(agent, cancella
                     .Where(kvp => kvp.Value != channel)
                     .Select(kvp => new ChannelReference(kvp.Value, kvp.Key));
             this._broadcastQueue.Enqueue(channelRefs, messages);
+
+            this.Logger.LogInformation("[{MethodName}] Invoked agent {AgentType}: {AgentId}", nameof(InvokeAgentAsync), agent.GetType(), agent.Id);
```
dmytrostruk, line 242: Same here, information about the beginning of operation (i.e. `invoking`) has level `Debug` while completion (i.e. `invoked`) has level `Information`. What is the reason of that?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -223,13 +252,19 @@ async Task<AgentChannel> GetOrCreateChannelAsync()
             AgentChannel channel = await this.SynchronizeChannelAsync(channelKey, cancellationToken).ConfigureAwait(false);
             if (channel == null)
             {
-                channel = await agent.CreateChannelAsync(cancellationToken).ConfigureAwait(false);
+                this.Logger.LogDebug("[{MethodName}] Creating channel for {AgentType}: {AgentId}", nameof(InvokeAgentAsync), agent.GetType(), agent.Id);
+
+                channel = await agent.CreateChannelAsync(this.LoggerFactory.CreateLogger(agent.GetType()), cancellationToken).ConfigureAwait(false);
+                channel.Logger = this.LoggerFactory.CreateLogger(channel.GetType());
```
dmytrostruk, line 258: Since you are passing `this.LoggerFactory.CreateLogger(agent.GetType())` to `CreateChannelAsync` method, why it's needed to set `channel.Logger` in the next line?

dotnet/src/Agents/Abstractions/AgentChat.cs
```diff
@@ -261,6 +296,8 @@ private void SetActivityOrThrow()
         int wasActive = Interlocked.CompareExchange(ref this._isActive, 1, 0);
         if (wasActive > 0)
         {
+            this.Logger.LogError("Unable to proceed while another agent is active.");
+
             throw new KernelException("Unable to proceed while another agent is active.");
```
dmytrostruk, line 301: It's better to create an instance of exception first, and then pass it to `LogError` and throw it later. Example:
```csharp
var exception = new KernelException("Unable to proceed while another agent is active.");
this.Logger.LogError(exception, exception.Message); // instead of exception.Message - custom message if needed.
throw exception;
```

If you pass `exception` to `LogError`, it may also capture exception stack trace, which is very useful during troubleshooting.

Same comment applies to other places where `LogError` is used.

dotnet/src/Agents/Abstractions/ChatHistoryKernelAgent.cs
```diff
@@ -17,13 +18,14 @@ protected internal sealed override IEnumerable<string> GetChannelKeys()
     }
 
     /// <inheritdoc/>
-    protected internal sealed override Task<AgentChannel> CreateChannelAsync(CancellationToken cancellationToken)
+    protected internal sealed override Task<AgentChannel> CreateChannelAsync(ILogger logger, CancellationToken cancellationToken)
     {
         return Task.FromResult<AgentChannel>(new ChatHistoryChannel());
     }
```
dmytrostruk, line 24: `ILogger` is added here and it's not used in a method. So, the idea of adding `ILogger` on class level should resolve this issue.

dotnet/src/Agents/Core/AgentGroupChat.cs
```diff
@@ -67,10 +71,25 @@ public async override IAsyncEnumerable<ChatMessageContent> InvokeAsync([Enumerat
...
+            this.Logger.LogDebug("[{MethodName}] Selecting agent: {StrategyType}", nameof(InvokeAsync), this.ExecutionSettings.SelectionStrategy.GetType());
+
+            Agent agent;
+            try
+            {
+                agent = await this.ExecutionSettings.SelectionStrategy.NextAsync(this.Agents, this.History, cancellationToken).ConfigureAwait(false);
+            }
+            catch
+            {
+                this.Logger.LogError("[{MethodName}] Unable to determine next agent.", nameof(InvokeAsync));
```
dmytrostruk, line 88: We can also pass `exception` instance to `LogError` here.

dotnet/src/Agents/Core/AgentGroupChat.cs
```diff
@@ -145,4 +172,17 @@ public AgentGroupChat(params Agent[] agents)
...
+    {
+        if (NullLogger.Instance == this.ExecutionSettings.SelectionStrategy.Logger)
+        {
+            this.ExecutionSettings.SelectionStrategy.Logger = this.LoggerFactory.CreateLogger(this.ExecutionSettings.SelectionStrategy.GetType());
+        }
+
+        if (NullLogger.Instance == this.ExecutionSettings.SelectionStrategy.Logger)
+        {
+            this.ExecutionSettings.TerminationStrategy.Logger = this.LoggerFactory.CreateLogger(this.ExecutionSettings.TerminationStrategy.GetType());
+        }
```
dmytrostruk, line 186: 1. It feels like `NullLogger.Instance` should be second operand here. We usually write `if (something == null)` and not `if (null == something)`.
2. Both conditions are the same. In line 183, there should be a check for `this.ExecutionSettings.TerminationStrategy.Logger`.
3. It feels like logger should be configured in all cases, even without checking if it's `NullLogger` or not.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Avoids a race condition for on-demand instantiation.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Line 170 is "Add**ing**"

All the PRE messages (-ING) are DEBUG.  All the POST messages (-ED) are INFORMATION.

crickman: image

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Same reason.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: One is "agent" typed.  The other is "channel" typed.  It translates into the capture diagnostics.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Will such an exception have a stack trace?

dotnet/src/Agents/Core/AgentGroupChat.cs
crickman: Its an common C++ convention.  I'll switch if its confusing.

It felt to me that re-invoking the factory may be sub-optimal perf as this will be most likely inside a loop.

dotnet/src/Agents/Abstractions/Agent.cs
crickman: Agent is really just a config object without any dynamism.  This is the only contract where agent may want to log something interesting (like the thread-id for OpenAIAssistant).  I initially didn't have this in the signature but it surfaced as part of design conversation.

Also, an agent may participate in multiple chats.  The entry-point for logger/logger-factor is the chat.  The agent isn't logically affinitized the the logger from any one chat...which is good because the agent really doesn't have much interesting to say / log.

dotnet/src/Agents/Abstractions/AgentChannel.cs
crickman: Channel is where all the dynamism is.  (Similiar to strategy.)  Agent doesn't have much interesting to say other than maybe when its creating a channel.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Kernel.cs#L133

dotnet/src/Agents/Abstractions/ChatHistoryKernelAgent.cs
crickman: Nothing interesting happening here.  Agents don't have much interesting so say / log.  (See previous comments).

Here where it is used:

https://github.com/microsoft/semantic-kernel/blob/feature_agent_logging/dotnet/src/Agents/OpenAI/OpenAIAssistantAgent.cs#L207

dotnet/src/Agents/Core/AgentGroupChat.cs
crickman: Updated condition order

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: The suggested pattern creates a warning that will fail the git-hub pipeline.  Maybe we can iteration on this.

I will say LogError does support being called with exception.  I've always used the overrides with exception parameter when _catching_ and exception.

dotnet/src/Agents/Core/AgentGroupChat.cs
crickman: Creates build warning / pipeline failure

dotnet/src/Agents/Abstractions/AgentChat.cs
dmytrostruk: > https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Kernel.cs#L133

Yes, I know that we have it in `Kernel.cs`, I was wondering if we should do it here as well for `ILogger` :)

dotnet/src/Agents/Abstractions/AgentChat.cs
dmytrostruk: > All the PRE messages (-ING) are DEBUG. All the POST messages (-ED) are INFORMATION.

And what is the reason of that? In ASP.NET, default log level is `Information`. So, by default, using agents I will see only `post` messages with results, but not when operation started. In order to see that, I will have to explicitly enable `Debug` level. Why can't we keep both messages at the same level (e.g. `Debug` or `Information`) and control them at the same time (i.e. turn them on or off)? 

Just want to understand the benefits of this approach. And whether it's a common practice across other libraries. As a developer, I would expect to see `Started/Finished` messages at the same log level. 

From `LogLevel` documentation:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=net-8.0#fields
image

If we compare `Debug` and `Information`, the difference would be whether the message is useful during debugging and whether it has or hasn't long-term value. I'm not sure if we can differentiate `pre/post` messages based on these criteria, unless I'm missing something.

dotnet/src/Agents/Abstractions/AgentChat.cs
dmytrostruk: > Will such an exception have a stack trace?

In this case no, but log message should contain exception type, which may be useful during debugging.

> The suggested pattern creates a warning that will fail the git-hub pipeline.

My fault, you shouldn't use variables/properties for `message` parameter:
```csharp
this.Logger.LogError(exception, "Unable to proceed while another agent is active.");
```

dotnet/src/Agents/Core/AgentGroupChat.cs
dmytrostruk: > Creates build warning / pipeline failure

Which exactly? I don't have any warnings locally:
image

dotnet/src/Agents/Core/AgentGroupChat.cs
dmytrostruk: > Both conditions are the same. In line 183, there should be a check for this.ExecutionSettings.TerminationStrategy.Logger.

This point is still unresolved:
image

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Would this be an impactful change?

dotnet/src/Agents/Core/AgentGroupChat.cs
crickman: Your original suggestion:

`this.Logger.LogError(exception, exception.Message);`

dotnet/src/Agents/Core/AgentGroupChat.cs
crickman: Copy paste typo.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Fixed

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Removed logging in alignment with our f2f call.

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Will add comments.

dotnet/src/Agents/Core/AgentGroupChat.cs
crickman: Updated

dotnet/src/Agents/Core/AgentGroupChat.cs
crickman: Updated

dotnet/src/Agents/Abstractions/AgentChat.cs
crickman: Updated

dmytrostruk (approved):

dmytrostruk (approved):

Issue #6026, opened 2024-04-29
.Net: Support for embedding dimensions


marcominerva:
**Azure.AI.OpenAI** NuGet package supports the `Dimensions` property for embeddings, to set the number of dimensions the resulting output embeddings should have. I see that the latest version of [AzureOpenAITextEmbeddingGenerationService](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.OpenAI/TextEmbedding/AzureOpenAITextEmbeddingGenerationService.cs) in **Microsoft.SemanticKernel.Connectors.OpenAI** does not provide a way to specify this value.

Is there any ETA about when will this feature be available?

stephentoub:
cc: @dmytrostruk

PR #6077, opened 2024-05-01
.Net: Added dimensions property to OpenAI embedding generation services

dmytrostruk:
### Motivation and Context

Resolves: https://github.com/microsoft/semantic-kernel/issues/6026

This PR contains changes to expose `dimensions` property which is supported by OpenAI and Azure .NET SDK:
https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-dimensions
image

### Contribution Checklist


dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
```diff
@@ -347,7 +348,8 @@ public static class OpenAIServiceCollectionExtensions
         string apiKey,
         string? serviceId = null,
         string? modelId = null,
-        HttpClient? httpClient = null)
+        HttpClient? httpClient = null,
+        int? dimensions = null)
```
stephentoub, line 352: Should dimensions be earlier in the argument list, e.g. before serviceId?

And do we think there might be other options in the future? Wondering if we should be considering some kind of options bag type.

dmytrostruk: > Should dimensions be earlier in the argument list, e.g. before serviceId?

Yes, and I like this approach better, but adding it as last parameter is less painful in terms of breaking changes, especially if users don't use named parameters.

> And do we think there might be other options in the future? Wondering if we should be considering some kind of options bag type.

With increasing number of arguments in these methods it would be better to rewrite these signatures and replace parameters with config models. In this case, it would be possible to easily add new parameters in the future, support optional parameters and don't care about parameters ordering. But I'm not sure if we want to do such change right now, and again, such refactoring will be breaking change.

We can use hybrid approach, leave parameters as they are today and add new model which will contain only new parameters, but then it will feel strange to pass big list of parameters and still initialize config class at the end.

RogerBarreto: > We can use hybrid approach, leave parameters as they are today and add new model which will contain only new parameters, but then it will feel strange to pass big list of parameters and still initialize config class at the end.

I like this option. 

Might consider [Obsolete] the parameterized one in favor of Options.

RogerBarreto:
Adding this parameter at the connector level seems to be an issue, we might consider having a ExecutionSettings for embeddings API... 

Any change on this matter I would add as experimental if the plan is to soon move this to be a setting per Invocation.

@westey-m

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
markwallace-microsoft: Added this to the backlog https://github.com/microsoft/semantic-kernel/issues/6081

dmytrostruk:
> Adding this parameter at the connector level seems to be an issue, we might consider having a ExecutionSettings for embeddings API...

@RogerBarreto I thought about that, but I also think that we can have both options supported. If you set vector dimensions once on connector level, you won't have to do that with each request. And I don't think that this parameter will change too much in runtime, because every time you change the dimension, your data may lose a context. 

`ExecutionSettings` for embedding API will be supported, here is ADR for this:
https://github.com/microsoft/semantic-kernel/pull/5277

> Any change on this matter I would add as experimental if the plan is to soon move this to be a setting per Invocation.

All embedding functionality is marked as `Experimental` at the moment.

dotnet/src/Connectors/Connectors.OpenAI/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,109 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
dmytrostruk: Yes.

RogerBarreto:
IF this is Experimental, suggest creating a new method with the dimension extra parameter as Experimental and avoid the Breaking Change.

.github/workflows/dotnet-build-and-test.yml
```diff
@@ -98,9 +98,9 @@ jobs:
           AzureOpenAI__DeploymentName: ${{ vars.AZUREOPENAI__DEPLOYMENTNAME }}
           AzureOpenAIEmbeddings__DeploymentName: ${{ vars.AZUREOPENAIEMBEDDING__DEPLOYMENTNAME }}
           AzureOpenAI__Endpoint: ${{ secrets.AZUREOPENAI__ENDPOINT }}
-          AzureOpenAIEmbeddings__Endpoint: ${{ secrets.AZUREOPENAI__ENDPOINT }}
+          AzureOpenAIEmbeddings__Endpoint: ${{ secrets.AZUREOPENAI_EASTUS__ENDPOINT }}
```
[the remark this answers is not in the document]
dmytrostruk: What is the source of this configuration? Because I can't find it in secrets/variables in GitHub settings. The reason why I switched this setting is because `text-embedding-3` models are supported in limited regions of Azure OpenAI resource, and `eastus` is one of them.

dmytrostruk:
> IF this is Experimental, suggest creating a new method marked as `Experimental` with the new dimension extra parameter to avoid the breaking change.

@RogerBarreto I thought about that as well :) But the thing is that there are 11 methods that were modified. In order to avoid breaking changes, I will need to add 11 more overload methods with new parameter. It's not a big problem, but since we agreed that at some point we will have to obsolete existing methods and replace them with `Config` classes, I'm not sure if we need to add new methods now in order to obsolete them later. 

Besides, this is breaking change only from API compatibility point of view, but not for compiler, that's the reason I added this parameter as last one. And since these are static methods, it's not possible to mock them in tests, so it should not break mocking scenarios.

dotnet/src/Connectors/Connectors.OpenAI/CompatibilitySuppressions.xml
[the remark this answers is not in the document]
dmytrostruk: Will be handled in scope of this task: https://github.com/microsoft/semantic-kernel/issues/6081

RogerBarreto (approved):

PR #6082, opened 2024-05-01
.Net: Add request uri and payload to RestApiOperationResponse

### LINKED ISSUE #6071: .Net: For OpenAPI plugins, the request URL and request payload should be included in the Metadata property of FunctionResult
For developers wanting to provide additional context about the result of a function call performed by an OpenAPI plugin, it would be helpful to provide the request information in the Metadata property bag.

markwallace-microsoft:
### Motivation and Context

Closes #6071 

### Description

Add a new `EnablePayloadInResponse` execution parameter which determines whether payload will be included in the `RestApiOperationResponse`. If true, the payload will be included in the response. Otherwise the payload will not be included and `RestApiOperationResponse.Payload` will be null. `RestApiOperationResponse.IncludesPayload` will be set to true if the payload is included in the response.

### Contribution Checklist


markwallace-microsoft:
Not planning to proceed with this option for now as there is a cost to duplicating the request data

dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiFunctionExecutionParameters.cs
```diff
@@ -54,6 +54,14 @@ public class OpenApiFunctionExecutionParameters
     /// </summary>
     public bool EnablePayloadNamespacing { get; set; }
 
+    /// <summary>
+    /// Determines whether payload will be included in the RestApiOperationResponse.
+    /// If true, the payload will be included in the response.
+    /// Otherwise the payload will not be included and RestApiOperationResponse.Payload will be null.
+    /// RestApiOperationResponse.IncludesPayload will be set to true if the payload is included in the response.
+    /// </summary>
+    public bool EnablePayloadInResponse { get; set; }
```
SergeyMenshykh, line 63: Is this property necessary? Why don't we always include the payload in the response? There are no expensive computations or allocations associated with it.

zengin:
@markwallace-microsoft can we add request method as well? There might be cases where URL is the same, but operation is different. That information might be useful in making sure that LLM didn't choose the wrong operation.

dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiFunctionExecutionParameters.cs
zengin: I am wondering if this could be renamed to something like "EnableRequestPayloadInResponse". Users may think this might be related to response payload. What do you think?

zengin (approved):

dotnet/src/Functions/Functions.OpenApi/Extensions/OpenApiFunctionExecutionParameters.cs
markwallace-microsoft: @stephentoub This PR is to address an ask that the request payload be included in the `RestApiOperationResponse` (which represents the response from a OpenAPI REST call). This is to support a specific use case where the application needs to display some debug information to end users. The proposed implementation adds a flag so that is behaviour is opt in so we don't include the request payload always as it may be large. Do you think this optimization is needed?

stephentoub: > Do you think this optimization is needed?

Do we already have the object that we'd be referencing from the response object, or do we not already have it and would need to otherwise create it?

Is the concern that someone might be keeping a result object alive for a long period of time not realizing it's rooting a large object?

My gut reaction is to not add such a property, and if it proves important, it could always be added later as opt-out.

dotnet/src/Functions/Functions.OpenApi/RestApiOperationRunner.cs
```diff
@@ -348,15 +357,15 @@ private JsonObject BuildJsonObject(IList<RestApiOperationPayloadProperty> proper
     /// </summary>
     /// <param name="payloadMetadata">The payload meta-data.</param>
     /// <param name="arguments">The payload arguments.</param>
-    /// <returns>The HttpContent representing the payload.</returns>
-    private HttpContent BuildPlainTextPayload(RestApiOperationPayload? payloadMetadata, IDictionary<string, object?> arguments)
+    /// <returns>The text payload and corresponding HttpContent.</returns>
+    private (object?, HttpContent) BuildPlainTextPayload(RestApiOperationPayload? payloadMetadata, IDictionary<string, object?> arguments)
```
RogerBarreto, line 361: nit.
Add named tuples to make it easier.

So instead of seeing Item1, Item2, the caller actually sees the provided names `Payload`, `Content` in the IDE.

```suggestion
    private (object? Payload, HttpContent Content) BuildPlainTextPayload(RestApiOperationPayload? payloadMetadata, IDictionary<string, object?> arguments)
```

RogerBarreto (approved):

RogerBarreto (approved):

Issue #5954, opened 2024-04-22
.Net: Specifying Function Calling Behavior in YAML


SergeyMenshykh:
It should be possible to specify function calling behavior in YAML files. This may require making the `FunctionCallBehavior` classes created within the scope of https://github.com/microsoft/semantic-kernel/issues/5253 {de}serialization friendly.

dmytrostruk:
bump

Issue #5253, opened 2024-03-01
.Net: Function Calling Abstractions


markwallace-microsoft:
```[tasklist]
### Tasks
- [x] Explore current function-calling model implementation & usage.
- [x] POC to understand the scope, feasibility, etc.
- [x] Share POC details to have every one of the same page and collect feedback.
- [x] Implement the abstraction taking into account the feedback for OpenAI connector only.
```

Krzysztof318:
I recommend looking at the implementation in gemini, it will make it easier to extract the abstraction, I tried to make it as similar as possible to openai's connector. https://github.com/microsoft/semantic-kernel/pull/5044

Issue #5953, opened 2024-04-22
.Net: Function Calling Abstraction for Gemini


SergeyMenshykh:
Use the function calling abstraction created is scope of this issue - https://github.com/microsoft/semantic-kernel/issues/5253 for the Gemini AI connector.

PR #6083, opened 2024-05-01
.Net: Function calling abstraction

SergeyMenshykh:
### Motivation and Context
Today, every AI connector in SK that supports function calling has its own implementation of tool call behavior model classes. These classes are used to configure the way connectors advertise and invoke functions. For example, the behavior classes can describe which functions should be advertised to LLM by a connector, whether the functions should be called automatically by the connector, or if the connector caller will invoke them himself manually, etc.

All the tool call behavior classes are the same in terms of describing the desired function call behavior. However, the classes have a mapping functionality to map the function call behavior to the connector-specific model classes, and that's what makes the function calling classes non-reusable between connectors. 

Additionally, today, it's not possible to specify function calling behavior declaratively in YAML, MD, and Kernel(config.json) prompts.

### Description

This PR introduces a few function choice behavior classes that allow configure function calling behavior in connector agnostic way:
- FunctionChoiceBehavior - an abstract class that used as a base class for all the choice behaviors described below. 
- AutoFunctionChoiceBehavior - inherits the FunctionChoiceBehavior class and contains configuration to tell LLM to decide itself whether to call kernel or provided function(s) or not.
- RequiredFunctionChoiceBehavior - inherits the FunctionChoiceBehavior class and contains configuration forcing LLM to call one or more kernel or provided functions.
- NoneFunctionChoiceBehavior - inherits the FunctionChoiceBehavior class and contains configuration to tell the model not to call functions and only generate a user-facing message.

Callers can specify one of the function choice behaviors using the new property, FunctionChoiceBehavior, which has been added to the PromptExecutionSettings class. Here's an example:
```csharp
PromptExecutionSettings settings = new PromptExecutionSettings() 
{ 
   FunctionChoiceBehavior = FunctionChoiceBehavior.Auto()
};

var result = await completionService.GetChatMessageContentAsync(chatHistory, settings, kernel);
```

The function choice behavior can be configured in JSON and YAML prompts as well:
```JSON
"execution_settings": {
    "service-gpt4": {
       "model_id": "gpt4",
       "temperature": 0.7,
       "function_choice_behavior": {
           "type": "auto",
           "functions": [
              "p1.f1"
           ]
        }
    }
}
```
```YAML
execution_settings:
  service-gpt4:
    model_id: gpt-4
    temperature: 1.0
    function_choice_behavior:
      type: auto
      functions:
      - p1.f1
```

Closes - https://github.com/microsoft/semantic-kernel/issues/5253, https://github.com/microsoft/semantic-kernel/issues/5954

ADR - https://github.com/microsoft/semantic-kernel/pull/6099

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/AutoFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,150 @@
...
+    /// </summary>
+    [JsonPropertyName("functions")]
+    public IEnumerable<string>? Functions
+    {
+        get => this._functionFQNs;
+        init
+        {
+            if (value?.Count() > 0 && this._functions?.Count() > 0)
+            {
+                throw new KernelException("Functions are already provided via the constructor.");
```
dmytrostruk, line 71: Can property just override what's in constructor? It's worth to provide some comments in code about this logic.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// <summary>
+    /// The default maximum number of function auto-invokes that can be made in a single user request.
+    /// </summary>
+    /// <remarks>
+    /// After this number of iterations as part of a single user request is reached, auto-invocation
+    /// will be disabled. This is a safeguard against possible runaway execution if the model routinely re-requests
+    /// the same function over and over.
+    /// </remarks>
+    protected const int DefaultMaximumAutoInvokeAttempts = 5;
```
dmytrostruk, line 30: I believe this should be configurable by user. I think there were some concerns about this variable being constant.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <inheritdoc/>
+    public override FunctionChoiceBehaviorConfiguration GetConfiguration(FunctionChoiceBehaviorContext context)
+    {
+        return new FunctionChoiceBehaviorConfiguration()
+        {
+            // By not providing either available or required functions, we are telling the model to not call any functions.
+            AvailableFunctions = null,
+            RequiredFunctions = null,
+        };
```
dmytrostruk, line 27: `public override FunctionChoiceBehaviorConfiguration GetConfiguration(FunctionChoiceBehaviorContext context) => new();`? Or you want to do that explicitly for better understanding?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,163 @@
...
+        // specify the kernel and the kernel must contain those functions.
+        if (autoInvoke && context.Kernel is null)
+        {
+            throw new KernelException("Auto-invocation for Required choice behavior is not supported when no kernel is provided.");
+        }
+
+        List<KernelFunction>? requiredFunctions = null;
+        bool allowAnyRequestedKernelFunction = false;
+
+        // Handle functions provided via constructor as function instances.
```
dmytrostruk, line 108: Logic in this method is very similar to logic in `AutoFunctionChoiceBehavior.GetConfiguration` method. Should we move it to the base class and re-use?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1327,9 +1304,9 @@ private void CaptureUsageDetails(CompletionsUsage usage)
     /// Processes the function result.
     /// </summary>
     /// <param name="functionResult">The result of the function call.</param>
-    /// <param name="toolCallBehavior">The ToolCallBehavior object containing optional settings like JsonSerializerOptions.TypeInfoResolver.</param>
+    /// <param name="promptExecutionSettings">The prompt execution settings.</param>
     /// <returns>A string representation of the function result.</returns>
-    private static string? ProcessFunctionResult(object functionResult, ToolCallBehavior? toolCallBehavior)
+    private static string? ProcessFunctionResult(object functionResult, OpenAIPromptExecutionSettings promptExecutionSettings)
```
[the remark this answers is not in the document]
SergeyMenshykh: There are no benefits at all. This code was changed, but later the change was partially rolled back, leaving the leftovers.

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAIFunctionsTests.cs
```diff
@@ -0,0 +1,587 @@
...
+        var invokedFunctions = new List<string>();
+
+#pragma warning disable CS0618 // Events are deprecated
+        void MyInvokingHandler(object? sender, FunctionInvokingEventArgs e)
+        {
+            invokedFunctions.Add(e.Function.Name);
+        }
+
+        kernel.FunctionInvoking += MyInvokingHandler;
+#pragma warning restore CS0618 // Events are deprecated
```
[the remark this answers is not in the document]
SergeyMenshykh: Thanks, fixed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1389,4 +1368,111 @@ private void CaptureUsageDetails(CompletionsUsage usage)
...
+    /// <param name="kernel">The <see cref="Kernel"/> to be used for function calling.</param>
+    /// <param name="executionSettings">Execution settings for the completion API.</param>
+    /// <param name="chatOptions">The chat completion options from the Azure.AI.OpenAI package.</param>
+    /// <param name="requestIndex">Request sequence index of automatic function invocation process.</param>
+    /// <returns>A tuple containing the AllowAnyRequestedKernelFunction and MaximumAutoInvokeAttempts settings.</returns>
+    private (bool? AllowAnyRequestedKernelFunction, int? MaximumAutoInvokeAttempts)? ConfigureFunctionCalling(Kernel? kernel, OpenAIPromptExecutionSettings executionSettings, ChatCompletionsOptions chatOptions, int requestIndex)
+    {
+        if (executionSettings.FunctionChoiceBehavior is not null && executionSettings.ToolCallBehavior is not null)
+        {
+            throw new ArgumentException("ToolBehaviors and ToolCallBehavior cannot be used together.");
```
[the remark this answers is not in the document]
SergeyMenshykh: > I also think that we should not throw an exception and use configuration based on priority. If both are specified, use FunctionChoiceBehavior as a new approach, which will overwrite ToolCallBehavior, we can mention this behavior in XML documentation.

Both solutions look good to me. The one that throws the exception notifies the caller immediately when the code is invoked, potentially saving time they would spend troubleshooting unexpected behavior. The one that relies on priority is more permissive and allows callers to use both solutions simultaneously, even at the cost of potential confusion or unexpected behavior. I would probably go with the proposed "priority" one taking into account its permissiveness. Let's see what others think.

> New approach also allows to specify configuration in json/yaml. So, if I have already existing working code with ToolCallBehavior, and I want to try new configuration in json/yaml, I will receive an exception at first try, because it will force me to remove ToolCallBehavior from my existing code. Choosing one property based on priority will require minimal changes from developers.

Today, as far as I understand, prompt execution settings provided via the code override the ones that come from prompts, so no exception will be thrown in this scenario. To allow execution settings specified in the prompt to be applied, either the settings specified in the code should be removed or a different service selector should be used.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1389,4 +1368,111 @@ private void CaptureUsageDetails(CompletionsUsage usage)
...
+
+                return result;
+            }
+
+            // If we have a required function, it means we want to force LLM to invoke that function.
+            if (config.RequiredFunctions is { } requiredFunctions && requiredFunctions.Any())
+            {
+                if (requiredFunctions.Count() > 1)
+                {
+                    throw new KernelException("Only one required function is allowed.");
```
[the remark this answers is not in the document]
SergeyMenshykh: > Is it a limitation in OpenAI only or the same is applicable in other connectors? 
- OpenAI initially supported only one required function, but recently added support for multiple functions - https://community.openai.com/t/new-api-feature-forcing-function-calling-via-tool-choice-required/731488. However, we cannot use it yet because the Azure.AI.OpenAI SDK has not exposed the new tool choice yet.  
- Mistral AI seems to support only one - https://docs.mistral.ai/capabilities/function_calling/#tool_choice  
- Gemini supports many - https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#tool-config

> It would be nice to see this limitation during compilation, rather than runtime, if possible of course.

It's an interesting problem to solve. Without going into the design, I see two options:  
- To have a set of choices that can be specified on a connector-specific prompt execution setting, so that the choices' usage can be enforced by the compiler.  
- To have a set of choices that can be supplied via the connector constructor and enforced by compiler as well.  
Both have their pros and cons and would require another iteration of design.
   
We can also consider other solutions to address the problem:  
* Use the first function from the list - this is not obvious and can be confusing for callers.  
* Execute the functions by the connector itself - this will solve the problem but will require a solution for obtaining arguments for the functions.

dotnet/src/Functions/Functions.Yaml/PromptExecutionSettingsTypeConverter.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        parser.MoveNext(); // Move past the MappingEnd event  
+        return executionSettings;
+    }
+
+    /// <inheritdoc/>
+    public void WriteYaml(IEmitter emitter, object? value, Type type)
+    {
+        throw new NotImplementedException();
+    }
```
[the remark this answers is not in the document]
SergeyMenshykh: We do import prompts but don't export them atm. I'm happy to add it if it's necessary.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
SergeyMenshykh: It's the default value that is used if no value provided for the `MaximumAutoInvokeAttempts` property. Users can easily configure the `MaximumAutoInvokeAttempts` either in the code or directly in prompts.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
SergeyMenshykh: Yes, the idea was to explicitly set them to null for clarity, even though null is the default value for both.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/AutoFunctionChoiceBehavior.cs
SergeyMenshykh: Today, the existing function-calling functionality supports **manual** invocation of kernel functions created via `KernelFunctionFactory.CreateFrom*` but non-imported into the kernel. Auto function invocation requires functions to be registered in the kernel, though.  
   
The new model partially follows the existing one and also supports non-imported into kernel functions for manual invocation. To register these non-imported functions, their **instances** should be passed as an argument to one of the choice behavior class constructors.  
   
Additionally, the new model supports kernel functions specified via prompts as a list of fully qualified function names - ["plugin1.function1", "plugin1.function2"]. The function list is assigned to the `Function` property during the deserialization process.  
   
Taking that context into account, there are a few options to consider:  
1. Stop supporting non-imported kernel functions. This way, all the functions passed to the new model class constructors will be converted to the list of fully qualified function names and assigned to the `Functions` property. The `GetConfiguration` method will verify if all the functions specified in the `Functions` property are imported into the kernel and throw an exception if any is non imported.  
2. Give preference to functions specified in the `Functions` property over those passed via the constructor. The `GetConfiguration` method will always use functions from the `Functions` property if available, and use the ones passed via the constructor otherwise.
3. Throw an exception if callers pass functions via constructor and via the `Functions` property. This is the implementation in this PR.

@stephentoub, are you aware of any use case that requires support for non-imported functions in the kernel?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,163 @@
...
+        else if (this.Functions is { } functionFQNs && functionFQNs.Any())
+        {
+            requiredFunctions = [];
+
+            foreach (var functionFQN in functionFQNs)
+            {
+                // Make sure that every function can be found in the kernel.
+                var name = FunctionName.Parse(functionFQN, FunctionNameSeparator);
+
+                if (!context.Kernel!.Plugins.TryGetFunction(name.PluginName, name.Name, out var function))
```
markwallace-microsoft, line 135: Why do we skip the autoInvoke check here but require it on line 112?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public const string TypeDiscriminator = "none";
+
+    /// <inheritdoc/>
+    public override FunctionChoiceBehaviorConfiguration GetConfiguration(FunctionChoiceBehaviorContext context)
+    {
+        return new FunctionChoiceBehaviorConfiguration()
+        {
+            // By not providing either available or required functions, we are telling the model to not call any functions.
+            AvailableFunctions = null,
```
markwallace-microsoft, line 25: We should allow passing functions to the model in this mode. In MistralAI and OpenAI it is valid to pass in a list of functions and set the tool choice to be none. It will respond with a message describing what function it would call and what parameters it would pass.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    /// <summary>
+    /// The functions that are available for the model to call.
+    /// </summary>
+    public IEnumerable<KernelFunction>? AvailableFunctions { get; init; }
+
+    /// <summary>
+    /// The functions that the model is required to call.
+    /// </summary>
+    public IEnumerable<KernelFunction>? RequiredFunctions { get; init; }
```
markwallace-microsoft, line 22: Do we ever set both of these?

Can we simplify and just provide the list of functions to be advertised to the LLM?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Represent function choice behavior configuration.
+/// </summary>
+[Experimental("SKEXP0001")]
+public sealed class FunctionChoiceBehaviorConfiguration
```
markwallace-microsoft, line 12: Should this contain an abstraction for the tool choice property?

How can I use tool choice none and still pass a list of functions?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,73 @@
...
+/// </summary>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "type")]
+[JsonDerivedType(typeof(AutoFunctionChoiceBehavior), typeDiscriminator: AutoFunctionChoiceBehavior.TypeDiscriminator)]
+[JsonDerivedType(typeof(RequiredFunctionChoiceBehavior), typeDiscriminator: RequiredFunctionChoiceBehavior.TypeDiscriminator)]
+[JsonDerivedType(typeof(NoneFunctionChoiceBehavior), typeDiscriminator: NoneFunctionChoiceBehavior.TypeDiscriminator)]
+[Experimental("SKEXP0001")]
+public abstract class FunctionChoiceBehavior
+{
+    /// <summary>The separator used to separate plugin name and function name.</summary>
+    protected const string FunctionNameSeparator = ".";
```
markwallace-microsoft, line 20: I think this is currently a `-`, why is this being changed?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
SergeyMenshykh: If the function is specified as a function FQN, then it must be registered in the kernel regardless of the invocation type. How and where else could the connector obtain the function metadata to advertise?  
   
On the other hand, if a function is specified as an instance, the requirement of being registered in the kernel only applies to functions in auto-invocation mode, as the connector will need to look it up in the kernel to auto-invoke it. For manual mode, the connector delegates function resolution to the caller that registered the function for invocation and possesses its instance as such.

I've added the comment to make it clearer.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
SergeyMenshykh: > Do we ever set both of these?

We set either one or another, but never both.

> Can we simplify and just provide the list of functions to be advertised to the LLM?

The list of the functions alone might not be enough. We would also need to pass the choice - auto, required, or none to connectors so they can map it to their model choice.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
SergeyMenshykh: > Should this contain an abstraction for the tool choice property?

The idea was to allow connectors use the choice behavior classes polymorphically via the parent class - `FunctionChoiceBehavior` so the connectors would not even be aware which specific choice behavior classes they are working with. Adding the property would go against the polymorphic nature of the solution and that can be fine if the solution doesn't allow us to achieve what we want. Or you are thinking putting an abstraction over the choice property?

> How can I use tool choice none and still pass a list of functions?

Advertising functions to model for the "none" choice were not originally planned. None of the documents I saw imply passing both:
* OpenAI - https://platform.openai.com/docs/guides/function-calling
   image
* Gemini - https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling
   image
* Mistral - https://docs.mistral.ai/capabilities/function_calling/
   image

However, I recognize that the model, ideally, should support it in case a caller wants to supply both.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
SergeyMenshykh: The "." character seems to be used by all SK prompt templates as a separator between the plugin name and function name. So, in this case, it is used for consistency within the same prompt - the same separator for functions inside the prompt template and function choice behavior. To avoid confusion when the `Functions` collection is used in the code, we can either exclude it from the public surface, allowing functions to be supplied only via the constructor but allowing its deserialization, or have a converter that changes "." to "-" separators during deserialization.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RequiredFunctionChoiceBehavior"/> class.
+    /// </summary>
+    /// <param name="functions">The subset of the <see cref="Kernel"/>'s plugins' functions information.</param>
+    public RequiredFunctionChoiceBehavior(IEnumerable<KernelFunction> functions)
+    {
+        this._functions = functions;
+    }
```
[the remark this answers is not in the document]
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
SergeyMenshykh: Fixed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/AutoFunctionChoiceBehavior.cs
SergeyMenshykh: Fixed, by initializing the property from functions provided via the class constructor.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
SergeyMenshykh: done

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,104 @@
...
+/// <summary>
+/// Represents <see cref="FunctionChoiceBehavior"/> that does not provides any <see cref="Kernel"/>'s plugins' function information to the model by default.
+/// This behavior forces the model to not call any functions and only generate a user-facing message.
+/// </summary>
+/// <remarks>
+/// Although this behavior prevents the model from calling any functions, the model can use the provided function information
+/// to describe how it would complete the prompt if it had the ability to call the functions.
+/// </remarks>
+[Experimental("SKEXP0001")]
+public sealed class NoneFunctionChoiceBehavior : FunctionChoiceBehavior
```
stephentoub, line 19: Can this be internal?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,104 @@
...
+
+                // Check if a function instance that was not imported into the kernel was provided through the constructor.
+                function = this._functions?.FirstOrDefault(f => f.Name == nameParts.Name && f.PluginName == nameParts.PluginName);
+                if (function is not null)
+                {
+                    availableFunctions.Add(function);
+                    continue;
+                }
+
+                throw new KernelException($"The specified function {functionFQN} is not available.");
```
stephentoub, line 85: For None, does it matter? Nothing needs to be invoked.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Represent <see cref="FunctionChoiceBehavior"/> that provides either all of the <see cref="Kernel"/>'s plugins' function information to the model or a specified subset.
+/// This behavior forces the model to always call one or more functions. The model will then select which function(s) to call.
+/// </summary>
+[Experimental("SKEXP0001")]
+public sealed class RequiredFunctionChoiceBehavior : FunctionChoiceBehavior
```
stephentoub, line 15: Internal?

dotnet/src/SemanticKernel.UnitTests/PromptTemplate/PromptTemplateConfigTests.cs
```diff
@@ -141,6 +142,108 @@ public void DeserializingExpectCompletion()
...
+        // Assert
+        Assert.NotNull(promptTemplateConfig);
+        Assert.Single(promptTemplateConfig.ExecutionSettings);
+
+        var executionSettings = promptTemplateConfig.ExecutionSettings.Single().Value;
+
+        var noneFunctionCallChoice = executionSettings.FunctionChoiceBehavior as NoneFunctionChoiceBehavior;
+        Assert.NotNull(noneFunctionCallChoice);
+    }
+
```
stephentoub, line 246: I assume you'll be updating the PR to handle the abstraction on at least Mistral and Gemini as well?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/AutoFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Represent <see cref="FunctionChoiceBehavior"/> that provides either all of the <see cref="Kernel"/>'s plugins' function information to the model or a specified subset.
+/// This behavior allows the model to decide whether to call the functions and, if so, which ones to call.
+/// </summary>
+[Experimental("SKEXP0001")]
+public sealed class AutoFunctionChoiceBehavior : FunctionChoiceBehavior
```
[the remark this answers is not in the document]
SergeyMenshykh: There are two reasons, both related to the deserialization aspect of the function choice behavior classes:

**1. Polymorphic deserialization:**
The function choice behavior classes are deserialized from JSON and YAML formats polymorphically. As such, deserialization API need to know about their types to be able to create their instances.  System.Text.Json.JsonSerializer can get the information from `JsonDerivedType` attribute the `FunctionChoiceBehavior` base class is annotated with. While the `YamlDotNet.Serialization.IDeserializer` implementation doesn't know anything about those attributes and requires a custom registration of the classes on the destabilizer - [WithTypeDiscriminatingNodeDeserializer](https://github.com/microsoft/semantic-kernel/blob/a3527d7aadb4ad75db437da8f6a3da3041a95ef4/dotnet/src/Functions/Functions.Yaml/PromptExecutionSettingsTypeConverter.cs#L31). The YAML prompts deserialization logic resides in the `Functions.Yaml` project and needs the function choice behavior classes that live in the `SemanticKernel.Abstractions` project to be public to access them.

A solution here could be either:
**a.** Not to use polymorphic deserialization 
**b.** YAML deserialization functionality can read the function choice behavior types from the `JsonDerivedType` attribute the `FunctionChoiceBehavior` base class is annotated with using the `System.Reflection.GetCustomAttribute` method and cache it.

**2. Deserialization of function names:** SK prompt templates can call functions specified in the templates using syntax that uses '.' as a separator between plugin and function names. While configuring function choice behavior in the same template, it makes sense to use the same separator. In the example below, both function references use '.' as a separator:  
   
```yaml  
template: Say hello world in {{languageUtils.getLanguage}} to the current user  
execution_settings:  
  service1:  
    temperature: 1.0  
    function_choice_behavior:  
      type: auto  
      functions:  
        - graph.getCurrentUser  
```  
   
However, there could potentially be custom prompt templates that use '_', '+', '-', etc., as a separator in both the prompt template and function choice behavior. 

The solution here could be either:  
**a.** Normalize different separators to the well-known one '-' at the deserialization phase. This solution is used in this PR and requires access to the `Functions` collection of the function choice behavior classes that need to be public so YAML type converters can access them to do the separator normalization.  
**b.** Add a new `FunctionNameSeparator` property to the `FunctionChoiceBehavior` class so the separator can be specified in the prompt. This way, no normalization is required because the separator will be set explicitly if it differs from the default prompt one, '.'.

I can try to prototype the 1.b & 2.b solutions to make the function choice behavior classes internal.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/AutoFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// <summary>
+    /// Indicates whether the functions should be automatically invoked by the AI service/connector.
+    /// </summary>
+    private readonly bool _autoInvoke = true;
+
+    /// <summary>
+    /// This class type discriminator used for polymorphic deserialization of the type specified in JSON and YAML prompts.
+    /// </summary>
+    public const string TypeDiscriminator = "auto";
```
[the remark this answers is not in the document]
SergeyMenshykh: One of the classes - [PromptExecutionSettingsTypeConverter](https://github.com/microsoft/semantic-kernel/blob/a3527d7aadb4ad75db437da8f6a3da3041a95ef4/dotnet/src/Functions/Functions.Yaml/PromptExecutionSettingsTypeConverter.cs#L31) that uses the constant lives in the `Functions.Yaml` project. But the `AutoFunctionChoiceBehavior` lives in `SemanticKernel.Abstractions` project. The const need to be public to allow the type converter to access it unless we want to go the `InternalsVisibleTo` way.

Alternatively, instead of getting the descriminator from the public const, it can be read from the `JsonPolymorphic` attribute the `FunctionChoiceBehavior` base class is annotated with using the System.Reflection.GetCustomAttribute method.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1544,4 +1513,154 @@ private void LogUsage(CompletionsUsage usage)
...
+            }
+
+            return result;
+        }
+
+        if (config.Choice == FunctionChoice.Required)
+        {
+            if (config.Functions is { } functions && functions.Any())
+            {
+                if (functions.Count() > 1)
```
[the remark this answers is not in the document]
SergeyMenshykh: The `FunctionChoiceBehaviorConfiguration.Functions` property is declared as public `IEnumerable<KernelFunction>? Functions { get; init; }`. Its type can be easily changed to IList or IReadOnlyList if necessary. Is the Count property more efficient than the Count() method? Or will it make the `if (config.Functions is { } functions && functions.Any())` statement more elegant, like `if (config.Functions is { Count: > 0 } functions)`?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorContext.cs
```diff
@@ -0,0 +1,17 @@
...
+/// <summary>
+/// The context to be provided by the choice behavior consumer in order to obtain the choice behavior configuration.
+/// </summary>
+[Experimental("SKEXP0001")]
+public class FunctionChoiceBehaviorContext
+{
+    /// <summary>
+    /// The <see cref="Kernel"/> to be used for function calling.
+    /// </summary>
+    public Kernel? Kernel { get; init; }
```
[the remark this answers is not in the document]
SergeyMenshykh: The reason is that no scenario has been identified that would need to reassign the property after it was initialized by a connector. It's no big deal to change it to a setter; I just would like to understand scenarios when it's useful or more optimal.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
SergeyMenshykh: Yep, it can as explained in the https://github.com/microsoft/semantic-kernel/pull/6083#discussion_r1622840916 comment.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
SergeyMenshykh: The code needs to be changed before it can be done. More details can be found in the https://github.com/microsoft/semantic-kernel/pull/6083#discussion_r1622840916 comment.

dotnet/src/SemanticKernel.UnitTests/PromptTemplate/PromptTemplateConfigTests.cs
SergeyMenshykh: I planned to update both in separate PRs to minimize the PR scope.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,62 @@
...
+public abstract class FunctionChoiceBehavior
+{
+    /// <summary>
+    /// Gets an instance of the <see cref="FunctionChoiceBehavior"/> that provides either all of the <see cref="Kernel"/>'s plugins' function information to the model or a specified subset.
+    /// This behavior allows the model to decide whether to call the functions and, if so, which ones to call.
+    /// </summary>
+    /// <param name="autoInvoke">Indicates whether the functions should be automatically invoked by the AI service/connector.</param>
+    /// <param name="functions">The subset of the <see cref="Kernel"/>'s plugins' function information.</param>
+    /// <returns>An instance of one of the <see cref="FunctionChoiceBehavior"/>.</returns>
+    public static FunctionChoiceBehavior AutoFunctionChoice(bool autoInvoke = true, IEnumerable<KernelFunction>? functions = null)
```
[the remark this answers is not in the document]
SergeyMenshykh: Yes, those definitely can be shortened.
CC: @markwallace-microsoft

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,62 @@
...
+
+/// <summary>
+/// Represents the base class for different function choice behaviors.
+/// </summary>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "type")]
+[JsonDerivedType(typeof(AutoFunctionChoiceBehavior), typeDiscriminator: AutoFunctionChoiceBehavior.TypeDiscriminator)]
+[JsonDerivedType(typeof(RequiredFunctionChoiceBehavior), typeDiscriminator: RequiredFunctionChoiceBehavior.TypeDiscriminator)]
+[JsonDerivedType(typeof(NoneFunctionChoiceBehavior), typeDiscriminator: NoneFunctionChoiceBehavior.TypeDiscriminator)]
+[Experimental("SKEXP0001")]
+public abstract class FunctionChoiceBehavior
```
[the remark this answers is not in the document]
SergeyMenshykh: Done

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,62 @@
...
+    public static FunctionChoiceBehavior AutoFunctionChoice(bool autoInvoke = true, IEnumerable<KernelFunction>? functions = null)
+    {
+        return new AutoFunctionChoiceBehavior(autoInvoke, functions);
+    }
+
+    /// <summary>
+    /// Gets an instance of the <see cref="FunctionChoiceBehavior"/> that provides either all of the <see cref="Kernel"/>'s plugins' function information to the model or a specified subset.
+    /// This behavior forces the model to always call one or more functions. The model will then select which function(s) to call.
+    /// </summary>
+    /// <param name="autoInvoke">Indicates whether the functions should be automatically invoked by the AI service/connector.</param>
```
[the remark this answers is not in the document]
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
SergeyMenshykh: Yes, right, nothing will be invoked. However, if a subset of functions is specified, the caller wants them to be advertised to the model. If any is not found, it means it does not exist because it's either not registered in the kernel or the function name is incorrect. In either case, it seems right to inform the caller about it, doesn't it?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    /// <param name="functions">The subset of the <see cref="Kernel"/>'s plugins' function information.</param>
+    /// <returns>An instance of one of the <see cref="FunctionChoiceBehavior"/>.</returns>
+    /// <remarks>
+    /// Although this behavior prevents the model from calling any functions, the model can use the provided function information
+    /// to describe how it would complete the prompt if it had the ability to call the functions.
+    /// </remarks>
+    public static FunctionChoiceBehavior NoneFunctionChoice(IEnumerable<KernelFunction>? functions = null)
+    {
+        return new NoneFunctionChoiceBehavior(functions ?? []);
```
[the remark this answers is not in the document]
SergeyMenshykh: All three function choice behavior classes are similar in the way they advertise the functions. All three advertise all kernel functions if the caller doesn't explicitly set a subset either through the constructor or the `Functions` property. If a subset is specified, then only the functions from the subset are advertised.

It's not clear from the comment what led to that conclusion. Is it the method's XML comment, the method/class name, or something else?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1544,4 +1513,154 @@ private void LogUsage(CompletionsUsage usage)
...
+        // Handling old-style tool call behavior represented by `OpenAIPromptExecutionSettings.ToolCallBehavior` property.
+        else if (executionSettings.ToolCallBehavior is { } toolCallBehavior)
+        {
+            result = this.ConfigureFunctionCalling(kernel, chatOptions, requestIndex, toolCallBehavior);
+        }
+
+        // Having already sent tools and with tool call information in history, the service can become unhappy ("[] is too short - 'tools'")
+        // if we don't send any tools in subsequent requests, even if we say not to use any.
+        // Similarly, if we say not to use any tool (ToolChoice = ChatCompletionsToolChoice.None) and dont provide any for the first request,
+        // the service fails with "'tool_choice' is only allowed when 'tools' are specified."
```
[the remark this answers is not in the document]
SergeyMenshykh: Yes, it is still the case; the service fails with the error:
image
I've tried to assign null to the `ToolChoice` property because it's null by default, even though it's not nullable, but it hasn't changed anything.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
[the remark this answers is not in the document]
SergeyMenshykh: All three static factory methods have been renamed.

dotnet/src/SemanticKernel.UnitTests/PromptTemplate/PromptTemplateConfigTests.cs
SergeyMenshykh: The abstraction will be applied to Gemini and mistral in scope of the issues:
- https://github.com/microsoft/semantic-kernel/issues/6555
- https://github.com/microsoft/semantic-kernel/issues/5953

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// This behavior forces the model to not call any functions and only generate a user-facing message.
+    /// </summary>
+    /// <param name="functions">The subset of the <see cref="Kernel"/>'s plugins' functions to provide to the model.
+    /// If null or empty, all <see cref="Kernel"/>'s plugins' functions are provided to the model.</param>
+    /// <returns>An instance of one of the <see cref="FunctionChoiceBehavior"/>.</returns>
+    /// <remarks>
+    /// Although this behavior prevents the model from calling any functions, the model can use the provided function information
+    /// to describe how it would complete the prompt if it had the ability to call the functions.
+    /// </remarks>
+    public static FunctionChoiceBehavior None(IEnumerable<KernelFunction>? functions = null)
```
RogerBarreto, line 63: nit.

```suggestion
    public static FunctionChoiceBehavior None(IEnumerable<KernelFunction>? specificFunctions = null)
```

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return new RequiredFunctionChoiceBehavior(autoInvoke, functions);
+    }
+
+    /// <summary>
+    /// Gets an instance of the <see cref="FunctionChoiceBehavior"/> that provides either all of the <see cref="Kernel"/>'s plugins' function information to the model or a specified subset.
+    /// This behavior forces the model to not call any functions and only generate a user-facing message.
+    /// </summary>
+    /// <param name="functions">The subset of the <see cref="Kernel"/>'s plugins' functions to provide to the model.
+    /// If null or empty, all <see cref="Kernel"/>'s plugins' functions are provided to the model.</param>
```
RogerBarreto, line 57: If empty I would throw `ArgumentException`.

```suggestion
    /// If not provided, defaults to all plugins functions currently added to the <see cref="Kernel"/> instance.</param>
```

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// Although this behavior prevents the model from calling any functions, the model can use the provided function information
+    /// to describe how it would complete the prompt if it had the ability to call the functions.
+    /// </remarks>
+    public static FunctionChoiceBehavior None(IEnumerable<KernelFunction>? functions = null)
+    {
+        return new NoneFunctionChoiceBehavior(functions ?? []);
+    }
+
+    /// <summary>Returns the configuration specified by the <see cref="FunctionChoiceBehavior"/>.</summary>
+    /// <param name="context">The caller context.</param>
```
RogerBarreto, line 69:
```suggestion
    /// <param name="context">The function choice caller context.</param>
```

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    /// <summary>
+    /// Creates a new instance of the <see cref="FunctionChoiceBehaviorConfiguration"/> class.
+    /// </summary>
+    internal FunctionChoiceBehaviorConfiguration()
+    {
+    }
+
+    /// <summary>
+    /// Represents an AI model's decision-making strategy for calling functions.
```
RogerBarreto, line 22:
```suggestion
    /// Represents an AI model's decision-making choice strategy for calling functions.
```

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public FunctionChoice Choice { get; internal set; }
+
+    /// <summary>
+    /// The functions available for AI model.
+    /// </summary>
+    public IEnumerable<KernelFunction>? Functions { get; internal set; }
```
RogerBarreto, line 29: We may want to expose `Count` property instead of enforcing iterations over the functions.

```suggestion
    public ICollection<KernelFunction>? Functions { get; internal set; }
```

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorContext.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// The context to be provided by the choice behavior consumer in order to obtain the choice behavior configuration.
+/// </summary>
+[Experimental("SKEXP0001")]
+public class FunctionChoiceBehaviorContext
```
RogerBarreto, line 11: Do we intent to have specializations of the context?

```suggestion
public sealed class FunctionChoiceBehaviorContext
```

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionNameFormatJsonConverter.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// A custom JSON converter for converting function names in a JSON array.
+/// This converter replaces dots used as a function name separator in prompts with hyphens when reading and back when writing.
+/// </summary>
+[Experimental("SKEXP0001")]
+public sealed class FunctionNameFormatJsonConverter : JsonConverter<IList<string>>
```
RogerBarreto, line 16: nit

Some providers may use other separators, not sure where this will be used, just by our abstractions?

We might consider passing what is the separator as an init ?

```suggestion
public sealed class FunctionNameFormatJsonConverter : JsonConverter<IList<string>>
```

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,106 @@
...
```
RogerBarreto, line 87: Syntax thing: We will never know if the function is available or not, but we surely knows we didn't find it. 

```suggestion
                throw new KernelException($"The specified function {functionFQN} was not found.");
```

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -93,7 +135,8 @@ public virtual PromptExecutionSettings Clone()
         return new()
         {
             ModelId = this.ModelId,
-            ExtensionData = this.ExtensionData is not null ? new Dictionary<string, object>(this.ExtensionData) : null
+            ExtensionData = this.ExtensionData is not null ? new Dictionary<string, object>(this.ExtensionData) : null,
+            FunctionChoiceBehavior = this.FunctionChoiceBehavior
```
RogerBarreto, line 139: nit. Properties first.

```suggestion
            FunctionChoiceBehavior = this.FunctionChoiceBehavior,
            ExtensionData = this.ExtensionData is not null ? new Dictionary<string, object>(this.ExtensionData) : null,
```

RogerBarreto (approved):

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/AutoFunctionChoiceBehavior.cs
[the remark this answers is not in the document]
SergeyMenshykh: Fixed by implementing solution 1.b and using '.' as the default separator for plugin and function names. If, later on, there is a need to support a different separator for a new prompt template format, it can be achieved either by solution 2.b or by custom deserialization of the new prompt template.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/AutoFunctionChoiceBehavior.cs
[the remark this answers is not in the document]
SergeyMenshykh: removed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
SergeyMenshykh: done

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Represent function choice behavior configuration.
+/// </summary>
+[Experimental("SKEXP0001")]
+public sealed class FunctionChoiceBehaviorConfiguration
```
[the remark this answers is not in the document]
SergeyMenshykh: The "ToolCallBehavior : FunctionChoiceBehavior" commit is rolled back, and the internal constructor with internal setters are added.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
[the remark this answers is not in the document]
SergeyMenshykh: The "ToolCallBehavior : FunctionChoiceBehavior" commit is rolled back, and the internal constructor is added.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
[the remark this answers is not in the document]
SergeyMenshykh: The type is changed to IReadOnlyList so the Count property is used instead.

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/NoneFunctionChoiceBehavior.cs
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionNameFormatJsonConverter.cs
SergeyMenshykh: The converted has been removed.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoice.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// <inheritdoc/>
+    public bool Equals(FunctionChoice other)
+        => string.Equals(this.Label, other.Label, StringComparison.OrdinalIgnoreCase);
+
+    /// <inheritdoc/>
+    public override int GetHashCode()
+        => StringComparer.OrdinalIgnoreCase.GetHashCode(this.Label ?? string.Empty);
+
+    /// <inheritdoc/>
+    public override string ToString() => this.Label ?? string.Empty;
```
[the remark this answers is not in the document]
SergeyMenshykh: Yep, thanks, fixed!

dotnet/src/Functions/Functions.Yaml/FunctionChoiceBehaviorTypesConverter.cs
```diff
@@ -0,0 +1,84 @@
...
+    private const char PromptFunctionNameSeparator = '.';
+
+    private const char FunctionNameSeparator = '-';
+
+    private static IDeserializer? s_deserializer;
+
+    /// <inheritdoc/>
+    public bool Accepts(Type type)
+    {
+#pragma warning disable SKEXP0001
```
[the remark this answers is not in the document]
SergeyMenshykh: The converted has been removed.

dotnet/src/Functions/Functions.Yaml/FunctionChoiceBehaviorTypesConverter.cs
```diff
@@ -0,0 +1,84 @@
...
+using YamlDotNet.Core;
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Allows custom deserialization for derivatives of <see cref="FunctionChoiceBehavior"/>.
+/// </summary>
+internal sealed class FunctionChoiceBehaviorTypesConverter : IYamlTypeConverter
```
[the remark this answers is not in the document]
SergeyMenshykh: The converted has been removed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1441,4 +1390,128 @@ private void CaptureUsageDetails(CompletionsUsage usage)
...
     }
+
+    /// <summary>
+    /// Configures the function calling functionality based on the provided parameters.
+    /// </summary>
+    /// <param name="kernel">The <see cref="Kernel"/> to be used for function calling.</param>
+    /// <param name="executionSettings">Execution settings for the completion API.</param>
+    /// <param name="chatOptions">The chat completion options from the Azure.AI.OpenAI package.</param>
+    /// <param name="requestIndex">Request sequence index of automatic function invocation process.</param>
+    private (bool? AllowAnyRequestedKernelFunction, int? MaximumAutoInvokeAttempts)? ConfigureFunctionCalling(Kernel? kernel, OpenAIPromptExecutionSettings executionSettings, ChatCompletionsOptions chatOptions, int requestIndex)
```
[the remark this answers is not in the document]
SergeyMenshykh: Parameters order has been changed. Regarding the method name, I would say that the responsibility of the method is to configure function-calling functionality by setting the `chatOptions.Tools` and `chatOptions.ToolChoice` properties based on the configuration. Additionally, it returns a few properties from the configuration. So, the name reflects the method's responsibility: configuration. We can also reflect the fact that it returns configuration by naming it `ConfigureFunctionCallingAndGetConfig`. Let me know what you think.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
SergeyMenshykh: Wouldn't it be too specific to add the "specific" prefix to the parameter name? :) How would the "specific" prefix help the caller to better reason about the functions that need to be passed as an argument? Isn't the fact that the caller put certain functions in the array to supply as an argument already making the functions in the array specific? I don't mind renaming it, just want to understand the benefits of it.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehaviorConfiguration.cs
SergeyMenshykh: It seems the word "choice" is redundant in this context. The term "decision-making strategy" already conveys the idea of making choices or decisions?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
SergeyMenshykh: Good point, we can do it. Alternatively, we may treat empty array as a way to turn off function calling:
- The `FunctionChoiceBehavior.Auto(functions: listOfFuncitons)` call would advertise only functions from listOfFuncitons.
- The `FunctionChoiceBehavior.Auto(functions: null)` call would advertise all functions registered on kernel.
- The `FunctionChoiceBehavior.Auto(functions: [])` call would advertise no functions and can be used to disable function calling.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/RequiredFunctionChoiceBehavior.cs
SergeyMenshykh: Fixed.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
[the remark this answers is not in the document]
SergeyMenshykh: Fixed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1571,4 +1540,154 @@ private void LogUsage(CompletionsUsage usage)
...
+        // what functions are available in the kernel.
+        var config = functionChoiceBehavior.GetConfiguration(new() { Kernel = kernel });
+
+        (bool? AllowAnyRequestedKernelFunction, int? MaximumAutoInvokeAttempts) result = new()
+        {
+            AllowAnyRequestedKernelFunction = config.AllowAnyRequestedKernelFunction,
+            MaximumAutoInvokeAttempts = config.Options.AutoInvoke ? MaximumAutoInvokeAttempts : 0,
+        };
+
+        if (config.Choice == FunctionChoice.Required && requestIndex >= MaximumUseAttempts)
```
moonbox3, line 1606: If defining the FunctionChoiceBehavior as `required`, the OpenAI model will continue to run tools until the max auto invoke attempts are hit. How are you handling this, where the function result was returned to the model and tool should not be called again? Is the idea that the max auto invoke attempts should be 1 for a tool choice configured as `required`?

moonbox3: As I am testing the `required` tool choice, no matter how I configure the max auto invoke attempts, I only get back a tool call as part of the Function Result. Versus `auto` where the model responds with a tool call, we send the tool call response and the model responds with the final answer. I'm curious to learn about how you're testing the `required` tool choice. Thanks.

SergeyMenshykh: > If defining the FunctionChoiceBehavior as required, the OpenAI model will continue to run tools until the max auto invoke attempts are hit. How are you handling this, where the function result was returned to the model and tool should not be called again?

The 'if' statement on line 1606 ensures that the method exits before executing the code that advertises the function to the model, thanks to the 'return' statement. As a result, the required choice behavior advertises the function only once during the first request and stops doing so for subsequent requests.

> Is the idea that the max auto invoke attempts should be 1 for a tool choice configured as required?

In addition to the MaximumAutoInvokeAttempts, which caps the number of auto-invocations per single user interaction, there's also the MaximumUseAttempts constant, which specifies the number of requests per single user interaction that should include functions in the request.

> As I am testing the required tool choice, no matter how I configure the max auto invoke attempts, I only get back a tool call as part of the Function Result. Versus auto where the model responds with a tool call, we send the tool call response and the model responds with the final answer. I'm curious to learn about how you're testing the required tool choice. Thanks.

Add/advertise/include functions only to the first request to the model and stop doing so for all subsequent ones for the required function choice behavior.

moonbox3: Thanks!

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1571,4 +1540,154 @@ private void LogUsage(CompletionsUsage usage)
...
+        (bool? AllowAnyRequestedKernelFunction, int? MaximumAutoInvokeAttempts)? result = null;
+
+        // If neither behavior specified, we don't need to do anything.
+        if (executionSettings.FunctionChoiceBehavior is null && executionSettings.ToolCallBehavior is null)
+        {
+            return result;
+        }
+
+        // If both behaviors are specified, we can't handle that.
+        if (executionSettings.FunctionChoiceBehavior is not null && executionSettings.ToolCallBehavior is not null)
```
stephentoub, line 1562: Reading this again, it's still really confusing to me that both of these things exist. If we're going to delete the existing one as part of OpenAI SDK update work, maybe for now we just mark it editor browsable never and obsolete with a message telling folks to use the new one?

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
stephentoub: The latter suggestion sounds reasonable.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
moonbox3: In Python, while still in v1, we're marking the `FunctionCallBehavior` as deprecated, and logging warnings to use the new `FunctionChoiceBehavior`. If a customer still configures the `FunctionCallBehavior` class, we're using a model validator to map the `FunctionCallBehavior` to `FunctionChoiceBehavior`. Anytime the old `FunctionCallBehavior` gets an update, so does `FunctionChoiceBehavior`. This allows us to only make decisions based off of `FunctionChoiceBehavior` and makes for easier cleanup when we do remove `FunctionCallBehavior`.

SergeyMenshykh: > Reading this again, it's still really confusing to me that both of these things exist. If we're going to delete the existing one as part of OpenAI SDK update work, maybe for now we just mark it editor browsable never and obsolete with a message telling folks to use the new one? Altough that's odd as well if the new surface area is marked experimental. Ugh.

We plan to remove the `ToolCallBehavior` property and the corresponding code that uses it as part of the OpenAI SDK update work. The PR will be retargeted and merged into the [feature-connectors-openai](https://github.com/microsoft/semantic-kernel/tree/feature-connectors-openaiurl) branch. 

The experimental attributes will be removed, as the new function choice behavior functionality replaces the tool call behavior that was released for a while.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
SergeyMenshykh: > The latter suggestion sounds reasonable, i.e. empty means no functions

There are two ways to handle an empty array. The first way is to disable function calling. The second way is to set `options.ToolChoice` to auto and assign an empty array to the `options.Tools` property. In the second case, AzureSDK will throw an exception stating that no tools were provided for the auto choice. ToolCallBehavior uses the first way, and I more inclined to follow it.

> I'd also be happy not having a list here at all, and just using what's in the kernel, as is the case today.

Do you mean to remove the functions parameter of the Auto static method? Today, tool call behavior allows to supply functions via the `ToolCallBehavior.EnableFunctions(IEnumerable<OpenAIFunction> functions, bool autoInvoke = false)` method.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
stephentoub: Ok, thanks

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
stephentoub: > ToolCallBehavior uses the first way, and I more inclined to follow it

Yes, I don't think we should throw.

> Today, tool call behavior allows to supply functions via the ToolCallBehavior.EnableFunctions(IEnumerable<OpenAIFunction> functions, bool autoInvoke = false) method.

Yeah.  Ok.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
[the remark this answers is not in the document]
SergeyMenshykh: We plan to remove the `ToolCallBehavior` so the functionality will be dropped as well.

dotnet/src/SemanticKernel.Abstractions/AI/FunctionChoiceBehaviors/FunctionChoiceBehavior.cs
```diff
@@ -0,0 +1,187 @@
...
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// Represents the base class for different function choice behaviors.
+/// </summary>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "type")]
+[JsonDerivedType(typeof(AutoFunctionChoiceBehavior), typeDiscriminator: "auto")]
+[JsonDerivedType(typeof(RequiredFunctionChoiceBehavior), typeDiscriminator: "required")]
+[JsonDerivedType(typeof(NoneFunctionChoiceBehavior), typeDiscriminator: "none")]
+public abstract class FunctionChoiceBehavior
```
crickman, line 16: Can we evaluate how this adapts to requirements for assistant-agent _tool-choice_?

https://platform.openai.com/docs/api-reference/runs/createRun#runs-createrun-tool_choice

I wonder if there's a way to not bind this base abstraction so tightly to chat-completion only?  Maybe there is an even more generic base abstraction?  The current form might not adapt so well for assistants.

stephentoub: Why would we not / are we not using the same content types for agents?

PR #6099, opened 2024-05-02
.Net: Function Calling Abstractions ADR

SergeyMenshykh:
### Motivation and Context
Today, every AI connector in SK that supports function calling has its own implementation of tool call behavior model classes. These classes are used to configure the way connectors advertise and invoke functions. For example, the behavior classes can describe which functions should be advertised to LLM by a connector, whether the functions should be called automatically by the connector, or if the connector caller will invoke them himself manually, etc.

All the tool call behavior classes are the same in terms of describing the desired function call behavior. However, the classes have a mapping functionality to map the function call behavior to the connector-specific model classes, and that's what makes the function calling classes non-reusable between connectors. For example, [the constructor of the ToolCallBehavior class](https://github.com/microsoft/semantic-kernel/blob/0c40031eb917bbf46c9af97897051f45e4084986/dotnet/src/Connectors/Connectors.OpenAI/ToolCallBehavior.cs#L165) references the [OpenAIFunction](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/OpenAIFunction.cs) class that lives in the Microsoft.SemanticKernel.Connectors.OpenAI namespace declared in the Connectors.OpenAI project. The classes can't be reused by, let's say, the Mistral AI connector without introducing an explicit project dependency from the Connectors.Mistral project to the Connectors.OpenAI project, a dependency we definitely don't want to have.

Additionally, today, it's not possible to specify function calling behavior declaratively in YAML, MD, and Kernel(config.json) prompts.

### Description

This ADR outlines design options (there's only one at the moment) to address the issues above. Related to https://github.com/microsoft/semantic-kernel/issues/5253

### Contribution Checklist


dmytrostruk (approved):

docs/decisions/0045-function-call-behavior.md
```diff
@@ -0,0 +1,326 @@
...
+public class FunctionChoiceBehaviorContext
+{
+    public Kernel? Kernel { get; init; }
+}
+
+public class FunctionCallChoiceConfiguration
+{
+    public IEnumerable<KernelFunctionMetadata>? AvailableFunctions { get; init; }
+    public IEnumerable<KernelFunctionMetadata>? RequiredFunctions { get; init; }
+    public bool? AllowAnyRequestedKernelFunction { get; init; }
```
markwallace-microsoft, line 70: Should this not be something like this

```csharp
public IEnumerable<KernelFunctionMetadata>? Functions { get; init; } // Functions to advertise to the LLM
public bool? IsRequired { get; init; } // True if the LLM must call one of the functions
public bool? IsNone { get; init; } // True if the LLM must not call one of the functions but instead should respond with a message
```

westey-m (approved):

docs/decisions/0045-function-call-behavior.md
SergeyMenshykh: The ADR is slightly behind the PR - https://github.com/microsoft/semantic-kernel/pull/6083, in which the class has been changed based on the latest feedback to:
image

I will update the ADR later to reflect the latest state of that PR.

docs/decisions/0045-function-call-behavior.md
```diff
@@ -0,0 +1,326 @@
...
+
+Additionally, today, it's not possible to specify function calling behavior declaratively in YAML, MD, and Kernel(config.json) prompts.
+
+## Decision Drivers
+
+- The same set of function call behavior model classes should be connector/mode-agnostic, allowing them to be used by all SK connectors that support function calling.
+- Function calling behavior should be specified in the `PromptExecutionSettings` base class rather than in connector-specific derivatives.
+- It should be possible and easy to specify function calling behavior in all already supported YAML, MD, and SK(config.json) prompts.
```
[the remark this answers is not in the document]
SergeyMenshykh: added, thanks.

docs/decisions/0045-function-call-behavior.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Function call choice classes 
+The `FunctionChoiceBehavior` class is abstract base class for all *FunctionChoiceBehavior classes:
+
+```csharp
+public abstract class FunctionChoiceBehavior
+{
+    public static FunctionChoiceBehavior AutoFunctionChoice(IEnumerable<KernelFunction>? functions = null, bool autoInvoke = true) { ... }
+    public static FunctionChoiceBehavior RequiredFunctionChoice(IEnumerable<KernelFunction>? functions = null, bool autoInvoke = true) { ... }
+    public static FunctionChoiceBehavior None { get { ... }; }
```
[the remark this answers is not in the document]
SergeyMenshykh: It should, fixed.

docs/decisions/0045-function-call-behavior.md
```diff
@@ -0,0 +1,326 @@
...
+
+    [JsonPropertyName("maximumUseAttempts")]
+    public int MaximumUseAttempts { get; init; } = DefaultMaximumUseAttempts;
+
+    public override FunctionChoiceBehaviorConfiguration GetConfiguration(FunctionChoiceBehaviorContext context)
+    {
+        // Handle functions provided via constructor as function instances.
+        ...
+        // Or, handle functions provided via the 'Functions' property as function fully qualified names.
+        ...
```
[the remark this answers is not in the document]
SergeyMenshykh: Both are used to specify functions to be advertised to the model. However, they are presented by different types - the ones passed through the constructor are instances of the `KernelFunction` type, while those passed through the `Functions` property are fully qualified function names represented by strings. This way, it's possible to advertise non-kernel functions (functions created but not imported to the kernel) and functions specified in YAML and JSON prompts where they are specified as strings.

docs/decisions/0045-function-call-behavior.md
moonbox3: Shouldn't the ADR be driving what is done in code, not the other way around? I think it makes it more difficult for other SK languages to fully understand the scope of the work if the code is first completed, and then the ADR is updated.

docs/decisions/0045-function-call-behavior.md
```diff
@@ -0,0 +1,322 @@
...
+Considering the hierarchical nature of the choice behavior model classes, polymorphic deserialization should be enabled for instances when functional choice behavior needs to be configured in JSON prompts - MD and Kernel (config.json) prompts, as well as YAML prompts - handlebars.
+```json
+{
+    ...
+    "execution_settings": {
+        "default": {
+            "temperature": 0.4,
+            "function_choice_behavior": {
+                "type": "auto", //possible values - auto, required, none
+                "functions": [
```
moonbox3, line 211: Is the following scenario correct? In the event that someone defines the functions in the JSON/yaml config and specifies the type as "auto?" In this example "plugin1-function1" is the only tool made available to the model even though the kernel has plugin1-function1, plugin2-function1, plugin3-function1?

If someone first configures the functions in the config, and later updates the filter to include other plugins/functions and these two fully qualified function lists now diverge, what functions list has higher precedence?

docs/decisions/0045-function-call-behavior.md
SergeyMenshykh: It should, but because of the complexity and many comments, despite the ADR presented and reviewed, the code evolved much faster and became a driver for the ADR.

docs/decisions/0045-function-call-behavior.md
SergeyMenshykh: > Is the following scenario correct? In the event that someone defines the functions in the JSON/yaml config and specifies the type as "auto?" In this example "plugin1-function1" is the only tool made available to the model even though the kernel has plugin1-function1, plugin2-function1, plugin3-function1?

Yes, that's the intention: whoever configures it explicitly states that only the 'plugin1-function1' function should be advertised and that the `auto` function choice behavior should be used, even though the other functions are available on kernel.

> If someone first configures the functions in the config, and later updates the filter to include other plugins/functions and these two fully qualified function lists now diverge, what functions list has higher precedence?

I would say that whatever is configured in the filter should have higher precedence over the configuration.

markwallace-microsoft (approved):

PR #6127, opened 2024-05-05
.Net: Samples - Fix array access in Handlebars syntax

johndowns:
### Description

This is a small change to fix what seems like some typos in the samples that use Handlebars syntax to select a default choice for an intent detection prompt template. The [Handlebars docs](https://handlebarsjs.com/guide/expressions.html#literal-segments) show that you access the first element by using the syntax `array.[0]`, but in these templates it's using `array[0]`.

I also tested and confirmed the current approach doesn't work, but the new approach in this PR does.

### Contribution Checklist


dmytrostruk (approved):

teresaqhoang (approved):

teresaqhoang:
Thanks for this fix @johndowns! Merging now

Issue #6063, opened 2024-04-30
Add a CI test leg for .NET Framework


stephentoub:
CI currently only tests on .NET Core (on multiple OSes). We should have at least one leg that validates .NET Framework (Windows).

dmytrostruk:
bump

PR #6148, opened 2024-05-07
.Net: Adds a memory connector for Azure Cosmos DB for NoSQL

Pilchie:
### Motivation and Context

Azure Cosmos DB is adding Vector Similarity APIs to the NoSQL project, and would like Semantic Kernel users to be able to leverage them.

### Description

This adds a Memory Connector implementation for Azure Cosmos DB's, including support for the new vector search functionality coming soon in Cosmos DB. It is mostly based off the existing connector for Azure Cosmos DB for Mongo DB vCore.

### Contribution Checklist


dotnet/Directory.Packages.props
```diff
@@ -85,6 +84,7 @@
     <PackageVersion Include="protobuf-net.Reflection" Version="3.2.12" />
     <PackageVersion Include="YamlDotNet" Version="15.1.2" />
     <!-- Memory stores -->
+    <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.40.9000-nightly-20240430" />
```
Pilchie, line 87: Need to wait for package to be published publicly to merge.

dotnet/nuget.config
```diff
@@ -4,9 +4,13 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="local" value="D:\code\VectorSearchDotNetSDKNugets" />
   </packageSources>
 
   <packageSourceMapping>
+    <packageSource key="local">
+      <package pattern="Microsoft.Azure.Cosmos*"/>
+    </packageSource>
```
Pilchie, line 13: This will be reverted when the NuGet package is published publicly.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Cosmos;
+
+/// <summary>
+/// This class provides a default implementation of System.Text.Json Cosmos Linq Serializer.
+/// </summary>
+public class CosmosSystemTextJsonSerializer : CosmosLinqSerializer
```
Pilchie, line 16: This will be added to the Azure Cosmos DB SDK soon, and we can remove this when it is.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+    /// <param name="applicationName">The application name to use in requests.</param>
+    public AzureCosmosDBNoSQLMemoryStore(
+        CosmosClient cosmosClient,
+        string databaseName,
+        VectorEmbeddingPolicy vectorEmbeddingPolicy,
+        IndexingPolicy indexingPolicy,
+        string? applicationName = null)
+    {
+        this._cosmosClient = cosmosClient;
+        this._cosmosClient.ClientOptions.ApplicationName = applicationName;
```
Pilchie, line 75: Cosmos DB folks - will this work after the `CosmosClient` has been created?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/Connectors.Memory.AzureCosmosDBNoSQL.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- THIS PROPERTY GROUP MUST COME FIRST -->
+    <AssemblyName>Microsoft.SemanticKernel.Connectors.AzureCosmosDBNoSQL</AssemblyName>
+    <RootNamespace>$(AssemblyName)</RootNamespace>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
markwallace-microsoft, line 7: Add `net8.0` ?

westey-m (changes requested):

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/Connectors.Memory.AzureCosmosDBNoSQL.csproj
Pilchie: Is there concrete benefit to doing so when the code doesn't depend on anything .NET 8?  That would require also multi-targeting the `IntegrationTests` project, or accepting test coverage only on .NET 8?

Or do you mean *changing* to `net8.0` instead of `netstandard2.0`?

stephentoub: The test projects could multitarget to .NET 8 and .NET 6, but the latter will be out of support in November, and there's rarely an execution difference between the two. The more impactful thing to test would be .NET Framework, for which there's currently no CI support. That's relevant regardless of multitargeting, and is tracked by https://github.com/microsoft/semantic-kernel/issues/6063.

As for concrete benefits, there's the consistency with the rest of the projects, but also that the compiler will see newer surface area, so for example it'll compile string interpolation in a more efficient manner, it'll bind to newer more efficient overloads that might exist, etc.

Pilchie: So, just to confirm @stephentoub, your recommendation is to multi-target this project to `netstandard2.0` and `net8.0`, and just leave the test project only testing the `net8.0` implementation?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+        await foreach (
+            var existingCollectionName in this.GetCollectionsAsync(cancellationToken)
+                .ConfigureAwait(false)
+        )
+        {
+            if (existingCollectionName == collectionName)
+            {
+                return true;
+            }
+        }
```
stephentoub, line 132: I see in GetCollectionsAsync that there's a query being specified. Would this be able to specify a more specific query just for the name and avoid then having to bring down and iterate through the collection names on the client?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+
+                if (localWhere.Length >= OR.Length)
+                {
+                    localWhere.Length -= OR.Length;
+
+                    var localQueryDefinition = new QueryDefinition($"""
+                    SELECT x.id,x.key,x.metadata,x.timestamp{(withEmbeddings ? ",x.embedding" : "")}
+                    FROM x
+                    WHERE {localWhere}
+                    """);
```
stephentoub, line 254: Nit: indent?
```suggestion
                        SELECT x.id,x.key,x.metadata,x.timestamp{(withEmbeddings ? ",x.embedding" : "")}
                        FROM x
                        WHERE {localWhere}
                        """);
```

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
stephentoub: Can it be internal until then?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        if (typeof(Stream).IsAssignableFrom(typeof(T)))
+        {
+            return (T)(object)stream;
+        }
+
+        using (stream)
+        {
+            using StreamReader reader = new(stream);
+            return JsonSerializer.Deserialize<T>(reader.ReadToEnd(), this._jsonSerializerOptions);
```
stephentoub, line 55: JsonSerializer.Deserializer has overloads that can deserialize directly from Stream. Can those be used instead of reading the whole contents into a string first?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    /// <inheritdoc/>
+    public override Stream ToStream<T>(T input)
+    {
+        MemoryStream streamPayload = new();
+        using Utf8JsonWriter writer = new(streamPayload);
+
+        JsonSerializer.Serialize(
```
stephentoub, line 65: Same here... could this Serialize directly to streamPayload?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
Pilchie: Done.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+    {
+        return $"{this.Key} - {this.SimilarityScore}";
+    }
+}
+
+/// <summary>
+/// Creates a new record that also serializes an "id" property.
+/// </summary>
+[DebuggerDisplay("{GetDebuggerDisplay()}")]
+public class MemoryRecordWithId : MemoryRecord
```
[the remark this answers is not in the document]
Pilchie: Yep, done!

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+        MemoryRecordMetadata metadata,
+        ReadOnlyMemory<float> embedding,
+        string? key,
+        DateTimeOffset? timestamp = null)
+        : base(metadata, embedding, key, timestamp)
+    {
+    }
+
+    /// <summary>
+    /// The similarity score returned.
```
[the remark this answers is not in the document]
Pilchie: Fixed, thanks!

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+
+/// <summary>
+/// Creates a new record with a similarity score.
+/// </summary>
+/// <param name="metadata"></param>
+/// <param name="embedding"></param>
+/// <param name="key"></param>
+/// <param name="timestamp"></param>
+[DebuggerDisplay("{GetDebuggerDisplay()}")]
+public class MemoryRecordWithSimilarityScore(
```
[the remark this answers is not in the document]
Pilchie: Yep, done!

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+    {
+        var options = new CosmosClientOptions
+        {
+            ApplicationName = applicationName ?? HttpHeaderConstant.Values.UserAgent,
+            Serializer = new CosmosSystemTextJsonSerializer(JsonSerializerOptions.Default),
+        };
+        this._cosmosClient = new CosmosClient(connectionString, options);
+        this._databaseName = databaseName;
+        this._vectorEmbeddingPolicy = vectorEmbeddingPolicy;
+        this._indexingPolicy = indexingPolicy;
```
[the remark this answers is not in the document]
Pilchie: In the latest change I added a check for the the `vectorEmbeddingPolicy`.  I didn't for the `indexingPolicy` since it can technically work, Cosmos DB will just end up doing a brute force search that will be slow and expensive. Maybe we should be more prescriptive than that though?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/Connectors.Memory.AzureCosmosDBNoSQL.csproj
stephentoub: Yes. And then hopefully very soon we get the .NET Framework test leg. @markwallace-microsoft, how hard would that be? @dmytrostruk, you were just mucking around in there... would we be able to add that quickly?

Until then, I get that there's a gap where the netstandard2.0 assets aren't being tested in CI, and that's unfortunate. If we want to try to augment CI with a test leg for .NET 6 before then, and multitarget the test projects to .NET 6, we could do so. I'm just pointing out that this issue isn't unique to this connector; the rest of the SK assemblies are in the same boat.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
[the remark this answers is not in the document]
westey-m: That sounds like the right balance to me. I had assumed it was required in indexingPolicy as well for vector search to work, but if it's not required by Cosmos DB, I don't think we should make it required in our sdk.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
Pilchie: Apparently the version of STJ our SDK depends on is too old, according to the discussion at https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4332/files#r1562040906.

I could address that by adding an explicit reference to a newer STJ here in this project if that makes sense to you?

westey-m (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,438 @@
...
+            // We're in the very large case, we'll need to split this into multiple queries.
+            // We add one to catch any fractional piece left in the last segment
+            var segments = (int)(ratio + 1);
+            var keyList = keys.ToList();
+            var keysPerQuery = keyList.Count / segments;
+            // Make a guess as to how long this query will be. We need at least 26 chars for each "OR" block, so
+            // put a few extra for the values of the keys.
+            var estimatedWhereLength = 30 * keysPerQuery;
+            var localWhere = new StringBuilder(estimatedWhereLength);
+            const string OR = " OR ";
```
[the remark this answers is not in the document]
Pilchie: Done

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
stephentoub: Ouch, STJ 4.6.0 is 5 years. The rest of the SK assemblies reference the 8.0 version:
https://github.com/microsoft/semantic-kernel/blob/0c89e0bd4314b4f4c913563258ffefadedab1afe/dotnet/Directory.Packages.props#L37
and since this assembly depends on some of those, I expect you're already being forced up. Might as well benefit from it :smile:

Pilchie: Yep, comes through the abstractions project.  Updated.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/Connectors.Memory.AzureCosmosDBNoSQL.csproj
Pilchie: Done.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
stephentoub: This overload:
https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.serialize?view=net-8.0#system-text-json-jsonserializer-serialize-1(system-io-stream-0-system-text-json-jsonserializeroptions)
Is that not showing up for some reason?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
Pilchie: Good catch - done.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/CosmosSystemTextJSonSerializer.cs
Pilchie: I think you meant 
https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.serialize?view=net-8.0#system-text-json-jsonserializer-serialize-1(system-io-stream-0-system-text-json-jsonserializeroptions).

It wasn't showing up because of the named parameters.  Fixed now.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
Pilchie: Answering my own question - this will work unless the `ComosClient` already has a connection established, after which it would be ignored. I think that's okay for this case.

dotnet/Directory.Packages.props
Pilchie: Fixed.

dotnet/nuget.config
Pilchie: Reverted.

Pilchie:
I believe this is ready to go now @westey-m, @markwallace-microsoft, @stephentoub.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/Connectors.Memory.AzureCosmosDBNoSQL.csproj
markwallace-microsoft: We'll look into extending the testing coverage. Likely this will be after Build.

stephentoub: Last night I got all the merged test projects building for net462 and all the unit test projects passing on netfx. https://github.com/microsoft/semantic-kernel/pull/6326 is the draft progress.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,462 @@
...
+
+    /// <inheritdoc/>
+    public async Task<bool> DoesCollectionExistAsync(
+        string collectionName,
+        CancellationToken cancellationToken = default)
+    {
+        using var feedIterator = this.
+            _cosmosClient
+            .GetDatabase(this._databaseName)
+            .GetContainerQueryIterator<string>($"SELECT VALUE(c.id) FROM c WHERE c.id = '{collectionName}'");
```
stephentoub, line 137: Any SQL injection attack concerns here?

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -0,0 +1,462 @@
...
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        const string OR = " OR ";
+
+        // Optimistically create the entire query string.
+        var whereClause = string.Join(OR, keys.Select(k => $"(x.id = \"{k}\" AND x.key = \"{k}\")"));
+        var queryDefinition = new QueryDefinition($"""
+            SELECT x.id,x.key,x.metadata,x.timestamp{(withEmbeddings ? ",x.embedding" : "")}
+            FROM x
+            WHERE {whereClause}
```
stephentoub, line 221: Same question

Pilchie: Replaced all.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
Pilchie: SQL injection is harder in Cosmos DB, since our SQL syntax doesn't allow modification, but I suppose it might be possible to construct something to disclose extra information, so I've switched all places that use argument data in queries to use parameterized queries.

stephentoub (approved):

Pilchie:
Thanks so much everyone!

PR #6150, opened 2024-05-07
.Net: Add model diagnostics to non-streaming APIs

TaoChenOSU:
### Motivation and Context

According to the [ADR](https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0044-OTel-semantic-convention.md), it's essential that SK provides the best developer experience while complying with the industry standards for observability in generative-AI-based applications.

### Description

This PR adds instrumentation to the chat completion and text completion endpoints in all AI connectors.

Streaming APIs will be worked on next.
The telemetry sample and the documentation will be updated in a separate PR.

> Note that this is an ongoing effort, i.e. metrics and more events will be added as the conventions evolve.

### Contribution Checklist


dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,244 @@
...
+        {
+            return;
+        }
+
+        activity.AddTags(
+            [
+                new(ModelDiagnosticsTags.FinishReason, GetFinishReasons(completions)),
+                new(ModelDiagnosticsTags.PromptToken, promptTokens ?? -1),
+                new(ModelDiagnosticsTags.CompletionToken, completionTokens ?? -1),
+                new(ModelDiagnosticsTags.ResponseId, GetResponseId(completions.FirstOrDefault())),
```
TaoChenOSU, line 177: Should we keep the tags with some default values if they are not available or we don't include the tags when the values aren't available?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -132,15 +134,31 @@ internal ClientCore(ILogger? logger = null)
 
         var options = CreateCompletionsOptions(text, textExecutionSettings, this.DeploymentOrModelName);
 
-        var responseData = (await RunRequestAsync(() => this.Client.GetCompletionsAsync(options, cancellationToken)).ConfigureAwait(false)).Value;
-        if (responseData.Choices.Count == 0)
+        Completions responseData;
+        IEnumerable<TextContent> responseContent;
+        using (var activity = ModelDiagnostics.StartCompletionActivity(this.Endpoint, this.DeploymentOrModelName, "OpenAI", text, executionSettings))
```
TaoChenOSU, line 139: Use a constant for "OpenAI"

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceMessageApiClient.cs
```diff
@@ -106,14 +107,27 @@ await foreach (var streamingChatContent in this.ProcessChatResponseStreamAsync(r
         string modelId = executionSettings?.ModelId ?? this._clientCore.ModelId;
         var endpoint = this.GetChatGenerationEndpoint();
         var request = this.CreateChatRequest(chatHistory, executionSettings);
+
         using var httpRequestMessage = this._clientCore.CreatePost(request, endpoint, this._clientCore.ApiKey);
+        using var activity = ModelDiagnostics.StartCompletionActivity(endpoint, modelId, "HuggingFace", chatHistory, executionSettings);
```
TaoChenOSU, line 112: Use a constant for "HuggingFace"

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -161,11 +162,29 @@ internal sealed class GeminiChatCompletionClient : ClientBase
 
         for (state.Iteration = 1; ; state.Iteration++)
         {
-            var geminiResponse = await this.SendRequestAndReturnValidGeminiResponseAsync(
-                    this._chatGenerationEndpoint, state.GeminiRequest, cancellationToken)
-                .ConfigureAwait(false);
+            GeminiResponse geminiResponse;
+            List<GeminiChatMessageContent> chatResponses;
+            using (var activity = ModelDiagnostics.StartCompletionActivity(
+                this._chatGenerationEndpoint, this._modelId, "Google", chatHistory, executionSettings))
```
TaoChenOSU, line 168: Use a constant for "Google"

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+    /// </summary>
+    private static Activity? StartCompletionActivity<T>(
+        Uri? endpoint,
+        string modelName,
+        string modelProvider,
+        T prompt,
+        PromptExecutionSettings? executionSettings,
+        Func<T, string> formatPrompt)
+    {
+        if (!IsModelDiagnosticsEnabled())
```
stephentoub, line 121:
```suggestion
        if (!IsModelDiagnosticsEnabled || !s_activitySource.HasListeners)
```

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+
+    # region Private
+    /// <summary>
+    /// Check if model diagnostics is enabled
+    /// Model diagnostics is enabled if either EnableModelDiagnostics or EnableSensitiveEvents is set to true.
+    /// </summary>
+    private static bool IsModelDiagnosticsEnabled()
+    {
+        return s_enableModelDiagnostics || s_enableSensitiveEvents;
+    }
```
stephentoub, line 77:
```suggestion
    private static bool IsModelDiagnosticsEnabled => s_enableModelDiagnostics;
```

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+internal static class ModelDiagnostics
+{
+    private static readonly string s_namespace = typeof(ModelDiagnostics).Namespace!;
+    private static readonly ActivitySource s_activitySource = new(s_namespace);
+
+    private const string EnableModelDiagnosticsSettingName = "Microsoft.SemanticKernel.Experimental.EnableModelDiagnostics";
+    private const string EnableSensitiveEventsSettingName = "Microsoft.SemanticKernel.Experimental.EnableModelDiagnosticsWithSensitiveData";
+
+    private static readonly bool s_enableModelDiagnostics = AppContextSwitchHelper.GetConfigValue(EnableModelDiagnosticsSettingName);
+    private static readonly bool s_enableSensitiveEvents = AppContextSwitchHelper.GetConfigValue(EnableSensitiveEventsSettingName);
```
stephentoub, line 23:
```suggestion
    private static readonly bool s_enableSensitiveEvents = AppContextSwitchHelper.GetConfigValue(EnableSensitiveEventsSettingName);
    private static readonly bool s_enableModelDiagnostics = s_enableSensitiveEvents 
 || AppContextSwitchHelper.GetConfigValue(EnableModelDiagnosticsSettingName);
```

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+    private static string GetFinishReasons(IEnumerable<KernelContent> completions)
+    {
+        var finishReasons = completions.Select(c =>
+        {
+            if (c.Metadata?.TryGetValue("FinishReason", out var finishReason) == true)
+            {
+                return finishReason;
+            }
+
+            return null;
```
stephentoub, line 201: Isn't this possibly going to result in a bunch of ", , , , , ," etc. from the string.Join?

dotnet/src/Agents/Abstractions/Agents.Abstractions.csproj
```diff
@@ -15,11 +15,15 @@
   <PropertyGroup>
     <!-- NuGet Package Settings -->
     <Title>Semantic Kernel Agents - Abstractions</Title>
-    <Description>Semantic Kernel Agents abstractions. This package is automatically installed by Semantic Kernel Agents packages if needed.</Description>
+    <Description>Semantic Kernel Agents abstractions. This package is automatically installed by
+      Semantic Kernel Agents packages if needed.</Description>
```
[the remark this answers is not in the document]
TaoChenOSU: This is changed by autoformat.

dotnet/src/Agents/Core/Agents.Core.csproj
```diff
@@ -16,12 +16,17 @@
   <PropertyGroup>
     <!-- NuGet Package Settings -->
     <Title>Semantic Kernel Agents - Core</Title>
-    <Description>Defines core set of concrete Agent and AgentChat classes, based on the Agent Abstractions.</Description>
+    <Description>Defines core set of concrete Agent and AgentChat classes, based on the Agent
+      Abstractions.</Description>
```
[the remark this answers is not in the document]
TaoChenOSU: This is also changed by autoformat.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
lmolkova, line 168: Does  `"Google"` constant go to `gen_ai.system`? It (and all the others) should be lowercase then. Language-specific preferences does not apply and we need to have common casing everywhere. Lowercase is the existing approach for otel

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -136,14 +137,27 @@ internal HttpRequestMessage CreatePost(object requestData, Uri endpoint, string?
         string modelId = executionSettings?.ModelId ?? this.ModelId;
         var endpoint = this.GetTextGenerationEndpoint(modelId);
         var request = this.CreateTextRequest(prompt, executionSettings);
+
         using var httpRequestMessage = this.CreatePost(request, endpoint, this.ApiKey);
+        using var activity = ModelDiagnostics.StartCompletionActivity(endpoint, modelId, "HuggingFace", prompt, executionSettings);
```
lmolkova, line 142:
```suggestion
        using var activity = ModelDiagnostics.StartCompletionActivity(endpoint, modelId, "hugging_face", prompt, executionSettings);
```
?

Mind sending a PR to extend the list of constants to the OTel semconv (lmk if you need any help fighting with tooling)?

dotnet/src/InternalUtilities/src/Diagnostics/ActivityExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Microsoft.SemanticKernel.Diagnostics;
+
+[ExcludeFromCodeCoverage]
+internal static class ActivityExtensions
+{
+    public static Activity? StartActivityWithTags(this ActivitySource source, string name, List<KeyValuePair<string, object?>> tags)
+    {
+        return source.StartActivity(
+            name,
+            ActivityKind.Internal,
```
lmolkova, line 16: according to the semconv it should be client. I'm violating it in my prototypes, not sure what to do about it for now, but you don't have to violate :)

dotnet/src/InternalUtilities/src/Diagnostics/ActivityExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+            Activity.Current?.Context ?? new ActivityContext(),
```
lmolkova, line 17:
```suggestion
            default,
```

the `StartActivity`  will use current if available or will create a new one

dotnet/src/InternalUtilities/src/Diagnostics/ActivityExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            activity.SetTag(tag.Key, tag.Value);
+        });
+
+        return activity;
+    }
+
+    public static Activity AttachSensitiveDataAsEvent(this Activity activity, string name, List<KeyValuePair<string, object?>> tags)
+    {
+        activity.AddEvent(new ActivityEvent(
```
lmolkova, line 33: I've been trying to make it work with EventSource or ILogger or OTel Logs and it's just impossible. So for .NET span events is the only choice for the time being 😓

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+namespace Microsoft.SemanticKernel;
+
+[ExcludeFromCodeCoverage]
+internal static class ModelDiagnostics
+{
+    private static readonly string s_namespace = typeof(ModelDiagnostics).Namespace!;
+    private static readonly ActivitySource s_activitySource = new(s_namespace);
+
+    private const string EnableModelDiagnosticsSettingName = "Microsoft.SemanticKernel.Experimental.EnableModelDiagnostics";
+    private const string EnableSensitiveEventsSettingName = "Microsoft.SemanticKernel.Experimental.EnableModelDiagnosticsWithSensitiveData";
```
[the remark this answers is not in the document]
lmolkova: maybe something like
`SEMANTICKERNEL_EXPERIMENTAL_GENAI_OTEL_ENABLE` 

and 

`SEMANTICKERNEL_EXPERIMENTAL_GENAI_OTEL_ENABLE_SENSITIVE_CONTENT`

`OTELLLMCONVENTION` is a bit hard to read :)

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+        if (activity is null || executionSettings?.ExtensionData is null)
+        {
+            return;
+        }
+
+        void TryAddTag(string key, string tag)
+        {
+            if (executionSettings.ExtensionData.TryGetValue(key, out var value))
+            {
+                activity.AddTag(tag, value);
```
lmolkova, line 90:
```suggestion
                activity.SetTag(tag, value);
```

Adding a tag supports duplication, setting a tag overwrites the existing (if any) value.
We made a mistake in the past and accidentally added the same attribute twice which broke some parts of the world. So `SetTag` is always safer :)

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
lmolkova: we should not include them at all

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+        {
+            if (c.Metadata?.TryGetValue("FinishReason", out var finishReason) == true)
+            {
+                return finishReason;
+            }
+
+            return null;
+        });
+
+        return $"[{string.Join(", ", finishReasons)}]";
```
lmolkova, line 204: 👍 
nit: any reason to do `"[stop, stop]"` instead of `"stop,stop"` ?
`[stop, stop]` is not a valid json, nothing will expect json, in many cases there is just one stop reason. 
Also I'd avoid ` `

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    private static string GetResponseId(KernelContent? completion)
+    {
+        if (completion?.Metadata?.TryGetValue("Id", out var id) == true)
+        {
+            return id as string ?? "N/A";
```
lmolkova, line 211: don't set the response id attribute if there is no Id

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+    /// <summary>
+    /// Tags used in model diagnostics
+    /// </summary>
+    private static class ModelDiagnosticsTags
+    {
+        // Activity tags
+        public const string System = "gen_ai.system";
+        public const string Operation = "gen_ai.operation.name";
+        public const string Model = "gen_ai.request.model";
+        public const string MaxToken = "gen_ai.request.max_token";
```
lmolkova, line 226:
```suggestion
        public const string MaxToken = "gen_ai.request.max_tokens";
```

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+        public const string TopP = "gen_ai.request.top_p";
+        public const string ResponseId = "gen_ai.response.id";
+        public const string ResponseModel = "gen_ai.response.model";
+        public const string FinishReason = "gen_ai.response.finish_reason";
+        public const string PromptToken = "gen_ai.response.prompt_tokens";
+        public const string CompletionToken = "gen_ai.response.completion_tokens";
+        public const string Prompt = "gen_ai.content.prompt";
+        public const string Completion = "gen_ai.content.completion";
+        public const string Address = "server.address";
+        public const string Port = "server.port";
```
lmolkova, line 237:
```suggestion
        public const string Port = "server.port";
        public const string ErrorType = "error.type";
```

it's the most useful attribute in my small world :)

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -132,15 +136,31 @@ internal ClientCore(ILogger? logger = null)
...
+        IEnumerable<TextContent> responseContent;
+        using (var activity = ModelDiagnostics.StartCompletionActivity(this.Endpoint, this.DeploymentOrModelName, ModelProvider, text, executionSettings))
         {
-            throw new KernelException("Text completions not found");
+            try
+            {
+                responseData = (await RunRequestAsync(() => this.Client.GetCompletionsAsync(options, cancellationToken)).ConfigureAwait(false)).Value;
+                if (responseData.Choices.Count == 0)
+                {
+                    throw new KernelException("Text completions not found");
```
lmolkova, line 148: you'd still want to record other non-choice related things on the activity (if any are available) which won't happen because of the exception

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -323,18 +343,34 @@ await foreach (Completions completions in response.ConfigureAwait(false))
...
+            using (var activity = ModelDiagnostics.StartCompletionActivity(this.Endpoint, this.DeploymentOrModelName, ModelProvider, chat, executionSettings))
             {
-                throw new KernelException("Chat completions not found");
+                try
+                {
+                    responseData = (await RunRequestAsync(() => this.Client.GetChatCompletionsAsync(chatOptions, cancellationToken)).ConfigureAwait(false)).Value;
+                    this.CaptureUsageDetails(responseData.Usage);
+                    if (responseData.Choices.Count == 0)
+                    {
+                        throw new KernelException("Chat completions not found");
```
lmolkova, line 356: same here, let's record the properties we can record even if there are no choices

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -323,18 +343,34 @@ await foreach (Completions completions in response.ConfigureAwait(false))
...
+                    }
+                }
+                catch (Exception ex)
+                {
+                    activity?.SetError(ex);
+                    throw;
+                }
+
+                responseContent = responseData.Choices.Select(chatChoice => this.GetChatMessage(chatChoice, responseData));
+                activity?.SetCompletionResponse(responseContent, responseData.Usage.PromptTokens, responseData.Usage.CompletionTokens);
```
lmolkova, line 366: do we handle the case when

- Streaming starts
- N chunks are received
- Connection drops and stream interrupts
?

We need to make sure to 
- report an activity with reasonable error.type
- flush received parts of the stream and report events for the partial context
- probably set finish reason to an error

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: This was actually my original implementation. And then I thought "what would happen if the someone accidentally change the order of the two statements?". It won't generate any error, but the behavior will be very different.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /// <summary>
+    /// Convert chat history to a string aligned with the OpenAI format
+    /// </summary>
+    private static string ToOpenAIFormat(IEnumerable<ChatMessageContent> chatHistory)
+    {
+        return $"[{string.Join(
+            ", \n",
+            chatHistory.Select(m => $"{{\"role:\" {m.Role}, \"content:\" {m.Content}}}"))}]";
```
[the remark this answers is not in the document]
TaoChenOSU: Yes, `role` can be the following:
- `system`
- `assistant`
- `user`
- `tool`

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: I created a new extension method `SetError` which will attach this attribute to the span.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
TaoChenOSU: Streaming will be my focus next. This one is for non-streaming APIs to keep the PRs relatively small.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: Tags should be dropped if the values are not available.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: Yes, it was a possibility. With the latest changes, it's no longer the case. The finish reason will be `N/A` if it's not found.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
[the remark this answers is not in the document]
TaoChenOSU: I went with the following:
- `Microsoft.SemanticKernel.Experimental.GenAI.EnableOTelDiagnostics`
- `Microsoft.SemanticKernel.Experimental.GenAI.EnableOTelDiagnosticsSensitive`
- `SEMANTICKERNEL_EXPERIMENTAL_GENAI_ENABLE_OTEL_DIAGNOSTICS`
- `SEMANTICKERNEL_EXPERIMENTAL_GENAI_ENABLE_OTEL_DIAGNOSTICS_SENSITIVE`

Let me know if these are clear enough for our users. I am happy to improve them further.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
TaoChenOSU: I think we should use `huggingface` (without the `_`) since `OpenAI` is `openai`.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: `HasListener()` check has been added to `IsModelDiagnosticsEnabled()`.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -323,18 +351,42 @@ await foreach (Completions completions in response.ConfigureAwait(false))
         for (int requestIndex = 1; ; requestIndex++)
         {
             // Make the request.
-            var responseData = (await RunRequestAsync(() => this.Client.GetChatCompletionsAsync(chatOptions, cancellationToken)).ConfigureAwait(false)).Value;
-            this.CaptureUsageDetails(responseData.Usage);
-            if (responseData.Choices.Count == 0)
+            ChatCompletions? responseData = null;
+            List<OpenAIChatMessageContent> responseContent;
+            using (var activity = ModelDiagnostics.StartCompletionActivity(this.Endpoint, this.DeploymentOrModelName, ModelProvider, chat, executionSettings))
```
stephentoub, line 356: As written, we're going to get a span around each roundtrip to the LLM, but if there are N roundtrips as part of the single user call to GetChatMessageContentsAsync, those N roundtrips won't be wrapped in a single activity. Can we add another layer of activity around the loop, such that all the activities for each constituent requests are grouped?

dotnet/src/Agents/Abstractions/Agents.Abstractions.csproj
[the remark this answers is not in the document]
stephentoub: Presumably it should be reverted.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -132,15 +136,39 @@ internal ClientCore(ILogger? logger = null)
...
+                    activity?
+                        .SetResponseId(responseData.Id)
+                        .SetPromptTokenUsage(responseData.Usage.PromptTokens)
+                        .SetCompletionTokenUsage(responseData.Usage.CompletionTokens);
+                }
+                throw;
+            }
+
+            responseContent = responseData.Choices.Select(choice => new TextContent(choice.Text, this.DeploymentOrModelName, choice, Encoding.UTF8, GetTextChoiceMetadata(responseData, choice))).ToList();
+            activity?.SetCompletionResponse(responseContent, responseData.Usage.PromptTokens, responseData.Usage.CompletionTokens);
```
stephentoub, line 166: Presumably we still want to set information on the activity about token usage and the like even if there's an error (like no choices, though I'm not sure in what situation that could happen)

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+    /// </summary>
+    public static Activity? StartCompletionActivity(Uri? endpoint, string modelName, string modelProvider, string prompt, PromptExecutionSettings? executionSettings)
+        => StartCompletionActivity(endpoint, modelName, modelProvider, prompt, executionSettings, prompt => prompt);
+
+    /// <summary>
+    /// Start a chat completion activity for a given model.
+    /// The activity will be tagged with the a set of attributes specified by the semantic conventions.
+    /// </summary>
+    public static Activity? StartCompletionActivity(Uri? endpoint, string modelName, string modelProvider, ChatHistory chatHistory, PromptExecutionSettings? executionSettings)
+        => StartCompletionActivity(endpoint, modelName, modelProvider, chatHistory, executionSettings, chatHistory => ToOpenAIFormat(chatHistory.AsEnumerable()));
```
stephentoub, line 48: Why is the `.AsEnumerable()` needed?

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    /// <summary>
+    /// Convert chat history to a string aligned with the OpenAI format
+    /// </summary>
+    private static string ToOpenAIFormat(IEnumerable<ChatMessageContent> chatHistory)
+    {
+        return $"[{string.Join(
+            ", \n",
+            chatHistory.Select(m => $"{{\"role:\" {m.Role}, \"content:\" {m.Content}}}"))}]";
```
stephentoub, line 125: It'd be nice to avoid all these intermediate allocations and instead format into a StringBuilder.

stephentoub (approved):

dotnet/src/Agents/Abstractions/Agents.Abstractions.csproj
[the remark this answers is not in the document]
TaoChenOSU: Done

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
TaoChenOSU: I agree that we should wrap all the auto tool calls in an activity. I will do that in a separate PR. I created an item to track the work: https://github.com/microsoft/semantic-kernel/issues/6176

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: You're right. It's not needed.

TaoChenOSU: Removed

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
TaoChenOSU: Right, it will attempt to set the token usage even when there are no choices. Please see the catch block above.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: Updated with the optimization. However, we cannot avoid the intermediate string objects completely as I realized I had to make them Json compatible (the content may contain line breaks).

alliscode (approved):

PR #6151, opened 2024-05-08
.Net: Example of Semantic Caching with Filters

dmytrostruk:
### Motivation and Context

This example shows how to achieve Semantic Caching with Filters.
`IPromptRenderFilter` is used to get rendered prompt and check in cache if similar prompt was already answered. If there is a record in cache, then previously cached answer will be returned to the user instead of making a call to LLM. If there is no record in cache, a call to LLM will be performed, and result will be cached together with rendered prompt. 
`IFunctionInvocationFilter` is used to update cache with rendered prompt and related LLM result.

Example includes in-memory, Redis and Azure Cosmos DB for MongoDB as caching stores. 
Common output which demonstrates that second execution is faster, because the result is returned from cache: 
```
First run: What's the tallest building in New York?
Elapsed Time: 00:00:03.828
Second run: What is the highest building in New York City?
Elapsed Time: 00:00:00.541
Result 1: The tallest building in New York is One World Trade Center, also known as Freedom Tower. It stands at 1,776 feet (541.3 meters) tall, including its spire.
Result 2: The tallest building in New York is One World Trade Center, also known as Freedom Tower. It stands at 1,776 feet (541.3 meters) tall, including its spire.
```

PR also contains a couple of fixes in Azure Cosmos DB for MongoDB connector and a couple of additions in public API:
1. Added `FunctionResult? Result` property to `PromptRenderContext`. By default it's `null`, because at prompt rendering stage there is no available result yet. But it's possible to set result with some value - in this case, prompt won't be sent to LLM. Instead, the result from filter will be returned.
2. Added `string? RenderedPrompt` to `FunctionResult` type as `Experimental`. By default it's `null`, and will be populated only when `KernelFunctionFromPrompt` is executed. This property will provide a couple of benefits:
   - It's an additional way how to observe rendered prompt which was sent to LLM during function invocation (today, it's possible to see it only through filter or trace logging).
   - Rendered prompt will be also available in function invocation/automatic function invocation filters, which is required for caching scenarios to store rendered prompt and LLM result together.

### Contribution Checklist


westey-m (approved):

dotnet/samples/Concepts/Caching/SemanticCachingWithFilters.cs
```diff
@@ -0,0 +1,253 @@
...
+namespace Caching;
+
+/// <summary>
+/// This example shows how to achieve Semantic Caching with Filters.
+/// <see cref="IPromptRenderFilter"/> is used to get rendered prompt and check in cache if similar prompt was already answered.
+/// If there is a record in cache, then previously cached answer will be returned to the user instead of making a call to LLM.
+/// If there is no record in cache, a call to LLM will be performed, and result will be cached together with rendered prompt.
+/// <see cref="IFunctionInvocationFilter"/> is used to update cache with rendered prompt and related LLM result.
+/// </summary>
+public class SemanticCachingWithFilters(ITestOutputHelper output) : BaseTest(output)
```
westey-m, line 19: We should also add a link to this from here:
https://github.com/microsoft/semantic-kernel/blob/main/dotnet/samples/Concepts/README.md

westey-m (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStore.cs
```diff
@@ -147,6 +147,8 @@ await this
         CancellationToken cancellationToken = default
     )
     {
+        record.Key = record.Metadata.Id;
```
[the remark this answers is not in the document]
dmytrostruk: Yes, that's correct if memory store is used directly. This line is a fix for abstractions like `ISemanticTextMemory` where `Key` and `Metadata.Id` are used incorrectly. If you search for `record.Key = record.Metadata.Id;` line in a codebase, you will find this line in most of the connectors :) I hope it's something that can be fixed during memory redesign, but for now I decided to apply the same fix as in other connectors to make this memory store work for example and avoid big number of changes.

dotnet/samples/Concepts/Caching/SemanticCachingWithFilters.cs
dmytrostruk: Good catch, thank you! Added.

markwallace-microsoft (approved):

PR #6160, opened 2024-05-08
Net: Add Net Sessions (Code Interpreter) Core Plugin and Demo Project

RogerBarreto:
### Contribution Checklist

### DIFF
--- a/dotnet/samples/Demos/CodeInterpreterPlugin/Program.cs
+++ b/dotnet/samples/Demos/CodeInterpreterPlugin/Program.cs
@@ -1,0 +1,76 @@
+﻿using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.ChatCompletion;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+using Microsoft.SemanticKernel.Plugins.Core.CodeInterpreter;
+using System.Text;
+
+#pragma warning disable SKEXP0050 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+
+var configuration = new ConfigurationBuilder()
+    .AddUserSecrets<Program>()
+    .AddEnvironmentVariables()
+    .Build();
+
+var apiKey = configuration["OpenAI:ApiKey"];
+var token = configuration["AzureContainerApps:BearerKey"];
+var endpoint = configuration["AzureContainerApps:Endpoint"];
+
+ArgumentNullException.ThrowIfNull(apiKey);
+ArgumentNullException.ThrowIfNull(token);
+ArgumentNullException.ThrowIfNull(endpoint);
+
+Task<string> TokenProvider() => Task.FromResult(token);
+
+var settings = new SessionPythonSettings()
+{
+    Endpoint = new Uri(endpoint),
+    SessionId = Guid.NewGuid().ToString()
+};
+
+Console.WriteLine("Hello, World!");
+
+var builder =
+    Kernel.CreateBuilder()
+    .AddOpenAIChatCompletion("gpt-3.5-turbo", apiKey);
+
+builder.Services.AddLogging(loggingBuilder => loggingBuilder.AddConsole());
+builder.Services.AddHttpClient();
+builder.Services.AddSingleton((sp)
+    => new SessionsPythonPlugin(
+        settings,
+        TokenProvider,
+        sp.GetRequiredService<IHttpClientFactory>(),
+        sp.GetRequiredService<ILoggerFactory>()));
+var kernel = builder.Build();
+
+kernel.Plugins.AddFromObject(kernel.GetRequiredService<SessionsPythonPlugin>());
+var chatCompletion = kernel.GetRequiredService<IChatCompletionService>();
+
+var chatHistory = new ChatHistory();
+
+StringBuilder fullAssistantContent = new();
+
+do
+{
+    Console.Write("\nUser: ");
+    var input = Console.ReadLine();
+    if (string.IsNullOrWhiteSpace(input)) { break; }
+
+    chatHistory.AddUserMessage(input);
+
+    Console.WriteLine("Assistant: ");
+    fullAssistantContent.Clear();
+    await foreach (var content in chatCompletion.GetStreamingChatMessageContentsAsync(
+        chatHistory,
+        new OpenAIPromptExecutionSettings { ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions },
+        kernel)
+        .ConfigureAwait(false))
+    {
+        Console.Write(content.Content);
+        fullAssistantContent.Append(content.Content);
+    }
+    chatHistory.AddAssistantMessage(fullAssistantContent.ToString());
+} while (true);
--- a/dotnet/samples/Demos/CodeInterpreterPlugin/Program.cs
+++ b/dotnet/samples/Demos/CodeInterpreterPlugin/Program.cs
@@ -30,7 +30,7 @@ var settings = new SessionPythonSettings()
     SessionId = Guid.NewGuid().ToString()
 };
 
-Console.WriteLine("Hello, World!");
+Console.WriteLine("=== Code Interpreter With Azure Container Apps Plugin Demo ===");
 
 var builder =
     Kernel.CreateBuilder()
--- a/dotnet/samples/Demos/CodeInterpreterPlugin/README.md
+++ b/dotnet/samples/Demos/CodeInterpreterPlugin/README.md
@@ -1,0 +1,33 @@
+# Semantic Kernel - Code Interpreter Plugin with Azure Container Apps
+
+This example demonstrates how to do AI Code Interpretetion using a Plugin with Azure Container Apps to execute python code in a container.
+
+## Configuring Secrets
+
+The example require credentials to access OpenAI and Azure Container Apps (ACA)
+
+If you have set up those credentials as secrets within Secret Manager or through environment variables for other samples from the solution in which this project is found, they will be re-used.
+
+### To set your secrets with Secret Manager:
+
+```
+dotnet user-secrets init
+
+dotnet user-secrets set "OpenAI:ApiKey" "..."
+
+dotnet user-secrets set "AzureContainerApps:BearerKey" " .. current token .. "
+dotnet user-secrets set "AzureContainerApps:Endpoint" " .. endpoint .. "
+```
+
+### To set your secrets with environment variables
+
+Use these names:
+
+```
+# OpenAI
+OpenAI__ApiKey
+
+# Azure Container Apps
+AzureContainerApps__BearerKey
+AzureContainerApps__Endpoint
+```
--- a/dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionPythonSettings.cs
+++ b/dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionPythonSettings.cs
@@ -1,0 +1,98 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.ComponentModel;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel.Plugins.Core.CodeInterpreter;
+
+/// <summary>
+/// Settings for a Python session.
+/// </summary>
+public class SessionPythonSettings
+{
+    /// <summary>
+    /// Determines if the input should be sanitized.
+    /// </summary>
+    [JsonIgnore]
+    public bool SanitizeInput { get; set; }
+
+    /// <summary>
+    /// The target endpoint.
+    /// </summary>
+    [JsonIgnore]
+    public Uri? Endpoint { get; init; }
+
+    /// <summary>
+    /// The session identifier.
+    /// </summary>
+    [JsonPropertyName("identifier")]
+    public string? SessionId { get; init; }
+
+    /// <summary>
+    /// Code input type.
+    /// </summary>
+    [JsonPropertyName("codeInputType")]
+    public CodeInputTypeSetting CodeInputType { get; set; }
+
+    /// <summary>
+    /// Code execution type.
+    /// </summary>
+    [JsonPropertyName("executionType")]
+    public CodeExecutionTypeSetting CodeExecutionType { get; set; }
+
+    /// <summary>
+    /// Timeout in seconds for the code execution.
+    /// </summary>
+    [JsonPropertyName("timeoutInSeconds")]
+    public int TimeoutInSeconds { get; set; } = 100;
+
+    /// <summary>
+    /// The Python code to execute.
+    /// </summary>
+    [JsonPropertyName("pythonCode")]
+    public string? PythonCode { get; set; }
+
+    /// <summary>
+    /// Code input type.
+    /// </summary>
+    [Description("Code input type.")]
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum CodeInputTypeSetting
+    {
+        /// <summary>
+        /// Code is provided as a inline string.
+        /// </summary>
+        [Description("Code is provided as a inline string.")]
+        [JsonPropertyName("inline")]
+        Inline
+    }
+
+    /// <summary>
+    /// Code input type.
+    /// </summary>
+    [Description("Code input type.")]
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum CodeExecutionTypeSetting
+    {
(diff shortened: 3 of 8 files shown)


dotnet/samples/Demos/README.md
```diff
@@ -7,4 +7,5 @@ Demonstration applications that leverage the usage of one or many SK features
 | Create Chat GPT Plugin | A simple plugin that uses OpenAI GPT-3 to chat |
 | Home Automation | This example demonstrates a few dependency injection patterns that can be used with Semantic Kernel. |
 | HuggingFace Image to Text | In this demonstration the application uses Semantic Kernel's HuggingFace ImageToText Service to fetch a descriptive analysis of the clicked image. |
-| Telemetry With Application Insights | Demo on how an application can be configured to send Semantic Kernel telemetry to Application Insights. |
\ No newline at end of file
+| Telemetry With Application Insights | Demo on how an application can be configured to send Semantic Kernel telemetry to Application Insights. |
+| Code Interpreter Plugin | A plugin that leverages Azure Container Apps service to run python code for AI Agents. |
```
alliscode, line 11: nit: Agents may be confusing since we have an agents package now.
```suggestion
| Code Interpreter Plugin | A plugin that leverages Azure Container Apps service to execute python code. |
```

dotnet/samples/Demos/CodeInterpreterPlugin/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    Endpoint = new Uri(endpoint),
+    SessionId = Guid.NewGuid().ToString()
+};
+
+Console.WriteLine("=== Code Interpreter With Azure Container Apps Plugin Demo ===");
+
+var builder =
+    Kernel.CreateBuilder()
+    .AddOpenAIChatCompletion("gpt-3.5-turbo", apiKey);
```
markwallace-microsoft, line 39: Should the model be configurable?

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+        Verify.NotNullOrWhiteSpace(code, nameof(code));
+
+        if (this._settings.SanitizeInput)
+        {
+            code = SanitizeCodeInput(code);
+        }
+
+        if (this._logger.IsEnabled(LogLevel.Information))
+        {
+            this._logger.LogInformation("Executing Python code: {Code}", code);
```
markwallace-microsoft, line 86: Should this be trace or debug logging?

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+                { "User-Agent", $"{HttpHeaderConstant.Values.UserAgent}/{s_assemblyVersion} (Language=dotnet)" }
+            },
+            Content = new StringContent(JsonSerializer.Serialize(requestBody), Encoding.UTF8, "application/json")
+        };
+
+        var response = await httpClient.SendAsync(request).ConfigureAwait(false);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            throw new HttpRequestException($"Failed to execute code. Status code: {response.StatusCode}");
```
markwallace-microsoft, line 108: Include the response body if any.

Should we also include the code?

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        var jsonElementResult = JsonSerializer.Deserialize<JsonElement>(await response.Content.ReadAsStringAsync().ConfigureAwait(false));
+
+        return $@"Result:
+{jsonElementResult.GetProperty("result").GetRawText()}
+Stdout:
+{jsonElementResult.GetProperty("stdout").GetRawText()}
+Stderr:
+{jsonElementResult.GetProperty("stderr").GetRawText()}";
```
markwallace-microsoft, line 118: Not for this PR but I wonder is there a better way to return this data e.g. as JSON

dotnet/samples/Demos/CodeInterpreterPlugin/README.md
```diff
@@ -0,0 +1,33 @@
...
+If you have set up those credentials as secrets within Secret Manager or through environment variables for other samples from the solution in which this project is found, they will be re-used.
+
+### To set your secrets with Secret Manager:
+
+```
+dotnet user-secrets init
+
+dotnet user-secrets set "OpenAI:ApiKey" "..."
+
+dotnet user-secrets set "AzureContainerApps:BearerKey" " .. current token .. "
```
moonbox3, line 18: I am curious what is the bearer key is for? The dotnet plugin has an auth callback, correct?

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+        // Response Body {"$id":"1","$values":[{"$id":"2","filename":"test-file.txt","size":17,"last_modified_time":"2024-05-07T22:24:00.7528208Z"}]}
+
+        return JsonSerializer.Deserialize<SessionRemoteFileMetadata>(JsonElementResult.GetProperty("$values")[0].GetRawText())!;
+    }
+
+    /// <summary>
+    /// Downloads a file from the current Session ID.
+    /// </summary>
+    /// <param name="remoteFilePath"> The path to download the file from, relative to `/mnt/data`. </param>
+    /// <param name="localFilePath"> The path to save the downloaded file to. If not provided, the file is returned as a BufferedReader. </param>
```
alliscode, line 173: nit: byte[]

alliscode (approved):

dotnet/samples/Demos/CodeInterpreterPlugin/README.md
RogerBarreto: It does and the demo uses the callback to provide the token from the configuration. In a normal scenario this would be handled by the user on acquiring the token.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    /// <summary>
+    /// Executes the provided Python code.
+    /// Start and end the code snippet with double quotes to define it as a string.
+    /// Insert \n within the string wherever a new line should appear.
+    /// Add spaces directly after \n sequences to replicate indentation.
+    /// Use \"" to include double quotes within the code without ending the string.
+    /// Keep everything in a single line; the \n sequences will represent line breaks
+    /// when the string is processed or displayed
```
markwallace-microsoft, line 62: nite: Missing full stop

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        var response = await httpClient.SendAsync(request).ConfigureAwait(false);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            throw new HttpRequestException($"Failed to upload file. Status code: {response.StatusCode}");
+        }
+
+        var JsonElementResult = JsonSerializer.Deserialize<JsonElement>(await response.Content.ReadAsStringAsync().ConfigureAwait(false));
+        // Response Body {"$id":"1","$values":[{"$id":"2","filename":"test-file.txt","size":17,"last_modified_time":"2024-05-07T22:24:00.7528208Z"}]}
```
markwallace-microsoft, line 164: Remove commented out code

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+            var fileContent = await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+            try
+            {
+                File.WriteAllBytes(localFilePath, fileContent);
+                return null;
+            }
+            catch (Exception ex)
+            {
+                throw new InvalidOperationException("Failed to write file to disk.", ex);
+            }
```
markwallace-microsoft, line 208: Return using the `fileContent` rather than read these again

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        using var httpClient = this._httpClientFactory.CreateClient();
+        httpClient.DefaultRequestHeaders.Add("Authorization", $"Bearer {(await this.GetAuthTokenAsync().ConfigureAwait(false))}");
+        httpClient.DefaultRequestHeaders.Add("User-Agent", $"{HttpHeaderConstant.Values.UserAgent}/{s_assemblyVersion} (Language=dotnet)");
+
+        var response = await httpClient.GetAsync($"{this._poolManagementEndpoint}python/files?identifier={this._settings.SessionId}").ConfigureAwait(false);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            throw new HttpRequestException($"Failed to list files. Status code: {response.StatusCode}");
```
markwallace-microsoft, line 234: Include response body (if any)

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        using var httpClient = this._httpClientFactory.CreateClient();
+        httpClient.DefaultRequestHeaders.Add("Authorization", $"Bearer {(await this.GetAuthTokenAsync().ConfigureAwait(false))}");
+        httpClient.DefaultRequestHeaders.Add("User-Agent", $"{HttpHeaderConstant.Values.UserAgent}/{s_assemblyVersion} (Language=dotnet)");
+
+        var response = await httpClient.GetAsync($"{this._poolManagementEndpoint}python/downloadFile?identifier={this._settings.SessionId}&filename={remoteFilePath}").ConfigureAwait(false);
+        if (!response.IsSuccessStatusCode)
+        {
+            throw new HttpRequestException($"Failed to download file. Status code: {response.StatusCode}");
```
markwallace-microsoft, line 194: Include response body (if any) as it may contain additional diagnostic information

markwallace-microsoft (commented):
The code has no unit tests

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
RogerBarreto: Moved to tracing, @moonbox3 might consider moving this too on python impl.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
RogerBarreto: Added the response body for clarity.

The code is provided in the log trace above. Didn't thought it would be necessary to provide again in the exception.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
RogerBarreto: Probably there's, I didn't want to change much from what was done in the python code, as it works.

dotnet/samples/Demos/CodeInterpreterPlugin/README.md
RogerBarreto: Changed the provider the use Azure.Identity to get the token and to make this closer to a real world scenario.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionPythonSettings.cs
```diff
@@ -0,0 +1,98 @@
...
+using System;
+using System.ComponentModel;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.SemanticKernel.Plugins.Core.CodeInterpreter;
+
+/// <summary>
+/// Settings for a Python session.
+/// </summary>
+public class SessionPythonSettings
```
westey-m, line 12: I see Endpoint is nullable, but I would expect it to be a required setting for the service to work.  is it like this because we have one class for both public settings and what we are passing internally to the service?  If so, should we split it, so that it's more obvious to users what settings they have to provide at programming time, rather than run time?

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,304 @@
...
+    public SessionsPythonPlugin(SessionPythonSettings settings, Func<Task<string>> authTokenProvider, IHttpClientFactory httpClientFactory, ILoggerFactory? loggerFactory = null)
+    {
+        Verify.NotNull(settings, nameof(settings));
+        Verify.NotNull(authTokenProvider, nameof(authTokenProvider));
+        Verify.NotNull(httpClientFactory, nameof(httpClientFactory));
+        Verify.NotNull(settings.Endpoint, nameof(settings.Endpoint));
+
+        this._settings = settings;
+
+        // Ensure the endpoint won't change by reference
```
westey-m, line 47: FYI, The Endpoint prop is init only, and Uri is immutable, so changing by reference shouldn't be possible in this case.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,304 @@
...
+        using var httpClient = this._httpClientFactory.CreateClient();
+        var requestBody = new
+        {
+            properties = this._settings.CloneForRequest(code)
+        };
+
+        using var request = new HttpRequestMessage(HttpMethod.Post, this._poolManagementEndpoint + "python/execute")
+        {
+            Headers = {
+                { "Authorization", $"Bearer {(await this.GetAuthTokenAsync().ConfigureAwait(false))}" },
```
westey-m, line 98: It's possible to add an auth header via a DelegatingHandler registered on the http client, so we don't necessarily need to provide that capability here, although I suppose requiring someone to implement their own DelegatingHandler may be a barrier to entry.  On the other hand, it may be annoying to anyone who has their own delegating handlers setup already for generating auth tokens.  Maybe consider making GetAuthTokenAsync optional, so that anyone who is using their own DelegatingHandler don't have to use this.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
```diff
@@ -0,0 +1,304 @@
...
+
+        return result;
+    }
+
+    /// <summary>
+    /// Retrieves the auth token for the session.
+    /// </summary>
+    private async Task<string> GetAuthTokenAsync()
+    {
+        if (string.IsNullOrWhiteSpace(this._authToken))
```
westey-m, line 258: Consider leaving token caching to the authTokenProvider, since they will know best how long the session will last, and whether the token may need to be renewed at some stage.

westey-m (approved):

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
RogerBarreto: Endpoint shouldn't change, moved those to the constructor.

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionsPythonPlugin.cs
RogerBarreto: Made the delegate as optional

dotnet/src/Plugins/Plugins.Core/CodeInterpreter/SessionPythonSettings.cs
RogerBarreto: Moved to be a required item.

Splitted the Model for request from the Plugin settings, to give a better idea of what is a setting in the plugin level.

alliscode (approved):

westey-m (approved):

PR #6171, opened 2024-05-09
.Net: Example of prompt PII detection logic with Filters and Microsoft Presidio

dmytrostruk:
### Motivation and Context

This example shows how to implement Personal Identifiable Information (PII) detection logic with Filters using Microsoft Presidio service: https://github.com/microsoft/presidio.

Example contains two filters:

First filter is responsible for PII detection in prompt by using Presidio Text Analyzer. If PII is detected, then the prompt won't be sent to LLM and custom result will be returned after function invocation.

Example output:
```
Prompt: John Smith has a card 1111 2222 3333 4444
Entity type: CREDIT_CARD. Score: 1
Entity type: PERSON. Score: 0.85
Exception: Prompt contains PII information. Operation is canceled.
```

Second filter is responsible for PII detection in prompt and updating the prompt by following specified rules before sending it to LLM. This filter uses Presidio Text Analyzer and Presidio Text Anonymizer. 

Example output:
```
LLM instructions: If prompt does not contain first and last name - return 'true'.
Prompt before anonymization : Hello world, my name is Jane Doe. My number is: 034453334
Prompt after anonymization : Hello world, my name is ANONYMIZED. My number is: 
Result: true
```

### Contribution Checklist

### DIFF
--- a/dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
+++ b/dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
@@ -446,7 +446,7 @@ public class PIIDetectionWithFilters(ITestOutputHelper output) : BaseTest(output
     /// <summary>
     /// Contact model for demonstration purposes.
     /// </summary>
-    private class Contact
+    private sealed class Contact
     {
         public string Name { get; set; }
         public string Phone { get; set; }
@@ -460,11 +460,12 @@ public class PIIDetectionWithFilters(ITestOutputHelper output) : BaseTest(output
     {
         [KernelFunction]
         public List<Contact> GetContacts()
-            => [
+            =>
+            [
                 new () { Name = "John Smith", Phone = "+1 (123) 456-7890", Position = "Developer" },
                 new () { Name = "Alice Doe", Phone = "+1 (987) 654-3120", Position = "Manager" },
                 new () { Name = "Emily Davis", Phone = "+1 (555) 555-5555", Position = "Designer" }
-                ];
+            ];
     }
 
     #endregion
--- a/dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
+++ b/dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
@@ -6,13 +6,14 @@ using System.Text.Json.Serialization;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
 
 namespace Filtering;
 
 public class PIIDetectionWithFilters(ITestOutputHelper output) : BaseTest(output)
 {
     [Fact]
-    public async Task PIIAnalyzerAsync()
+    public async Task PromptAnalyzerAsync()
     {
         var builder = Kernel.CreateBuilder();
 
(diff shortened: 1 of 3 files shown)


dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
```diff
@@ -0,0 +1,404 @@
...
+            TestConfiguration.AzureOpenAI.ChatDeploymentName,
+            TestConfiguration.AzureOpenAI.Endpoint,
+            TestConfiguration.AzureOpenAI.ApiKey);
+
+        // Add logging
+        var logger = this.LoggerFactory.CreateLogger<PIIDetectionWithFilters>();
+        builder.Services.AddSingleton<ILogger>(logger);
+
+        // Add Microsoft Presidio Text Analyzer service and configure HTTP client for it. Text Analyzer results are required for Text Anonymizer input.
+        builder.Services.AddHttpClient<PresidioTextAnalyzerService>(client => { client.BaseAddress = new Uri("http://localhost:5001"); });
```
westey-m, line 92: You've already registered this above.

westey-m (approved):

SergeyMenshykh (approved):

dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
```diff
@@ -0,0 +1,404 @@
...
+        // Example 1: Use prompt with PII
+        var result1 = await kernel.InvokePromptAsync("John Smith drivers license is AC432223");
+        logger.LogInformation("Result: {Result}", result1.ToString());
+
+        /* 
+        Prompt: John Smith drivers license is AC432223
+        Entity type: PERSON. Score: 0.85
+        Entity type: US_DRIVER_LICENSE. Score: 0.6499999999999999
+        Result: Prompt contains PII information. Operation is canceled. 
+        */
```
markwallace-microsoft, line 60: Capturing the data about what went wrong but I'm unsure about returning it as a normal result.

If the goal is to detect PII and then have the operation fail it might be better to throw an exception with the same information. At the moment things fail silently.

Also in a real world sample using a credit card number might be better because it's common to want to send the users name to the LLM (score 0.85) but very unlikely to want to send a credit card number. So it would be useful to show a threshold where user names would pass but credit card numbers would fail.

dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
```diff
@@ -0,0 +1,404 @@
...
+
+        var kernel = builder.Build();
+
+        // Define instructions for LLM how to react when certain conditions are met for demonstration purposes
+        var executionSettings = new OpenAIPromptExecutionSettings
+        {
+            ChatSystemPrompt = "If prompt does not contain first and last name - return 'true'."
+        };
+
+        var result = await kernel.InvokePromptAsync("Hello world, my name is Jane Doe. My number is: 034453334", new(executionSettings));
```
markwallace-microsoft, line 119: A more compelling demonstration of anonymizing peoples names would be for a case where you are uploading customer data. How about having a prompt which inserts data from a function which simulates reading customer data from a database and you want to anonymise the names and redact all of the phone numbers.

markwallace-microsoft (commented):
Not for this PR but the following samples would be great:

1. Detecting passwords or secrets in a prompt
2. An image redaction sample

dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
dmytrostruk: Correct, but this is different test with new kernel builder. I was thinking to put it in common place and re-use, but this example file is already big enough with Presidio HTTP service clients, so I decided to keep it simple and configure it again. If there will be a 3rd test with the same logic, I will definitely put it in common place to respect Rule of Three :)

dotnet/samples/Concepts/Filtering/PIIDetectionWithFilters.cs
dmytrostruk: > Capturing the data about what went wrong but I'm unsure about returning it as a normal result. If the goal is to detect PII and then have the operation fail it might be better to throw an exception with the same information. At the moment things fail silently.

Good point, I changed the logic to throw an exception instead. Although I left a comment that it's possible to override `context.Result` to avoid breaking function execution. I believe it still should be useful for automatic function calling scenarios.

> Also in a real world sample using a credit card number might be better because it's common to want to send the users name to the LLM (score 0.85) but very unlikely to want to send a credit card number. So it would be useful to show a threshold where user names would pass but credit card numbers would fail.

Done, I used fake credit card number as an example. The threshold is 0.9 now, and user name with score 0.85 passes while credit card with score 1 fails.

markwallace-microsoft (approved):

PR #6177, opened 2024-05-09
Azure CosmosDB MongoDB vCore Memory Store Bug Fixes

aayush3011:
### Motivation and Context
Fixing some issues with config and memory store for Azure CosmosDB MongoDB vCore.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBConfig.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBConfig.cs
@@ -82,5 +82,6 @@ public class AzureCosmosDBMongoDBConfig
         this.NumberOfConnections = 16;
         this.EfConstruction = 64;
         this.EfSearch = 40;
+        this.Dimensions = 1536;
     }
 }
--- a/dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStore.cs
+++ b/dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStore.cs
@@ -48,16 +48,15 @@ public class AzureCosmosDBMongoDBMemoryStore : IMemoryStore, IDisposable
     /// and other properties required for vector search.
     /// </summary>
     public AzureCosmosDBMongoDBMemoryStore(
-        IMongoClient mongoClient,
+        MongoClient mongoClient,
         string databaseName,
         AzureCosmosDBMongoDBConfig config
     )
     {
-        MongoClientSettings settings = mongoClient.Settings;
         this._config = config;
-        settings.ApplicationName = this._config.ApplicationName;
-        this._mongoClient = new MongoClient(settings);
+        this._mongoClient = mongoClient;
         this._mongoDatabase = this._mongoClient.GetDatabase(databaseName);
+        
     }
 
     /// <inheritdoc/>
--- a/dotnet/src/IntegrationTests/Connectors/Memory/AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStoreTestsFixture.cs
+++ b/dotnet/src/IntegrationTests/Connectors/Memory/AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStoreTestsFixture.cs
@@ -32,10 +32,13 @@ public class AzureCosmosDBMongoDBMemoryStoreTestsFixture : IAsyncLifetime
         var connectionString = GetSetting(configuration, "ConnectionString");
         this.DatabaseName = "DotNetSKTestDB";
         this.CollectionName = "DotNetSKTestCollection";
+        AzureCosmosDBMongoDBConfig config = new AzureCosmosDBMongoDBConfig();
+        config.Dimensions = 3;
+        
         this.MemoryStore = new AzureCosmosDBMongoDBMemoryStore(
             connectionString,
             this.DatabaseName,
-            new AzureCosmosDBMongoDBConfig()
+            config
         );
     }
 


dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStore.cs
```diff
@@ -48,15 +48,13 @@ AzureCosmosDBMongoDBConfig config
...
+        MongoClient mongoClient,
         string databaseName,
         AzureCosmosDBMongoDBConfig config
     )
     {
-        MongoClientSettings settings = mongoClient.Settings;
         this._config = config;
-        settings.ApplicationName = this._config.ApplicationName;
-        this._mongoClient = new MongoClient(settings);
+        this._mongoClient = mongoClient;
```
dmytrostruk, line 57: Previous logic worked because even though user passes `mongoClient`, we created our own instance of `mongoClient` based on settings. Now, we are just setting it as it is, and this is a problem, because we have a logic that disposes `mongoClient.Cluster`. The rule is - if we didn't create it, we should not dispose it. You can check how it's handled in Kusto or Milvus connectors.

dotnet/src/IntegrationTests/Connectors/Memory/AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStoreTestsFixture.cs
```diff
@@ -32,17 +32,19 @@ public AzureCosmosDBMongoDBMemoryStoreTestsFixture()
         var connectionString = GetSetting(configuration, "ConnectionString");
         this.DatabaseName = "DotNetSKTestDB";
         this.CollectionName = "DotNetSKTestCollection";
+        AzureCosmosDBMongoDBConfig config = new AzureCosmosDBMongoDBConfig();
+        config.Dimensions = 3;
```
dmytrostruk, line 36:
```suggestion
        AzureCosmosDBMongoDBConfig config = new() { Dimensions = 3 };
```

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBConfig.cs
```diff
@@ -73,5 +73,21 @@ public class AzureCosmosDBMongoDBConfig(int dimensions)
     /// The size of the dynamic candidate list for search (40 by default). A higher value provides better recall at
     /// the cost of speed.
     /// </summary>
-    public int EfSearch { get; set; } = 40;
+    public int EfSearch { get; set; }
+
+    /// <summary>
+    /// Initialize the AzureCosmosDBMongoDBConfig with default values
+    /// </summary>
+    public AzureCosmosDBMongoDBConfig()
```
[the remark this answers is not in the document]
aayush3011: I updated the PR.

dotnet/src/IntegrationTests/Connectors/Memory/AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStoreTests.cs
```diff
@@ -86,6 +86,13 @@ public async Task ItCanGetNearestMatchesAsync(int limit, bool withEmbeddings)
         var memoryStore = this._fixture.MemoryStore;
         var searchEmbedding = DataHelper.VectorSearchTestEmbedding;
         var nearestMatchesExpected = DataHelper.VectorSearchExpectedResults;
+        var records = DataHelper.VectorSearchTestRecords;
+
+        await memoryStore.CreateCollectionAsync(collectionName);
```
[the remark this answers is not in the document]
aayush3011: I updated the PR.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStore.cs
aayush3011: Fixed in next iteration.

dotnet/src/IntegrationTests/Connectors/Memory/AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStoreTestsFixture.cs
aayush3011: I updated the PR.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStore.cs
dmytrostruk: That makes sense, but since with the new code we are using the client that was passed in constructor, we cannot dispose it now in `Dispose` method. The following logic should be added:
1. When user passes `connectionString` in constructor - we create our own client, use it for operations and we dispose it.
2. When user passes instance of `MongoClient` - we use this client for operations, and we do not dispose it.

aayush3011: Adding a boolean global flag **_disposing** which is set True when user passes connectionString 

1. Set to True  - When user passes `connectionString` in constructor
2. Set to False - When user passes instance of `MongoClient`

And use this flag in the Dispose method call.

dmytrostruk: Sounds good, thank you. By the way, no need to set it to `false` explicitly, since it's `false` by default. Setting it to `true` when user passes `connectionString` should be enough. Thanks again!

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBMongoDB/AzureCosmosDBMongoDBMemoryStore.cs
```diff
@@ -318,7 +318,10 @@ protected virtual void Dispose(bool disposing)
     {
         if (disposing)
         {
-            this._mongoClient.Cluster.Dispose();
+            if (this._ownsMongoClient)
+            {
+                this._mongoClient.Cluster.Dispose();
+            }
         }
```
dmytrostruk, line 325: You can combine 2 conditions here:
```csharp
if (disposing && this._ownsMongoClient)
{
    this._mongoClient.Cluster.Dispose();
}
```

dmytrostruk (approved):

markwallace-microsoft (approved):

PR #6196, opened 2024-05-12
.Net: Cross-language tests

glahaye:
### Motivation and Context
We want to ensure we have the same functionality across C#, Python and Java.

### Description
These tests generate the requests sent to OpenAI from SK in C#. SK in Python and Java can compare against these requests and verify that they are exactly the same.

### Contribution Checklist


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -325,7 +325,7 @@ await foreach (Completions completions in response.ConfigureAwait(false))
             // Make the request.
             var responseData = (await RunRequestAsync(() => this.Client.GetChatCompletionsAsync(chatOptions, cancellationToken)).ConfigureAwait(false)).Value;
             this.CaptureUsageDetails(responseData.Usage);
-            if (responseData.Choices.Count == 0)
+            if (responseData.Choices is null || responseData.Choices.Count == 0)
```
stephentoub, line 328: In what situation does this end up being null?

glahaye: When receiving invalid data back from OpenAI following a request. I was able to simulate this happening by providing a custom HttpMessageHandler to a custom HttpClient so I'll grant you this shouldn't happen much normally.

Also, it looks like the latest version of this file doesn't have this line so this might be a moot point after I re-synch.

moonbox3 (approved):

alliscode (approved):

PR #6200, opened 2024-05-12
.Net: Time Plugin Demo

RogerBarreto:
# Time Plugin - Demo Application

This is an example how you can easily use Plugins with the Power of Auto Function Calling from AI Models. 

Here we have a simple Time Plugin created in C# that can be called from the AI Model to get the current time.


dotnet/SK-dotnet.sln
```diff
@@ -294,14 +294,16 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "PromptTemplates.Liquid.Unit
...
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ContentSafety", "samples\Demos\ContentSafety\ContentSafety.csproj", "{6EF9663D-976C-4A27-B8D3-8B1E63BA3BF2}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ContentSafety", "samples\Demos\ContentSafety\ContentSafety.csproj", "{6EF9663D-976C-4A27-B8D3-8B1E63BA3BF2}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Concepts", "samples\Concepts\Concepts.csproj", "{925B1185-8B58-4E2D-95C9-4CA0BA9364E5}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FunctionInvocationApproval", "samples\Demos\FunctionInvocationApproval\FunctionInvocationApproval.csproj", "{6B56D8EE-9991-43E3-90B2-B8F5C5CE77C2}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CodeInterpreterPlugin", "samples\Demos\CodeInterpreterPlugin\CodeInterpreterPlugin.csproj", "{3ED53702-0E53-473A-A0F4-645DB33541C2}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TimePlugin", "TimePlugin\TimePlugin.csproj", "{F312FCE1-12D7-4DEF-BC29-2FF6618509F3}"
```
dmytrostruk, line 305: Looks like location is incorrect both in file system and .NET solution.

dotnet/TimePlugin/Program.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
dmytrostruk, line 1: Do we need separate project for this example? Can we add it as integration test under `Concepts` folder? If the point is to be able to accept user input (like from Console in this case), maybe it's worth to add it to .NET notebooks?

dotnet/TimePlugin/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+public class TimeInformationPlugin
+{
+    /// <summary>
+    /// Retrieves the current time in UTC.
+    /// </summary>
+    /// <returns>The current time in UTC. </returns>
+    [KernelFunction, Description("Retrieves the current time in UTC.")]
+    public string GetCurrentUtcTime()
+        => DateTime.UtcNow.ToString("R");
+}
```
stephentoub, line 66: This is what I do for lots of my own demos. But it's strange to have this example when there's still the TimePlugin in the src. Is that going away? Will that be replaced with a simpler one like this?

RogerBarreto: I added to the source to simplify the Demo scope and dependencies as this is intended to be a very simple and small demo for Blog post.

dotnet/TimePlugin/Program.cs
RogerBarreto: We already have, I'm adding this Demo as an ask from @sophialagerkranspandey, to support a blog post on how simple is to use a Plugin with Auto Function Calling with minimum setup effort.

dotnet/SK-dotnet.sln
RogerBarreto: Fixed thanks!

dotnet/TimePlugin/Program.cs
markwallace-microsoft, line 1: Should this be in the samples\Demos folder?

markwallace-microsoft (commented):
@RogerBarreto is there a reason this is not being created as a Learn site sample and then having a Blog post pointing to it?
In the Learn site we can have a tutorial that shows this sample in each language and tutorials posted there will be more discoverable.

dotnet/TimePlugin/Program.cs
RogerBarreto: Sure, good catch.

RogerBarreto:
@sophialagerkranspandey ^^^

As I answered into Dmytro question, the overall idea is that a Demo is a small and well defined program that all you have to do is to configure OpenAI Key and execute it without having knowledge about Unit testing to see the output.

markwallace-microsoft (approved):

dmytrostruk (approved):

PR #6218, opened 2024-05-13
.Net: Enable CreateFromType/Object to work with closed generics

stephentoub:
https://github.com/microsoft/semantic-kernel/pull/6206#issuecomment-2107167732


dotnet/src/SemanticKernel.Core/Functions/KernelPluginFactory.cs
```diff
@@ -101,4 +101,18 @@ public static KernelPlugin CreateFromObject(object target, string? pluginName =
...
+    {
+        string name = type.Name;
+
+        int genericParamIndex = name.IndexOf('`');
+        if (genericParamIndex >= 0)
+        {
+            name = name.Substring(0, genericParamIndex);
+        }
+
+        return KernelFunctionFromMethod.SanitizeMetadataName(name);
```
dmytrostruk, line 116: Do we need to include generic type name in plugin name, so LLM can work with same plugin, but different generic type?
```csharp

builder.Plugins.AddFromType<GenericPlugin<int>>();
builder.Plugins.AddFromType<GenericPlugin<string>>();

var kernel = builder.Build();
```

dmytrostruk: I guess it will work with custom plugin name, so maybe we can add support for it when using `AddFromType`. 
```csharp

var intPlugin = new GenericPlugin<int>();
var stringPlugin = new GenericPlugin<string>();

builder.Plugins.AddFromObject(intPlugin, "GenericPluginInt");
builder.Plugins.AddFromObject(stringPlugin, "GenericPluginString");

var kernel = builder.Build();
```

dmytrostruk (approved):

RogerBarreto:
Suggest adding an example on how we can retrieve a generic in the `Concepts\Functions\KernelMethod_Types.cs`.

I may be trying the open generics here, but would be super nice if we could use that as well, without necessarily specifying the `T` at the class level.

```csharp
public class MyPlugin
{
    [KernelFunction]
    public T GenericResult<T>(T input) => input;

    [KernelFunction]
    public Task<T> GenericResultAsync<T>(T input) => Task.FromResult(input);
}
```

stephentoub:
> I may be trying the open generics here, but would be super nice if we could use that as well, without necessarily specifying the T at the class level.

This won't work today (even with my change) and is more complicated.

dotnet/src/SemanticKernel.Core/Functions/KernelPluginFactory.cs
stephentoub: Updated how the name is derived.

dmytrostruk (approved):

markwallace-microsoft (approved):

PR #6223, opened 2024-05-13
.Net: Add support for tool choice none and any

markwallace-microsoft:
### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
+++ b/dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
@@ -62,27 +62,24 @@ public abstract class MistralAIToolCallBehavior
     /// <returns>
     /// The <see cref="MistralAIToolCallBehavior"/> that may be set into <see cref="MistralAIPromptExecutionSettings.ToolCallBehavior"/>
     /// to indicate that the specified functions should be made available to the model.
+    /// The model is forced to call a function from the list of functions provided.
     /// </returns>
-    public static MistralAIToolCallBehavior EnableFunctions(IEnumerable<KernelFunction> functions, bool autoInvoke = false)
+    public static MistralAIToolCallBehavior RequiredFunctions(IEnumerable<KernelFunction> functions, bool autoInvoke = false)
     {
         Verify.NotNull(functions);
-        return new EnabledFunctions(functions, autoInvoke);
+        return new AnyFunction(functions, autoInvoke);
     }
 
-    /*
-    /// <summary>Gets an instance that will request the model to use the specified function.</summary>
-    /// <param name="function">The function the model should request to use.</param>
-    /// <param name="autoInvoke">true to attempt to automatically handle function call requests; otherwise, false.</param>
-    /// <returns>
-    /// The <see cref="MistralAIToolCallBehavior"/> that may be set into <see cref="MistralAIPromptExecutionSettings.ToolCallBehavior"/>
-    /// to indicate that the specified function should be requested by the model.
-    /// </returns>
-    public static MistralAIToolCallBehavior RequireFunction(KernelFunction function, bool autoInvoke = false)
-    {
-        Verify.NotNull(function);
-        return new RequiredFunction(function, autoInvoke);
-    }
-    */
+    /// <summary>
+    /// Gets an instance that will both provide all of the <see cref="Kernel"/>'s plugins' function information
+    /// to the model but not any function call requests.
+    /// </summary>
+    /// <remarks>
+    /// When successful, tool call requests from the model become an implementation detail, with the service
+    /// handling invoking any requested functions and supplying the results back to the model.
+    /// If no <see cref="Kernel"/> is available, no function information will be provided to the model.
+    /// </remarks>
+    public static MistralAIToolCallBehavior NoKernelFunctions { get; } = new NoneKernelFunctions();
 
     /// <summary>Initializes the instance; prevents external instantiation.</summary>
     private MistralAIToolCallBehavior(bool autoInvoke)
@@ -140,10 +137,20 @@ public abstract class MistralAIToolCallBehavior
         internal override void ConfigureRequest(Kernel? kernel, ChatCompletionRequest request)
         {
             var functionsMetadata = kernel?.Plugins?.GetFunctionsMetadata();
-
             if (functionsMetadata is null)
             {
                 return;
+            }
+
+            // If auto-invocation is specified, we need a kernel to be able to invoke the functions.
+            // Lack of a kernel is fatal: we don't want to tell the model we can handle the functions
+            // and then fail to do so, so we fail before we get to that point. This is an error
+            // on the consumers behalf: if they specify auto-invocation with any functions, they must
+            // specify the kernel and the kernel must contain those functions.
+            bool autoInvoke = base.MaximumAutoInvokeAttempts > 0;
+            if (autoInvoke && kernel is null)
+            {
+                throw new KernelException($"Auto-invocation with {nameof(KernelFunctions)} is not supported when no kernel is provided.");
             }
 
             request.ToolChoice = "auto";
@@ -160,17 +167,11 @@ public abstract class MistralAIToolCallBehavior
     /// <summary>
     /// Represents a <see cref="MistralAIToolCallBehavior"/> that provides a specified list of functions to the model.
     /// </summary>
-    internal sealed class EnabledFunctions : MistralAIToolCallBehavior
+    internal sealed class AnyFunction(IEnumerable<KernelFunction> functions, bool autoInvoke) : MistralAIToolCallBehavior(autoInvoke)
     {
-        private readonly IEnumerable<KernelFunctionMetadata>? _kernelFunctionMetadata;
-        //private readonly ChatCompletionsFunctionToolDefinition[] _functions;
+        private readonly IEnumerable<KernelFunctionMetadata>? _kernelFunctionMetadata = functions.Select(f => f.Metadata);
 
-        public EnabledFunctions(IEnumerable<KernelFunction> functions, bool autoInvoke) : base(autoInvoke)
-        {
-            this._kernelFunctionMetadata = functions.Select(f => f.Metadata);
-        }
-
-        public override string ToString() => $"{nameof(EnabledFunctions)}(autoInvoke:{this.MaximumAutoInvokeAttempts != 0}): {string.Join(", ", this._kernelFunctionMetadata.Select(f => f.Name))}";
+        public override string ToString() => $"{nameof(AnyFunction)}(autoInvoke:{this.MaximumAutoInvokeAttempts != 0}): {string.Join(", ", this._kernelFunctionMetadata.Select(f => f.Name))}";
 
         internal override void ConfigureRequest(Kernel? kernel, ChatCompletionRequest request)
         {
@@ -179,16 +180,15 @@ public abstract class MistralAIToolCallBehavior
                 return;
             }
 
-            bool autoInvoke = base.MaximumAutoInvokeAttempts > 0;
-
             // If auto-invocation is specified, we need a kernel to be able to invoke the functions.
             // Lack of a kernel is fatal: we don't want to tell the model we can handle the functions
             // and then fail to do so, so we fail before we get to that point. This is an error
             // on the consumers behalf: if they specify auto-invocation with any functions, they must
             // specify the kernel and the kernel must contain those functions.
+            bool autoInvoke = base.MaximumAutoInvokeAttempts > 0;
             if (autoInvoke && kernel is null)
             {
-                throw new KernelException($"Auto-invocation with {nameof(EnabledFunctions)} is not supported when no kernel is provided.");
+                throw new KernelException($"Auto-invocation with {nameof(RequiredFunctions)} is not supported when no kernel is provided.");
             }
 
             foreach (var metadata in this._kernelFunctionMetadata)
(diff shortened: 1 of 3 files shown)


dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
```diff
@@ -62,27 +62,24 @@ public abstract class MistralAIToolCallBehavior
...
+    /// <summary>
+    /// Gets an instance that will both provide all of the <see cref="Kernel"/>'s plugins' function information
+    /// to the model but not any function call requests.
+    /// </summary>
+    /// <remarks>
+    /// When successful, tool call requests from the model become an implementation detail, with the service
+    /// handling invoking any requested functions and supplying the results back to the model.
+    /// If no <see cref="Kernel"/> is available, no function information will be provided to the model.
+    /// </remarks>
+    public static MistralAIToolCallBehavior NoKernelFunctions { get; } = new NoneKernelFunctions();
```
dmytrostruk, line 82: Should we call it `NoneKernelFunctions` to align it with `none` tool choice behavior name?

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
```diff
@@ -179,16 +180,15 @@ internal override void ConfigureRequest(Kernel? kernel, ChatCompletionRequest re
...
             // If auto-invocation is specified, we need a kernel to be able to invoke the functions.
             // Lack of a kernel is fatal: we don't want to tell the model we can handle the functions
             // and then fail to do so, so we fail before we get to that point. This is an error
             // on the consumers behalf: if they specify auto-invocation with any functions, they must
             // specify the kernel and the kernel must contain those functions.
+            bool autoInvoke = base.MaximumAutoInvokeAttempts > 0;
             if (autoInvoke && kernel is null)
             {
-                throw new KernelException($"Auto-invocation with {nameof(EnabledFunctions)} is not supported when no kernel is provided.");
+                throw new KernelException($"Auto-invocation with {nameof(RequiredFunctions)} is not supported when no kernel is provided.");
```
dmytrostruk, line 191: Here, for exception message we use `RequiredFunctions`, but in `ToString` override we use `AnyFunction`, which looks a little bit inconsistent. Would it be possible to use only one name?

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
```diff
@@ -163,7 +189,31 @@ public async Task ValidateGetChatMessageContentsWithAutoInvokeAsync()
...
+        };
+        var response = await service.GetChatMessageContentsAsync(chatHistory, executionSettings, kernel);
+
+        // Assert
+        Assert.NotNull(response);
+        Assert.Single(response);
+        Assert.Contains("GetWeather", response[0].Content, System.StringComparison.Ordinal);
+    }
+
+    [Fact] // (Skip = "This test is for manual verification.")
```
dmytrostruk, line 215:
```suggestion
    [Fact]
```

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
```diff
@@ -172,6 +222,32 @@ public async Task ValidateGetChatMessageContentsWithAutoInvokeAndInvocationFilte
...
+        // Assert
+        Assert.NotNull(response);
+        Assert.Single(response);
+        Assert.Equal(3, chatHistory.Count);
+        Assert.Equal(2, chatHistory[1].Items.Count);
+        Assert.True(chatHistory[1].Items[1] is FunctionCallContent);
+        Assert.Equal("GetWeather", ((FunctionCallContent)chatHistory[1].Items[1]).FunctionName);
+    }
+
+    [Fact] // (Skip = "This test is for manual verification.")
```
dmytrostruk, line 242:
```suggestion
    [Fact]
```

dmytrostruk (approved):

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
```diff
@@ -62,27 +62,24 @@ public abstract class MistralAIToolCallBehavior
...
     /// The <see cref="MistralAIToolCallBehavior"/> that may be set into <see cref="MistralAIPromptExecutionSettings.ToolCallBehavior"/>
     /// to indicate that the specified functions should be made available to the model.
+    /// The model is forced to call a function from the list of functions provided.
     /// </returns>
-    public static MistralAIToolCallBehavior EnableFunctions(IEnumerable<KernelFunction> functions, bool autoInvoke = false)
+    public static MistralAIToolCallBehavior RequiredFunctions(IEnumerable<KernelFunction> functions, bool autoInvoke = false)
     {
         Verify.NotNull(functions);
-        return new EnabledFunctions(functions, autoInvoke);
+        return new AnyFunction(functions, autoInvoke);
```
RogerBarreto, line 70: When using Required are you using Auto? Shoudn't it be `RequiredFunction` ?

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
RogerBarreto, line 82: I would keep the same same, either `No` for both or `None`
```suggestion
    public static MistralAIToolCallBehavior NoneKernelFunctions { get; } = new NoneKernelFunctions();
```

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
```diff
@@ -62,27 +62,24 @@ public abstract class MistralAIToolCallBehavior
...
-    /// </returns>
-    public static MistralAIToolCallBehavior RequireFunction(KernelFunction function, bool autoInvoke = false)
-    {
-        Verify.NotNull(function);
-        return new RequiredFunction(function, autoInvoke);
-    }
-    */
+    /// <summary>
+    /// Gets an instance that will both provide all of the <see cref="Kernel"/>'s plugins' function information
+    /// to the model but not any function call requests.
```
RogerBarreto, line 75: Are you sure when is `None` we send the list of functions still?
```suggestion
    /// to the model but will not perform any call to them.
```

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
```diff
@@ -140,12 +137,22 @@ internal sealed class KernelFunctions : MistralAIToolCallBehavior
...
             {
                 return;
             }
 
+            // If auto-invocation is specified, we need a kernel to be able to invoke the functions.
+            // Lack of a kernel is fatal: we don't want to tell the model we can handle the functions
+            // and then fail to do so, so we fail before we get to that point. This is an error
+            // on the consumers behalf: if they specify auto-invocation with any functions, they must
+            // specify the kernel and the kernel must contain those functions.
+            bool autoInvoke = base.MaximumAutoInvokeAttempts > 0;
```
RogerBarreto, line 150: Avoid using base and you can use `this`, unless you intend to hide `MaximumAutoInvokeAttempts` at some point.

```suggestion
            bool autoInvoke = this.MaximumAutoInvokeAttempts > 0;
```

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
```diff
@@ -160,17 +167,11 @@ internal override void ConfigureRequest(Kernel? kernel, ChatCompletionRequest re
     /// <summary>
     /// Represents a <see cref="MistralAIToolCallBehavior"/> that provides a specified list of functions to the model.
     /// </summary>
-    internal sealed class EnabledFunctions : MistralAIToolCallBehavior
+    internal sealed class AnyFunction(IEnumerable<KernelFunction> functions, bool autoInvoke) : MistralAIToolCallBehavior(autoInvoke)
```
RogerBarreto, line 170: Maybe consider adding Choice to those to avoid confusion on the `Auto` I was almost thinking this was auto until I saw the `bool autoInvoke`.

```suggestion
    internal sealed class AnyFunctionChoice(IEnumerable<KernelFunction> functions, bool autoInvoke) : MistralAIToolCallBehavior(autoInvoke)
```

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
```diff
@@ -70,10 +70,10 @@ public async Task ValidateInvokeChatPromptAsync()
             .AddMistralChatCompletion(model!, apiKey!)
             .Build();
 
-        const string ChatPrompt = @"
+        const string ChatPrompt = """
             <message role=""system"">Respond in French.</message>
             <message role=""user"">What is the best French cheese?</message>
-        ";
+        """;
```
RogerBarreto, line 76: Using the new C# 12 literal, you don't need to escape `"` quotes anymore
```suggestion
        const string ChatPrompt = """
            <message role="system">Respond in French.</message>
            <message role="user">What is the best French cheese?</message>
        """;
```

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
```diff
@@ -130,12 +128,40 @@ public async Task ValidateGetChatMessageContentsHasToolCallsResponseAsync()
...
+        var executionSettings = new MistralAIPromptExecutionSettings { ToolCallBehavior = MistralAIToolCallBehavior.EnableKernelFunctions };
+        var response = await service.GetChatMessageContentsAsync(chatHistory, executionSettings, kernel);
+
+        // Assert
+        Assert.NotNull(response);
+        Assert.Single(response);
+        Assert.Equal("tool_calls", response[0].Metadata?["FinishReason"]);
+    }
+
+    [Fact] // (Skip = "This test is for manual verification.")
```
RogerBarreto, line 140: Skip back this and others?

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
```diff
@@ -200,21 +277,22 @@ public sealed class WeatherPlugin
         [Description("Get the current weather in a given location.")]
         public string GetWeather(
             [Description("The city and department, e.g. Marseille, 13")] string location
-            ) => $"Weather in {location} is sunny and 18 celsius";
+            ) => $"Weather in {location} is sunny and 18 Celsius";
+    }
+
+    public sealed class AnonymousPlugin
```
RogerBarreto, line 283:
```suggestion
    private sealed class AnonymousPlugin
```

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
markwallace-microsoft: `RequiredFunction` is the name we are using for the behaviour when the model must call a function. MistralAU uses the tool choice "any" for this hence the name `AnyFunction`. But `AnyFunction` is internal and no exposed as part of the public API. 
That said this code will get replaced with Sergeys function calling abstraction code once that is merged.

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
markwallace-microsoft: This is a reminder that these tests need to be skipped when merging to main

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
markwallace-microsoft: The names collide if they are the same. 

I used "None" in the internal class because that matches the tool choice. For the property I used "No" because that's a better name for the public API, the name `NoneKernelFunctions` isn't good English.

I'll be replacing these with the common implementation Sergey is working on so I'll leave as is for now.

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
markwallace-microsoft: Yes, the model responds with a message describing which function it would call e.g.

E.g. `"Sure, let me check the weather for you.\n\n[{\"name\": \"get_weather\", \"arguments\": {\"location\": \"Paris, 75\"}}}]"`

If a developer doesn't want any tool call behaviour, then they just don't set the property and then no functions get advertised to the model.

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
markwallace-microsoft: This change causes this error

```
/__w/semantic-kernel/semantic-kernel/dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs(283,26): error CA1812: 'MistralAIChatCompletionTests.AnonymousPlugin' is an internal class that is apparently never instantiated. If so, remove the code from the assembly. If this class is intended to contain only static members, make it 'static' (Module in Visual Basic). (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1812) [/__w/semantic-kernel/semantic-kernel/dotnet/src/IntegrationTests/IntegrationTests.csproj]
```

Given this is a test class I think it's OK to leave as is.

dotnet/src/IntegrationTests/Connectors/MistralAI/ChatCompletion/MistralAIChatCompletionTests.cs
markwallace-microsoft: Leaving as is while this is in a feature branch as I'm running these tests a lot. I will resolve this when I create the PR to merge to main.

dotnet/src/Connectors/Connectors.MistralAI/MistralAIToolCallBehavior.cs
markwallace-microsoft: See the comment above on naming and plans to update this code

PR #6241, opened 2024-05-14
.Net: Fix filters cloning when registered via Kernel properties

dmytrostruk:
### Motivation and Context

Based on: https://github.com/microsoft/semantic-kernel/discussions/6240

Since filters are cloned when they are registered through DI container, in the same way they should be cloned when registered through Kernel properties (i.e. `kernel.FunctionInvocationFilters`). 

### Contribution Checklist

### DIFF
--- a/dotnet/src/SemanticKernel.Abstractions/Kernel.cs
+++ b/dotnet/src/SemanticKernel.Abstractions/Kernel.cs
@@ -114,6 +114,9 @@ public sealed class Kernel
             FunctionInvoked = this.FunctionInvoked,
             PromptRendering = this.PromptRendering,
             PromptRendered = this.PromptRendered,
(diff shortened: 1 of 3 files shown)


dotnet/src/SemanticKernel.Abstractions/Kernel.cs
```diff
@@ -114,6 +114,9 @@ public sealed class Kernel
             FunctionInvoked = this.FunctionInvoked,
             PromptRendering = this.PromptRendering,
             PromptRendered = this.PromptRendered,
```
stephentoub, line 116: Unrelated to this PR, but can we get rid of the events now?

stephentoub (approved):

dotnet/src/SemanticKernel.Abstractions/Kernel.cs
dmytrostruk: Good question :) I actually did it during filters v2 implementation, but then we decided to temporarily revert them back. I think the decision was to remove them when we graduate filters to be non-experimental.

markwallace-microsoft (approved):

PR #6242, opened 2024-05-14
.Net: [WIP] OTel model diagnostics: streaming APIs

TaoChenOSU:
### Motivation and Context

Previously (https://github.com/microsoft/semantic-kernel/pull/6150) we added support for OTel (LLM semantic conventions) to non-streaming APIs in the AI connectors. This PR adds that to streaming APIs.

### Description

1. Add OTel (LLM semantic conventions) to streaming APIs.
2. Update the telemetry sample to use streaming APIs along with non-streaming ones..

### Contribution Checklist


dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -226,15 +226,33 @@ internal sealed class GeminiChatCompletionClient : ClientBase
...
+                    using var httpRequestMessage = await this.CreateHttpRequestAsync(state.GeminiRequest, this._chatStreamingEndpoint).ConfigureAwait(false);
+                    // We cannot dispose these two objects leaving the try-catch block because we need them to read the response stream
+                    httpResponseMessage = await this.SendRequestAndGetResponseImmediatelyAfterHeadersReadAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
+                    responseStream = await httpResponseMessage.Content.ReadAsStreamAndTranslateExceptionAsync().ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    activity?.SetError(ex);
+                    throw;
+                }
```
stephentoub, line 245: What about exceptions that might emerge from the foreach loop below?

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -226,15 +226,33 @@ internal sealed class GeminiChatCompletionClient : ClientBase
...
+
+                await foreach (var messageContent in this.GetStreamingChatMessageContentsOrPopulateStateForToolCallingAsync(state, responseStream, cancellationToken).ConfigureAwait(false))
+                {
+                    activity?.AddStreamingContent(messageContent);
+                    yield return messageContent;
+                }
+
+                activity?.EndStreaming();
+                httpResponseMessage.Dispose();
+                responseStream.Dispose();
```
stephentoub, line 255: These are no longer going to be called in the case of an exception above.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -169,18 +169,31 @@ internal HttpRequestMessage CreatePost(object requestData, Uri endpoint, string?
...
 
         await foreach (var streamingTextContent in this.ProcessTextResponseStreamAsync(responseStream, modelId, cancellationToken).ConfigureAwait(false))
         {
+            activity?.AddStreamingContent(streamingTextContent);
             yield return streamingTextContent;
         }
+
+        activity?.EndStreaming();
+        httpResponseMessage.Dispose();
+        responseStream.Dispose();
```
stephentoub, line 196: Same comments here (and presumably for the other connectors)

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -226,15 +226,33 @@ internal sealed class GeminiChatCompletionClient : ClientBase
...
+            using (var activity = ModelDiagnostics.StartCompletionActivity(
+                this._chatGenerationEndpoint, this._modelId, ModelProvider, chatHistory, executionSettings))
             {
-                yield return messageContent;
+                HttpResponseMessage httpResponseMessage;
+                Stream responseStream;
+                try
+                {
+                    using var httpRequestMessage = await this.CreateHttpRequestAsync(state.GeminiRequest, this._chatStreamingEndpoint).ConfigureAwait(false);
+                    // We cannot dispose these two objects leaving the try-catch block because we need them to read the response stream
```
stephentoub, line 237: Why can't the disposal be done in a finally block down at the end where things are currently being disposed?

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -35,6 +36,26 @@ internal static class ModelDiagnostics
...
+        s_activityListener.ActivityStopped = activity =>
+        {
+            // Called when an activity is stopped. Clean up the streaming content in case `EndStreaming` is not called.
+            // This action needs to be idempotent as the event may be fired multiple times.
+            if (activity.Id is not null)
+            {
+                s_streamingContents.Remove(activity.Id);
+            }
+        };
+        ActivitySource.AddActivityListener(s_activityListener);
```
stephentoub, line 56: Using a listener here seems like unnecessary overhead and complication. Can't the client just have a StringBuilder that it appends to and then passes that off at the end?

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
TaoChenOSU: We cannot have a return yield in a try-catch block thus we need to keep the objects valid going into the foreach loop.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
stephentoub: Note as well that Dictionary is not thread-safe. Concurrency here could end up corrupting s_streamingContents.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
TaoChenOSU: I will dispose them in the catch block.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
TaoChenOSU: I will dispose them in the catch block.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
stephentoub: > We cannot have a return yield in a try-catch 

But you can in a try/finally.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
TaoChenOSU: I think we can only rely on exception handling further up the stack in this case, which is outside the scope of these activities.

stephentoub: I don't understand why. Can you elaborate?

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: I think it could be beneficial to us if we do handle merging the chunks and handling the formatting in the diagnostics class as opposed to in each client. For example, we don't have to merge the chunks if model diagnostics is not enabled. And we perform the checks in a single place. 

In addition, we will need other metadata as well, not just the texts.

stephentoub: You don't have to do it in each client. You just need to have the state stored in the client and pass it in to a method that does the work.

stephentoub: (Regardless, what's currently there can't be merged; it's not safe.)

TaoChenOSU: > Note as well that Dictionary is not thread-safe. Concurrency here could end up corrupting s_streamingContents.

Since the keys of the dictionary are the activity IDs, which are unique across all processes, would concurrency be a concern here?

stephentoub: Dictionary is not thread safe. You can't safely read/write to it from multiple threads concurrently, regardless of which keys they're using.

stephentoub: (It's also a possible leak. If, for example, an async enumerator is just dropped rather than being pushed to a final state, I expect this is leaking data into a Dictionary and it'll never be cleaned up.)

TaoChenOSU: Would `ConcurrentDictionary` make it thread safe?

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -136,9 +194,43 @@ private static string ToOpenAIFormat(IEnumerable<ChatMessageContent> chatHistory
...
+        }
+        sb.Append(']');
+
+        return sb.ToString();
+    }
+
+    /// <summary>
+    /// Convert tool calls to a string aligned with the OpenAI format
+    /// </summary>
+    private static string ToOpenAIFormat(ChatMessageContentItemCollection chatMessageContentItems)
```
TaoChenOSU, line 213: Although Gemini currently supports function calling, it does not attach the tool call information in the item list.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
stephentoub: It'd avoid corrupting the dictionary for concurrent mutations to the dictionary itself. I suspect you could still end up in a situation where multiple threads are trying to modify the list for the same activity, e.g. if some code called an async method but didn't await it and then proceeded to do something else, both forks of execution will share the same async locals and activity.

This is still a possible leak, though.

And is still unnecessary overhead.

It's clever, but I don't think it's the right approach.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
stephentoub: You mean it doesn't add the function calling information to the chat history? If so, that sounds like a bug.
cc: @Krzysztof318

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
TaoChenOSU: Since we cannot yield return in a try-catch block, we can't catch exceptions raised within the foreach loop and associate them to the activities.

stephentoub: You can, you just can't use the nice foreach syntax. Instead of something like:
```C#
await foreach (var x in Foo())
    yield return x;
```
you'd write:
```C#
IAsyncEnumerator<T> e = Foo().GetEnumerator();
try
{
while (true)
{
    try
    {
        if (!await e.MoveNextAsync()) break;
    }
    catch (Exception exc) { ... }

    yield return e.Current;
}
}
finally
{
    await e.DisposeAsync();
}
```
We can decide it's not worth it, but it is possible.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: Yeah, it doesn't use the `FunctionCallContent` to store function call information in the `ChatMessageContent`. It's not a bug. You can cast the response to a more concrete type (`GeminiChatMessageContent`) to get the function calling stuff. 

I talked to Roger, he said a contributor was planning to setup `FunctionCallContent` in Gemini but the contributor is on vacation. I can add it. Just want to get the OpenAI part done before working on that.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
TaoChenOSU: Thanks! I think this is ok. Should be straightforward. I will make the changes.

dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
```diff
@@ -99,16 +114,17 @@ public static async Task Main()
     /// </summary>
     private static readonly ActivitySource s_activitySource = new("Telemetry.Example");
 
-    private const string AzureOpenAIChatServiceKey = "AzureOpenAIChat";
-    private const string GoogleAIGeminiChatServiceKey = "GoogleAIGeminiChat";
-    private const string HuggingFaceChatServiceKey = "HuggingFaceChat";
+    private const string AzureOpenAIServiceKey = "AzureOpenAI";
+    private const string GoogleAIGeminiServiceKey = "GoogleAIGemini";
+    private const string HuggingFaceServiceKey = "HuggingFace";
```
stephentoub, line 119: @markwallace-microsoft has a PR out to merge the mistral connector into main. Presumably once that's in we'll want to update it as well with all of this tracing.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -226,15 +226,57 @@ internal sealed class GeminiChatCompletionClient : ClientBase
...
+            using (var activity = ModelDiagnostics.StartCompletionActivity(
+                this._chatGenerationEndpoint, this._modelId, ModelProvider, chatHistory, executionSettings))
             {
-                yield return messageContent;
+                HttpResponseMessage? httpResponseMessage = null;
+                Stream? responseStream = null;
+                try
+                {
+                    using var httpRequestMessage = await this.CreateHttpRequestAsync(state.GeminiRequest, this._chatStreamingEndpoint).ConfigureAwait(false);
+                    // We cannot dispose these two objects leaving the try-catch block because we need them to read the response stream
```
stephentoub, line 237: Delete this comment now?

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -226,15 +226,57 @@ internal sealed class GeminiChatCompletionClient : ClientBase
...
+                            if (!await responseEnumerator.MoveNextAsync())
+                            {
+                                break;
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            activity?.SetError(ex);
+                            throw;
+                        }
```
stephentoub, line 268: @jcouv, just FYI, an example of needing to work around the inability to yield within a try that has a catch.

dotnet/src/Connectors/Connectors.Google/Core/Gemini/Clients/GeminiChatCompletionClient.cs
```diff
@@ -226,15 +226,57 @@ internal sealed class GeminiChatCompletionClient : ClientBase
...
+
+                        streamedContents.Add(responseEnumerator.Current);
+                        yield return responseEnumerator.Current;
+                    }
+                }
+                finally
+                {
```
stephentoub, line 275: responseEnumerator should also be disposed of in the finally

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -169,17 +169,54 @@ internal HttpRequestMessage CreatePost(object requestData, Uri endpoint, string?
...
+        catch (Exception ex)
+        {
+            activity?.SetError(ex);
+            httpResponseMessage?.Dispose();
+            responseStream?.Dispose();
+            throw;
+        }
 
-        await foreach (var streamingTextContent in this.ProcessTextResponseStreamAsync(responseStream, modelId, cancellationToken).ConfigureAwait(false))
+        var responseEnumerator = this.ProcessTextResponseStreamAsync(responseStream, modelId, cancellationToken)
```
stephentoub, line 190: Same comments as above

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceMessageApiClient.cs
```diff
@@ -85,17 +85,54 @@ internal sealed class HuggingFaceMessageApiClient
...
-
-        using var responseStream = await response.Content.ReadAsStreamAndTranslateExceptionAsync()
-            .ConfigureAwait(false);
+        using var activity = ModelDiagnostics.StartCompletionActivity(endpoint, modelId, this._clientCore.ModelProvider, chatHistory, executionSettings);
+        HttpResponseMessage? httpResponseMessage = null;
+        Stream? responseStream = null;
+        try
+        {
+            using var httpRequestMessage = this._clientCore.CreatePost(request, endpoint, this._clientCore.ApiKey);
+            // We cannot dispose these two objects leaving the try-catch block because we need them to read the response stream
```
stephentoub, line 94: Same comments. I won't comment on it further, just replicate whatever fixes you make across all the connectors :)

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -627,25 +661,66 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
...
+                catch (Exception ex)
+                {
+                    activity?.SetError(ex);
+                    throw;
+                }
 
-                // If we're intending to invoke function calls, we need to consume that function call information.
-                if (autoInvoke)
+                var responseEnumerator = response.ConfigureAwait(false).GetAsyncEnumerator();
+                List<OpenAIStreamingChatMessageContent> streamedContents = [];
```
stephentoub, line 680: It'd be nice if we only paid to store all these items if diagnostics are enabled, e.g. change this to be `activity is not null ? [] : null` and then the Adds to be more like `streamedContents?.Add(openAIStreamingChatMessageContent);`.

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -136,9 +148,43 @@ private static string ToOpenAIFormat(IEnumerable<ChatMessageContent> chatHistory
...
+    {
+        var sb = new StringBuilder();
```
stephentoub, line 169: This needn't create it sown StringBuilder. Just pass the callers in.

alliscode (approved):

PR #6275, opened 2024-05-15
.Net: fixed extension data in Model diagnostics

TaoChenOSU:
### Motivation and Context

Previously when an AI client starts a model diagnostics activity, it passes in an execution setting that is not parsed to a setting that is specific to the client. This creates an issue where the some of the settings cannot be read by the diagnostics module.

### Description

Pass in the parsed setting to the diagnostics module. The diagnostics module will the serialize the object and deserialize it to `PromptExecutionSettings` to get the extension data.

image

### Contribution Checklist

### DIFF
--- a/dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
+++ b/dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
@@ -48,7 +48,6 @@ public sealed class Program
             .AddSource("Microsoft.SemanticKernel*")
             .AddSource("Telemetry.Example")
             .AddAzureMonitorTraceExporter(options => options.ConnectionString = connectionString)
-            .AddOtlpExporter(options => options.Endpoint = new Uri("http://localhost:4317"))
             .Build();
 
         using var meterProvider = Sdk.CreateMeterProviderBuilder()
--- a/dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
+++ b/dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
@@ -297,6 +297,9 @@ public sealed class Program
                         GoogleAIGeminiServiceKey => new GeminiPromptExecutionSettings()
                         {
                             Temperature = 0,
+                            // Not show casing the AutoInvokeKernelFunctions behavior for Gemini due the following issue:
+                            // https://github.com/microsoft/semantic-kernel/issues/6282
+                            // ToolCallBehavior = GeminiToolCallBehavior.AutoInvokeKernelFunctions
                         },
                         HuggingFaceServiceKey => new HuggingFacePromptExecutionSettings()
                         {
--- a/dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
+++ b/dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
@@ -48,6 +48,7 @@ public sealed class Program
             .AddSource("Microsoft.SemanticKernel*")
             .AddSource("Telemetry.Example")
             .AddAzureMonitorTraceExporter(options => options.ConnectionString = connectionString)
+            .AddOtlpExporter(options => options.Endpoint = new Uri("http://localhost:4317"))
             .Build();
 
         using var meterProvider = Sdk.CreateMeterProviderBuilder()
@@ -294,8 +295,14 @@ public sealed class Program
                             Temperature = 0,
                             ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions
                         },
-                        GoogleAIGeminiServiceKey => new GeminiPromptExecutionSettings(),
-                        HuggingFaceServiceKey => new HuggingFacePromptExecutionSettings(),
+                        GoogleAIGeminiServiceKey => new GeminiPromptExecutionSettings()
+                        {
+                            Temperature = 0,
+                        },
+                        HuggingFaceServiceKey => new HuggingFacePromptExecutionSettings()
+                        {
+                            Temperature = 0,
+                        },
                         _ => null,
                     };
 
--- a/dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
+++ b/dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
@@ -134,7 +134,8 @@ internal sealed class HuggingFaceClient
         var endpoint = this.GetTextGenerationEndpoint(modelId);
         var request = this.CreateTextRequest(prompt, executionSettings);
 
-        using var activity = ModelDiagnostics.StartCompletionActivity(endpoint, modelId, this.ModelProvider, prompt, executionSettings);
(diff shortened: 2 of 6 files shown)


dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
```diff
@@ -294,8 +294,14 @@ private sealed class AIServiceSelector : IAIServiceSelector
                             Temperature = 0,
                             ToolCallBehavior = ToolCallBehavior.AutoInvokeKernelFunctions
                         },
-                        GoogleAIGeminiServiceKey => new GeminiPromptExecutionSettings(),
-                        HuggingFaceServiceKey => new HuggingFacePromptExecutionSettings(),
+                        GoogleAIGeminiServiceKey => new GeminiPromptExecutionSettings()
```
alliscode, line 297: Is this change needed?

TaoChenOSU: If this is not set, it won't have a default and will be null. This makes sure the activity of the Gemini run has at least one execution setting in this sample app.

alliscode (approved):

dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
```diff
@@ -294,8 +294,14 @@ private sealed class AIServiceSelector : IAIServiceSelector
...
+                        {
+                            Temperature = 0,
+                        },
```
stephentoub, line 300: What's the reason for not including tool calling here?

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
```diff
@@ -134,7 +134,8 @@ internal HttpRequestMessage CreatePost(object requestData, Uri endpoint, string?
         var endpoint = this.GetTextGenerationEndpoint(modelId);
         var request = this.CreateTextRequest(prompt, executionSettings);
 
-        using var activity = ModelDiagnostics.StartCompletionActivity(endpoint, modelId, this.ModelProvider, prompt, executionSettings);
+        var huggingFaceExecutionSettings = HuggingFacePromptExecutionSettings.FromExecutionSettings(executionSettings);
```
stephentoub, line 137: Should the LogTextGenerationUsage later in the method use this one instead of the original?

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceMessageApiClient.cs
```diff
@@ -85,7 +85,8 @@ internal sealed class HuggingFaceMessageApiClient
         var request = this.CreateChatRequest(chatHistory, executionSettings);
         request.Stream = true;
 
-        using var activity = ModelDiagnostics.StartCompletionActivity(endpoint, modelId, this._clientCore.ModelProvider, chatHistory, executionSettings);
+        var huggingFaceExecutionSettings = HuggingFacePromptExecutionSettings.FromExecutionSettings(executionSettings);
```
stephentoub, line 88: The original execution settings don't actually seem to be used anywhere in the implementation. Is that a bug? Or if it's by design, does it make sense to log information that's not actually used?

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -109,16 +119,23 @@ public static bool IsModelDiagnosticsEnabled()
...
+    private static void AddOptionalTags<T>(Activity? activity, T? executionSettings) where T : PromptExecutionSettings
     {
-        if (activity is null || executionSettings?.ExtensionData is null)
+        if (activity is null || executionSettings is null)
+        {
+            return;
+        }
+
+        // Serialize and deserialize the execution settings to get the extension data
+        var deserializedSettings = JsonSerializer.Deserialize<PromptExecutionSettings>(JsonSerializer.Serialize(executionSettings));
```
stephentoub, line 130: Is this always going to do the right thing? i.e. the exact same names are used for all execution settings for MaxToken, Temperature, and TopP?

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
```diff
@@ -194,13 +211,13 @@ private static void ToOpenAIFormat(StringBuilder sb, ChatMessageContentItemColle
     /// Start a completion activity and return the activity.
     /// The `formatPrompt` delegate won't be invoked if events are disabled.
     /// </summary>
-    private static Activity? StartCompletionActivity<T>(
+    private static Activity? StartCompletionActivity<T1, T2>(
```
stephentoub, line 214: How about TPrompt and TPromptExecutionSettings?

dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
TaoChenOSU: I found an issue with Gemini when invoking a Kernel function with auto invoke enabled. I will file an issue in a moment.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceMessageApiClient.cs
TaoChenOSU: It's a bit strange in the HuggingFace connector. The original execution settings is being parsed inside the `CreateChatRequest` method and all the parameters are stored in the `request`.

I can actually parse the settings to `HuggingFacePromptExecutionSettings` outside of `CreateChatRequest` and pass in the result, to avoid repeating the operation.

dotnet/samples/Demos/TelemetryWithAppInsights/Program.cs
stephentoub: Ok. I suggest we leave the line in but commented out and with the issue link next to it... it can then be uncommented when the issue is fixed (hopefully soon :smile:)

dotnet/src/InternalUtilities/src/Diagnostics/ModelDiagnostics.cs
TaoChenOSU: Yes, so far. We just need to be very careful about this. I am planning to add unit tests for all the model diagnostics code.

dotnet/src/Connectors/Connectors.HuggingFace/Core/HuggingFaceClient.cs
TaoChenOSU: Good catch! Yes, in theory. But it won't change anything ATM. I will change it for consistency.

stephentoub (approved):
one unaddressed comment, otherwise lgtm

TaoChenOSU:
> one unaddressed comment, otherwise lgtm

Added the line back with a comment and the link to the issue.

markwallace-microsoft (approved):

Issue #6081, opened 2024-05-01
.Net: Consider adding Config classes for use with IKernelBuilder and IServiceCollection AddXXX methods


markwallace-microsoft:
One problem with the IKernelBuilder and IServiceCollection AddXXX extension methods is that they all have a very long list of parameters. So adding new parameters is a breaking change and we are inclined to keep the existing parameter order (to reduce impact of the breaking change) even if it doesn't make sense.

The goal of this task is to consider adding XXXConfig classes and add new extension methods which just take an instance of XXXConfig. All of the old extension methods can be refactored to call this.

markwallace-microsoft:
No plans to do this at present

PR #6283, opened 2024-05-16
.Net: Adding Config classes for use with IKernelBuilder and IServiceCollection

akordowski:
### Motivation and Context

This PR introduce Config classes as described in the issue #6081.

### Description

For this PR I was inspiered by playing with the [microsoft/chat-copilot](https://github.com/microsoft/chat-copilot) project and looking into the configuration in the [microsoft/kernel-memory](https://github.com/microsoft/kernel-memory) project. The new classes and logic are copied or inspired by the code in the `kernel-memory` project. Some of the properties don't match with the C# convention, but I kept them as they are to have consistent API  (e.g. `APIType`, `APIType` -> by C# convention should be rather `ApiType`, `ApiType`).

With this proposed addition we will have a similar API and developer experience through the  [microsoft/semantic-kernel](https://github.com/microsoft/semantic-kernel/) and the [microsoft/kernel-memory](https://github.com/microsoft/kernel-memory) projects.

The following classes are inspired by:

- `KernelConfig` - [`KernelMemoryConfig`](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Configuration/KernelMemoryConfig.cs)
- `AzureOpenAIConfig` - [`AzureOpenAIConfig`](https://github.com/microsoft/kernel-memory/blob/main/extensions/AzureOpenAI/AzureOpenAIConfig.cs)
- `OpenAIConfig` - [`OpenAIConfig`](https://github.com/microsoft/kernel-memory/blob/main/extensions/OpenAI/OpenAIConfig.cs)
- `ConfigurationException` - [`ConfigurationException`](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Configuration/ConfigurationException.cs)

Thank you for considering of this PR. Looking forward to your feedback.

### Contribution Checklist


akordowski:
**Note:** I also have seen that in the [`OpenAIServiceCollectionExtensions`](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs) class some parameters are not consistent, e.g. `TokenCredential credentials` and `TokenCredential credential`. I haven't changed it yet but I think it would be good to have consistent naming.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
```diff
@@ -213,14 +214,125 @@ public static class OpenAIServiceCollectionExtensions
...
+        AzureOpenAIConfig config,
+        string? serviceId = null,
+        HttpClient? httpClient = null)
+    {
+        Verify.NotNull(builder);
+        Verify.NotNull(config);
+
+        config.Validate();
+
+        switch (config.Auth)
```
dmytrostruk, line 236: Let's avoid using enums for this configuration. I think we can create hierarchy of configuration classes, where base class will contain common properties like `deploymentName` etc, and different derived types will contain configuration-specific properties like `TokenCredential` or `string ApiKey`. I think the main goal would be to create the same set of overloads we have today, but with specific config models. This approach will also decrease the probability of incorrect configuration when `AuthType` is unknown. With specific classes we won't need to throw an exception.

Otherwise, if enum value is `TokenCredential`, we will have to verify that the property `TokenCredential` was initialized and so on. The more parameters will be added in the future, the harder it will be to maintain the validation. But having specific config models should simplify the usage and the maintenance.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
```diff
@@ -213,14 +214,125 @@ public static class OpenAIServiceCollectionExtensions
...
+    /// <param name="builder">The <see cref="IKernelBuilder"/> instance to augment.</param>
+    /// <param name="config">Required configuration for Azure OpenAI.</param>
+    /// <param name="serviceId">A local identifier for the given AI service.</param>
+    /// <param name="httpClient">The HttpClient to use with this service.</param>
+    /// <returns>The same instance as <paramref name="builder"/>.</returns>
+    public static IKernelBuilder AddAzureOpenAITextGeneration(
+        this IKernelBuilder builder,
+        AzureOpenAIConfig config,
+        string? serviceId = null,
+        HttpClient? httpClient = null)
```
dmytrostruk, line 229: I think we also want to put `serviceId` and `httpClient` to config model as nullable. If we introduce config models, I think no additional parameters should be present in the method. It will also simplify the usage, because if user already initialize config object, they won't have to pass `serviceId` and `httpClient` parameters separately.

dotnet/src/Connectors/Connectors.OpenAI/Configuration/AzureOpenAIConfig.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// <summary>
+    /// OpenAI API type, e.g. text completion, chat completion, image generation, etc.
+    /// </summary>
+    public APITypes APIType { get; set; } = APITypes.ChatCompletion;
+
+    /// <summary>
+    /// Azure authentication type.
+    /// </summary>
+    public AuthTypes Auth { get; set; }
```
dmytrostruk, line 37: As already mentioned, it would be nice to avoid using these enums. For `AuthTypes` - define specific classes. For `APITypes` - I'm not sure we need it at all.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
```diff
@@ -213,14 +214,125 @@ public static class OpenAIServiceCollectionExtensions
...
 
+    /// <summary>
+    /// Adds an Azure OpenAI text generation service with the specified configuration.
+    /// </summary>
+    /// <param name="builder">The <see cref="IKernelBuilder"/> instance to augment.</param>
+    /// <param name="config">Required configuration for Azure OpenAI.</param>
+    /// <param name="serviceId">A local identifier for the given AI service.</param>
+    /// <param name="httpClient">The HttpClient to use with this service.</param>
+    /// <returns>The same instance as <paramref name="builder"/>.</returns>
+    public static IKernelBuilder AddAzureOpenAITextGeneration(
```
dmytrostruk, line 225: I noticed that in some places you marked new methods as experimental, thanks for that! In scope of this work, we need to make every new class and method as experimental for initial period of time.

dotnet/src/SemanticKernel.Core/Configuration/KernelConfig.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.SemanticKernel;
+
+public class KernelConfig
```
dmytrostruk, line 8: Not sure what is the purpose of this class, but our goal would be to refactor only extension methods for `IKernelBuilder` and `IServiceCollection` `Add...` methods.

dmytrostruk (commented):
Thanks for your contribution! I left initial comments. I also think that we need some alignment between `Config` and `Options` naming, because today we have some classes that are `Options`:
[KernelFunctionFromMethodOptions.cs](https://github.com/microsoft/semantic-kernel/blob/a5a38b8fc132afc31d2be85373752c1abb95ba99/dotnet/src/SemanticKernel.Core/Functions/KernelFunctionFromMethodOptions.cs)

Taking into account that `Options` is used as .NET pattern for strongly-typed settings, I would prefer to use it for extension methods as well:
- https://learn.microsoft.com/en-us/dotnet/core/extensions/options
- https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.ioptions-1

cc: @markwallace-microsoft @matthewbolanos

akordowski:
Thank you for your feedback!

The proposed implementation reflects the implementation of the configuration in the `microsoft/kernel-memory` project. As I understand it the `semantic-kernel` and the `kernel-memory` projects are in the same domain. From developers perspective I would appreciate to have a consistent way to configure things (not _‘here we do it this way and here the other way’_).

I don’t know if the `semantic-kernel` and the `kernel-memory` projects are developed by the same team or different teams, but maybe the team(s) could agree on a single design which could/should be used across the projects. This way we would have a consistent API and it would result in a better developer experience.

I hope I was able to explain my thoughts. What do you think?

> Taking into account that Options is used as .NET pattern for strongly-typed settings, I would prefer to use it for extension methods as well:

I’m aware of that pattern (and I agree), but the `OpenAIServiceCollectionExtensions` already uses in the ` AddAzureOpenAIChatCompletion ` method a `AzureOpenAIChatCompletionWithDataConfig` config parameter. Introducing `Option` classes may lead to confusion and were not consistent with the current implementation.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
akordowski: Please see my [comment](https://github.com/microsoft/semantic-kernel/pull/6283#issuecomment-2117053217) below.

dotnet/src/Connectors/Connectors.OpenAI/Configuration/AzureOpenAIConfig.cs
akordowski: Please see my [comment](https://github.com/microsoft/semantic-kernel/pull/6283#issuecomment-2117053217) below.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
akordowski: I was not sure about that but I think it is a good idea. I will move it adjust it.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
akordowski: Thanks for the hint. I addeed the missing attributes.

dmytrostruk:
> From developers perspective I would appreciate to have a consistent way to configure things (not _‘here we do it this way and here the other way’_).

I 100% agree, but we can always improve things in both projects if needed :)
I mentioned my concerns about using enums, and I believe we can make it simpler, comparing to using a lot of `switch/case` statements.

> I’m aware of that pattern (and I agree), but the OpenAIServiceCollectionExtensions already uses in the AddAzureOpenAIChatCompletion method a AzureOpenAIChatCompletionWithDataConfig config parameter.

This class is going to be deprecated at some point, since current version of `chat completion with data` functionality was added at the point when this feature wasn't available in Azure .NET SDK. Since it's available now, we are going to use it directly, which means that `WithData` classes will be deprecated.

dotnet/src/SemanticKernel.Core/Configuration/KernelConfig.cs
dmytrostruk: So far, we didn't see a requirement for such class. And without usage examples it's hard to understand the benefits. My recommendation would be to stick to the original issue that mentioned only updates in `Add...` extension methods. It will simplify the review and merging of this PR. 

If you feel that such class will provide some benefits, you can create new issue with all the details, including usage examples, and we can handle it separately. Hope that makes sense!

akordowski: I have deleted the class.

dotnet/src/Connectors/Connectors.OpenAI/Configuration/AzureOpenAIConfig.cs
akordowski: As requested I created specific classes for the configuration and removed the enums. Please take a look.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
akordowski: I moved the `ServiceId` property to the config class and adjusted the methods.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
akordowski: Adjusted as requested. Please take a look.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIServiceCollectionExtensions.cs
```diff
@@ -213,14 +213,124 @@ public static class OpenAIServiceCollectionExtensions
...
+        Verify.NotNull(config);
+
+        config.Validate();
+
+        builder.AddAzureOpenAITextGeneration(
+            deploymentName: config.Deployment,
+            endpoint: config.Endpoint,
+            credentials: config.TokenCredential!,
+            serviceId: config.ServiceId,
+            modelId: config.Deployment,
```
dmytrostruk, line 268: In current extension methods `deploymentName` and `modelId` properties are separate, same as on Azure side. We should keep that logic with new configuration models as well.

akordowski:
> Please let us know if you feel comfortable to resolve all comments. Otherwise, it's totally fine to leave this PR as it is, and our team will continue the work to resolve the issue.

I would try to resolve the comments but the unit tests I would maybe left for the team :satisfied:

> 1. Based on team discussion, we would like to proceed with `Options` naming instead of `Config`.

I have some questions regarding this.

If we create option classes for each modality type (as you suggested [above](https://github.com/microsoft/semantic-kernel/pull/6283#discussion_r1613824393)) it would lead to initialization like this:

```csharp
var options = new AzureOpenAITextGenerationOptions(ApiKey / Token)
{
	DeploymentName = "...",
    ModelId = "...",
    ...
}
```

- Or should the option classes be initialized only over the constructor?
- If yes should the properties still be `get/set` or only `get`?
- As we are going with the `Options` pattern 
  - should I keep the `Validate()` method in the option class
  - or rather add property attributes to validate the object by the `ConfigurationBuilder`
  - or remove the validation and use constructor attribute validation instead?

The approach would also require a check for the `ApiKey/Token` in the extension methods in the `OpenAIServiceCollectionExtensions` class.

```csharp
public static IKernelBuilder AddAzureOpenAITextGeneration(
    this IKernelBuilder builder,
    AzureOpenAITextGenerationOptions options)
{
    if (!string.IsNullOrWhiteSpace(options.ApiKey))
    {
        builder.Services.AddMethod(options.ApiKey))
    }
    else if (options.Token is not null)
    {
        builder.Services.AddMethod(options.Token))
    }
    else
    {
        // This case may never occour as we would have to set the auth via the constructor
        throw new InvalidOperationException("Please provide API key or Token authentication.");
    }
}
```

Please provide details on how the team wish it to be implemented. Thank you!

akordowski:
@dmytrostruk Any further instructions how I should proceed so we can finalize this PR? (See my comments above)

dmytrostruk:
> Any further instructions how I should proceed so we can finalize this PR? (See my comments above)

@akordowski We have to wait with this PR a little bit, since there is parallel work going on to upgrade our OpenAI connector to use separate OpenAI .NET SDK. That part of work would also impact configuration changes, so maybe it will be done in the same scope. Since we are in the middle of investigation phase, I will keep you updated here on the progress. 

In all cases, many thanks for your contribution! I will mark this PR as draft for now.

akordowski:
> We have to wait with this PR a little bit

Thank you for the update. Feel free to reach out if I can help to finalize the PR.

> In all cases, many thanks for your contribution!

You're very welcome!

markwallace-microsoft:
@akordowski Thanks for the contribution. We are going to close this PR for now and use it when we pick up the work for #6081.

PR #6297, opened 2024-05-16
.Net: Add activities to MistralClient

stephentoub:
Replicates the ModelDiagnostics stuff to the MistralAI chat completion service implementation.

I still need to test it. Best I can say now is it compiles :)

cc: @markwallace-microsoft, @TaoChenOSU


TaoChenOSU (approved):

TaoChenOSU:
LGTM, I think we should still test it to make sure the metadata and the execution settings are there in the activities.

markwallace-microsoft (approved):

Issue #6277, opened 2024-05-15
.Net: OpenAIPromptExecutionSettings does not support enabling logprobs


abhiksingla:
Currently, OpenAIPromptExecutionSettings does not support enabling logprobs in the request. It would be great, if the functionality could be enabled and logprobs be returned in the repsonse.

PR #6300, opened 2024-05-16
.Net: Added logprobs property to OpenAIPromptExecutionSettings

dmytrostruk:
### Motivation and Context

Fixes: https://github.com/microsoft/semantic-kernel/issues/6277

https://platform.openai.com/docs/api-reference/chat/create#chat-create-logprobs

### Contribution Checklist


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1036,7 +1036,7 @@ private static CompletionsOptions CreateCompletionsOptions(string text, OpenAIPr
             Echo = false,
             ChoicesPerPrompt = executionSettings.ResultsPerPrompt,
             GenerationSampleCount = executionSettings.ResultsPerPrompt,
-            LogProbabilityCount = null,
+            LogProbabilityCount = executionSettings.LogProbabilitiesCount,
```
markwallace-microsoft, line 1039: Should we use the same name for the property?

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -254,6 +254,39 @@ public int ResultsPerPrompt
         }
     }
 
+    /// <summary>
+    /// Whether to return log probabilities of the output tokens or not.
+    /// If true, returns the log probabilities of each output token returned in the `content` of `message`.
+    /// </summary>
+    [Experimental("SKEXP0010")]
+    [JsonPropertyName("logprobs")]
+    public bool? EnableLogProbabilities
```
stephentoub, line 263: Why `bool?` rather than `bool`?

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -254,6 +254,39 @@ public int ResultsPerPrompt
...
+    public int? LogProbabilitiesCount
+    {
+        get => this._logProbabilitiesCount;
+
+        set
+        {
+            this.ThrowIfFrozen();
+            this._logProbabilitiesCount = value;
+        }
+    }
```
stephentoub, line 288: What about the consumption of the logprobs that come back in the response? Does that already work and they're just surfaced somewhere in the Metadata?

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
markwallace-microsoft, line 263: Elsewhere we match the property name with the JSON property name. Why are we changing the pattern for these two properties?

I'd expect them to be `Logprobs` and `TopLogprobs`

dmytrostruk: I used the same type as in Azure .NET SDK. Also, OpenAI documentation explicitly says that it's bool or null:
image

Today, Azure .NET SDK sends always `null`. I thought it's better to keep that behavior rather than start sending `false`, even though it may not change overall behavior.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
dmytrostruk: Yes, today they are already included in response metadata:
https://github.com/microsoft/semantic-kernel/blob/33f278c477c264aa1e84caedd9e41cd1ea926582/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs#L248

https://github.com/microsoft/semantic-kernel/blob/33f278c477c264aa1e84caedd9e41cd1ea926582/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs#L268

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
dmytrostruk: Azure SDK uses `LogProbabilityCount` for text generation and `LogProbabilitiesPerToken` for chat completion. I decided to go with `LogProbabilitiesCount` as something in the middle.

dmytrostruk: Renamed to `Logprobs` and `TopLogprobs`.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
dmytrostruk: Renamed to `Logprobs` and `TopLogprobs`.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
dmytrostruk: Added integration test that verifies consumption behavior.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
stephentoub: Thanks. Just seems odd to explicitly have three representable states when there's actually only two. It's strange on the OpenAI definition as well, especially with it defaulting to false rather than null.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
stephentoub: Cool, thanks.

SergeyMenshykh (approved):

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -254,6 +254,39 @@ public int ResultsPerPrompt
...
+
+        set
+        {
+            this.ThrowIfFrozen();
+            this._logprobs = value;
+        }
+    }
+
+    /// <summary>
+    /// An integer specifying the number of most likely tokens to return at each token position, each with an associated log probability.
```
markwallace-microsoft, line 275: Any reason not t use the description from the OpenAI documentation?

`An integer between 0 and 20 specifying the number of most likely tokens to return at each token position, each with an associated log probability. logprobs must be set to true if this parameter is used.`

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -254,6 +254,39 @@ public int ResultsPerPrompt
         }
     }
 
+    /// <summary>
+    /// Whether to return log probabilities of the output tokens or not.
+    /// If true, returns the log probabilities of each output token returned in the `content` of `message`.
+    /// </summary>
+    [Experimental("SKEXP0010")]
```
markwallace-microsoft, line 261: I don't think we need to make this experimental as this is just a wrapper around what is in the OpenAI API.

Is there any uncertainty here?

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -254,6 +254,39 @@ public int ResultsPerPrompt
...
+        {
+            this.ThrowIfFrozen();
+            this._logprobs = value;
+        }
+    }
+
+    /// <summary>
+    /// An integer specifying the number of most likely tokens to return at each token position, each with an associated log probability.
+    /// </summary>
+    [Experimental("SKEXP0010")]
```
markwallace-microsoft, line 277: I don't think we need to make this experimental as this is just a wrapper around what is in the OpenAI API.

Is there any uncertainty here?

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
dmytrostruk: Mostly because of integer range difference. OpenAI documentation says `integer between 0 and 20`, while Azure .NET SDK says `integer between 0 and 5`:
https://github.com/Azure/azure-sdk-for-net/blob/576ace699a0b61721762d5898e6d6a201d4861a0/sdk/openai/Azure.AI.OpenAI/src/Generated/ChatCompletionsOptions.cs#L195
image


And I tested both limitations using OpenAI and Azure OpenAI endpoints and they work as mentioned in documentation. Taking into account that we use `OpenAIPromptExecutionSettings` for both OpenAI and Azure OpenAI, I decided to not mention that limitation. But we can specify both limitations if needed. We can also add `logprobs must be set to true if this parameter is used.` part.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
dmytrostruk: > I don't think we need to make this experimental as this is just a wrapper around what is in the OpenAI API.

@markwallace-microsoft I followed the same approach we used when added `Seed` field in this PR:
https://github.com/microsoft/semantic-kernel/pull/4162/files#diff-fffa27d3df6f16d9f8a41044ade5f40246999ce8de832769b7ca91f8b94891d4R73
image

I also think that even if it's just a wrapper, it's better to mark it experimental, keep it for a couple of weeks and in case there are no issues - remove `Experimental` attribute. It gives us some space for bug fixing if any.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
markwallace-microsoft: Thanks, that's good to know

markwallace-microsoft (approved):

markwallace-microsoft (approved):

Issue #5818, opened 2024-04-09
.Net: Support FunctionCallContent for OpenAI streaming API


SergeyMenshykh:
Two new types - `FunctionCallRequestContent` and `FunctionCallResultContent` have been added to support connector-agnostic function calling model for non-streaming API: **Requirements**: https://github.com/microsoft/semantic-kernel/issues/5153. **PR**: https://github.com/microsoft/semantic-kernel/pull/5800

This item tracks work required to add connector-agnostic function calling model for **streaming** API.

PR #6305, opened 2024-05-17
.Net: Include streaming tool call information in model diagnostics

TaoChenOSU:
### Motivation and Context

Tool call information is currently not included in the model diagnostics when using the streaming APIs.

### Description

1. Record OpenAI tool call information in model diagnostics for the streaming API.
2. If there is no tool call information, do not record an empty entry.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
@@ -724,7 +724,7 @@ internal abstract class ClientCore
                     toolCalls = OpenAIFunctionToolCall.ConvertToolCallUpdatesToChatCompletionsFunctionToolCalls(
                         ref toolCallIdsByIndex, ref functionNamesByIndex, ref functionArgumentBuildersByIndex);
                     // Translate all entries into FunctionCallContent instances for diagnostics purposes.
-                    functionCallContents = activity is null ? null : toolCalls.Select(this.GetFunctionCallContent).ToArray();
+                    functionCallContents = ModelDiagnostics.IsSensitiveEventsEnabled() ? toolCalls.Select(this.GetFunctionCallContent).ToArray() : null;
                 }
                 finally
                 {
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
@@ -662,6 +662,8 @@ internal abstract class ClientCore
             string? streamedName = null;
             ChatRole? streamedRole = default;
             CompletionsFinishReason finishReason = default;
+            ChatCompletionsFunctionToolCall[]? toolCalls = null;
+            FunctionCallContent[]? functionCallContents = null;
 
             using (var activity = ModelDiagnostics.StartCompletionActivity(this.Endpoint, this.DeploymentOrModelName, ModelProvider, chat, chatExecutionSettings))
             {
@@ -717,10 +719,16 @@ internal abstract class ClientCore
                         streamedContents?.Add(openAIStreamingChatMessageContent);
                         yield return openAIStreamingChatMessageContent;
                     }
+
+                    // Translate all entries into ChatCompletionsFunctionToolCall instances.
+                    toolCalls = OpenAIFunctionToolCall.ConvertToolCallUpdatesToChatCompletionsFunctionToolCalls(
+                        ref toolCallIdsByIndex, ref functionNamesByIndex, ref functionArgumentBuildersByIndex);
+                    // Translate all entries into FunctionCallContent instances for diagnostics purposes.
+                    functionCallContents = activity is null ? null : toolCalls.Select(this.GetFunctionCallContent).ToArray();
                 }
                 finally
                 {
-                    activity?.EndStreaming(streamedContents);
+                    activity?.EndStreaming(streamedContents, functionCallContents);
                     await responseEnumerator.DisposeAsync();
                 }
             }
@@ -738,10 +746,6 @@ internal abstract class ClientCore
             // Get any response content that was streamed.
             string content = contentBuilder?.ToString() ?? string.Empty;
 
-            // Translate all entries into ChatCompletionsFunctionToolCall instances.
-            ChatCompletionsFunctionToolCall[] toolCalls = OpenAIFunctionToolCall.ConvertToolCallUpdatesToChatCompletionsFunctionToolCalls(
-                ref toolCallIdsByIndex, ref functionNamesByIndex, ref functionArgumentBuildersByIndex);
-
             // Log the requests
             if (this.Logger.IsEnabled(LogLevel.Trace))
             {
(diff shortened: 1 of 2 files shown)


dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -717,10 +719,16 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
                         streamedContents?.Add(openAIStreamingChatMessageContent);
                         yield return openAIStreamingChatMessageContent;
                     }
+
+                    // Translate all entries into ChatCompletionsFunctionToolCall instances.
+                    toolCalls = OpenAIFunctionToolCall.ConvertToolCallUpdatesToChatCompletionsFunctionToolCalls(
+                        ref toolCallIdsByIndex, ref functionNamesByIndex, ref functionArgumentBuildersByIndex);
+                    // Translate all entries into FunctionCallContent instances for diagnostics purposes.
+                    functionCallContents = activity is null ? null : toolCalls.Select(this.GetFunctionCallContent).ToArray();
```
stephentoub, line 727: Rather than doing all this translation, can we just get the original response from the Azure SDK object? Doesn't it make it available as a raw response content off of the relevant objects? It'd be both cheaper and higher fidelity.

stephentoub: Is checking `activity is null` the right check here, or do we need to check if the sensitive model diagnostics are enabled? It'd be good to avoid doing such work except in the case where it's actually going to be used.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -755,7 +759,17 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
             // Add the original assistant message to the chatOptions; this is required for the service
             // to understand the tool call responses.
             chatOptions.Messages.Add(GetRequestMessage(streamedRole ?? default, content, streamedName, toolCalls));
-            chat.Add(new OpenAIChatMessageContent(streamedRole ?? default, content, this.DeploymentOrModelName, toolCalls, metadata) { AuthorName = streamedName });
+            // Add the result message to the caller's chat history
+            var newChatMessageContent = new OpenAIChatMessageContent(streamedRole ?? default, content, this.DeploymentOrModelName, toolCalls, metadata)
+            {
+                AuthorName = streamedName
+            };
+            // Add the tool call messages to the new chat message content for diagnostics purposes.
```
stephentoub, line 767: Why is this necessary when toolCalls is already being provided to the content ctor?

stephentoub:
Do the other connectors that currently support function calling (Gemini, Mistral) have a similar problem?

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
TaoChenOSU: The reason I didn't do that is because I would have to do the parsing inside the model diagnostic module, which is supposed to be agnostic to the model specifics.

The `FunctionCallContent` class carries enough information to represent the tool calls and is a common class. It's used in the non-streaming API, and AFAIK the Gemini connector is going to adapt to it soon.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
TaoChenOSU: It's because the message is of a derived type of the `ChatMessageContent` type, which doesn't not have the tool calls property.

TaoChenOSU:
> Do the other connectors that currently support function calling (Gemini, Mistral) have a similar problem?

I intentionally did not record the tool calls information in the Gemini connector, since it's going to adapt to the `FunctionCallContent` and migrate to the new function call abstractions.

TaoChenOSU:
> Do the other connectors that currently support function calling (Gemini, Mistral) have a similar problem?

Mistral has the same issue with its streaming API. I will address it in a separate PR.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
TaoChenOSU: > Is checking `activity is null` the right check here, or do we need to check if the sensitive model diagnostics are enabled? It'd be good to avoid doing such work except in the case where it's actually going to be used.

Nice catch! I will address this

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
stephentoub: Isn't this going to interfere with non-diagnostic uses, though? This is the user's ChatHistory object, and this is now changing the ChatMessageContent to now have Items that it didn't previously have. That doesn't seem right. @SergeyMenshykh?

SergeyMenshykh: > Isn't this going to interfere with non-diagnostic uses, though? This is the user's ChatHistory object, and this is now changing the ChatMessageContent to now have Items that it didn't previously have. That doesn't seem right. @SergeyMenshykh?

@TaoChenOSU, do you use the FunctionCallContent items from the chat message content items for model diagnostics? Or are you adding them because similar functionality exists for non-streaming APIs?  
   
Either way, this change will be required to support function call content classes for streaming. I would have added it anyway, within the scope of the task - https://github.com/microsoft/semantic-kernel/issues/5818, to communicate the fact of the function call by LLM to callers in a connector-agnostic manner.  
   
I don't expect callers to be impacted by the change. On the contrary, they will be able to get function call info from the chat message item's Items collection in a connector-agnostic way rather than casting the chat message to the OpenAI one and accessing the info via the ToolCalls property.

TaoChenOSU: It’s required for model diagnostics.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -717,10 +719,16 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
                         streamedContents?.Add(openAIStreamingChatMessageContent);
                         yield return openAIStreamingChatMessageContent;
                     }
+
+                    // Translate all entries into ChatCompletionsFunctionToolCall instances.
+                    toolCalls = OpenAIFunctionToolCall.ConvertToolCallUpdatesToChatCompletionsFunctionToolCalls(
+                        ref toolCallIdsByIndex, ref functionNamesByIndex, ref functionArgumentBuildersByIndex);
+                    // Translate all entries into FunctionCallContent instances for diagnostics purposes.
+                    functionCallContents = ModelDiagnostics.IsSensitiveEventsEnabled() ? toolCalls.Select(this.GetFunctionCallContent).ToArray() : null;
```
markwallace-microsoft, line 727: Do we need to check if we have an `activity`

TaoChenOSU: That's implicitly checked. The `ModelDiagnostics` module makes sure that the activity is null when model diagnostics is disabled. And `IsSensitiveEventEnabled` checks if model diagnostics is enabled and the collection of sensitive event is enabled.

SergeyMenshykh, line 727: The functionCallContents variable should always be initialized by the value returned by GetFunctionCallContent, regardless of whether logging of sensitive events is enabled or not. Otherwise, callers will have to enable sensitive events to have the function call content items in the chat message.  
   
The "if" logic can be moved a few lines below, where the list is passed to the EndStreaming method. Depending on the IsSensitiveEventsEnabled, either the functionCallContents value or null will be passed as an argument.  
   
Since the function call content classes for streaming are not officially supported yet, this behavior may be acceptable for now and can be changed later when adding call content classes for streaming.

TaoChenOSU: Agreed. I think we can remove this when we have more consistency across the connectors. I may modify this behavior when I work on the Mistral connector.

markwallace-microsoft (approved):

PR #6309, opened 2024-05-17
.Net: Include request info in HttpOperationException

markwallace-microsoft:
### Contribution Checklist


westey-m (approved):

dotnet/src/SemanticKernel.Abstractions/Http/HttpOperationException.cs
```diff
@@ -58,4 +58,19 @@ public HttpOperationException(HttpStatusCode? statusCode, string? responseConten
     /// Gets or sets the content of the HTTP response.
     /// </summary>
     public string? ResponseContent { get; set; }
+
+    /// <summary>
+    /// Gets the method used for the HTTP request.
+    /// </summary>
+    public string? RequestMethod { get; set; }
```
SergeyMenshykh, line 65: Overall, it looks good to me. I think it may be beneficial for callers to have the extra information. The only concern I have is that some of the SK components might not provide this additional information. For instance, some services may use Service Client libraries/SDKs that do not expose the request details externally, so they will not be able to add them to the exception.

markwallace-microsoft: I have updated the comment to reflect this limitation

SergeyMenshykh (approved):

PR #6319, opened 2024-05-17
.Net Kernel Contents Graduation

### LINKED ISSUE #5625: Graduate Content Types
Checklist to be completed when graduating an experimental feature

- [ ] Verify the Content Types functionality is complete (Matthew to provide requirements)
- [ ] ​Verify API docs are complete and arrange to have them published
- [ ] Make appropriate updates to Learn docs​
- [ ] Make appropriate updates to Samples​​
- [ ] Verify there are no serious open Issues​​
- [ ] Verify there are sample implementations​ for the main use cases
- [ ] Update table in EXPERIMENTS.md 
- [ ] Remove SKEXP​ flag from Filters code

RogerBarreto:
### Motivation and Context

⚠️ Breaking changes on non-experimental types **ImageContent**

Resolves #5625

For a brief time this changes will keep the content below as experimental.

- BinaryContent
- AudioContent
- ImageContent 
- FunctionCallContent
- FunctionRequestContent

Changes:

### **BinaryContent** 
- Removed providers for lazy loading content, simplifying its usage and APIs.
- Removed `Stream` constructor to avoid IDisposable resource consumption or bad practices.
- Added `Uri` dedicated for Referenced Uri information
- Added `DataUri` property which can be set or get dynamically (auto generated if you created the 
content using byte array with a mime type) 
  Setting a `DataUri` will automatically update the `MimeType` property and add any extra metadata that may be available in the data scheme definition.
- Added a required `mimeType` property to the ByteArray constructor, to encourage passing the mimeType when creating BinaryContent directly or from specializations.
- Added `Data` property which can be set or get dynamically (auto generated if you created the content using a data uri format)
  Setting a Data on an existing BinaryContent will also reflect on the getter of `DataUri` for the given content.
- Added DataUri and Base64 validation when setting DataUri on the contents.

### ⚠️ **ImageContent** Fixes bugs and inconsistency behavior:
- Setting the Data of an image doesn't change the current data uri and vice versa, allowing the sema image content instance to have different binary data to representations.
- When an Image could not have DataUri and Uri in the same instance, this limits scenarios where you have the image data but want to have a reference to where that content is from.
- Wasn't possible to create an Image from a data uri greater than the size limit of .Net System.Uri type here:
[https://github.com/dotnet/runtime/issues/96544](https://github.com/dotnet/runtime/issues/96544).

### **FunctionResultContent**
- Update `Id` property to `CallId`.


dotnet/src/SemanticKernel.Abstractions/Contents/KernelContent.cs
```diff
@@ -38,21 +43,17 @@ public abstract class KernelContent
     /// <summary>
     /// The model ID used to generate the content.
     /// </summary>
+    [JsonPropertyName("modelId")]
```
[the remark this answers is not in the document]
RogerBarreto: You are right, I considered adding this breaking change, but after double checking our contents, none of them has `JsonPropertyName` definition, I will revert this change.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
```diff
@@ -3,118 +3,310 @@
...
+        // Validate the dataUri format
+        var parsedDataUri = DataUriParser.Parse(dataUri);
+
+        // Overwrite the mimetype to the DataUri.
+        this.MimeType = parsedDataUri.MimeType;
+
+        // If parameters where provided in the data uri, updates the content metadata.
+        if (parsedDataUri.Parameters.Count != 0)
+        {
+            this.UpdateDataUriParametersToMetadata(parsedDataUri);
```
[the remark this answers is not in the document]
RogerBarreto: After researching for the `Data Uri` specification and understanding this is accepted, since we have a parser already implemented adding this made sense when using the `Data Uri` on its full glance and seems handy way of not losing data or enforcing the caller to interate over those to add manually to metadata. 
The parser also has a special logic to only interate over data uri parameters if any is present.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
```diff
@@ -3,118 +3,310 @@
...
+    /// Initializes a new instance of the <see cref="BinaryContent"/> class from a byte array.
+    /// </summary>
+    /// <param name="data">Byte array content</param>
+    /// <param name="mimeType">The mime type of the content</param>
+    public BinaryContent(
+        ReadOnlyMemory<byte> data,
+        string? mimeType)
+    {
+        Verify.NotNull(data, nameof(data));
+        if (data.IsEmpty)
```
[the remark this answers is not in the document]
RogerBarreto: The previous `ImageContent` had this check, I thought it did make sense to have it the same way you are suggesting not having nullable constructors. 

Either way I can ignore it

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
```diff
@@ -3,118 +3,310 @@
...
+    public Uri? Uri
     {
+        get => this.GetUri();
+        set => this.SetUri(value);
     }
 
     /// <summary>
-    /// Initializes a new instance of the <see cref="BinaryContent"/> class.
+    /// Gets the DataUri of the content.
     /// </summary>
```
westey-m, line 45:
```suggestion
    /// </summary>
    /// <remarks>
    /// This is a uri that contains the content in a base64 encoded binary format and should
    /// use the "data" URL scheme as defined <see href="https://www.rfc-editor.org/rfc/rfc2397">here</see>.
    /// </remarks>
```

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
```diff
@@ -3,118 +3,310 @@
...
     /// The binary content.
     /// </summary>
-    public ReadOnlyMemory<byte>? Content { get; set; }
+    [JsonIgnore, Obsolete("Use Data instead")]
+    public ReadOnlyMemory<byte>? Content => this.Data;
 
     /// <summary>
-    /// Initializes a new instance of the <see cref="BinaryContent"/> class.
+    /// Gets the referenced Uri of the content.
     /// </summary>
```
westey-m, line 34:
```suggestion
    /// </summary>
    /// <remarks>
    /// This is a uri that points at the location of the content if it is available for reading from an external source.
    /// </remarks>```

File: dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs [hunk 6]
```diff
@@ -3,118 +3,310 @@
...
     /// </remarks>
+    [JsonConstructor]
+    public BinaryContent()
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="BinaryContent"/> class referring to an external uri.
+    /// </summary>
+    public BinaryContent(Uri? uri = null)
```
westey-m, line 87: Missing uri param xml comment.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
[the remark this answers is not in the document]
westey-m: I'd suggest keeping the check until we have a good use case for storing null data, since it's not a breaking change to remove the check, but it is to add it back in again later if we remove it now.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
```diff
@@ -3,118 +3,310 @@
...
+        var newMetadata = this.Metadata as Dictionary<string, object?>;
+        if (newMetadata is null)
+        {
+            // Clone the current read only metadata;
+            newMetadata = new Dictionary<string, object?>();
+            if (this.Metadata is not null)
+            {
+                foreach (var property in this.Metadata)
+                {
+                    newMetadata[$"data-uri-{property.Key}"] = property.Value;
```
westey-m, line 196: If I'm reading this right, it seems like this bit of code will only execute if Metadata is not castable to Dictionary<string, object>, and in that case we copy everything that was in the IReadOnlyDictionary<string, object?> to the new dictionary, but we rename each value by prepending `data-uri-`.  I don't understand why we are doing this.  Am I missing something?

dotnet/src/InternalUtilities/src/Text/DataUriParser.cs
```diff
@@ -0,0 +1,168 @@
...
+            model.MimeType = metadataParts[0];
+        }
+
+        for (int i = 1; i < metadataParts.Length; i++)
+        {
+            var part = metadataParts[i];
+            if (part!.Contains("="))
+            {
+                var keyValue = part.Split('=');
+                if (keyValue.Length == 2)
```
westey-m, line 113: Since we checked above that part contains `=`, wouldn't there always be two keyvalue parts?

dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs
```diff
@@ -109,11 +109,22 @@ public async Task DeleteFileAsync(string id, CancellationToken cancellationToken
     /// <remarks>
     /// Files uploaded with <see cref="OpenAIFilePurpose.Assistants"/> do not support content retrieval.
     /// </remarks>
-    public BinaryContent GetFileContent(string id, CancellationToken cancellationToken = default)
+    public async Task<BinaryContent> GetFileContentAsync(string id, CancellationToken cancellationToken = default)
```
westey-m, line 112: Would it make sense to keep the non async version and obsolete it?

westey-m (approved):

dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs
RogerBarreto: The service is experimental, I/O operations like in this service I lean towards having Async over non-async.

dotnet/src/InternalUtilities/src/Text/DataUriParser.cs
RogerBarreto: No, in a invalid dataUri, the user could provide more than one `=` and this would be a Format Error, good catch on this one added another check and throwing scenario for this.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
RogerBarreto: You are right, good catch there, I added a UT to ensure I don't rename existing non-data-uri parameters and changed the logic on this block.

dotnet/src/SemanticKernel.Abstractions/Contents/BinaryContent.cs
RogerBarreto: I will keep the description simple, as this Uri can be there just as informative or extra, not necessarily implying the caller will do something specific with it.

dotnet/src/Connectors/Connectors.OpenAI/CompatibilitySuppressions.xml
```diff
@@ -22,6 +22,13 @@
     <Right>lib/netstandard2.0/Microsoft.SemanticKernel.Connectors.OpenAI.dll</Right>
     <IsBaselineSuppression>true</IsBaselineSuppression>
   </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Microsoft.SemanticKernel.Connectors.OpenAI.OpenAIFileService.GetFileContent(System.String,System.Threading.CancellationToken)</Target>
+    <Left>lib/netstandard2.0/Microsoft.SemanticKernel.Connectors.OpenAI.dll</Left>
+    <Right>lib/netstandard2.0/Microsoft.SemanticKernel.Connectors.OpenAI.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
```
markwallace-microsoft, line 31: I need to review this PR and understand the breaking change.

Are there any options that will allow us introduce the new pattern and also avoid the breaking change?

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionResultContent.cs
```diff
@@ -16,7 +16,7 @@ public sealed class FunctionResultContent : KernelContent
     /// The function call ID.
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string? Id { get; }
+    public string? CallId { get; }
```
markwallace-microsoft, line 19: Why do we need this change?

We should be avoiding breaking changes.

dotnet/src/SemanticKernel.Abstractions/Contents/ImageContent.cs
```diff
@@ -1,27 +1,17 @@
...
 
 namespace Microsoft.SemanticKernel;
 
 /// <summary>
 /// Represents image content.
 /// </summary>
-public sealed class ImageContent : KernelContent
+#pragma warning disable SKEXP0010 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+public class ImageContent : BinaryContent
+#pragma warning restore SKEXP0010 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
```
markwallace-microsoft, line 13: Should this be `SKEXP0001` ?

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionResultContent.cs
RogerBarreto: This was decided in the ADR discussion regarding the confusion on this being a `Call` or a `Result` identifier.

dotnet/src/Connectors/Connectors.OpenAI/CompatibilitySuppressions.xml
RogerBarreto: Added a detailed description of the behavioral breaking changes associated with the `ImageContent`

crickman (approved):

dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs
```diff
@@ -109,11 +109,22 @@ public async Task DeleteFileAsync(string id, CancellationToken cancellationToken
...
     {
         Verify.NotNull(id, nameof(id));
+        var (stream, mimetype) = await this.StreamGetRequestAsync($"{this._serviceUri}/{id}/content", cancellationToken).ConfigureAwait(false);
 
-        return new BinaryContent(() => this.StreamGetRequestAsync($"{this._serviceUri}/{id}/content", cancellationToken));
+        using (stream)
```
crickman, line 117: Is there truely no way to avoid duplicating the stream to memory.  The file might be quite large.  This could create undue memory pressure for production services.

RogerBarreto: I will be adding the `Stream` property to the `BinaryContent` in a following up PR for those scenarios where we don't want to allocate huge amounts of memory when using `BinaryContents`

dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs
```diff
@@ -109,11 +109,22 @@ public async Task DeleteFileAsync(string id, CancellationToken cancellationToken
...
+        {
```
crickman, line 118: image

I'd assert this would be every scenario (why copy stream if not required).  In any case, is there a tracking issue for this follow-up?

RogerBarreto: > I'd assert this would be every scenario (why copy stream if not required). In any case, is there a tracking issue for this follow-up?

Not sure I understand your message (why copy stream if not required).

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -339,15 +339,19 @@ internal ClientCore(ILogger? logger = null)
         PromptExecutionSettings? executionSettings,
         CancellationToken cancellationToken)
     {
-        Verify.NotNull(content.Data);
+        var audioData = content.Data!.Value;
```
dmytrostruk, line 342: You removed `Verify.NotNull` and used `content.Data!`, but `content.Data` can still be null (`AudioContent` has empty constructor). In previous implementation this case would throw `ArgumentNullException`, but I tested it now and it throws `System.InvalidOperationException : Nullable object must have a value.` without information which parameter is not initialized.

dotnet/src/Connectors/Connectors.OpenAI/Connectors.OpenAI.csproj
```diff
@@ -8,6 +8,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <NoWarn>$(NoWarn);NU5104;SKEXP0001,SKEXP0010</NoWarn>
     <EnablePackageValidation>true</EnablePackageValidation>
+    <ApiCompatGenerateSuppressionFile>true</ApiCompatGenerateSuppressionFile>
```
dmytrostruk, line 11: Do we need this? I can see it only in 2 assemblies now, but in order to generate suppression files when you know there are breaking changes, you can run `dotnet build /p:ApiCompatGenerateSuppressionFile=true` and it should do the job.
If we want to generate suppression files every time in case if needed for any assembly, it's probably better to specify it on `Directory.Build.props` level.

dotnet/src/Connectors/Connectors.OpenAI/Files/OpenAIFileService.cs
```diff
@@ -156,8 +167,7 @@ public async Task<OpenAIFileReference> UploadContentAsync(BinaryContent fileCont
 
         using var formData = new MultipartFormDataContent();
         using var contentPurpose = new StringContent(this.ConvertPurpose(settings.Purpose));
-        using var contentStream = await fileContent.GetStreamAsync().ConfigureAwait(false);
-        using var contentFile = new StreamContent(contentStream);
+        using var contentFile = new ByteArrayContent(fileContent.Data!.Value.ToArray());
```
dmytrostruk, line 170: Same here, should we add `Verify.NotNull(fileContent.Data);` at the beginning of the method?

dotnet/src/InternalUtilities/src/Text/DataUriParser.cs
```diff
@@ -0,0 +1,172 @@
...
+#pragma warning disable CA1847 // Use StringBuilder.Append when concatenating strings
+
+namespace Microsoft.SemanticKernel.Text;
+
+/// <summary>
+/// Data Uri Scheme Parser based on RFC 2397.
+/// https://datatracker.ietf.org/doc/html/rfc2397
+/// </summary>
+[ExcludeFromCodeCoverage]
+static internal class DataUriParser
```
dmytrostruk, line 18: nit :)
```suggestion
internal static class DataUriParser
```

dotnet/src/SemanticKernel.Abstractions/SemanticKernel.Abstractions.csproj
```diff
@@ -6,6 +6,7 @@
     <TargetFrameworks>net8.0;netstandard2.0</TargetFrameworks>
     <NoWarn>$(NoWarn);SKEXP0001</NoWarn>
     <EnablePackageValidation>true</EnablePackageValidation>
+    <ApiCompatGenerateSuppressionFile>true</ApiCompatGenerateSuppressionFile>
```
dmytrostruk, line 9:
```suggestion
```

dmytrostruk (approved):

dotnet/samples/Concepts/AudioToText/OpenAI_AudioToText.cs
```diff
@@ -42,7 +42,7 @@ public async Task AudioToTextAsync()
         // Read audio content from a file
         await using var audioFileStream = EmbeddedResource.ReadStream(AudioFilename);
         var audioFileBinaryData = await BinaryData.FromStreamAsync(audioFileStream!);
-        AudioContent audioContent = new(audioFileBinaryData);
+        AudioContent audioContent = new(audioFileBinaryData, mimeType: null);
```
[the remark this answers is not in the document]
RogerBarreto: I'm leaving this as non optional for 2 reasons:

1. Every time we instantiate a binary with bytearray the caller has provide the mimetype and don't forget about it if they have. 
2. Leaving this as optional this can make this property unintentionally neglected.
3. In some scenario's other API's don't provide us with enough detail about the mimeType and for those the caller has to thing how to get it or if not possible provide `null` knowing that eventually he won't be able to get a `data-uri` from this.

Issue #6233, opened 2024-05-13
.Net: Use Fluid package to parse liquid template


LittleLittleCloud:
Replace the liquid template engine from Scriban to Fluid.Core
- Fluid.Core: https://github.com/sebastienros/fluid

PR #6320, opened 2024-05-17
.Net: Update LiquidPromptTemplate to use Fluid instead of Scriban

stephentoub:
Closes https://github.com/microsoft/semantic-kernel/issues/6233
cc: @sebastienros, @LittleLittleCloud, @JakeRadMSFT


dotnet/src/Extensions/PromptTemplates.Liquid/LiquidPromptTemplate.cs
```diff
@@ -231,27 +235,51 @@ public static HashSet<string> InferInputs(Template template)
             return visitor._variables;
         }
 
-        public override void Visit(ScriptVariableGlobal node)
+        public override Statement Visit(Statement statement)
```
sebastienros, line 238: Nice! This could almost be part of a default visitor, but at the same time it's easy to implement. What I was suggesting wat to override `VisitForExpression` (and also VisitAssignExpression) but this technique is more generic and works as well.

sebastienros (approved):

dotnet/src/Extensions/PromptTemplates.Liquid/LiquidPromptTemplate.cs
```diff
@@ -55,12 +57,8 @@ public LiquidPromptTemplate(PromptTemplateConfig config, bool allowDangerouslySe
 
         // Parse the template now so we can check for errors, understand variable usage, and
         // avoid having to parse on each render.
-        this._liquidTemplate = Template.ParseLiquid(config.Template);
-        if (this._liquidTemplate.HasErrors)
-        {
-            throw new ArgumentException($"The template could not be parsed:{Environment.NewLine}{string.Join(Environment.NewLine, this._liquidTemplate.Messages)}");
-        }
-        Debug.Assert(this._liquidTemplate.Page is not null);
+        this._liquidTemplate = s_parser.Parse(config.Template) ??
```
sebastienros, line 60: Not sure if you need to report the errors, then is an overload that returns them.

Ultimately you can even provide a visitor that would detect undefined members. Something that was asked quite frequently, might come OOB at some point now that we have visitors.

markwallace-microsoft (approved):

westey-m (approved):

PR #6347, opened 2024-05-21
Simplify README.md

matthewbolanos:
Simplifying readme until further changes are available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,45 +108,6 @@ Finally, refer to our API references for more details on the C# and Python APIs:
 - [C# API reference](https://learn.microsoft.com/en-us/dotnet/api/microsoft.semantickernel?view=semantic-kernel-dotnet)
 - Python API reference (coming soon)
 
-## Chat Copilot: see what's possible with Semantic Kernel
-
-If you're interested in seeing a full end-to-end example of how to use Semantic Kernel, check out
-our [Chat Copilot](https://github.com/microsoft/chat-copilot) reference application. Chat Copilot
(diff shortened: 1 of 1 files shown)


stephentoub:
Can we also clean up earlier in the README, e.g. where the beginning of it highlights planners even though that's no longer a focus?

stephentoub (approved):

stephentoub:
> Can we also clean up earlier in the README, e.g. where the beginning of it highlights planners even though that's no longer a focus?

If it were me, I'd replace everything before the "Getting started with Semantic Kernel" with something like this:

---

[Semantic Kernel](https://learn.microsoft.com/en-us/semantic-kernel/overview/) is a set of libraries for easily integrating AI into applications and services implemented with C#, Java, and Python.

It provides:
- abstractions for AI services (such as chat, text to images, audio to text, etc.) and memory stores
- implementations of those abstractions for services from [OpenAI](https://platform.openai.com/docs/introduction), [Azure](https://azure.microsoft.com/en-us/products/ai-services/openai-service), [Hugging Face](https://huggingface.co/), and more, and for a multitude of vector databases, such as those from Chroma, Qdrant, Milvus, and Azure
- a common representation for plugins, which can then be orchestrated automatically by AI
- the ability to create such plugins from a multitude of sources, including from OpenAPI specifications, prompts, and arbitrary code written in the target language
- extensible support for prompt management and rendering, including built-in handling of common formats like Handlebars and Liquid
- and a wealth of functionality layered on top of these abstractions, such as filters, dependency injection integration, and more.

alliscode (approved):

PR #6358, opened 2024-05-22
Update README with some suggestions for the opening description

stephentoub:
Suggested rewrite of the beginning information describing what is SK


stephentoub:
@matthewbolanos, curious if you think this is headed in the right direction?

Issue #5887, opened 2024-04-16
.Net: Create ADR for the updated Memory Connector design


markwallace-microsoft:
# Tasks
- [x] Create ADR
- [x] Azure AI Search POC
- [x] Redis POC
- [x] Qadrant POC

## Context and Problem Statement

Semantic Kernel has a collection of connectors to popular Vector databases e.g. Azure AI Search, Chroma, Milvus, ...
Each Memory connector implements a memory abstraction defined by Semantic Kernel and allows developers to easily intergrate Vector databases into their applications.
The current abstractions are experimental and the purpose of this ADR is to progress the design of the abstractions so that they can graduate to non experimental status.

## Decision Drivers

- API surface must be easy to use and intuitive
    - Alignment with other patterns in the SK
- Design must allow Memory Plugins to be easily instantiated with any connector
- Design must support all Kernel content types
- Design must allow for database specific configuration
- All NFR's to be production ready are implemented
- Basic CRUD operations must be supported so that connectors can be used in a polymorphic manner
- Official Database Clients must be used where available
- Dynamic database schema must be supported
- Dependency injection must be supported
- Azure-ML YAML format must be supported
- Breaking glass scenarios must be supported

## Questions

- Primary use cases are described in the RAG document.
- Prioritised list of Memory connectors
    - P1: Azure AI Search, Weaviate, Redis
    - P2: Cosmos dB, Kusto, SQL Server, LlamaIndex*
    - P3: Milvus, Pinecone, Chroma
    - Remove database that do not support vector similarity search

* For LlamaIndex this is an integration story with their service offering

## Resources

- https://github.com/microsoft/kernel-memory
- https://github.com/microsoft/semantic-kernel/issues?q=is%3Aopen+is%3Aissue+label%3A%22memory+connector%22
- LlamIndex, LangChain, Haystack

PR #6364, opened 2024-05-22
.Net: Add Updated Memory Connector ADR and IMemoryRecordService interface with related classes.

westey-m:
Adding an ADR for evolving memory connectors.

POC implementations of the included interface can be viewed here:
- https://github.com/westey-m/semantic-kernel-1/blob/poc-vector-store/dotnet/src/Connectors/Connectors.Memory.AzureAISearch/AzureAISearchVectorDBRecordService%7BTDataModel%7D.cs
- https://github.com/westey-m/semantic-kernel-1/blob/poc-vector-store/dotnet/src/Connectors/Connectors.Memory.Qdrant/QdrantVectorDBRecordService%7BTDataModel%7D.cs
- https://github.com/westey-m/semantic-kernel-1/blob/poc-vector-store/dotnet/src/Connectors/Connectors.Memory.Redis/RedisVectorDBRecordService%7BTDataModel%7D.cs

Related work item:
#5887 

### Motivation and Context

See included ADR for more motivation and context.

### Description

See included ADR for more more info.

### Contribution Checklist


eavanvalkenburg (commented):
small note

SergeyMenshykh:
LGTM

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,749 @@
...
+2. The `IMemoryStore` implementations are opinionated around key encoding / decoding and collection name sanitization, which limits its usability by customers with existing data sets.
+
+Responsibilities:
+
+|Functional Area|Cardinality|Significance to Semantic Kernel|
+|-|-|-|
+|Collection/Index create|An implementation per store type and model|Valuable when building a store and adding data|
+|Collection/Index list/exists/delete|An implementation per store type|Valuable when building a store and adding data|
+|Data Storage and Retrieval|An implementation per store type|Valueble when building a store and adding data|
+|Vector Search|An implementation per store type, model and search type|Valuable for RAG scenarios|
```
dluc, line 32: I would argue this is not just about RAG, but one of the main goals of "memory" as a feature. For instance, search by similarity can be used in these scenarios:
* when receiving user input, search for similar input to understand the User Intent
* when planning for a given goal, search for relevant functions to execute
* when building a prompt with examples (few shots learning), search for relevant examples
* when saving a "memory", search for similar memories to be merged/updated
* when validating user input, search for contradicting facts

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,749 @@
...
+}
+```
+
+### Actions
+
+1. The `IMemoryStore` should be split into four different interfaces, one for each responsibility.
+2. The **Data Storage and Retrieval** and **Vector Search** areas should allow typed access to data and support any schema that is currently available in the customer's data store.
+3. The collection / index create functionality should allow developers to create their own implementations and support creating first party collections for built in functionality. Each implementation would be for a specific schema and data store type.
+4. The collection / index list/exists/delete functionality should allow management of any collection regardless of schema. There should be one implementation for each data store type.
+5. Remove opinionated behaviors from connectors. The opinionated behavior limits the ability of these connectors to be used with pre-existing vector databases. As far as possible these behaviors should be moved into decorators or be injectable.  Examples of opinionated behaviors:
```
dluc, line 77: One reason for past opionated behavior was the ability to switch from one DB to another, without changing code. 
For instance, Qdrant supports only Guid/Numeric IDs, while Azure AI Search supports strings. When working locally as a developer I need to use Qdrant to save money, and when I deploy my code I need to use Azure AI Search to scale and all the other benefits. 
As a developer I think the ideal solution allows to switch from Qdrant to Azure AI Search without changing code (e.g. GUID vs strings), for example just changing the dependency injection settings.

dluc: On a related note, there should be a consistent set of exceptions thrown by connectors, so that try-catch blocks are agnostic about the underlying connector used.

dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecordAttributes/KeyAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.SemanticKernel.Memory;
+
+/// <summary>
+/// Attribute to mark a property on a class as the key under which data is stored in a vector store.
+/// </summary>
+[Experimental("SKEXP0001")]
+[AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+public sealed class KeyAttribute : Attribute
```
markwallace-microsoft, line 13: Any concerns that this generic attribute name could conflict (same for others)?

Should we prefix the attribute names with `MemoryRecord`?

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+Variation on option 3. 
+
+```cs
+
+interface IMemoryCollectionCreateService
+{
+    virtual Task CreateCollectionAsync(string name, CancellationToken cancellationToken = default);
+}
+
+interface IMemoryCollectionUpdateService
```
markwallace-microsoft, line 487: Naming is a little odd given this provides Read and Delete but no Update

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+{
+    IAsyncEnumerable<string> ListCollectionNamesAsync(CancellationToken cancellationToken = default);
+    Task<bool> CollectionExistsAsync(string name, CancellationToken cancellationToken = default);
```
markwallace-microsoft, line 490: Why not  `GetCollectionAsync` and return some metadata?

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+```cs
+
+interface IMemoryCollectionCreateService
+{
+    virtual Task CreateCollectionAsync(string name, CancellationToken cancellationToken = default);
+}
+
+interface IMemoryCollectionUpdateService
+{
+    IAsyncEnumerable<string> ListCollectionNamesAsync(CancellationToken cancellationToken = default);
```
markwallace-microsoft, line 489: Should we allow returns additional metadata for a collection?

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+Same as option 3 / 4, plus:
+
+```cs
+
+interface IMemoryService : IMemoryCollectionCreateService, IMemoryCollectionService, IMemoryRecordService
+{    
+}
+
+// Create a static factory that produces one of these, so only the interface is public, not the class.
+internal class CombinedMemoryService<TDataModel>(IMemoryCollectionCreateService creation, IMemoryCollectionService collections, IMemoryRecordService<TDataModel> records): IMemoryService
```
markwallace-microsoft, line 531: Is `IMemoryCollectionService` already a combination of `IMemoryCollectionCreateService` and `IMemoryCollectionUpdateService`?

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+    }
+```
+
+#### Option 2 - Normalization in decorator
+
+- Pros: Allows normalization to vary separately from the record service.
+- Pros: No code executed when no normalization required.
+- Pros: Easy to package matching encoders/decoders together.
+- Pros: Easier to obsolete encoding/normalization as a concept.
+- Cons: Need to implement the full MemoryRecordService interface.
```
markwallace-microsoft, line 579: Is this really a Con based on our current requirements or should this be Neutral? 

We can introduce an `IReadOnlyMemoryRecordService` later if needed?

dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecordAttributes/KeyAttribute.cs
westey-m: These will certainly clash.  There are similar attributes in a number of other libraries.  Both EF and the Odata libraries e.g. have a KeyAttribute.  I had what you are proposing first but the length of the name wasn't great.  That said, it's a tradeoff between having the long names, and requiring people to re-map the class names if there are clashes which also sucks.  Happy to go with the consensus here.

docs/decisions/0045-updated-vector-store-design.md
westey-m: Makes sense, I've included a few more of the examples here in the doc to clarify.

markwallace-microsoft (approved):
I've added some minor comments bit otherwise this looks great

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+1. The `IMemoryStore` interface has four responsibilities with different cardinalities.
+2. The `IMemoryStore` interface only supports a fixed schema for data storage, retrieval and search, which limits its usability by customers with existing data sets.
+2. The `IMemoryStore` implementations are opinionated around key encoding / decoding and collection name sanitization, which limits its usability by customers with existing data sets.
+
+Responsibilities:
+
+|Functional Area|Cardinality|Significance to Semantic Kernel|
+|-|-|-|
+|Collection/Index create|An implementation per store type and model|Valuable when building a store and adding data|
+|Collection/Index list/exists/delete|An implementation per store type|Valuable when building a store and adding data|
```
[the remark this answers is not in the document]
westey-m: We actually don't support read right now for collections, i.e. there is no simple get.  I'll update the line there to clarify.  It should really say `Collection/Index list names, exists and delete`, since the closest we get to read is listing the names of indices.  I've previously updated the interface method name to ListCollectionNames already since it was ambiguous too.

You may ask why we don't have a simple get.  If we had that what you retrieve will essentially be a complex schema specific to that implementation and it would need to be converted to a common type.  I think having a common index definition that captures all of the available functionality would be tricky, but certainly having a cut down version is possible.  That is fine when creating, since anything that isn't in the definition just isn't supported, but on get, we would lose information, and that may not be expected for users.

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+
+### Support for different mappers
+
+Mapping between data models and the storage models can also require custom logic depending on the type of data model and storage model involved.
+
+I'm therefore proposing that we allow mappers to be injectable for each `MemoryRecordService` instance. The interfaces for these would vary depending
+on the storage models used by each vector store and any unique capabilities that each vector store may have, e.g. qdrant can operate in `single` or
+`multiple named vector` modes, which means the mapper needs to know whether to set a single vector or fill a vector map.
+
+In addition to this, we should build first party mappers for each of the vector stores, which will cater for built in, generic models or use metadata to perform the mapping.
```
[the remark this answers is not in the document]
westey-m: `If we have support for arbitrary data models do we need this as a P1 requirement?`
When I say `we should build first party mappers`, it's a mapper that supports arbitrary data models that I'm referring to.
That said, these mappers are making some assumptions around what the consumer data model will look like.  E.g. it expects the consumer data model to more or less be a flat list of fields.  Supporting pluggable mappers that can convert to some other model with e.g. a Key, VectorsDictionary, DataDictionary, will be needed for break glass scenarios.  Make sense?

docs/decisions/0045-updated-vector-store-design.md
westey-m: As discussed before, agreed, this name has issues.  Do you have an alternative suggestion?

docs/decisions/0045-updated-vector-store-design.md
westey-m: Typically, the only thing valuable here to return is the index configuration.  That varies significantly between different dbs, so having a common description would mean we would lose data as we convert.  Losing data is bad if we also support create, since a developer could rightly assume that they should be able to get, modify and create, but they would get something that deviates from the original in more ways than the modifications they made.

docs/decisions/0045-updated-vector-store-design.md
westey-m: What metadata do you have in mind?

docs/decisions/0045-updated-vector-store-design.md
westey-m: Yeah, that's the idea.  What I'm illustrating here is the ability to create a class that combines the two in a way that makes it easy to just add the bits that are different.  E.g. you grab the update piece from the library, but add your own create implementation.

SergeyMenshykh (approved):

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: proposed
+contact: westey-m
+date: 2024-05-01
+deciders: sergeymenshykh, markwallace, rbarreto, dmytrostruk, westey-m, matthewbolanos
+consulted: stephentoub, dluc
```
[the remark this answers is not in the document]
westey-m: Thanks, fixed.

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,810 @@
...
+
+## Decision Drivers
+
+From GitHub Issue:
+- API surface must be easy to use and intuitive
+- Alignment with other patterns in the SK
+- - Design must allow Memory Plugins to be easily instantiated with any connector
+- Design must support all Kernel content types
+- Design must allow for database specific configuration
+- All NFR's to be production ready are implemented
```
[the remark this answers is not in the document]
westey-m: FYI: This is the list from the associated issue. Can you give an example of what you have in mind?

docs/decisions/0045-updated-vector-store-design.md
westey-m: Updating to clarify.  This isn't a major con, but I'm just contrasting it with option 3, where we can just pass the translation functions and we don't need to do a full wrapper class.

dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecordAttributes/KeyAttribute.cs
dmytrostruk: @westey-m Specifically for `Key` attribute, can we use existing one from `System.ComponentModel.DataAnnotations`? 
https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.keyattribute

We can still keep `VectorAttribute` and others as new ones on our side, but re-use existing ones which we get out of the box in .NET.

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,811 @@
...
+
+#### Option 4 - Separated collection and record management with collection create separate from other operations, with collection management aggregation class on top.
+
+Variation on option 3. 
+
+```cs
+
+interface IMemoryCollectionCreateService
+{
+    virtual Task CreateCollectionAsync(string name, CancellationToken cancellationToken = default);
```
dmytrostruk, line 485: In this option, the signature of `CreateCollectionAsync` method lacks information about collection/index schema (although, some configuration exists in Option 1: `Task CreateCollectionAsync(CollectionConfig collectionConfig, CancellationToken cancellationToken = default);`, it would be interesting to learn more about `CollectionConfig` idea.)

ADR describes how to annotate data model with attributes, whether it's a `Key` or `Vector` field etc, but we don't use this information in `CreateCollectionAsync`. Should we make this signature generic, so it will accept data model and use attributes to identify collection fields?

Below in Option 4, there is example with `abstract Task CreateCollectionAsync` and implementations: `class AzureAISearchChatHistoryMemoryService`, `class AzureAISearchSemanticCacheMemoryService` etc, so it looks like separate class per collection. 

I'm a little bit worried about this approach, because I'm not sure how it will correlate with `TDataModel` from record abstraction:
1. Every time I create a new `TDataModel` to be used in record abstraction (e.g. `Contact`, `Location`, `Message`), do I need to create a new class for collection manipulation (e.g. `AzureAISearchContactCollectionService`, `AzureAISearchLocationCollectionService` etc?
2. If at some point of time I will add a new property to `TDataModel` or rename existing one, should I replicate it in related `CollectionCreateService` class to update collection schema? If schema update is a case, should we consider adding `UpdateCollectionAsync` method as well?

It feels like collection and record abstractions should operate with custom schema in similar way. Having custom schema in record abstraction means that we can update it at any time for record operations, but these changes won't be replicated in actual collection, which may be a problem. Because I will create a collection programmatically with some schema, but as soon as I update schema in code, I will need to manually update collection schema in db. 

Let me know what you think about it. If all these scenarios are out-of-scope, we should also mention that in ADR.

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,811 @@
...
+
+public class StoreOptions
+{
+    public Func<string, string>? EncodeKey { get; init; }
+    public Func<string, string>? DecodeKey { get; init; }
+    public Func<string, string>? SanitizeCollectionName { get; init; }
+}
+```
+
+### Option 4 - Normalization via custom mapper
```
dmytrostruk, line 607:
```suggestion
#### Option 4 - Normalization via custom mapper
```

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,811 @@
...
+- Cons: Harder to package matching encoders/decoders together.
+
+```cs
+public class AzureAISearchMemoryRecordService<TDataModel>(StoreOptions options);
+
+public class StoreOptions
+{
+    public Func<string, string>? EncodeKey { get; init; }
+    public Func<string, string>? DecodeKey { get; init; }
+    public Func<string, string>? SanitizeCollectionName { get; init; }
```
dmytrostruk, line 603: I'm wondering if we need normalization logic only for key and collection name or for other fields as well? Maybe we should introduce some optional field mapper/transformer that will work for any property?

The great thing about custom schema and vector DB abstractions is the ability to use any of vector DB implementations with the same schema and easily switch between them. Just like `Key` property format may be compatible with Qdrant but not with Azure AI Search, similar situations may happen when supporting different property types, especially datetimes, boolean flags (previously there was an issue with some connector that used `0` and `1` for boolean and then switched to `true`/`false`). It feels like optional mapper should handle any data model property. This should provide more flexibility for users.

dotnet/src/SemanticKernel.Abstractions/Memory/IMemoryRecordService{TKey,TDataModel}.cs
```diff
@@ -0,0 +1,80 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SemanticKernel.Memory;
+
+/// <summary>
+/// An interface for adding, updating, deleting and retrieving records from a memory store.
```
markwallace-microsoft, line 11: - Should this abstraction be an API for interacting with vector data stores only rather than the more generic memory store?
- Should we restrict what we store to always having associated Vector field(s)?

dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecordAttributes/KeyAttribute.cs
westey-m: @dmytrostruk, I am concerned that we may need to support additional properties on the Key and other attributes to support create and filtered search scenarios.  E.g. Azure AI Search has many options you can set on a Key, see https://learn.microsoft.com/en-us/azure/search/search-get-started-vector#create-a-vector-index
If any of these scenarios make sense across most of the databases, it would make sense for us to support them too.

Having our own attribute will allow us to add these as/if needed without requiring users to switch to a new type.

I've updated the doc with my reasoning as well, but let's chat more directly later.

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,811 @@
...
+Variation on option 3. 
+
+```cs
+
+interface IMemoryCollectionCreateService
+{
+    virtual Task CreateCollectionAsync(string name, CancellationToken cancellationToken = default);
+}
+
+interface IMemoryCollectionUpdateService
```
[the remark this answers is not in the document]
westey-m: Thanks for the feedback.  My reasoning was in the Decision Outcome section, but I added more detail just now, to try and explain it in more detail.

The main reason for splitting out collection create is because I expect that there will be multiple implementations of this, even for the same database type, while list, exists and delete are the same across all and will only require a single implementation.

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,811 @@
...
+{
+    IAsyncEnumerable<string> ListCollectionNamesAsync(CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
westey-m: The reason I changed it was because we cannot make assumptions about how many collections may exist. Depending on how they are defined, especially in a relational database, if .e.g. you have a collections table and a records table, there could be a very large number.  `IAsyncEnumerable` allows getting data from the store in pages without needing to make the caller wait until all are retrieved before returning.

docs/decisions/0045-updated-vector-store-design.md
westey-m: I've updated the Option 3 to clarify a bit:
```cs
// Implement a generic version of create that takes a configuation that should work for 80% of cases.
class AzureAISearchConfiguredCollectionCreateService(CollectionCreateConfig collectionConfig): IMemoryCollectionCreateService;

// Allow custom implementations of create for break glass scenarios for outside the 80% case.
class AzureAISearchChatHistoryCollectionCreateService: IMemoryCollectionCreateService;
class AzureAISearchSemanticCacheCollectionCreateService: IMemoryCollectionCreateService;
```

The thinking was to not have a create config in the method parameter as in option 1, since it'll be hard to cover all the different features from all the different dbs.  Instead, we can have a generic create implementation that takes a create config in the constructor, but there can be other create implementations too, that take different configuration.

Having a single config for the record and create has benefits and drawbacks.  On the one hand it means that you may need to add create configuration even if you don't want to use create, or if you are implementing a custom create method.  On the other hand, it avoids the issues you are raising above, where you may need to update two configs when something changes.  Let's discuss this more in person later.

stephentoub:
cc: @roji, @ajcvickers

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,860 @@
...
+I'm proposing that we allow two ways in which to provide the information required to map data between the consumer data model and storage data model.
+First is a set of configuration objects that capture the types of each field. Second would be a set of attributes that can be used to decorate the model itself
+and can be converted to the configuration objects, allowing a single execution path.
+Additional configuration properties can easily be added for each type of field as required, e.g. IsFilterable or IsFullTextSearchable, allowing us to also create an index from the provided configuration.
+
+I'm also proposing that even though similar attributes already exist in other systems, e.g. System.ComponentModel.DataAnnotations.KeyAttribute, we create our own.
+We will likely require additional properties on all these attributes that are not currently supported on the existing attributes, e.g. whether a field is or
+should be filterable. Requiring users to switch to new attributes later will be disruptive.
+
+Here is what the attributes would look like, plus a sample use case.
```
stephentoub, line 315: How does this relate to anything in EF today? Is there opportunity for convergence?

docs/decisions/0045-updated-vector-store-design.md
```diff
@@ -0,0 +1,860 @@
...
+}
+
+// Create a static factory that produces one of these, so only the interface is public, not the class.
+internal class CombinedMemoryService<TDataModel>(IMemoryCollectionCreateService creation, IMemoryCollectionService collections, IMemoryRecordService<TDataModel> records): IMemoryService
+{
+}
+
+```
+
+#### Decision Outcome
```
stephentoub, line 546: I don't see any example consumption code in this design. What does it look like for typical scenarios to query for the appropriate interfaces from DI and use them? For example, the above shows multiple implementations of MemoryCollectionService... how are those being registered in DI, how are those being queried for, etc.?

dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecordAttributes/KeyAttribute.cs
dmytrostruk: Makes sense, thank you!

docs/decisions/0045-updated-vector-store-design.md
westey-m: The way I'm thinking about it is that EF's focus is on relational / SQL like databases, but here we need to support simple index / vector stores as well, so I think the focus is both broader, in terms of what stores are supported, and narrower in terms of what functionality is supported across all these.  The lowest common denominator between all these is certainly a lot lower than what EF is targeting.  Makes sense?

docs/decisions/0045-updated-vector-store-design.md
westey-m: I haven't designed the DI experience yet. I was hoping to focus on the core data interfaces first. That said, the scenario you mention is very important to get right. Let's chat about it directly, I'd like to get your input on what you'd like to see here.

dotnet/src/SemanticKernel.Abstractions/Memory/IMemoryRecordService{TKey,TDataModel}.cs
westey-m: I think the definition of a vector store is becoming more difficult since all databases and indexes seem to be adding vector support right now.  Requiring users to specify a vector field is also tricky, since there may be cases where the embedding is generated in the service from a provided field, and on query the service again generates an embedding from the query text, does the comparison and returns the text.  The client may never have to do anything with vectors.
I think therefore Memory makes sense in that it restricts the use cases to memory access.

docs/decisions/0045-updated-vector-store-design.md
westey-m: As discussed in person, I'll work on this in the next iteration.

docs/decisions/0045-updated-vector-store-design.md
westey-m: Let's keep thinking about this, and try and address before going to main.

dotnet/src/SemanticKernel.Abstractions/Memory/IMemoryRecordService{TKey,TDataModel}.cs
westey-m: Still getting input from various people on this as well, so let's keep iterating on this.

docs/decisions/0045-updated-vector-store-design.md
dluc: > especially if they have already created their own database and want to read/write to it using our code

I think we're on the same page here. Bring your own datasource is a top priority. My preference on this matter is to address this with configuration ((which field is what, aka mappers). I've seen attempts at using custom classes to map custom schemas, and the code get very complex very quickly, e.g. due to an excessive use of generics that are coupled to specific storage types. In my experience, relying on configuration makes it very easy to switch storage without changing code, and allows to use the same approach across different programming languages.

PR #6393, opened 2024-05-24
.Net: Support records with no key, fix issue with Relevance

### LINKED ISSUE #6379: Error while saving to Semantic Memory using NoSQL
I am trying to store into Semantic Memory using below code

```
public async Task SaveEmbeddingsAsync(string data, string id)
{
    try
    {
        await meomoryWithCosmosNoSQL.SaveReferenceAsync(
            collection: MemoryCollectionName,
            text: data,
            externalId:id,
            externalSourceName:"RecipeApp"
            );


    }
    catch (Exception ex)
    {
        Console.WriteLine(ex.ToString());   
    }
}
```
I am getting the below exception and can’t make out what it refers to as input name.

"Microsoft.Azure.Cosmos.CosmosException : Response status code does not indicate success: BadRequest (400); Substatus: 0; ActivityId: 76cf2457-164b-49c0-b249-6e929bef8dfc; Reason: (Message: {\"Errors\":[\"The input name '' is invalid. 

@Pilchie

Pilchie:
### Motivation and Context

When using SemanticTextMemory.SaveReferenceAsync, no value is passed for the `Key` to use. In the Cosmos DB Memory Connector, we treat the "Key" as the Id and PartitionKey, and so it has to have a value. Address this by generating a `GUID` to use if none are provided.

### Description

* Assign a GUID key when the passed in `MemoryRecord.Key` is null or empty so that CosmosDB always has a valid PartitionKey.
* While adding test coverage for this I noticed two things:
1. `withEmbeddings` is ignored in `AzureCosmosDBNoSQLMemoryStore.GetAsync`. It's not clear whether it's worth changing this, as point reads like the current one tend to be less expensive than queries in Cosmos DB.
2. The SimilarityScore returned from `GetNearestAsync` wasn't actually a similarity, it was a cosine distance.  Normalize that.

Fixes #6379.

### Contribution Checklist


dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -332,7 +339,7 @@ FROM x
             {
                 if (memoryRecord.SimilarityScore >= minRelevanceScore)
                 {
-                    yield return (memoryRecord, memoryRecord.SimilarityScore);
+                    yield return (memoryRecord, 1 - memoryRecord.SimilarityScore);
```
Pilchie, line 342: I'm following up on whether we should always do this, or if it depends on the configured distance function in Cosmos DB.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -164,6 +164,12 @@ await this._cosmosClient
         MemoryRecord record,
         CancellationToken cancellationToken = default)
     {
+        // In some cases we're expected to generate the key to use. Do so if one isn't provided.
+        if (string.IsNullOrEmpty(record.Key))
+        {
+            record.Key = Guid.NewGuid().ToString();
```
Pilchie, line 170: Is this valid to do? We're modifying the input param here.

dmytrostruk: This is temporary solution which exists in other memory connectors as well. We will remove it as soon as new memory abstraction will be in place.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
Pilchie: @jcodella - any thoughts about this? Do we need to remember the distance function passed in and have a different operation here?

Pilchie: Restricted to just `Cosine` and corrected this calculation.  @jcodella does that seem reasonable.

Pilchie:
@dmytrostruk - can you take another look here?  I restricted the provider to only use `Cosine` distance in order to be able to calculate the `0..1` relevance score.

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -71,14 +71,22 @@ public class AzureCosmosDBNoSQLMemoryStore : IMemoryStore, IDisposable
         VectorEmbeddingPolicy vectorEmbeddingPolicy,
         IndexingPolicy indexingPolicy)
     {
-        if (!vectorEmbeddingPolicy.Embeddings.Any(e => e.Path == "/embedding"))
+        var embedding = vectorEmbeddingPolicy.Embeddings.FirstOrDefault(e => e.Path == "/embedding");
+        if (embedding is null)
         {
             throw new InvalidOperationException($"""
                 In order for {nameof(GetNearestMatchAsync)} to function, {nameof(vectorEmbeddingPolicy)} should
                 contain an embedding path at /embedding. It's also recommended to include a that path in the
```
dmytrostruk, line 79:
```suggestion
                contain an embedding path at /embedding. It's also recommended to include that path in the
```

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -71,14 +71,22 @@ public class AzureCosmosDBNoSQLMemoryStore : IMemoryStore, IDisposable
...
                 {nameof(indexingPolicy)} to improve performance and reduce cost for searches.
                 """);
         }
+        else if (embedding.DistanceFunction != DistanceFunction.Cosine)
```
dmytrostruk, line 83: It looks like we are accepting `VectorEmbeddingPolicy` but only with following conditions:
1. One of the embeddings should contain `/embedding` path.
2. Embedding with `/embedding` path should have `Cosine` distance function, and no other options are allowed.

We use this policy to create new collection. In this case, can we use `Cosine` distance function and `/embedding` path by default, and instead of accepting `VectorEmbeddingPolicy` in constructor, we accept only those parameters that have a support for different values, for example `Dimensions`?

We can add constructor overloads with more configuration in the future, if connector will support it, but for now, it will minimize error probability. As a user, I won't have to check the source code of this connector or manually try to guess which distance function is supported.

Let me know if that makes sense.

Pilchie: I considered that, I could definitely switch to a model that just accepts `DataType` and `Dimensions`. The main reason I didn't is because this actually takes a collection of `Embeddings`, so maybe it's possible that someone would want some other VectorIndex in the container.

I'm definitely open to changing though as it does seem somewhat as though Semantic Kernel expects this container to be mostly a black box.

dmytrostruk: > this actually takes a collection of Embeddings, so maybe it's possible that someone would want some other VectorIndex in the container.

But Embedding object should have `/embedding` path and we will take just first one, even if there are multiple of them in collection. It seems like a lot of limitations are in-place for such configuration model. We can leave it as it is but provide another constructor overload that will accept `DataType` and `Dimensions` for simple configuration.

Now I'm wondering, do we need to extend support for `Euclidean` and `DotProduct` functions? I think we have similar request for Chroma connector, to support more than one distance function, is it doable in this connector?

Pilchie: *This connector* will always use the vectors at `/embedding`, but the user could use the `CosmosClient` with other connectors, or directly to access other embeddings.  But I think you're right that it's worth adding an overload that just takes `DataType` and `Dimensions`.

> Now I'm wondering, do we need to extend support for Euclidean and DotProduct functions?

The issue I ran into is that each distance function has its own range of values:

* [cosine](https://en.wikipedia.org/wiki/Cosine_similarity), which has values from -1 (least similar) to +1 (most similar).
* [dotproduct](https://en.wikipedia.org/wiki/Dot_product), which has values from -inf (least similar) to +inf (most similar).
* [euclidean](https://en.wikipedia.org/wiki/Euclidean_distance), which has values from 0 (most similar) to +inf (least similar).

The current connectors rely on the idea of a relevance from 0-1, which is a pretty lossy scaling for `Euclidean` and `DotProduct`.

dmytrostruk: Makes sense, thanks! I think we should come up with some strategy for distance functions that are not similar to 0-1 scaling, because we already received some requests to support not only `cosine` but other functions as well. 
Tagging @westey-m @markwallace-microsoft for visibility. 

As for now, `cosine` function should be enough and we can extend it later as soon as new memory design will be ready.

westey-m: I think the approach here makes sense until we get to the new design, which would allow custom schemas.
One other thing to consider here is to also verify DataType, since only the public model only supports float32, so if the user picks float16, they will lose precision (maybe ok), but if they pick Uint8 or Int8, that won't work at all.

westey-m (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
Pilchie: Which public model?  There is no guarantee that customers are using this with an OpenAI embedding model, right?

westey-m: Sorry, I meant data model, see
https://github.com/microsoft/semantic-kernel/blob/048be8c86766f882a48c32f4a6c985fc7f6d653d/dotnet/src/SemanticKernel.Abstractions/Memory/MemoryRecord.cs#L22
We only support `ReadOnlyMemory<float>` for the embedding property today.

Pilchie: Got it - I do think getting `float16` from a model, putting it into a `ReadOnlyMemory<float>` for the API, but having it be stored in the db as `float16` is a useful scenario though. It possibly means less data over the wire, and definitely less storage in the DB. But I'm fine either way - if you want me to fix it to `float32` I can do that.

westey-m: I agree float16 and float32 are ok, worse case someone loses some precision if they try and store float32 but use float16 in the config, but hopefully they know what they are doing there 😄.  I think we should throw if someone picks Uint8 or Int8 though.  My understanding is that that will not get you any meaningful results.

Pilchie: Done - new bikeshed is what's the right exception type. I went with `NotSupportedException`, but could see an argument for `ArgumentException` too.

westey-m: Thanks, looks good to me.

westey-m (approved):

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -22,11 +23,62 @@ namespace Microsoft.SemanticKernel.Connectors.AzureCosmosDBNoSQL;
 /// </summary>
 public class AzureCosmosDBNoSQLMemoryStore : IMemoryStore, IDisposable
 {
+    private const string EMBEDDING_PATH = "/embedding";
```
dmytrostruk, line 26:
```suggestion
    private const string EmbeddingPath = "/embedding";
```

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -22,11 +23,62 @@ namespace Microsoft.SemanticKernel.Connectors.AzureCosmosDBNoSQL;
...
+                }),
+            databaseName,
+            new VectorEmbeddingPolicy(
+                [
+                    new Embedding
+                    {
+                        DataType = vectorDataType,
+                        Dimensions = dimensions,
+                        DistanceFunction = DistanceFunction.Cosine,
+                        Path = EMBEDDING_PATH,
```
dmytrostruk, line 66:
```suggestion
                        Path = EmbeddingPath,
```

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -22,11 +23,62 @@ namespace Microsoft.SemanticKernel.Connectors.AzureCosmosDBNoSQL;
...
+                    }
+                ]),
+            new IndexingPolicy
+            {
+                VectorIndexes = new Collection<VectorIndexPath> {
+                    new()
+                    {
+                        Path = EMBEDDING_PATH,
```
dmytrostruk, line 74:
```suggestion
                        Path = EmbeddingPath,
```

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -71,14 +123,29 @@ public class AzureCosmosDBNoSQLMemoryStore : IMemoryStore, IDisposable
         VectorEmbeddingPolicy vectorEmbeddingPolicy,
         IndexingPolicy indexingPolicy)
     {
-        if (!vectorEmbeddingPolicy.Embeddings.Any(e => e.Path == "/embedding"))
+        var embedding = vectorEmbeddingPolicy.Embeddings.FirstOrDefault(e => e.Path == EMBEDDING_PATH);
```
dmytrostruk, line 126:
```suggestion
        var embedding = vectorEmbeddingPolicy.Embeddings.FirstOrDefault(e => e.Path == EmbeddingPath);
```

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
```diff
@@ -71,14 +123,29 @@ public class AzureCosmosDBNoSQLMemoryStore : IMemoryStore, IDisposable
...
+        if (embedding is null)
         {
             throw new InvalidOperationException($"""
                 In order for {nameof(GetNearestMatchAsync)} to function, {nameof(vectorEmbeddingPolicy)} should
-                contain an embedding path at /embedding. It's also recommended to include a that path in the
+                contain an embedding path at {EMBEDDING_PATH}. It's also recommended to include that path in the
```
dmytrostruk, line 131:
```suggestion
                contain an embedding path at {EmbeddingPath}. It's also recommended to include that path in the
```

dmytrostruk (approved):
Small updates on constant naming. We use PascalCase for constants:
https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions

dotnet/src/Connectors/Connectors.Memory.AzureCosmosDBNoSQL/AzureCosmosDBNoSQLMemoryStore.cs
Pilchie: done

dmytrostruk (approved):

PR #5964, opened 2024-04-22
Java: .Net: Add AssemblyAI file service

Swimburger:
### Description

Add AssemblyAI file service that accepts a `Stream` as input for file uploading.
By adding the file service, it removes the need for `AudioStreamContent` which I removed.
In a separate PR I can add a `StreamingAudioContent` class which would take audio from a microphone for example, and send it to AssemblyAI's Streaming STT.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/semantic-kernel/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #6406, opened 2024-05-27
.Net: Add AssemblyAI file service Remake from #5964

RogerBarreto:
### Contribution Checklist


dotnet/src/Connectors/Connectors.AssemblyAI/Files/AssemblyAIFileService.cs
```diff
@@ -0,0 +1,55 @@
...
+            logger: loggerFactory?.CreateLogger(this.GetType()));
+    }
+
+    /// <summary>
+    /// Upload a file.
+    /// </summary>
+    /// <param name="stream">The file stream</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The file metadata.</returns>
+    public async Task<AudioContent> UploadAsync(Stream stream, CancellationToken cancellationToken = default)
```
RogerBarreto, line 49:
```suggestion
    public async Task<AudioContent> UploadAsync(Stream stream, CancellationToken cancellationToken = default)
```
@Swimburger Consider changing the signature of this service to use `IAsyncEnumerable<byte[]>` as the input.

This will avoid `IDisposable` contents to be sent to the service and being managed by both caller / internals.

Swimburger: I don't have permission to update your PR. 

Either way, I only added this remark to the `UploadAsync` method:
```csharp
    /// <remarks>
    /// Caller responsible for disposal of the stream.
    /// </remarks>
```

Since the point of this service is to upload files, the expected type used should be `Stream`.
When users pass a stream to a method, it is expected by precedence that the `Stream` should be disposed by the caller, unless there's an additional parameter that lets you control this behavior.

I considered using `BinaryContent` but the implementation of retrieving the content from this object can potentially duplicate the memory.

RogerBarreto:
@Swimburger, added minor changes to overall code.

- Stopped using `JsonDocument` in favor of `JsonElement` for performance
- Changed structure from `Files` to `Services`, all public services should be in the root namespace.
- Changes `Client` to be `Core` in line with our other connectors.
- Update Namespaces on UnitTests.
- Sealing specialized class.

Swimburger:
Those changes look great. I didn't know about the perf difference, thanks for sharing and updating!

PR #6416, opened 2024-05-27
.Net Simplify configuration by ServiceId on Multi Model Scenarios.

RogerBarreto:
### Motivation and Context

Setting multiple execution settings is not simple and demands creating a dictionary on the caller side to set directly into the `ExecutionSettings` setter property.

This change adds a `ServiceId` property to the execution settings which will be used during the initialization and deserialization to set the expected `Key` in the dictionary as well as the setting for filtering and executing a service specific function invocation.

With this change were also added new constructors for `PromptTemplateConfig`, `KernelArguments` accepting multiple `PromptExecutionSettings` as well as added multiple for `Kernel.CreateFunctionFromPrompt` and `KernelFunctionFromPrompt.Create`

### ServiceId Settings

Before:

```csharp
KernelArguments arguments = [];
arguments.ExecutionSettings = new Dictionary<string, PromptExecutionSettings>()
{
    { serviceId, new PromptExecutionSettings() }
};
var result = await kernel.InvokePromptAsync(prompt, arguments);
```

After:

```csharp
var result = await kernel.InvokePromptAsync(prompt, new(new PromptExecutionSettings { ServiceId = serviceId }));
```

### ModelIds Settings 

Before:

```csharp
string[] modelIds = ["model1", "model2", ...];
var modelSettings = new Dictionary<string, PromptExecutionSettings>();
foreach (var modelId in modelIds)
{
    modelSettings.Add(modelId, new PromptExecutionSettings() { ModelId = modelId });
}
var promptConfig = new PromptTemplateConfig(prompt) { Name = "HelloAI", ExecutionSettings = modelSettings };

var function = kernel.CreateFunctionFromPrompt(promptConfig);
```

After:

```csharp
string[] modelIds = ["model1", "model2", ...];
var function = kernel.CreateFunctionFromPrompt(prompt, modelIds.Select((modelId, index) => new PromptExecutionSettings { ServiceId = $"service-{index}", ModelId = modelId }));
```

The same can be done for ServiceId settings:

```csharp
string[] serviceIds = ["service1", "service2"... ];
var function = kernel.CreateFunctionFromPrompt(prompt, serviceIds.Select(serviceId => new PromptExecutionSettings { ServiceId = serviceId }));
```


dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -27,6 +27,22 @@ public class PromptExecutionSettings
     /// </remarks>
     public static string DefaultServiceId => "default";
 
+    /// <summary>
+    /// Service identifier.
+    /// This identifies the service these settings are configured for e.g., openai, ollama, huggingface, etc.
```
markwallace-microsoft, line 32: The service id is an arbitrary identifier so include examples like `azure_openai_eastus, ...`

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -36,12 +37,32 @@ public KernelArguments()
...
         if (executionSettings is not null)
         {
-            this.ExecutionSettings = new Dictionary<string, PromptExecutionSettings>() { { PromptExecutionSettings.DefaultServiceId, executionSettings } };
+            var newExecutionSettings = new Dictionary<string, PromptExecutionSettings>(executionSettings.Count);
+            foreach (var settings in executionSettings)
+            {
+                var targetServiceId = settings.ServiceId ?? PromptExecutionSettings.DefaultServiceId;
+                if (newExecutionSettings.ContainsKey(targetServiceId))
+                {
+                    throw new ArgumentException("When adding multiple execution settings, the service id needs to be provided and be unique for each.");
```
markwallace-microsoft, line 59: The service id doesn't need to be provided because when not the default one is used. If provided it must be unique and only one default is allowed. So either one long message or one for each case. The message should include the service id so the developer can see which one is causing the issue.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -65,7 +86,34 @@ public KernelArguments(IDictionary<string, object?> source, Dictionary<string, P
...
+                if (kv.Key != kv.Value.ServiceId)
+                {
+                    if (!string.IsNullOrWhiteSpace(kv.Value.ServiceId))
+                    {
+                        throw new ArgumentException($"Service id '{kv.Value.ServiceId}' must match the key '{kv.Key}'.", nameof(this.ExecutionSettings));
+                    }
+
+                    kv.Value.ServiceId = kv.Key;
+                }
+            }
```
markwallace-microsoft, line 114: What happens if someone sets this value and then mutates the Dictionary later? Should we create a copy to prevent this?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -65,7 +86,34 @@ public KernelArguments(IDictionary<string, object?> source, Dictionary<string, P
     /// <summary>
     /// Gets or sets the prompt execution settings.
```
markwallace-microsoft, line 87: Add remarks explaining the checks that are performed when we set.

dotnet/src/SemanticKernel.Abstractions/PromptTemplate/PromptTemplateConfig.cs
```diff
@@ -187,6 +187,25 @@ public List<InputVariable> InputVariables
         {
             Verify.NotNull(value);
             this._executionSettings = value;
+
```
markwallace-microsoft, line 190: Update the API document and do we need a defensive copy here?

dotnet/src/SemanticKernel.UnitTests/Functions/OrderedAIServiceSelectorTests.cs
```diff
@@ -75,8 +75,8 @@ public void ItGetsAIServiceConfigurationForTextGenerationByServiceId()
         Kernel kernel = builder.Build();
 
         var promptConfig = new PromptTemplateConfig() { Template = "Hello AI" };
-        var executionSettings = new PromptExecutionSettings();
-        promptConfig.AddExecutionSettings(executionSettings, "service2");
+        promptConfig.AddExecutionSettings(new(), "service2");
+        var executionSettings = promptConfig.ExecutionSettings["service2"];
```
markwallace-microsoft, line 79: Why make this change? It looks less efficient.

markwallace-microsoft:
Some general feedback

1. When invoking a prompt a likely use case is that you want to target a specific `modelId` rather than `serviceId`. Can you include an example for this if we don't already have one.
1. If I want to invoke a prompt `KernelFunction` and specify the `modelId` or `serviceId` to use, do we have an example for this?

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
RogerBarreto: Yeah, I did this in the PromptExecutionsettings and skipped this, good catch.

dotnet/src/SemanticKernel.UnitTests/Functions/OrderedAIServiceSelectorTests.cs
RogerBarreto: As we internally clone the `ExecutionSettings` and set the `ServiceId` automatically, the original code fails the test because the comparison `Assert.Equivalent` tries to compare the settings without a `ServiceId` with the cloned Settings with an auto-set `ServiceId`.

RogerBarreto:
@markwallace-microsoft

> When invoking a prompt a likely use case is that you want to target a specific modelId rather than serviceId. Can you include an example for this if we don't already have one.

Good point, I added all possible combinations of samples with PreConfigured and per Invocation and simplified a bit more the samples to 2 lines max.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -65,7 +91,38 @@ public KernelArguments(IDictionary<string, object?> source, Dictionary<string, P
...
+    {
+        get => this._executionSettings;
+        set
+        {
+            // Clone the settings to avoid reference changes.
+            this._executionSettings = value is IDictionary<string, PromptExecutionSettings> dictionary
+                ? new Dictionary<string, PromptExecutionSettings>(dictionary)
+                : value?.ToDictionary(kv => kv.Key, kv => kv.Value);
+
+            if (this._executionSettings is not null && this._executionSettings.Count != 0)
```
SergeyMenshykh, line 108:
```suggestion
            if (this._executionSettings is {Count: >0})
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -65,7 +91,38 @@ public KernelArguments(IDictionary<string, object?> source, Dictionary<string, P
...
+    /// </remarks>
+    public IReadOnlyDictionary<string, PromptExecutionSettings>? ExecutionSettings
+    {
+        get => this._executionSettings;
+        set
+        {
+            // Clone the settings to avoid reference changes.
+            this._executionSettings = value is IDictionary<string, PromptExecutionSettings> dictionary
+                ? new Dictionary<string, PromptExecutionSettings>(dictionary)
+                : value?.ToDictionary(kv => kv.Key, kv => kv.Value);
```
dmytrostruk, line 106: Having `IReadOnlyDictionary<string, PromptExecutionSettings>` as property type means that if user has custom implementation of `IReadOnlyDictionary<string, PromptExecutionSettings>`, it's also possible to set it here. But with this logic, when I will try to set `MySettingsDictionary` which implements both `IReadOnlyDictionary` and `IDictionary` or extends `Dictionary<string, PromptExecutionSettings>`, it will override it with `System.Collections.Generic.Dictionary`. 

```csharp
public class MySettingsDictionary : IDictionary<string, PromptExecutionSettings>, IReadOnlyDictionary<string, PromptExecutionSettings>
{
    // Implementation
}

[Fact]
public void Test()
{
    var mySettings = new MySettingsDictionary();
    var arguments = new KernelArguments() { ExecutionSettings = mySettings };

    Console.WriteLine(arguments.ExecutionSettings.GetType());
    // Before change: Namespace.MySettingsDictionary
    // After change: System.Collections.Generic.Dictionary
}
```

Not sure if it's a common case, but if somebody does this today, this change looks like behavioral breaking change. Even if nobody does this, this behavior doesn't look intuitive, because the property is abstracted, but it will always be `Dictionary`, so it makes sense to keep property type as `Dictionary<string, PromptExecutionSettings>`.

RogerBarreto: The change was made to prevent reference changes after you set the `KernelArgument` settings, which prevents unexpected reference impact. 

That said this is an expected behavioral change where only break implementations that leverage an undesirable behavior.

The internal dictionary will still be exposed as `IReadOnlyDictionary` which keeps public API exposure also intact.

On your scenario this would only be a breaking change if the specialized Dictionary don't respect the Liskov Substitution Principle (Intentionally changing the expected Dictionary behavior) which also is a anti-pattern and ensuring code quality from the caller side.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -36,12 +38,36 @@ public KernelArguments()
...
         if (executionSettings is not null)
         {
-            this.ExecutionSettings = new Dictionary<string, PromptExecutionSettings>() { { PromptExecutionSettings.DefaultServiceId, executionSettings } };
+            var newExecutionSettings = new Dictionary<string, PromptExecutionSettings>();
+            foreach (var settings in executionSettings)
+            {
+                var targetServiceId = settings.ServiceId ?? PromptExecutionSettings.DefaultServiceId;
+                if (newExecutionSettings.ContainsKey(targetServiceId))
+                {
+                    var exceptionMessage = (targetServiceId == PromptExecutionSettings.DefaultServiceId)
```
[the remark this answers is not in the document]
RogerBarreto: Followed @markwallace-microsoft suggestion on this one, prior was using something on those lines 😊

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
dmytrostruk: > On your scenario this would only be a breaking change if the specialized Dictionary don't respect the Liskov Substitution Principle (Intentionally changing the expected Dictionary behavior) which also is a anti-pattern and ensuring code quality from the caller side.

@RogerBarreto I think on our side we cannot be sure if specific SOLID principle or other pattern is respected on user side or not. But from the usage perspective, it may be confusing when I set specialized Dictionary (and I can do that, because public API allows me to do so), but it will be ignored and `System.Collections.Generic.Dictionary` will be used instead in all cases.

Imagine there is a scenario, when every time the new setting is added to the dictionary, user wants to verify if prompt execution settings are valid and log warning/throw exception if not. Today, it's possible to do that by implementing dictionary abstraction and set that extended implementation to `KernelArguments.ExecutionSettings`. But with new change, we will override it and user-defined custom validation and logging will be ignored. Just want to make sure if this sounds acceptable, maybe it's worth to communicate that through PR description and/or release notes.

In this case, it feels natural to convert `ExecutionSettings` property type to `Dictionary<string, PromptExecutionSettings>` (like in `PromptTemplateConfig.cs` for example), so we explicitly say that custom dictionary is not acceptable here, but we can't do that, because it will be a breaking change on compilation time. By modifying setter to always create `Dictionary`, this is breaking change in runtime, which may be even harder to troubleshoot. 

I understand the reason why it's implemented in this way, I just want to make sure that we understand the consequences.

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -27,6 +27,22 @@ public class PromptExecutionSettings
     /// </remarks>
     public static string DefaultServiceId => "default";
 
+    /// <summary>
+    /// Service identifier.
+    /// This identifies the service these settings are configured for e.g., azure_openai_eastus, openai, ollama, huggingface, etc.
+    /// </summary>
+    [JsonPropertyName("service_id")]
+    public string? ServiceId
```
dmytrostruk, line 35: Do we need to mark new properties and methods in this PR as `Experimental` for some period of time?

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -26,57 +27,92 @@ public async Task RunAsync()
                 serviceId: "OpenAIChat")
             .Build();
 
-        await RunByServiceIdAsync(kernel, "AzureOpenAIChat");
-        await RunByModelIdAsync(kernel, TestConfiguration.OpenAI.ChatModelId);
-        await RunByFirstModelIdAsync(kernel, "gpt-4-1106-preview", TestConfiguration.AzureOpenAI.ChatModelId, TestConfiguration.OpenAI.ChatModelId);
+        // Preconfigured function settings
+        await PreconfiguredFunctionSettingsByFirstModelIdAsync(kernel, ["gpt-4-1106-preview", TestConfiguration.AzureOpenAI.ChatModelId, TestConfiguration.OpenAI.ChatModelId]);
```
markwallace-microsoft, line 31: The names of these methods don't describe what each function does. From reading the code it looks like this method takes a list of modelIds and then executes the prompt with the first matching service. So a name like `InvokePromptWithFirstMatchingModelIdAsync` would be clearer. Each function should also have a comment explaining what it does.

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -26,57 +27,92 @@ public async Task RunAsync()
...
+        await PreconfiguredFunctionSettingsByFirstServiceIdAsync(kernel, ["NotFound", "AzureOpenAIChat", "OpenAIChat"]);
+        await PreconfiguredFunctionSettingsByModelIdAsync(kernel, TestConfiguration.OpenAI.ChatModelId);
+        await PreconfiguredFunctionSettingsByServiceIdAsync(kernel, "AzureOpenAIChat");
+
+        // Per invocation settings
+        await InvocationSettingsByServiceIdAsync(kernel, "AzureOpenAIChat");
```
markwallace-microsoft, line 37: I think `InvokePromptByServiceIdAsync` would be clearer. Again a comment for each of these functions explaining what it does.

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -7,6 +7,7 @@ namespace ChatCompletion;
 
 public class Connectors_WithMultipleLLMs(ITestOutputHelper output) : BaseTest(output)
 {
+    private const string ChatPrompt = "Hello AI, what can you do for me?";
```
markwallace-microsoft, line 10: The comment for this method needs to be updated

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
markwallace-microsoft: Also consider making each of the functions into it's own `Fact` so that they can be run is isolation

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
RogerBarreto: This is a double-edged sword, keep the original reference and change it or not. The only way I see to keep the original reference is to avoid auto-set `ServiceId` information and consider that value as optional for every case and not representing at all what is the current Key in the dictionary... not sure what impact this will have but seems to be a more simplistic approach... I will investigate that.

RogerBarreto: @dmytrostruk After discussing with the team and valid considerations I did some changes to this PR and dropped the approach on `Auto-Setting` the `ServiceId` property internally which was the whole reason as well as the problem of `safe` copying/cloning. 

As a solution I just made this new property `Optional` allowing just to use it when provided and while keeping the reference as is.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -65,7 +90,30 @@ public KernelArguments(IDictionary<string, object?> source, Dictionary<string, P
...
+    /// When setting, the service id of each <see cref="PromptExecutionSettings"/> must match the key in the dictionary.
+    /// </remarks>
+    public IReadOnlyDictionary<string, PromptExecutionSettings>? ExecutionSettings
+    {
+        get => this._executionSettings;
+        set
+        {
+            if (this._executionSettings is { Count: > 0 })
+            {
+                foreach (var kv in this._executionSettings!)
```
dmytrostruk, line 104: Should this be performed on incoming `value` rather than existing `this._executionSettings`? Looks like it's the case in `PromptTemplateConfig.cs`:
```suggestion
            if (value is { Count: > 0 })
            {
                foreach (var kv in value!)
```

dmytrostruk (approved):

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
+        Console.WriteLine($"======== Model Ids: {string.Join(", ", modelIds)} ========");
 
+        var function = kernel.CreateFunctionFromPrompt(ChatPrompt, modelIds.Select((modelId, index) => new PromptExecutionSettings { ServiceId = $"service-{index}", ModelId = modelId }));
         var result = await kernel.InvokeAsync(function);
+
+        Console.WriteLine(result.GetValue<string>());
+    }
+
+    /// <summary>
+    /// Create a function with a predefined configuration to run for a specific model id.
```
markwallace-microsoft, line 125: How about this description:
```
Shows how to invoke a KernelFunction and specify the model id of the AI Service the function will use.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
+        Console.WriteLine($"======== Model Id: {modelId} ========");
+
+        var function = kernel.CreateFunctionFromPrompt(ChatPrompt);
+        var result = await kernel.InvokeAsync(function, new(new PromptExecutionSettings { ModelId = modelId }));
+
+        Console.WriteLine(result.GetValue<string>());
+    }
+
+    /// <summary>
+    /// Create a function with a predefined configuration to run for a specific service id.
```
markwallace-microsoft, line 141: How about this description:
```
Shows how to invoke a KernelFunction and specify the service id of the AI Service the function will use.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
 
-        var modelSettings = new Dictionary<string, PromptExecutionSettings>();
-        foreach (var modelId in modelIds)
-        {
-            modelSettings.Add(modelId, new PromptExecutionSettings() { ModelId = modelId });
-        }
-        var promptConfig = new PromptTemplateConfig(prompt) { Name = "HelloAI", ExecutionSettings = modelSettings };
+    /// <summary>
+    /// Create a function with a predefined configuration to preferably run for a list specific model ids where the
+    /// first model id that is found respecting the order of the options provided will be used.
```
markwallace-microsoft, line 108: How about this description:
```
Shows how to create a KernelFunction from a prompt and specify the model ids of the preferred AI services.
When the function is invoked the AI Service will be selected based on the order of the provided model ids.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
         Console.WriteLine($"======== Model Ids: {string.Join(", ", modelIds)} ========");
 
-        var prompt = "Hello AI, what can you do for me?";
+        var result = await kernel.InvokePromptAsync(ChatPrompt, new(modelIds.Select((modelId, index) => new PromptExecutionSettings { ServiceId = $"service-{index}", ModelId = modelId })));
+
+        Console.WriteLine(result.GetValue<string>());
+    }
+
+    /// <summary>
+    /// Create a function with a predefined configuration and invoke at later moment.
```
markwallace-microsoft, line 91: How about this description:
```
Shows how to create a KernelFunction from a prompt and specify the service ids of the preferred AI services.
When the function is invoked the AI Service will be selected based on the order of the provided service ids.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
-           {
-               ModelId = modelId
-           }));
         Console.WriteLine(result.GetValue<string>());
     }
 
-    private async Task RunByFirstModelIdAsync(Kernel kernel, params string[] modelIds)
+    /// <summary>
+    /// Invoke the prompt function to preferably run for a list of specific model ids where the
+    /// first model id that is found respecting the order of the options provided will be used.
```
markwallace-microsoft, line 76: How about this description:
```
Shows how to invoke a prompt and specify the model ids of the preferred AI services.
When the prompt is executed the AI Service will be selected based on the order of the provided model ids.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
 
-        var prompt = "Hello AI, what can you do for me?";
+        var result = await kernel.InvokePromptAsync(ChatPrompt, new(new PromptExecutionSettings() { ModelId = modelId }));
+
+        Console.WriteLine(result.GetValue<string>());
+    }
+
+    /// <summary>
+    /// Invoke the prompt function to preferably run for a list specific service ids where the
+    /// first service id that is found respecting the order of the options provided will be used.
```
markwallace-microsoft, line 60: How about this description:
```
Shows how to invoke a prompt and specify the service ids of the preferred AI services.
When the prompt is executed the AI Service will be selected based on the order of the provided service ids.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
+                    .AddOpenAIChatCompletion(
+                        modelId: TestConfiguration.OpenAI.ChatModelId,
+                        apiKey: TestConfiguration.OpenAI.ApiKey,
+                        serviceId: "OpenAIChat")
+                    .Build();
     }
 
-    private async Task RunByServiceIdAsync(Kernel kernel, string serviceId)
+    /// <summary>
+    /// Invoke the prompt function to run for a specific service id.
```
markwallace-microsoft, line 28: How about this description:
```
Shows how to invoke a prompt and specify the service id of the preferred AI service.
When the prompt is executed the AI Service with the matching service id will be selected.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
```diff
@@ -1,82 +1,156 @@
...
-        {
-            { serviceId, new PromptExecutionSettings() }
-        };
-        var result = await kernel.InvokePromptAsync(prompt, arguments);
         Console.WriteLine(result.GetValue<string>());
     }
 
-    private async Task RunByModelIdAsync(Kernel kernel, string modelId)
+    /// <summary>
+    /// Invoke the prompt function to run for a specific model id.
```
markwallace-microsoft, line 44: How about this description:
```
Shows how to invoke a prompt and specify the model id of the preferred AI service.
When the prompt is executed the AI Service with the matching model id will be selected.
```

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
markwallace-microsoft: Can you also add an example showing what happens if the preferred service is not available (for both the service id and model id cases)

dotnet/src/SemanticKernel.Abstractions/AI/PromptExecutionSettings.cs
```diff
@@ -27,6 +28,26 @@ public class PromptExecutionSettings
     /// </remarks>
     public static string DefaultServiceId => "default";
 
+    /// <summary>
+    /// Service identifier.
+    /// This identifies the service these settings are configured for e.g., azure_openai_eastus, openai, ollama, huggingface, etc.
+    /// </summary>
+    /// <remarks>
+    /// When provided, this service identifier will be the key in a dictionary collection of execution settings for both <see cref="KernelArguments"/> and <see cref="PromptTemplateConfig"/>.
```
markwallace-microsoft, line 36: Also example what happens when this value is not provided.

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -36,12 +37,36 @@ public KernelArguments()
...
         {
-            this.ExecutionSettings = new Dictionary<string, PromptExecutionSettings>() { { PromptExecutionSettings.DefaultServiceId, executionSettings } };
+            var newExecutionSettings = new Dictionary<string, PromptExecutionSettings>();
+            foreach (var settings in executionSettings)
+            {
+                var targetServiceId = settings.ServiceId ?? PromptExecutionSettings.DefaultServiceId;
+                if (newExecutionSettings.ContainsKey(targetServiceId))
+                {
+                    var exceptionMessage = (targetServiceId == PromptExecutionSettings.DefaultServiceId)
+                        ? $"Default service id '{PromptExecutionSettings.DefaultServiceId}' must not be duplicated."
```
markwallace-microsoft, line 60: This message will also be used if the developer doesn't specify a service id. So this message could be confusing if I didn't explicitly set default so we should have a message something like.

```
Multiple prompt execution settings with the default service id '{...}' or no service id have been provided. Specify a single default prompt execution settings and provide a unique service id for all other instances.
```

dotnet/src/SemanticKernel.Abstractions/Functions/KernelArguments.cs
```diff
@@ -36,12 +37,36 @@ public KernelArguments()
...
+                        : $"Service id '{settings.ServiceId}' must not be duplicated and should match the key '{targetServiceId}'.";
```
markwallace-microsoft, line 61: How about 
```
Multiple prompt execution settings with the service id '{...}' have been provided. Provide a unique service id for all instances.
```

dotnet/src/Experimental/Orchestration.Flow/FlowOrchestrator.cs
```diff
@@ -73,6 +73,6 @@ public class FlowOrchestrator
         }
 
         var executor = new FlowExecutor(this._kernelBuilder, this._flowStatusProvider, this._globalPluginCollection, this._config);
-        return await executor.ExecuteFlowAsync(flow, sessionId, input, kernelArguments ?? new KernelArguments(null)).ConfigureAwait(false);
+        return await executor.ExecuteFlowAsync(flow, sessionId, input, kernelArguments ?? new KernelArguments()).ConfigureAwait(false);
```
markwallace-microsoft, line 76: Is this change required or optional?

dotnet/samples/Concepts/ChatCompletion/Connectors_WithMultipleLLMs.cs
RogerBarreto: Not sure about this one, as we don't use samples for showing wrong paths. I can add an assert and create a new sample for an undesired outcome though.

dotnet/src/Experimental/Orchestration.Flow/FlowOrchestrator.cs
RogerBarreto: Required

markwallace-microsoft (approved):

Issue #6404, opened 2024-05-27
.Net: Terminate auto function invocation with streaming does not return last function response


AssafYo:
When calling termination of automatic function invocation, we've noticed an inconsistency between the results of GetChatMessageContentAsync and GetStreamingChatMessageContentsAsync.
GetChatMessageContentAsync is functioning as expected and returns the last function call result. However, GetStreamingChatMessageContentsAsync does not include the last function call result in its response.

You could consider adding something like: 

"yield return new OpenAIStreamingChatMessageContent(chat.Last().Role, chat.Last().Content);" 

before the:
"yield break;" 

https://github.com/microsoft/semantic-kernel/blob/65515468f8aba286ed012d0407e8b56b877e2a14/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs#L870

markwallace-microsoft:
We need to ensure this behaviour is consistent across Google and Mistral connectors also

HillaPeter:
@SergeyMenshykh can u update when you will release a new version with this fix?

SergeyMenshykh:
Hi @HillaPeter, we plan to create a new release early next week.

SergeyMenshykh:
Hi @HillaPeter, the fix has been released and available in the latest version of SK - https://www.nuget.org/packages/Microsoft.SemanticKernel/1.14.0

PR #6428, opened 2024-05-28
.Net: Return result of the function executed before termination for streaming API

SergeyMenshykh:
Fixes: https://github.com/microsoft/semantic-kernel/issues/6404

Today, the SK chat completion streaming API does not return the result of a function executed before termination, whereas the non-streaming version does return the result. This PR resolves this issue by returning the result of a function executed before termination was requested.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
+++ b/dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
@@ -859,7 +859,7 @@ internal abstract class ClientCore
 
                 AddResponseMessage(chatOptions, chat, streamedRole, toolCall, metadata, stringResult, errorMessage: null, this.Logger);
 
-                // If filter requested termination, breaking request iteration loop.
+                // If filter requested termination, returning latest function result and breaking request iteration loop.
                 if (invocationContext.Terminate)
                 {
                     if (this.Logger.IsEnabled(LogLevel.Debug))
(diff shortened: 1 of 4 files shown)


dmytrostruk (approved):
@SergeyMenshykh Just for the context, I decided to not return function result in streaming termination scenario on purpose, because I wasn't sure if returning entire result as one chunk is a right approach, and function result is still accessible via `ChatHistory` input parameter, when using `IChatCompletionService`.

Based on created issue, it looks like this is expected behavior by developers, so it's probably worth to add it. Thanks for adding this!

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -859,14 +859,17 @@ static void AddResponseMessage(ChatCompletionsOptions chatOptions, ChatHistory c
...
                 {
                     if (this.Logger.IsEnabled(LogLevel.Debug))
                     {
                         this.Logger.LogDebug("Filter requested termination of automatic function invocation.");
                     }
 
+                    var lastChatMessage = chat.Last();
+
+                    yield return new OpenAIStreamingChatMessageContent(lastChatMessage.Role, lastChatMessage.Content);
                     yield break;
```
RogerBarreto, line 873: Unreachable code.
```suggestion
```

RogerBarreto (approved):

PR #6449, opened 2024-05-30
.Net: Support function call content classes for AI streaming API

SergeyMenshykh:
This PR is the second and last one in the series of PRs to leverage `FunctionCallContent` and `FunctionResultClass` classes to represent function calls and function results in a service/connector/model agnostic way. The first PR: [.Net: Function call content types](https://github.com/microsoft/semantic-kernel/pull/5800) added the classes and updated the **non-streaming** AI API to use them. This PR updates the **streaming** API to utilize the classes.

Related ADR: [Function call content](https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0041-function-call-content.md)
Related issue: [.Net: Support FunctionCallContent for streaming](https://github.com/microsoft/semantic-kernel/issues/5818)

Note: More tests and examples will be added in scope of this PR.

### Contribution Checklist


dotnet/src/SemanticKernel.Abstractions/Contents/StreamingFunctionCallUpdateContent.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// <summary>
+    /// The function name that can come as full or partial.
+    /// </summary>
+    public string? Name { get; }
+
+    /// <summary>
+    /// The function arguments that can come as full or partial.
+    /// </summary>
+    public string? Arguments { get; }
```
eavanvalkenburg, line 26: I'm doubting if this is needed at all, in python we store the arguments as string in the FunctionCallContent class and then only when we want to use them do we parse to a dict/KernelArguments, and we have a update method (add) to combine mulitple items together for streaming, isn't that a bit simpler?

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -17,9 +19,53 @@ namespace Microsoft.SemanticKernel;
...
+            }
+        }
+    }
+
+    /// <summary>
+    /// Chat message content items.
+    /// </summary>
+    [JsonIgnore]
+    [Experimental("SKEXP0001")]
+    public StreamingKernelContentItemCollection Items
```
dmytrostruk, line 64: Can we add an example how this property should be used?

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContentBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Text;
+using System.Text.Json;
+
+namespace Microsoft.SemanticKernel;
+
+/// <summary>
+/// A builder class for creating <see cref="FunctionCallContent"/> objects from incremental function call updates represented by <see cref="StreamingFunctionCallUpdateContent"/>.
+/// </summary>
+[Experimental("SKEXP0001")]
+public sealed class FunctionCallContentBuilder
```
dmytrostruk, line 16: Same here, we should add an example that shows how to use this class.

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAIToolsTests.cs
```diff
@@ -460,6 +460,253 @@ public async Task ConnectorAgnosticFunctionCallingModelClassesCanBeUsedForAutoFu
...
+            var textContent = new StringBuilder();
+
+            await foreach (var streamingContent in sut.GetStreamingChatMessageContentsAsync(chatHistory, settings, kernel))
+            {
+                textContent.Append(streamingContent.Content);
+                authorRole ??= streamingContent.Role;
+                fccBuilder.Add(streamingContent);
+            }
+
+            var functionCalls = fccBuilder.Build().ToArray();
```
RogerBarreto, line 492: I would change the builder to be `IAsyncEnumerable` and used like this:

Preferably this could be an extension method to a Collection of `FunctionCallContent` 

i.e.: 

```csharp
public FunctionCallContentCollection : IEnumerable<FunctionCallContent> {}

public static class FunctionCallContentCollectionExtensions {
    public static IAsyncEnumerable<StreamingChatMessageContent> AddFromStreamingChatMessageContents(this FunctionCallContentCollection 
 collection, IAsyncEnumerable<StreamingChatMessageContent> extractFrom)
    //extraction logic here..
}


FunctionCallContentCollection functionCalls = [];
await foreach (var update in functionCalls.AddFromStreamingChatContents(sut.GetStreamingChatMessageContentsAsync(chatHistory, settings, kernel)) 
{ 
    textContent.Append(streamingContent.Content);
    authorRole ??= streamingContent.Role;
}

```

```suggestion
            var fccBuilder = new FunctionCallContentBuilder(sut.GetStreamingChatMessageContentsAsync(chatHistory, settings, kernel));
            var textContent = new StringBuilder();

            await foreach (var streamingContent in fccBuilder)
            {
                textContent.Append(streamingContent.Content);
                authorRole ??= streamingContent.Role;
            }

            var functionCalls = fccBuilder.FunctionCalls();
```

SergeyMenshykh: It's a valid option, but I would like to understand its benefits before proceeding with it. In comparison to the option implemented in this PR that uses only one class (builder) for collecting function bits and pieces, the suggested option adds three classes: a builder that is IAsyncEnumerable, a function content collection, and the collection extension method(s) class to SK public API surface. I'm hesitant to go with it unless it provides additional benefits.

On a side note, I originally considered a similar design/option where there would be just an extension method, like this: `... IAsyncEnumerable<StreamingChatMessageContent> GetFunctionCalls(this IAsyncEnumerable<StreamingChatMessageContent> content, out FunctionCallContent[] functionCalls)` that could be elegantly used like this:
```csharp
FunctionCallContent[]? functionCalls = null;

await foreach (var streamingContent in sut.GetStreamingChatMessageContentsAsync(chatHistory, settings, kernel).GetFunctionCalls(out functionCalls))
{
    textContent.Append(streamingContent.Content);
    authorRole ??= streamingContent.Role;
}

if (functionCalls.Any())
{
    ...
}
```
However, C# does not allow having out parameters for iterators and async methods, resulting in CS1988 (Async methods cannot have ref, in, or out parameters) and CS1623 (Iterators cannot have ref, in, or out parameters).

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingFunctionCallUpdateContent.cs
SergeyMenshykh: As far as I understand, in SK python, `FunctionCallContent` is used for both streaming and non-streaming scenarios. In SK .NET, this is not the case.

SK .NET has two distinct hierarchies of content classes: one for streaming content, represented by the `StreamingKernelContent` base class, and the other for non-streaming content, represented by `KernelContent`. Content classes from each hierarchy are designed to represent not only the specifics of the content - such as image, audio, or function - but also the specifics of the way the content is communicated to API callers - streaming or non-streaming. Content classes belonging to the streaming hierarchy represent partial content, updates, or chunks that need to be combined together to create full content. Conversely, non-streaming classes represent fully baked content that doesn't need to be assembled together and can be used as is.

Back to the `FunctionCallContent`. The class inherits the `KernelContent` class and as such belongs to the non-streaming hierarchy and represents a fully-fledged function call. To use it for streaming purposes as well as for non-streaming ones, it would need to be inherited from the `StreamingKernelContent` base class as well. .NET doesn't allow multiple inheritance and as such it's not possible to have one class belonging to two hierarchies. 

Even if it were possible to have `FunctionCallContent` inherit two classes, it would have to be redesigned because its public `InvokeAsync` method would not work in a way callers might have expected if it was called on partial/streaming content.

eavanvalkenburg: @SergeyMenshykh  thanks for the clarification, the big difference is the fact that in python we can inherit from multiple classes, so we have chosen long ago to base everything off `KernelContent` and we created the `StreamingContentMixin` to create the classes that come from streaming api's. This makes sure we have similar behavior when looking at something like StreamingTextContent vs TextContent and allows our type checking code to compare something to TextContent where it can then be either Streaming or not. So good to understand why we have this difference and also means there is no need at this time to add StreamingFunctionCallContent to Python.

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
SergeyMenshykh: Yes, I have added them. Please check OpenAI_ChatCompletionStreaming.StreamTextContentAsync and OpenAI_ChatCompletionStreaming.StreamFunctionCallContentAsync.

dotnet/src/SemanticKernel.Abstractions/Contents/FunctionCallContentBuilder.cs
SergeyMenshykh: Added, please check - OpenAI_FunctionCalling.RunStreamingChatAPIWithManualFunctionCallingAsync.

RogerBarreto (approved):

dotnet/samples/Concepts/ChatCompletion/OpenAI_ChatCompletionStreaming.cs
```diff
@@ -1,27 +1,36 @@
...
 namespace ChatCompletion;
 
-// The following example shows how to use Semantic Kernel with streaming Chat Completion
+/// <summary>
+/// These examples demonstrate the ways different content types are streamed by OpenAI LLM via the chat completion service.
+/// </summary>
 public class OpenAI_ChatCompletionStreaming(ITestOutputHelper output) : BaseTest(output)
 {
+    /// <summary>
+    ///This example demonstrates how the chat completion service streams chat using the OpenAI LLM.
```
markwallace-microsoft, line 15: /// This example demonstrates chat completion streaming using OpenAI.

dotnet/samples/Concepts/ChatCompletion/OpenAI_ChatCompletionStreaming.cs
```diff
@@ -1,27 +1,36 @@
...
-        Console.WriteLine("======== Open AI - ChatGPT Streaming ========");
+        Console.WriteLine("======== Open AI Chat Completion Streaming ========");
 
         OpenAIChatCompletionService chatCompletionService = new(TestConfiguration.OpenAI.ChatModelId, TestConfiguration.OpenAI.ApiKey);
 
         return this.StartStreamingChatAsync(chatCompletionService);
     }
 
+    /// <summary>
+    ///This example demonstrates how the chat completion service streams chat using the Azure OpenAI LLM.
```
markwallace-microsoft, line 28: /// This example demonstrates chat completion streaming using Azure OpenAI.

dotnet/samples/Concepts/ChatCompletion/OpenAI_ChatCompletionStreaming.cs
```diff
@@ -32,24 +41,98 @@ public Task AzureOpenAIChatStreamSampleAsync()
         return this.StartStreamingChatAsync(chatCompletionService);
     }
 
+    /// <summary>
+    /// This example demonstrates how the chat completion service streams text content.
+    /// It shows how to access the response update via StreamingChatMessageContent.Content property
+    /// and alternatively via the StreamingChatMessageContent.Items property.
+    /// </summary>
+    [Fact]
+    public async Task StreamTextContentAsync()
```
markwallace-microsoft, line 50: Why is this sample structured differently from the previous two?

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -17,9 +18,52 @@ namespace Microsoft.SemanticKernel;
...
+        get
+        {
+            var textContent = this.Items.OfType<StreamingTextContent>().FirstOrDefault();
+            return textContent?.Text;
+        }
+        set
+        {
+            if (value is null)
+            {
+                return;
```
markwallace-microsoft, line 34: Getting returns the first item but setting to null does nothing, is that what we want?


If I had only one item in the list should we remove it?

dotnet/src/SemanticKernel.Abstractions/AI/ChatCompletion/StreamingKernelContentItemCollection.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// Removes the content item at the specified index from the collection.
+    /// </summary>
+    /// <param name="index">The index of the content item to remove.</param>
+    public void RemoveAt(int index) => this._items.RemoveAt(index);
+
+    bool ICollection<StreamingKernelContent>.IsReadOnly => false;
+
+    IEnumerator IEnumerable.GetEnumerator() => this._items.GetEnumerator();
+
+    IEnumerator<StreamingKernelContent> IEnumerable<StreamingKernelContent>.GetEnumerator() => this._items.GetEnumerator();
```
markwallace-microsoft, line 132: Do we need API documentation for the above?

SergeyMenshykh: We do, thanks, added.

dotnet/samples/Concepts/ChatCompletion/OpenAI_ChatCompletionStreaming.cs
SergeyMenshykh: Thanks, fixed.

dotnet/samples/Concepts/ChatCompletion/OpenAI_ChatCompletionStreaming.cs
SergeyMenshykh: Thanks, fixed.

dotnet/samples/Concepts/ChatCompletion/OpenAI_ChatCompletionStreaming.cs
SergeyMenshykh: The previous two examples show the streaming concept for both OpenAI and AzureOpenAI connectors, even though the streaming code is 99% the same. This sample and the next one demonstrate streaming text and function content without repeating it twice—once per connector. I'm happy to add four more methods: two public ones as tests and two private ones shared by the public methods. Alternatively, we can restructure the samples above.

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
SergeyMenshykh: > If I had only one item in the list should we remove it?

Good catch. I fixed it slightly differently. I don't remove the item but just update the first item's text/content to null if null is assigned to the property so both the setter and getter have consistent behavior. I'll also register a task to handle a similar case in the `ChatMessageContent` class.

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -55,12 +121,15 @@ public StreamingChatMessageContent(AuthorRole? role, string? content, object? in
...
     }
 
     /// <inheritdoc/>
     public override string ToString() => this.Content ?? string.Empty;
 
     /// <inheritdoc/>
     public override byte[] ToByteArray() => this.Encoding.GetBytes(this.ToString());
+
+    private StreamingKernelContentItemCollection? _items;
+    private Encoding _encoding = Encoding.UTF8;
```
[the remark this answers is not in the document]
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingFunctionCallUpdateContent.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <inheritdoc />
+    public override string ToString()
+    {
+        return nameof(StreamingFunctionCallUpdateContent);
+    }
+
+    /// <inheritdoc />
+    public override byte[] ToByteArray()
+    {
+        return [];
```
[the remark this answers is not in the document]
SergeyMenshykh: fixed

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -37,7 +81,29 @@ public class StreamingChatMessageContent : StreamingKernelContent
...
+                return textContent.Encoding;
+            }
+
+            return this._encoding;
+        }
+        set
+        {
+            this._encoding = value;
+
+            var textContent = this.Items.OfType<StreamingTextContent>().FirstOrDefault();
```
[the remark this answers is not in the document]
SergeyMenshykh: Originally, StreamingChatMessageContent could work with only one piece of content - a string represented by the Content property, which could be encoded according to the Encoding property. Both the Content and Encoding properties were related to one piece of content: a string.

Now, with the addition of the Items property, the Content and Encoding properties have become convenient shortcuts to the first text element in the Items collection. Despite the change in their internal implementation, their semantics have been preserved - they still work with only one piece of text content.

So, I would keep Encoding as a convenient shortcut to preserve its semantics and align it with the Content property.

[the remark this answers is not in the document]
SergeyMenshykh: I've updated the property XML comment to say that it's a shortcut.

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
```diff
@@ -17,9 +18,52 @@ namespace Microsoft.SemanticKernel;
...
+            if (value is null)
+            {
+                return;
+            }
+
+            var textContent = this.Items.OfType<StreamingTextContent>().FirstOrDefault();
+            if (textContent is not null)
+            {
+                textContent.Text = value;
+                textContent.Encoding = this.Encoding;
```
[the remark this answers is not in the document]
SergeyMenshykh: I'm not sure I follow. Both the Content and Encoding properties are shortcuts to a first text content item. As either changes, the first text content item is changed as well, correspondingly, to preserve the properties' semantics as described here - https://github.com/microsoft/semantic-kernel/pull/6449#discussion_r1633893892

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1391,58 +1349,106 @@ private OpenAIChatMessageContent GetChatMessage(ChatChoice chatChoice, ChatCompl
...
+
+        if (toolCall is ChatCompletionsFunctionToolCall functionCall)
         {
-            InnerContent = toolCall,
-            Exception = exception
-        };
+            // Add an item of type FunctionResultContent to the ChatMessageContent.Items collection in addition to the function result stored as a string in the ChatMessageContent.Content property.  
+            // This will enable migration to the new function calling model and facilitate the deprecation of the current one in the future.
+            var functionName = FunctionName.Parse(functionCall.Name, OpenAIFunction.NameSeparator);
+            message.Items.Add(new FunctionResultContent(functionName.Name, functionName.PluginName, functionCall.Id, result));
```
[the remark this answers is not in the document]
SergeyMenshykh: Neither of those was ever added to either the parent `ChatMessageContent` instance or the `FunctionResultContent` instance. They can be added later if/when needed.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1391,58 +1349,106 @@ private OpenAIChatMessageContent GetChatMessage(ChatChoice chatChoice, ChatCompl
...
+        result ??= errorMessage ?? string.Empty;
+        chatOptions.Messages.Add(new ChatRequestToolMessage(result, toolCall.Id));
 
-        return new FunctionCallContent(
-            functionName: functionName.Name,
-            pluginName: functionName.PluginName,
-            id: toolCall.Id,
-            arguments: arguments)
+        // Add the tool response message to the chat history.
+        var message = new ChatMessageContent(role: AuthorRole.Tool, content: result, metadata: new Dictionary<string, object?> { { OpenAIChatMessageContent.ToolIdProperty, toolCall.Id } });
```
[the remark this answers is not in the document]
SergeyMenshykh: It was never there, nor was metadata. Similar to - https://github.com/microsoft/semantic-kernel/pull/6449#discussion_r1634441002

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
```diff
@@ -1391,58 +1349,106 @@ private OpenAIChatMessageContent GetChatMessage(ChatChoice chatChoice, ChatCompl
     {
         var message = new OpenAIChatMessageContent(chatChoice.Message, this.DeploymentOrModelName, GetChatChoiceMetadata(responseData, chatChoice));
 
-        foreach (var toolCall in chatChoice.Message.ToolCalls)
+        message.Items.AddRange(this.GetFunctionCallContent(chatChoice.Message.ToolCalls));
```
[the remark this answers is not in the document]
SergeyMenshykh: Fixed, thanks.

dotnet/samples/Concepts/ChatCompletion/OpenAI_ChatCompletionStreaming.cs
SergeyMenshykh: Decided to resolve it later.

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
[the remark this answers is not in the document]
SergeyMenshykh: Discussed offline and agreed to keep as is.

dotnet/src/SemanticKernel.Abstractions/Contents/StreamingChatMessageContent.cs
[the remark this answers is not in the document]
SergeyMenshykh: Discussed offline and agreed not to set encoding when updating the text. Fixed.

dotnet/src/IntegrationTests/Connectors/OpenAI/OpenAIToolsTests.cs
SergeyMenshykh: Discussed offline and decided to keep the builder.

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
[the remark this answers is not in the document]
SergeyMenshykh: Discussed offline and agreed to track the issue separately: https://github.com/microsoft/semantic-kernel/issues/6665

dotnet/src/Connectors/Connectors.OpenAI/AzureSdk/ClientCore.cs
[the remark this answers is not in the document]
SergeyMenshykh: Discussed offline and agreed to track the issue separately: https://github.com/microsoft/semantic-kernel/issues/6665

markwallace-microsoft (approved):

Issue #6448, opened 2024-05-30
.Net: AzureOpenAIChatCompletionWithDataService is using the wrong Azure OpenAI URL


Cobra86:
I'm trying to test the AzureOpenAIChatCompletionWithDataService, I am receiving an "Error occurred on chat completion with data request execution: Response status code does not indicate success: 404 (resource Not Found)" when trying to access the Azure OpenAI. It seems to be generating the wrong URL.
`openai.azure.com/openai/deployments/gpt-4-model/extensions/chat/completions?api-version=2024-03-01-preview`

I can confirm the resource  is working with Microsoft.SemanticKernel.Core, also i was able to change the api version as the default one is 2024-02-01

**Platform**
 - OS: Window 11
 - IDE: Visual Studio
 - Language: C#
 - Source: Microsoft.SemanticKernel.Connectors.OpenAI 1.13.0

Issue #6112, opened 2024-05-03
.Net: Update the Chat Completion With Data


markwallace-microsoft:
- [ ] Change implementation to use the Azure OpenAI rather than a custom client
- [ ] Fix the outstanding issues listed below

Chat completion "with data" support in Azure .NET SDK:
https://github.com/Azure/azure-sdk-for-net/tree/49b0ace48d35cb4b2ace93b517036b660fc12ca4/sdk/openai/Azure.AI.OpenAI#use-your-own-data-with-azure-openai

Chat completion "with data" issues:

- [.Net: Unable to configure DataSource as much as a vanilla HTTP request to OpenAI · Issue #4787 · microsoft/semantic-kernel (github.com)](https://github.com/microsoft/semantic-kernel/issues/4787)
- [.Net: [BUG] When working with Semantic Kernel, while invoking ChatCompletion API, unable to specify data sources · Issue #6061 · microsoft/semantic-kernel (github.com)](https://github.com/microsoft/semantic-kernel/issues/6061)
- [.Net: AzureOpenAIChatCompletionWithDataService is using the wrong Azure OpenAI URL](https://github.com/microsoft/semantic-kernel/issues/6448)

Issue #6061, opened 2024-04-30
.Net: [BUG] When working with Semantic Kernel, while invoking ChatCompletion API, unable to specify data sources


jsquire:
# Issue Transfer

This issue has been transferred from the Azure SDK for .NET repository, [#43783](https://github.com/Azure/azure-sdk-for-net/issues/43783).

### Library name and version

Microsoft.SemanticKernel, Microsoft.SemanticKernel.ChatCompletion

### Describe the bug

When calling the OpenAI's chat completion API, the SDK doesn't provide an option to specify extra field such as "data_sources"

https://learn.microsoft.com/en-us/azure/ai-services/openai/references/on-your-data?tabs=rest#context

### Expected behavior

the SKD should support additional parameter to chat completions API

### Actual behavior

the SKD doesn't support additional parameter to chat completions API

### Reproduction Steps

```csharp
Microsoft.SemanticKernel.ChatCompletion.IChatCompletionService.GetStreamingChatMessageContentsAsync()
```

This method doesn't accept additional parameters to query data on custom data sources

### Environment

VS 2022
.NET 6.0
Windows

jsquire:
Thanks, @rupeshtech, but as stated, I am just the person who transferred this issue to the proper repository.   You'll want to discuss with @chetan-d-mahajan, the author of the original issue.

Rtoribiog:
hello @rupeshtech ,   thanks for the comment we are going to give a try and we will let you know.  @chetan-d-mahajan

chetan-d-mahajan:
Hi @rupeshtech 

Thanks for your comment.

I tried with your suggestion and it doesn't work, the response from OpenAI API is empty which means that it isn't able to connect & fetch the data from the Azure AI search index.

Seems like the .NET SKD doesn't support data source yet.
https://github.com/microsoft/semantic-kernel/issues/6112

<img width="391" alt="image" src="https://github.com/microsoft/semantic-kernel/assets/26963448/0e0ae393-c893-4cf1-8d2c-4fd1f470e2cb">

@Rtoribiog

Issue #4787, opened 2024-01-30
.Net: Unable to configure DataSource as much as a vanilla HTTP request to OpenAI


santanor:
**Describe the bug**
Semantic Kernel offers the ability to connect to a DataSource by using AzureOpenAIChatCompletionWithDataConfig I can't see a way to edit some of the parameters such as the `querytype`. 

```
        
        var config = new AzureOpenAIChatCompletionWithDataConfig()
        {
            CompletionEndpoint = openAiConf["Endpoint"] ?? string.Empty,
            CompletionApiKey = openAiConf["Key"] ?? string.Empty,
            CompletionApiVersion = openAiConf["ApiVersion"] ?? string.Empty,
            CompletionModelId = openAiConf["Deployment"] ?? string.Empty,
            DataSourceEndpoint = dataSourceConf["Endpoint"] ?? string.Empty,
            DataSourceIndex = dataSourceConf["IndexName"] ?? string.Empty,
            DataSourceApiKey = dataSourceConf["Key"] ?? string.Empty
        };
```

In a basic OpenAI request there are a lot of parameters that one can play around with, fields mapping, query type, topn.... 

```
"dataSources": [
    {
      "type": "AzureCognitiveSearch",
      "parameters": {
        "endpoint": "",
        "indexName": "",
        "semanticConfiguration": "default",
        "queryType": "semantic",
        "fieldsMapping": {
        "contentFieldsSeparator": "\n",
        "contentFields": [
            "content"
        ],
        "filepathField": "filepath",
        "titleField": "title",
        "urlField": "url",
        "vectorFields": [
            "contentVector"
        ]
        },
        "inScope": true,
        "roleInformation": "You are an AI assistant that helps people find information.",
        "filter": null,
        "strictness": 3,
        "topNDocuments": 2,
        "key": "",
        "embeddingDeploymentName": ""
      }
    }
```

I tried putting them in the `execution_settings` of the config.json, but no luck. 

Am I missing something? Any ideas?

**Expected behavior**
It would be nice to be able to edit those parameters in some way. 

**Platform**
 - OS: Mac
 - IDE: Rider
 - Language: C#
 - Source: NuGet 1.1.0

**Additional context**

dmytrostruk:
Hi @santanor, thank you for reporting this issue! `AzureOpenAIChatCompletionWithData` related classes will be obsolete soon, as we will replace it with functionality that is available in Azure .NET SDK. There is a possibility to set query type and other related properties:
[...]

PR #6454, opened 2024-05-31
.Net: Added AzureChatExtensionsOptions property and deprecated WithData classes

dmytrostruk:
### Motivation and Context

Fixes: https://github.com/microsoft/semantic-kernel/issues/4787
Fixes: https://github.com/microsoft/semantic-kernel/issues/6061
Fixes: https://github.com/microsoft/semantic-kernel/issues/6448
Fixes: https://github.com/microsoft/semantic-kernel/issues/6112

Discussion: https://github.com/microsoft/semantic-kernel/discussions/6016#discussioncomment-9459380

There are two reasons why this change is required:
1. Azure .NET SDK supports `chat completion on your data` functionality out-of-the-box, which means that now we don't need custom implementation.
2. `Chat completion on your data` new API has [breaking changes](https://learn.microsoft.com/en-us/azure/ai-services/openai/references/on-your-data), which makes current `WithData` classes non-functional. 

### Description

1. `AzureChatExtensionsOptions` property was added to `OpenAIPromptExecutionSettings`. Initial idea was to add it on `AzureOpenAIChatCompletionService` class level (in all constructors and extension methods), but it looks like this configuration contains some properties which should be useful per request (e.g. index name, search filter etc). Later, we can also add it on service class level, if needed (although I would do that when Options classes for all `AddChatCompletion...`  extension methods will be introduced, to avoid more parameters in these methods at the moment).
2. `AzureChatExtensionsOptions` property is marked with `JsonIgnore` attribute, because it's still needs to be investigated if this configuration is serializable. It supports not only Azure AI Search configuration, but also [other configurations](https://learn.microsoft.com/en-us/azure/ai-services/openai/references/on-your-data#data-source) like Azure Cosmos DB for MongoDB, Pinecone etc. We can release it without JSON serialization first, to unblock code-first customers, and we can add support for JSON serialization later (i.e. `config.json` files), if there will be such request.

Note: It looks like `OpenAIPromptExecutionSettings` is not ideal place for Azure-specific `AzureChatExtensionsOptions` property. I was thinking to add derived `AzureOpenAIPromptExecutionSettings : OpenAIPromptExecutionSettings` class (and unmark `OpenAIPromptExecutionSettings` as sealed), but this looks like some scope of work which should be implemented when OpenAI and Azure OpenAI functionality will be separated by SDKs. 

### Contribution Checklist


.github/workflows/dotnet-build-and-test.yml
```diff
@@ -81,7 +81,7 @@ jobs:
         run: |
           export UT_PROJECTS=$(find ./dotnet -type f -name "*.UnitTests.csproj" | grep -v -E "(Experimental.Orchestration.Flow.UnitTests.csproj|Experimental.Assistants.UnitTests.csproj)" | tr '\n' ' ')
           for project in $UT_PROJECTS; do
-            dotnet test -c ${{ matrix.configuration }} $project --no-build -v Normal --logger trx --collect:"XPlat Code Coverage" --results-directory:"TestResults/Coverage/" -- DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.ExcludeByAttribute=ObsoleteAttribute,GeneratedCodeAttribute,CompilerGeneratedAttribute,ExcludeFromCodeCoverageAttribute
+            dotnet test -c ${{ matrix.configuration }} $project --no-build -v Normal --logger trx --collect:"XPlat Code Coverage" --results-directory:"TestResults/Coverage/" -- DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.ExcludeByAttribute=GeneratedCodeAttribute,CompilerGeneratedAttribute,ExcludeFromCodeCoverageAttribute
```
markwallace-microsoft, line 84: Curious why this change is needed for this PR?

markwallace-microsoft (approved):

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
```diff
@@ -287,6 +287,23 @@ public int ResultsPerPrompt
         }
     }
 
+    /// <summary>
+    /// An abstraction of additional settings for chat completion, see https://learn.microsoft.com/en-us/dotnet/api/azure.ai.openai.azurechatextensionsoptions.
+    /// This property is compatible only with Azure OpenAI.
+    /// </summary>
+    [Experimental("SKEXP0010")]
+    [JsonIgnore]
+    public AzureChatExtensionsOptions? AzureChatExtensionsOptions
```
SergeyMenshykh, line 296: I wonder whether an abstraction layer over the `AzureChatExtensionsOptions` class was considered as an option to isolate service API consumers from internal implementation details?

.github/workflows/dotnet-build-and-test.yml
dmytrostruk: `Obsolete` attribute was added to code coverage filtering in recent PRs. As I marked bunch of classes as `Obsolete`, the code coverage decreased, but I think it's incorrect. Even though we marked some classes as `Obsolete`, they are still in-place, and somebody may use it. Which means that even if we do not recommend using these classes anymore, we still need to test them and include them into code coverage, especially if these classes are located in v1 assembly (without `alpha` or `preview` suffix). That's why I removed `Obsolete` attribute from code coverage filtering. Let me know if that makes sense.

stephentoub: >  As I marked bunch of classes as Obsolete, the code coverage decreased

You mean the percentage of covered lines/branches decreased? Were the classes you marked as obsolete better covered than the average? I'd have expected the WithData stuff to be, on average, less well covered, and thus I'd have expected marking them obsolete and hiding them from coverage would have actually increased percentage covered.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
dmytrostruk: Great question! I was thinking about it, but the number of possible configuration properties for this functionality is huge, here is just some of them:
image

I think it would be difficult to abstract these parameters and maintain them if anything changes in the future. I also think that abstraction could be useful if similar settings would exist in other connectors as well, but as for now it's Azure-specific settings, so I decided to expose `AzureChatExtensionsOptions` class directly, in similar way how we do that with `OpenAIClient` in Azure/OpenAI chat completion service constructors.

.github/workflows/dotnet-build-and-test.yml
dmytrostruk: > You mean the percentage of covered lines/branches decreased?

So, without `Obsolete` attribute branch coverage for OpenAI assembly is 80.4%. With `Obsolete` attribute it's 79.5%.

> Were the classes you marked as obsolete better covered than the average?

It could be the case. Which means that we can add more tests to other OpenAI functionality to increase code coverage in more important places. But I'm still not sure if we need to exclude `Obsolete` classes from code coverage report, since deprecated functionality is still in place.

dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs
SergeyMenshykh: I think this solution is good for now, considering the large number of classes that require abstraction to achieve the isolation we might never need. As you mentioned, let's revise the model later when we have more connectors using similar settings.

SergeyMenshykh (approved):

Issue #6451, opened 2024-05-30
Python: Kernel function arguments can handle lists only if the items are strings


stefan521:
**Describe the bug**
Version: 1.0.3

Kernel functions can handle `list[str]` but not lists of other types like `list[int]` or `list[Person]`

If the list contains more complex types then the metadata is inferred incorrectly.

**To Reproduce**
Define this plugin:
```python
from typing import Annotated
from semantic_kernel.functions import kernel_function
from semantic_kernel.kernel_pydantic import KernelBaseModel


class Person(KernelBaseModel):
    first_name: str
    last_name: str


class CreditScorePlugin:
    @kernel_function(name="CreditScore", description="Calculate the credit score.")
    def credit_score_breaks_1(
        self,
        people: Annotated[list[Person], "the people to calculate the credit score for"]
    ) -> float:
        return 9.3

```

The request sent to the Azure LLM contains this incorrect tool definition:
```json
{
    "type": "function",
    "function": {
        "name": "score-CreditScore",
        "description": "Calculate the credit score.",
        "parameters": {
            "type": "object",
            "properties": {
                "people": {
                    "type": "array",
                    "description": "the people to calculate the credit score for",
                    "items": {
                        "type": "string"
                    }
                }
            },
            "required": [ "people" ]
        }
    }
}
```

**Expected behavior**
I expect the tool definition to look like this:
```json
{
    "type": "function",
    "function": {
        "name": "score-CreditScore",
        "description": "Calculate the credit score.",
        "parameters": {
            "type": "object",
            "properties": {
                "people": {
                    "type": "array",
                    "description": "the people to calculate the credit score for",
                    "items": {
                        "type": "object",
                        "properties": {
                            "first_name": {
                                "type": "string"
                            },
                            "last_name": {
                                "type": "string"
                            }
                        }
                    }
                }
            },
            "required": [ "people" ]
        }
    }
}
```

**Screenshots**
NA

**Platform**
 - OS: [Mac]
 - IDE: [PyCharm]
 - Language: [Python]
 - Source: [version 1.0.3]

**Additional context**
The `KernelJsonSchemaBuilder` loses the type of the array items in `semantic_kernel.schema.kernel_json_schema_builder.KernelJsonSchemaBuilder.get_json_schema`

PR #6469, opened 2024-05-31
Python: properly handle types while building the schema

### LINKED ISSUE #6466: Python: Kernel Function Decorator loses the types of arguments during parsing
**Describe the bug**

While parsing the kernel function's arguments, we recurse to the type's base origin and then set that as the `type_object`. For a type like `list[MyCustomClass]` only setting the `type_object` as `list` has us lose out on the other valuable information of the list's type, which is `MyCustomClass` and what attributes it contains within. This then leads us to build an incomplete schema when creating the kernel function parameters and ultimately when we send the JSON schema to the model during tool calls.

moonbox3:
### Motivation and Context

While parsing the kernel function's arguments, we recurse to the type's base origin and then set that as the `type_object`. For a type like `list[MyCustomClass]` only setting the `type_object` as `list` has us lose out on the other valuable information of the list's type, which is `MyCustomClass` and what attributes it contains within. This then leads us to build an incomplete schema when creating the kernel function parameters and ultimately when we send the JSON schema to the model during tool calls.

### Description

This PR:
- Closes #6466 
- Refactors the kernel function decorator code to parse arguments/parameters. We now grab the `type_object` of the parameter/argument in question, which is used to build the JSON schema.
- Fixes the schema building for the types we build. We were missing `required` parameters as well.
- Attempts to handle as many edge cases as possible, although there are probably more that we aren't handling right now -- for example properly handling all schemas for Unions is tricky.
- Adds unit tests to make sure we are building the JSON schema as expected. All unit, integration and sample tests are passing.
- Handles a `TypeError` exception during auto function calling where the model sends back an argument like `2`, which doesn't cause JSON parsing to fail, but will fail when trying to update the argument dictionary.

### Contribution Checklist


python/semantic_kernel/connectors/ai/open_ai/services/utils.py
```diff
@@ -14,61 +16,48 @@
...
     return {
         "type": "function",
         "function": {
             "name": metadata.fully_qualified_name,
             "description": metadata.description or "",
             "parameters": {
                 "type": "object",
-                "properties": {param.name: parse_schema(param.schema_data) for param in metadata.parameters},
+                "properties": {
+                    param.name: param.schema_data for param in metadata.parameters
```
eavanvalkenburg, line 59: I guess this means that the schema data is fully compatible with openai, will that be universal enough? maybe check with the Sergey!

python/semantic_kernel/functions/kernel_function_decorator.py
```diff
@@ -78,7 +74,44 @@ def decorator(func: Callable[..., object]) -> Callable[..., object]:
...
+def _get_underlying_type(annotation: Any) -> Any:
+    """Get the underlying type of the annotation."""
+    if isinstance(annotation, types.UnionType):
+        args = annotation.__args__
+        non_none_types = [arg for arg in args if arg is not type(None)]
+        return non_none_types[0] if non_none_types else None
+    if hasattr(annotation, "__origin__"):
+        if annotation.__origin__ is Union:
+            args = get_args(annotation)
+            non_none_types = [arg for arg in args if arg is not type(None)]
```
eavanvalkenburg, line 86: I think you should be able to use NoneType for this (typing.NoneType)

eavanvalkenburg (approved):
Fine for now, but we do want to check how to handle different formats for new models like Mistral!

python/semantic_kernel/connectors/ai/open_ai/services/utils.py
moonbox3: This will need a bit of a re-vamp as we do expand away from OpenAI, yes.

juliomenendez (approved):

stefan521:
closes https://github.com/microsoft/semantic-kernel/issues/6451

